PR #143, opened 2015-10-26
Add Spring Cloud Config Server

gregturn:
Spring Cloud Config Server makes it possible to externalize and centralize property files with a single github repo.

For reference, see http://cloud.spring.io/spring-cloud-config/spring-cloud-config.html

To run this, you need to run a separate config server. I have an app at https://github.com/gregturn/config-server. I suggest moving/copying this into github.com/spinnaker and rebudding as "fig". You just run it as:

```
SPRING_CLOUD_CONFIG_SERVER_GIT_URI=https://github.com/<gregturn>/<spinnaker-config>  ./gradlew bootRun
```

(See mine at https://github.com/gregturn/spinnaker-config/)

This server process runs locally on port 8888. Then, when you fire up clouddriver locally,it will reach out to that process and consume property files found inside that github repo as if they were local. In production, you can override "localhost:8888" via FIG_URL to talk to a production version of this config server.

This makes it possible to have all your properties files in one place and outside the actual source code, yet still under version control of each user of Spinnaker.

P.S. There are options to secure the config server as well as encrypt secrets inside the github repo.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -84,6 +84,7 @@ subprojects { project ->
 
   dependencies {
     compile spinnaker.dependency('groovy')
+    compile 'org.springframework.cloud:spring-cloud-starter-config:1.0.3.RELEASE'
     spinnaker.group('test')
   }
 }
--- a/clouddriver-web/src/main/resources/bootstrap.yml
+++ b/clouddriver-web/src/main/resources/bootstrap.yml
@@ -1,0 +1,6 @@
+spring:
+  application:
+    name: clouddriver
+  cloud:
+    config:
+      uri: ${figUrl:http://localhost:8888}


gregturn:
@cfieber Are you up for considering making Spring Cloud Config Server an optional setting in the various modules?

The work I have on standby is here:

* https://github.com/gregturn/clouddriver/commit/6aee44e2f6e711f0686b0c34fc837ad144d520a7
* https://github.com/gregturn/echo/commit/a980073ec874d339ba73f144c22b1d48232faad8
* https://github.com/gregturn/front50/commit/111aa97dbb8d72904d7e14cde344e7a197b8bfd1
* https://github.com/gregturn/gate/commit/f11e351115e7ae6d0039f8d1286b20deeff93137
* https://github.com/gregturn/igor/commit/80a029a0892480bfc8afc68be5530a4d63c85cbc
* https://github.com/gregturn/orca/commit/b52a76500a5ba1d12091be6ef7d47e2af68a7ac9

This sets up each client to optional flip this on via -`Dspring.cloud.config.enabled=true` `-Dspring.cloud.config.uri=<gateUrl>/config` (except **gate**).

**gate** itself can be flipped on via `-Dspring.cloud.config.server.bootstrap=true` and point at a source via `-Dspring.cloud.config.server.git.uri=<github>`. **gate** would bootstrap itself via that repository, meaning it would need to startup first. From there, all the other process can launch.

By putting the config server inside **gate**, namespaced underneath `/config`, **deck** can optionally consume it as well due to existing CORS settings. @anotherchrisberry could help on that front.

If there is agreement, then I can turn these branches into pull requests.

Thoughts?

PR #185, opened 2015-11-16
Without these dependencies, everything compiles, unit tests run, and …

claymccoy:
…it runs locally.


ttomsu:
This PR appears to break my local build. Confirmed by `git checkout HEAD~1` and then `./gradlew test` and everything works again.

PR #186, opened 2015-11-17
manage package files consistently with other builds

cfieber:
cc: @dstengle @dpeterka


cfieber:
no changes to the packaged result, just consistency cleanup

PR #193, opened 2015-11-17
Add default CF settings for running locally.

gregturn:


cfieber:
this seems out of date with the change to s3 deploy artifacts?

gregturn:
I'll fix this PR as well.

gregturn:
@cfieber Updated.

PR #194, opened 2015-11-17
Have Cloud Foundry deployments detect the next sequence for a server group

gregturn:
For a long time, CF has used the version number of the Jenkins job as the basis for rolling the version. That hack is finally ripped out and replaced with detection of the next subsequent version number, insuring clean deployments. This also requires in orca to support as well.


gregturn:
Superceded by #209

PR #197, opened 2015-11-19
/applications now supports relationship expansion (?expand=true)

ajordens:
- Expanding the CLUSTERS relationship takes time and is unnecessary in a lot of situations (perhaps all)


ajordens:
This should have no impact on anything but AWS (and perhaps TITAN although I left the default unfiltered behavior in place).

@duftler @gregturn 

This should sufficiently improve the speed of /applications such that we can serve directly out of it rather than the fallback cache in gate.

cfieber:
nice!

PR #198, opened 2015-11-19
AWS and Titan application providers will now filter cache identifiers…

ajordens:
… rather than return _all_


ajordens:
@tomaslin PTAL

PR #208, opened 2015-11-29
Use account name for environment and accountType when not explicitly-…

duftler:
…specified for Google accounts.

Configure Spring DefaultConversionService so we can tokenize lists of strings via @Value annotations.
Should be merged at the same time as: https://github.com/spinnaker/spinnaker/pull/634
@cfieber please review.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/CloudDriverConfig.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/CloudDriverConfig.groovy
@@ -33,6 +33,8 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import org.springframework.context.annotation.Import
+import org.springframework.core.convert.ConversionService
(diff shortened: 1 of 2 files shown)


gregturn:
San we strive to move towards beans annotated with @ConfigueationProperties instead of leaving on @Value injected settings? I think it pulls settings together in a more logical fashion instead if having settings strewn throughout the code.

clouddriver-security/src/main/groovy/com/netflix/spinnaker/clouddriver/security/config/SecurityConfig.groovy
```diff
@@ -29,4 +31,10 @@ class SecurityConfig {
   CredentialsInitializerSynchronizable noopCredentialsInitializerSynchronizable() {
     new NoopCredentialsInitializerSynchronizable()
   }
+
+  // Allows @Value annotation to tokenize a list of strings.
+  @Bean
+  ConversionService conversionService() {
```
cfieber: could you move this into CloudDriverConfig in clouddriver-core? since its a context wide change it feels more appropriate there

cfieber:
@gregturn fair point

cfieber:
LGTM

duftler:
Done.

PR #209, opened 2015-12-01
Migrate Cloud Foundry deployment to new approach

gregturn:


gregturn:
@cfieber @duftler Please review? 

It's 99% cloud foundry specific, except I created an OperationPoller (and test case) based on Google's implementation and made it generic enough for reuse, putting it in kato-core.

kato/kato-core/src/main/groovy/com/netflix/spinnaker/kato/helpers/OperationPoller.groovy
```diff
@@ -0,0 +1,93 @@
...
+    task.updateStatus basePhase, "Done operating on $resourceString."
+
+    operation
+  }
+
+  private Object pollOperation(Closure operation, Closure ifDone, long timeoutSeconds) {
+    int totalTimePollingSeconds = 0
+    boolean timeoutExceeded = false
+
+    // Fibonacci backoff in seconds, up to googleConfigurationProperties.asyncOperationMaxPollingIntervalSeconds interval.
```
cfieber: is that comment accurate (the source of asyncOperationMaxPollingIntervalSeconds?). Seems like if this moved into kato-core the config prop to support it should be generic

cfieber: ah yeah I see where you are wiring it up, just should just remove the `googleConfigurationProperties.`

gregturn: Old comment. It's configured by constructor injection, and each bean can tap their own prop file.

kato/kato-core/src/main/groovy/com/netflix/spinnaker/kato/helpers/OperationPoller.groovy
```diff
@@ -0,0 +1,93 @@
...
+import com.netflix.spinnaker.kato.exception.OperationTimedOutException
+
+/**
+ * A poller with an upper time limit combined with a Fibonacci-based backoff.
+ * Let's you wrap the operation in a Groovy closure and the "if complete" operation in another.
+ * The "if complete" closure is fed the results of the first.
+ */
+class OperationPoller {
+
+  int asyncOperationTimeoutSecondsDefault
```
cfieber: I'd suggest making these final and giving this class a constructor, or slapping `@Immutable` on it

cfieber:
LGTM if you can address those minor comments and squash I'll merge it

gregturn:
@cfieber Comments addressed and squashed.

PR #216, opened 2015-12-14
Declare k8s client dependency in clouddriver-kubernetes instead of in…

duftler:
… clouddriver.
@lwander please review.
@ttomsu please confirm this fixes the issue you're seeing.


ttomsu:
Fixes my issue. Merging.

PR #219, opened 2015-12-16
Set AutoScalingWorker.metaClass to null in cleanupSpec() so other tes…

duftler:
…ts that rely on AutoScalingWorker.metaClass.deploy() don't fail.


duftler:
@ajordens please review.

ajordens:
LGTM

PR #221, opened 2015-12-17
Add buildpack and memory support to CF

gregturn:
* Add ability to customize the buildpack for a CF deployment
* Add ability to customize memory settings for a CF deployment
* Polish up some details so one off deployments don't blow up under the assumption of a trigger when there is none


gregturn:
@duftler @ajordens please review

duftler:
LGTM.

PR #223, opened 2015-12-18
dependency update to fix empty body PUT to eureka for status change

cfieber:
fixes an error caused by kubernetes-client transitively upgrading okhttp to a version that errors on an empty body PUT with the default client


ajordens:
LGTM

PR #224, opened 2015-12-18
Allow extra fields in user data

AMeng:
It is useful to be able to inject additional Frigga fields in user data. This adds support for:

 * revision
 * countries
 * devPhase
 * hardware
 * zone


AMeng:
Tests pass locally. I'm not sure what that test failure is about.

duftler:
Tried this branch locally and the tests all passed for me as well. Maybe push a commit with just whitespace to get travis to try again.

AMeng:
@duftler, looks like it passed this time. Weird.

cfieber:
yeah I need to figure out what is up with the CI flakiness (I suspect it is something to do with running in a more resource-constrained container causing timeouts..)

cfieber:
LGTM. I will sort out what to do with the duplication just below your change where it explicitly stick in the `NETFLIX_` environment variables in the presence of those name attributes.

It is worth noting that we sort of consciously stopped explicitly supporting those fields from the UI/deployment perspective - the deploy UI doesn't let you explicitly call out those components of a ServerGroup name it just lumps them all under freeFormDetails (same with the deploy JSON payload, it only has a freeFormDetails) so in order to take advantage of this for now you'd need to explicitly format that section of the server group name.

If you rebase I'm happy to merge

AMeng:
> in order to take advantage of this for now you'd need to explicitly format that section of the server group name

Yeah that's exactly what we plan to do. We use revision numbers for our deploys, so we add `r0XX` to our ASG names.

PR #229, opened 2015-12-21
Merge oort-core and oort-web into clouddriver-core and clouddriver-we…

duftler:
…b, respectively.
@ajordens or @cfieber please review.


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/WebConfig.groovy
```diff
@@ -48,17 +55,24 @@ import javax.servlet.Filter
 @Configuration
 @Import([
   SimpleCORSFilter,
+  ApplicationsController,
```
ajordens: I suspect scanning `com.netflix.spinnaker.clouddriver.controllers` is going to be a less cumbersome path forward than explicit @Imports.

settings.gradle
```diff
@@ -18,7 +18,7 @@ rootProject.name="clouddriver"
 
 def subs = [
   kato: ["kato-web", "kato-aws", "kato-manual", "kato-jedis", "kato-docker", "kato-cf", "kato-titan", "kato-azure"],
-  oort: ["oort-core", "oort-web", "oort-aws", "oort-gce", "oort-bench", "oort-titan", "oort-cf", "oort-azure"],
+  oort: ["oort-web", "oort-aws", "oort-gce", "oort-bench", "oort-titan", "oort-cf", "oort-azure"],
```
ajordens: does oort-web still exist, looks like it was folded into clouddriver-web?

duftler: I've kept kato-web, mort-web and oort-web so the subprojects they bring in are not added directly into clouddriver-web. Especially for kato-web, it's a handful of dependencies. Thought it was cleaner this way, and can kill off the remaining {kato|mort|oort}-web .gradle files once the subprojects have all been themselves merged into clouddriver.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/WebConfig.groovy
duftler: Done.

settings.gradle
ajordens: That's reasonable.

ajordens:
LGTM

PR #234, opened 2015-12-22
Move kato-cf/oort-cf into clouddriver-cf

gregturn:
Wrote a lot of missing tests to ensure kato had reasonable test coverage. With that in place, migrated all of kato-cf into clouddriver-cf and verified it still works. Also moved oort-cf.


gregturn:
@duftler @cfieber please review.

duftler:
Reading now...

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/config/CloudFoundryConfig.groovy
```diff
@@ -37,9 +44,53 @@ import org.springframework.scheduling.annotation.EnableScheduling
...
+
+	@Bean
+	@ConditionalOnMissingBean(CloudFoundryClientFactory)
+	CloudFoundryClientFactory cloudFoundryClientFactory() {
+		new DefaultCloudFoundryClientFactory()
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(CloudFoundryDeployHandler)
+	DeployHandler deployHandler(CloudFoundryClientFactory clientFactory) {
```
cfieber: this bean should probably get a different name (`cloudFoundryDeployHandler`?) - it seems ripe for one of those hard to diagnose spring context init problems when someone else drops in a bean of the same name

duftler:
Ah, looks like @cfieber is already on this.

cfieber:
LGTM - can you squash it after that bean rename

and thanks for beefing up the test coverage - nice to see!

gregturn:
Fixed bean definition, squashed and rebased against master.

And more testing is coming. Now with CF supporting most of the things I see needed, I can continue upping test coverage. I also need to refactor certain parts to lend themselves to better testing,

PR #235, opened 2015-12-22
Introduce the concept of a description processor that runs prior to t…

ajordens:
…he AtomicOperation being created

- Supports a migration scenario wherein old-style inputs need conversion (region vs. regions) prior to execution (also eliminates if/else from the Converter classes)
- Netflix will extending upon this to support overriding iamRole's for deployment

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/OperationsController.groovy
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/OperationsController.groovy
@@ -53,7 +53,7 @@ class OperationsController {
   @Autowired OrchestrationProcessor orchestrationProcessor
   @Autowired Registry registry
   @Autowired (required = false) Collection<AllowedAccountsValidator> allowedAccountValidators = []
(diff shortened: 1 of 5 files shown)


ajordens:
PTAL @cfieber 

@duftler / @ttomsu - Thinking that this is also an opportunity to migrate inputs from old to new styles (thinking region vs. regions, etc.). Doing this sort of migration at the edge would let us get away from if/else in either the `Operation` or `Converter` ...

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/OperationsController.groovy
```diff
@@ -51,6 +53,7 @@ class OperationsController {
   @Autowired OrchestrationProcessor orchestrationProcessor
   @Autowired Registry registry
   @Autowired (required = false) Collection<AllowedAccountsValidator> allowedAccountValidators = []
+  @Autowired (required = false) Collection<AtomicOperationDescriptionPreProcessor> atomicOperationDescriptionPreProcessors = []
```
cfieber: would be worth making this a `List` - Spring will respect an `@Order` annotation on any preprocessor and the List would end up sorted by the `@Order` items. If for some reason we want a lower or higher precedence pre-processor we would get that for free

cfieber:
LGTM

PR #236, opened 2015-12-22
Adds a migration path from legacy UDF to newer UDF for AWS UserData generation

cfieber:
The legacy UDF mechanism supports assembling a UserData script customized by account/application/
environment. We are migrating away from this (it is generally not a good idea in the spirit of
immutable infrastructure to have logic in UserData), to a single UserData template file in which
we just token replace environment variables. The logic to customize behaviour based on those
values is baked into the image.

An application can explicitly opt in or out of the legacyUdf format by setting a legacyUdf flag
in the application metadata in front50. Additionally there is a global default when the preference
is unspecified for the application is configured via udf.defaultLegacyUdf property

@dstengle @AMeng I know you both have looked at UserData configuration - are you guys taking advantage at all of the current (what would be come the legacyUdf=true) behaviour of composing multiple files together to customize user data based on application or environment?

Once we finish migration our user data will just be a single file (udf0) that looks like:
````bash
NETFLIX_ACCOUNT="%%account%%"
NETFLIX_ACCOUNT_TYPE="%%accounttype%%"
NETFLIX_ENVIRONMENT="%%env%%"
NETFLIX_APP="%%app%%"
NETFLIX_STACK="%%stack%%"
NETFLIX_CLUSTER="%%cluster%%"
NETFLIX_DETAIL="%%detail%%"
NETFLIX_AUTO_SCALE_GROUP="%%autogrp%%"
NETFLIX_LAUNCH_CONFIG="%%launchconfig%%"
EC2_REGION="%%region%%"
````

@AMeng based on #224 I'm sort of assuming you are doing something similar (just environment variables)

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/LocalFileUserDataProvider.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/LocalFileUserDataProvider.groovy
@@ -16,7 +16,9 @@
 package com.netflix.spinnaker.clouddriver.aws.deploy.userdata
 
 import com.netflix.frigga.Names
+import com.netflix.spinnaker.clouddriver.core.services.Front50Service
 import org.springframework.beans.factory.annotation.Autowired
+import retrofit.RetrofitError
 
 class LocalFileUserDataProvider implements UserDataProvider {
   private static final INSERTION_MARKER = '\nexport EC2_REGION='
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/LocalFileUserDataProvider.groovy [hunk 1]
```diff
@@ -24,38 +26,80 @@ class LocalFileUserDataProvider implements UserDataProvider {
...
+          return localFileUserDataProperties.defaultLegacyUdf
+        }
+        throw re
+      }
+    }
+
+    final int maxRetry = 5
+    final int retryBackoff = 500
+    final Set<Integer> retryStatus = [429, 500]
+    for (int i = 0; i < maxRetry; i++) {
```
cfieber: probably an opportunity to address retry a bit more generally (I know there is also a bunch of custom retry logic in `DiscoverySupport`)

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/LocalFileUserDataProvider.groovy
```diff
@@ -24,38 +26,80 @@ class LocalFileUserDataProvider implements UserDataProvider {
...
-    def rawUserData = assembleUserData(names, region, account)
-    replaceUserDataTokens localFileUserDataProperties.useAccountNameAsEnvironment, names, launchConfigName, region, account, environment, accountType, rawUserData
+    boolean legacyUdf = isLegacyUdf(account, names.app)
+    def rawUserData = assembleUserData(legacyUdf, names, region, account)
+    replaceUserDataTokens legacyUdf, names, launchConfigName, region, account, environment, accountType, rawUserData
   }
 
-  String assembleUserData(Names names, String region, String environment) {
-    def udfRoot = localFileUserDataProperties.udfRoot
+  String assembleUserData(boolean legacyUdf, Names names, String region, String account) {
```
cfieber: renamed environment -> account to be clear that is what we were templating on (not a change in behaviour)

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/LocalFileUserDataProviderSpec.groovy
```diff
@@ -17,42 +17,47 @@
...
     when:
     def userData = localFileUserDataProvider.getUserData(ASG_NAME, LAUNCH_CONFIG_NAME, REGION, ACCOUNT, ENVIRONMENT, ACCOUNT_TYPE)
 
     then:
-    userData == getFormattedUserData()
+    userData == getFormattedUserData(expectedEnvironment)
+
+    where:
+    legacyUdf | expectedEnvironment
+    true      | ACCOUNT
```
cfieber: in the legacy format we didn't have account and environment as separate concepts, so where we were templating environment we were really templating the account name

riltsken:
@AMeng and I were going down the route of using this same schema for user data files, but I think we would ideally like to pass the userdata from an upstream stage in the pipeline to the deploy stage. ex:

```
Jenkins CI job -> Deploy job

Edit pipeline json
{
  "imageId": "${imageId}"
  "userData": "${userData}"
}
```

Where the jenkins CI job would produce the imageId and userData per application based on some yaml config in a repo. Cases we currently have userdata for are:

Application Revision
JVM Options
Environment Variables

We are actually about to propose a PR which adds `userData` here in the `DefaultLaunchConfigurationBuilder`. Would also like some feedback if that is appropriate or not. This PR here would basically not affect us if so.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/LocalFileUserDataProvider.groovy
cfieber: I'm not going to address this in the scope of this change - generic retrofit exception handling/retry belongs in a kork-retrofit library for broader use across spinnaker services

cfieber:
@riltsken supporting a userData attribute of a `basicAmazonDeployDescription` (and `modifyAwsLaunchConfigDescription`) would make sense to me. Would you envision entirely skipping the UserDataProvider, or passing it the value from the `Description` payload and still having it do token replacement?

(either way I don't think it affects this PR, I'm just curious about the direction of the upcoming PR)

riltsken:
We would append after the UDP as overrides to the defaults.

PR #239, opened 2015-12-28
Only set accountName on description if it is not already set.

duftler:


duftler:
@ttomsu please review.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/converters/GoogleAtomicOperationConverterHelper.groovy
```diff
@@ -23,7 +23,9 @@ class GoogleAtomicOperationConverterHelper {
   static Object convertDescription(Map input,
                                    AbstractAtomicOperationsCredentialsSupport credentialsSupport,
                                    Class targetDescriptionType) {
-    input.accountName = input.credentials
+    if (!input.accountName) {
```
ttomsu: One of the two hardest CS problems - naming. I've seen credentials, account, and now accountName. Can we pick one and stick with it?

@cfieber @ajordens

PR #240, opened 2015-12-29
Make Google cloneServerGroup use source.region instead of source.zone.

duftler:


duftler:
@ttomsu please review.

PR #246, opened 2016-01-07
Add tests for Cloud Foundry

gregturn:
* Upgrade cglib to 3.2.0 to support mocking jets3t
* Don't worry about CloudFoundryResourceRetriever since plans are to migrate to CATS

Increases coverage of CF bits from 15% to 57%.


gregturn:
@cfieber please review

PR #247, opened 2016-01-07
Fix handling of server group sequence numbers when rolling over from …

duftler:
…999 back to 000.

Fix handling of sequence numbers when next name is already taken (keep looking until an open slot is found).
Determine most-recent server group by considering creationTimestamp instead of the sequence number.
Throw exception when namespace is exhausted.
@ttomsu please review.

### DIFF
--- a/clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtilSpec.groovy
+++ b/clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtilSpec.groovy
@@ -26,10 +26,14 @@ import com.google.api.services.compute.model.Image
 import com.google.api.services.compute.model.ImageList
 import com.google.api.services.compute.model.Instance
 import com.google.api.services.compute.model.InstanceAggregatedList
+import com.google.api.services.compute.model.InstanceGroupManager
+import com.google.api.services.compute.model.InstanceGroupManagerList
 import com.google.api.services.compute.model.InstancesScopedList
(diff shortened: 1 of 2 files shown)


clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtilSpec.groovy
```diff
@@ -430,4 +439,64 @@ class GCEUtilSpec extends Specification {
...
+    setup:
+      def computeMock = Mock(Compute)
+      def credentials = new GoogleCredentials(PROJECT_NAME, computeMock)
+      def regionsMock = Mock(Compute.Regions)
+      def regionsGetMock = Mock(Compute.Regions.Get)
+      def regionReal = new Region(zones: [ZONE_URL_1, ZONE_URL_2])
+      def instanceGroupManagersMock = Mock(Compute.InstanceGroupManagers)
+      def instanceGroupManagersListZone1Mock = Mock(Compute.InstanceGroupManagers.List)
+      def instanceGroupManagersListZone2Mock = Mock(Compute.InstanceGroupManagers.List)
+      def instanceGroupManagerReal997 = new InstanceGroupManager(name: "$CLUSTER_NAME-v997", creationTimestamp: "2015-12-24T12:06:46.103-08:00")
```
ttomsu: For simplicity, can you make the creation timestamps something very vanilla, like "2000-01-01T00:00:00.997-00:00" where 997 corresponds to its respective cluster version

duftler:
Done. PTAL.

PR #252, opened 2016-01-13
Handle case when no server groups presently exist in cluster.

duftler:
Handle case when most recently-created server group does not define sequential number segment in its name.
@skim1420 please review.


skim1420:
lgtm

PR #257, opened 2016-01-19
provider/kubernetes: Allow for more flexible resource requests/limits.

lwander:
Turns out you can submit requests/limits that omit either CPU or memory, and don't supply a `m/Mi/Gi` suffix. @duftler @skim1420


duftler:
LGTM.

PR #258, opened 2016-01-19
provider/kubernetes: Support multiple namespaces

lwander:


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/DeployKubernetesAtomicOperation.groovy
```diff
@@ -21,6 +21,7 @@ import com.netflix.spinnaker.clouddriver.data.task.TaskRepository
 import com.netflix.spinnaker.clouddriver.deploy.DeploymentResult
 import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.DeployKubernetesAtomicOperationDescription
 import com.netflix.spinnaker.clouddriver.kubernetes.deploy.KubernetesUtil
+import com.netflix.spinnaker.clouddriver.kubernetes.deploy.exception.KubernetesIllegalArgumentException
```
duftler: Sort-order is off here.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/DeployKubernetesAtomicOperation.groovy
```diff
@@ -59,10 +60,19 @@ class DeployKubernetesAtomicOperation implements AtomicOperation<DeploymentResul
...
     task.updateStatus BASE_PHASE, "Initializing creation of replication controller."
 
+
+    task.updateStatus BASE_PHASE, "Looking up provided namespace..."
+
     def credentials = description.kubernetesCredentials
+    def namespace = description.namespace ? description.namespace : "default"
+
+    if (!credentials.isRegisteredNamespace(namespace)) {
+      throw new KubernetesIllegalArgumentException("Namespace $namespace was not provided as a registered namespace within Clouddriver. Registered namespaces are ${credentials.getNamespaces()}.")
```
duftler: What is the behavior if exactly one namespace is configured for the given account within clouddriver, it is not the `default` namespace, and the request elides the namespace attribute?

I also think the error message might be more helpful if s/within Clouddriver/for account $theAccountName.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/validators/StandardKubernetesAttributeValidator.groovy
```diff
@@ -97,6 +97,14 @@ class StandardKubernetesAttributeValidator {
     }
   }
 
+  def validateNamespace(String value, String attribute) {
+    // Namespace is optional, empty taken to mean 'default'
```
duftler: Missing terminating period.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/health/KubernetesHealthIndicator.groovy
```diff
@@ -64,8 +64,10 @@ class KubernetesHealthIndicator implements HealthIndicator {
         KubernetesCredentials kubernetesCredentials = accountCredentials.credentials
 
         // This verifies that the specified credentials are sufficient to
-        // access the referenced Kuberentes master endpoint.
-        kubernetesCredentials.client.namespaces().withName(kubernetesCredentials.namespace).get();
+        // access the referenced Kubernetes master endpoint.
+        for (String namespace in kubernetesCredentials.getNamespaces()) {
```
duftler: Could use `kubernetesCredentials.namespaces.each {}` here instead.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesCredentials.java
```diff
@@ -18,20 +18,22 @@
...
   public KubernetesClient getClient() {
     return client;
   }
 
-  public String getNamespace() {
-    return namespace;
-  }
+  public List<String> getNamespaces() { return namespaces; }
+
+  public Boolean isRegisteredNamespace(String namespace) { return namespaces.contains(namespace); }
```
duftler: Looks like this can throw NPE if the constructor is passed a `null` `namespaces` reference.

clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/validators/StandardKubernetesAttributeValidatorSpec.groovy
```diff
@@ -473,48 +473,95 @@ class StandardKubernetesAttributeValidatorSpec extends Specification {
 
   void "cpu accept"() {
     setup:
+    def errorsMock = Mock(Errors)
```
duftler: Indentation is off in these 2 methods.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/DeployKubernetesAtomicOperation.groovy
lwander: The request fails, which is what I want to have happen; however, I agree it might not be obvious to someone new to the API. The motivation was: I wanted to imply `default` only when no namespace was declared. This applies to both the credentials setup and the incoming requests. This seems reasonable, since this is generally how a "default" is handled. However, I also wanted to allow users to specifically restrict access to `default`, so requests that either imply or explicitly require creating a replication controller in `default` fail when they need to.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesCredentials.java
lwander: That's true - but it can't happen in the context of clouddriver - if `namespaces` is `null` in the clouddriver config, this object will have been instantiated with `namespaces = ["default"]`. I'll place a better exception here though just in case.

duftler:
LGTM.

PR #260, opened 2016-01-20
When retrying deletion of target pool, give up and propagate exceptio…

duftler:
…n when max number of retries have been attempted.
Increase max number of retries from 5 to 60.


duftler:
@ewiseblatt please review.

ewiseblatt:
LGTM

PR #261, opened 2016-01-20
Consolidate AWS AMI lookup logic.

AMeng:
Currently AMI lookups happen in three different places. Two of them have duplicated logic, and the third looks up the AMI in a different way. This consolidates them all to lookup the AMI the same way by using a new method.

The important fix here is to allow the deploy pipeline stage to lookup AMIs that are accessible to the account, but not necessarily owned by the account. This was previously not possible and would give an error:

image

This happens even though the account is able to view the AMI in the AWS console, as it is shared with the account.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AmiIdResolver.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AmiIdResolver.groovy
@@ -9,7 +9,7 @@ import java.util.regex.Pattern
 class AmiIdResolver {
   private static final Pattern amiIdPattern = Pattern.compile('^ami-[0-9a-f]+$')
 
-  public static ResolvedAmiResult resolveAmiId(AmazonEC2 amazonEC2, String region, String nameOrId, String owner = null, String launcher = null) {
+  private static ResolvedAmiResult resolveAmiId(AmazonEC2 amazonEC2, String region, String nameOrId, String owner = null, String launcher = null) {
     def req = new DescribeImagesRequest()
     if (amiIdPattern.matcher(nameOrId).matches()) {
       req.withImageIds(nameOrId)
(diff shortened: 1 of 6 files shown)


AMeng:
This change has led to a different error when trying to deploy:

image

This line seems to be causing it:
```groovy
sourceAmazonEC2.modifyImageAttribute(new ModifyImageAttributeRequest().withImageId(resolvedAmi.amiId).withLaunchPermission(new LaunchPermissionModifications()
        .withAdd(new LaunchPermission().withUserId(targetCredentials.accountId))))
```

It seems to me that if the image is found in a different account, it should not try to modify its permissions.

cfieber:
I like the consolidation of the AmiIdResolver

I suspect the new error is a result of orca injecting a ModifyAsgLaunchPermission operation in a deploy where the target account != the orca default.bake.account

In our model, we have one account where we run our bakery and create images owned by that account (our testing account). People deploy there, and promote images from there into other accounts.
See: https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/CreateDeployTask.groovy#L123
and: https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/kato/tasks/CreateDeployTask.groovy#L51

Can you describe the model a bit more in your configuration around which account owns the image, which account you are deploying into, and if spinnaker is managing both or just the target account?

AMeng:
We are dealing with two accounts here, called: `master` and `production`.

Spinnaker lives in `master` and deploys applications in `production`. Most of our AMIs live in `master` and that case works fine. We also have a few legacy AMIs that live in `production`. This is the case where the deploy is failing. Even though we've granted the `master` account access to the `production`  AMIs, we are still seeing these issues.

anotherchrisberry:
@AMeng @cfieber checking in on this PR - any outstanding issues or questions?

cfieber:
@AMeng This change should be able to go in as is (its a nice cleanup of the code) - if you want to rebase it I can merge

The broader issue exists in orca where we are injecting the 'allowLaunch' operation any time we see a deploy against the non bake account

tomaslin:
what's the status of this PR?

AMeng:
Sorry for the delay. I am trying to rebase but the build is timing out.

AMeng:
As @cfieber said, this should be fine to merge. It doesn't solve the original issue but does clean up the code a bit.

PR #263, opened 2016-01-23
Merge sequence-handling logic from GCEUtil into AbstractServerGroupNa…

duftler:
…meResolver.

Will still handle the rollover from v999 to v000 properly, but will also handle the situation where the slot immediately following the most recently created server group is already taken.
This routine will hunt for the first available slot following the most recently created server group.
Will throw an exception if the namespace is exhausted.
Refactored AbstractServerGroupNameResolver a bit to make it easy for providers to reuse this logic.
Applied to GCE, AWS & Titan.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/services/AsgService.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/services/AsgService.groovy
@@ -21,6 +21,7 @@ import com.google.common.collect.Iterables
 import com.netflix.frigga.Names
 import com.netflix.spinnaker.clouddriver.aws.model.AutoScalingProcessType
 import com.netflix.spinnaker.clouddriver.aws.model.AwsResultsRetriever
+import com.netflix.spinnaker.clouddriver.helpers.AbstractServerGroupNameResolver
 import groovy.transform.Canonical
 
 @Canonical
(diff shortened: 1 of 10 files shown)


duftler:
Tested various deployments with AWS and GCE.

gregturn:
This PR is awesome!

duftler:
Thanks @gregturn.

duftler:
Rebased.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/services/AsgService.groovy
```diff
@@ -57,6 +58,29 @@ class AsgService {
...
+      }
+
+      @Override
+      protected List<AutoScalingGroup> accessResult(DescribeAutoScalingGroupsResult result) {
+        result.autoScalingGroups.findAll { AutoScalingGroup asg ->
+          def names = Names.parseName(asg.autoScalingGroupName)
+          return names.cluster == clusterName
+        }
+      }
+    }.retrieve(new DescribeAutoScalingGroupsRequest()).collect { AutoScalingGroup asg ->
```
ajordens: I'm going to merge this down as I'm working in a near identical area to avoid the need to make an empty DescribeAutoScalingGroupsRequest.

In one of our large regions, this takes upwards of 2 minutes+ to finish processing.

I've got changes that make the previous ASG lookup go against the cache ... PR will be coming shortly once I've incorporated these changes.

PR #265, opened 2016-01-26
The AWSServerGroupNameResolver should use previously cached cluster d…

ajordens:
…ata to determine the current (and therefore next) server group name

- Original behavior involved fetching all server groups and took (many) minutes to complete in a sufficiently large region/account
- Expects that the clouddriver cluster cache is up-to-date


ajordens:
The crux of this change is a discovery that the original `asgService.getAncestorAsg()` took 2 minutes+ to complete on a sufficiently loaded region/account.

I've migrated to using the cached data (_available via the `AmazonClusterProvider`_) to make a much faster determination of what the latest server group is and therefore what the next increment should be.

I did have to refactor a few pieces after pulling in @duftler's changes so please don't merge until I run it through another set of live testing.

That said, please take a look @cfieber / @duftler when you have a chance.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AWSServerGroupNameResolver.groovy
```diff
@@ -45,6 +52,55 @@ class AWSServerGroupNameResolver extends AbstractServerGroupNameResolver {
 
   @Override
   List<AbstractServerGroupNameResolver.TakenSlot> getTakenSlots(String clusterName) {
-    return asgService.getTakenSlots(clusterName)
```
ajordens: `asgService.getTakenSlots` fetched all server groups in order to determine what was available/taken ... switching to using cached cluster data.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AWSServerGroupNameResolver.groovy
```diff
@@ -45,6 +52,55 @@ class AWSServerGroupNameResolver extends AbstractServerGroupNameResolver {
...
+  String resolveNextServerGroupName(String application, String stack, String details, Boolean ignoreSequence) {
+    def originalNextServerGroupName = super.resolveNextServerGroupName(application, stack, details, ignoreSequence)
+    def nextServerGroupName = originalNextServerGroupName
+    def clusterName = combineAppStackDetail(application, stack, details)
+
+    if (nextServerGroupName) {
+      while (true) {
+        // this resolver uses cached data to determine the next server group name so we should verify it does not already
+        // exist before blindly using it
+        def nextServerGroup = asgService.getAutoScalingGroup(nextServerGroupName)
```
ajordens: Honestly this might be overkill ... but because we're using cached data, I've opt'd to double check that the `nextServerGroupName` does not actually already exist.

If it does, we increment and try again.

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AWSServerGroupNameResolverSpec.groovy
```diff
@@ -0,0 +1,136 @@
...
+    def resolver = new AWSServerGroupNameResolver(account, region, asgService, clusterProviders)
+
+    and:
+    def clusterName = 'application-stack-details'
+
+    when:
+    resolver.resolveNextServerGroupName('application', 'stack', 'details', false)
+
+    then:
+    1 * amazonClusterProvider.getCluster('application', account, clusterName) >> {
```
ajordens: This is likely never to happen in reality ... but if someone does happen to have 1000 server groups then we need to fail.

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/SimpleCluster.groovy
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+
+package com.netflix.spinnaker.clouddriver.aws.deploy
+
+import com.netflix.spinnaker.clouddriver.model.Cluster
+import com.netflix.spinnaker.clouddriver.model.LoadBalancer
+import com.netflix.spinnaker.clouddriver.model.ServerGroup
+
+class SimpleCluster implements Cluster {
```
duftler: This class looks quite generic. Any reason it can't be in `clouddriver.deploy` or `clouddriver.deploymodel`?

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/SimpleServerGroup.groovy
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+
+package com.netflix.spinnaker.clouddriver.aws.deploy
+
+import com.netflix.spinnaker.clouddriver.model.Instance
+import com.netflix.spinnaker.clouddriver.model.ServerGroup
+
+class SimpleServerGroup implements ServerGroup {
```
duftler: Same comment here about scoping to aws vs making it generic.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/helpers/AbstractServerGroupNameResolver.groovy
```diff
@@ -44,6 +47,17 @@ abstract class AbstractServerGroupNameResolver extends NameBuilder {
...
+    def latestServerGroup = takenSlots?.max { a, b ->
+      a.createdTime <=> b.createdTime
+    }
+
+    return latestServerGroup
+  }
+
   String resolveNextServerGroupName(String application, String stack, String details, Boolean ignoreSequence) {
     Integer nextSequence = 0
     String clusterName = combineAppStackDetail(application, stack, details)
```
duftler: Looks like the next 6 lines can be replaced with a call to `resolveLatestServerGroupName()`.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/helpers/AbstractServerGroupNameResolver.groovy
```diff
@@ -72,10 +86,14 @@ abstract class AbstractServerGroupNameResolver extends NameBuilder {
     if (stepCounter == SEQUENTIAL_NUMBERING_NAMESPACE_SIZE) {
       throw new IllegalArgumentException("All server group names for cluster $clusterName in $region are taken.")
     }
-
     return generateServerGroupName(application, stack, details, nextSequence, ignoreSequence)
   }
 
+  protected int generateNextSequence(String serverGroupName) {
+    Names parts = Names.parseName(serverGroupName)
+    return ((parts.sequence ?: 0) + 1) % SEQUENTIAL_NUMBERING_NAMESPACE_SIZE
```
duftler: Maybe have this line in just one helper method to be called here and on line 77?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AWSServerGroupNameResolver.groovy
duftler: Wonder if it might be simpler to do a force cache refresh of the identified 'next' server group and then call the original routine again (and do that until the same result is returned twice, or we circle around, or something like that...). Feels like that would similarly address the race window, but haven't thought too deeply about it.

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/SimpleCluster.groovy
ajordens: Fair point ... I did this specifically for test cases (there's a handful of similar scenarios) but opt'd to extract for re-use across a couple tests.

I'll commit a follow-on that moves these to static classes within `Cluster` and `ServerGroup`. Likely only used for test cases but it's a little less duplication I suppose.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/helpers/AbstractServerGroupNameResolver.groovy
ajordens: Done

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/helpers/AbstractServerGroupNameResolver.groovy
ajordens: Done

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AWSServerGroupNameResolver.groovy
ajordens: I'm not sure it makes life any simpler to call a forceCacheRefresh ... it would add a chunk more boiler plate here to force refresh than just going directly to AWS ... > 99% of the time there isn't going to be a newer ASG (_teams aren't pushing new server groups that close to each other, unless by unintentional accident_).

Thoughts @cfieber?

cfieber: forceCacheRefresh is going to trigger enough separate AWS API calls if it hits and our goal is to quickly determine whether we have a good name so I'm ok with just doing the single lookup here

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AWSServerGroupNameResolver.groovy
```diff
@@ -45,6 +52,55 @@ class AWSServerGroupNameResolver extends AbstractServerGroupNameResolver {
...
+  }
+
+  @Override
+  String resolveNextServerGroupName(String application, String stack, String details, Boolean ignoreSequence) {
+    def originalNextServerGroupName = super.resolveNextServerGroupName(application, stack, details, ignoreSequence)
+    def nextServerGroupName = originalNextServerGroupName
+    def clusterName = combineAppStackDetail(application, stack, details)
+
+    if (nextServerGroupName) {
+      while (true) {
```
cfieber: is it worth just giving this some fixed number of attempts?

cfieber: thinking being that the time it takes to do 1000 distinct getAutoScalingGroup calls is probably going to blow things out and we should have just failed after 5? attempts

PR #266, opened 2016-01-26
added os = LINUX to clouddriver-web.gradle for building rpms. does no…

skorten:
…t affect deb, option is ignored


cfieber:
@skorten have a look at the conversation on #264 and lets figure out the plan there

skorten:
good plan

skorten:
FWIW, this change is just so the rpms build with the correct headers so they can be installed by rpm without using the --ignoreos switch.  WIthout this change rpm complains about "unsupported OS".

cfieber:
yeah I ran into that as well, I think there might be some preInstall/postInstall stuff to do along with this to make the RPMs 1st class

skorten:
agreed.

PR #267, opened 2016-01-26
updates DiscoveryCachingAgent to only add one Health entry in the appropriate account

cfieber:
previously, the DiscoveryCachingAgent didn't try to figure out which account an instance was in, for all accounts that shared a discovery server

this change looks at the metadata to get the accountId and compares that against the known credentials, and only if that matches caches a single health entry in the appropriate account

@ajordens ptal


ajordens:
Solid LGTM.

Squash and merge away.

PR #269, opened 2016-02-02
Log the onDemand cache result (for debugging purposes)

ajordens:


ajordens:
PTAL @anotherchrisberry - This might help down the capacity inconsistency.

PR #270, opened 2016-02-02
Index GCE Windows images.

duftler:
@ewiseblatt please review.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
@@ -25,8 +25,6 @@ import com.google.api.client.http.HttpRequest
 import com.google.api.client.http.HttpRequestInitializer
 import com.google.api.services.compute.Compute
 import com.google.api.services.compute.model.*
-import com.netflix.frigga.NameValidation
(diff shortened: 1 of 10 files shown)


duftler:
I verified that the pruned list of Windows images matches that returned by `gcloud`.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -50,7 +47,8 @@ class GCEUtil {
 
   // TODO(duftler): This list should not be static, but should also not be built on each call.
   static final List<String> baseImageProjects = ["centos-cloud", "coreos-cloud", "debian-cloud", "google-containers",
-                                                 "opensuse-cloud", "rhel-cloud", "suse-cloud", "ubuntu-os-cloud"]
+                                                 "opensuse-cloud", "rhel-cloud", "suse-cloud", "ubuntu-os-cloud",
```
ewiseblatt: Would it make sense for this list to be a config property?
That way you could "easily" prune the list to things the organization may care about, as well as possibly add (or entirely replace with) other projects exporting sanctioned images for the local organization(s).

duftler: Done.

PR #272, opened 2016-02-03
Avoid evicting cached server groups that are newer than what Edda kno…

ajordens:
…ws about

Backstory:
It is possible for Edda to return a set of server groups with an `mtime` of N that do not accurately reflect the true state of AWS at time N.

Workaround:
Whenever we see an Edda result that appears to be newer than a previously ForceCache'd result, fetch the server group directly from AWS and compare a few attributes:
- desired/min/max sizes
- suspended processes

If the direct read and edda match, we are free to evict the ForceCache'd result.

If they do not match, we'll keep the ForceCache'd result.

### DIFF
--- a/oort/oort-aws/src/main/groovy/com/netflix/spinnaker/oort/aws/provider/agent/ClusterCachingAgent.groovy
+++ b/oort/oort-aws/src/main/groovy/com/netflix/spinnaker/oort/aws/provider/agent/ClusterCachingAgent.groovy
@@ -398,7 +398,20 @@ class ClusterCachingAgent implements CachingAgent, OnDemandAgent, AccountAware, 
     def usableOnDemandCacheDatas = []
     providerCache.getAll(ON_DEMAND.ns, asgs.collect { Keys.getServerGroupKey(it.autoScalingGroupName, account.name, region) }).each {
       if (it.attributes.cacheTime < start) {
-        evictableOnDemandCacheDatas << it
+        if (account.eddaEnabled) {
+          def asgFromEdda = asgs.find { asg -> it.id.endsWith(":${asg.autoScalingGroupName}") }
+          def asgFromAws = loadAutoScalingGroup(asgFromEdda.autoScalingGroupName, true)
+
+          if (areSimilarAutoScalingGroups(asgFromEdda, asgFromAws)) {
+            log.info("Evicting previous onDemand value for ${asgFromEdda.autoScalingGroupName} (${flattenAutoScalingGroup(asgFromEdda)} vs ${flattenAutoScalingGroup(asgFromAws)}")
+            evictableOnDemandCacheDatas << it
+          } else {
+            log.info("Preserving previous onDemand value for ${asgFromEdda.autoScalingGroupName} (${flattenAutoScalingGroup(asgFromEdda)} vs ${flattenAutoScalingGroup(asgFromAws)}")
+            usableOnDemandCacheDatas << it
+          }
+        } else {
+          evictableOnDemandCacheDatas << it
+        }
       } else {
         usableOnDemandCacheDatas << it
       }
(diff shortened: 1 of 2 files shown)


oort/oort-aws/src/main/groovy/com/netflix/spinnaker/oort/aws/provider/agent/ClusterCachingAgent.groovy
```diff
@@ -554,6 +575,19 @@ class ClusterCachingAgent implements CachingAgent, OnDemandAgent, AccountAware,
...
+    return [
+      desiredCapacity   : asg.desiredCapacity,
+      minSize           : asg.minSize,
+      maxSize           : asg.maxSize,
+      suspendedProcesses: asg.suspendedProcesses*.processName.sort()
+    ]
+  }
+
+  private static boolean areSimilarAutoScalingGroups(AutoScalingGroup asg1, AutoScalingGroup asg2) {
+    return flattenAutoScalingGroup(asg1) == flattenAutoScalingGroup(asg2)
```
ajordens: Just using a simple heuristic of things that we particular care about being current (capacities and suspended processes).

Anything else worthwhile?

ajordens:
This improved the consistency of my test rather significantly ...

That said, it also uncovered another race condition where we get a ForceCache refresh in between determining what cache updates to apply and the actual writes completing.

`ClusterCachingAgent:399`

ajordens:
PTAL @cfieber @anotherchrisberry

oort/oort-aws/src/main/groovy/com/netflix/spinnaker/oort/aws/provider/agent/ClusterCachingAgent.groovy
cfieber: that seems reasonable to me

oort/oort-aws/src/main/groovy/com/netflix/spinnaker/oort/aws/provider/agent/ClusterCachingAgent.groovy
```diff
@@ -546,6 +559,14 @@ class ClusterCachingAgent implements CachingAgent, OnDemandAgent, AccountAware,
     }
   }
 
+  private AutoScalingGroup loadAutoScalingGroup(String autoScalingGroupName, boolean skipEdda) {
+    def skipEddaClients = new AmazonClients(amazonClientProvider, account, region, true)
```
cfieber: this could just be amazonClientProvider.getAutoScaling

cfieber:
LGTM

oort/oort-aws/src/main/groovy/com/netflix/spinnaker/oort/aws/provider/agent/ClusterCachingAgent.groovy
```diff
@@ -398,7 +398,20 @@ class ClusterCachingAgent implements CachingAgent, OnDemandAgent, AccountAware,
     def usableOnDemandCacheDatas = []
     providerCache.getAll(ON_DEMAND.ns, asgs.collect { Keys.getServerGroupKey(it.autoScalingGroupName, account.name, region) }).each {
       if (it.attributes.cacheTime < start) {
-        evictableOnDemandCacheDatas << it
+        if (account.eddaEnabled) {
```
cfieber: we may want to look at reworking this to collect a list of ASG names and do the loadAutoScalingGroup as one call with all the ASGs we want to compare

ajordens: I contemplated that ... and it may still be worthwhile but from looking at clouddriver logs, we don't evict multiple server groups in a single caching cycle very often.

cfieber: makes sense to keep it this way then (simpler)

PR #274, opened 2016-02-04
Server group and security group implementation

scotmoor:
Implementation of server group (scale set) creation, including subnets, as well as security groups


scotmoor:
Please disregard this PR. Wrong target branch

PR #277, opened 2016-02-05
An endpoint that provides a view into pending onDemand requests

ajordens:
- ClusterCachingAgent is the only Agent currently surfacing results


ajordens:
The plan is that the ForceCacheRefresh tasks in orca consume this endpoint and wait until:
- the refreshed server group isn't in the list
- the refreshed server group is in the list BUT it has a cacheTime/processedTime newer than when the forceCache operation occurred

cfieber:
LGTM

PR #278, opened 2016-02-05
provider/kubernetes: Create namespaces defined by credentials

lwander:
I realized it is a bit of a pain to create namespaces with `kubectl`, so I allow clouddriver to create them. 

@duftler, this should be a quicky.


duftler:
Please add a logging call to indicate that each namespace is being created.
LGTM after adding the logging call. Feel free to self-merge.

PR #280, opened 2016-02-05
s/asgName/serverGroupName

ajordens:
- orca is sending both asgName and serverGroupName ... this allows us to remove the deprecated asgName


ajordens:
Lets us remove a deprecated field in orca

PR #281, opened 2016-02-06
provider/kubernetes: Fix bug where cache wasn't being force refreshed correctly

lwander:
This was happening because the caching agent was responding to the `handle(type, name)` call correctly for the types in could cache.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesServerGroupCachingAgent.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesServerGroupCachingAgent.groovy
@@ -42,7 +42,7 @@ class KubernetesServerGroupCachingAgent implements CachingAgent, OnDemandAgent, 
   @Deprecated
   private static final String LEGACY_ON_DEMAND_TYPE = 'KubernetesCluster'
 
-  private static final String ON_DEMAND_TYPE = 'Cluster'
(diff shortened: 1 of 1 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesServerGroupCachingAgent.groovy
```diff
@@ -42,7 +42,7 @@ class KubernetesServerGroupCachingAgent implements CachingAgent, OnDemandAgent,
   @Deprecated
   private static final String LEGACY_ON_DEMAND_TYPE = 'KubernetesCluster'
 
-  private static final String ON_DEMAND_TYPE = 'Cluster'
+  private static final List<String> ON_DEMAND_TYPES = ['ServerGroup', 'Cluster', 'Instance']
```
duftler: I only did a quick read of this, but I would have thought this would be `ON_DEMAND_TYPE = 'ServerGroup'`.

lwander: The thing is this agent is responsible for server groups, clusters, and instances. When the top level `CachingAgent` class queries each `CachingAgent` subclass, it decides whether or not to force a cache refresh based on the response of `handle(type, name)`. Maybe @cfieber knows what the right thing is to do here.

lwander:
Closing since I have an upcoming PR that fixes this.

PR #283, opened 2016-02-07
Add support for creating/cloning GCE Autoscalers on server groups.

duftler:


duftler:
@lwander please review.

duftler:
Relates to https://github.com/spinnaker/spinnaker/issues/700.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -248,6 +249,22 @@ class GCEUtil {
...
 
+  static Autoscaler queryZonalAutoscaler(String projectName,
+                                         String zone,
+                                         String serverGroupName,
+                                         GoogleCredentials credentials) {
+    try {
+      return credentials.compute.autoscalers().get(projectName, zone, serverGroupName).execute()
+    } catch (GoogleJsonResponseException e) {
+      // 404 is thrown if the autoscaler does not exist in the given zone. Any other exception needs to be propagated.
+      if (e.getStatusCode() != 404) {
```
lwander: For one reason or another, we verify that a `404` occurred due to the resource actually missing in [Terraform](https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_compute_instance.go#L560), might be helpful to do that here too.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -425,6 +481,50 @@ class GCEUtil {
...
+      return new Autoscaler(name: serverGroupName,
+                            zone: migCreateOperation.zone,
+                            target: migCreateOperation.targetLink,
+                            autoscalingPolicy: gceAutoscalingPolicy)
+    }
+  }
+
+  static void calibrateTargetSizeWithAutoscaler(BasicGoogleDeployDescription description) {
+    description.autoscalingPolicy.with {
+      if (description.targetSize < minNumReplicas) {
```
lwander: Do you have a verification handler that makes sure these number make sense?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
lwander: (As opposed to the API endpoint not being found)

duftler: I've added an additional check to verify that the response was a `404` _and_ the `details` are populated. I also verified that trying to hit a non-existent endpoint leaves the `details` block null.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
duftler: I've added some additional validation to ensure that the `max` is not less than the `min` (and unit tests for same). Was already verifying that they were both non-negative.

duftler:
PTAL.

lwander:
LGTM

PR #284, opened 2016-02-07
CacheController will now return either 200 / 202 responses

ajordens:
- 200 if the cache update has been applied immediately
- 202 if the cache update has been queued for application (with the expectation that a caller will poll for completion)

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy
@@ -54,24 +54,27 @@ class CatsOnDemandCacheUpdater implements OnDemandCacheUpdater {
   }
 
   @Override
-  void handle(String type, Map<String, ? extends Object> data) {
+  OnDemandCacheUpdater.OnDemandCacheStatus handle(String type, Map<String, ? extends Object> data) {
     Collection<OnDemandAgent> onDemandAgents = onDemandAgents.findAll { it.handles(type) }
-    handle(type, onDemandAgents, data)
+    return handle(type, onDemandAgents, data)
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


ajordens:
Relates to https://github.com/spinnaker/orca/pull/725

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy
```diff
@@ -93,6 +96,8 @@ class CatsOnDemandCacheUpdater implements OnDemandCacheUpdater {
         log.warn("$agent.providerName/$agent.onDemandAgentType failed to handle on demand update for $type", e)
       }
     }
+
+    return hasOnDemandResults ? OnDemandCacheUpdater.OnDemandCacheStatus.PENDING : OnDemandCacheUpdater.OnDemandCacheStatus.SUCCESSFUL
```
lwander: I'm confused about the semantics here. Any result returned from `agent.handle` is written to the cache [here](https://github.com/spinnaker/clouddriver/pull/284/files#diff-52b84ddde4b53349570a957e33dce33fL77). Why does that qualify the result as `PENDING`?

ajordens: There are race conditions at play wherein a force cache refresh could come in at the ~exact same time as a normal caching cycle and be lost.

More specifically, it can take a non-zero amount of time to bundle up and apply the cache writes and that exposes a window of time where an on-demand update could be applied and then be overwritten with stale data.

It's the stale data after a force cache refresh that we need to avoid happening.

lwander: Ah, I missed the orca PR. This makes sense, thank you

PR #286, opened 2016-02-08
Implement CATS-based caching for Cloud Foundry

gregturn:
* Update CloudFoundryLoadBalancerController to use CATS
* Replace map-based providers with Cats-based providers
* Write extensive unit tests for maximum coverage (reached 94%!)


lwander:
This is probably broken with the latest orca-clouddriver contract. See my PR #290 for how to fix it. Specifically [here](https://github.com/spinnaker/clouddriver/pull/290/files#diff-d351e9063de0113282f12053759d3826R209) and [here](https://github.com/spinnaker/clouddriver/pull/290/files#diff-d351e9063de0113282f12053759d3826R251).

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/agent/ClusterCachingAgent.groovy
```diff
@@ -0,0 +1,504 @@
...
+    ]
+
+    log.info("onDemand cache refresh (data: ${data}, evictions: ${evictions})")
+    return new OnDemandAgent.OnDemandResult(
+        sourceAgentType: getOnDemandAgentType(), cacheResult: cacheResult, evictions: evictions
+    )
+  }
+
+  @Override
+  Collection<Map> pendingOnDemandRequests(ProviderCache providerCache) {
```
lwander: This needs to return what's in your `ON_DEMAND` namespace.

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/cache/Keys.groovy
```diff
@@ -0,0 +1,106 @@
...
+  static Map<String, String> parse(String key) {
+    def parts = key.split(':')
+
+    if (parts.length < 2) {
+      return null
+    }
+
+    def result = [provider: parts[0], type: parts[1]]
+
+    switch (result.type) {
```
cfieber: do the other `Keys` classes short-circuit out here if provider is not a match? (feels like they should if they don't)

duftler: We have this: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/cache/Keys.groovy#L54-L56

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/controllers/CloudFoundryLoadBalancerController.groovy
```diff
@@ -17,83 +17,78 @@
...
   List<CloudFoundryLoadBalancerAccount> list() {
-    getSummaryForLoadBalancers(cfResourceRetriever.loadBalancersByAccount).values() as List
+    Collection<String> loadBalancers = cacheView.getIdentifiers(LOAD_BALANCERS.ns)
+    getSummaryForLoadBalancers(loadBalancers).values() as List
   }
 
   @RequestMapping(value = "/{name:.+}", method = RequestMethod.GET)
   CloudFoundryLoadBalancerSummary get(@PathVariable String name) {
-    getSummaryForLoadBalancers(cfResourceRetriever.loadBalancersByAccount).get(name)
+    def searchKey = Keys.getLoadBalancerKey(name, '*', '*')  + "*"
```
cfieber: this seems wrong - I think your getLoadBalancerKey is `account, region, loadBalancerName` so it should be `Keys.getLoadBalancerKey('*', '*', name)` and not require the + '*'

If you do that it then won't require the findAll block

cfieber: my bad, the ordering is correct, but I don't understand why the +  '*' or the findAll afterword is necessary - the globbing should accomplish a match on name by default

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/controllers/CloudFoundryLoadBalancerController.groovy
```diff
@@ -17,83 +17,78 @@
...
-            loadBalancerName: name
-          ]
-
-          loadBalancerDetail.ipAddress = cfLoadBalancer.nativeRoute.host + '.' + cfLoadBalancer.nativeRoute.domain.name
-          loadBalancerDetail.dnsname = cfLoadBalancer.nativeRoute.host + '.' + cfLoadBalancer.nativeRoute.domain.name
-
-          return [ loadBalancerDetail ]
-        }
-      }
+    def searchKey = Keys.getLoadBalancerKey(name, account, region) + '*'
```
cfieber: same question here: why the + '*' and findAll

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/controllers/CloudFoundryLoadBalancerController.groovy
```diff
@@ -17,83 +17,78 @@
...
   @Autowired
-  CloudFoundryResourceRetriever cfResourceRetriever
+  CloudFoundryLoadBalancerController(Cache cacheView) {
+    this.cacheView = cacheView
+  }
 
   @RequestMapping(method = RequestMethod.GET)
   List<CloudFoundryLoadBalancerAccount> list() {
-    getSummaryForLoadBalancers(cfResourceRetriever.loadBalancersByAccount).values() as List
+    Collection<String> loadBalancers = cacheView.getIdentifiers(LOAD_BALANCERS.ns)
```
cfieber: I think this will suffer the same problem that the AWS version does in that in a multi cats enabled cloudprovider world the CacheView is an aggregate of all the cloudprovider data, so getIdentifiers(LOAD_BALANCERS.ns) is going to return results from all providers.

I was going to add a Keys.getCloudProviderKey() that would essentially return provider:* (in this case `cf:*`) so that this could change to `cacheView.filterIdentifiers(LOAD_BALANCERS.ns, Keys.getCloudProviderKey())` 

That said, I don't know that we actually have anything that exercises this method besides tests and I think there is a proposal on the table to cull all the unused bits out of here from @ttomsu which might also resolve this specific issue

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/model/CloudFoundryApplicationInstance.groovy
```diff
@@ -40,7 +41,33 @@ class CloudFoundryApplicationInstance implements Instance, Serializable {
...
+      case InstanceState.DOWN:
+        return HealthState.Down
+      case InstanceState.STARTING:
+        return HealthState.Starting
+      case InstanceState.RUNNING:
+        return HealthState.Up
+      case InstanceState.CRASHED:
+        return HealthState.Down
+      case InstanceState.FLAPPING:
+        return HealthState.Unknown
```
cfieber: I'd be tempted to map this to Down

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/model/CloudFoundryServerGroup.groovy
```diff
@@ -128,7 +131,7 @@ class CloudFoundryServerGroup implements ServerGroup, Serializable {
 
       @Override
       Map<String, Object> getBuildInfo() {
-        buildInfo
+        bi
```
cfieber: where does this come from?

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/agent/ClusterCachingAgent.groovy
```diff
@@ -0,0 +1,504 @@
...
+  }
+
+  @Override
+  String getProviderName() {
+    CloudFoundryProvider.PROVIDER_NAME
+  }
+
+  @Override
+  String getAgentType() {
+    "${account.name}/${account.org}/${account.space}/${ClusterCachingAgent.simpleName}"
```
cfieber: I don't think it hurts anything to do so, but you don't need to partition your agentType on org and space if there is only 1 agent per account (in AWS there is one agent per account+region)

in the CF credentials model can there be multiple org/space associated with one account?

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/config/CloudFoundryProviderConfig.groovy
```diff
@@ -0,0 +1,108 @@
...
+  class CloudFoundrySynchronizerTypeWrapper implements ProviderSynchronizerTypeWrapper {
+    @Override
+    Class getSynchronizerType() {
+      CloudFoundryProviderSynchronizer
+    }
+  }
+
+  class CloudFoundryProviderSynchronizer {}
+
+  @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
```
cfieber: looks like all this is wired up ok but it would be worth confirming that if you edit account configuration and hit the config-refresh endpoint that everything re-initializes properly

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/view/CatsApplicationProvider.groovy
```diff
@@ -0,0 +1,82 @@
...
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Component
+
+import static com.netflix.spinnaker.clouddriver.cf.cache.Keys.Namespace.APPLICATIONS
+import static com.netflix.spinnaker.clouddriver.cf.cache.Keys.Namespace.CLUSTERS
+/**
+ * @author Greg Turnquist
+ */
+@Component
+class CatsApplicationProvider implements ApplicationProvider {
```
cfieber: I'd be ok with just calling this CloudFoundryApplicationProvider (and I would rename CatsApplicationProvider in the aws module at some point)

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/cache/Keys.groovy
gregturn: Fixed

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/model/CloudFoundryServerGroup.groovy
gregturn: I think it's copied from Google.

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/agent/ClusterCachingAgent.groovy
gregturn: No, I've coded CF to map one account onto a single org+space.

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/view/CatsApplicationProvider.groovy
gregturn: Fixed (along with instance, cluster, and load balancer parts).

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/config/CloudFoundryProviderConfig.groovy
gregturn: Edit what where?

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/model/CloudFoundryApplicationInstance.groovy
gregturn: Fixed.

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/controllers/CloudFoundryLoadBalancerController.groovy
gregturn: Sure there is nothing in deck that calls on this?

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/config/CloudFoundryProviderConfig.groovy
cfieber: in the yml configuration that is backing clouddriver. essentially this supports the equivalent of spring-cloud-context and refresh scope for caching agents

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/controllers/CloudFoundryLoadBalancerController.groovy
gregturn: Fixed

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/controllers/CloudFoundryLoadBalancerController.groovy
gregturn: Fixed

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/model/CloudFoundryServerGroup.groovy
cfieber: I meant specifically the `bi` field, I didn't see it in the class

gregturn: Line 126 has it

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/agent/ClusterCachingAgent.groovy
gregturn: Fixed.

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/config/CloudFoundryProviderConfig.groovy
gregturn: I need to learn about this behavior more. Would like to craft a unit test that exerts this. Might need to tap into https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-testing.html#boot-features-testing-spring-boot-applications

gregturn: Couldn't write a test case. But when simply running clouddriver, I would edit clouddriver-local.yml to add another account. Upon hitting the refresh endpoint, I didn't see the synchronizeCloudFoundryProvider get invoked again to rebuild anything. So not sure I'm groking how this works.

P.S. This is a feature that works out of the box with Spring CLoud Config Server.

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/model/CloudFoundryServerGroup.groovy
cfieber: ah sorry - I was tricked by where the diff cut off and didn't notice the extra indentation

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/security/CloudFoundryCredentialsInitializer.groovy
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  @Override
+  String getCredentialsSynchronizationBeanName() {
+    return "synchronizeCloudFoundryAccounts"
+  }
+
+  @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
+  @Bean
+  List<?> synchronizeCloudFoundryAccounts(CloudFoundryConfigurationProperties cloudFoundryConfigurationProperties,
```
cfieber: This is the entry point for the configuration refresh

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/config/CloudFoundryProviderConfig.groovy
cfieber: > P.S. This is a feature that works out of the box with Spring CLoud Config Server.

Not really - the interesting bit here is unscheduling and rescheduling the caching agent processes. This code will be nearly identical if we bring in spring cloud config with the exception of changing the current prototype_scopes to refresh_scope

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/security/CloudFoundryCredentialsInitializer.groovy
cfieber: @duftler mind taking a look over this and see if anything jumps out as missing for the config-refresh stuff?

gregturn:
@cfieber In prepping for devnexus, this branch is running into issues regarding "Wait for Down Instances" and "Force Refresh", so don't merge yet.

P.S. I have to shelve this and make sure master branch works.

gregturn:
I fixed the gaps and even used this branch in my demo, so I have operationally verified it works with deployments and rollbacks.

PTAL @cfieber

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/agent/ClusterCachingAgent.groovy
```diff
@@ -0,0 +1,526 @@
...
+  private static final String LEGACY_ON_DEMAND_TYPE = 'CloudFoundry' // not relevant for CF
+
+  private static final String ON_DEMAND_TYPE = 'ServerGroup'
+
+  private static final TypeReference<Map<String, Object>> ATTRIBUTES = new TypeReference<Map<String, Object>>() {}
+
+  static final Set<AgentDataType> types = Collections.unmodifiableSet([
+      AUTHORITATIVE.forType(SERVER_GROUPS.ns),
+      AUTHORITATIVE.forType(CLUSTERS.ns),
+      AUTHORITATIVE.forType(APPLICATIONS.ns),
```
cfieber: An application can span multiple providers, so this should be `INFORMATIVE.forType(APPLICATIONS.ns)`

clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/agent/ClusterCachingAgent.groovy
```diff
@@ -0,0 +1,526 @@
...
+  @Deprecated
+  private static final String LEGACY_ON_DEMAND_TYPE = 'CloudFoundry' // not relevant for CF
+
+  private static final String ON_DEMAND_TYPE = 'ServerGroup'
+
+  private static final TypeReference<Map<String, Object>> ATTRIBUTES = new TypeReference<Map<String, Object>>() {}
+
+  static final Set<AgentDataType> types = Collections.unmodifiableSet([
+      AUTHORITATIVE.forType(SERVER_GROUPS.ns),
+      AUTHORITATIVE.forType(CLUSTERS.ns),
```
cfieber: A cluster can span multiple providers as well (application/stack/detail = cluster)

cfieber:
LGTM after the AgentDataTypes are updated in CF ClusterCachingAgent

PR #291, opened 2016-02-08
Ensure that in the non-edda case that every forceCache request is pro…

ajordens:
…cessed at least once

- The orca ServerGroupRefreshCacheTask will poll until it sees the request has been processed


ajordens:
Addresses a situation that @lwander pointed out.

No OnDemand record will be evicted until it's been processed at least once, edda or not.

PR #295, opened 2016-02-09
Add Azure implementation of Security groups and Server groups

scotmoor:
Includes implementation of network and subnet support, bug fixes, and some cleanup


clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/loadbalancer/cache/AzureLoadBalancerCachingAgent.groovy
```diff
@@ -0,0 +1,104 @@
...
+import com.netflix.spinnaker.clouddriver.azure.resources.common.cache.Keys
+import com.netflix.spinnaker.clouddriver.azure.resources.common.cache.provider.AzureInfrastructureProvider
+import com.netflix.spinnaker.clouddriver.azure.resources.loadbalancer.model.AzureLoadBalancerDescription
+import com.netflix.spinnaker.clouddriver.azure.security.AzureCredentials
+import groovy.util.logging.Slf4j
+
+import static com.netflix.spinnaker.cats.agent.AgentDataType.Authority.AUTHORITATIVE
+
+@Slf4j
+class AzureLoadBalancerCachingAgent implements CachingAgent, AccountAware {
```
ajordens: You guys will likely need to support OnDemand support ... to handle immediate updates when a forceCacheRefresh task happens.

milismsft: Yes, I already start looking into adding that support given the latest round of PR changes affecting the other cloud providers.

ajordens:
I didn't review in significant depth (3101 loc change makes that difficult) but I didn't see anything venturing outside of azure packages and artifacts so I suspect we're ok.

Will need a re-base.

scotmoor:
re-base complete

scotmoor:
PTAL - Anything else we need to do?

PR #296, opened 2016-02-09
provider/aws&kubernetes: fix cache eviction race condition

lwander:
Fixes bug where edda is disabled and the cache is repopulated with old on demand results.

@ajordens

### DIFF
--- a/oort/oort-aws/src/main/groovy/com/netflix/spinnaker/oort/aws/provider/agent/ClusterCachingAgent.groovy
+++ b/oort/oort-aws/src/main/groovy/com/netflix/spinnaker/oort/aws/provider/agent/ClusterCachingAgent.groovy
@@ -220,7 +220,7 @@ class ClusterCachingAgent implements CachingAgent, OnDemandAgent, AccountAware, 
     }
 
     def cacheResult = metricsSupport.transformData {
-      buildCacheResult(onDemandData.asgs, onDemandData.scalingPolicies, onDemandData.scheduledActions, onDemandData.subnetMap, [:], [])
+      buildCacheResult(onDemandData.asgs, onDemandData.scalingPolicies, onDemandData.scheduledActions, onDemandData.subnetMap, [:], [], Long.MAX_VALUE)
     }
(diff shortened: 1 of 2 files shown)


oort/oort-aws/src/main/groovy/com/netflix/spinnaker/oort/aws/provider/agent/ClusterCachingAgent.groovy
```diff
@@ -480,7 +481,7 @@ class ClusterCachingAgent implements CachingAgent, OnDemandAgent, AccountAware,
 
     for (AutoScalingGroup asg : asgs) {
       def onDemandCacheData = onDemandCacheDataByAsg ? onDemandCacheDataByAsg[Keys.getServerGroupKey(asg.autoScalingGroupName, account.name, region)] : null
-      if (onDemandCacheData) {
+      if (onDemandCacheData && onDemandCacheData.attributes.cacheTime >= start) {
```
ajordens: `start` doesn't look like the right variable to be using here

lwander: whoops, typo - should by `cacheStart`. I'll fix that.

ajordens:
Can you clarify what scenario this came up in?

Issue that there's an old onDemand entry w/ processedCount == 0?

lwander:
Yeah exactly, that entry will be added to the usable entries to have the `processedTime` set, even though they may be stale at this point.

ajordens:
I think this actually causes a problem in the edda case ... specifically when we determine that edda results are older than the cache result (but have a `start` that is actually newer).

What is the issue if it takes an extra cache cycle for the newest data to get folded in (and the old OnDemand data to get evicted)?

lwander:
Ah that's right, I was fixing the case where the list of `usableOnDemandCacheData` now contains entries that have a `cacheTime` later than the start of the load data cycle, but didn't have `processedCount` set yet. So when edda is disabled, stale cache data may be used. I guess it's too much trouble in the edda enabled case to worry about saving a cache cycle. I'll close this.

PR #297, opened 2016-02-09
Temporarily disable cats-caching of Google load balancers.

duftler:


duftler:
Discussed and LGTM'd offline with @ttomsu .

PR #299, opened 2016-02-09
Restrict AWS load balancer cache lookup via 'aws:' so it does not pic…

duftler:
…k up other providers' load balancers.


duftler:
@ajordens please review.
@ttomsu @lwander fyi

ajordens:
LGTM

This sounds like the same issue that @milismsft was running in to the other day.

PR #303, opened 2016-02-10
provider/kubernetes: Populate image pull secrets by default.

lwander:


lwander:
Reopening in proper branch

PR #306, opened 2016-02-11
Add GoogleServerGroup CachingAgent + Provider

ttomsu:
There are two commits in this PR:

1. Moves the `DefaultCacheDataBuilder` from docker to main `cats` package for reuse across all modules
1. Adds the GoogleServerGroup CachingAgent + Provider (which uses the object from above)

This caching agent is a little trickier than the LoadBalancer one, but I think it adds a lot of readability/maintainability by pulling all immediately related API callbacks together into 1 file. Currently, they're spread over numerous files and in the `GoogleResourceRetriever`

@duftler @lwander PTAL

@spinnaker/reviewers FYI


ttomsu:
Fixed up the build issues. I ended up putting the CacheResultBuilder in `clouddriver-google` with intentions to pull it into a common module once i have evictions and on demand supported.

PR #311, opened 2016-02-12
Adds GoogleInstance caching agent + provider

ttomsu:
Also introduces HEALTH namespace, which will be tackled next. 

@duftler @lwander PTAL

@spinnaker/reviewers FYI


lwander:
LGTM

PR #307, opened 2016-02-11
provider/kubernetes: Upsert load balancer

lwander:
Implements upsert load balancer for kubernetes.

Load balancers are represented as kuberentes services (and can optionally be backed by a cloud provider's load balancer, when possible). 

Relationships between load balancers and instances (pods) are maintained using labels & selectors.

@duftler

PR #312, opened 2016-02-12
provider/kubernetes: Cache & expose kubernetes services as load balancers.

lwander:
Implements & adds a load balancer `CachingAgent`, well as a load balancer `Provider`.

This PR is most useful in conjunction with #307.

### DIFF
--- a/clouddriver-kubernetes/log
+++ b/clouddriver-kubernetes/log
@@ -1,0 +1,67 @@
+Inferred project: clouddriver, version: 1.49.0-SNAPSHOT
+Publication nebula not found in project :.
+[buildinfo] Not using buildInfo properties file for this build.
+None of the specified publications matched for project ':' - nothing to publish.
+:cats:cats-core:compileJava UP-TO-DATE
+:cats:cats-core:compileGroovy SKIPPED
+:cats:cats-core:processResources UP-TO-DATE
+:cats:cats-core:classes UP-TO-DATE
+:cats:cats-core:writeManifestProperties
+:cats:cats-core:jar
+:clouddriver-security:compileJava UP-TO-DATE
+:clouddriver-security:compileGroovy
+:clouddriver-security:processResources UP-TO-DATE
+:clouddriver-security:classes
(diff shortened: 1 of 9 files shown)


lwander:
@duftler, @ttomsu

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesLoadBalancer.groovy
```diff
@@ -26,6 +27,26 @@ class KubernetesLoadBalancer implements LoadBalancer, Serializable {
...
+    this.accountName = accountName
+  }
+
+  KubernetesLoadBalancer(Service service, List<KubernetesServerGroup> serverGroupList, String accountName) {
+    this.service = service
+    this.name = service.metadata.name
+    this.namespace = service.metadata.namespace
+    this.region = this.namespace
+    this.accountName = accountName
+    this.serverGroups = serverGroupList.collect { KubernetesServerGroup serverGroup
```
ttomsu: Would `serverGroupList.groupBy { it.name }` work here?

lwander: didn't know that existed, good idea.

lwander: Doesn't work in this case since we are returning a set of maps, not just a single map.

ttomsu:
LGTM

clouddriver-kubernetes/log
```diff
@@ -0,0 +1,67 @@
+Inferred project: clouddriver, version: 1.49.0-SNAPSHOT
```
duftler: @lwander Looks like your log file got checked-in by accident.

PR #315, opened 2016-02-14

heph:
Creates autoscaling groups with `tags` defined from a key-value map. Not sure how y`all want to handle exceptions here, but if the user passes in more than allowed the AWS API throws a LimitExceededException. Could handle it with a validator, but would rather not hard-code to a limit that might change.


claymccoy:
Looks like just a test failure in CopyLastAsgAtomicOperationUnitSpec. Can you run this locally and see if you get the same problem?

claymccoy:
It would be fine to validate and make the size configurable and default to the AWS limit. It also wouldn't be out of line to let the AWS exception happen.

PR #317, opened 2016-02-15
Cache Google subnets via cats.

duftler:
Add cloudProvider attribute to Subnet interface.
Modify SubnetController list() entrypoint to return a map of cloudProvider -> list-of-subnets.
Add listByCloudProvider() entrypoint to SubnetController.
Update AmazonSubnetProvider to only retrieve 'aws' identifiers.
Update GoogleNetworkProvider to only retrieve 'gce' identifiers.

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/SubnetController.groovy
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/SubnetController.groovy
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.clouddriver.controllers
 import com.netflix.spinnaker.clouddriver.model.Subnet
 import com.netflix.spinnaker.clouddriver.model.SubnetProvider
 import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RequestMethod
 import org.springframework.web.bind.annotation.RestController
@@ -35,6 +36,15 @@ class SubnetController {
     subnetProviders.collectMany {
       it.all
     } as Set
+  }
+
+  @RequestMapping(method = RequestMethod.GET, value = "/{cloudProvider}")
+  Set<Subnet> listByCloudProvider(@PathVariable String cloudProvider) {
+    subnetProviders.findAll { subnetProvider ->
+      subnetProvider.type == cloudProvider
+    } collectMany {
+      it.all
+    }
   }
 
   // TODO: implement the rest
--- a/build.gradle
+++ b/build.gradle
@@ -34,7 +34,7 @@ allprojects {
   apply plugin: 'groovy'
 
   spinnaker {
-    dependenciesVersion = "0.26.0"
+    dependenciesVersion = "0.29.0"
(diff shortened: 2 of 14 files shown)


duftler:
@ajordens Please review. Have tested this behind deck and verified no breaking changes.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/SubnetController.groovy
```diff
@@ -31,10 +32,27 @@ class SubnetController {
   List<SubnetProvider> subnetProviders
 
   @RequestMapping(method = RequestMethod.GET)
-  Set<Subnet> list() {
-    subnetProviders.collectMany {
+  Map<String, Set<Subnet>> list() {
```
ajordens: Feels like we need to preserve the original `Set<Subnet>` response and avoid a breaking change. The `Subnet` should have a `type` attribute allowing a prospective caller to differentiate by type if necessary.

Similarly, it seems like filtering by `cloudProvider` could also be accomplished by a filter query parameter? That said, I don't believe we're terribly consistent on filters vs. path params so if you have a reasonable "it's like this already" argument for `/{cloudProvider}` than I'm convinced.

Gate is deserving of a `v2` treatment with consistency in mind.

duftler: I found this style (i.e. /{cloudProvider}) in use by CacheController, ClusterController, NetworkController and OperationsController.
The cloudProvider attribute can be used for filtering if necessary by the caller, but would be a breaking change to rely on the caller to do that anyway.
Chose to explicitly break (and fix) gate as the only caller of clouddriver, and not break any callers of gate.
This overall pattern matches that of NetworkController.

duftler: Just spotted your comment on the companion PR. Let me have another look at it, and will ping this thread again shortly.

duftler:
Changed to preserve the original behavior of `list()`.
This entry-point isn't used by gate after the pending gate pr is merged, so nothing needs to change on the calling side.
This pr needs to be deployed first.

duftler:
It preserves the original behavior of `list()`, but each subnet contains the `cloudProvider` attribute now.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/SubnetController.groovy
ajordens: Argument makes sense to me.

ajordens:
LGTM with this being a non-breaking change now.

ajordens:
With regards to the `cloudProvider` attribute, it's `type` for Cluster/ServerGroup/LoadBalancer.

I'd suggest going with `type` vs. `cloudProvider` for consistency.

ajordens:
LGTM otherwise.

duftler:
Replaced `cloudDriver` with `type`.
@JargoonPard fyi Had to update `AzureSubnet` and `AzureSubnetProvider` as well to satisfy the interfaces.

JargoonPard:
Thanks @duftler

PR #319, opened 2016-02-15

heph:
Creates autoscaling groups with `tags` defined from a key-value map. Tagging @claymccoy, this is a resub cos I suck at git :-)


start_dev.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
duftler: Don't think you meant to include `start_dev.sh` in the pr.

heph: I did not. :-)

PR #326, opened 2016-02-17
Add support to retrieve the Azure VM images specified in the config.y…

milismsft:
Fix bug related to checking the status of the current deployment
Add work around to make a REST call in order to retrieve current error message in case deployment failed
Consolidate duplicated code into one function
Thrown AtomicOperationException when there was an error with the current operation
Add fix for orca exception when editing an existing security group
Fix regression in Azure Load Balancer creation caused by the recent refactoring in the upstream branch
Merge latest PR payload

### DIFF
--- a/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureBaseClient.groovy
+++ b/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureBaseClient.groovy
@@ -16,18 +16,28 @@
 
 package com.netflix.spinnaker.clouddriver.azure.client
 
+import com.fasterxml.jackson.databind.DeserializationFeature
 import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.databind.SerializationFeature
 import com.microsoft.windowsazure.Configuration
+import com.microsoft.windowsazure.exception.ServiceException
 import com.microsoft.windowsazure.management.configuration.ManagementConfiguration
+import com.netflix.spinnaker.clouddriver.azure.common.AzureUtilities
 import com.netflix.spinnaker.clouddriver.azure.security.AzureCredentials
 import groovy.transform.CompileStatic
+import groovy.util.logging.Slf4j
+import org.apache.commons.io.IOUtils
+import org.apache.http.HttpHost
+import org.apache.http.HttpResponse
+import org.apache.http.HttpStatus
+import org.apache.http.client.methods.HttpGet
+import org.apache.http.impl.client.HttpClientBuilder
 
+@Slf4j
 @CompileStatic
 public abstract class AzureBaseClient {
   final String subscriptionId
 
-  static ObjectMapper mapper = new ObjectMapper().configure(SerializationFeature.INDENT_OUTPUT, true)
+  static ObjectMapper mapper = new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
 
   protected AzureBaseClient(String subscriptionId) {
     this.subscriptionId = subscriptionId
@@ -45,5 +55,55 @@ public abstract class AzureBaseClient {
     } catch (Exception e) {
       throw new RuntimeException("Failed to create Azure Resource Management Service", e)
     }
+  }
+
+  public <T> T getAzureRESTJson(AzureCredentials creds, String baseUrl, String targetUrl, List<String> queryParameters, Class<T> valueType){
+    String url = AzureUtilities.getAzureRESTUrl(subscriptionId, baseUrl, targetUrl, queryParameters)
+
+    try {
+      // create HTTP Client
+      def config = buildConfiguration(creds)
+      HttpClientBuilder httpClientBuilder = config.create(HttpClientBuilder.class)
+
+      String proxyHost = System.getProperty("http.proxyHost")
+      String proxyPort = System.getProperty("http.proxyPort")
+      if ((proxyHost != null) && (proxyPort != null)) {
+        HttpHost proxy = new HttpHost(proxyHost, Integer.parseInt(proxyPort))
+        if (proxy != null) {
+          httpClientBuilder.setProxy(proxy)
+        }
+      }
+
+      def httpClient = httpClientBuilder.build()
+
+      // Create HTTP transport objects
+      HttpGet httpRequest = new HttpGet(url)
+
+      // Set Headers
+      httpRequest.setHeader("Content-Type", "application/json")
+
+      // Send Request
+      HttpResponse httpResponse = httpClient.execute(httpRequest)
+      int statusCode = httpResponse.getStatusLine().getStatusCode()
+      if (statusCode != HttpStatus.SC_OK) {
+        ServiceException ex = ServiceException.createFromJson(httpRequest, null, httpResponse, httpResponse.getEntity())
+        throw ex
+      }
+
+      // Deserialize Response
+      def entity = httpResponse.getEntity()
+      def responseContent = entity.getContent()
+      String resultJson = IOUtils.toString(responseContent)
+
+      if (resultJson) {
+        return mapper.readValue(resultJson, valueType)
+      }
+    }
+    catch (Exception e) {
+      log.info("getVMImagesAll -> Unexpected exception " + e.toString())
+      throw new RuntimeException("Unable to make Azure REST call to ${url}", e)
+    }
+
+    null
   }
 }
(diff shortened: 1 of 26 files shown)


milismsft:
@ajordens @cfieber can you please review and merge if there are no issues?

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureBaseClient.groovy
```diff
@@ -46,4 +57,53 @@ public abstract class AzureBaseClient {
...
   }
+
+  public String getAzureRESTJson(AzureCredentials creds, String baseUrl, String targetUrl, List<String> queryParameters){
+    String result = null
+    String url = AzureUtilities.getAzureRESTUrl(subscriptionId, baseUrl, targetUrl, queryParameters)
+
+    try {
+      // create HTTP Client
+      def config = buildConfiguration(creds)
+      HttpClientBuilder httpClientBuilder = config.create(HttpClientBuilder.class)
```
ajordens: HttpClients are usually created once and re-used.

It seems strange to creating a new one from within this method.

Is there a particular type that should be returned vs. a `String` (I've not looked far enough in the review yet to see where it's being consumed)

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureBaseClient.groovy
```diff
@@ -46,4 +57,53 @@ public abstract class AzureBaseClient {
...
+
+      String proxyHost = System.getProperty("http.proxyHost");
```
ajordens: `System.getProperty` doesn't seem quite right. 

Are these not properties that would otherwise end up in a yml config block?

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureBaseClient.groovy
```diff
@@ -46,4 +57,53 @@ public abstract class AzureBaseClient {
...
+      // Send Request
+      HttpResponse httpResponse = httpClient.execute(httpRequest);
+      int statusCode = httpResponse.getStatusLine().getStatusCode();
+      if (statusCode != HttpStatus.SC_OK) {
+        ServiceException ex = ServiceException.createFromJson(httpRequest, null, httpResponse, httpResponse.getEntity());
+        throw ex;
+      }
+
+      // Deserialize Response
+      if (statusCode == HttpStatus.SC_OK) {
```
ajordens: Unnecessary comparison ... you're guaranteed to have SC_OK.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureBaseClient.groovy
ajordens: Also, reason you're using httpclient vs. retrofit or OkHttp? (later being more standard within Spinnaker)

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureBaseClient.groovy
```diff
@@ -46,4 +57,53 @@ public abstract class AzureBaseClient {
...
+        def entity = httpResponse.getEntity()
+        def responseContent = entity.getContent()
+        result = IOUtils.toString(responseContent)
+      }
+    }
+    catch (Exception e) {
+      log.info("getVMImagesAll -> Unexpected exception " + e.toString())
```
ajordens: Suspect that this is redundant given you're throwing an exception immediately afterwards)

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/common/model/AzureDeploymentOperation.groovy
```diff
@@ -0,0 +1,150 @@
...
+   * @param deploymentName the name of the deployment
+   * @return a Collection of error messages capturing deployment related errors
+   */
+  static List<String> checkDeploymentOperationStatus(Task task, String opsName, AzureCredentials creds, String resourceGroupName, String deploymentName) {
+    def errList = new ArrayList<String>()
+    Map<String, Boolean> resourceCompletedState = new HashMap<String, Boolean>()
+    String deploymentState = AzureResourceManagerClient.DeploymentState.DEPLOYING
+    Boolean checkDeployment = true
+
+    while (checkDeployment) {
```
ajordens: I'd like to see an upper bounds on how long you're going to loop for ... rather than the infinite wait

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/common/model/AzureDeploymentOperation.groovy
```diff
@@ -0,0 +1,150 @@
...
+            // TODO remove the REST call and use d.properties.satusMessage.error.message instead
+
+            String responseContent = creds.resourceManagerClient.
+              getAzureRESTJson(
+                              creds,
+                              "https://management.azure.com",
+                              "resourceGroups/${resourceGroupName}/providers/Microsoft.Resources/deployments/${deploymentName}/operations",
+                              ["api-version=2015-11-01"]
+                              )
+            AzureDeploymentOperation updatedDeploymentOperation = getObjectFromJson(responseContent)
```
ajordens: Seems like the kind of method where you should be passing AzureDeploymentOperation into getAzureXXX() and having it do the conversion?

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureBaseClient.groovy
milismsft: This is a temporary code and only added to work around current limitation of the Azure Java SDK. The source code behind it is pretty much exactly the same with what is currently executed when we create the respective Azure clients (resource manager, network and compute). 
The HTTP client builder requires the proper Azure credentials to construct a proper token; the token can expire so the safest thing is to attempt to get a new one every time (though the same token might be returned if it has not expired).
The method is general enough and should be able to handle any REST calls into Azure and return the JSON string corresponding to the call; this string must be converted to the proper object depending on what REST call is made. At the moment there's only one object that we are interested in and that is the deployment operation object in order to extract the proper error message. Without this the only viable option to understand why a deployment failed is to go into the Azure Portal.
I'm expecting to retouch this code the moment we migrate to the new SDK, hopefully remove it.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureBaseClient.groovy
milismsft: Good catch, I will correct this.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/common/model/AzureDeploymentOperation.groovy
milismsft: Makes sense; I'll add some upper bound timeout.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/common/model/AzureDeploymentOperation.groovy
milismsft: Good suggestion. I'll see if I can get it worked out.

milismsft:
@ajordens PTAL

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/vmimage/view/AzureVMImageLookupController.groovy
```diff
@@ -0,0 +1,189 @@
...
+                  version: vmImage.version,
+                  account: parts.account,
+                  region: parts.region
+                )
+    }
+
+    null
+  }
+
+  @ResponseStatus(value = HttpStatus.NOT_FOUND, reason = 'Image not found')
```
ajordens: Minor detail but you probably don't want to include a `reason` as it will override any specific message you provide when throwing the exception.

PR #332, opened 2016-02-19
Weaves together instance and load balancer healths

ttomsu:
This PR includes numerous changes:

1. Introduces model objects to represent instance health and load balancer health (see `c.n.s.c.google.model.health` package)
1. Updates the LoadBalancerCachingAgent and InstanceCachingAgent to use these models 
1. Wires them together in the GoogleInstanceProvider to actually be presented to client users
1. Removes the HEALTH namespace, as it's no longer used.
1. Introduces a new providerImpl property to allow easy switching between old and new providers to help speed development.

@duftler @lwander PTAL. It's kind of a mishmash of changes (and changes in my mind) that have accumulated over the week.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleHealthCheck.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleHealthCheck.groovy
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.google.model
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import groovy.transform.Canonical
+
+@Canonical
+class GoogleHealthCheck {
+  String requestPath
+  int port
+
+  // Attributes
+  int checkIntervalSec
+  int timeoutSec
+  int unhealthyThreshold
+  int healthyThreshold
+
+  @JsonIgnore
+  View getView() {
+    new View()
+  }
+
+  class View implements Serializable {
+    int checkIntervalSec = checkIntervalSec
+    int timeoutSec = timeoutSec
+    int unhealthyThreshold = unhealthyThreshold
+    int healthyThreshold = healthyThreshold
+
+    String getTarget() {
+      port ? "HTTP:${port}${requestPath ?: '/'}" : null
+    }
+
+  }
+}
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/health/GoogleInstanceHealth.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/health/GoogleInstanceHealth.groovy
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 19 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleHealthCheck.groovy
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  class View implements Serializable {
+    int checkIntervalSec = checkIntervalSec
+    int timeoutSec = timeoutSec
+    int unhealthyThreshold = unhealthyThreshold
+    int healthyThreshold = healthyThreshold
+
+    String getTarget() {
+      port ? "HTTP:${port}${requestPath ?: '/'}" : null
```
duftler: Wonder if it makes sense to build this in a safer way? What if requestPath has a value but does not have a leading This comment really applies everywhere we are building urls, not just here. We can worry about this in a later PR.
We should probably be building these urls similar to this: https://github.com/cloudera/director-google-plugin/blob/master/provider/src/main/java/com/cloudera/director/google/util/Urls.java#L64-L80

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/health/GoogleInstanceHealth.groovy
```diff
@@ -0,0 +1,59 @@
...
+@Canonical
+class GoogleInstanceHealth extends GoogleHealth implements Health, Serializable {
+
+  final Type type = Type.Google
+  final HealthClass healthClass = HealthClass.platform
+
+  Status status
+
+  enum Status {
+    PROVISIONING,
```
lwander: Are these specific to Google, or generic to Spinnaker?

duftler:
LGTM.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/health/GoogleLoadBalancerHealth.groovy
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  static class LBHealthSummary {
+    // These aren't the most descriptive names, but it's what's expected in Deck.
+    String loadBalancerName
+    String instanceId
+    ServiceStatus state
+
+    String getDescription() {
+      state == ServiceStatus.OutOfService ?
```
lwander: Wouldn't the description be `"Healthy"`, rather than `null`?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleLoadBalancerCachingAgent.groovy
```diff
@@ -69,63 +74,175 @@ class GoogleLoadBalancerCachingAgent extends AbstractGoogleCachingAgent {
...
   CacheResult loadData(ProviderCache providerCache) {
-    List<GoogleLoadBalancer> loadBalancers = getLoadBalancers()
+    List<GoogleLoadBalancer2> loadBalancers = getLoadBalancers()
     buildCacheResult(providerCache, loadBalancers)
   }
 
-  List<GoogleLoadBalancer> getLoadBalancers() {
-    List<GoogleLoadBalancer> loadBalancers = new ArrayList<GoogleLoadBalancer>()
+  List<GoogleLoadBalancer2> getLoadBalancers() {
+    List<GoogleLoadBalancer2> loadBalancers = new ArrayList<GoogleLoadBalancer2>()
```
lwander: I think you can just use `List<type> x = []` here.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/health/GoogleInstanceHealth.groovy
ttomsu: Specific to GCE instances.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/health/GoogleLoadBalancerHealth.groovy
ttomsu: I don't think it's used if it's not out of service, but I'll add it

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleLoadBalancerCachingAgent.groovy
ttomsu: Added in another PR

PR #333, opened 2016-02-19
provider/docker: Fixed loading of docker credentials

lwander:
Fixes a silly typo


duftler:
LGTM (and impressed that you snuck those in).

PR #334, opened 2016-02-19
Allow for customizing block devices

riltsken:
Related to https://github.com/spinnaker/spinnaker/issues/721

Currently the defaults are hardcoded and unable to be bypassed. This commit allows for passing down custom block devices if they are specified in the pipeline json description.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
@@ -107,9 +107,8 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
         }
       }
 
-      def blockDeviceMappingForInstanceType = BlockDeviceConfig.blockDevicesByInstanceType[description.instanceType]
-      if (blockDeviceMappingForInstanceType) {
-        description.blockDevices = blockDeviceMappingForInstanceType
+      if (description.blockDevices == null) {
+        description.blockDevices = BlockDeviceConfig.blockDevicesByInstanceType[description.instanceType]
       }
 
       // find by 1) result of a previous step (we performed allow launch)


anotherchrisberry:
I don't see anything worrying here but defer to @cfieber on merging.

anotherchrisberry:
ping @cfieber

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
```diff
@@ -107,9 +107,8 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
         }
       }
 
-      def blockDeviceMappingForInstanceType = BlockDeviceConfig.blockDevicesByInstanceType[description.instanceType]
-      if (blockDeviceMappingForInstanceType) {
-        description.blockDevices = blockDeviceMappingForInstanceType
+      if (!description.blockDevices) {
```
duftler: Does this allow for an explicitly-empty set of block devices (not an aws expert, but wouldn't it be a reasonable use-case for one to request a server group whose instances won't be associated with any additional block devices)?

If this is a reasonable use-case, this should probably be `if (description.blockDevices == null)`.

If not reasonable, feel free to disregard this comment.

riltsken: Yep you are correct. I attempted to create the following with empty block device mappings via the aws cli:

m3.medium using hvm_instance_store
c4.large using hvm_ebs

And both resulted in using some defaults amazon provided.

duftler:
Travis timed-out. Please re-push the same branch to kick it off again.

riltsken:
I'll try pushing tomorrow as the build keeps timing out after several attempts.

anotherchrisberry:
There should be a "restart" button to the right of the Travis job if you click on the "Details" link above, which will restart the build:
screen shot 2016-02-25 at 5 54 19 pm

(It's possible that's not visible to everyone, or only visible to logged in users or something)

duftler:
Nice, didn't know that.

murarisumit:
@duftler What modification I need to make in the json file, so that extra disks aren't added ?

jcusa:
I had the same issue where I needed to override the default block size. Looking at this PR, it wasn't immediately clear to me what I needed to do to make it work.  After some digging, hopefully info here will save time for others (like me. :smile: ). 

Lets say you are picking m4.xlarge instance type for deployment. By default, /dev/sdb is set to 80. You want to use 120 instead. To accomplish that, you would go to the spinnaker pipeline that does the deployment, open/edit its json. then, go to the section for "clusters" (search for "instanceType" to jump to that location in the json) In that section, you can add a block like this: 

"blockDevices": [
            {
              "deviceName": "/dev/sdb",
              "size": 120
            }
          ],
add  more as needed. 

Please note, this is just a sample, you will want to refer to[ this ](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonBlockDevice.groovy)to see other values you can set per your specific requirement. Also, take a look at what values are set by default for different instance types: blockDevicesByInstanceType [here](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/BlockDeviceConfig.groovy#L56)

PR #335, opened 2016-02-19
provider/kubernetes: Use credentials from ~/.kube/config

lwander:
The user can now point to a kube config file, default is `~./kube/config` to authenticate with the kubernetes endpoint.

@duftler


duftler:
LGTM. Feel free to self-merge if/when metal travis decides the tests have passed.

duftler:
Btw, I think in general that `String.isEmpty()` makes the intent of the code a bit clearer than using `.length() == 0`.

lwander:
Switched to `.length()`.

lwander:
`.isEmpty()`* whoops

PR #337, opened 2016-02-22
Hooks up Security groups to cluster provider

ttomsu:
Based off of https://github.com/spinnaker/clouddriver/pull/332 - Please only review 2nd commit.

1. Makes GoogleSecurityGroupCachingAgent extend and use the niceties of AbstractGoogleCachingAgent.
  1. Also makes use of CacheResultBuilder
1. Fill in the list of security groups a server group is a member of from the cluster provider's view. Work still needs to be done to hook up security groups within individual instances.

@duftler @lwander PTAL


duftler:
LGTM.

PR #343, opened 2016-02-23
provider/kubernetes: Fix omitted namespaces bug

lwander:
@duftler, @skim1420 - fixes what you pointed out.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesNamedAccountCredentials.java
```diff
@@ -121,12 +114,20 @@ private KubernetesCredentials buildCredentials() {
     if (this.namespaces == null || this.namespaces.isEmpty()) {
       this.namespaces = Collections.singletonList(config.getNamespace());
     }
+
```
duftler: I believe you can drop the `this.` throughout this method.

PR #348, opened 2016-02-23
provider/kubernetes: Retrieve credentials for server group

lwander:
The `credentials` field is needed for clone server group operations to resolve valid images for the given namespace. 

@duftler


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesServerGroup.groovy
```diff
@@ -30,6 +30,7 @@ class KubernetesServerGroup implements ServerGroup, Serializable {
   String name
   String type = "kubernetes"
   String region
+  String credentials
```
duftler: Wouldn't using `account` be more consistent?

duftler:
Looks like the unit tests need to be updated to reflect the modified constructor.

lwander:
I changed every use of `credentials` when referring to an account name to `account`. `credentials` is still used to refer to the class containing the authentication details & client.

duftler:
Looks like Metal Travis has now decided to not even try.
LGTM, once you can get Metal Travis on board.

PR #354, opened 2016-02-26
Cherry-pick: Allow for customizing block devices

duftler:
Currently the defaults are hardcoded and unable to be bypassed. This commit
allows for passing down custom block devices if they are specified in the pipeline json description.


duftler:
travis....

duftler:
Cherry-picked commit from #334.

murarisumit:
@duftler What modification I need to make in the json file, so that extra drives aren't added ?

duftler:
You should just need to ensure that your Create/Clone requests specify
block devices via this attribute:
https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy#L54

PR #357, opened 2016-02-27
provider/kubernetes: Split imageId into registry, repository and tag.

lwander:
The purpose of this is to always display the registry an image came from in Deck.

@duftler

### DIFF
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/provider/agent/DockerRegistryImageCachingAgent.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/provider/agent/DockerRegistryImageCachingAgent.groovy
@@ -32,12 +32,7 @@ import groovy.util.logging.Slf4j
 import static java.util.Collections.unmodifiableSet
 
 @Slf4j
-class DockerRegistryImageCachingAgent implements CachingAgent, OnDemandAgent, AccountAware {
-  @Deprecated
-  private static final String LEGACY_ON_DEMAND_TYPE = 'DockerRegistryTaggedImage'
-
-  private static final String ON_DEMAND_TYPE = 'TaggedImage'
-
+class DockerRegistryImageCachingAgent implements CachingAgent, AccountAware {
   static final Set<AgentDataType> types = unmodifiableSet([
     AgentDataType.Authority.AUTHORITATIVE.forType(Keys.Namespace.TAGGED_IMAGE.ns)
   ] as Set)
@@ -45,16 +40,13 @@ class DockerRegistryImageCachingAgent implements CachingAgent, OnDemandAgent, Ac
   private DockerRegistryCredentials credentials
   private DockerRegistryCloudProvider dockerRegistryCloudProvider
   private String accountName
-  private OnDemandMetricsSupport metricsSupport
 
   DockerRegistryImageCachingAgent(DockerRegistryCloudProvider dockerRegistryCloudProvider,
                                   String accountName,
-                                  DockerRegistryCredentials credentials,
-                                  Registry registry) {
+                                  DockerRegistryCredentials credentials) {
     this.dockerRegistryCloudProvider = dockerRegistryCloudProvider
     this.accountName = accountName
     this.credentials = credentials
-    this.metricsSupport = new OnDemandMetricsSupport(registry, this, "${Keys.Namespace.provider}:${ON_DEMAND_TYPE}".toString())
   }
 
   @Override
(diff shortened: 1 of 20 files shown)


clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/provider/agent/DockerRegistryImageCachingAgent.groovy
```diff
@@ -79,46 +71,6 @@ class DockerRegistryImageCachingAgent implements CachingAgent, OnDemandAgent, Ac
     DockerRegistryProvider.PROVIDER_NAME
   }
 
-  @Override
```
lwander: Removed because this caching agent doesn't every handle onDemand requests.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
```diff
@@ -89,4 +95,48 @@ class KubernetesApiAdaptor {
   Namespace createNamespace(Namespace namespace) {
     client.namespaces().create(namespace)
   }
+
+  static KubernetesContainerDescription fromContainer(Container container) {
```
lwander: This makes it easier to reconstruct Deploy/Clone commands on the UI.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/KubernetesUtil.groovy
```diff
@@ -26,6 +27,7 @@ class KubernetesUtil {
   static String SECURITY_GROUP_LABEL_PREFIX = "security-group-"
   static String LOAD_BALANCER_LABEL_PREFIX = "load-balancer-"
   static String REPLICATION_CONTROLLER_LABEL = "replication-controller"
+  static String DEFAULT_REGISTRY = "gcr.io"
```
duftler: Maybe this should be externalized as config? Could make it optional by using `@Value` here and giving it the same default value.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/description/servergroup/DeployKubernetesAtomicOperationDescription.groovy
```diff
@@ -46,9 +46,17 @@ class KubernetesContainerPort {
...
+  String tag
+  String registry
+}
+
+@AutoClone
+@Canonical
 class KubernetesContainerDescription {
   String name
-  String image
+  KubernetesImageDescription imageDescription
```
duftler: Seems like this would still read ok as `image`, but don't feel too strongly about it.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/KubernetesUtil.groovy
lwander: This is what Kubernetes picks as its default, so if the registry is left empty using regular kubectl, you'll deploy an image from gcr.io

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/KubernetesUtil.groovy
```diff
@@ -48,6 +50,59 @@ class KubernetesUtil {
...
+      image.tag = "latest"
+    }
+
+    if (!image.repository) {
+      throw new IllegalArgumentException("Image descriptions must provide a repository.")
+    }
+  }
+
+  static String getImageId(KubernetesImageDescription image) {
+    "$image.registry/$image.repository:$image.tag"
```
duftler: Can just have this method delegate to the one below.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/KubernetesUtil.groovy
lwander: But I can externalize it if you think that's best.

clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/KubernetesUtilSpec.groovy
```diff
@@ -0,0 +1,75 @@
...
+
+class KubernetesUtilSpec extends Specification {
+  private final String REGISTRY1 = 'gcr.io'
+  private final String REGISTRY2 = 'localhost:5000'
+  private final String REPOSITORY1 = 'ubuntu'
+  private final String REPOSITORY2 = 'library/nginx'
+  private final String TAG1 = '1.0'
+  private final String TAG2 = 'mytag'
+
+  void "should correctly build an image description 1"() {
```
duftler: All of these test cases can probably be combined into one that uses `@Unroll` and `where:`.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/KubernetesUtil.groovy
duftler: Could also define a `toString()` directly on `KubernetesImageDescription` and then have that delegate to `getImageId()` below. Would let you drop a bunch of explicit calls to this `getImageId(KubernetesImageDescription)`.

duftler:
LGTM after addressing comments and fixing the NPE's that caused the Travis build to fail.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/KubernetesUtil.groovy
duftler: Feels like having the option to specify what default is used when the registry is omitted from a request is reasonable. And the default default (heh) can still be `gcr.io`.

PR #359, opened 2016-02-27
provider/kubernetes: Fix loading of resource amounts in API Adaptor.

lwander:


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
```diff
@@ -101,11 +101,11 @@ class KubernetesApiAdaptor {
     containerDescription.name = container?.name
     containerDescription.imageDescription = KubernetesUtil.buildImageDescription(container?.image)
     containerDescription.limits = new KubernetesResourceDescription()
-    containerDescription.limits.cpu = container?.resources?.limits?.cpu
-    containerDescription.limits.memory = container?.resources?.limits?.memory
+    containerDescription.limits.cpu = container?.resources?.limits?.cpu?.amount
```
duftler: This block might read cleaner using something like:
```
container?.resources.with {
  containerDescription.limits = new KubernetesResourceDescription(
    cpu: limits?.cpu?.amount,
    memory: limits?.memory?.amount,
  )
  containerDescription.requests = new KubernetesResourceDescription(
    cpu: requests?.cpu?.amount,
    memory: requests?.memory?.amount,
  )
}
```
This is rough and did not dump this into the ide, so if you decide to use it, you'll need to test it a bit.

duftler:
LGTM irrespective of whether you accept the suggestion or not.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
lwander: Looks neater, I'll go with that.

PR #360, opened 2016-02-27
provider/kubernetes: Correctly assign load balancer type

lwander:
Turns out `type` was already a field in use by Orca.


lwander:
@duftler

duftler:
LGTM.

PR #362, opened 2016-02-28
provider/docker: Cache content digests for triggers.

lwander:
This feeds into docker registry triggers coming later.

@duftler

### DIFF
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/DockerRegistryConfiguration.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/DockerRegistryConfiguration.groovy
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.clouddriver.docker.registry
 import com.netflix.spinnaker.clouddriver.docker.registry.config.DockerRegistryConfigurationProperties
 import com.netflix.spinnaker.clouddriver.docker.registry.health.DockerRegistryHealthIndicator
 import com.netflix.spinnaker.clouddriver.docker.registry.security.DockerRegistryCredentialsInitializer
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.beans.factory.config.ConfigurableBeanFactory
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.boot.context.properties.ConfigurationProperties
@@ -27,6 +28,7 @@ import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.ComponentScan
 import org.springframework.context.annotation.Configuration
 import org.springframework.context.annotation.Import
+import org.springframework.context.annotation.PropertySource
 import org.springframework.context.annotation.Scope
 import org.springframework.scheduling.annotation.EnableScheduling
 
@@ -35,6 +37,7 @@ import org.springframework.scheduling.annotation.EnableScheduling
 @EnableScheduling
 @ConditionalOnProperty('dockerRegistry.enabled')
 @ComponentScan(["com.netflix.spinnaker.clouddriver.docker.registry"])
+@PropertySource(value = "classpath:META-INF/clouddriver-core.properties", ignoreResourceNotFound = true)
 @Import([ DockerRegistryCredentialsInitializer ])
 class DockerRegistryConfiguration {
   @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
@@ -47,5 +50,10 @@ class DockerRegistryConfiguration {
   @Bean
   DockerRegistryHealthIndicator dockerRegistryHealthIndicator() {
     new DockerRegistryHealthIndicator()
+  }
+
+  @Bean
+  String dockerApplicationName(@Value('${Implementation-Version:Unknown}') String implementationVersion) {
+    "Spinnaker/$implementationVersion"
   }
 }
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
@@ -17,8 +17,10 @@
 package com.netflix.spinnaker.clouddriver.docker.registry.api.v2.client
 
 import com.google.gson.GsonBuilder
+import com.netflix.spinnaker.clouddriver.docker.registry.DockerRegistryConfiguration
 import com.netflix.spinnaker.clouddriver.docker.registry.api.v2.auth.DockerBearerToken
 import com.netflix.spinnaker.clouddriver.docker.registry.api.v2.auth.DockerBearerTokenService
+import org.springframework.beans.factory.annotation.Autowired
 import retrofit.Callback
 import retrofit.RestAdapter
 import retrofit.RetrofitError
@@ -37,6 +39,9 @@ class DockerRegistryClient {
   private GsonConverter converter
   private String basicAuth
 
+  @Autowired
+  String dockerApplicationName
+
   public getBasicAuth() {
     return basicAuth
   }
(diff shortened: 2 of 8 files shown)


clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/DockerRegistryConfiguration.groovy
```diff
@@ -48,4 +51,9 @@ class DockerRegistryConfiguration {
   DockerRegistryHealthIndicator dockerRegistryHealthIndicator() {
     new DockerRegistryHealthIndicator()
   }
+
+  @Bean
+  String dockerApplicationName(@Value('${Implementation-Version:Unknown}') String implementationVersion) {
```
duftler: Might be just about time to 'promote' this factory method out of GCE and Kubernetes and into core.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -52,32 +57,76 @@ class DockerRegistryClient {
...
+  public String getDigest(String name, String tag) {
+    def response = request({
+      registryService.getManifest(name, tag, dockerApplicationName)
+    }, { token ->
+      registryService.getManifest(name, tag, token, dockerApplicationName)
+    }, name)
+
+    def headers = response.headers
+
+    def digest = headers.find {
```
duftler: Can `headers

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -86,31 +135,29 @@ class DockerRegistryClient {
...
         token = "Bearer ${dockerToken.bearer_token ?: dockerToken.token}"
-        response = registryService.getTags(repository, token)
+        response = withToken(token)
       } else {
         throw error
       }
     }
 
-    return (DockerRegistryTags) converter.fromBody(response.body, DockerRegistryTags)
+    response
```
duftler: Please use `return` here to be consistent with the original methods in this class.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -86,31 +135,29 @@ class DockerRegistryClient {
...
-      }
-    }
-    return true
+  public void isV2() {
+    request({
+      registryService.checkVersion(dockerApplicationName)
+    }, { token ->
+      registryService.checkVersion(token, dockerApplicationName)
+    }, "v2 version check")
+    null
```
duftler: Is this extraneous now?

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -86,31 +135,29 @@ class DockerRegistryClient {
...
-  public Boolean isV2() {
-    try {
-      registryService.checkVersion()
-    } catch (RetrofitError error) {
-      if (error.response.status != 401) {
-        return false
-      }
-    }
-    return true
+  public void isV2() {
```
duftler: Please add a comment describing the behavior of this method when the service is/isn't 'V2'. Most readers will expect a method starting with `is` to return a `boolean`.

clouddriver-docker/src/test/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClientSpec.groovy
```diff
@@ -41,9 +41,9 @@ class DockerRegistryClientSpec extends Specification {
 
   void "DockerRegistryClient should validate that it is pointing at a v2 endpoint."() {
     when:
-      Boolean result = client.isV2()
+      client.isV2()
 
     then:
-      result
+      true
```
duftler: Also makes me think the return type is `boolean`.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
lwander: I think I'll rename it to `checkV2Availability` or something. There are many ways this check can fail (no v2 endpoint, network down, server failed, bad credentials, etc...) so a boolean didn't seem to cut it.

clouddriver-docker/src/test/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClientSpec.groovy
lwander: I figured this would make it clear that it wasn't `boolean`, as I'm ignoring the (non existant) return value and making the test happy with an explicit `true`.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
lwander: `null`? It makes the linter happy, otherwise it thinks I'm returning the result of `request(..)`.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
lwander: Good point, they totally could be.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/DockerRegistryConfiguration.groovy
lwander: Last time I tried, I couldn't get the core module to load this value. But I'll give it another shot.

duftler: Feel free to work on that in a follow-on PR.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
duftler: Ok. Maybe add a comment saying so? At a glance, it looks funny to see a lone `null` at the end of a `void` method.

clouddriver-docker/src/test/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClientSpec.groovy
duftler: I can't really see what this test tests then. Seems like there should a version that passes, and a version that fails? Or is it demonstrating something else?

lwander: This test actually hits docker hub, and will fail with an exception if I wrote this method incorrectly. Unfortunately it will also fail if docker hub goes down, but I don't see any other sort of meaningful test to write here.

duftler: Sounds like a solid test. Please just add a comment describing what it does and how failures may happen.

lwander:
@duftler, PTAL

duftler:
LGTM.

PR #364, opened 2016-02-29
Define new useAmiBlockDeviceMappings attribute on BasicAmazonDeployDe…

duftler:
…scription.

If set, the AMI block device mappings are used for the newly-provisioned server group.


ajordens:
LGTM

PR #365, opened 2016-02-29
Adding support for aws proxy for deployments not having direct intern…

ajaijoy-bmc:
…et access
1. Updated clouddriver.yml with comments for adding proxy params including windows ntlm support.
2. Updated amazon credentials classes for the same.
3. Updated junits and ran junits.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.clouddriver.aws.security;
 
 import com.amazonaws.AmazonWebServiceClient;
 import com.amazonaws.ClientConfiguration;
+import com.amazonaws.Protocol;
 import com.amazonaws.auth.AWSCredentialsProvider;
 import com.amazonaws.handlers.RequestHandler2;
 import com.amazonaws.regions.Region;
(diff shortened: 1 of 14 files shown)


ajaijoy:
@ajordens  Please let me know your thoughts.

cfieber:
Feels to me like the proxy configuration shouldn't be attached to the `NetflixAmazonCredentials` object

If you are managing many accounts, the proxy configuration for all of them should be the same (its about outbound connectivity from the host not the specific account being managed)

I would update the `AmazonClientProvider.Builder` to allow passing proxy configuration, update `AmazonClientProvider` to take proxy configuration as a constructor arg, and update `com.netflix.spinnaker.clouddriver.aws.AwsConfiguration` to inject the proxy configuration when it registers the `AmazonClientProvider` bean

You could make `AmazonCredentials.AWSProxy` an `@ConfigurationProperties` bound bean

ajordens:
Attaching it one level higher in the hierarchy (aws vs aws.accounts) is a good call and is probably a simpler implementation / configuration.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
```diff
@@ -274,6 +275,45 @@ public AmazonCloudWatch getCloudWatch(NetflixAmazonCredentials amazonCredentials
     }
   }
 
+  private boolean isNTLMProxy(NetflixAmazonCredentials amazonCredentials){
+
+    boolean isNTLMProxy = false;
+
+    AmazonCredentials.AWSProxy proxy = amazonCredentials.getAWSProxy();
+
+    if(proxy != null){
```
ajordens: Minor code formatting issues, spaces between `)` and `{`.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
```diff
@@ -274,6 +275,45 @@ public AmazonCloudWatch getCloudWatch(NetflixAmazonCredentials amazonCredentials
...
+        try{
+          Integer.parseInt(proxy.getProxyPort());
+        }catch(NumberFormatException nfe){
+           isProxy = false;
+        }
+
+
+    }
+
+    return isProxy;
```
ajordens: `return proxy != null && proxy.isValid()'

Might be simpler if these methods were actually on `AWSProxy`.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
```diff
@@ -274,6 +275,45 @@ public AmazonCloudWatch getCloudWatch(NetflixAmazonCredentials amazonCredentials
...
+    AmazonCredentials.AWSProxy proxy = amazonCredentials.getAWSProxy();
+
+    if(proxy != null){
+
+      if(proxy.getProxyHost() != null && proxy.getProxyPort() != null && proxy.getProxyDomain() != null && proxy.getProxyWorkstation() != null )
+        isNTLMProxy = true;
+
+    }
+
+    return isNTLMProxy;
```
ajordens: similar to below, a `proxy.isNTLM()` seems reasonable.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
```diff
@@ -281,6 +321,38 @@ public AmazonCloudWatch getCloudWatch(NetflixAmazonCredentials amazonCredentials
     ClientConfiguration clientConfiguration = new ClientConfiguration();
     clientConfiguration.setRetryPolicy(retryPolicy);
 
+    if(isProxyConfigMode(amazonCredentials)){
```
ajordens:
```
if (proxy != null) {
   proxy.apply(clientConfiguration);
}
```

Or something along those lines might be more easily tested.

ajordens:
Thanks for digging into this @ajaijoy-bmc!

ajaijoy:
Thanks will do the changes.

ajaijoy-bmc:
@cfieber  when i add @ConfigurationProperties to AmazonCredentials$AWSProxy without @Bean it says no bean found, @Bean doesnt compile. Any ideas?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
ajaijoy: sure will fix.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
ajaijoy: agree, will move them to AWSProxy

ajaijoy-bmc:
Closing this pull request and creating new pull request since i changed the design in the second attempt.

PR #369, opened 2016-03-01
Updates AWS create/clone server group and modify launch config to support explicitly specifying classicLinkSecurityGroups

cfieber:
@ajordens PTAL

@claymccoy fyi


ajordens:
LGTM

PR #372, opened 2016-03-02
Clone AWS user data (if not provided) along with other ancestor asg a…

duftler:
…nd launch configuration attributes.
@ajordens or @cfieber please review.
@riltsken Would appreciate your input as well.


duftler:
@riltsken pointed out this earlier relevant thread: https://github.com/spinnaker/clouddriver/pull/237#discussion_r48362556

Will work through that and figure out the right thing to do.

duftler:
This has been updated to feature-flag the `UserDataProviders` behavior. There is a new `udf.enabled` flag that defaults to `true`.
If this flag is explicitly set to `false`, both `CopyLastAsgAtomicOperation` and `ModifyAsgLaunchConfigurationOperation` will copy over any user data from the ancestor.
In the case of `CopyLastAsgAtomicOperation`, the ancestor user data is only copied over if no user data is explicitly provided on the clone request.
@cfieber PTAL

duftler:
Should be no change in behavior unless the `udf.enabled` config property is explicitly set to `false`.

duftler:
Rebased.

cfieber:
LGTM

PR #378, opened 2016-03-03
[Provider: Titus ] Use instaces(Min/Max/Desired) and add a resize operation

tomaslin:


tomaslin:
@dzapata @cfieber PTAL

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/RegionScopedTitusClient.java
```diff
@@ -156,6 +156,14 @@ public Task getTask(String taskId) {
     }
 
     @Override
+    public void resizeJob(ResizeJobRequest resizeJobRequest){
+        if(resizeJobRequest.getUser() == null){
```
dzapata: In which instances would this be null?

dzapata:
LGTM

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/RegionScopedTitusClient.java
tomaslin: in the cases where authentication is disabled. Could happen when Titus is OSS'd

tomaslin: also, duplicating the functionality in submitjobs here

PR #381, opened 2016-03-05
provider/kubernetes: Reassign missing credentials field

lwander:
When `.convert(input, ...)` is called multiple times on the same request, and `input` only provides `credentials`, and not the `account` field, the second call to `convert` will fail since the first call deletes `credentials`. This fixes that.

Also, I added a namespace field to the Kubernetes server group model to make it easier to lookup.

@duftler

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/converters/KubernetesAtomicOperationConverterHelper.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/converters/KubernetesAtomicOperationConverterHelper.groovy
@@ -17,6 +17,10 @@
 package com.netflix.spinnaker.clouddriver.kubernetes.deploy.converters
 
 import com.fasterxml.jackson.databind.DeserializationFeature
+import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.KubernetesAtomicOperationDescription
+import com.netflix.spinnaker.clouddriver.kubernetes.deploy.exception.KubernetesIllegalArgumentException
+import com.netflix.spinnaker.clouddriver.kubernetes.security.KubernetesCredentials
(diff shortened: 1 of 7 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/converters/KubernetesAtomicOperationConverterHelper.groovy
```diff
@@ -24,17 +28,23 @@ class KubernetesAtomicOperationConverterHelper {
                                    AbstractAtomicOperationsCredentialsSupport credentialsSupport,
                                    Class targetDescriptionType) {
     def account = input.account as String
-    input.remove('credentials')
+    def optAccount = input.remove('credentials')
```
duftler: Feel like I should know this, but what is the `opt` in `optAccount`?

duftler: Optional?

duftler: If so, maybe `removedAccount` or `deletedAccount` would be clearer. I can live with it how it is though.

duftler:
LGTM after responding to comment.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/converters/KubernetesAtomicOperationConverterHelper.groovy
lwander: `removedAccount` sounds better to me, I'll fix it up.

PR #385, opened 2016-03-08
provider/docker: Replace empty repository list with _catalog results.

lwander:
@duftler 

Just for you, @tomaslin :wink: 

When a user omits the `repositories` field of a `dockerRegistry` config in `clouddriver.yml`, clouddriver will now try to populate the list of repositories by calling the `/_catalog` endpoint of the docker registry.

### DIFF
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerTokenService.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerTokenService.groovy
@@ -136,7 +136,7 @@ class DockerBearerTokenService {
     def tokenService = realmToService.get(realm)
 
     if (tokenService == null) {
-      def builder = new RestAdapter.Builder().setEndpoint(realm).setLogLevel(RestAdapter.LogLevel.FULL).build()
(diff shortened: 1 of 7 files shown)


tomaslin:
LGTM. Wonder if we want this behavior to be opt-in so not to take down a registry by accident.

lwander:
Hah, maybe, it's only called once per registry per clouddriver initialization, and the bigger registries disable the `/_catalog` endpoint for exactly that reason. I'd rather keep the config simple.

tomaslin:
don't we want this to be called every time the listData is called? Otherwise if I push a new image to my repo, I will never see it

lwander:
Ah, you're totally right. I wouldn't be worried about the the size of the response to the `/_catalog` request, since the number of repositories will still be an order of magnitude or two lower than the number of tags & digests retrieved. 

But, if we estimate 1KB per manifest (which we need for the digest), and we are pointing at a largeish registry with ~1,000 repositories of 100 tags each we are looking at ~100MB of data per refresh cycle. Clouddriver refreshes every 30s, so that's about 3MB/s. I don't think that's enough to knock over a registry, but will probably get you throttled if it's public. WDYT?

tomaslin:
TBH I'm less concerned about DDoSing a registry than about the images not being indexed.

lwander:
That's true - I meant you're right, I'll fix the PR to pick up changes to the repo list with `/_catalog`, but I'm not sure we need an extra flag turn on that behavior.

lwander:
It now refreshes the repo list on every cache load cycle.

tomaslin:
should change the loglevel of docker client to none

PR #388, opened 2016-03-08
updates RedisCache to use hashing to eliminate writes

cfieber:
I went with the approach of keeping the hashing logic all contained inside RedisCache - it has the advantage of being nicely behind the black box, but the disadvantage that a client of the cache can't take advantage of the hash info to figure out if it can avoid doing work.

This change is mostly backwards compatible with an older clouddriver running in parallel - the one edge case I could see is

1. new code caches something + computes the value
2. value changes, old code caches it and doesn't update the hashes
3. value changes back to original
4. new code runs and doesn't update the value because the hash matches

it seems pretty edge-casey but I would still delete the `*:hashes` keys after rolling this in 

@ajordens @duftler PTAL

### DIFF
--- a/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cache/RedisCache.java
+++ b/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cache/RedisCache.java
@@ -33,6 +33,9 @@ import redis.clients.jedis.ScanParams;
 import redis.clients.jedis.ScanResult;
 
 import java.io.IOException;
+import java.nio.charset.Charset;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -48,6 +51,19 @@ import java.util.Set;
 
 public class RedisCache implements WriteableCache {
 
+    public interface CacheMetrics {
+        void merge(String prefix, String type, int itemCount, int keysWritten, int relationshipCount, int hashMatches, int hashUpdates);
+        class NOOP implements CacheMetrics {
+            @Override
+            public void merge(String prefix, String type, int itemCount, int keysWritten, int relationshipCount, int hashMatches, int hashUpdates) {
+                //noop
+            }
+        }
+    }
+
+    private static final String DIGEST_ALGORITHM = "SHA1";
+    private static final String HASH_CHARSET = "UTF8";
+
     private static final int DEFAULT_SCAN_SIZE = 50000;
 
     private static final TypeReference<Map<String, Object>> ATTRIBUTES = new TypeReference<Map<String, Object>>() {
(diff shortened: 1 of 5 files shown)


cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cache/RedisCache.java
```diff
@@ -267,17 +304,48 @@ public CacheData get(String type, String id, CacheFilter cacheFilter) {
...
-        public MergeOp(Set<String> relNames, List<String> keysToSet) {
+        public MergeOp(Set<String> relNames, List<String> keysToSet, Map<byte[], byte[]> hashesToSet, int skippedWrites) {
             this.relNames = relNames;
             this.keysToSet = keysToSet;
+            this.hashesToSet = hashesToSet;
+            this.skippedWrites = skippedWrites;
+        }
+    }
+
+    private boolean hashCheck(Map<String, byte[]> hashes, String id, String serializedValue, MessageDigest digest, List<String> keys, Map<byte[], byte[]> updatedHashes) {
```
ajordens: Javadoc comment on this method would be helpful for our future selfs (since it mutates a parameter).

ajordens:
Approach seems sane and jives with how I thought we'd tackle the problem (hash based on serialized attributes).

I'm curious to see it run in the staging environment.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/RedisCacheConfig.groovy
```diff
@@ -47,7 +47,7 @@ class RedisCacheConfig {
     JedisSource jedisSource,
     ObjectMapper objectMapper,
     @Value('${redis.maxMsetSize:250000}') int maxMsetSize) {
-    new RedisNamedCacheFactory(jedisSource, objectMapper, maxMsetSize)
+    new RedisNamedCacheFactory(jedisSource, objectMapper, maxMsetSize, null)
```
ajordens: What are your thoughts for the non-NOOP `CacheMetrics`? It would be nice to have an indication of how many cache writes we're performing.

cfieber: I was logging it at first (it was skipping a whole ton of items) but I went the `CacheMetrics` route because it let me actually confirm via a test case and it opens the door to wiring something up to export the metrics out via Spectator if we feel like we really need them.

ajordens: Ya, that's what I figured. It would be nice to Spectator wired in ... doesn't have to block us from moving this forward but I'd like to see cache metrics on our dashboards sooner than later. Reasonable?

Seems like these metrics would be much more valuable then what we have right now around how often caching agents run.

duftler:
Sorry for the late response. Seems reasonable to me.
Since I might be missing some context here, this is an optimization and is not fixing a known issue, right?

PR #390, opened 2016-03-09
create app security groups on every ASG creation

claymccoy:


cfieber:
This feels way too opinionated to go in as is

claymccoy:
I will put in the config that makes it optional this morning. I could also make parts of the construction optional. But some of that is tricky, like connecting the ELB and app security groups.

PR #391, opened 2016-03-09
Add support for finding/viewing Azure custom images

milismsft:
Add support for finding/viewing Azure custom images given a set of storage SCS, region, relative path and ostype as specified in config.yml file.
Fix bug/regression whlie retrieving the Azure Market Store VM images.


milismsft:
@spinnaker/netflix-reviewers @ajordens @duftler PTAL

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/security/AzureNamedAccountCredentials.groovy
```diff
@@ -69,21 +73,20 @@ public class AzureNamedAccountCredentials implements AccountCredentials<AzureCre
...
 
   private AzureCredentials buildCredentials() {
-
-    return new AzureCredentials(this.tenantId, this.clientId, this.appKey, this.subscriptionId)
+    new AzureCredentials(this.tenantId, this.clientId, this.appKey, this.subscriptionId)
   }
 
   private static List<AzureRegion> buildRegions(List<String> regions) {
-    return regions?.collect {new AzureRegion(it)}
+    regions?.collect {new AzureRegion(it)}
```
lwander: Is it OK to return `null` here? Would `regions?.collect {new AzureRegion(it)} ?: []` be safer?

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/security/AzureNamedAccountCredentials.groovy
```diff
@@ -91,7 +94,7 @@ public class AzureNamedAccountCredentials implements AccountCredentials<AzureCre
 
     public AzureRegion(String name) {
       if (name == null) {
-        throw new NullPointerException("name");
+        throw new NullPointerException("name")
```
lwander: Seems like a misuse of `NPE`, since I believe it's supposed to be thrown on access of a `null` object, not when one is discovered. Maybe an `IllegalArgumentException` is better?

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/vmimage/view/AzureVMImageLookupController.groovy
```diff
@@ -90,50 +125,67 @@ class AzureVMImageLookupController {
...
+        true))
+
+
+    if (!lookupOptions.customOnly) {
+      // return a list of virtual machine images as read from the config.yml file
+      accountCredentialsProvider.getAll().each { creds ->
+        if (creds instanceof AzureNamedAccountCredentials && (!lookupOptions.account || lookupOptions.account.isEmpty() || lookupOptions.account == creds.accountName)) {
+          creds.vmImages.each { vmImage ->
+            def imageName = vmImage.offer + "-" + vmImage.sku + "(Recommended)"
+            if (lookupOptions.q == null || imageName.toLowerCase().contains(lookupOptions.q.toLowerCase())) {
```
lwander: This should be 
`if (!lookupOptions.q || ...)`
otherwise, `lookupOptions.q == ""` will always be contained in `imageName`, which might not be expected behavior.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/vmimage/view/AzureVMImageLookupController.groovy
```diff
@@ -61,15 +62,31 @@ class AzureVMImageLookupController {
...
+      return result
+    }
+
     // return a list of virtual machine images as read from the config.yml file
-    // TODO: retrieve the list of virtual machine images from the azure respective cache
     accountCredentialsProvider.getAll().each { creds ->
       if(creds instanceof AzureNamedAccountCredentials && creds.accountName == account) {
         creds.vmImages.each { vmImage ->
-          def imageName = vmImage.offer + " " + vmImage.sku + " (${vmImage.publisher})"
+          def imageName = vmImage.offer + "-" + vmImage.sku + "(Recommended)"
```
lwander: Is this the only place this name generation logic is used? It might be helpful to extract into a utility class.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/vmimage/view/AzureVMImageLookupController.groovy
```diff
@@ -61,15 +62,31 @@ class AzureVMImageLookupController {
   List<AzureNamedImage> getVMImage(@PathVariable String account, @PathVariable String region, @PathVariable String imageId) {
```
lwander: I'm not familiar with the Azure platform, so could you explain when someone will want to hit this endpoint over the `azure/images/find` endpoint?

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/security/AzureNamedAccountCredentials.groovy
milismsft: yep, agree, it's safer to return an empty list here.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/security/AzureNamedAccountCredentials.groovy
milismsft: agree

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/vmimage/view/AzureVMImageLookupController.groovy
milismsft: if lookupOptions.q is null or an empty string we do want to return all the recommended images that are specified in the config.yml file.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/vmimage/view/AzureVMImageLookupController.groovy
milismsft: yes, this is the only place.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/vmimage/view/AzureVMImageLookupController.groovy
milismsft: Triggering the pipeline to create a new servergroup (Azure VMSS) we will use the name of the image passed in in order to find its other details (URI, OS type). The search is for the exact name while the other "find" entry is for all the images containing the passed name; we could have the deck sort through the returned list though. 

BTW there are three categories of images we expect the user to select from in the deck:
- custom images (VHDs) from Azure storage blobs that the user/app administrator specified in the config.yml file
- select Azure Market Store VM images that the  user/app administrator specified in the config.yml file
- any available Azure Market Store VM images only if find also passes "configOnly=0" and "q.length() > 3" and image name contains "q"

lwander: What type of trigger will you be using in that case?

Thanks for the summary

milismsft: Current plan is to use jenkins/igor -> echo to trigger the pipeline; we don't plan to use "bakery" at the moment. BTW I just start looking into these, setting them up and figure out the exact workflow from the AWS side so I'm still ramping up.

lwander: Ah I see. So where will the user specify the image name to be deployed in the pipeline workflow? Usually it's done with a bake, or a findImage stage. findImage will resolve a server group and its image, which should provide enough details to specify what will be deployed.

milismsft: Our first approach to this is to start with an expected hard coded image name at first until we can figure out how to get Jenkins to pass the real thing. Implementation of how the image will be built and get dropped into the Azure Storage will follow after. The trigger is supposed to be a very simple one for now, just to notify the app when a new version of the image got uploaded in order to trigger the pipeline; I'm still working on the details so I can't be more specific at the moment.

lwander: Ok, SGTM

lwander:
LGTM

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/vmimage/view/AzureVMImageLookupController.groovy
duftler: Any particular reason we shouldn't extend Rosco to support baking Azure images? I recall we discussed this briefly a while back, and there does exist an Azure builder for Packer.

PR #401, opened 2016-03-10
non transient non serializable fixes

ajaijoy-bmc:
non transient non serializable fixes


cfieber:
We are serializable these model objects with Jackson. I believe marking a field as transient will exclude it from serialization.

ajaijoy-bmc:
ok got it. how do you recommend we fix this fbugs high priority "Non-transient non-serializable instance field in serializable class". Also wouldnt the unit tests have failed ?

ajaijoy-bmc:
Just checked a stackoverflow post, "The reason Jackson serializes the transient member is because the getters are used to determine what to serialize, not the member itself"

PR #402, opened 2016-03-10
Bug fix for Azure SDK deployment operations

scotmoor:
Use specific version of Azure SDK client-runtime library from SNAPSHOT repo. This version contains fix for bug in template deployment operations.

This is a temporary solution until the bug fix is released as part of the upcoming beta2 release.


scotmoor:
@duftler @ajordens @cfieber PTAL

cfieber:
Do we have an ETA on this change going away (I'm ok with it if it is short term)

LGTM

scotmoor:
The current P.O.R. is the end of the month.

PR #404, opened 2016-03-10
Change the caching a bit they said. It will be easy they said.

cfieber:
* Adds a key to disable hashing
* Pipelines evictions calls
* Writes the backup index before the main index so we can swap backup overtop of main

@ajordens PTAL


cfieber:
so appropriate that this is #404

PR #408, opened 2016-03-11
Adds proper server group list to GCE load balancer's new provider impl.

ttomsu:
Fixes a handful of other issues found during testing, like when relationship keys are missing.

For convenience, I'm caching the instance name AND instance zone. Zone is only required to support filtering in the UI.

@duftler @lwander PTAL.


duftler:
LGTM.

PR #409, opened 2016-03-11
provider/kubernetes: Properly report health to Deck

lwander:
@duftler

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/HealthState.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/HealthState.groovy
@@ -2,5 +2,5 @@
 package com.netflix.spinnaker.clouddriver.model
 
 enum HealthState {
-  Up, Down, Unknown, Starting, OutOfService
+  Up, Down, Unknown, Starting, OutOfService, Succeeded, Failed
 }
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesHealth.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesHealth.groovy
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.kubernetes.model
+
+import com.netflix.spinnaker.clouddriver.kubernetes.deploy.KubernetesUtil
+import com.netflix.spinnaker.clouddriver.model.Health
+import com.netflix.spinnaker.clouddriver.model.HealthState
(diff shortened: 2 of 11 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/HealthState.groovy
```diff
@@ -2,5 +2,5 @@
 package com.netflix.spinnaker.clouddriver.model
 
 enum HealthState {
-  Up, Down, Unknown, Starting, OutOfService
+  Up, Down, Unknown, Starting, OutOfService, Succeeded, Failed
```
lwander: I added these two states in anticipation of short running jobs, which can end successfully or fail. @ajordens, @cfieber, @tomaslin WDYT?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesHealth.groovy
```diff
@@ -0,0 +1,44 @@
...
+  KubernetesHealth(Pod pod) {
+    source = "Pod"
+    def phase = pod.status.phase
+    state = phase == "Pending" ? HealthState.Starting :
+      phase == "Running" ? HealthState.Up :
+        phase == "Succeeded" ? HealthState.Succeeded :
+          phase == "Failed" ? HealthState.Failed : HealthState.Unknown
+  }
+
+  KubernetesHealth(String name, ContainerStatus containerState) {
```
duftler: Seems like this parameter should be named `containerStatus`.

duftler:
LGTM.
Metal travis stalled.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/HealthState.groovy
tomaslin: Feels like we're trying to shoehorn some sort of result into the health state. Is there a need to keep the container around if it is no longer executing?

tomaslin: What is the difference between succeeded vs up and failed vs down

lwander: You might want to know if the container running some batch processing task succeeded or failed. In the case where your management system is set to always restart containers on failure (for something like a web server), you'll never see this state. But if your containers are configured to run a task, and exit gracefully, you'll want to know what happened to them. 

The difference is:

`up` signifies the container (or vm) is running properly and passing health checks, whereas `succeeded` tells you the container is no longer running, but has exited successfully (i.e. no signals thrown, exit code 0, etc...).

`down` signifies the container is failing health checks, and should probably be restarted, whereas `failed` tells you the container exited, but not successfully.

lwander: I am up for moving this into a separate enum. Maybe `JobState`? Can be `Running, Succeeded, Failed`?

duftler: I'd think there are similar states/results defined for pipelines.

lwander:
Whoops, broke some of my unit tests.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/HealthState.groovy
lwander: Then again, maybe they shouldn't live in separate enums since the states are all still mutually exclusive.

tomaslin: I'm happy to keep it as it 

* as long as Lars and only Lars maintains it forever

lwander: :sweat:

PR #411, opened 2016-03-12
use trusty VM for travis build instead of container infrastructure

cfieber:
based on a recommendation from travis-ci support to deal with build timeouts


cfieber:
promising:
```
Elapsed time 5 min 42 sec
```

PR #414, opened 2016-03-13
provider/kubernetes: Instance caching agent

lwander:
Needed for instances that are detached from any server group. Will be useful for operations like abandon, and searching for pods not managed by replication controllers.

@duftler


duftler:
Unit test failed.

duftler:
LGTM once tests are fixed.

PR #416, opened 2016-03-13
provider/kubernetes: (De)register instances

lwander:
@duftler 

register/deregister instances by manipulating the labels that attach them to load balancers.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesInstanceCachingAgent.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesInstanceCachingAgent.groovy
@@ -82,8 +82,8 @@ class KubernetesInstanceCachingAgent implements  CachingAgent, AccountAware {
 
     Map<String, MutableCacheData> cachedInstances = MutableCacheData.mutableCacheMap()
 
-    pods.each { pod ->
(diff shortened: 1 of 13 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesInstanceCachingAgent.groovy
```diff
@@ -82,8 +82,8 @@ class KubernetesInstanceCachingAgent implements  CachingAgent, AccountAware {
 
     Map<String, MutableCacheData> cachedInstances = MutableCacheData.mutableCacheMap()
 
-    pods.each { pod ->
-      def rc = pod.metadata.labels[KubernetesUtil.REPLICATION_CONTROLLER_LABEL] ?: Keys.DETACHED_POD
+    for (Pod pod : pods) {
+      def rc = pod.metadata.labels[KubernetesUtil.REPLICATION_CONTROLLER_LABEL] ?: ''
```
duftler: Is `Keys.DETACHED_POD` still used elsewhere?

What's the reason to switch from `.each` to `for`?

lwander: Whoops, it's not, I'll delete it.

Groovy bug; we've seen this before where a closure sometimes can't access certain class attributes. In this case, `accountName` was null inside the loop, but correct outside it.

duftler:
LGTM after removing `Keys.DETACHED_POD`.

PR #417, opened 2016-03-14
Makes two Google caching agents extend the common abstract class

ttomsu:
Includes Network & Subnet caching agents.

Enables WRITE_DATES_AS_TIMESTAMP serialization feature only once, instead of within each caching agent.

@duftler @lwander PTAL

tag:  https://github.com/spinnaker/spinnaker/issues/712


duftler:
LGTM.

PR #420, opened 2016-03-14
When merging items into the cache, perform the operation in batches

cfieber:
* reduce redis operation size (time to execute the op on the redis size)
* reduce the computed redis operation memory footprint (on the client side)

@ajordens PTAL

This probably eliminates the need to also paginate msets but I left that in there


ajordens:
LGTM

Want to also set maxMsetSize in clouddriver.yml when you roll this out. It's still set artificially high from when we were OOO.

PR #423, opened 2016-03-15
provider/kubernetes: Destroy server group

lwander:
@duftler.

I moved the shared generic server group description from enable/disable into a `KubernetesServerGroupDescription`.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
```diff
@@ -61,6 +61,10 @@ class KubernetesApiAdaptor {
     client.replicationControllers().inNamespace(namespace).withName(name).scale(size)
   }
 
+  boolean hardDestroyReplicationController(String namespace, String name) {
```
lwander: 'hard' destroy because this destroys its pods as well.

duftler:
LGTM.

PR #426, opened 2016-03-15
Implementation for caching server groups

scotmoor:
Implemented caching for server groups plus some minor fixes


scotmoor:
wrong repo

PR #427, opened 2016-03-15
Makes all callbacks implement FailureLogger

ttomsu:
@lwander @duftler PTAL


lwander:
LGTM!

PR #428, opened 2016-03-15
provider/docker: Allow user to specify password file

lwander:
This makes authenticating with gcr.io easier, since it requires your full GCP json key.


lwander:
@duftler

duftler:
LGTM.

PR #431, opened 2016-03-16
provider/docker: Paginage catalog results

lwander:
@tomaslin

### DIFF
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
@@ -19,19 +19,15 @@ package com.netflix.spinnaker.clouddriver.docker.registry.api.v2.client
 import com.google.gson.GsonBuilder
 import com.netflix.spinnaker.clouddriver.docker.registry.api.v2.auth.DockerBearerToken
 import com.netflix.spinnaker.clouddriver.docker.registry.api.v2.auth.DockerBearerTokenService
+import com.netflix.spinnaker.clouddriver.docker.registry.api.v2.exception.DockerRegistryOperationException
 import com.squareup.okhttp.OkHttpClient
 import org.springframework.beans.factory.annotation.Autowired
-import org.springframework.beans.factory.annotation.Value
 import retrofit.RestAdapter
 import retrofit.RetrofitError
 import retrofit.client.OkClient
 import retrofit.client.Response
 import retrofit.converter.GsonConverter
-import retrofit.http.GET
-import retrofit.http.HEAD
-import retrofit.http.Header
-import retrofit.http.Headers
-import retrofit.http.Path
+import retrofit.http.*
 
 import java.util.concurrent.TimeUnit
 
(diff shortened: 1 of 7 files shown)


lwander:
Tested on a local docker registry with a paginate size of `n=1,2,3` for a handful of repositories.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -49,6 +51,9 @@ class DockerRegistryClient {
   @Value('${dockerRegistry.client.timeout:60000}')
   int clientTimeout
 
+  @Value('${dockerRegistry.client.paginateSize:100}')
+  int paginateSize
```
lwander: @tomaslin, I actually can't seem to get this value to be read properly. Regardless of if I have a `dockerRegistry.client.paginateSize` field in my config, it's always set to 0. I had to manually override it for testing. Any ideas?

tomaslin: what are you putting in your config?

lwander:
```
dockerRegistry:
  client:
    paginateSize: 1
```

Even without that, it should default to 100, right?

gregturn: If it was loaded through Spring, yes. If not, Java would default an int to 0.

tomaslin: yeah, the dockerRegistryClient is not being loaded with Spring it seems

gregturn: DockerRegistryClient has no @Component-based annotations, so unless it is being explicitly configured somewhere via @bean, it won't get picked up by component scanning.

tomaslin:
@lwander any updates on this? feels the solution would be to move the values you're injecting into the providerconfig or make the client into a proper spring bean.

lwander:
Whoops, forgot about this. I'll take a look now.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/config/DockerRegistryConfigurationProperties.groovy
```diff
@@ -37,6 +37,10 @@ class DockerRegistryConfigurationProperties {
     String address
     // How many threads to cache all provided repos on. Really only useful if you have a ton of repos.
     int cacheThreads
+    // Timeout time in milliseconds for this repository. Default is 60,000 (1 minute).
+    long clientTimeoutMillis
+    // Paginate size for the docker repository /_catalog endpoint. Default is 100
+    int paginateSize
```
lwander: @tomaslin I moved clientTimeout into clientTimeoutMillis to per-account configuration. Getting the beans to work was too much of a pain, as the autowired dependencies aren't available until after the constructor runs.

tomaslin: SGTM

lwander: Cool, going to rebase and merge this then.

PR #433, opened 2016-03-16
Make CatsSearchProvider work with cached GCE data in Redis

ttomsu:
I needed to add `region` back into the instance key because it's needed to form the target URL.

@duftler @lwander PTAL.


lwander:
LGTM

PR #438, opened 2016-03-16
Stop using cloudFoundryCloudProvider.id

gregturn:


clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/agent/ClusterCachingAgent.groovy
```diff
@@ -86,7 +86,7 @@ class ClusterCachingAgent implements CachingAgent, OnDemandAgent, AccountAware {
     this.cloudFoundryClientFactory = cloudFoundryClientFactory
     this.cloudFoundryCloudProvider = cloudFoundryCloudProvider
```
cfieber: doesn't look like this needs to be passed in here anymore

gregturn:
@cfieber Based on your feedback, I spotted a couple other places that no longer needed it. Updated the unit test.

gregturn:
@cfieber please review.

cfieber:
LGTM

gregturn:
@cfieber rebased.

PR #440, opened 2016-03-17
don't skip edda just because we make modify call to AWS

claymccoy:
@cfieber 
I skipped edda just because I reused the client to modify the ingress. It appears that that was merely superstition. I think this will fix the rate limiting that we sometimes see when migrating certain ASGs in prod.
This is small and should really go to prod ASAP if it makes sense.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/SecurityGroupLookupFactory.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/SecurityGroupLookupFactory.groovy
@@ -59,7 +59,7 @@ class SecurityGroupLookupFactory {
       if (description.vpcId) {
         request.withVpcId(description.vpcId)
       }
-      final amazonEC2 = amazonClientProvider.getAmazonEC2(credentials, region, true)
+      final amazonEC2 = amazonClientProvider.getAmazonEC2(credentials, region)
       final result = amazonEC2.createSecurityGroup(request)
       final newSecurityGroup = new SecurityGroup(ownerId: credentials.accountId, groupId: result.groupId,
         groupName: description.name, description: description.description, vpcId: description.vpcId)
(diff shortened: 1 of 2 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/SecurityGroupLookupFactory.groovy
```diff
@@ -81,7 +81,7 @@ class SecurityGroupLookupFactory {
 
     private List<SecurityGroup> getSecurityGroups(String accountName, AmazonEC2 amazonEC2) {
       List<SecurityGroup> securityGroupsForAccount = securityGroupsByAccount[accountName]
-      if (!securityGroupsForAccount) {
+      if (securityGroupsForAccount == null) {
```
claymccoy: While I am here. Because fuck groovy truthiness, I really meant "== null" but was just lazy.

cfieber:
The implication of this is that those calls are going to return the cached view of security groups so someone can't use them in the context of doing a create or modify (after the change).

Does that seem like a concern based on how they are used?

claymccoy:
Yes. That is a deal breaker. That is why I specifically asked about this on hipchat last night.
This is how I thought it worked originally. I'll create a cached one just for the describe call. Thanks.

cfieber:
The client can execute a modify call it is just that any read operation is always going to return the cached view of the data not the aws api view of the data

claymccoy:
It sounds like your two comments here are contradictory.

"Someone can't use them in the context of doing a create or modify"

"Client can execute a modify"

cfieber:
You changed the code so that getSecurityGroupByName will not find a security group that hasn't hit the cache yet which would be surprising behavior in this part of the codebase.

claymccoy:
No it is intended, it was always a cache and it always did that. It just stuffs old sgs in securityGroupsByAccount for future use. I'll add documentation.

I am only concerned with the writes being handled this way. And concerned about the conflicts on that specific point.

PR #441, opened 2016-03-17
Filter keyset by account and region.

duftler:
There were attempts being made against aws accounts to query load balancers in one region that actually existed in another (api throttling came into play as well).
Adding a `printStackTrace()` in the `catch` block will show the exceptions being thrown prior to this change.


duftler:

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonLoadBalancerInstanceStateCachingAgent.groovy
```diff
@@ -95,7 +95,7 @@ class AmazonLoadBalancerInstanceStateCachingAgent implements CachingAgent,Health
   CacheResult loadData(ProviderCache providerCache) {
     log.info("Describing items in ${agentType}")
     def loadBalancing = amazonClientProvider.getAmazonElasticLoadBalancing(account, region)
-    def loadBalancerKeys = getCacheView().filterIdentifiers(LOAD_BALANCERS.ns, Keys.getLoadBalancerKey('*', '*', '*', '*'))
+    def loadBalancerKeys = getCacheView().filterIdentifiers(LOAD_BALANCERS.ns, Keys.getLoadBalancerKey('*', account.name, region, '*'))
```
cfieber: I don't know why we are reading load balancer ids from the cache here - this means we wouldn't pick up health for an elb until we had cached it.

There should be a loadBalancing.describeLoadBalancers we could call and page through to get the list of ELB ids

This change makes sense as is though

duftler:
Good to merge then?

PR #443, opened 2016-03-17
provider/kubernetes: Mark disabled server groups w/o load balancers as unknown

lwander:
When server groups are deleted from Deck, they must first be disabled. Already disabled server groups were showing up as healthy, making the disable operation a NoOp, causing the delete operation to hang indefinitely. 

@ewiseblatt @duftler


duftler:
LGTM.

PR #444, opened 2016-03-17
New impl of GoogleLoadBalancerController

ttomsu:
Getting close to being able to disable `GoogleResourceRetriever`!

@duftler @lwander PTAL

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController2.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController2.groovy
@@ -1,0 +1,145 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.google.controllers
+
+import com.fasterxml.jackson.annotation.JsonIgnore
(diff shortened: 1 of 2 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController2.groovy
```diff
@@ -0,0 +1,145 @@
...
+  static class GoogleLoadBalancerSummary {
+    String account
+    String region
+    String name
+    String type = GoogleCloudProvider.GCE
+  }
+
+  static class GoogleLoadBalancerDetails {
+    Long createdTime
+    String dnsname
```
duftler: `dnsName`?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController2.groovy
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @JsonProperty("regions")
+    List<GoogleLoadBalancerAccountRegion> getRegions() {
+      mappedRegions.values() as List
+    }
+  }
+
+  static class GoogleLoadBalancerAccountRegion {
+    String name
```
duftler: If this is the region, shouldn't we name it `region`?

duftler: eh, I guess I can live with it. Looked weird at first read, but I suppose it identifying the _name_ of the region is ok.

duftler:
LGTM after addressing the one (potentially) relevant comment.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController2.groovy
ttomsu: Old impl has all lowercase.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController2.groovy
ttomsu: Agreed on it looking weird, but it matches the old impl's output verbatim.

PR #446, opened 2016-03-17
provider/kubernetes: One last health tweak...

lwander:
@duftler 

I'm now labeling the health types by which kubernetes construct they come from (service, pod, container). This allows my to set `interestingHealthProviders` accordingly. For example, during disable, the only health construct that's relevant is `KubernetesService`, since pods and containers have separate notions of health.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesHealth.groovy
```diff
@@ -24,28 +23,33 @@ import io.fabric8.kubernetes.api.model.Pod
 
 class KubernetesHealth implements Health {
   HealthState state
-  String source
-  String type = "Kubernetes" // All healths are reported by Kubernetes
+  final String source
+  final String type
+  final String healthClass = "platform"
```
duftler: Nice.

duftler:
LGTM.

PR #448, opened 2016-03-18
Made 'type' a proper Enum in OnDemandAgent/CacheUpdater interfaces

ttomsu:
Considering the `type` of each OnDemandAgent is limited in scope to either ServerGroups, SecurityGroups, or LoadBalancers, having a free form String felt extraneous, so I made it a proper Enum.

Based on the OortService/MortService interfaces in Orca, the deprecated/legacy stuff is just dead code and has been dropped in this PR. 

@duftler @lwander @cfieber @ajordens @JargoonPard @gregturn - This change touches all providers. PTAL.

tag:  https://github.com/spinnaker/spinnaker/issues/712

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/OnDemandAgent.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/OnDemandAgent.groovy
@@ -34,10 +34,26 @@ interface OnDemandAgent {
     Map<String, Collection<String>> evictions = [:]
   }
 
-  @Deprecated
-  boolean handles(String type)
+  enum OnDemandType {
+    ServerGroup,
+    SecurityGroup,
+    LoadBalancer,
 
-  boolean handles(String type, String cloudProvider)
+    static OnDemandType fromString(String s) {
+      if (s) {
+        String mod = s.toLowerCase().trim()
+        for (OnDemandType type : values()) {
+          String typeMod = type.toString().toLowerCase()
+          if (typeMod == mod) {
+            return type
+          }
+        }
+      }
+      throw new IllegalArgumentException("Cannot create OnDemandType from String '${s}'")
+    }
+  }
+
+  boolean handles(OnDemandType type, String cloudProvider)
 
   OnDemandResult handle(ProviderCache providerCache, Map<String, ? extends Object> data)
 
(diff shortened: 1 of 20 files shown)


lwander:
LGTM

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/CacheController.groovy
```diff
@@ -29,22 +30,12 @@ class CacheController {
...
-    }
-    new ResponseEntity(HttpStatus.ACCEPTED)
-  }
-
   @RequestMapping(method = RequestMethod.POST, value = "/{cloudProvider}/{type}")
   ResponseEntity handleOnDemand(@PathVariable String cloudProvider,
                                 @PathVariable String type,
                                 @RequestBody Map<String, ? extends Object> data) {
-    def cacheStatus = onDemandCacheUpdaters.find { it.handles(type, cloudProvider) }?.handle(type, cloudProvider, data)
+    OnDemandAgent.OnDemandType onDemandType = OnDemandAgent.OnDemandType.fromString(type)
```
cfieber: Need to handle the IllegalArgumentException here for an unknown type

Previously that would result in a null cacheStatus then a 200 OK

arguably a 404 would be more appropriate

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/OnDemandAgent.groovy
```diff
@@ -34,10 +34,26 @@ interface OnDemandAgent {
...
 
-  @Deprecated
-  boolean handles(String type)
+  enum OnDemandType {
+    ServerGroup,
+    SecurityGroup,
+    LoadBalancer,
+
+    static OnDemandType fromString(String s) {
+      if (s) {
```
ajordens:
```
def onDemandType = values().find { it.toString().equalsIgnoreCase(type) }
if (!onDemandType) {
    throw IAE(...)
}
return onDemandType
```

?

ttomsu: Good catch. I originally had declared them as SERVER_GROUP, so I had a "replace('_', '')" call in there.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/CacheController.groovy
ttomsu: Added exception handler that returns 404

ajordens: Is there any known situation where we're sending an invalid type across from orca?

With this throwing a 404 it'll probably now result in the task/pipeline failing.

ttomsu: That's probably a better situation than silently failing with a 200. I don't know of any situation where Orca sends over any value other than the 3 listed.

PR #452, opened 2016-03-19
add an option to attach app, stack, and cluster scoped vpc security groups to the classic link interface if they exist during deployment

cfieber:
also fixes a clone across regions or accounts (resolves groupIds to names, then re-looks up those names in the destination account's classic link vpc)

@ajordens PTAL

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
@@ -18,7 +18,9 @@ package com.netflix.spinnaker.clouddriver.aws.deploy.handlers
 
 import com.amazonaws.services.autoscaling.model.BlockDeviceMapping
 import com.amazonaws.services.autoscaling.model.DescribeAutoScalingGroupsRequest
+import com.amazonaws.services.ec2.model.DescribeSecurityGroupsRequest
 import com.google.common.annotations.VisibleForTesting
+import com.netflix.frigga.Names
 import com.netflix.spinnaker.clouddriver.aws.AwsConfiguration
 import com.netflix.spinnaker.clouddriver.aws.security.NetflixAmazonCredentials
 import com.netflix.spinnaker.clouddriver.data.task.Task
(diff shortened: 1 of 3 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
```diff
@@ -101,21 +103,80 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
...
+            if (description.freeFormDetails) {
+              String clusterGroup = "${description.application}-${description.stack ?: ''}-${description.freeFormDetails}"
+              if (!classicLinkGroupNames.contains(clusterGroup)) {
+                groupNamesToLookUp.add(clusterGroup)
+              }
+            }
+            if (groupNamesToLookUp && classicLinkGroupNames.size() < deployDefaults.maxClassicLinkSecurityGroups) {
+              def appGroups = regionScopedProvider.securityGroupService.getSecurityGroupIds(groupNamesToLookUp, classicLinkVpcId, false)
+              for (String name : groupNamesToLookUp) {
+                if (appGroups.containsKey(name) && classicLinkGroupNames.size() < deployDefaults.maxClassicLinkSecurityGroups) {
```
ajordens: Any sense in logging what groups we're aren't bring over because > 5?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
```diff
@@ -101,21 +103,80 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
...
+                groupNamesToLookUp.add(stackGroup)
+              }
+            }
+            if (description.freeFormDetails) {
+              String clusterGroup = "${description.application}-${description.stack ?: ''}-${description.freeFormDetails}"
+              if (!classicLinkGroupNames.contains(clusterGroup)) {
+                groupNamesToLookUp.add(clusterGroup)
+              }
+            }
+            if (groupNamesToLookUp && classicLinkGroupNames.size() < deployDefaults.maxClassicLinkSecurityGroups) {
```
ajordens: In theory the size check isn't necessary since your'e doing it again below (but would save a `getSecurityGroupIds` call I suppose).

ajordens:
LGTM.

Normally I'd says having a test around the new behavior would be beneficial but I'll leave that to you.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
```diff
@@ -101,21 +103,80 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
...
+          // if we have provided groups and a vpcId, resolve them back to names to handle the case of cloning
+          // from a Server Group in a different region
+          if (description.classicLinkVpcId && description.classicLinkVpcSecurityGroups) {
+            def groupIds = classicLinkGroupNames.findAll { it.matches(~/sg-[0-9a-f]+/) } ?: []
+            classicLinkGroupNames.removeAll(groupIds)
+            if (groupIds) {
+              def describeSG = new DescribeSecurityGroupsRequest().withGroupIds(groupIds)
+              def provider = sourceRegionScopedProvider ?: regionScopedProvider
+              def resolvedNames = provider.amazonEC2.describeSecurityGroups(describeSG).securityGroups.findResults {
+                if (it.vpcId == classicLinkVpcId && groupIds.contains(it.groupId)) {
```
ajordens: How would you get back a groupId that wasn't in the original `groupIds` that you were describing with? Not sure the secondary check is necessary?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
cfieber: yeah good call

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
cfieber: the getSecurityGroupIds call can be really slow (throttling I guess?) so that is what I was looking to avoid

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
cfieber: I wasn't sure if the DescribeSecurityGroups would fail on an unknown groupId or just return the subset that were found. Because they potentially come in from an API request I figured validating they were legit during the lookup made sense.

PR #454, opened 2016-03-19
Check for Java 1.8 before starting.

ewiseblatt:
@cfieber 

I have a similar change queued up for all the other microservices. I want to run the first one by you before I send out the others (mostly whether this is the right place) to put it.

### DIFF
--- a/clouddriver-web/clouddriver-web.gradle
+++ b/clouddriver-web/clouddriver-web.gradle
@@ -63,9 +63,16 @@ jar {
 }
 
 startScripts {
-  defaultJvmOpts = applicationDefaultJvmArgs + ["-Dspring.config.location=/opt/spinnaker/config/"]
+   defaultJvmOpts = applicationDefaultJvmArgs + ["-Dspring.config.location=/opt/spinnaker/config/"]
   doLast {
     unixScript.text = unixScript.text.replace('DEFAULT_JVM_OPTS=', '''\
+        java_version=$(java -version 2>&1|head -1)
+        releasenum=$(echo $java_version | sed 's/[^\\.]*\\.\\([^\\.]*\\)\\..*/\\1/')
+        if [[ "$releasenum" != "8" ]]; then
+          >&2 echo "Java 1.8 is required. You are running $java_version."
+          exit -1
+        fi
+
         if [ -f /etc/default/spinnaker ]; then
           set -a
           . /etc/default/spinnaker


ewiseblatt:
ping

clouddriver-web/clouddriver-web.gradle
```diff
@@ -63,9 +63,14 @@ jar {
 }
 
 startScripts {
-  defaultJvmOpts = applicationDefaultJvmArgs + ["-Dspring.config.location=/opt/spinnaker/config/"]
+   defaultJvmOpts = applicationDefaultJvmArgs + ["-Dspring.config.location=/opt/spinnaker/config/"]
   doLast {
     unixScript.text = unixScript.text.replace('DEFAULT_JVM_OPTS=', '''\
+        java_version=$(java -version 2>&1|head -1)
+        if [[ "$java_version" != *"1.8.0"* ]]; then
```
tomaslin: what happens when someone installs java 9 in 400 days?

gregturn: I think we should split the java_version on ".", and pull out the second field, since it IS the major version now, and run with that.

cfieber: I like that idea

ewiseblatt:
how about this?

anotherchrisberry:
ping - any opinions on this PR?

ajordens:
This LGTM ... if there are no other comments I'd rebase and merge.

clouddriver-web/clouddriver-web.gradle
```diff
@@ -63,9 +63,16 @@ jar {
 }
 
 startScripts {
-  defaultJvmOpts = applicationDefaultJvmArgs + ["-Dspring.config.location=/opt/spinnaker/config/"]
+   defaultJvmOpts = applicationDefaultJvmArgs + ["-Dspring.config.location=/opt/spinnaker/config/"]
```
ajordens: This extra indention is probably unintentional

anotherchrisberry:
The `startScripts` block was removed by @cfieber in 5/16 - should we close this PR?

ewiseblatt:
Sure. I hadnt noticed this was LGTM'd. I had variants for all the other
services and was waiting on this at the time before proceeding on those.
Was startScripts removed from the others as well?

PR #457, opened 2016-03-21
Add frontendpool to loadbalancer template

JargoonPard:
Minor change to enable dynamic IP address allocation.

@spinnaker/netflix-reviewers @spinnaker/google-reviewers PTAL

PR #458, opened 2016-03-22
Add frontendpool to loadbalancer template

duftler:


duftler:
Cherry-pick commit from #457.

PR #459, opened 2016-03-22
provider/kubernetes: Block until kubernetes state is consistent

lwander:
@ttomsu 

When a destructive Kubernetes operation completes (excluding destroy), there is a chance that the state the user has requested has not propagated throughout the entire cluster yet. 

To see why this is dangerous, imagine a user does the following in a pipeline:

Resize RC to 0 -> Destroy RC

Meanwhile, the resize completes, however the RC's state is not reflected in the entire k8s cluster. When the destroy happens, it might destroy the RC while it still has some non-zero size in some part of the k8s cluster. The pods are orphaned, and resources are leaked.

To avoid this, I implemented an exponential backoff strategy waiting for the generation numbers of the requested and current kubernetes objects to match.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
@@ -23,14 +23,44 @@ import io.fabric8.kubernetes.api.model.*
 import io.fabric8.kubernetes.api.model.extensions.Ingress
 import io.fabric8.kubernetes.client.KubernetesClient
 
+import java.awt.image.ReplicateScaleFilter
+import java.util.concurrent.TimeUnit
+
 class KubernetesApiAdaptor {
   KubernetesClient client
+
+  static final int RETRY_COUNT = 20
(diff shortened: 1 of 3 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
```diff
@@ -33,6 +40,29 @@ class KubernetesApiAdaptor {
...
+    def current = getReplicationController(desired.metadata.namespace, desired.metadata.name)
+
+    def wait = RETRY_INITIAL_WAIT_MILLIS
+    def attempts = 0
+    while (current.status.observedGeneration < desired.status.observedGeneration) {
+      attempts += 1
+      if (attempts > RETRY_COUNT) {
+        return false
+      }
+      sleep(wait)
```
ttomsu: nit: put the wait assignment next to this statement (group the same variable usage)

ttomsu:
LGTM

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
lwander: SGTM

PR #460, opened 2016-03-22
provider/kubernetes: Provide missing instances to kubernetes load balancers

lwander:
@ttomsu - addresses the TODO you left me after you cleaned up the load balancer providers.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesLoadBalancer.groovy
```diff
@@ -54,15 +54,28 @@ class KubernetesLoadBalancer implements LoadBalancer, Serializable {
...
+                health: [
+                    state: instance.healthState.toString()
+                ]
+            )
+          } else {
+            return (LoadBalancerInstance) null // Groovy generics need to be convinced all control flow paths return the same object type
+          }
+        } as Set,
+        detachedInstances: serverGroup?.instances?.findResults { instance ->
+          if (!instance.isAttached(this.name)) {
```
ttomsu: return instance.isAttached(name) ? null : instance.name

ttomsu:
LGTM

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesLoadBalancer.groovy
lwander: I kept the code structure in line with the chunk above, to make it clearer that this effectively partitions the set of instances.

PR #461, opened 2016-03-22
Adds retry support when calling edda for AWS resources

cfieber:
@spinnaker/netflix-reviewers ptal

tested by unplugging the network cable while it was running...


ajordens:
`EDDA_RETRY_BACKOFF_MILLIS` seems tight to me ... why not just do `retryAttempts * EDDA_RETRY_BASE_MILLIS`?

Do we really care about 1..10ms vs. 50ms? 

Any fear of going with much lower socket/connect timeouts?

cfieber:
I was mainly trying to introduce some jitter into the backoff.

Could probably go as low as 1 second for the connect timeout, and could probably go lower on the socket timeout, but would basically be guessing there. I'll experiment with some lower values

cfieber:
did some testing with lower socket timeout and it started tripping on the larger payloads (images mainly)

I moved all the config into a bean that is `@ConfigurationProperties` bound so it is tunable

PR #462, opened 2016-03-22
provider/kubernetes: Load balancer description attribute

lwander:
The idea is to retrieve the operation that would have created this load balancer, making it easier to edit a load balancer in deck, and unifying the data models within the kubernetes provider. 

@ttomsu or @duftler

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
@@ -18,6 +18,8 @@ package com.netflix.spinnaker.clouddriver.kubernetes.api
 
 import com.netflix.frigga.Names
 import com.netflix.spinnaker.clouddriver.kubernetes.deploy.KubernetesUtil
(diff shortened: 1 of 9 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
```diff
@@ -177,6 +179,39 @@ class KubernetesApiAdaptor {
...
+    loadBalancerDescription.serviceType = service.spec.type
+
+    loadBalancerDescription.externalIps = service.spec.externalIPs ?: []
+    loadBalancerDescription.ports = service.spec.ports?.collect { port ->
+      new KubernetesNamedServicePort(
+          name: port.name,
+          protocol: port.protocol,
+          port: port.port ?: 0,
+          targetPort: port.targetPort?.intVal ?: 0,
+          nodePort: port.nodePort ?: 0
```
ttomsu: Since these are just declared as `int`s, they'll default to 0 without the need of the elvis operator

ttomsu: Nevermind my previous comment

ttomsu:
LGTM

PR #464, opened 2016-03-22
Reduce aws image caching

cfieber:
Right now we index the same public images in every account clouddriver manages.

This change updates it to only index them once per region - they will still show up in the search dropdown but in our case this eliminates on the order of 1.5 million keys in redis.

This is a step towards image caching cleanup that will only index the images actually owned by an account, in each account, but that is a bit more invasive as the Key structure for IMAGES and NAMED_IMAGES has to change to remove the account name component.

The three commits in this change all stand alone so I wasn't intending to squash this except to address any feedback towards the last commit into that commit


ajordens:
LGTM

PR #467, opened 2016-03-23
provider/kubernetes: Prepare for deploying spinnaker with spinnaker

lwander:
These additions allow Spinnaker to be deployed on Kubernetes using Spinnaker 

Most important was the addition of secrets as volume sources, which is how we will be deploying configuration.

@duftler or @ttomsu


lwander:
ping @ttomsu

ttomsu:
LGTM

PR #469, opened 2016-03-23
get security group and skip cache if we need to on upsert

claymccoy:
In the uncommon case that the security group we are upserting appears between the time that we look for it and create it, we will get an AWS error. We now look again skipping the cache.  We don't want to skip the cache here in the normal case because it usually isn't required and has throttled AWS when creating many security groups at once during a VPC migration in prod.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/SecurityGroupLookupFactory.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/SecurityGroupLookupFactory.groovy
@@ -63,18 +63,18 @@ class SecurityGroupLookupFactory {
       if (description.vpcId) {
         request.withVpcId(description.vpcId)
       }
-      final amazonEC2 = amazonClientProvider.getAmazonEC2(credentials, region)
+      final amazonEC2 = amazonClientProvider.getAmazonEC2(credentials, region, true)
       final result = amazonEC2.createSecurityGroup(request)
       final newSecurityGroup = new SecurityGroup(ownerId: credentials.accountId, groupId: result.groupId,
         groupName: description.name, description: description.description, vpcId: description.vpcId)
       new SecurityGroupUpdater(newSecurityGroup, amazonEC2)
     }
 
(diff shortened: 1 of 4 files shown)


claymccoy:
@cfieber This will fix chex.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/SecurityGroupLookupFactory.groovy
```diff
@@ -63,18 +63,18 @@ class SecurityGroupLookupFactory {
...
-      final amazonEC2 = amazonClientProvider.getAmazonEC2(credentials, region)
+      final amazonEC2 = amazonClientProvider.getAmazonEC2(credentials, region, true)
       final result = amazonEC2.createSecurityGroup(request)
       final newSecurityGroup = new SecurityGroup(ownerId: credentials.accountId, groupId: result.groupId,
         groupName: description.name, description: description.description, vpcId: description.vpcId)
       new SecurityGroupUpdater(newSecurityGroup, amazonEC2)
     }
 
-    SecurityGroupUpdater getSecurityGroupByName(String accountName, String name, String vpcId) {
+    SecurityGroupUpdater getSecurityGroupByName(String accountName, String name, String vpcId, skipCache = false) {
```
cfieber: want to stick a type on this method signature

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/SecurityGroupLookupFactory.groovy
```diff
@@ -83,9 +83,9 @@ class SecurityGroupLookupFactory {
       null
     }
 
-    private List<SecurityGroup> getSecurityGroups(String accountName, AmazonEC2 amazonEC2) {
+    private List<SecurityGroup> getSecurityGroups(String accountName, AmazonEC2 amazonEC2, skipCache) {
```
cfieber: and this one

cfieber:
LGTM after nits addressed, thanks!

PR #471, opened 2016-03-24
Removes account from AWS image and named image cache key.

cfieber:
AMI Id uniquely identifies an image within a region. This change just hardcodes a value where previously
we would put the account name (allows for compatible key parsing between versions older than this change).

Images are now only cached by the image caching agent if they are owned by that agent's account.


ajordens:
Needs a rebase if we're still thinking this would be beneficial.

anotherchrisberry:
ping

cfieber:
notdeadyet

ajordens:
Thoughts on reviving this thing ... looks close.

Would complete my image tagging work (as I'd really only like to tag in the owning account)

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ImageCachingAgent.groovy
```diff
@@ -102,10 +102,10 @@ class ImageCachingAgent implements CachingAgent, AccountAware, DriftMetric {
     Collection<CacheData> namedImageCacheData = new ArrayList<>(images.size())
 
     for (Image image : images) {
-      if (includePublicImages || !image.public) {
+      if (includePublicImages || !image.public || image.ownerId == account.accountId) {
```
ajordens: Might need some love in the case where `clouddriver` isn't indexing the owning account.

cfieber:
Yeah I want to get this in - need to do some testing mainly to make sure things behave ok (and it was waiting for the dual mode redis to handle to key format change)

ajordens:
Bump bump.

ewiseblatt:
please reap this.

PR #472, opened 2016-03-24
provider/kubernetes: Namespace create bug

lwander:
Originally, I was only recreating namespaces that were missing the cluster health mechanism. If a docker registry specifies a namespace it can deploy to that doesn't exist, the credential initialization fails.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesCredentials.java
```diff
@@ -59,6 +58,14 @@ public KubernetesCredentials(KubernetesApiAdaptor apiAdaptor,
 
       for (int j = 0; j < affectedNamespaces.size(); j++) {
```
duftler: Since you're not using the index elsewhere, why not use a for-each loop here?
`for (String inNamespace : affectedNamespaces)`

duftler:

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesCredentials.java
lwander: SGTM

PR #475, opened 2016-03-25
GCE OnDemandAgents + GCE LB display issues

ttomsu:
@lwander - note the comment in the code that includes your idea to make on demand caching simpler. These impls are good enough, IMO, until that's ready.

@duftler, @jtk54 FYI 


lwander:
LGTM

PR #476, opened 2016-03-27
provider/kubernetes: Move translation of resources out of adaptor

lwander:
@ttomsu pointed out bloat in the `KubernetesApiAdaptor` class, so I pulled it apart into `KubernetesApiAdaptor` which acts as a proxy to make API calls through, and `KubernetesApiConverter` which translates API resources into the resource descriptions used by Deck and Clouddriver to create/edit server groups, security groups, and load balancers.


duftler:
LGTM.

PR #479, opened 2016-03-28
Support for specifying an explicit ASG sequence number

ajordens:
This handles the scenario where an API user would like to bulk create (or clone) multiple ASGs


ajordens:

FYI @spinnaker/google-reviewers, not sure if this is something you'd like to support in GCE. Specifically, we have an internal API user that needs/would like to create multiple ASGs simultaneously. 

I believe exposing the `sequence` number was far simpler than the alternative of supporting either a batch mode in the existing clone operation or a brand new batch-capable operation.

duftler:
Seems like a good idea. We'll do the same for GCE.

lwander:
And Kubernetes!

PR #480, opened 2016-03-28
Fix NoSuchElementException from GCE on demand cache handler

ttomsu:
citest's Google server group, kato, and smoke tests all pass without any a stack traces after these changes. 

@duftler @lwander


lwander:
LGTM

PR #481, opened 2016-03-28
provider/kubernetes: Fix up error reporting

lwander:
Previously, exceptions from the client library were being suppressed in the logging, and were too verbose in the UI. This fixes that by logging the exceptions, and only propagating the "message" field of the client exception forward.

@duftler @ttomsu


ttomsu:
LGTM

PR #484, opened 2016-03-29
provider/kubernetes: Trying to catch a groovy string bug

lwander:
A user has noticed this code path failing due to a gstring conversion error that I can't seem to replicate. Hopefully this fixes it.

@duftler


duftler:
LGTM.

PR #485, opened 2016-03-29
Turn on GCE Redis-based provider impl

ttomsu:
The day is finally here! This turns on the new Redis based caching implementation for GCE. 

Obviously if you're having troubles, flip this value back to "old" and file a bug report, assigning @ttomsu. 

@spinnaker/google-reviewers FYI


duftler:
Nice.
LGTM.

PR #488, opened 2016-03-30
Fixes issue with order of GCE httpHealthCheck and instance health che…

ttomsu:
…ck calls.

@lwander - I'm able to see load balancer health after this fix. Nice find.

@spinnaker/google-reviewers FYI


lwander:
LGTM - really @ewiseblatt's testing & your searching found it ;)

PR #489, opened 2016-03-30
Add destroy server group ops and other minor bug fixes

milismsft:
Add destroy server group ops. 
Cleanup resources that are created as part of the create server group in case of a failure during the creation.
Add retries support when deleting a resource in Azure (work around for random timeouts when accessing Azure resources in the Azure Cloud).
Fix an exception in clouddriver console when the application has no clusters.
Add type as part of the deployment name.

### DIFF
--- a/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureComputeClient.groovy
+++ b/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureComputeClient.groovy
@@ -20,7 +20,9 @@ import com.microsoft.azure.CloudException
 import com.microsoft.azure.credentials.ApplicationTokenCredentials
 import com.microsoft.azure.management.compute.ComputeManagementClient
 import com.microsoft.azure.management.compute.ComputeManagementClientImpl
+import com.microsoft.azure.management.compute.VirtualMachineScaleSetsOperations
 import com.microsoft.azure.management.compute.models.VirtualMachineScaleSet
+import com.microsoft.rest.ServiceResponse
 import com.netflix.spinnaker.clouddriver.azure.resources.servergroup.model.AzureInstance
 import com.netflix.spinnaker.clouddriver.azure.resources.servergroup.model.AzureServerGroupDescription
 import com.netflix.spinnaker.clouddriver.azure.resources.vmimage.model.AzureVMImage
@@ -152,6 +154,20 @@ public class AzureComputeClient extends AzureBaseClient {
       log.warn("ServerGroup: ${e.message} (${serverGroupName} was not found)")
     }
     null
+  }
+
+  ServiceResponse<Void> destroyServerGroup(String resourceGroupName, String serverGroupName) {
+    VirtualMachineScaleSetsOperations ops = getAzureOps(
+      client.&getVirtualMachineScaleSetsOperations, "Get operations object", "Failed to get operation object") as VirtualMachineScaleSetsOperations
+
+    deleteAzureResource(
+      ops.&delete,
+      resourceGroupName,
+      serverGroupName,
+      null,
+      "Delete Server Group ${serverGroupName}",
+      "Failed to delete Server Group ${serverGroupName} in ${resourceGroupName}"
+    )
   }
 
   Collection<AzureInstance> getServerGroupInstances(String resourceGroupName, String serverGroupName) {
--- a/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureNetworkClient.groovy
+++ b/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureNetworkClient.groovy
@@ -18,8 +18,12 @@ package com.netflix.spinnaker.clouddriver.azure.client
 
 import com.microsoft.azure.CloudException
 import com.microsoft.azure.credentials.ApplicationTokenCredentials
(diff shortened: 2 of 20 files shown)


milismsft:

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureComputeClient.groovy
```diff
@@ -154,6 +155,38 @@ public class AzureComputeClient extends AzureBaseClient {
     null
   }
 
+  ServiceResponse<Void> destroyServerGroup(String resourceGroupName, String serverGroupName) {
+    ServiceResponse<Void> result = null
+    // The API call might return a timeout exception or some other Azure CloudException that is not
+    //   related to the operation we are trying to execute; retry couple of times and if the final
+    //   retry fails then rethrow
+    long operationRetry = 0
+    while (operationRetry < AZURE_ATOMICOPERATION_RETRY) {
```
ajordens: You probably want to introduce a sleep between cycles, and perhaps some logging to know that retries are happening.

See `AmazonClientInvocationHandler` for an example of how we've done this similarly.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureNetworkClient.groovy
```diff
@@ -184,13 +184,83 @@ class AzureNetworkClient extends AzureBaseClient {
...
     def publicIpAddressName = AzureUtilities.getResourceNameFromID(loadBalancer.frontendIPConfigurations.first().getPublicIPAddress().id)
-    client.getLoadBalancersOperations().delete(resourceGroupName, loadBalancerName)
 
-    client.getPublicIPAddressesOperations().delete(resourceGroupName, publicIpAddressName)
+    ServiceResponse<Void> result = null
+    // The API call might return a timeout exception or some other Azure CloudException that is not
+    //   related to the operation we are trying to execute; retry couple of times and if the final
+    //   retry fails then rethrow
+    long operationRetry = 0
+    while (operationRetry < AZURE_ATOMICOPERATION_RETRY) {
```
ajordens: Looks like this behavior is used in a handful of places, I'd suggest a method (taking a closure) that wraps the retry behavior and just calls out to the closure to actually execute the client call.

This is primarily towards eliminating copy & pasted code.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureResourceManagerClient.groovy
```diff
@@ -62,12 +62,13 @@ class AzureResourceManagerClient extends AzureBaseClient {
...
                                                 String region,
                                                 String resourceName,
+                                                String resourceType,
                                                 Map<String, String> templateParams = [:]) {
 
     // TODO validate that all callers invoke this themselves, then remove this call
     initializeResourceGroupAndVNet(credentials, resourceGroupName, null, region)
 
-    String deploymentName = resourceName + AzureUtilities.NAME_SEPARATOR +"deployment"
+    String deploymentName = resourceName + AzureUtilities.NAME_SEPARATOR + resourceType + AzureUtilities.NAME_SEPARATOR +"deployment"
```
ajordens: Not a big deal but `[resourceName, resourceType, "deployment"].join(AzureUtilities.NAME_SEPARATOR]` would be another way to write this.

ajordens:
Few comments but this is all within azure packages so LGTM.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureComputeClient.groovy
milismsft: Makes sense; I can add the sleeps in between the calls.
BTW can you point me into AmazonClientInvocationHandler what method should I check? I did not notice anyone that has similar retries.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureNetworkClient.groovy
milismsft: I've made a note to refactor these into one single method; I've initially started with something much simpler than added the extra lines.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureResourceManagerClient.groovy
milismsft: Makes sense; I'll do the change.

milismsft:
@spinnaker/netflix-reviewers @ajordens  PTAL

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureComputeClient.groovy
milismsft: @ajordens I moved most of the duplicated code that handles the retries into the base class; the two methods will retry on both getting the operation object and the actual delete op. 
In time I will continue to build on top of that and change some more operations to use the same pattern (see get like methods).

ajordens:
Needs a quick little rebase

milismsft:
@ajordens  rebase done; can you help with merging it in?

PR #492, opened 2016-03-30
Introduced a ReservedInstances caching agent (runs every 10 minutes)

ajordens:
- Updated ReservationReport caching agent to use the cached ReservedInstances data rather than calling AWS directly
- Adjusted ReservationReport caching agent to only run every minute

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservationReportCachingAgent.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservationReportCachingAgent.groovy
@@ -66,6 +66,7 @@ class ReservationReportCachingAgent implements CachingAgent, CustomScheduledAgen
   ])
 
   private final Scheduler scheduler
+  private final ApplicationContext ctx
   private Cache cacheView
 
 
@@ -74,13 +75,12 @@ class ReservationReportCachingAgent implements CachingAgent, CustomScheduledAgen
   final ObjectMapper objectMapper
   final AccountReservationDetailSerializer accountReservationDetailSerializer
 
-  @Autowired
-  ApplicationContext ctx
 
   ReservationReportCachingAgent(AmazonClientProvider amazonClientProvider,
                                 Collection<NetflixAmazonCredentials> accounts,
                                 ObjectMapper objectMapper,
-                                Scheduler scheduler) {
+                                Scheduler scheduler,
+                                ApplicationContext ctx) {
     this.amazonClientProvider = amazonClientProvider
     this.accounts = accounts
 
@@ -90,6 +90,7 @@ class ReservationReportCachingAgent implements CachingAgent, CustomScheduledAgen
 
     this.objectMapper = objectMapper.copy().enable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS).registerModule(module)
     this.scheduler = scheduler
+    this.ctx = ctx
   }
 
   @Override
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservationReportCachingAgent.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservationReportCachingAgent.groovy
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.clouddriver.aws.provider.agent
 
 import com.amazonaws.services.ec2.model.DescribeInstancesRequest
 import com.amazonaws.services.ec2.model.DescribeReservedInstancesRequest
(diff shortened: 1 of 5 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservationReportCachingAgent.groovy
```diff
@@ -253,6 +287,13 @@ class ReservationReportCachingAgent implements CachingAgent {
     }
   }
 
+  private Cache getCacheView() {
```
ajordens: `AmazonLoadBalancerInstanceStateCachingAgent` is also doing this to late bind to `Cache`.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservationReportCachingAgent.groovy
```diff
@@ -184,8 +210,16 @@ class ReservationReportCachingAgent implements CachingAgent {
         log.info("Fetching reservation report for ${credentials.name}:${region.name}")
 
         def amazonEC2 = amazonClientProvider.getAmazonEC2(credentials, region.name)
-        def reservedInstancesResult = amazonEC2.describeReservedInstances(new DescribeReservedInstancesRequest())
-        reservedInstancesResult.reservedInstances.findAll {
+
+        def cacheView = getCacheView()
+        def reservedInstances = cacheView.getAll(
```
ajordens: There is some chance that on the very first cycle that the `ReservedInstancesCachingAgent` won't have completed yet for each region.

I anticipate pre-seeding the redis (by running the **ONLY** the `ReservedInstancesCachingAgent` locally) and just rolling out as normal.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservedInstancesCachingAgent.groovy
```diff
@@ -0,0 +1,123 @@
...
+import org.slf4j.LoggerFactory
+
+import java.util.concurrent.TimeUnit
+
+import static com.netflix.spinnaker.cats.agent.AgentDataType.Authority.AUTHORITATIVE
+import static com.netflix.spinnaker.clouddriver.aws.data.Keys.Namespace.RESERVED_INSTANCES
+
+class ReservedInstancesCachingAgent implements CachingAgent, CustomScheduledAgent, AccountAware {
+  private static final TypeReference<Map<String, Object>> ATTRIBUTES = new TypeReference<Map<String, Object>>() {}
+  private static final long DEFAULT_POLL_INTERVAL_MILLIS = TimeUnit.MINUTES.toMillis(10)
```
ajordens: Didn't see a need to run this super frequently.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservationReportCachingAgent.groovy
```diff
@@ -31,51 +32,51 @@ import com.netflix.spinnaker.cats.agent.AgentDataType
...
-  Collection<AgentDataType> getProvidedDataTypes() {
-    types
-  }
 
   final AmazonClientProvider amazonClientProvider
   final Collection<NetflixAmazonCredentials> accounts
   final ObjectMapper objectMapper
   final AccountReservationDetailSerializer accountReservationDetailSerializer
 
+  @Autowired
```
cfieber: why autowire this instead of just constructor injecting it like the rest of its dependencies? I'm pretty sure ApplicationContext is available where this thing gets instantiated

ajordens:
With this PR, I see consistent < 20s invocation times of the `ReservationReportCachingAgent`.

I've spot checked the generated report against production and see no discernible difference.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservationReportCachingAgent.groovy
ajordens: Ya, that is silly and just borrowed from `AmazonLoadBalancerInstanceStateCachingAgent`.

I'll fix both.

cfieber: could probably actually just inject the Cache directly

ajordens: I didn't chase it down but this is all trying to avoid injecting the Cache that is in the middle of being constructed.

PR #495, opened 2016-04-01
filter instance relationships for significant performance improvement on large server groups

cfieber:


ajordens:
LGTM

PR #496, opened 2016-04-01
Fix block device mappings when cloning asg

ajordens:


ajordens:
Verified by cloning from c3.2xlarge -> c4.2xlarge -> c3.2xlarge and seeing the block device mappings get updated to the correct instance-specific defaults.

ajordens:
PTAL @cfieber or @claymccoy

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
```diff
@@ -80,5 +81,7 @@ class BasicAmazonDeployDescription extends AbstractAmazonCredentialsDescription
     String region
     String asgName
     Boolean useSourceCapacity
+
+    LaunchConfiguration launchConfiguration
```
cfieber: this doesn't feel like the right object to include in our API model class

cfieber: we already pull the source launch config (I think) during deploy, so we should be able to get at it without attaching it in here I think

cfieber: https://github.com/ajordens/clouddriver/blob/fix-block-device-mappings-when-cloning-asg/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy#L293 right before convertBlockDevices

ajordens: Agreed ... I'll move `buildBlockDeviceMappings` there and parameterize in the `sourceLaunchConfiguration`.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
```diff
@@ -295,6 +296,8 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
     )
 
     description.blockDevices = description.blockDevices != null ? description.blockDevices : convertBlockDevices(sourceLaunchConfiguration.blockDeviceMappings)
+    description.blockDevices = buildBlockDeviceMappings(description, sourceLaunchConfiguration)
```
ajordens: Will find a cleaner solution for this later tonight when I get home.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/LoggingInstrumentation.groovy
```diff
@@ -33,7 +33,8 @@ class LoggingInstrumentation implements ExecutionInstrumentation {
 
   @Override
   void executionCompleted(Agent agent) {
-    logger.info("${agent.providerName}:${agent.agentType} completed")
+    def durationMs = Math.round((System.nanoTime() - MetricInstrumentation.getAgentStartTimeNs(agent)) / 1000000)
```
cfieber: I'd probably opt for TimeUnit.NANOSECONDS.toMillis here

cfieber: I'm not sure there is a solid guarantee that this executionCompletion is going to happen before the MetricInstrumentation executionCompletion (which would have cleared the startTime out of the ThreadLocal)

cfieber: I'd be pretty ok with changing ExecutionInstrumentation callback to get passed the elapsed time and have the AgentScheduler runnable deal with tracking the elapsed time and killing off that whole dirty threadlocal

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/LoggingInstrumentation.groovy
```diff
@@ -33,7 +35,8 @@ class LoggingInstrumentation implements ExecutionInstrumentation {
 
   @Override
   void executionCompleted(Agent agent) {
-    logger.info("${agent.providerName}:${agent.agentType} completed")
+    def durationMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - MetricInstrumentation.getAgentStartTimeNs(agent))
```
cfieber: I'm not sure there is a solid guarantee that this executionCompletion is going to happen before the MetricInstrumentation executionCompletion (which would have cleared the startTime out of the ThreadLocal)
I'd be pretty ok with changing ExecutionInstrumentation callback to get passed the elapsed time and have the AgentScheduler runnable deal with tracking the elapsed time and killing off that whole dirty threadlocal

cfieber:
LGTM

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/LoggingInstrumentation.groovy
ajordens: Ya, I haven't seen it yet run with a cleared ThreadLocal but that could just be happenstance locally.

I'm just gassing that ThreadLocal, seems like the sensible thing to do.

PR #497, opened 2016-04-01
Fix the setting of the serverGroup.disabled property to take into acc…

duftler:
…ount whether there are load balancers defined in the server group metadata.


duftler:
@ttomsu please review.

PR #501, opened 2016-04-03
ReservedInstances caching agent only caches 'active' RIs

ajordens:
- Additional logging in ReservationReport caching agent


cfieber:
LGTM

PR #503, opened 2016-04-03
Adds NodeStatusProvider for ClusteredAgentScheduler.

cfieber:
ClusteredAgentScheduler will only run agents if the node is enabled.

Adds EurekaStatusNodeStatusProvider to determine node status from eureka status.


ajordens:
LGTM

PR #504, opened 2016-04-04
provider/kubernetes: Fix image pull secret address

lwander:
Using the full v2 endpoint was preventing Kubernetes from correctly authenticating with some private registries

@duftler


duftler:
LGTM.

PR #505, opened 2016-04-04
Fix deserialization error when moving from on demand namespace.

ttomsu:
Copied from Kubernetes impl.

@lwander @jtk54 PTAL


lwander:
LGTM

jtk54:
LGTM also

PR #506, opened 2016-04-04
Consider serverGroup.disabled when returned server group disabled/ena…

duftler:
…bled status as part of load balancer view.
@ttomsu please review.


ttomsu:
I guess my assertion proved to be false. D'oh! LGTM.

PR #510, opened 2016-04-06
Use 'region' instead of 'zone' on Cloud Foundry server group operations.

duftler:


duftler:
@gregturn please review.

gregturn:
LGTM

PR #511, opened 2016-04-06
Move Eureka out of AWS and set as external Health Provider

tomaslin:
This PR comes from a need to make eureka available to Titus as well as AWS.

The changes are as follows:

* The HealthProvidingCachingAgent class is now part of core
* There is a new clouddriver-eureka package
* EurekaCachingAgent will reach into each provider to get keys via the EurekaAwareProvider interface
* Instance and cluster health endpoints will check against the ExternalHealthProvidingAgent


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/config/AwsProviderConfig.groovy
```diff
@@ -144,33 +138,20 @@ class AwsProviderConfig {
             )
           }
         }
-
         if (credentials.discoveryEnabled) {
-          discoveryAccounts[credentials.discovery][region.name].add(credentials)
+          awsProvider.addEurekaCredential(credentials)
```
lwander: What does it mean for eureka to need credentials?

tomaslin: it's probably not the right name, but this translates the AWS credential name from the accountId field in the discovery entry

tomaslin: addCredentialNameForEurekaLookup?

clouddriver-eureka/clouddriver-eureka.gradle
```diff
@@ -0,0 +1,3 @@
+dependencies {
```
lwander: What are your thoughts on making this module contain all supported service discovery types (starting with eureka), since there is some interest here at Google to add support for Consul

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/config/AwsProviderConfig.groovy
lwander: Oh I see. Name is up to you

lwander:
>  The HealthProvidingCachingAgent class is now part of core

Is it necessary for other providers to conform to this interface to provide health, or is it only needed to allow external health providers to provide health?

tomaslin:
Any other providers that want to provide health would need to conform to this, EddaLoadBalancingCachingAgent and AmazonLoadBalancerInstanceStateCachingAgent for example are not external agents

clouddriver-eureka/clouddriver-eureka.gradle
tomaslin: not sure what the advantage of having consul being in here would be. If anything, would expect different libraries to be self-contained so the dependency doesn't exist amongst them.

lwander: That makes sense

lwander:
LGTM

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/AwsProvider.groovy
```diff
@@ -134,4 +130,34 @@ class AwsProvider extends AgentSchedulerAware implements SearchableProvider {
       return cacheView.getAll(INSTANCES.ns, possibleInstanceIdentifiers)*.id
     }
   }
+
+  // Eureka provider support
+  private Map<String, String> accounIdToAccountNameLookup = [:]
```
cfieber: I would kill this off and just make the supporting methods use AccountCredentialRepository. There are a small enough number of accounts in a system that linearly searching them for every call is not going to be significant. There is support for reloading the configuration that this would have to support otherwise

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/config/AwsProviderConfig.groovy
```diff
@@ -71,21 +67,21 @@ class AwsProviderConfig {
...
-                           amazonCloudProvider,
-                           amazonClientProvider,
-                           accountCredentialsRepository,
-                           objectMapper,
-                           discoveryApiFactory,
-                           eddaApiFactory,
-                           ctx,
-                           registry,
-                           reservationReportScheduler)
+      amazonCloudProvider,
```
cfieber: would prefer not to see code reformatting unrelated to the change at hand in PRs

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/AwsProvider.groovy
```diff
@@ -134,4 +130,34 @@ class AwsProvider extends AgentSchedulerAware implements SearchableProvider {
...
+
+  @Override
+  Boolean isProviderForEurekaRecord(Map<String, Object> attributes) {
+    attributes.containsKey('accountId') && attributes.get('titusTaskId') == null
```
cfieber: can we do something where this class doesn't have to know about `titusTaskId`? an `@Ordered` set of `EurekaAwareProvider` and the first one that returns true for `isProviderForEurekaRecord` wins?

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/EurekaProviderConfiguration.groovy
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.context.annotation.PropertySource
+import org.springframework.context.annotation.Scope
+
+import java.util.regex.Pattern
+
+@Configuration
+@EnableConfigurationProperties
+@ConditionalOnProperty('eureka.provider.enabled')
+@ComponentScan(["com.netflix.spinnaker.clouddriver.eureka"])
+@PropertySource(value = "classpath:META-INF/clouddriver-core.properties", ignoreResourceNotFound = true)
```
cfieber: do we need this here? I think it was used in the google provider to get `Implementation-Version` to pass along with a user-agent header

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/EurekaProviderConfiguration.groovy
```diff
@@ -0,0 +1,70 @@
...
+  @ConfigurationProperties("eureka.provider")
+  EurekaAccountConfigurationProperties eurekaConfigurationProperties() {
+    new EurekaAccountConfigurationProperties()
+  }
+
+  @Autowired
+  ObjectMapper objectMapper
+
+  @Autowired
+  EurekaApiFactory discoveryApiFactory
```
cfieber: rename to eurekaApiFactory

cfieber: can we replace all `discovery` with `eureka` - at the very least within this module

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/EurekaProviderConfiguration.groovy
```diff
@@ -0,0 +1,70 @@
...
+
+  @Bean
+  EurekaCachingProvider eurekaCachingProvider(EurekaAccountConfigurationProperties eurekaAccountConfigurationProperties, List<EurekaAwareProvider> eurekaAwareProviderList) {
+    List<EurekaCachingAgent> agents = []
+    eurekaAccountConfigurationProperties.accounts.each { EurekaAccountConfigurationProperties.EurekaAccount accountConfig ->
+      accountConfig.regions.each { region ->
+        String discoveryHost = accountConfig.readOnlyUrl.replaceAll(Pattern.quote('{{region}}'), region)
```
cfieber: what was the thought around pulling the region templating handling up to here? if you want to keep it here, then I think `region` is no longer required on the createApi method

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/api/EurekaConfig.groovy
```diff
@@ -36,8 +36,8 @@ class DiscoveryConfig {
   }
 
   @Bean
-  DiscoveryApiFactory discoveryApiFactory(Converter discoveryConverter) {
-    new DiscoveryApiFactory(discoveryConverter)
+  EurekaApiFactory discoveryApiFactory(Converter discoveryConverter) {
```
cfieber: `s/discovery/eureka/g`

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/model/discovery/EurekaInstance.groovy
```diff
@@ -97,7 +99,8 @@ class DiscoveryInstance implements Health {
       vipAddress,
       secureVipAddress,
       lastUpdatedTimestamp,
-      asgName)
+      asgName,
+      metadata?.titusTaskId)
```
cfieber: wonder if we should just have `EurekaInstance` have the whole Metadata and DataCenterInfo maps so that we don't have to encode provider specific values into this object

cfieber:
this would have been better as two PRs - one to convert Titus instance caching to remove account/region and just use taskId as the key, and the other to extract the Eureka bit into its own module. Not suggesting you change it now but in the future one concern per PR would be handy

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/AwsProvider.groovy
tomaslin: can just check for accountId since titus discovery entries don't have one

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/EurekaProviderConfiguration.groovy
tomaslin: done

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/EurekaProviderConfiguration.groovy
tomaslin: the old way was deriving regions from the account settings, but we're loading them independently here.

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/model/discovery/EurekaInstance.groovy
tomaslin: will do this at a later PR since Jackson doesn't make it easy and don't want to waste too much time on this

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/AwsProvider.groovy
```diff
@@ -134,4 +133,28 @@ class AwsProvider extends AgentSchedulerAware implements SearchableProvider {
       return cacheView.getAll(INSTANCES.ns, possibleInstanceIdentifiers)*.id
     }
   }
+
+  // Eureka provider support
+  @Override
+  Boolean isProviderForEurekaRecord(Map<String, Object> attributes) {
+    attributes.containsKey('accountId')
```
cfieber: worth checking that we have an existing account for this id as well?

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/model/EurekaInstance.groovy
```diff
@@ -97,7 +99,8 @@ class DiscoveryInstance implements Health {
       vipAddress,
       secureVipAddress,
       lastUpdatedTimestamp,
-      asgName)
+      asgName,
+      metadata?.titusTaskId)
```
cfieber: still feels like it would be nice to not have titusTaskId defined in here (and I'd be ok with removing all the AWS specific bits in 'dataCenterInfo.metadata' too) - maybe we just add a metadata attribute to EurekaInstance that is set to dataCenterInfo.metadata + metadata ?  Since it gets converted back to a map when passed to the provider interface it should mostly work the same but with less cloud-provider specific opinions in this module.

wdyt?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/AwsProvider.groovy
tomaslin: this is done later when we retrieve the instanceKey

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/model/EurekaInstance.groovy
tomaslin: goes away when we use instanceId for titus but worth considering datacenterinfo.metadata

cfieber:
LGTM

PR #515, opened 2016-04-07
Hooks up requiredGroupMembership to Google credentials

ttomsu:
@duftler @jtk54 PTAL


duftler:
Looks like the constructor that was being called before this change is no longer being used anywhere. If that's in fact the case, please kill it off as part of this PR.
LGTM after addressing the one comment.

duftler:
Ping @ttomsu .

PR #516, opened 2016-04-08
core: Add process & job providers & models

lwander:
This complements https://github.com/spinnaker/clouddriver/pull/513, allowing jobs to be queried. I'm representing jobs as composed of "processes", borrowing language from https://github.com/spinnaker/spinnaker/issues/818. Since "processes" are effectively containers, they share the same `HealthState` enum with VMs and containers running in server groups. Jobs have a separate notion of success/failure depending on the state of the processes they manage. 

I'll likely be adding to this as in separate PRs as I decide what needs to be forwarded to Deck to visualize Jobs and Processes. 

@tomaslin @duftler could you PTAL?

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/Job.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/Job.groovy
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.model
+
+/**
+ * A Job is a collection of Processes. The Job defines how many processes may run, and the failure/success conditions
+ * leading to a terminal status.
+ *
+ * See Process.groovy for an overview of Processes.
+ */
+interface Job {
+  String getName()
+
+  String getAccount()
+
+  String getId()
+
+  String getLocation()
+
+  JobState getJobState()
+
+  Set<Process> getProcesses()
+
+  Long getLaunchTime()
+
+  Long getFinishTime()
+
+  Set<String> getLoadBalancers()
+
+  Set<String> getSecurityGroups()
+}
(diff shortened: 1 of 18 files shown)


duftler:
Would be a good time to add some additional swagger annotations.
LGTM.
@tomaslin any concerns?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/JobProvider.groovy
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.model
+
+public interface JobProvider<T extends Job> {
+  String getPlatform()
+
+  T getJob(String account, String location, String id)
```
tomaslin: should account then be a field in Job and Process?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/Job.groovy
```diff
@@ -0,0 +1,35 @@
...
+ * leading to a terminal status.
+ *
+ * See Process.groovy for an overview of Processes.
+ */
+interface Job {
+  String getName()
+
+  String getId()
+
+  String getLocation()
```
tomaslin: if a job is a collection of processes, should I not be able to get a list of processes from a job?

lwander: Yup, I figured I'd be adding these fields as I realized I needed them when implementing the Kubernetes job/process providers, but I've added them now.

lwander:
I added job & process controllers for Kubernetes, as well as the swagger endpoints. @duftler

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/JobController.groovy
```diff
@@ -0,0 +1,69 @@
...
+
+  @Autowired
+  MessageSource messageSource
+
+  @ApiOperation(value = "Get a Job", notes = "Composed of many running `Process` objects")
+  @RequestMapping(value = "/{account}/{location}/{id:.+}", method = RequestMethod.GET)
+  Job getJob(@ApiParam(value = "Account job was created by", required = true) @PathVariable String account,
+             @ApiParam(value = "Namespace, region, or zone job is running in", required = true) @PathVariable String location,
+             @ApiParam(value = "Unique identifier of job being looked up", required = true) @PathVariable String id) {
+    Collection<Job> jobMatches = jobProviders.findResults {
```
duftler: Wouldn't `findResult` work here? Then you wouldn't need the `.first()` bit.

duftler:
This is looking really nice. LGTM. Defer to Tomas on merging.

lwander:
@tomaslin could you take one last look at this?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/JobProvider.groovy
lwander: It's missing from Instance and Server group, and will always be known by whoever is doing the lookup. So I left it out here.

tomaslin:
LGTM please squash before merge

PR #517, opened 2016-04-12
cats: New scheduler

lwander:
@cfieber @duftler

For an explanation of how this scheduler works, see the top of the ClusteredSortAgentScheduler.java

### DIFF
--- a/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredAgentScheduler.java
+++ b/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredAgentScheduler.java
@@ -20,6 +20,7 @@ import com.netflix.spinnaker.cats.agent.*;
 import com.netflix.spinnaker.cats.module.CatsModuleAware;
 import com.netflix.spinnaker.cats.redis.JedisSource;
 import com.netflix.spinnaker.cats.thread.NamedThreadFactory;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import redis.clients.jedis.Jedis;
 
 import java.util.Arrays;
@@ -27,12 +28,9 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 
+@SuppressFBWarnings
 public class ClusteredAgentScheduler extends CatsModuleAware implements AgentScheduler, Runnable {
     private final JedisSource jedisSource;
     private final NodeIdentity nodeIdentity;
--- a/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
+++ b/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
@@ -1,0 +1,297 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.cats.redis.cluster;
+
+import com.netflix.spinnaker.cats.agent.*;
+import com.netflix.spinnaker.cats.module.CatsModuleAware;
+import com.netflix.spinnaker.cats.redis.JedisSource;
+import com.netflix.spinnaker.cats.thread.NamedThreadFactory;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import redis.clients.jedis.Jedis;
+
+import java.util.*;
+import java.util.concurrent.*;
+
+/*
+ * The idea behind this scheduler is simple. Every agent it owns is always in one of two sorted sets,
+ * WORKING, or WAITING. If it is in the WORKING set, it's being run, if it's in the WAITING set, it is waiting to run.
+ * These sets are sorted by time, and the rank is time at which they are ready to run (in the WAITING set), or have
+ * timed out (in the WORKING set). Therefore, if we do a `zrangebyscore SETNAME -inf CURRENT_TIME`, we get a list
+ * of all agents that have expired. Those are then atomically pulled from the set, and run by the scheduler.
+ *
+ * The amortized cost of this scheduler is much lower than the original ClusteredAgentScheduler, since during each
+ * cache interval every key will only be removed from Redis once. If the interval is 60s, and the agent polls every 1s,
+ * we already have a (30s / 1) * (# of clouddrivers) factor of improvement.
+ */
+@SuppressFBWarnings
+public class ClusteredSortAgentScheduler extends CatsModuleAware implements AgentScheduler, Runnable {
+  private final JedisSource jedisSource;
+  private final NodeStatusProvider nodeStatusProvider;
+  private final AgentIntervalProvider intervalProvider;
+  private final ExecutorService agentWorkPool;
+
+  private static final Integer NOW = 0;
+  private static final int REDIS_REFRESH_PERIOD = 30;
+  private int runCount = 0;
+
+  private final Logger log;
+
+  private Map<String, AgentWorker> agents;
+  private Optional<Semaphore> runningAgents;
+
+  // This code assumes that every agent being run is in exactly either the WAITING or WORKING set.
+  private static final String WAITING_SET = "WAITZ";
+  private static final String WORKING_SET = "WORKZ";
+  private static final String ADD_AGENT_SCRIPT = "addAgentScript";
+  private static final String SWAP_SET_SCRIPT = "swapSetScript";
+  private static final String CONDITIONAL_SWAP_SET_SCRIPT = "conditionalSwapSetScript";
+
+  private ConcurrentHashMap<String, String> scriptShas;
+
+  public ClusteredSortAgentScheduler(JedisSource jedisSource, NodeStatusProvider nodeStatusProvider, AgentIntervalProvider intervalProvider, Integer parallelism) {
+    this.jedisSource = jedisSource;
+    this.nodeStatusProvider = nodeStatusProvider;
+    this.agents = new ConcurrentHashMap<>();
+    this.intervalProvider = intervalProvider;
+    this.log = LoggerFactory.getLogger(getClass());
+
+    if (parallelism == 0 || parallelism < -1) {
+      throw new IllegalArgumentException("Argument 'parallelism' must be positive, or -1 (for unlimited parallelism).");
+    } else if (parallelism > 0) {
+      this.runningAgents = Optional.of(new Semaphore(parallelism));
+    } else {
+      this.runningAgents = Optional.empty();
+    }
+
+    scriptShas = new ConcurrentHashMap<>();
+    storeScripts();
+
+    this.agentWorkPool = Executors.newCachedThreadPool(new NamedThreadFactory(AgentWorker.class.getSimpleName()));
+    Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory(ClusteredSortAgentScheduler.class.getSimpleName()))
+      .scheduleAtFixedRate(this, 0, 1, TimeUnit.SECONDS);
+  }
+
+  private void storeScripts() {
+    try (Jedis jedis = jedisSource.getJedis()) {
+      // When we switch an agent from one set to another, we first make sure it exists in the set we are removing it
+      // from, and then we perform the swap. If this check fails, the thread performing the swap does not get ownership
+      // of the agent.
+      // Swap happens from KEYS[1] -> KEYS[2] with the agent type being ARGV[1], and the score being ARGV[2].
+      scriptShas.put(SWAP_SET_SCRIPT, jedis.scriptLoad(
+          "if redis.call('zrank', KEYS[1], ARGV[1]) ~= nil then\n" +
+          "  redis.call('zrem', KEYS[1], ARGV[1])\n" +
+          "  return redis.call('zadd', KEYS[2], ARGV[2], ARGV[1])\n" +
+          "else return nil end\n"
+      ));
+
+      scriptShas.put(CONDITIONAL_SWAP_SET_SCRIPT, jedis.scriptLoad(
+          "if redis.call('zrank', KEYS[1], ARGV[1]) == ARGV[3] then\n" +
+          "  redis.call('zrem', KEYS[1], ARGV[1])\n" +
+          "  return redis.call('zadd', KEYS[2], ARGV[2], ARGV[1])\n" +
+          "else return nil end\n"
+      ));
+
+      // If the agent isn't present in either the WAITING or WORKING sets, it's safe to add. If it's present in either,
+      // it's being worked on or was recently run, so leave it be.
+      // KEYS[1] and KEYS[2] are checked for inclusion. If the agent is in neither ARGV[1] is added to KEYS[1] with score
+      // ARGV[2].
+      scriptShas.put(ADD_AGENT_SCRIPT, jedis.scriptLoad(
+        "if redis.call('zrank', KEYS[1], ARGV[1]) ~= nil then\n" +
+        "  if redis.call('zrank', KEYS[2], ARGV[1]) ~= nil then\n" +
+        "    return redis.call('zadd', KEYS[1], ARGV[2], ARGV[1])\n" +
+        "  else return nil end\n" +
+        "else return nil end\n"));
+    }
+  }
+
+  private String getScriptSha(String scriptName, Jedis jedis) {
+    String scriptSha = scriptShas.get(scriptName);
+    if (scriptSha == null) {
+      storeScripts();
+      scriptSha = scriptShas.get(scriptName);
+      if (scriptSha == null) {
+        throw new RuntimeException("Failed to load caching scripts.");
+      }
+    }
+
+    if (!jedis.scriptExists(scriptSha)) {
+      storeScripts();
+    }
+
+    return scriptShas.get(scriptName);
+  }
+
+  @Override
+  public void schedule(Agent agent, AgentExecution agentExecution, ExecutionInstrumentation executionInstrumentation) {
+    if (agent instanceof AgentSchedulerAware) {
+      ((AgentSchedulerAware)agent).setAgentScheduler(this);
+    }
+
+    if (!(agentExecution instanceof CachingAgent.CacheExecution)) {
+      throw new IllegalArgumentException("Sort scheduler requires agent executions to be of type CacheExecution");
+    }
+
+    agents.put(agent.getAgentType(), new AgentWorker(agent, (CachingAgent.CacheExecution)agentExecution, executionInstrumentation, this));
+    try (Jedis jedis = jedisSource.getJedis()) {
+      jedis.evalsha(getScriptSha(ADD_AGENT_SCRIPT, jedis), 2, WAITING_SET, WORKING_SET, agent.getAgentType(), score(NOW));
+    }
+  }
+
+  @Override
+  public void run() {
+    if (!nodeStatusProvider.isNodeEnabled()) {
+      return;
+    }
+    try {
+      saturatePool();
+    } catch (Throwable t) {
+      log.error("Failed to run caching agents", t);
+    } finally {
+      runCount++;
+    }
+  }
+
+  private static String score(long offset) {
+    return String.format("%d", TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()) + offset);
+  }
+
+  private String acquireAgent(Agent agent) {
+    try (Jedis jedis = jedisSource.getJedis()) {
+      String score = score(intervalProvider.getInterval(agent).getTimeout());
+      if (jedis.evalsha(getScriptSha(SWAP_SET_SCRIPT, jedis),
+          Arrays.asList(WAITING_SET, WORKING_SET),
+          Arrays.asList(agent.getAgentType(), score)) != null) {
+        return score;
+      } else {
+        return null;
+      }
+    }
+  }
+
+  private boolean conditionalReleaseAgent(Agent agent, String acquireScore) {
+    try (Jedis jedis = jedisSource.getJedis()) {
+      if (jedis.evalsha(getScriptSha(CONDITIONAL_SWAP_SET_SCRIPT, jedis),
+          Arrays.asList(WORKING_SET, WAITING_SET),
+          Arrays.asList(agent.getAgentType(), score(intervalProvider.getInterval(agent).getInterval()),
+              acquireScore)) != null) {
+        return true;
+      } else {
+        return false;
+      }
+    }
+  }
+
+  private boolean releaseAgent(Agent agent) {
+    try (Jedis jedis = jedisSource.getJedis()) {
+      if (jedis.evalsha(getScriptSha(SWAP_SET_SCRIPT, jedis),
+          Arrays.asList(WORKING_SET, WAITING_SET),
+          Arrays.asList(agent.getAgentType(), score(intervalProvider.getInterval(agent).getInterval()))) != null) {
+        return true;
+      } else {
+        return false;
+      }
+    }
+  }
+
+  private void saturatePool() {
+    try (Jedis jedis = jedisSource.getJedis()) {
+      // Occasionally repopulate the agents in case redis went down. If they already exist, this is a NOOP
+      if (runCount % REDIS_REFRESH_PERIOD == 0) {
+        for (String agent : agents.keySet()) {
+          jedis.evalsha(getScriptSha(ADD_AGENT_SCRIPT, jedis), 2, WAITING_SET, WORKING_SET, agent, score(NOW));
+        }
+      }
+
+      // First cull threads in the WORKING set that have been there too long (TIMEOUT time).
+      Set<String> oldKeys = jedis.zrangeByScore(WORKING_SET, "-inf", score(NOW));
+      for (String key : oldKeys) {
+        // Ignore result, since if this agent was released between now and the above jedis call, our work was done
+        // for us.
+        AgentWorker worker = agents.get(key);
+        if (worker != null) {
+          releaseAgent(worker.agent);
+        }
+      }
+
+      // Now look for agents that have been in the queue for at least INTERVAL time.
+      List<String> keys = new ArrayList<>();
+      keys.addAll(jedis.zrangeByScore(WAITING_SET, "-inf", score(NOW)));
+      Set<AgentWorker> workers = new HashSet<>();
+
+      // Loop until we either run out of threads to use, or agents (which are keys) to run.
+      while (!keys.isEmpty() && runningAgents.map(Semaphore::tryAcquire).orElse(true)) {
+        String agent = keys.remove(0);
+
+        AgentWorker worker = agents.get(agent);
+        String score;
+        if (worker != null && (score = acquireAgent(worker.agent)) != null) {
+          // This score is used to determine if the worker thread running the agent is allowed to store its results.
+          // If on release of this agent, the scores don't match, this agent was rescheduled by a separate thread.
+          worker.setScore(score);
+          workers.add(worker);
+        }
+      }
+
+      for (AgentWorker worker : workers) {
+        agentWorkPool.submit(worker);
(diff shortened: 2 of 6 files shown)


cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,204 @@
...
+ * we already have a 60 x 1 x # of clouddrivers factor of improvement.
+ */
+class ClusteredSortAgentScheduler extends CatsModuleAware implements AgentScheduler, Runnable {
+  private final JedisSource jedisSource;
+  private final NodeStatusProvider nodeStatusProvider;
+  private final ExecutorService agentWorkPool;
+  private final Integer parallelism;
+
+  private static Integer NOW = 0;
+  private static Integer INTERVAL = 30;
```
cfieber: there are a couple of agents (I think) that override the default interval. There is an AgentIntervalProvider interface that returns the interval given an Agent

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,204 @@
...
+  private static Integer TIMEOUT = 500;
+
+  private Map<String, AgentWorker> agents;
+  private volatile Integer runningAgents;
```
cfieber: AtomicInteger here?

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,204 @@
...
+      Set<AgentWorker> workers = new HashSet<>();
+
+      // Loop until we either run out of threads to use, or agents (which are keys) to run.
+      while ((parallelism == -1 || runningAgents != parallelism) && !keys.isEmpty()) {
+        String agent = keys.remove(0);
+
+        AgentWorker worker = agents.getOrDefault(agent, null);
+        if (worker != null && acquireAgent(agent)) {
+          workers.add(worker);
+          runningAgents++;
```
cfieber: I don't see the corresponding runningAgents-- anywhere

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,204 @@
...
+      try {
+        executionInstrumentation.executionStarted(agent);
+        long startTime = System.nanoTime();
+        agentExecution.executeAgent(agent);
+        executionInstrumentation.executionCompleted(agent, TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startTime));
+      } catch (Throwable cause) {
+        executionInstrumentation.executionFailed(agent, cause);
+      } finally {
+        scheduler.releaseAgent(agent.getAgentType());
+        scheduler.runningAgents--;
```
cfieber: nevermind! but maybe it would make sense to put the decrement into the releaseAgent method

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
cfieber: actually seems like java.util.concurrent.Semaphore would work well for this usage

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,204 @@
...
+  private static String score(long offset) {
+    // Precision will be lost in this cast once 2^53 seconds have elapsed since epoch (~ 285 Million years from 1970).
+    // I'm optimistic about Spinnaker's success, but not optimistic enough to see this is as a problem.
+    return String.format("%f", (double)(TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()) + offset));
+  }
+
+  // If the agent isn't present in either the WAITING or WORKING sets, it's safe to add. If it's present in either,
+  // it's being worked on or was recently run, so leave it be.
+  private static String atomicAddScript() {
+    return System.out.format("if redis.call('zrangebyscore', KEYS[1], ARGS[1]) != nil then\n" +
```
cfieber: for these scripts you can register them once and get back a sha, then use evalsha to call them (in that case you'd need to get rid of the string.format and just pass `score(NOW)` as an additional parameter

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,204 @@
...
+        "  redis.call('zrem', KEYS[1], ARGS[1])\n" +
+        "  return redis.call('zadd', KEYS[2], ARGS[1], '%s')\n" +
+        "else return nil end\n", score).toString();
+  }
+
+  private boolean acquireAgent(String type) {
+    try (Jedis jedis = jedisSource.getJedis()) {
+      return jedis.eval(atomicSwapSetsScript(score(TIMEOUT)),
+          Arrays.asList(new String[]{WAITING_SET, WORKING_SET}),
```
cfieber: Arrays.asList(WAITING_SET, WORKING_SET) should work (its got a varargs form)

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
lwander: Ah, good idea.

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
lwander: Semaphore will block, no? That would break the behavior I have below in `saturatePool`.

And why `AtomicInteger` over `volatile`? I know very little Java

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
lwander: Good idea, I can increment in the `acquireAgent` method then too

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
lwander: Oh, that's much neater, thanks!

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
lwander: Is that the idiomatic way to do this?

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
cfieber: semaphore has a non blocking tryAcquire (it would mean you'd be getting the ticket to schedule something before actually knowing there is something to schedule, so in the case of no agent to run, it would also need to release)

volatile gives you a consistent view of the value, but without also locking you don't have a guarantee that the value is still the same when you go to update (the ++ operation is not an atomic operation, its a read, increment, write), AtomicInteger has an atomic increment (without a lock to boot)

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
cfieber: it is more efficient in that 1) less bits over the wire (only passing the hash instead of the full script), and 2) the script is compiled once into redis's LUA memory and just invoked repeatedly
doing it this way where the actual executed script changes each time might end up creating a bunch of garbage in redis of previous scripts that will never execute again. at the very least its probably worth making the script itself static and parameterizing where it is String.formatting

duftler: I used the `evalsha` approach here: https://github.com/spinnaker/rosco/blob/master/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy

lwander:
It seems to happily cache my Kubernetes, Docker and GCP infrastructure. For now I should probably only enable this for Kubernetes, and then we'll roll it out later if needed?

cfieber:
I would be ok with just a global feature flag that toggles between implementations for all agents

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,232 @@
...
+class ClusteredSortAgentScheduler extends CatsModuleAware implements AgentScheduler, Runnable {
+  private final JedisSource jedisSource;
+  private final NodeStatusProvider nodeStatusProvider;
+  private final AgentIntervalProvider intervalProvider;
+  private final ExecutorService agentWorkPool;
+  private final Integer parallelism;
+  private Integer runCount = 0;
+  private final Integer REDIS_REFRESH_PERIOD = 30; // How many iterations to wait between repopulating redis with potentially missing keys
+
+  private static Integer NOW = 0;
```
cfieber: Can make this final

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,232 @@
...
+
+    scriptShas = new HashMap<>();
+    storeScripts();
+
+    this.agentWorkPool = Executors.newCachedThreadPool(new NamedThreadFactory(AgentWorker.class.getSimpleName()));
+    Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory(ClusteredSortAgentScheduler.class.getSimpleName()))
+      .scheduleAtFixedRate(this, 0, 1, TimeUnit.SECONDS);
+  }
+
+  private void storeScripts() {
```
cfieber: Should add a call to this on some mod of runCount. In the case of a redis crash that would recover the scripts and resume caching

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,232 @@
...
+  private static final String WORKING_SET = "WORKZ";
+  private static final String ADD_AGENT_SCRIPT = "addAgentScript";
+  private static final String SWAP_SET_SCRIPT = "swapSetScript";
+
+  private HashMap<String, String> scriptShas;
+
+  public ClusteredSortAgentScheduler(JedisSource jedisSource, NodeStatusProvider nodeStatusProvider, AgentIntervalProvider intervalProvider, Integer parallelism) {
+    this.jedisSource = jedisSource;
+    this.nodeStatusProvider = nodeStatusProvider;
+    this.agents = new HashMap<>();
```
cfieber: This should be ConcurrentHashMap (accessed during run and during schedule calls)

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,232 @@
...
+  private final NodeStatusProvider nodeStatusProvider;
+  private final AgentIntervalProvider intervalProvider;
+  private final ExecutorService agentWorkPool;
+  private final Integer parallelism;
+  private Integer runCount = 0;
+  private final Integer REDIS_REFRESH_PERIOD = 30; // How many iterations to wait between repopulating redis with potentially missing keys
+
+  private static Integer NOW = 0;
+
+  private Map<String, AgentWorker> agents;
```
cfieber: final here too

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,232 @@
...
+ * cache interval every key will only be removed from Redis once. If the interval is 60s, and the agent polls every 1s,
+ * we already have a (30s / 1) * (# of clouddrivers) factor of improvement.
+ */
+class ClusteredSortAgentScheduler extends CatsModuleAware implements AgentScheduler, Runnable {
+  private final JedisSource jedisSource;
+  private final NodeStatusProvider nodeStatusProvider;
+  private final AgentIntervalProvider intervalProvider;
+  private final ExecutorService agentWorkPool;
+  private final Integer parallelism;
+  private Integer runCount = 0;
```
cfieber: This can be primitive int to avoid boxing

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,232 @@
...
+        for (String type : agents.keySet()) {
+          jedis.evalsha(scriptShas.get(ADD_AGENT_SCRIPT), 2, WAITING_SET, WORKING_SET, type, score(NOW));
+        }
+      }
+      // First cull threads in the WORKING set that have been there too long (TIMEOUT time).
+      Set<String> oldKeys = jedis.zrangeByScore(WORKING_SET, "-inf", score(NOW));
+      for (String key : oldKeys) {
+        // Ignore result, since if this agent was released between now and the above jedis call, our work was done
+        // for us.
+        AgentWorker worker = agents.getOrDefault(key, null);
```
cfieber: Just .get here will return null if there is no worker in the map

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,232 @@
...
+
+  @Override
+  public void run() {
+    if (!nodeStatusProvider.isNodeEnabled()) {
+      return;
+    }
+    try {
+      saturatePool();
+    } catch (Throwable t) {
+      t.printStackTrace(System.err);
```
cfieber: Let's get a proper slf4j logger in here for this

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
lwander: agents is modified though by `schedule` outside of the constructor

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
lwander: What's boxing? Google isn't very helpful in this case

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
cfieber: the map isn't reassigned though - final just means that there is only a single assignment to that field at construction time

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
cfieber: boxing is conversion from primitive int to Integer object and back which will happen on the increment operation

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
lwander: Ah, now I see how final works. thanks!

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,236 @@
...
+          Arrays.asList(agent.getAgentType(), score(intervalProvider.getInterval(agent).getTimeout()))) != null;
+    }
+  }
+
+  private boolean releaseAgent(Agent agent) {
+    try (Jedis jedis = jedisSource.getJedis()) {
+      if (jedis.evalsha(scriptShas.get(SWAP_SET_SCRIPT),
+          Arrays.asList(WORKING_SET, WAITING_SET),
+          Arrays.asList(agent.getAgentType(), score(intervalProvider.getInterval(agent).getTimeout()))) != null) {
+        runningAgents.release();
```
cfieber: NPE if parallelism = -1

maybe change runningAgents to Optional<Semaphore> and do `runningAgents.ifPresent(Semaphore::release)` here and below `runningAgents.map(Semaphore::tryAcquire).orElse(true)`

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,236 @@
...
+
+  private void saturatePool() {
+    try (Jedis jedis = jedisSource.getJedis()) {
+      // Periodically repopulate redis with everything this agent is aware of (in case redis was flushed or crashed).
+      if (runCount % REDIS_REFRESH_PERIOD == 0) {
+        for (String type : agents.keySet()) {
+          jedis.evalsha(scriptShas.get(ADD_AGENT_SCRIPT), 2, WAITING_SET, WORKING_SET, type, score(NOW));
+        }
+
+        storeScripts();
```
cfieber: maybe do this before the for loop - on the chance it was needed it would mean the script gets repopulated before the evalsha call in the loop

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredAgentScheduler.java
```diff
@@ -87,7 +83,8 @@ private void runAgents() {
         activeAgents.putAll(thisRun);
         for (final Map.Entry<String, Long> toRun : thisRun.entrySet()) {
             final AgentExecutionAction exec = agents.get(toRun.getKey());
-            agentExecutionPool.submit(new AgentJob(toRun.getValue(), exec, this));
+            Future s = agentExecutionPool.submit(new AgentJob(toRun.getValue(), exec, this));
+            assert !s.isCancelled();
```
lwander: FindBugs was complaining that I was ignoring the result of these futures. (not sure why it was failing here, since this was otherwise untouched). Is there a neater way to do this? @cfieber

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
lwander: Used a variation of @duftler's approach to only repopulate scripts when they can't be found.

cats/cats-core/src/main/java/com/netflix/spinnaker/cats/mem/InMemoryCache.java
```diff
@@ -332,7 +332,7 @@ private void handleDelim(String s) {
                             toInit();
                             break;
                         default:
-                            throw new IllegalStateException("Unhandled delimiter in init state: " + s);
+                            throw new IllegalStateException(String.format("Unhandled delimiter in init state: %s", s));
```
lwander: Another place where FindBugs started complaining.

cfieber: Odd - must be something specific to your jvm version or something (@ttomsu had disabled it at one point mentioning this line). It works for me and works on Travis-CI

Feel free to ignore and I will see if I can figure out the cause and a way to conditionally disable it instead of blanket disabling

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
cfieber: Nice

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/RedisCacheConfig.groovy
```diff
@@ -43,6 +44,11 @@ import java.util.concurrent.TimeUnit
 @Configuration
 @ConditionalOnProperty('redis.connection')
 class RedisCacheConfig {
+  @Value('${redis.scheduler}')
```
cfieber: I think this will complain if the property doesn't exist :- change it to redis.scheduler:default and it will use default if the property is not set

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/RedisCacheConfig.groovy
```diff
@@ -88,12 +94,18 @@ class RedisCacheConfig {
   @Bean
   @ConditionalOnProperty(value = 'caching.writeEnabled', matchIfMissing = true)
   AgentScheduler agentScheduler(JedisSource jedisSource, @Value('${redis.connection:redis://localhost:6379}') String redisConnection, AgentIntervalProvider agentIntervalProvider, NodeStatusProvider nodeStatusProvider) {
-    URI redisUri = URI.create(redisConnection)
-    String redisHost = redisUri.getHost()
-    int redisPort = redisUri.getPort()
-    if (redisPort == -1) {
-      redisPort = 6379
+    if (schedulerType == null || schedulerType.isEmpty() || schedulerType.equalsIgnoreCase('default')) {
```
cfieber: Can use groovy-truth on a string for null or empty ( if !schedulerType ... )

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,252 @@
...
+    try (Jedis jedis = jedisSource.getJedis()) {
+      return jedis.evalsha(getScriptSha(SWAP_SET_SCRIPT),
+          Arrays.asList(WAITING_SET, WORKING_SET),
+          Arrays.asList(agent.getAgentType(), score(intervalProvider.getInterval(agent).getTimeout()))) != null;
+    }
+  }
+
+  private boolean releaseAgent(Agent agent) {
+    try (Jedis jedis = jedisSource.getJedis()) {
+      if (jedis.evalsha(getScriptSha(SWAP_SET_SCRIPT),
```
cfieber: I think there is a potential issue in here if an agent runs longer than the timeout value:

clouddriver1 schedules agent1 who has a timeout of 5 minutes
Agent1 starts
5 minutes pass so agent1 times out
Clouddriver2 schedules agent1
Agent1 on clouddriver1 finishes and releases the agent entry that is actually the lock for clouddriver2 agent1
ClouddriverN schedules agent1 again

This is what the NodeIdentityProvider in the other scheduler tried to account for (the lock was agentId->nodeId)

Don't have an immediate suggestion to fix other than incorporating some aspect of that behaviour into this scheduler (which might play nice with someone externally requesting an on demand run too)

lwander: I think the best way to fix that would be to prevent Agent1 from recording its results if it fails to release the lock (can't move Agent1.type from WORKING->WAITING).

lwander: See new implementation

lwander:
@cfieber is this ready to go?

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,297 @@
...
+        "if redis.call('zrank', KEYS[1], ARGV[1]) ~= nil then\n" +
+        "  if redis.call('zrank', KEYS[2], ARGV[1]) ~= nil then\n" +
+        "    return redis.call('zadd', KEYS[1], ARGV[2], ARGV[1])\n" +
+        "  else return nil end\n" +
+        "else return nil end\n"));
+    }
+  }
+
+  private String getScriptSha(String scriptName) {
+    try (Jedis jedis = jedisSource.getJedis()) {
```
cfieber: move the try-with-resources bit to just contain the call to jedis (otherwise it has a connection from the pool, then gets another connection from the pool inside storeScripts) Alternatively have storeScripts accept the Jedis object as an arg?

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,297 @@
...
+          // This score is used to determine if the worker thread running the agent is allowed to store its results.
+          // If on release of this agent, the scores don't match, this agent was rescheduled by a separate thread.
+          worker.setScore(score);
+          workers.add(worker);
+        }
+      }
+
+      for (AgentWorker worker : workers) {
+        Future s = agentWorkPool.submit(worker);
+        assert !s.isCancelled(); // This is only to appease findBugs
```
cfieber: there is an `@SuppressFBWarnings` annotation you could use instead of the assert here

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,297 @@
...
+      } catch (Throwable cause) {
+        executionInstrumentation.executionFailed(agent, cause);
+      } finally {
+        // Regardless of success or failure, we need to try and release this agent. If the release is successful (we
+        // own this agent), and a result was created, we can store it.
+        if (scheduler.conditionalReleaseAgent(agent, acquireScore) && result != null) {
+          agentExecution.storeAgentResult(agent, result);
+        }
+
+        scheduler.runningAgents.ifPresent(Semaphore::release);
```
cfieber: should do this before the call to conditionalReleaseAgent (which could fail if there is an error talking to redis) or alternatively do this inside conditionalReleaseAgent and keep all access to the semaphore inside the scheduler

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
cfieber: new impl looks good, thanks!

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredAgentScheduler.java
cfieber: would rather see `@SuppressFBWarnings` than the random assert - and need to figure out why that warning shows up only for you folks (I suspect something in the JVM you guys use)

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -0,0 +1,297 @@
...
+    }
+  }
+
+  private String getScriptSha(String scriptName) {
+    try (Jedis jedis = jedisSource.getJedis()) {
+      String scriptSha = scriptShas.get(scriptName);
+      if (scriptSha == null) {
+        storeScripts();
+        scriptSha = scriptShas.get(scriptName);
+        assert scriptSha != null;
```
cfieber: woudl prefer to see an exception thrown here with a descriptive message (easier to filter logging on than an AssertionError)

cfieber:
my last batch of comments were pretty minor, LGTM after those are addressed, looking forward to trying it out

PR #521, opened 2016-04-13
Fixes server group key issue and on demand caching issue

ttomsu:
I goofed a few places where zone and region are used to generate the server group key in the Redis impl. This, understandably caused some issues - some nuanced and some pretty obvious.

@duftler discovered the region/zone key issue, but there was also a more nefarious issue lurking. TL;DR: don't have your On Demand handler say it's authoritative. When you do this, and your cache update only includes 1 entry, the infrastructure takes this as a sign to delete the rest of the keys in that namespace. This caused server group flapping in the UI and general instability in any stage that included `ServerGroupForceCacheRefresh`

@duftler @jtk54 PTAL


ttomsu:
@lwander

lwander:
LGTM

duftler:
LGTMA.

PR #531, opened 2016-04-14
provider/kubernetes: Include version in UA

lwander:


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/KubernetesConfiguration.groovy
```diff
@@ -47,6 +48,11 @@ class KubernetesConfiguration {
   }
 
   @Bean
+  String kubernetesApplicationName(@Value('${Implementation-Version:Unknown}') String implementationVersion) {
```
duftler: I think you'll need this (https://github.com/spinnaker/clouddriver/blob/master/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/GoogleConfiguration.groovy#L44) to get that property populated with the version from the gradle build.

PR #527, opened 2016-04-14
provider/cloudfoundry: Handle credentials for a clean install

gregturn:

PR #532, opened 2016-04-15
Add flag to build uber jar

duftler:
Deploying clouddriver to Cloud Foundry requires an uber jar. This adds a flag "springBoot.repackage" that can be used on the command line to build an uber jar.

Cherry-picks #465.


duftler:
Cherry-picks #527.

PR #533, opened 2016-04-15
Fixes startup warnings of the use of a deprecated API 

ttomsu:
(by removing the API)

@spinnaker/reviewers @lwander PTAL


lwander:
Looks good to me!

PR #536, opened 2016-04-15
register required resource providers in Azure.

scotmoor:
Add logic to check resource provider status and, if necessary, register the provider for the subscription


scotmoor:

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureBaseClient.groovy
```diff
@@ -122,4 +121,20 @@ public abstract class AzureBaseClient {
     responseStatusCode == HttpURLConnection.HTTP_NO_CONTENT || responseStatusCode == HttpURLConnection.HTTP_NOT_FOUND
   }
 
+  /***
+   * register the resource provider in the subscription
+   * @param resourceManagerClient - an instance of the AzureResourceManagerClient
+   */
+  void register(AzureResourceManagerClient resourceManagerClient) {
+    if (resourceManagerClient && !"".equals(providerNamespace)) {
```
ajordens: `if (resourcesManagerClient && providerNamespace) {` would probably read a little nicer and also handles the situation where providerNamespace is "" or null.

PR #539, opened 2016-04-19
cats & kubernetes: on-demand locking

lwander:
@cfieber @duftler 

OnDemand cache updates (for agents that implement AgentSchedulerAware) using the new Sort scheduler no longer need the OnDemand namespace. This make these force cache refreshes much faster: 

fast-scheduler

To make sure that the Kubernetes provider keeps working regardles of the scheduler, I left the OnDemand namespace code in for now.

### DIFF
--- a/cats/cats-core/src/main/java/com/netflix/spinnaker/cats/agent/AgentLock.java
+++ b/cats/cats-core/src/main/java/com/netflix/spinnaker/cats/agent/AgentLock.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.cats.agent;
+
+public class AgentLock {
+  private final Agent agent;
+
+  public AgentLock(Agent agent) {
+    this.agent = agent;
+  }
+
+  public Agent getAgent() {
+    return agent;
+  }
+}
--- a/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentLock.java
+++ b/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentLock.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.cats.redis.cluster;
+
+import com.netflix.spinnaker.cats.agent.Agent;
+import com.netflix.spinnaker.cats.agent.AgentLock;
+
+public class ClusteredSortAgentLock extends AgentLock {
+  // The score the agent was acquired with (Used to ensure we own this agent on release).
+  private final String acquireScore;
+  // The score the agent was release from the WAITING set with (Used to ensure it is readded to the WAITING set with the right score).
+  private final String releaseScore;
+
+  ClusteredSortAgentLock(Agent agent, String acquireScore, String releaseScore) {
+    super(agent);
+    this.acquireScore = acquireScore;
+    this.releaseScore = releaseScore;
+  }
+
+  public String getAcquireScore() {
+    return acquireScore;
+  }
+
+  public String getReleaseScore() {
+    return releaseScore;
+  }
+}
--- a/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
+++ b/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
@@ -40,7 +40,7 @@ import java.util.concurrent.*;
  * we already have a (30s / 1) * (# of clouddrivers) factor of improvement.
  */
 @SuppressFBWarnings
-public class ClusteredSortAgentScheduler extends CatsModuleAware implements AgentScheduler, Runnable {
+public class ClusteredSortAgentScheduler extends CatsModuleAware implements AgentScheduler<ClusteredSortAgentLock>, Runnable {
   private final JedisSource jedisSource;
   private final NodeStatusProvider nodeStatusProvider;
   private final AgentIntervalProvider intervalProvider;
@@ -59,6 +59,7 @@ public class ClusteredSortAgentScheduler extends CatsModuleAware implements Agen
   private static final String WAITING_SET = "WAITZ";
   private static final String WORKING_SET = "WORKZ";
   private static final String ADD_AGENT_SCRIPT = "addAgentScript";
+  private static final String VALID_SCORE_SCRIPT = "validScoreScript";
   private static final String SWAP_SET_SCRIPT = "swapSetScript";
   private static final String CONDITIONAL_SWAP_SET_SCRIPT = "conditionalSwapSetScript";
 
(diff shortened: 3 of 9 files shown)


cfieber:
If an OnDemand is successfully executed with locking, does that reset the execution of the overall agent run back to its interval?

If I have a ClusterCachingAgent that should run every 30 seconds to load all server groups, if I get an OnDemand every 30 seconds or less will the ClusterCachingAgent ever do its full run?

lwander:
Ah, good call, it will push back the execution of the next LoadData cycle (which we probably don't want). I'll fix that.

lwander:
@cfieber The lock is now released with the score it had before the lock was acquired (so the interval isn't messed up each time an ondemand update happens).

ajordens:
Needs a rebase

ajordens:
Can we also find a suitable class to document the semantics of the locking ... it'll be useful for future selfs that end up looking this code in 2 months.

lwander:
What do you mean by document the semantics in a class? I can annotate the `AgentScheduler` with more comments to make it clearer what the `tryLock` and `tryRelease` are doing.

ajordens:
Sorry, just looking for some docs. On the AgentScheduler is probably a good
spot.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy
```diff
@@ -57,13 +61,27 @@ class CatsOnDemandCacheUpdater implements OnDemandCacheUpdater {
...
         def providerCache = catsModule.getProviderRegistry().getProviderCache(agent.providerName)
         OnDemandAgent.OnDemandResult result = agent.handle(providerCache, data)
         if (result) {
+          if (atomicScheduler && !(scheduler.tryRelease(lock))) {
+            hasOnDemandResults = true // force Orca to retry
+          }
           if (result.cacheResult) {
-            hasOnDemandResults = !(result.cacheResult.cacheResults ?: [:]).values().flatten().isEmpty()
+            hasOnDemandResults = !(result.cacheResult.cacheResults ?: [:]).values().flatten().isEmpty() && !atomicScheduler
             agent.metricsSupport.cacheWrite {
```
cfieber: this releases the lock before write the results. if that triggers the normal caching run for the agent, it may not see the on demand node when it runs (in our case for AWS we will still have to inspect that during our normal caching cycle to make sure our polled data is at least as new as our onDemand data)

I think I'd err on the side of doing release after write 

also probably want to stick the tryRelease call in a finally block to make sure we release if some exception happens during handle

cfieber: I guess releasing after write loses the check as to whether the lock is still valid before writing... hmm I still think I'm leaning towards release after write though

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy
```diff
@@ -57,13 +61,27 @@ class CatsOnDemandCacheUpdater implements OnDemandCacheUpdater {
   OnDemandCacheUpdater.OnDemandCacheStatus handle(OnDemandAgent.OnDemandType type, Collection<OnDemandAgent> onDemandAgents, Map<String, ? extends Object> data) {
     boolean hasOnDemandResults = false
     for (OnDemandAgent agent : onDemandAgents) {
+      boolean atomicScheduler = false;
+      AgentScheduler scheduler = null
+      if (agent instanceof AgentSchedulerAware) {
+        scheduler = agent.agentScheduler
```
cfieber: can we actually have multiple schedulers in the system? ( I think not the way it is currently set up )

if the CatsOnDemandCacheUpdater had a reference to the scheduler at the class level, the agents wouldn't have to implement AgentSchedulerAware and the return of an empty list for pendingOnDemand (currently in the k8s agent) could be done in this class

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -154,6 +154,28 @@ public void schedule(Agent agent, AgentExecution agentExecution, ExecutionInstru
   }
 
   @Override
+  public ClusteredSortAgentLock tryLock(Agent agent) {
+    // This score determines how much time the agent had left in the set it's currently in.
+    String releaseScore = agentScore(agent);
```
cfieber: we would only care about score in the waiting set in this case (acquire would fail if the agent is in working)

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy
lwander: Yeah, I thought about this, and figured that since the since the `.handle()` call is on the order of seconds, while the time between the release and the write is on the order of a few instructions, this shouldn't be a problem (most of the time). If we release after write we miss out on fixing the race condition where `handle()` times out and writes old data. I thank for that reason, this current approach is safer. If we hit the current race condition, we are writing data captured from the same source within less than a second of each other. If we release after the write, we risk writing data that was captured after a timeout occurred, meaning minute-old data could overwrite fresh data.

Ideally we'd need an extra lock just around the write for this to be really atomic, perhaps on a per-resource basis even, but that's a lot of transactions. 

WDYT?

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -154,6 +154,28 @@ public void schedule(Agent agent, AgentExecution agentExecution, ExecutionInstru
...
+    String acquireScore = acquireAgent(agent);
```
cfieber: it would be worth making the lua script for acquireAgent return both releaseScore and acquireScore which would make determining these atomic. I think all that would have to change in the lua script is storing the value of the original zrank call and returning it at the end

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
lwander: Yes that's true - I wrote that comment because `releaseScore` can refer to either set. However, since it's only useful when `releaseScore` came from the WAITING set I'll fix the comment to make it clearer what the intention is.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy
cfieber: worth adding an isValid on AgentLock that tests whether the lock is still held without releasing it? could call that before the write and release after the write in that case (and doesn't seem too much more complex)

cats/cats-core/src/main/java/com/netflix/spinnaker/cats/agent/AgentLock.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.cats.agent;
+
+public class AgentLock {
+  Agent agent;
```
cfieber: make this private final

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentLock.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package com.netflix.spinnaker.cats.redis.cluster;
+
+import com.netflix.spinnaker.cats.agent.Agent;
+import com.netflix.spinnaker.cats.agent.AgentLock;
+
+public class ClusteredSortAgentLock extends AgentLock {
+  // The score the agent was acquired with (Used to ensure we own this agent on release).
+  String acquireScore;
```
cfieber: private final on these too

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy
lwander: It's possible that 
```
assert(lock.isValid())
assert(!lock.isValid())
```
passes, but I guess that's even less likely than either of the above scenarios. I'll go with the `isValid()`.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy
lwander: Yeah there is only one scheduler at a time, but I figured that's an implicit assumption and wanted to avoid adding code that will break in weird ways if two are running at a time. WDYT?

cfieber: I'd be willing to go with the constraint that we have a single scheduler just to ensure the proper semantics when the scheduler is atomic are captured in this one class

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
lwander: Oh, can lua return multiple values?

cfieber: not sure, but the current return value from the swap script isn't being (other than a not null check to ensure the call worked) so it could be changed to return the zrank score as the indicator of success

lwander:
Fixed comments. I left it as two commit for now to see what was changed, I'll squash before we merge.

cfieber:
Nice work - LGTM!

PR #541, opened 2016-04-20
Mutating AWS server group calls operate only on `asgs`.

duftler:
All manner of asgName|serverGroupName, region|regions and asgs are supported on the requests and coerced to asgs via LegacyToAsgsDescriptionPreProcessor.
AWS resize is handled specially via its own dedicated ResizeAsgDescriptionPreProcessor due to the presence of `capacity` within `asgs`.
asgName and regions are deprecated on all the descriptions.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/DeleteAsgTagsDescription.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/DeleteAsgTagsDescription.groovy
@@ -16,8 +16,12 @@
 package com.netflix.spinnaker.clouddriver.aws.deploy.description
 
 class DeleteAsgTagsDescription extends AbstractAmazonCredentialsDescription {
-  List<AsgDescription> asgs = []
+
+  String serverGroupName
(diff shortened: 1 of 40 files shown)


duftler:
As discussed, these changes are necessary so that orca can be modified to pass `region` on these calls instead of `regions`.
Will first need to massage the Azure operations to accept `region` instead of `regions` (starting on that now).

duftler:
This PR does not contain any breaking changes w.r.t. to callers.

milismsft:
I thought that having multiple regions rather then one region allows me when creating pipeline stages to select/specify multiple ones and execute certain operations on them simultaneously (such as destroy, enable, disable). Was that not the case?

duftler:
Yes, that still works (and will continue to work) fine. The reason that works is that orca synthesizes individual clouddriver calls (one per region) from the original multi-region request.

duftler:
As it is now, orca always passes single-item `regions` lists to clouddriver for those stages.

duftler:
Rebased.

ajordens:
LGTM

I'll do a bit more testing on the affected Descriptions before we push it out to `prod`.

duftler:
Cool. Good to merge?

PR #542, opened 2016-04-20
Mutating Azure server group operations now handle either `regions` or…

duftler:
… `region`.
`regions` is deprecated in favor of `region`.
@milismsft please review.

### DIFF
--- a/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/model/EnableDisableDestroyAzureServerGroupDescription.groovy
+++ b/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/model/EnableDisableDestroyAzureServerGroupDescription.groovy
@@ -18,5 +18,8 @@ package com.netflix.spinnaker.clouddriver.azure.resources.servergroup.model
 
 class EnableDisableDestroyAzureServerGroupDescription extends AzureServerGroupDescription{
   String serverGroupName
+  String region
+
+  @Deprecated
   List<String> regions
 }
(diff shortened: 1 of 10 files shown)


duftler:
Make sure to view the diff for this pr with `?w=1` like so: https://github.com/spinnaker/clouddriver/pull/542/files?w=1

Otherwise, the unimportant whitespace delta (from removing the surrounding `regions` loops in the 3 operations) will drown out the important details.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/model/EnableDisableDestroyAzureServerGroupDescription.groovy
```diff
@@ -18,5 +18,8 @@ package com.netflix.spinnaker.clouddriver.azure.resources.servergroup.model
 
 class EnableDisableDestroyAzureServerGroupDescription extends AzureServerGroupDescription{
   String serverGroupName
+  String region
```
milismsft: region is already defined in the base class AzureResourceOpsDescription (all our descriptions are expected to extend that class)

duftler: Ah, missed that. I've fixed it and re-pushed the branch.

milismsft:
How does RegionsToRegionDescriptionPreProcessor come in play? Don't we need to call it in the converter classes (i.e. DestroyAzureServerGroupAtomicOperationConverter)?

Looking at the code we also use regions for some of the operations on the security group and load balancer. Do we need to change those as well?

duftler:
You can see in this PR (https://github.com/spinnaker/clouddriver/pull/235) how this code is called in `OperationsController` just before the validators are called.

> Looking at the code we also use regions for some of the operations on the security group and load balancer. Do we need to change those as well?

Nope. These changes specifically relate to coming changes in orca that are contained to the mutating server group operations.

milismsft:
LGTM then :-)

PR #544, opened 2016-04-20
Drop `region` from EnableDisableDestroyAzureServerGroupDescription si…

duftler:
…nce it is already defined on the superclass.


duftler:
Thought I had added this to the prior PR, but missed it somehow.
@milismsft fyi

PR #550, opened 2016-04-22
provider/google: All mutating server group operations now take region…

duftler:
… (zone is deprecated).


duftler:

ttomsu:
LGTM

PR #553, opened 2016-04-23
core: Prefix job controller path with containing app name

lwander:
Need by Deck to display jobs by application.

@duftler


lwander:
I added a second commit switching processes to instances. Since they are effectively the same thing (with even the same health states), it was causing too much grief in deck to have instances and processes have separate names. I'll squash before we merge.

duftler:
LGTM.

PR #555, opened 2016-04-25
cats: Use redis rather than system for timestamp

lwander:
Suggested change by @jtk54 & @duftler 

@cfieber


cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -220,8 +220,13 @@ public void run() {
     }
   }
 
-  private static String score(long offset) {
-    return String.format("%d", TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()) + offset);
+  private static String score(Jedis jedis, long offset) {
+    List<String> times = jedis.time();
```
lwander: I grab the time here rather than inside the lua script, since Redis does not allow non-deterministic calls to be made prior to a write inside a single transaction.

cfieber:
LGTM

PR #556, opened 2016-04-25
Implement Azure Application Gateway create template deployment

milismsft:
Implement Azure Application Gateway create template deployment (couple test cases included as well).

An Application Gateway resource provides an Azure-managed HTTP load-balancing solution based on layer-7 load balancing. These fit better in the Spinnaker load balancer model than the Azure Load Balancer resource that is currently used because it better suits transitioning from one active version of the server group to another. For more info about Azure application gateways and load balancer and how they compare see https://azure.microsoft.com/en-us/documentation/articles/application-gateway-introduction/

### DIFF
--- a/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/loadbalancer/view/AzureLoadBalancerProvider.groovy
+++ b/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/loadbalancer/view/AzureLoadBalancerProvider.groovy
@@ -78,6 +78,7 @@ class AzureLoadBalancerProvider implements LoadBalancerProvider<AzureLoadBalance
       name: loadBalancerDescription.loadBalancerName,
       region: loadBalancerDescription.region,
       vnet: loadBalancerDescription.vnet?: "vnet-unassigned",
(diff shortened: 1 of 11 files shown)


milismsft:

milismsft:
ping

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/loadbalancer/view/AzureLoadBalancerProvider.groovy
```diff
@@ -78,6 +78,7 @@ class AzureLoadBalancerProvider implements LoadBalancerProvider<AzureLoadBalance
       name: loadBalancerDescription.loadBalancerName,
       region: loadBalancerDescription.region,
       vnet: loadBalancerDescription.vnet?: "vnet-unassigned",
+      subnet: loadBalancerDescription.subnet?: "subnet-unassigned",
```
duftler: Might read easier with a space before the elvis operator. Can always change this in a subsequent PR.

duftler:
LGTM.
(All changes are contained to azure packages.) 
Will merge when travis completes.

PR #558, opened 2016-04-25
Consolidate version of Jackson

gregturn:
Running clouddriver in uber JAR mode generated quirky errors in Jackson that didn't happen in my IDE. Spotted that different modules were using different versions of Jackson, k8 being on 2.7 and all the others on 2.5. This patch tweaks the build system to use one version of Jackson.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,6 +71,12 @@ subprojects { project ->
           if (it.requested.group == 'org.apache.xbean') {
             it.useVersion '4.3'
           }
+          if (it.requested.group == 'com.fasterxml.jackson.core'
+                  || it.requested.group == 'com.fasterxml.jackson.datatype'
+                  || it.requested.group == 'com.fasterxml.jackson.dataformat'
+                  || it.requested.group == 'com.fasterxml.jackson.module') {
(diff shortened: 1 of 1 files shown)


gregturn:
@lwander PTAL

build.gradle
```diff
@@ -71,6 +71,12 @@ subprojects { project ->
           if (it.requested.group == 'org.apache.xbean') {
             it.useVersion '4.3'
           }
+          if (it.requested.group == 'com.fasterxml.jackson.core'
+                  || it.requested.group == 'com.fasterxml.jackson.datatype'
+                  || it.requested.group == 'com.fasterxml.jackson.dataformat'
+                  || it.requested.group == 'com.fasterxml.jackson.module') {
+            it.useVersion '2.5.4'
```
cfieber: I'm almost certain that this is not the only module that is going to be all over the place in versions in the graph. If the uberjar builder can't successfully deal with that then it is going to need an alternative solution than just explicitly version managing dependencies because that will get way out of hand

cfieber:
I don't think this is the right approach to solving this issue

PR #565, opened 2016-04-27
provider/kubernetes: Allow empty apps to be deleted

lwander:
When `CLUSTER` wasn't authoritative, it would persist cached relationships (for deleted server groups/load balancers), causing the application controller to return old clusters. Since this is only surfaced when summary application details are loaded, it went unnoticed until those details were used by orca to determine if an app can be deleted. This fixes that.

@duftler


duftler:
LGTM.

PR #566, opened 2016-04-28
provider/google: Return GCE regions as a list of maps to be more cons…

duftler:
…istent with how AWS regions are returned.


duftler:

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.java
```diff
@@ -75,8 +75,16 @@ public String getCloudProvider() {
       return GoogleCloudProvider.GCE;
     }
 
-    public Map<String, List<String>> getRegions() {
-      return regionToZonesMap;
+    public List<Map> getRegions() {
```
lwander: Seems like an odd name since this returns ` [ (region, [zones]) ]`

lwander:
LGTM

PR #569, opened 2016-04-29
provider/google: GoogleServerGroupCachingAgent will now handle on dem…

duftler:
…and cache requests that omit zone.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.java
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.java
@@ -99,6 +99,10 @@ public class GoogleNamedAccountCredentials implements AccountCredentials<GoogleC
       return null;
     }
 
+    public List<String> getZonesFromRegion(String region) {
+      return region != null && regionToZonesMap != null ? regionToZonesMap.get(region) : null;
+    }
+
     public GoogleCredentials getCredentials() {
       return credentials;
     }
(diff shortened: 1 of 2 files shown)


duftler:
@ttomsu or @lwander please review.

lwander:
LGTM, but maybe @ttomsu should take a look too since I could have missed something specific to google's resource model.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.java
```diff
@@ -99,6 +99,14 @@ public String regionFromZone(String zone) {
       return null;
     }
 
+    public List<String> getZonesFromRegion(String region) {
+      if (region == null || regionToZonesMap == null) {
```
ttomsu:
```
return regionToZonesMap != null ? regionToZonesMap.get(region) : null
```

ttomsu: Or rename this to a groovy class (since it's in the `groovy/` dir) and do
```
return regionToZoneMap?.get(region)
```

ttomsu:
LGTM

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.java
duftler: Done.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.java
```diff
@@ -99,6 +99,10 @@ public String regionFromZone(String zone) {
       return null;
     }
 
+    public List<String> getZonesFromRegion(String region) {
+      return region != null && regionToZonesMap != null ? regionToZonesMap.get(region) : null;
```
ttomsu: You don't need the `region != null`, because there is no `null` key in the map anyway.

duftler: It wasn't clear to me exactly what kind of Map regionToZonesMap will be. As
the javadocs for Map.get() indicate a Map impl may NPE on a null key, I
chose to play it safe.

PR #573, opened 2016-04-30
provider/google: Squash 404 when autoscaler is not found during on de…

duftler:
…mand cache update.


duftler:

PR #578, opened 2016-05-04
provider/kubernetes: Cache clusters for both jobs & servergroups

lwander:
Since both jobs & server groups belong to (mutually exclusive) sets of clusters, the cache needs to reflect this. By making the cluster cache key in redis aware of whether the cluster owns server groups or jobs, both the server group and job caching agents can be authoritative for clusters in the same namespace.

@ttomsu PTAL.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/cache/Keys.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/cache/Keys.groovy
@@ -67,10 +67,11 @@ class Keys {
       case Namespace.CLUSTERS.ns:
         def names = Names.parseName(parts[4])
         result << [
-            application: parts[3],
             account: parts[2],
-            name: parts[4],
-            cluster: parts[4],
+            application: parts[3],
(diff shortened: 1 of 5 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/cache/Keys.groovy
```diff
@@ -69,8 +69,9 @@ class Keys {
         result << [
             application: parts[3],
             account: parts[2],
-            name: parts[4],
-            cluster: parts[4],
+            name: parts[5],
+            cluster: parts[5],
+            category: parts[4],
```
ttomsu: `category` is pretty generic. I wouldn't know what the heck it meant upon first glance. `clusterType`? `jobType`? `podOwner`? `clusterF$%`?

lwander: It mirrors "category" in deck, where it's becoming a more pervasive way to distinguish between server groups and jobs. I'll leave a comment here.

.. `clusterF$%`, really?

lwander:
@ttomsu LGTU?

ttomsu:
LGTM

PR #583, opened 2016-05-06
Update Scale Set template to create a load balancer and associate AG 

scotmoor:
Updates the ARM template to set the backend address pool of the scale set accordingly
 - Create backend address pool in Application Gateway for server group if it doesn't exist
 - Move creation of internal load balancer associated with scale set into ARM template.
 - Minor fixes

### DIFF
--- a/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/appgateway/model/AzureAppGatewayDescription.groovy
+++ b/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/appgateway/model/AzureAppGatewayDescription.groovy
@@ -75,7 +75,7 @@ class AzureAppGatewayDescription extends AzureResourceOpsDescription {
     description.serverGroups = appGateway.tags?.serverGroups?.split(" ")
     description.vnet = appGateway.tags?.vnet
     description.createdTime = appGateway.tags?.createdTime?.toLong()
-    description.tags = appGateway.tags
+    description.tags = appGateway.tags ?: [:]
     description.region = appGateway.location
 
     appGateway.requestRoutingRules.each { rule ->
--- a/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/ops/CreateAzureServerGroupAtomicOperation.groovy
+++ b/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/ops/CreateAzureServerGroupAtomicOperation.groovy
@@ -56,8 +56,8 @@ class CreateAzureServerGroupAtomicOperation implements AtomicOperation<Map> {
     String resourceGroupName = null
     String virtualNetworkName = null
     String subnetName = null
-    String loadBalancerName = null
     String serverGroupName = null
+    //String appGatewayPoolID = null
 
     try {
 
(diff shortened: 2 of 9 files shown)


scotmoor:
@JargoonPard @milismsft @spinnaker/netflix-reviewers PTAL

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/ops/CreateAzureServerGroupAtomicOperation.groovy
```diff
@@ -96,43 +96,46 @@ class CreateAzureServerGroupAtomicOperation implements AtomicOperation<Map> {
...
-        AzureLoadBalancerResourceTemplate.getTemplate(lbDescription),
-        resourceGroupName,
-        description.region,
-        description.loadBalancerName,
-        "loadBalancer")
-      errList = AzureDeploymentOperation.checkDeploymentOperationStatus(task, BASE_PHASE, description.credentials, resourceGroupName, deployment.name)
+      // get the app gateway. Verify that it can be used for this server group/cluster. add an app pool if it doesn't already exist
+      def appGateway = description.credentials.networkClient.getAppGateway(resourceGroupName, description.loadBalancerName)
+      if (appGateway) {
+        if (appGateway.tags?.cluster && appGateway.tags?.cluster != description.clusterName) {
```
milismsft: We could move this check into getAppGatewayBackendPool() implementation and remove one call to get the application gateway object. I will do that as part of the enable/disable/app gateway backend address pool management check-in.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/appgateway/model/AzureAppGatewayDescription.groovy
```diff
@@ -30,6 +30,7 @@ class AzureAppGatewayDescription extends AzureResourceOpsDescription {
   List<String> serverGroups
   List<AzureAppGatewayHealthcheckProbe> probes = []
   List<AzureAppGatewayRule> rules = []
+  List<AzureAppGatewayBackendAddressPool> backendAddressPools = []
```
milismsft: I'm not sure how we are going to use this list because it's not something we are going to bring it into the deck UI. Was this added for debugging purpose only? (in which case it makes total sense)

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/templates/AzureServerGroupResourceTemplate.groovy
```diff
@@ -386,15 +436,19 @@ class AzureServerGroupResourceTemplate {
    */
   static class NetworkInterfaceIPConfigurationsProperty {
     NetworkInterfaceIPConfigurationSubnet subnet
+    ArrayList<AppGatewayBackendAddressPool> ApplicationGatewayBackendAddressPools = []
     ArrayList<LoadBalancerBackendAddressPool> loadBalancerBackendAddressPools = []
+    ArrayList<LoadBalancerInboundNatPoolId> loadBalancerInboundNatPools = []
 
-    /**
+    /**you specify the backend address pool the "rul
```
milismsft: typo?

milismsft:
LGTM after addressing one comment I made offline to Scot.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/ops/CreateAzureServerGroupAtomicOperation.groovy
scotmoor: I intentionally separated this out from the getAppGatewayBackendPool method in the client as this is "business logic" outside the scope of the responsibilities of the clients and should remain that way. Unfortunately without some additional refactoring, this will require a second call to retrieve the Application Gateway, but the perf hit for this 1 additional call is negligible. If we see a huge perf degradation from this additional call then we should look into the refactoring.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/appgateway/model/AzureAppGatewayDescription.groovy
scotmoor: Part of a change that I removed. will remove this as well.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/ops/CreateAzureServerGroupAtomicOperation.groovy
milismsft: "_this is "business logic" outside the scope of the responsibilities of the clients and should remain that way._"
Though I might agree with the principal you're trying to set here, I don't think we could do that anyway; for instance appGateway.tags.cluster will be set via a "client" call anyway. But let's not worry about this for now, I just wanted to give you a heads up that this might get refactored later on.

scotmoor:

ajordens:
Only touches azure classes so LGTM ... will need a rebase tho.

scotmoor:
@spinnaker/netflix-reviewers rebased

PR #596, opened 2016-05-12
provider/google: Retrieve instance urls from cached server group inst…

duftler:
…ead of making live calls during Terminate, TerminateAndDecrement, AbandonAndDecrement operations.

provider/google: Don't require zone to query machine type.


duftler:
@lwander please review.

lwander:
Nice. LGTM

PR #598, opened 2016-05-12
provider/kubernetes: merge server group & job cluster

lwander:
When returning a single cluster from the cluster endpoint, we should merge the jobs & server groups inside that cluster rather than arbitrarily return one of the two clusters.

@duftler


duftler:
LGTMIS (I suppose).

lwander:
lambda

PR #606, opened 2016-05-13
Metrics around RI availability

ajordens:


ajordens:
I've been running this in prestaging and it seems to hang together.

Downside is that we don't really have a way of only publishing metrics from a single instance/leader ... I was able to get around it by using a `max` aggregate (vs. the default `sum`).

cfieber:
LGTM

PR #611, opened 2016-05-17
Fix an issue while editing an existing load balancer

milismsft:
Fix an issue while editing an existing load balancer (aka Azure Application Gateway); the public Ip resources was not handled properly in the deployment template.

Also part of this change I've disabled the caching for Azure Market Store VM images and Azure L4 Load Balancers (it reduces the redis cache footprint too); none of these resources are not exposed to the "deck" at the moment.


milismsft:
@spinnaker/netflix-reviewers @scotmoor PTAL

duftler:
LGTM.

PR #612, opened 2016-05-18
provider/google: Also consider account when filtering for security gr…

duftler:
…oups with no target tags.


duftler:
@ttomsu or @lwander please review.

lwander:
LGTM

PR #615, opened 2016-05-18
Make discovery enable / disable available for titus

tomaslin:


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AtomicOperations.groovy
```diff
@@ -41,6 +41,8 @@ final class AtomicOperations {
   public static final String ATTACH_CLASSIC_LINK_VPC = "attachClassicLinkVpc"
   public static final String REGISTER_INSTANCES_WITH_LOAD_BALANCER = "registerInstancesWithLoadBalancer"
   public static final String DEREGISTER_INSTANCES_FROM_LOAD_BALANCER = "deregisterInstancesFromLoadBalancer"
+  public static final String ENABLE_INSTANCES_IN_DISCOVERY = "enableInstancesInDiscovery"
```
cfieber: lets `s/discovery/eureka` these

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/discovery/AwsEurekaSupport.groovy
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    def regionScopedProvider = regionScopedProviderFactory.forRegion(credentials, region)
+    def eureka = regionScopedProvider.eureka
+    return eureka
+  }
+
+  @VisibleForTesting
+  @PackageScope
+  boolean verifyInstanceAndAsgExist(def credentials,
```
cfieber: can we generic this instead of `def`

cfieber:
LGTM

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AtomicOperations.groovy
tomaslin: will change enum name in this PR and then the operation name in subsequent one since it would require deck -> orca -> clouddriver change

PR #619, opened 2016-05-19
provider/kubernetes: More descriptive error messages

lwander:
Makes it easier to decipher errors in the logs with multiple accounts configured doing multiple operations.

@duftler

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
@@ -30,16 +30,18 @@ import java.util.concurrent.TimeUnit
 @Slf4j
 class KubernetesApiAdaptor {
   KubernetesClient client
+  String account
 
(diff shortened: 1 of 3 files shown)


duftler:
Nice, I like it.
Might read cleaner with a helper method to append the ` in $namespace.` piece.
Also, I noticed in the earlier pr that you were calling `.toString()` on an exception string message, but aren't here. Spidey-sense says this may have something to do with the gstring cast issue that you were fighting, but not sure.

lwander:
Ah yes, weirdly it hasn't crashed without `toString()`, but maybe I should add it just in case...

The helper `in $namespace` feels weird because some of these exceptions arise from manipulating namespaces.

lwander:
Ah, it didn't crash because the parameter to `super` in `KubernetesOperationException` does a cast.

duftler:
I had pictured something like:
`throwException(e, "Replace Ingress $name")
`throwException(e, "Get Replication Controllers")`

lwander:
hmm - I guess I could overload `throwException` to both take and not take a `namespace` parameter to handle both the non-namespace and namespace operations the same. I'll reopen this.

lwander:
Nvm, looks like reopening creates an extra PR on top of the one to make that change. I'll just open a new one.

PR #620, opened 2016-05-19
provider/kubernetes: Even better error messages

lwander:
I wound up returning the exception rather than throwing it in the method, because the intellij linter couldn't tell that calling a method that results in an exception will never return.

See https://github.com/spinnaker/clouddriver/pull/619 for context.

@duftler


duftler:
LGTM.

PR #621, opened 2016-05-20
provider/kubernetes: Extensions API bug workaround

lwander:
@duftler 

See https://github.com/fabric8io/kubernetes-client/issues/408 for context.

Recreating the client re-registers the adaptors in the above issue, temporarily fixing the problem.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
```diff
@@ -77,41 +78,53 @@ class KubernetesApiAdaptor {
     return true
   }
 
+  /*
+   * Delete once https://github.com/fabric8io/kubernetes-client/issues/408 is resolved
```
duftler: Could drop a `TODO(lwander):` here.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/SharedMutex.groovy
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2016 Netflix, Inc.
```
duftler: Copyright is off.

duftler:
LGTM.

PR #623, opened 2016-05-20
update spinnaker-dependencies to 0.35.0

cfieber:
fixes NPE in aws retry/backoff handlers when generating metrics tags


cfieber:

ajordens:
WCGW LGTM

PR #630, opened 2016-05-24
Feature client refactor

varikin:
Added support for both v2 and v3 Openstack authentication. 

We are going to start working on tests, but we wanted to get a feel for this approach while working on tests.


clouddriver-web/config/clouddriver.yml
```diff
@@ -255,6 +255,16 @@ google:
 
 openstack:
   enabled: false
+  accounts:
+    - name: test
+      environment: test
+      accountType : v2
+      master: test
```
lwander: This isn't a test file, but instead a fallback for config properties. Rather than name everything "test" I'd recommend either providing sane defaults, or omitting values you can't generically apply to each client.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/security/OpenstackNamedAccountCredentials.groovy
```diff
@@ -0,0 +1,81 @@
...
+  @JsonIgnore
+  final String password
+  final String tenantName
+  final String domainName
+  final String endpoint
+  final List<String> requiredGroupMembership
+  final OpenstackCredentials credentials
+
+  OpenstackNamedAccountCredentials(String accountName,
+                                          String environment,
```
lwander: Indentation is off

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/security/OpenstackNamedAccountCredentials.groovy
```diff
@@ -0,0 +1,81 @@
...
+                                          String username,
+                                          String password,
+                                          String tenantName,
+                                          String domainName,
+                                          String endpoint) {
+    this(accountName, environment, accountType, master, username, password, null, tenantName, domainName, endpoint)
+  }
+
+  OpenstackNamedAccountCredentials(String accountName,
+                                          String environment,
```
lwander: Indentation is off

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackProviderFactory.groovy
```diff
@@ -0,0 +1,59 @@
...
+    OSClient osClient
+    OpenstackClientProvider provider
+    if (AccountType.V2.value() == credentials.accountType) {
+      osClient = OSFactory.builderV2()
+        .endpoint(credentials.endpoint)
+        .credentials(credentials.username, credentials.password)
+        .tenantId(credentials.tenantName)
+        .authenticate()
+      provider = new OpenstackClientV2Provider(osClient)
+    } else {
```
lwander: Maybe check against `credentials.accountType` here as well, and throw an error in a separate `else` if it's not `v3` or `v2`.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackProviderFactory.groovy
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Builds the appropriate {@link OpenstackClientProvider} based on the configuration.
+ */
+class OpenstackProviderFactory {
+
+  static OpenstackClientProvider createProvider(OpenstackNamedAccountCredentials credentials) {
+    OSClient osClient
+    OpenstackClientProvider provider
+    if (AccountType.V2.value() == credentials.accountType) {
```
lwander: I think groovy allows you to use a switch statement on strings

lwander:
deps.txt probably shouldn't be in this PR?

lwander:
Looks like some of your unit tests are failing. Also - can you guys squash the history before the merge? Feel free to keep a commit from each author to assign credit.

lwander:
Is the cert issue you mentioned worked out? If so I'll merge now.

varikin:
Yes, there is an insecure option to have Openstack4j not verify SSL with the default to enable verification. We also just caught a Openstack v3 error that is included in that latest push

PR #634, opened 2016-05-25
Moves and generalizes AmazonAllowedAccountsValidator

ttomsu:
This will enable required group membership validation across all providers, assuming their `credentials` object supports it.

@spinnaker/google-reviewers PTAL

@cfieber @ajordens FYI


clouddriver-security/src/main/groovy/com/netflix/spinnaker/clouddriver/security/DefaultAllowedAccountsValidator.groovy
```diff
@@ -0,0 +1,66 @@
...
+  void validate(String user, Collection<String> allowedAccounts, Object description, Errors errors) {
+    if (!accountCredentialsProvider.all.find { it.requiredGroupMembership }) {
+      // no accounts have group restrictions so no need to validate / log
+      return
+    }
+
+    def json = null
+    try {
+      json = OBJECT_MAPPER.writeValueAsString(description)
+    } catch (Exception ignored) {
```
lwander: What's going on here?

clouddriver-security/src/test/groovy/com/netflix/spinnaker/clouddriver/security/DefaultAllowedAccountsValidatorSpec.groovy
```diff
@@ -0,0 +1,98 @@
...
+    when:
+    def description = new TestDescription(credentials: credentialsWithRequiredGroup)
+
+    validator.validate("TestUser", allowedAccounts, description, errors)
+
+    then:
+    rejectValueCount * errors.rejectValue("credentials", "unauthorized", _)
+
+    where:
+    allowedAccounts                      || rejectValueCount
```
lwander: This syntax seems off, I think you only only need a single `|`

lwander:
LGTM post comments

clouddriver-security/src/main/groovy/com/netflix/spinnaker/clouddriver/security/DefaultAllowedAccountsValidator.groovy
ttomsu: Looks like it's used only for debugging. This was migrated from the other version.

clouddriver-security/src/test/groovy/com/netflix/spinnaker/clouddriver/security/DefaultAllowedAccountsValidatorSpec.groovy
ttomsu: It's legit -https://spockframework.github.io/spock/docs/1.0/data_driven_testing.html#_syntactic_variations

ajordens: Should use `||` when separating inputs from expected outputs.

clouddriver-security/src/main/groovy/com/netflix/spinnaker/clouddriver/security/DefaultAllowedAccountsValidator.groovy
ajordens: You changed the copyright ... you own it now!

ajordens:
From our earlier conversation ... I don't imagine this is going to be a useful class long term with the move to per-application and per-account read/write permissions (_split out into a separate service_).

That said, if there's some short-term benefit for locking down write-access than great!

ttomsu:
It's a step forward in the right direction, which is why I generalized this class. Once everything in Gate is merged and released, we can finally have a good, even if not yet complete, story how to lock down access.

PR #638, opened 2016-05-26
Azure bug fix for disableSG and isDisable implementation

milismsft:
Fix for a bug in disable server group; during destroy SG we call disableop and if a different SG is active, that gets disabled as well.
Added couple more logging lines to keep track of the "orchestration" progress when creating or disabling SGs.
Implemented isDisabled for a server group; also changed enable and disable ops to take advantage of this check and fast track the op if no updates are needed.


milismsft:
@spinnaker/netflix-reviewers @duftler PTAL and merge if LGTY ;-)

ajordens:
Needs a rebase

milismsft:
rebased

PR #639, opened 2016-05-26
It's log!

cfieber:
What rolls down stairs
alone or in pairs,
and over your neighbor's dog?
What's great for a snack,
And fits on your back?
It's log, log, log

It's log, it's log,
It's big, it's heavy, it's wood.
It's log, it's log, it's better than bad, it's good."

Everyone wants a log
You're gonna love it, log
Come on and get your log
Everyone needs a log
log log log

@lwander PTAL


lwander:
how did that get there!

lwander:
Oh boy I left that there since February

PR #642, opened 2016-05-26
Operation to terminate Openstack instance.

drmaas:
This is just a noop with all the framework to get the operation invoked.

terminate instance changes

Successfully deleted instance

adding specs for delete operations

Tests for OpenstackClientProvider handle request.

checkpoint

undo changes to kubernetes validation, some validation logic duplicated in openstack

### DIFF
--- a/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
+++ b/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
@@ -16,35 +16,56 @@
 
 package com.netflix.spinnaker.clouddriver.openstack.client
 
+import com.netflix.spinnaker.clouddriver.openstack.deploy.exception.OpenstackOperationException
+import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperations
 import org.openstack4j.api.OSClient
+import org.openstack4j.model.common.ActionResponse
 
 /**
  * Provides access to the Openstack API.
+ *
+ * TODO tokens will need to be regenerated if they are expired.
  */
 abstract class OpenstackClientProvider {
 
-  OSClient client
-
-  OpenstackClientProvider(OSClient client) {
-    this.client = client
-  }
-
-  //TODO test
   /**
    * Delete an instance.
    * @param instanceId
    * @return
    */
-  def deleteInstance(String instanceId) {
-    client.compute().servers().delete(instanceId)
+  void deleteInstance(String instanceId) {
+    handleRequest(AtomicOperations.TERMINATE_INSTANCES) {
+      client.compute().servers().delete(instanceId)
+    }
   }
+
+  /**
+   * Handler for an openstack4j request.
+   * @param closure
+   * @return
+   */
+  ActionResponse handleRequest(String operation, Closure closure) {
+    ActionResponse result
+    try {
+      result = closure()
+    } catch (Exception e) {
+      throw new OpenstackOperationException(operation, e)
+    }
+    if (!result.isSuccess()) {
+      throw new OpenstackOperationException(result, operation)
+    }
+    result
+  }
+
+  /**
+   * Thread-safe way to get client.
+   * @return
+   */
+  abstract OSClient getClient()
 
   /**
    * Get a new token id.
    * @return
    */
   abstract String getTokenId()
-
-  //TODO stuff common to v2 and v3
-
 }
--- a/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/description/instance/OpenstackInstancesDescription.groovy
+++ b/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/description/instance/OpenstackInstancesDescription.groovy
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2016 Target, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.openstack.deploy.description.instance
+
+import com.netflix.spinnaker.clouddriver.openstack.deploy.description.OpenstackAtomicOperationDescription
+
+class OpenstackInstancesDescription extends OpenstackAtomicOperationDescription {
+  List<String> instanceIds
+}
--- a/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/exception/OpenstackOperationException.groovy
+++ b/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/exception/OpenstackOperationException.groovy
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2016 Target, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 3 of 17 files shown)


clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
```diff
@@ -16,35 +16,56 @@
...
-    client.compute().servers().delete(instanceId)
+  void deleteInstance(String instanceId) {
+    handleRequest(AtomicOperations.TERMINATE_INSTANCES) {
+      client.compute().servers().delete(instanceId)
+    }
   }
 
   /**
-   * Get a new token id.
+   * Handler an openstack4j request.
```
varikin: Handler *for* an openstack4j request

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/ops/instance/TerminateOpenstackInstancesAtomicOperation.groovy
```diff
@@ -0,0 +1,60 @@
...
+   * curl -X GET -H "Accept: application/json" localhost:7002/task/1
+   */
+  @Override
+  Void operate(List priorOutputs) {
+    task.updateStatus BASE_PHASE, "Initializing terminate instances operation..."
+
+    log.info(description.toString())
+
+    description.instanceIds.each {
+      log.info("Deleting $it")
```
varikin: @drmaas Should we change this logging to debug or remove it now? I know we initially added it while trying to get this working.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/validators/OpenstackAttributeValidator.groovy
```diff
@@ -0,0 +1,107 @@
+package com.netflix.spinnaker.clouddriver.openstack.deploy.validators
```
varikin: Missed the license on this file.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/validators/instance/TerminateOpenstackInstancesDescriptionValidator.groovy
```diff
@@ -0,0 +1,46 @@
...
+import com.netflix.spinnaker.clouddriver.openstack.deploy.description.instance.OpenstackInstancesDescription
+import com.netflix.spinnaker.clouddriver.openstack.deploy.validators.OpenstackAttributeValidator
+import com.netflix.spinnaker.clouddriver.security.AccountCredentialsProvider
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.validation.Errors
+
+/**
+ *
+ */
+class TerminateOpenstackInstancesDescriptionValidator extends DescriptionValidator<OpenstackInstancesDescription> {
```
varikin: @drmaas I think we missed a test for this class. We should ensure perform the proper validation for this op.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
```diff
@@ -16,35 +16,56 @@
...
 
+import com.netflix.spinnaker.clouddriver.openstack.deploy.exception.OpenstackOperationException
+import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperations
 import org.openstack4j.api.OSClient
+import org.openstack4j.model.common.ActionResponse
 
 /**
  * Provides access to the Openstack API.
+ *
+ * TODO tokens will need to be regenerated if they are expired.
```
lwander: How long do tokens usually last? This feels like something the underlying API should provide?

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/description/instance/OpenstackInstancesDescription.groovy
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.openstack.deploy.description.instance
+
+import com.netflix.spinnaker.clouddriver.openstack.deploy.description.OpenstackAtomicOperationDescription
+
+class OpenstackInstancesDescription extends OpenstackAtomicOperationDescription {
+  List<String> instanceIds
```
lwander: Does `region` need to show up somewhere to scope this request?

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/exception/OpenstackOperationException.groovy
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.openstack.deploy.exception
+
+import groovy.transform.InheritConstructors
+import org.openstack4j.model.common.ActionResponse
+
+@InheritConstructors
+class OpenstackOperationException extends RuntimeException {
```
lwander: Very nice

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/ops/instance/TerminateOpenstackInstancesAtomicOperation.groovy
lwander: I recommend logging everything through this [`task`](https://github.com/spinnaker/clouddriver/pull/642/files#diff-92e8324c29a36a9c52049c9286b2cc1aR48) object, since it also reports output to Orca which can be read through Gate/Deck. Makes it easy to see why something went wrong just by looking through the frontend under the "Tasks" section.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
varikin: From a quick glance at documentation, 24 hours, though I need to verify that between v2 (uses access object) and v3 (uses token). Though I do think there is an expiration date on the token or access object that can be checked.

I would like to move the logic that connects and gets the token out of the factory and into `OpenstackClientProvider.getClient`. Then instead of a simple getter, it automatically handles reauthenticating for a new token if needed.

All this is to deal with using the `OSClient` object in a multi-threaded environment. Openstack4j seems to tie that object to a thread (well, throws errors if it is shared across threads), so we have to create a new client object for the openstack session using the cached token.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/description/instance/OpenstackInstancesDescription.groovy
varikin: For this operation, no, since we are dealing with instance ids which are unique across regions. We expect to be adding to this as needed for other operations, depending on the needs.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
drmaas: Agreed, I will leave the TODO and we will need to address at some point.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/validators/instance/TerminateOpenstackInstancesDescriptionValidator.groovy
drmaas: Added test

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/validators/OpenstackAttributeValidator.groovy
drmaas: added license

lwander:
LGTM - ready to merge after this is squashed

PR #631, opened 2016-05-25
Makes all AWS services interactions done via AmazonClientProvider

zaletniy:
Proxy configuration was not used in DefaultAWSAccountInfoLookup,
as result clouddriver-web was not able to start with 'Connection
timeout' error.

PR #643, opened 2016-05-26
Makes all AWS services interactions done via AmazonClientProvider

cfieber:
Some rework of #631

Closes #631


cfieber:
@zaletniy FYI (just a couple minor tweaks to your original implementation and a bit of cleanup in AmazonClientProvider to consolidate all the different ways that a cliend can be requested)

PR #646, opened 2016-05-27
Implemented openstack instance reboot and fixed terminate validation bug

drmaas:


clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/validators/instance/RebootOpenstackInstancesDescriptionValidator.groovy
```diff
@@ -0,0 +1,46 @@
...
+class RebootOpenstackInstancesDescriptionValidator extends DescriptionValidator<OpenstackInstancesDescription> {
+
+  @Autowired
+  AccountCredentialsProvider accountCredentialsProvider
+
+  @Override
+  void validate(List priorDescriptions, OpenstackInstancesDescription description, Errors errors) {
+    def validator = new OpenstackAttributeValidator("rebootOpenstackInstancesAtomicOperationDescription", errors)
+
+    //validate credentials
```
duftler: This comment and the next don't really seem to add information. No need to include them in that case.

duftler:
LGTM after addressing the minor comment. I'd lean towards striking those same comments from the other validator as well.

derekolk:
Cool ... I will remove those comments in both validators.  Give me a second to address :)

duftler:
LGTM.

PR #650, opened 2016-05-31
provider/docker: Make tracking digests optional

lwander:
@duftler 
@tomaslin FYI

By default tracking digest changes is now disabled. The end goal is to trigger pipelines only from tags being added to a specific repository. Once this is finished (changes to igor inbound), this will be the only place this behavior needs to be configured.

When enabled, there is a slight change. If the registry responds with a 404 during the digest lookup the tag is removed from the cache as before. If any other error is encountered, the tag is left in the cache but without a digest. This will be handled gracefully in igor (digests flipping between null & some value won't trigger pipelines).


duftler:
LGTM.

PR #655, opened 2016-06-03
Adds FakeGoogleCredentials for unit tests to stop relying on user bei…

ttomsu:
…ng gcloud auth'ed

Tested with:
```
$ gcloud auth revoke
$ ./gradlew clean test
```

@duftler PTAL
@cfieber - I don't know how to properly add a dependency on another modules _test_ classes. Does this approach look right to you?

### DIFF
--- a/clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/security/FakeGoogleCredentials.groovy
+++ b/clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/security/FakeGoogleCredentials.groovy
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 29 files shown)


clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/security/FakeGoogleCredentials.groovy
```diff
@@ -0,0 +1,45 @@
...
+    this("unitTestProject")
+  }
+
+  FakeGoogleCredentials(String project) {
+    super(project, ComputeVersion.V1)
+  }
+
+  @Override
+  HttpTransport buildHttpTransport() {
+    return new MockHttpTransport.Builder().build()
```
duftler: Nice.

cfieber:
Just taking a quick look at an alternative to referencing that test class

duftler:
LGTM once @cfieber is happy with the test dependency handling.

cfieber:
have a look at https://github.com/ttomsu/clouddriver/pull/1 - it makes sure the test runtime dependencies come along for the ride in addition to the classes, and it is supposed to play nicer with IDEs.

otherwise LGTM, thanks for the fix!

PR #656, opened 2016-06-03
Updates CopyLastAsgAtomicOperation to pull more from the source field.

cfieber:
The minimum requred payload becomes `credentials` + `source` to make a copy of an existing server group, and then any additional fields to make changes from the `source`'s configuration. `credentials` is required because asserting that exists is baked pretty far up the operation handling stack (could theoretically have pulled that from `source.account`)
Normalizes security groups to names when filling defaults that cross an account or zone
Resets keypair to the defaultKeyPair when crossing account or region and the keyPair wasn't explicitly passed

Also runs the validation after the final `BasicAmazonDeployDescription` has been constructed.

It is possible to issue a cloneServerGroup and the source not be found - so long as it is a fully formed deploy it will succeed, otherwise it will fail validation. Previously it would just fail with an NPE when the source could not be located

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/OperationsController.groovy
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/OperationsController.groovy
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.clouddriver.controllers
 import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.clouddriver.data.task.Task
 import com.netflix.spinnaker.clouddriver.deploy.DescriptionValidationErrors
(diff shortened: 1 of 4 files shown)


cfieber:

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/OperationsController.groovy
```diff
@@ -201,9 +202,4 @@ class OperationsController {
     AtomicOperation atomicOperations
     Errors errors
   }
-
-  @Canonical
-  static class ValidationException extends RuntimeException {
```
cfieber: moved this into clouddriver-core as `DescriptionValidationException`

PR #657, opened 2016-06-03
Added support for default aws keypairs

danielpeach:
Added support for default aws keypairs. @lwander @duftler PTAL


cfieber:
This seems like an unnecessary change

The defaultKeypairTemplate doesn't necessarily have to have any placeholders, so if you have a static value, just configure it there

If you need to override something at the account level you can also specify the keypair directly on the account (with or without templating)

cfieber:
@danielpeach does my comment above make sense? I'm going to close this PR but if I'm missing something you are trying to address please let me know

danielpeach:
Yes -- not a problem.

PR #658, opened 2016-06-03
expose per-deploy option to change udf generation behaviour

cfieber:
right now we don't have a great story for people looking to migrate from legacy udf to the newer style, this flag allows someone to deploy with the specific behaviour before flipping the flag at the application level


cfieber:

ajordens:
LGTM

PR #660, opened 2016-06-04
provider/google: Avoid NPE in load balancer caching agent.

duftler:
provider/google: Avoid NPE in load balancer controller when no matching load balancer is found.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController.groovy
@@ -73,6 +73,10 @@ class GoogleLoadBalancerController {
       view.account == account && view.region == region
     }
 
+    if (!view) {
+      return []
+    }
+
     [new GoogleLoadBalancerDetails(loadBalancerName: view.name,
                                    createdTime: view.createdTime,
                                    dnsname: view.ipAddress,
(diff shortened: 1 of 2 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleLoadBalancerCachingAgent.groovy
```diff
@@ -234,7 +234,7 @@ class GoogleLoadBalancerCachingAgent extends AbstractGoogleCachingAgent implemen
   class HttpHealthCheckCallback<HttpHealthCheck> extends JsonBatchCallback<HttpHealthCheck> implements FailureLogger {
 
     GoogleLoadBalancer googleLoadBalancer
-    TargetPool targetPool
+    def targetPool
```
duftler: We were encountering what looked like this: https://jira.atlassian.com/plugins/servlet/mobile#issue/CLOV-1779/comment/778908

Cache population was failing with NPEs resulting in no populated load balancers.

duftler:
TBR @ttomsu .

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController.groovy
```diff
@@ -73,6 +73,10 @@ class GoogleLoadBalancerController {
       view.account == account && view.region == region
     }
 
+    if (!view) {
+      return []
```
ttomsu: This should probably throw a 404 to be consistent, no?

duftler: I did it this way to be consistent with the entry-point that just takes `name`, and with the behavior of `aws/loadBalancers`.

PR #662, opened 2016-06-06
provider/google: Rely on ComputeVersion enum instead of exposing alph…

duftler:
…aListed boolean.


duftler:
@ttomsu please review.

ttomsu:
LGTM

PR #665, opened 2016-06-07
provider/google: Remove google-containers image source

lwander:
`google-containers` is no longer included in the list of images `gcloud images list` returns, so we stop supplying it in the default list of images.

@duftler


duftler:
LGTM.

PR #670, opened 2016-06-09
Delete load balancer operation.

derekolk:
Implemented openstack delete load balancer operation.


clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
```diff
@@ -282,6 +284,42 @@ abstract class OpenstackClientProvider {
   }
 
   /**
+   * Remove load balancer pool.
+   * @param region
```
lwander: While we're normally not great a generating code docs, it feels strange to see these `param` annotations with descriptions of the variables being passed in.

lwander:
LGTM, any objections to merging?

ashleykasim:
Looks good to me as well, @lwander go ahead and merge

ashleykasim:
https://github.com/spinnaker/spinnaker/issues/915

PR #671, opened 2016-06-09
Azure create subnet fixes and other small tweaks

milismsft:
Moved Azure Subnet and VirtualNetwork description creation from AzureNetworkClient into the respective description classes and disabled the Azure Subnet caching agent for now; we already get the subnets through the Azure network caching agents.
Default to premium storage.
Added ipv4 representation for the subnet address prefix which will be used to calculate next subnet address prefix using a randomly selected seed.
Compute and select the next subnet using the cached virtual network and using a randomly generated seed.
Added test cases for AzureVirtualNetworkDescription class and methods and the new methods in the AzureUtilities.

### DIFF
--- a/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/network/view/AzureNetworkProvider.groovy
+++ b/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/network/view/AzureNetworkProvider.groovy
@@ -27,7 +27,12 @@ import com.netflix.spinnaker.clouddriver.azure.resources.network.model.AzureVirt
 import com.netflix.spinnaker.clouddriver.model.NetworkProvider
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestMethod
+import org.springframework.web.bind.annotation.RestController
 
+@RestController
 @Component
 class AzureNetworkProvider implements NetworkProvider<AzureNetwork> {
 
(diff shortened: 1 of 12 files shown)


milismsft:

milismsft:
ping

ajordens:
Nothing outside of azure packages touched so LGTM.

In general, it'd be nice to have another LGTM from Microsoft/Azure do a once over on the Azure-specific changes as it's unlikely that anyone on `netflix-reviewers` has appropriate context (ie. no idea what Premium storage is).

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/network/view/AzureNetworkProvider.groovy
```diff
@@ -52,6 +57,17 @@ class AzureNetworkProvider implements NetworkProvider<AzureNetwork> {
     cacheView.getAll(Keys.Namespace.AZURE_NETWORKS.ns, RelationshipCacheFilter.none()).collect(this.&fromCacheData)
   }
 
+  @RequestMapping(value = "/azure/vnet/{account}/{region}/{name:.+}", method = RequestMethod.GET)
```
scotmoor: Why are we exposing endpoints in the provider class and not through a controller? Who is expected to be calling this outside the two methods in cloud driver?

milismsft: I've added his mapping for testing purposes since the other mapping only provides us with a summary of the vnet and not the full description of the vnet cached object; but it can be removed as it's not needed outside of that scope (deck does not use it at the moment) and it's easy to add it back if we'll need it later.

scotmoor:
LGTM

milismsft:
Ping

PR #676, opened 2016-06-10
Docker registry: prevent errors from getTags terminating polling agent

tomaslin:
sometimes the docker registry will have incorrect indexing of repositories. Seeing that it currently terminates the poling process whereas it should just be ignored.


tomaslin:
@lwander PTAL

lwander:
LGTM

PR #685, opened 2016-06-15
provider/aws: Replace one missed amazonCloudProvider.id with AmazonCl…

duftler:
…oudProvider.AWS. Without this fix, no provider-specific subnets can be retrieved.


duftler:
@cfieber fyi

cfieber:
shame

PR #691, opened 2016-06-18
provider/cloudfoundry: Add version specific route for canary testing

gregturn:
To support canary testing, Cloud Foundry-based server groups need a version-specific route ALSO mapped during deployment. This is also handled during enable/disable server group operations.


gregturn:
@spinnaker/reviewers PTAL

lwander:
LGTM

PR #693, opened 2016-06-21
Feature refactor create servergroup and servergroup clone

drmaas:
Closes spinnaker/spinnaker#937

### DIFF
--- a/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
+++ b/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
@@ -18,31 +18,33 @@ package com.netflix.spinnaker.clouddriver.openstack.client
 
 import com.netflix.spinnaker.clouddriver.openstack.deploy.exception.OpenstackOperationException
 import com.netflix.spinnaker.clouddriver.openstack.deploy.exception.OpenstackProviderException
+import com.netflix.spinnaker.clouddriver.openstack.deploy.exception.OpenstackResourceNotFoundException
 import com.netflix.spinnaker.clouddriver.openstack.domain.LoadBalancerPool
 import com.netflix.spinnaker.clouddriver.openstack.domain.PoolHealthMonitor
+import com.netflix.spinnaker.clouddriver.openstack.domain.ServerGroupParameters
 import com.netflix.spinnaker.clouddriver.openstack.domain.VirtualIP
-import com.netflix.spinnaker.clouddriver.openstack.deploy.exception.OpenstackResourceNotFoundException
 import org.apache.commons.lang.StringUtils
 import org.openstack4j.api.Builders
 import org.openstack4j.api.OSClient
 import org.openstack4j.api.networking.ext.LoadBalancerService
 import org.openstack4j.model.common.ActionResponse
-import org.openstack4j.model.compute.IPProtocol
 import org.openstack4j.model.compute.FloatingIP
+import org.openstack4j.model.compute.IPProtocol
 import org.openstack4j.model.compute.RebootType
+import org.openstack4j.model.compute.SecGroupExtension
+import org.openstack4j.model.compute.Server
 import org.openstack4j.model.heat.Resource
+import org.openstack4j.model.heat.Stack
+import org.openstack4j.model.heat.StackCreate
 import org.openstack4j.model.network.NetFloatingIP
 import org.openstack4j.model.network.Port
 import org.openstack4j.model.network.ext.HealthMonitor
 import org.openstack4j.model.network.ext.HealthMonitorType
 import org.openstack4j.model.network.ext.LbMethod
 import org.openstack4j.model.network.ext.LbPool
+import org.openstack4j.model.network.ext.Member
 import org.openstack4j.model.network.ext.Protocol
 import org.openstack4j.model.network.ext.Vip
-import org.openstack4j.model.compute.SecGroupExtension
-import org.openstack4j.model.compute.Server
-import org.openstack4j.model.heat.Stack
-import org.openstack4j.model.network.ext.Member
 
 import java.lang.reflect.UndeclaredThrowableException
 import java.util.regex.Matcher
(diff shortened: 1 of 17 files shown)


drmaas:
@spinnaker/reviewers

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
```diff
@@ -526,7 +544,7 @@ abstract class OpenstackClientProvider {
       client.useRegion(region).heat().stacks().getStackByName(stackName)
     }
     if (!stack) {
-      throw new OpenstackResourceNotFoundException("Unable to find stack $stackName in region $region")
+      throw new OpenstackProviderException("Unable to find stack $stackName in region $region")
```
lwander: Do you explicitly cast `.toString()` in the constructor to whatever exception this subclasses? I just remembered that my custom Kubernetes exceptions originally failed because they expected a Java string, but got a Groovy string due to the `$templating` I use.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/ops/servergroup/CloneOpenstackAtomicOperation.groovy
```diff
@@ -43,44 +42,62 @@ class CloneOpenstackAtomicOperation implements AtomicOperation<DeploymentResult>
...
+    deployDescription.serverGroupParameters.with {
+      image = description.serverGroupParameters?.image ?: ancestorParams.image
+      instanceType = description.serverGroupParameters?.instanceType ?: ancestorParams.instanceType
+      maxSize = description.serverGroupParameters?.maxSize ?: ancestorParams.maxSize
+      minSize = description.serverGroupParameters?.minSize ?: ancestorParams.minSize
+      networkId = description.serverGroupParameters?.networkId ?: ancestorParams.networkId
+      poolId = description.serverGroupParameters?.poolId ?: ancestorParams.poolId
+      securityGroups = description.serverGroupParameters?.securityGroups ?: ancestorParams.securityGroups
+    }
+    deployDescription.disableRollback = description.disableRollback ?: false
```
lwander: Why not take the ancestor value?

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/ops/servergroup/DeployOpenstackAtomicOperation.groovy
```diff
@@ -18,14 +18,22 @@ package com.netflix.spinnaker.clouddriver.openstack.deploy.ops.servergroup
...
+import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperations
+import org.apache.commons.io.IOUtils
+import org.openstack4j.model.network.ext.LbPool
+
+class DeployOpenstackAtomicOperation implements AtomicOperation<DeploymentResult> {
 
-@Slf4j
-class DeployOpenstackAtomicOperation implements AtomicOperation<Void> {
   private final String BASE_PHASE = "DEPLOY"
+  private final String TEMPLATE_FILE = 'asg.yaml'
```
lwander: Can you comment on what these strings mean?

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/validators/servergroup/CloneOpenstackAtomicOperationValidator.groovy
```diff
@@ -35,39 +35,45 @@ class CloneOpenstackAtomicOperationValidator extends DescriptionValidator<CloneO
...
-    if (description.timeoutMins) {
-      helper.validateNonNegative(description.timeoutMins, "timeoutMins")
+  //TODO this is copy-paste from DeployOpenstackAtomicOperationValidator
+  def validateServerGroup(OpenstackAttributeValidator validator, ServerGroupParameters parameters) {
+    String prefix = "serverGroupParameters"
+    parameters.with {
+      validator.validateNotEmpty(instanceType, "${prefix}.instanceType")
+      validator.validateNotEmpty(image, "${prefix}.image")
+      validator.validateNotNull(maxSize, "${prefix}.maxSize")
+      validator.validatePositive(maxSize, "${prefix}.maxSize")
```
lwander: Seems like your `validatePositive` now also checks that the value isn't `null`, so those `NotNull` checks seem superfluous

clouddriver-openstack/src/main/resources/asg.yaml
```diff
@@ -0,0 +1,79 @@
+heat_template_version: 2015-04-30
```
lwander: What's the strategy for distributing these/making the customizable?

lwander:
LGTM after comments are addressed.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
drmaas: We're using ```@InheritConstructors```, which seem to work fine.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/ops/servergroup/CloneOpenstackAtomicOperation.groovy
drmaas: It is not available from the ```org.openstack4j.model.heat.Stack``` object.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/validators/servergroup/CloneOpenstackAtomicOperationValidator.groovy
drmaas: good catch

clouddriver-openstack/src/main/resources/asg.yaml
drmaas: Good question - I think for now we want to provide 'the standard' way of being able to configure an autoscaling group, with the hardcoded templates. Later on add in the feature to provide custom templates.

Overriding the default via configuration is a good idea, as long as people do their diligence to honor the properties that the template can expect to be given to it. The Openstack API is finicky when properties are provided but not used, and doesn't work at all when properties are not provided but expected.

Being able to pass in the template via free-form text is also a good idea, but again it would need to honor the expected parameters. We could use the freeform details field to store the template string.

Added a comment in the code per ^^^.

lwander:
LGTM, will merge once squashed

PR #697, opened 2016-06-21
UpsertSecurityGroup should go straight to AWS to retrieve security gr…

ajordens:
…oups (vs using edda if available)

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/SecurityGroupLookupFactory.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/SecurityGroupLookupFactory.groovy
@@ -3,6 +3,8 @@ package com.netflix.spinnaker.clouddriver.aws.model
 import com.amazonaws.services.ec2.AmazonEC2
 import com.amazonaws.services.ec2.model.AuthorizeSecurityGroupIngressRequest
 import com.amazonaws.services.ec2.model.CreateSecurityGroupRequest
+import com.amazonaws.services.ec2.model.DescribeSecurityGroupsRequest
+import com.amazonaws.services.ec2.model.Filter
 import com.amazonaws.services.ec2.model.IpPermission
 import com.amazonaws.services.ec2.model.RevokeSecurityGroupIngressRequest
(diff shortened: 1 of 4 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/SecurityGroupLookupFactory.groovy
```diff
@@ -40,7 +42,7 @@ class SecurityGroupLookupFactory {
     private final String region
     private final ImmutableSet<NetflixAmazonCredentials> accounts
 
-    private final Map<String, List<SecurityGroup>> securityGroupsByAccount = [:]
+    private final Map<String, SecurityGroup> securityGroupByName = [:]
```
ajordens: Honestly not sure the importance of keep an instance lifetime cache of security groups.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/SecurityGroupLookupFactory.groovy
```diff
@@ -70,28 +72,31 @@ class SecurityGroupLookupFactory {
...
-      final securityGroup = getSecurityGroups(accountName, amazonEC2, skipCache).find {
+
+      def amazonEC2 = amazonClientProvider.getAmazonEC2(credentials, region, true)
+      def cachedSecurityGroupKey = name.toLowerCase() + "." + vpcId
+      def cachedSecurityGroup = securityGroupByName.get(cachedSecurityGroupKey)
+      if (cachedSecurityGroup) {
+        return new SecurityGroupUpdater(cachedSecurityGroup, amazonEC2)
+      }
+
+      def describeSecurityGroupsRequest = new DescribeSecurityGroupsRequest().withFilters(
```
ajordens: Verified that this results in only a single call to AWS (ie. it doesn't fetch the world)

ajordens:
PTAL @cfieber 

This came about as a result of script users that:
- create a new security group
- immediately upsert security group to add self-referencial ingress

With `edda` in the mix, there's a chance that the subsequent upsert would use stale data from edda and fail validation.

PR #698, opened 2016-06-21
provider/cloudfoundry: Set default repository details

gregturn:
clouddriver.yml, the default settings for using CF, needs to have repository username/password fields set.


gregturn:
@spinnaker/reviewers PTAL

duftler:
LGTM.

PR #699, opened 2016-06-21
provider/google: added name field to GoogleHealthCheck

nwwebb:
@lwander @duftler 

Added a name field to the GoogleHealthCheck class and updated the agent to provide this field. I need the name of the health check in order to properly serialize it. Terraform uses the name to determine if it needs to create a new resource or edit an existing one. The name of a health check is determined in part by the time it was created, so it is not possible to reconstruct this name from the current fields in the model. Adding the name field fixes this issue.


lwander:
LGTM. WDYT Matt?

duftler:
LGTM.
Needs a rebase.

lwander:
:confetti_ball:

PR #702, opened 2016-06-22
provider/cloudfoundry: Fix createdTime for out of order instances

gregturn:
When cloudfoundry fetches createdTime for a server group running multiple instances, it simply grabs the first one. It should actually look at all instances and grab the oldest.


gregturn:
@spinnaker/reviewers PTAL pending passing the CI checks.

lwander:
LGTM

PR #704, opened 2016-06-22
Adds an agent to reconcile the classic link security groups.

cfieber:
During a deploy the classiclink and app specific groups can be attached.
This agent looks for instances that do not conform and adds the security groups.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/agent/ReconcileClassicLinkSecurityGroupsAgent.java
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/agent/ReconcileClassicLinkSecurityGroupsAgent.java
@@ -1,0 +1,232 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/agent/ReconcileClassicLinkSecurityGroupsAgent.java
```diff
@@ -0,0 +1,229 @@
...
+      .forEach(i -> {
+        List<String> existingClassicLinkGroups = i.getGroups().stream()
+          .map(GroupIdentifier::getGroupId)
+          .collect(Collectors.toList());
+
+        int maxNewGroups = deployDefaults.getMaxClassicLinkSecurityGroups() - existingClassicLinkGroups.size();
+        i.getTags()
+          .stream()
+          .filter(t -> AUTOSCALING_TAG.equals(t.getKey()))
+          .map(Tag::getValue)
```
ajordens: Might be a little difficult to follow what's going on here ... maybe a quick example comment for each map:

```
.map(Tag::getValue)  // clouddriver-main-v001
...
.map(this::getSecurityGroupNames) // clouddriver
```

ajordens:
Feels like a few inline comments around the agent would help those without intimate knowledge of what's going on.

cfieber:
broke up stream processing into more logical chunks and added a few more test cases

cfieber:
@ajordens let me know if this looks more reasonable

ajordens:
LGTM

PR #705, opened 2016-06-23
provider/kubernetes: respond to readiness probe failures

lwander:
@duftler 

When the container is running, but `ready := false`, the probe is failing, and the container should be marked as down.


lwander:
pointed out by @cholin-bmc

duftler:
LGTM.

cholin-bmc:
Thanks for the fix

PR #706, opened 2016-06-23
(gce) updated vcpu max list

danielpeach:
- (gce) updated vcpu max list.
- also updated link to corresponding list in deck.

@duftler PTAL


duftler:
LGTM.

PR #707, opened 2016-06-23
provider/aws: Define BasicAmazonDeployDescription.Capacity{min/max/de…

duftler:
…sired} as Integers instead of as ints. This allows the null-checks to work properly.


duftler:
@cfieber please review.

cfieber:
LGTM

PR #710, opened 2016-06-24
Removed version from URIs to support Spring Cloud Eureka. Fixed #956

ervansetiawan:
As discussed with @cfieber via Slack, this change removes the Eureka API version from the URI and let that be configured as part of the DiscoveryURL in the clouddriver.yml instead so we can support Spring Cloud Eureka.


cfieber:
LGTM thanks

I will hit all the buttons once the build finishes

PR #716, opened 2016-06-27
provider/kubernetes: s/launchTime/createdTime for consistency

lwander:
@danielpeach just for you


lwander:
Server groups surface `createdTime`, whereas jobs surface `launchTime`, which makes them difficult to render consistently in Deck. This change fixes that.

PR #721, opened 2016-06-28
Documentation to include notes about primaryCredentials

edwinavalos:
I believe this is accurate from my testing. Rosco bakes will have an ownership of `providers.aws.primaryCredentials.name` when the default account doesn't exist.

There was also some weird whitespacing changes in this PR making it longer than I had intended, but it doesn't have any effect on rendering, let me know if I should fix it to minimize the changes to what I touched.


BetaKevin:
👍

PR #727, opened 2016-06-29
Feature lb network

drmaas:
Have the floating IP get pulled from the pool or allocated based on the network id used.

### DIFF
--- a/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
+++ b/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
@@ -37,6 +37,7 @@ import org.openstack4j.model.heat.Stack
 import org.openstack4j.model.heat.StackCreate
 import org.openstack4j.model.heat.StackUpdate
 import org.openstack4j.model.network.NetFloatingIP
+import org.openstack4j.model.network.Network
 import org.openstack4j.model.network.Port
 import org.openstack4j.model.network.Subnet
 import org.openstack4j.model.network.ext.HealthMonitor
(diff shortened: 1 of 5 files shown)


drmaas:
@spinnaker/reviewers

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
```diff
@@ -346,6 +347,34 @@ abstract class OpenstackClientProvider {
...
+   * Get an unallocated IP from the network, or if none are found, try to create a new floating IP in the network.
+   * @param region
+   * @param networkName
+   * @return
+   */
+  FloatingIP getOrCreateFloatingIp(final String region, final String networkName) {
+    handleRequest {
+      FloatingIP ip = getRegionClient(region).compute().floatingIps().list().find { !it.fixedIpAddress }
+      if (!ip) {
+        ip = client.useRegion(region).compute().floatingIps().allocateIP(networkName)
```
lwander: Will this throw if it fails to allocate, or return `null`?

lwander:
LGTM post comment

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/client/OpenstackClientProvider.groovy
drmaas: Good question, it throws a ClientResponseException with a status code of 403, which we catch and rethrow as an OpenstackProviderException.

lwander: In that case, shouldn't this throw an exception if `ip` is `null` here?

drmaas: IP can be null from the first call. The second call shouldn't return null if it allocates a new IP without throwing an exception. I can add a check though just to be sure.

PR #729, opened 2016-07-01
provider/kubernetes: Probe HTTP headers

lwander:
@danielpeach FYI

@duftler PTAL

This was missing from the fabric8 API when I first added probes (so it's in the deploy description, but previously couldn't be used). Now it'll actually create probes with HTTP headers attached.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
@@ -240,6 +240,15 @@ class KubernetesApiConverter {
               containerBuilder = containerBuilder.withScheme(get.uriScheme)
             }
 
+            if (get.httpHeaders) {
+              def headers = get.httpHeaders.collect() {
+                def builder = new HTTPHeaderBuilder()
+                return builder.withName(it.name).withValue(it.value).build()
+              }
+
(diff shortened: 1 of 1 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
```diff
@@ -515,6 +524,9 @@ class KubernetesApiConverter {
     kubernetesHttpGetAction.path = httpGet.path
     kubernetesHttpGetAction.port = httpGet.port?.intVal
     kubernetesHttpGetAction.uriScheme = httpGet.scheme
+    kubernetesHttpGetAction.httpHeaders = httpGet.httpHeaders?.collect() {
+      new KeyValuePair(name: it.name, value: it.value)
```
duftler: Using `return` here would be more consistent.

duftler:
LGTM.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
lwander: All my 1-line [closures](https://github.com/spinnaker/clouddriver/pull/729/files/3ef2ee1db06ace437c5b63cdf461763ed9f484b6#diff-64938dfb6417f4870a851790fe3a70a2R460) omit `return`

PR #730, opened 2016-07-01
provider/kubernetes: Looser coupling of labels

lwander:
The idea is to allow custom user-defined RC -> Pod labels to render correctly in spinnaker. This loosens my arbitrary naming-convention requirements to get infrastructure to show up correctly in Deck. For example, `kube-system` components now have their pods rendered correctly: 

kubdns

@duftler PTAL


duftler:
Unit test failed.
LGTM once failing test is addressed.

PR #731, opened 2016-07-01
Titus: enable / disable server group

tomaslin:


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/RegionScopedTitusClient.java
```diff
@@ -172,6 +172,14 @@ public void resizeJob(ResizeJobRequest resizeJobRequest){
     }
 
     @Override
+    public void activateJob(ActivateJobRequest activateJobRequest){
```
tomaslin: endpoint to activate / deactive titus jobs

PR #740, opened 2016-07-07
provider/google: Retry delete SG on timeout

lwander:
We are seeing timeouts during large numbers of MIG deletes. To avoid
inappropriate failures we retry the delete operation once. If this works at
scale we can extend this logic to other operations prone to timeouts.

@duftler, @ttomsu PTAL

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/DestroyGoogleServerGroupAtomicOperation.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/DestroyGoogleServerGroupAtomicOperation.groovy
@@ -16,16 +16,26 @@
 
 package com.netflix.spinnaker.clouddriver.google.deploy.ops
 
+import com.google.api.client.googleapis.json.GoogleJsonResponseException
+import com.google.api.services.compute.Compute
 import com.netflix.spinnaker.clouddriver.data.task.Task
 import com.netflix.spinnaker.clouddriver.data.task.TaskRepository
 import com.netflix.spinnaker.clouddriver.google.deploy.GCEUtil
 import com.netflix.spinnaker.clouddriver.google.deploy.GoogleOperationPoller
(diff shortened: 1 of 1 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/DestroyGoogleServerGroupAtomicOperation.groovy
```diff
@@ -88,34 +136,31 @@ class DestroyGoogleServerGroupAtomicOperation implements AtomicOperation<Void> {
...
+      if (isRegional) {
+        googleOperationPoller.waitForRegionalOperation(compute, project, region, instanceGroupOperationName, null, task,
           "regional instance group $serverGroupName", BASE_PHASE)
-    } else {
-      googleOperationPoller.waitForZonalOperation(compute, project, zone, instanceGroupOperationName, null, task,
+      } else {
+        googleOperationPoller.waitForZonalOperation(compute, project, zone, instanceGroupOperationName, null, task,
           "zonal instance group $serverGroupName", BASE_PHASE)
+      }
+      null
```
ttomsu: Necessary?

ttomsu:
LGTM

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/DestroyGoogleServerGroupAtomicOperation.groovy
lwander: Groovy tries to return the `Operation` result coming from either of the `googleOperationPoller`s otherwise. The original code does this too for similar reasons [here](https://github.com/spinnaker/clouddriver/pull/740/files/d2302fbebddb0ac750cbd111d85d57f9b4d3d3e9#diff-c74180f4c1b005c9f9b2b0453de7fa5dL119).

PR #746, opened 2016-07-08
provider/google: Targeted cluster load

lwander:
@ttomsu PTAL
@duftler FYI

This change prevents loading every cluster from Redis on a single-cluster lookup.


duftler:
LGTM.

ttomsu:
LGTM2

PR #749, opened 2016-07-09
Surface a view of image tags that exist across all AMIs

ajordens:
This PR introduces the ability to both tag images and search for images by 1..* tags.

The current implementation is AWS-specific, but it is an AtomicOperation and open to implementation by other cloud providers.

Relates to https://github.com/spinnaker/gate/pull/255.

There will be `orca` changes forthcoming that provide an UpsertMachineImageTagsStage.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/controllers/AmazonNamedImageLookupController.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/controllers/AmazonNamedImageLookupController.groovy
@@ -30,6 +30,8 @@ import org.springframework.web.bind.annotation.RequestMethod
 import org.springframework.web.bind.annotation.ResponseStatus
 import org.springframework.web.bind.annotation.RestController
 
+import javax.servlet.http.HttpServletRequest
+
 import static com.netflix.spinnaker.clouddriver.core.provider.agent.Namespace.IMAGES
 import static com.netflix.spinnaker.clouddriver.core.provider.agent.Namespace.NAMED_IMAGES
 
@@ -57,18 +59,15 @@ class AmazonNamedImageLookupController {
       throw new ImageNotFoundException("${imageId} not found in ${account}/${region}")
     }
     Collection<String> namedImageKeys = cd.relationships[NAMED_IMAGES.ns]
-    Collection<Map> imageTags = cd.attributes.tags
-
     if (!namedImageKeys) {
       throw new ImageNotFoundException("Name not found on image ${imageId} in ${account}/${region}")
     }
 
-    Collection<CacheData> namedImages = cacheView.getAll(NAMED_IMAGES.ns, namedImageKeys)
-    render(namedImages, null, imageTags, region)
+    render(null, [cd], null, region)
   }
 
   @RequestMapping(value = '/find', method = RequestMethod.GET)
-  List<NamedImage> list(LookupOptions lookupOptions) {
+  List<NamedImage> list(LookupOptions lookupOptions, HttpServletRequest request) {
     if (lookupOptions.q == null || lookupOptions.q.length() < MIN_NAME_FILTER) {
       throw new InsufficientLookupOptionsException(EXCEPTION_REASON)
     }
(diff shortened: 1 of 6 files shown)


ajordens:
Still WIP.

I have a set of related changes that add a new UpsertAmiTags operation ... with the goal of supporting users that wish to tag AMIs and lookup previously tagged AMIs by 1..* key/value tags.

I need to verify that I can send arbitrary tag:key=value parameters from `gate` -> `clouddriver`.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/controllers/AmazonNamedImageLookupController.groovy
```diff
@@ -95,11 +94,51 @@ class AmazonNamedImageLookupController {
...
+  private List<NamedImage> render(Collection<CacheData> namedImages, Collection<CacheData> images, String requestedName = null, String requiredRegion = null) {
     Map<String, NamedImage> byImageName = [:].withDefault { new NamedImage(imageName: it) }
+
+    cacheView.getAll(IMAGES.ns, namedImages.collect {
+      (it.relationships[IMAGES.ns] ?: []).collect {
+        it
+      }
+    }.flatten() as Collection<String>).each {
+      // present tags as a merged set across all AMIs
+      byImageName[it.attributes.name as String].tags.putAll(
```
ajordens: Previously we did not show tags when querying by name ... this change is a conscious decision to not break tags down by regional AMI but rather show the merged set of tags across all AMIs.

Thoughts?

tomaslin: it will prevent ability to do tag-based promotions of ami ( i.e, test -> candidate -> release ) for people that do one region and then all other 2, won't it?

ajordens: Not sure how legitimate that use case is ... in talking with JSJ they (CDE) don't differentiate by region.

That is to say they promote the bits (_aka image by name_) but can still choose to roll it out to one region before another.

From the tagging perspective, I was planning on applying the tags across all AMIs and ideally not requiring an end-user to select regions.

Do you know of other use-cases with a desire to tag/promote right down to a regional AMI? (_this would effectively mean there are different named AMIs tagged `production` for example_)

ajordens: Pushed an alternative implementation that introduces `tagsByImageId` and filters out any regional AMIs that don't pass the tag filter.

It makes differences between regional AMI tags more apparent, probably a good thing even if it's considered unintentional / an edge-case.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/UpsertAmiTagsAtomicOperation.groovy
```diff
@@ -0,0 +1,97 @@
...
+  UpsertAmiTagsAtomicOperation(UpsertAmiTagsDescription description) {
+    this.description = description
+  }
+
+  @Override
+  Void operate(List priorOutputs) {
+    def descriptor = "${description.credentials.name}/${description.amiName}"
+    task.updateStatus BASE_PHASE, "Initializing Upsert ASG Tags operation for ${descriptor}..."
+
+    description.regions.each { String region ->
```
cfieber: should we make region part of the payload, or just tag the image by name in every region in which it exists?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/controllers/AmazonNamedImageLookupController.groovy
```diff
@@ -95,11 +94,65 @@ class AmazonNamedImageLookupController {
...
+      request
+    )
+  }
+
+  /**
+   * Apply tag-based filtering to the list of named images.
+   *
+   * ie. /aws/images/find?q=PackageName&tag:engine=spinnaker&tag:stage=released
+   */
+  private List<NamedImage> filter(List<NamedImage> namedImages, HttpServletRequest httpServletRequest) {
```
cfieber: I'd probably pull the request parameter parsing out of here and pass in the tag filters as an arg to this method

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/UpsertAmiTagsAtomicOperation.groovy
ajordens: My thought was to have orca lookup the image and send down all regions that it exists in (vs having clouddriver figure it out) ... suppose it could go both ways.

cfieber:
we have some code that specifically works around the deduping behaviour from our (non-`rosco`) bakery where an image name can end up with a `1` appended to it in the case of a duplicate bake - because those can end up getting deployed during a bake and deploy pipeline

feels like we should do something to handle that case in here where we are dealing with imageName as well (tag all images that start with the requested name, or explicitly match `name`1)

ajordens:
Minor PR feedback but did not add anything to deal with how the Netflix bakery appends `1` to de-duplicate a bake.

In theory I don't believe this will be a huge concern as the tagging operations will run in the context of a bake and should be able to supply the correct image name (with `1` or not).

Will revisit if this does become an issue.

PR #750, opened 2016-07-11
Added Openstack security group caching.

varikin:
Closes spinnaker/spinnaker#969

### DIFF
--- a/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/model/OpenstackSecurityGroup.groovy
+++ b/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/model/OpenstackSecurityGroup.groovy
@@ -17,12 +17,17 @@
 
 package com.netflix.spinnaker.clouddriver.openstack.model
 
+import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonInclude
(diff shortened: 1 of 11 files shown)


varikin:
@spinnaker/reviewers PTAL

lwander:
LGTM

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/model/OpenstackSecurityGroup.groovy
```diff
@@ -37,8 +42,64 @@ class OpenstackSecurityGroup implements SecurityGroup {
...
+
+  private static AddressableRange buildAddressableRangeFromCidr(String cidr) {
+    if (!cidr) {
+      return null
+    }
+
+    def rangeParts = cidr.split('/') as List
+
+    // If the cidr just a single IP address, use 32 as the mask
+    if (rangeParts.size() == 1) {
```
varikin: I had it elsewhere and moved it here as a static method. I thought I needed it in 2 locations, but I guess I only needed it in one place, but kept it as a static method anyways. I had pulled this logic from the Google security group caching.

lwander: Sounds good

PR #751, opened 2016-07-11
Add support for retrieving the VM user name and password from KeyVaul…

scotmoor:
…t when creating Scale Sets


scotmoor:
This introduces a "breaking change" which will require all users of Azure to setup keyvault. Need to discuss correct way to proceed. 

Closing for now

PR #759, opened 2016-07-13
Azure server group caching bug fixes

milismsft:
When the user deletes the server group through the Azure Portal web site, it causes the Redis entries to be in a corrupted state. One side effect of it is Spinnaker continues to render "ghost" server groups that can become confusing.
Also fixed an issue with how deck is expecting load balancers and security groups associated with a server group to be passed (some core functionality is expecting a list rather than a string).

### DIFF
--- a/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/cache/AzureServerGroupCachingAgent.groovy
+++ b/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/cache/AzureServerGroupCachingAgent.groovy
@@ -85,7 +85,45 @@ class AzureServerGroupCachingAgent extends AzureCachingAgent {
       it.attributes.processedCount = (it.attributes.processedCount ?: 0) + 1
     }
 
-    result
+    if (result.cacheResults[AZURE_SERVER_GROUPS.ns]) {
+      result
+    } else {
+      // run the cache cleanup routine on an empty server group list only for now
+      removeDeadCacheEntries(result, providerCache)
+    }
+  }
+
+  CacheResult removeDeadCacheEntries(CacheResult cacheResult, ProviderCache providerCache) {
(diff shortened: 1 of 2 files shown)


milismsft:
@scotmoor @spinnaker/netflix-reviewers PTAL

milismsft:
ping

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/cache/AzureServerGroupCachingAgent.groovy
```diff
@@ -85,7 +85,45 @@ class AzureServerGroupCachingAgent extends AzureCachingAgent {
       it.attributes.processedCount = (it.attributes.processedCount ?: 0) + 1
     }
 
-    result
+    if (result.cacheResults[AZURE_SERVER_GROUPS.ns]) {
+      result
+    } else {
+      // run the cache cleanup routine on an empty server group list only for now
```
scotmoor: What if the user has deleted a server group from the portal but there are still other server groups that are getting cached? How will this handle "cleaning up" the deleted server group if it only runs on an empty list?

duftler:
@scotmoor @milismsft Is this PR ready to go?

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/cache/AzureServerGroupCachingAgent.groovy
milismsft: The "dead/stale" server group entries are added to the eviction list so if the "cacheResult" collection contains at least one non-empty list we successfully process the eviction list and the dead entries are removed. It's only for the case where the "cacheResult" collection is made of empty lists that we fail to properly evict/delete stale cache entries (this issue might be in the general code - still looking to see what's causing it).

scotmoor:
LGTM

PR #762, opened 2016-07-14
Added default caches for openstack.

varikin:


varikin:
@spinnaker/reviewers PTAL

lwander:
LGTM

PR #764, opened 2016-07-14
provider/openstack: Instance type operation and caching

derekolk:
closes https://github.com/spinnaker/spinnaker/issues/990

### DIFF
--- a/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/cache/Keys.groovy
+++ b/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/cache/Keys.groovy
@@ -33,6 +33,7 @@ class Keys {
     NETWORKS,
     SUBNETS,
     INSTANCES,
+    INSTANCE_TYPES,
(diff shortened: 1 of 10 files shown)


derekolk:
@spinnaker/reviewers PTAL @lwander

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/cache/Keys.groovy
```diff
@@ -154,4 +158,8 @@ class Keys {
   static String getImageKey(String imageId, String account, String region) {
     "${ID}:${Namespace.IMAGES}:${account}:${region}:${imageId}"
   }
+
+  static String getInstanceTypeKey(String instanceType, String account, String region) {
+    "${ID}:${Namespace.INSTANCE_TYPES}:${instanceType}:${account}:${region}"
```
varikin: We seem to be inconsistent with order in these keys. I don't think it is a problem, just noting it.

lwander:
LGTM, nice

PR #765, opened 2016-07-14
provider/consul: Config options

lwander:
I've made a different choice here w.r.t. configuring consul for a specific provider when compared to the eureka support. The idea here is that consul is added/configured once per account per provider, because it's easy to imagine a multi-cloud situation where either 
  1. each cloud is running a separate consul installation
  2. the cloud's consul installations are connected into one consul installation (multi [datacenter](https://www.consul.io/docs/guides/datacenters.html)).

If we make the ability to configure consul a top level option like it is for eureka, we can't have separate consul installations for each clouddriver account. This is an issue for consul, since it's intended to be cloud-agnostic. 

Ultimately, the `clouddriver-consul` package will provide all the utility for caching, reporting health, etc... but the actual cache and health results will need to be surfaced per-provider for this reason. 

PTAL @duftler @ttomsu @jtk54

### DIFF
--- a/clouddriver-consul/src/main/groovy/com/netflix/spinnaker/clouddriver/consul/config/ConsulConfig.groovy
+++ b/clouddriver-consul/src/main/groovy/com/netflix/spinnaker/clouddriver/consul/config/ConsulConfig.groovy
@@ -14,22 +14,12 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.clouddriver.consul.api.v1.model
+package com.netflix.spinnaker.clouddriver.consul.config
 
-class CheckResult {
-  String Node
-  String CheckID
-  String Name
-  Status Status
-  String Notes
-  String Output
-  String ServiceID
-  String ServiceName
-
-  enum Status {
-    passing,
-    critical,
-    warning,
-    unknown,
-  }
+class ConsulConfig {
+  boolean enabled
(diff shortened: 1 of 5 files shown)


clouddriver-consul/src/main/groovy/com/netflix/spinnaker/clouddriver/consul/config/ConsulConfig.groovy
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.consul.config
+
+class ConsulConfig {
+  boolean enabled
+  // required: reachable Consul server endpoints
+  List<String> servers
```
jtk54: If this is a server IP/DNS I would make that more explicit in the naming.

jtk54:
LGTM after comments.

clouddriver-consul/src/main/groovy/com/netflix/spinnaker/clouddriver/consul/config/ConsulConfig.groovy
lwander: Consul docs generally refer to the endpoint as the "server". I updated the comment to make it clearer, but I'd rather stay consistent with the official docs: https://www.consul.io/docs/guides/servers.html

PR #770, opened 2016-07-15
(k8s) prevent container hostPort from defaulting to zero

danielpeach:
(k8s) prevent container hostPort from defaulting to zero (instead of null)
@lwander PTAL


lwander:
good call, LGTM

PR #771, opened 2016-07-15
provider/consul: enable/disable op

lwander:
@duftler @ttomsu @jtk54 PTAL

### DIFF
--- a/clouddriver-consul/src/main/groovy/com/netflix/spinnaker/clouddriver/consul/api/v1/services/AgentApi.groovy
+++ b/clouddriver-consul/src/main/groovy/com/netflix/spinnaker/clouddriver/consul/api/v1/services/AgentApi.groovy
@@ -51,4 +51,7 @@ interface AgentApi {
 
   @PUT("/v1/agent/service/maintenance/{serviceId}")
   Response maintenance(@Path("serviceId") String serviceId, @Query("enable") boolean enable, @Query("reason") String reason)
+
+  @PUT("/v1/agent/maintenance")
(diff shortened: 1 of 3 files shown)


clouddriver-consul/src/main/groovy/com/netflix/spinnaker/clouddriver/consul/api/v1/services/AgentApi.groovy
```diff
@@ -49,6 +49,6 @@ interface AgentApi {
   @PUT("/v1/agent/service/deregister/{serviceId}")
   Response deregisterService(@Path("serviceId") String serviceId)
 
-  @PUT("/v1/agent/service/maintenance/{serviceId}")
-  Response maintenance(@Path("serviceId") String serviceId, @Query("enable") boolean enable, @Query("reason") String reason)
+  @PUT("/v1/agent/maintenance")
```
lwander: The reason I changed this is because there are two `maintenance` endpoints, and the one dealing with specific services isn't useful to us.

jtk54:
LGTM

clouddriver-consul/src/main/groovy/com/netflix/spinnaker/clouddriver/consul/api/v1/services/AgentApi.groovy
lwander: Just kidding - we realized that the per-service enable/disable will be necessary for (de)register instance with load balancer.

PR #775, opened 2016-07-18
provider/kubernetes: Fix loading of LBs

lwander:
The cause was that when load balancers were cached and loaded before the
first server group cache cycle, we would pass `null` server group objects into
the load balancer view. These would then be sent to Deck, causing the unrolling
of server groups to fail.

@duftler


duftler:
LGTM.

PR #779, opened 2016-07-18
provider/consul: Provide a way to get default config

lwander:
@duftler


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.groovy
```diff
@@ -121,7 +121,10 @@ class GoogleNamedAccountCredentials implements AccountCredentials<GoogleCredenti
     }
 
     Builder consulConfig(ConsulConfig consulConfig) {
-      this.consulConfig = consulConfig
+      if (consulConfig && consulConfig.enabled) {
```
duftler: `if (consulConfig?.enabled) {`

duftler:
LGTM.

PR #783, opened 2016-07-18
Provide subnet instead of network to server group

drmaas:
Network is looked up via subnet and both are provided to the heat template.

closes spinnaker/spinnaker#994


drmaas:
@spinnaker/reviewers ptal

lwander:
LGTM

PR #786, opened 2016-07-19
Docker registry

tgracchus:
Equipare the docker cli behaviour and the clouddriver behavior 

When login to a docker registry, the official behaviour is defined here
https://docs.docker.com/registry/spec/auth/token/

1. So, the first think it is trying to do is a connection to /v2/ endpoint
2. The server respond with a 401 with the auth details in the response
3. Then, the client try to auth.

In this first request is very docker cli and clouddriver differs.

Docker:
Docker client sends this first request without Authorization Header, see line 65 of attached log.
[nginx.txt](https://github.com/spinnaker/clouddriver/files/370858/nginx.txt)

CloudDriver:
Sends the query with the Authorization Header

This slightly difference makes our clouddriver to be down, cause the Docker Health Check always fails.

So, this PR pretends to make clouddriver behave in the same way as the docker client

### DIFF
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
@@ -22,6 +22,8 @@ import com.netflix.spinnaker.clouddriver.docker.registry.api.v2.auth.DockerBeare
 import com.netflix.spinnaker.clouddriver.docker.registry.api.v2.exception.DockerRegistryOperationException
 import com.squareup.okhttp.OkHttpClient
 import groovy.util.logging.Slf4j
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
 import retrofit.RestAdapter
 import retrofit.RetrofitError
@@ -34,6 +36,9 @@ import java.util.concurrent.TimeUnit
 
 @Slf4j
 class DockerRegistryClient {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DockerRegistryClient)
+
   private DockerBearerTokenService tokenService
 
   public String address
@@ -198,15 +203,27 @@ class DockerRegistryClient {
    * This method will hit the /v2/ endpoint of the configured docker registry. If it this endpoint is up,
    * it will return silently. Otherwise, an exception is thrown detailing why the endpoint isn't available.
    */
+
   public void checkV2Availability() {
+    try {
+      doCheckV2Availability()
+    } catch (RetrofitError error) {
+      Response response = doCheckV2Availability(tokenService.basicAuthHeader)
+      if (!response){
+        LOG.error "checkV2Availability", error
+        throw error
+      }
+    }
+    // Placate the linter (otherwise it expects to return the result of `request()`)
+    null
+  }
+
+  private Response doCheckV2Availability(String basicAuthHeader = null) {
     request({
-      registryService.checkVersion(tokenService.basicAuthHeader, dockerApplicationName)
+      registryService.checkVersion(basicAuthHeader, dockerApplicationName)
     }, { token ->
       registryService.checkVersion(token, dockerApplicationName)
     }, "v2 version check")
-
-    // Placate the linter (otherwise it expects to return the result of `request()`)
-    null
   }
 
   /*


duftler:
@lwander

lwander:
Retrofit ignores `null` parameters, so if you're not setting auth details those two requests should be identical.

tgracchus:
@lwander

Agreed with that.

But in line
https://github.com/spinnaker/clouddriver/blob/master/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerTokenService.groovy#L175
The basicAuth is used again. It's true it works with docker repository since it gives you a token regardless if you are auth or not.

But, in the case the basicAuth is needed, because it's a corporate repository. Then it will be null and the query will fail.

In our case, this corporate repository is Artifactory. 
Does it make sense?

lwander:
I see what you mean, but I don't think this fix is correct: 

I don't think this modified flow will work when someone has basic auth enabled without a token server. The first request will fail, due to a lack of credentials, and the fact that there is no token server: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerTokenService.groovy#L162. The retry will then fail as well here: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy#L232.

tgracchus:
@lwander 
Please, can you take a look at the new proposed solution?

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -198,15 +201,27 @@ class DockerRegistryClient {
...
+        Response response = doRequest(null)
+        if(response != null){
+          throw error
+        }
+    }
+    // Placate the linter (otherwise it expects to return the result of `request()`)
+    null
+  }
+
+  private Response doRequest(String basicAuthHeader) {
```
lwander: Why is the v2 version check the only request using this function? I'd rename the function or apply it to other requests

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -198,15 +201,27 @@ class DockerRegistryClient {
...
    * it will return silently. Otherwise, an exception is thrown detailing why the endpoint isn't available.
    */
+
   public void checkV2Availability() {
+    try {
+      doRequest(tokenService.basicAuthHeader)
+    } catch (RetrofitError error) {
+        LOG.warn "checkV2Availability", error
+        Response response = doRequest(null)
+        if(response != null){
```
lwander: Missing space before `(response...`

lwander:
It doesn't look like this will break existing workflows, but this still doesn't match the way docker does this request like you posted above (this still sends the auth header first).

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
lwander: `if (response)` is the groovy way to do this.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
tgracchus: Just Renamed.

tgracchus:
It follows the docker workflow now, ie:
1. Request without auth credentials
2. If 401 look for token and query with token
3. If error again, retry 1 and 2 without auth credentials

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
tgracchus: changed!!

tgracchus:
@lwander please, could you take a look at it?

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -198,15 +203,27 @@ class DockerRegistryClient {
    * This method will hit the /v2/ endpoint of the configured docker registry. If it this endpoint is up,
    * it will return silently. Otherwise, an exception is thrown detailing why the endpoint isn't available.
    */
+
   public void checkV2Availability() {
+    try {
+      doCheckV2Availability()
+    } catch (RetrofitError error) {
+      LOG.warn "checkV2Availability", error
```
lwander: This will lead to a lot of wasted log space; we shouldn't `warn` everytime the registry requires authentication (I expect most registries will).

lwander:
LGTM after comment, please rebase as well

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
tgracchus: changed to error

tgracchus:
rebased

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
lwander: I don't think error is appropriate either, when this is the expected registry auth flow, right? An error message should be provided when the auth flow fails, not during it I think.

tgracchus: ok, so, is it ok if a move log it if no response is available and the second try with auth do not throw an expcetion?

lwander: The way you have it now looks good, thanks!

tgracchus: I guess it would be logged twice, one by the DockerRegistry and another time by the caller. It is still ok?

PR #787, opened 2016-07-19
provider/consul: cleanup API usage

lwander:
This allows us to instantiate each consul API with the `config` object stored in the provider credentials.

### DIFF
--- a/clouddriver-consul/src/main/groovy/com/netflix/spinnaker/clouddriver/consul/api/v1/ConsulCatalog.groovy
+++ b/clouddriver-consul/src/main/groovy/com/netflix/spinnaker/clouddriver/consul/api/v1/ConsulCatalog.groovy
@@ -17,23 +17,16 @@
 package com.netflix.spinnaker.clouddriver.consul.api.v1
 
 import com.netflix.spinnaker.clouddriver.consul.api.v1.services.CatalogApi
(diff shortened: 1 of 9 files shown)


lwander:
@duftler PTAL

clouddriver-consul/src/main/groovy/com/netflix/spinnaker/clouddriver/consul/api/v1/ConsulCatalog.groovy
```diff
@@ -26,11 +28,14 @@ import java.util.concurrent.TimeUnit
 class ConsulCatalog {
   public CatalogApi api
 
-  ConsulCatalog(String serverBaseUrl, Long timeout) {
+  ConsulCatalog(ConsulConfig config) {
+    if (!config.enabled) {
```
duftler: The bodies of the constructor in this class and the next are identical except for the arg to `.create()`. Can the common code be extracted somewhere? Maybe they should share a common base class and then put the common code in the parent class's constructor?

lwander: Good idea

lwander:
@duftler updated with your suggestion

duftler:
Nice. LGTM.

PR #788, opened 2016-07-19
provider/google: added L7 LB feature flag in configuration.

jtk54:
Added L7 feature flag in configuration. @duftler or @lwander or @ttomsu PTAL.

The yaml configuration to enable L7 LB (in clouddriver-local.yml) looks like:
```yml
google:
  enabled: true
  - name: my-account
    project: my-gcp-project
    jsonPath: /path/to/service/account/json/key.json
    httpLoadBalancingEnabled: true
```

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/config/GoogleConfigurationProperties.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/config/GoogleConfigurationProperties.groovy
@@ -30,6 +30,7 @@ class GoogleConfigurationProperties {
     String accountType
     String project
     boolean alphaListed
+    boolean httpLoadBalancingEnabled // TODO(jacobkiefer): Feature flag for L7 development.
     String jsonPath
     List<String> imageProjects
     List<String> requiredGroupMembership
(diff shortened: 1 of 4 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/config/GoogleConfigurationProperties.groovy
```diff
@@ -30,6 +30,7 @@ class GoogleConfigurationProperties {
     String accountType
     String project
     boolean alphaListed
+    boolean httpLoadBalancingEnabled // TODO(jacobkiefer): Feature flag for L7 development.
```
ttomsu: Seems to me like you'd want a global flag to do this, not just per account. 

I just put it at the top level of the `google` yaml stack: https://github.com/spinnaker/clouddriver/commit/17bc7b24a605a4c7331313c0f61d5d07d5d9449f#diff-f3efbf7dc0f934c8fb5a295b1bf8e629R56

jtk54: I would like a sanity check to make sure the L3 behavior is the same with and without L7 alongside.

duftler:
LGTM.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/config/GoogleConfigurationProperties.groovy
duftler: I agree that being able to compare behavior with/without this enabled will likely be useful during development.

PR #791, opened 2016-07-20
Documentation around Eureka health provider

daltonconley:
Adding a simple README to describe how to configure the eureka health provider in clouddriver.


cfieber:
thanks! mind squashing those commits and then LGTM

PR #792, opened 2016-07-20
provider/google: removed line causing error in serialize atomic operation

nwwebb:
Removed a line that was causing an error when trying to serialize server groups. PTAL @duftler  @lwander


lwander:
LGTM

duftler:
LGTMA

PR #796, opened 2016-07-21
(gce) fixed validation for autoscaler custom metrics

danielpeach:
(gce) fixed validation for autoscaler custom metrics
Utilization target for custom metrics just needs to be greater than zero.
@duftler PTAL


duftler:
LGTM.

PR #800, opened 2016-07-25
(gce) renamed delete autoscaler classes

danielpeach:
@duftler please review. I verified that the operation still works as expected.


duftler:
LGTM.

PR #803, opened 2016-07-26
quick stab at request times for titus API calls

cfieber:
@tomaslin PTAL

not super happy with putting the requestName strings on all those calls but the alternatives seemed worse (mining the stack trace) or maybe not possible ( inspecting the Retrofit annotations when creating the call object)


cfieber:
@tomaslin added  titusAccount and titusRegion tags

tomaslin:
LGTM

PR #805, opened 2016-07-26
Add network attributes to load balancer requests

drmaas:
Also add in subnet id.


drmaas:
@spinnaker/reviewers @derekolk @varikin enjoy

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/model/OpenstackLoadBalancer.groovy
```diff
@@ -30,30 +30,35 @@ import org.openstack4j.model.network.ext.LbPool
...
+  String method
+  String ip
+  Integer externalPort
+  String subnetId
+  String subnetName
+  String networkId
+  String networkName
+  Set<PoolHealthMonitor> healthChecks
+  Set<LoadBalancerServerGroup> serverGroups = Sets.newConcurrentHashSet()
+
```
derekolk: I have been using builders instead of static methods ... Feels like it gives more flexibility for each use case.

drmaas: There's only 1 use case - adding a builder feels like extra framework code.

lwander:
LGTM

PR #823, opened 2016-07-28
provider/google: cleaned up unnecessary enum conversion.

jtk54:
@duftler FYI.


duftler:
LGTM.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/loadbalancing/GoogleLoadBalancedBackend.groovy
```diff
@@ -39,17 +39,6 @@ class GoogleLoadBalancedBackend {
    */
   String serverGroupUrl
 
-  static BalancingMode toBalancingMode(String mode) {
```
duftler: Can't you now drop the constant `Strings` up above and do comparisons directly against the `enum` values?

jtk54: Yep, I'll follow up with a cleanup PR once I do the 'on fire' bugs.

PR #826, opened 2016-07-28
provider/google: fixed default service population in L7.

jtk54:
Also made 'queryAllLoadBalancers' return all cached LBs, not just in application scope. @duftler PTAL.


duftler:
LGTM.

PR #827, opened 2016-07-28
provider/google: cross application LBs returned from LB provider

jtk54:
if associated with a server group from our application. @duftler PTAL.


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/view/GoogleLoadBalancerProvider.groovy
```diff
@@ -49,8 +49,16 @@ class GoogleLoadBalancerProvider implements LoadBalancerProvider<GoogleLoadBalan
     def pattern = Keys.getLoadBalancerKey("*", "*", "${application}*")
     def identifiers = cacheView.filterIdentifiers(LOAD_BALANCERS.ns, pattern)
 
+    def applicationServerGroups = cacheView.getAll(
+        SERVER_GROUPS.ns,
+        cacheView.filterIdentifiers(SERVER_GROUPS.ns, "gce:*:${application}-*")
```
duftler: Can use `GoogleCloudProvider.GCE` instead of `"gce"` here.

duftler:

PR #830, opened 2016-07-29
provider/google: L7 support in server group create.

jtk54:
@duftler PTAL.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
@@ -30,6 +30,7 @@ import com.netflix.spinnaker.clouddriver.google.deploy.GoogleOperationPoller
 import com.netflix.spinnaker.clouddriver.google.deploy.description.BasicGoogleDeployDescription
 import com.netflix.spinnaker.clouddriver.google.model.GoogleServerGroup
 import com.netflix.spinnaker.clouddriver.google.model.callbacks.Utils
+import com.netflix.spinnaker.clouddriver.google.model.loadbalancing.GoogleHttpLoadBalancingPolicy
 import com.netflix.spinnaker.clouddriver.google.model.loadbalancing.GoogleLoadBalancerType
 import com.netflix.spinnaker.clouddriver.google.provider.view.GoogleLoadBalancerProvider
 import com.netflix.spinnaker.clouddriver.google.provider.view.GoogleSecurityGroupProvider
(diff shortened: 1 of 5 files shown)


jtk54:
This also adds support for setting `balancingMode` and `maxUtilization` or `maxRatePerInstance` upon creation. @danielpeach will add support in the frontend shortly.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/loadbalancing/GoogleHttpLoadBalancingPolicy.groovy
```diff
@@ -0,0 +1,37 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.clouddriver.google.model.loadbalancing
+
+class GoogleHttpLoadBalancingPolicy {
+  static final String UTILIZATION = "UTILIZATION"
```
duftler: Let's drop these `String` constants in favor of working directly with the `enum` values.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -265,11 +266,29 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
         Backend sourceBackend = backendService.backends.find { Backend backend ->
           Utils.getLocalName(backend.group) == description.source.serverGroupName
         }
-        if (!sourceBackend) {
+        if (description.source.serverGroupName && !sourceBackend) {
```
duftler: Even though `source` is assigned a value at initialization of the containing `description`, someone could still make it explicitly null in the request. So let's use `?.` between `description` and `source`. Here, and the couple of other spots where `source` is dereferenced.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -265,11 +266,29 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
...
           GCEUtil.updateStatusAndThrowNotFoundException("Backend for ancestor server group $description.source.serverGroupName not found.", task, BASE_PHASE)
         }
 
-        def backendToAdd = new Backend(sourceBackend)
+        Backend backendToAdd
+        def loadBalancingPolicy = description.loadBalancingPolicy
+        if (sourceBackend) {
+          backendToAdd = new Backend(sourceBackend)
+        } else {
```
duftler: Can't this `else` and the child `if` be combined into an `else if`?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/loadbalancing/GoogleHttpLoadBalancingPolicy.groovy
jtk54: Comparison directly between the `enum` and `String` values won't work, but I'll use `valueOf` and get rid of the constants.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleHttpLoadBalancerCachingAgent.groovy
```diff
@@ -376,11 +376,13 @@ class GoogleHttpLoadBalancerCachingAgent extends AbstractGoogleCachingAgent impl
...
           new GoogleLoadBalancedBackend(
               serverGroupUrl: backend.group,
-              balancingMode: backend.balancingMode,
-              maxRatePerInstance: backend.balancingMode == GoogleLoadBalancedBackend.RATE ?
-                  backend.maxRatePerInstance : null,
-              maxUtilization: backend.balancingMode == GoogleLoadBalancedBackend.UTILIZATION ?
-                  backend.maxUtilization : null,
+              policy: new GoogleHttpLoadBalancingPolicy(
+                  balancingMode: backend.balancingMode,
+                  maxRatePerInstance: GoogleHttpLoadBalancingPolicy.BalancingMode.valueOf(backend.balancingMode) == GoogleHttpLoadBalancingPolicy.BalancingMode.RATE ?
```
duftler: Why not capture the balancing mode `enum` once and then use it in both comparisons? Would make the lines quite a bit shorter too.

duftler:
LGTM after addressing final comment.

PR #832, opened 2016-07-29
provider/google: Use GoogleNamedAccountCredentials to fetch serial co…

duftler:
…nsole output instead of GoogleCredentials.


duftler:
fyi @ttomsu

PR #833, opened 2016-07-29
Fixed race condition on getting openstack client token.

varikin:


varikin:
@spinnaker/reviewers PTAL

varikin:
@spinnaker/reviewers PTAL?

PR #835, opened 2016-07-29
provider/google: set named ports on clone/create.

jtk54:
so the server groups will actually receive traffic. @duftler PTAL.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
@@ -32,12 +32,15 @@ import com.netflix.spinnaker.clouddriver.google.model.GoogleServerGroup
 import com.netflix.spinnaker.clouddriver.google.model.callbacks.Utils
 import com.netflix.spinnaker.clouddriver.google.model.loadbalancing.GoogleHttpLoadBalancingPolicy
 import com.netflix.spinnaker.clouddriver.google.model.loadbalancing.GoogleLoadBalancerType
+import com.netflix.spinnaker.clouddriver.google.provider.view.GoogleClusterProvider
 import com.netflix.spinnaker.clouddriver.google.provider.view.GoogleLoadBalancerProvider
 import com.netflix.spinnaker.clouddriver.google.provider.view.GoogleSecurityGroupProvider
+import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 
 @Component
+@Slf4j
 class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescription> {
 
   // TODO(duftler): This should move to a common location.
@@ -50,6 +53,9 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
 
   @Autowired
   private GoogleConfigurationProperties googleConfigurationProperties
+
+  @Autowired
+  private GoogleClusterProvider googleClusterProvider
 
   @Autowired
   private GoogleConfiguration.DeployDefaults googleDeployDefaults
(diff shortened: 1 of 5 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -204,16 +204,46 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
...
+        .setBaseInstanceName(serverGroupName)
+        .setInstanceTemplate(instanceTemplateUrl)
+        .setTargetSize(description.targetSize)
+        .setTargetPools(targetPools)
+        .setAutoHealingPolicies(autoHealingPolicy)
+
+    def hasBackendServices = description.instanceMetadata &&
+        description.instanceMetadata.containsKey(GoogleServerGroup.View.BACKEND_SERVICE_NAMES)
+
+    if (hasBackendServices && (description?.loadBalancingPolicy || description.source.serverGroupName))  {
```
duftler: Null-safe navigation on `description.source.serverGroupName`.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -204,16 +204,46 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
...
+      NamedPort namedPort = null
+      def sourceManagerName = description.source.serverGroupName
+      if (sourceManagerName) {
+        def sourceInstanceGroupManager
+        if (isRegional) {
+          sourceInstanceGroupManager = compute.regionInstanceGroupManagers().get(project, description.source.region, sourceManagerName).execute()
```
duftler: I think a better approach might be to cache the named ports on server groups in the caching agents, and then retrieve the source server group from the cache here instead of making live calls.

duftler: This approach should also let you kill off your `TODO` below regarding the zone caveat.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -204,16 +204,46 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
...
+        } else {
+          // TODO(jacobkiefer): This will work only if the new zone is the same as the old zone. This will only be a problem for ad-hoc.
+          sourceInstanceGroupManager = compute.instanceGroupManagers().get(project, description.zone, sourceManagerName).execute()
+        }
+        if (!sourceInstanceGroupManager) {
+          GCEUtil.updateStatusAndThrowNotFoundException("Could not locate source group manager ${sourceManagerName} to update named port.", task, BASE_PHASE)
+        }
+        namedPort = sourceInstanceGroupManager.namedPorts.find { it.name == GoogleHttpLoadBalancingPolicy.HTTP_PORT_NAME }
+      } else {
```
duftler: Won't this fail if the source server group exists but does not define the named port? Don't we want to create the named port on the new server group even if the source server group didn't have it defined?

jtk54: Yep, I'll fix this so we can mix the specifications: i.e. clone but specify load balancing policy.

jtk54:
Replaced live calls with cache calls and just warn and create a default if we can't resolve a named port.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -204,16 +210,48 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
...
+      NamedPort namedPort = null
+      def sourceGroupName = description?.source?.serverGroupName
+      // Note: this favors the explicitly specified load balancing policy over the source server group.
+      if (sourceGroupName && !description?.loadBalancingPolicy) {
+        def sourceServerGroup = googleClusterProvider.getServerGroup(description.accountName, description.source.region, sourceGroupName)
+        if (!sourceServerGroup) {
+          log.warn("Could not locate source server group ${sourceGroupName} to update named port.")
+        }
+        namedPort = new NamedPort(
+            name: GoogleHttpLoadBalancingPolicy.HTTP_PORT_NAME,
```
duftler: Can't the creator of the load balancer choose a port other than 'http'? If so, don't we need to get the name from the load balancer's backend service?

jtk54: Each backend service only knows about one named port, so unless you're running a monolith with everything on one machine and routing to different ports on the same machine from multiple backend services, I think it makes sense to stick with the default name. If needed, we can add that in v2.

duftler: Sgtm

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -204,16 +210,48 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
...
+            port: sourceServerGroup?.namedPorts[(GoogleHttpLoadBalancingPolicy.HTTP_PORT_NAME)] ?: GoogleHttpLoadBalancingPolicy.HTTP_DEFAULT_PORT,
+        )
+      } else {
+        namedPort = new NamedPort(
+            name: GoogleHttpLoadBalancingPolicy.HTTP_PORT_NAME,
+            port: description?.loadBalancingPolicy?.listeningPort ?: GoogleHttpLoadBalancingPolicy.HTTP_DEFAULT_PORT
+        )
+      }
+      if (!namedPort) {
+        log.warn("Could not locate named port on either load balancing policy or source server group. Setting default named port.")
```
duftler: Nice.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -204,16 +210,48 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
...
+        .setBaseInstanceName(serverGroupName)
+        .setInstanceTemplate(instanceTemplateUrl)
+        .setTargetSize(description.targetSize)
+        .setTargetPools(targetPools)
+        .setAutoHealingPolicies(autoHealingPolicy)
+
+    def hasBackendServices = description.instanceMetadata &&
+        description.instanceMetadata.containsKey(GoogleServerGroup.View.BACKEND_SERVICE_NAMES)
+
+    if (hasBackendServices && (description?.loadBalancingPolicy || description?.source?.serverGroupName))  {
```
duftler: Just realized I led you astray with the `description?.source`. `description` itself can't be `null`, so it's actually just `source` that needs to be followed by `?.`. And description doesn't need it anywhere here.

duftler: Feel free to clean that up in a follow-on PR or when you're back in the neighborhood.

duftler:
LGTM.

PR #838, opened 2016-08-01
Added L7 loadBalancingPolicy to server group view for clone.

jtk54:
@duftler @danielpeach PTAL.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleServerGroup.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleServerGroup.groovy
@@ -23,7 +23,9 @@ import com.google.api.services.compute.model.AutoscalingPolicy
 import com.google.api.services.compute.model.InstanceGroupManagerActionsSummary
 import com.google.api.services.compute.model.InstanceGroupManagerAutoHealingPolicy
 import com.netflix.spinnaker.clouddriver.google.GoogleCloudProvider
-import com.netflix.spinnaker.clouddriver.google.model.loadbalancing.GoogleLoadBalancerView
+import com.netflix.spinnaker.clouddriver.google.deploy.GCEUtil
+import com.netflix.spinnaker.clouddriver.google.model.callbacks.Utils
+import com.netflix.spinnaker.clouddriver.google.model.loadbalancing.*
 import com.netflix.spinnaker.clouddriver.model.HealthState
 import com.netflix.spinnaker.clouddriver.model.Instance
 import com.netflix.spinnaker.clouddriver.model.ServerGroup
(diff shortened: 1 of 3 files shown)


jtk54:
Some context: in the clone wizard, we want to populate the L7 "load balancing policy" section with the settings of the previous server group. This PR provides that information so it can be copied.

jtk54:
Update: the load balancing policy is the same for each backend service a server group joins, so we need to report only the policy itself. I'll refactor and ping when done.

jtk54:
Ping -- this is cleaned up and ready to review @duftler @danielpeach.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleServerGroup.groovy
```diff
@@ -113,6 +115,37 @@ class GoogleServerGroup {
...
+      def loadBalancers = GoogleServerGroup.this.loadBalancers
+      loadBalancers.each { GoogleLoadBalancerView view ->
+        if (GoogleLoadBalancerType.valueOf(view.loadBalancerType) == GoogleLoadBalancerType.HTTP) {
+          GoogleHttpLoadBalancer.View httpView = view as GoogleHttpLoadBalancer.View
+          def backendServices = Utils.getBackendServicesFromHttpLoadBalancerView(httpView)
+          backendServices.each { GoogleBackendService backendService ->
+            def foundBackends = backendService.backends.findAll { GoogleLoadBalancedBackend backend ->
+              GCEUtil.getLocalName(backend.serverGroupUrl) == GoogleServerGroup.this.name
+            }
+            // Note: we configure the same load balancing policy for all backends on a create/clone,
```
lwander: What if a server group created outside of spinnaker is being cloned?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleServerGroup.groovy
```diff
@@ -113,6 +115,37 @@ class GoogleServerGroup {
           null
     }
 
+    /**
+     * @return The load balancing policy containing the
```
lwander: Strange newline

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleServerGroup.groovy
```diff
@@ -113,6 +115,37 @@ class GoogleServerGroup {
...
+            }
+            // Note: we configure the same load balancing policy for all backends on a create/clone,
+            // so we can just return the policy from the first one we find.
+            if (foundBackends) {
+              def foundBackend = foundBackends?.first()
+              loadBalancingPolicy = foundBackend?.policy
+              if (loadBalancingPolicy) {
+                loadBalancingPolicy.listeningPort = GoogleServerGroup.this.namedPorts.get(GoogleHttpLoadBalancingPolicy.HTTP_PORT_NAME)
+              }
+              return loadBalancingPolicy
```
lwander: This `return` is in a closure, and will exit the `backendServices.each` loop rather than `getLoadBalancingPolicy()`.

jtk54: Short circuiting is still the desired behavior -- we'll still return the correct information with the last `return`.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleServerGroup.groovy
jtk54: _Stranger Newlines_: A Netflix Original

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleServerGroup.groovy
jtk54: This is intended to to be used as the suggestion in the server group wizard for load balancing policy. If a server group outside is created and added to L7 outside of Spinnaker, this will return the first load balancing policy we find. _Maybe_ we could count the backends and report the most common configuration, but it seems like overkill at this point, since you can configure this anyway.

lwander: Ah I see, makes sense.

lwander: I guess the method name and comment should reflect that this is just one of many applicable load balancing policies, since that's not clear without this above comment.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleServerGroup.groovy
lwander: Hmm the outer loop keeps looping though, and `return`ing an ignored value is a bit odd.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleServerGroup.groovy
```diff
@@ -113,6 +115,37 @@ class GoogleServerGroup {
           null
     }
 
+    /**
+     * @return The load balancing policy containing the * capacity metrics and named ports this server
```
lwander: `*`

jtk54:
Making a follow-up PR to clean this up a bit.

PR #854, opened 2016-08-05
Adding back the ability to override environment and accounttype.

zanthrash:
Putting back because the wrong NETFILX_ENVIRONMENT is being written to userdata for the SEG account.

NETFILX_ENVIRONMENT is getting set to 'seg' and should be set to 'prod'

@robfletcher you put this in and rolled it back b/c @anotherchrisberry said that it has some OSS implications.  Please review and add concerns.


zanthrash:
Closing: Duplicate of https://github.com/spinnaker/clouddriver/pull/713

PR #855, opened 2016-08-05
Kubernetes : Return image list back in buildInfo

bfarrell:
First PR for [#1047](https://github.com/spinnaker/spinnaker/issues/1047)

Sends the image information back so that it can be displayed on cluster and project page.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesClusterProvider.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesClusterProvider.groovy
@@ -23,17 +23,17 @@ import com.netflix.spinnaker.cats.cache.CacheData
 import com.netflix.spinnaker.cats.cache.CacheFilter
 import com.netflix.spinnaker.cats.cache.RelationshipCacheFilter
 import com.netflix.spinnaker.clouddriver.kubernetes.cache.Keys
+import com.netflix.spinnaker.clouddriver.kubernetes.deploy.KubernetesUtil
 import com.netflix.spinnaker.clouddriver.kubernetes.model.KubernetesCluster
-import com.netflix.spinnaker.clouddriver.kubernetes.model.KubernetesInstance
 import com.netflix.spinnaker.clouddriver.kubernetes.model.KubernetesLoadBalancer
-import com.netflix.spinnaker.clouddriver.kubernetes.model.KubernetesSecurityGroup
 import com.netflix.spinnaker.clouddriver.kubernetes.model.KubernetesServerGroup
 import com.netflix.spinnaker.clouddriver.model.ClusterProvider
 import com.netflix.spinnaker.clouddriver.model.ServerGroup
-import io.fabric8.kubernetes.api.model.Pod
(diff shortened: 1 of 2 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesClusterProvider.groovy
```diff
@@ -148,6 +147,15 @@ class KubernetesClusterProvider implements ClusterProvider<KubernetesCluster> {
       serverGroup.loadBalancers?.each {
         serverGroup.securityGroups.addAll(securityGroups[it])
       }
+
+      def imageList = new ArrayList()
+      for (def container : serverGroup.deployDescription.containers) {
+        imageList.add(container.imageDescription.repository + ":" + container.imageDescription.tag)
```
lwander: Could you use this method instead for consistency? https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/KubernetesUtil.groovy#L103

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesClusterProvider.groovy
```diff
@@ -148,6 +147,15 @@ class KubernetesClusterProvider implements ClusterProvider<KubernetesCluster> {
...
+      }
+      if (!imageList.isEmpty()) {
+        Map buildInfo = [images: imageList]
```
lwander: Why not just attach an empty image list? Also, an empty image list is cause for concern since an RC shouldn't be deployable without containers.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesClusterProvider.groovy
```diff
@@ -148,6 +147,15 @@ class KubernetesClusterProvider implements ClusterProvider<KubernetesCluster> {
       serverGroup.loadBalancers?.each {
         serverGroup.securityGroups.addAll(securityGroups[it])
       }
+
+      def imageList = new ArrayList()
```
lwander: The groovy way to do this is `def imageList = []`.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesClusterProvider.groovy
bfarrell: I had done it so I could check on buildInfo.images in the UI but you are right, that if its empty then we likely have a problem.  Ill remove the isEmpty check.

bfarrell:
working on rebase

lwander:
LGTM, ready for a merge?

lwander:
(git will squash this for us, no need to do it by hand)

bfarrell:
Ah great, then gtg.

PR #860, opened 2016-08-08
provider/google: Don't try to resolve target pool if forwarding rule …

duftler:
…targets a vpn gateway.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleLoadBalancerCachingAgent.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleLoadBalancerCachingAgent.groovy
@@ -193,12 +193,16 @@ class GoogleLoadBalancerCachingAgent extends AbstractGoogleCachingAgent implemen
         loadBalancers << newLoadBalancer
 
         if (forwardingRule.target) {
-          def targetPoolName = Utils.getLocalName(forwardingRule.target)
-          def targetPoolsCallback = new TargetPoolCallback(googleLoadBalancer: newLoadBalancer,
-                                                           httpHealthChecksRequest: httpHealthChecksRequest,
-                                                           instanceHealthRequest: instanceHealthRequest)
+          def forwardingRuleTokens = forwardingRule.target.split("/")
 
-          compute.targetPools().get(project, region, targetPoolName).queue(targetPoolsRequest, targetPoolsCallback)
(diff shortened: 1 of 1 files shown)


duftler:
@lwander please review.

duftler:
Or @jtk54

danielpeach:
LGTM

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleLoadBalancerCachingAgent.groovy
```diff
@@ -193,12 +193,16 @@ class GoogleLoadBalancerCachingAgent extends AbstractGoogleCachingAgent implemen
...
 
         if (forwardingRule.target) {
-          def targetPoolName = Utils.getLocalName(forwardingRule.target)
-          def targetPoolsCallback = new TargetPoolCallback(googleLoadBalancer: newLoadBalancer,
-                                                           httpHealthChecksRequest: httpHealthChecksRequest,
-                                                           instanceHealthRequest: instanceHealthRequest)
+          def forwardingRuleTokens = forwardingRule.target.split("/")
 
-          compute.targetPools().get(project, region, targetPoolName).queue(targetPoolsRequest, targetPoolsCallback)
+          if (forwardingRuleTokens[forwardingRuleTokens.size() - 2] != "targetVpnGateways") {
```
lwander: Doesn't this make you a little nervous?

duftler: The url pattern shouldn't change since it is observed by all of our platform resources.
And the `[]` lookup uses `getAt()` under the covers. That method won't throw `ArrayIndexOutOfBoundsExceptions`. It will return `null` if you're off of the end of the collection, and the final token if you're into the negatives.
So feels ok to me.

duftler: Worth pointing out here that I'd rather we had a more formalized way to determine the type of the thing being targeted, but I'm unable to find one. So we're stuck with string-work.

lwander: Sounds good, I didn't know `[]` used `getAt()`

PR #864, opened 2016-08-08
need to ensure we get instances that have never been classiclinked when reconciling classic link groups

cfieber:
starts with a describeInstances call to build the complete set of EC2 classic instances


ajordens:
LGTM

robfletcher:
LGTM

PR #870, opened 2016-08-09
provider/google new autoscaling policy fix

danielpeach:
@duftler please review

The behavior for updating an autoscaling policy and creating a new autoscaling policy is currently inconsistent when a field in the description is an empty object.


duftler:
LGTM.

PR #874, opened 2016-08-10
provider/google: Drop .with construct since it was causing on-demand …

duftler:
…update to fail.


duftler:
@jtk54 TBR. Pushing this through as it is blocking some operations.

jtk54:
LGTM

PR #876, opened 2016-08-10
provider/google: Added retry function, applied it to L7 updates.

jtk54:
@lwander and/or @duftler please review. This takes care of both the 400 and 412 retries, and covers the 404 acceptance in destroy.


lwander:
Nice work, LGTM

PR #880, opened 2016-08-11
provider/google: Log platform error if retry fails repeatedly.

jtk54:
@lwander please review @duftler FYI.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
@@ -838,7 +838,7 @@ class GCEUtil {
             log.warn "Retry $action of $resource encountered ${jsonException.statusCode}, treating as success..."
             return
           } else if (jsonException.statusCode in retryCodes) {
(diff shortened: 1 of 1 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -842,11 +844,13 @@ class GCEUtil {
...
           }
-        } catch (GoogleOperationTimedOutException __) {
+        } catch (GoogleOperationTimedOutException timeout) {
+          t = timeout
           log.warn "Retry $action timed out again, trying again..."
         }
       }
-      throw new GoogleOperationException("Failed to delete $resource after #$tries")
+      log.warn("Failed to $action $resource after #$tries.")
+      throw t
```
lwander: It's a little strange that this exception only reflects the last failure of possibly many

jtk54: There's at least a log trace of the error codes. I could also log the messages, but that would get quite noisy.

lwander:
LGTM

PR #883, opened 2016-08-11
`createLaunchConfiguration` should retry on failure

ajordens:
- Protects against "Invalid IamInstanceProfile" errors when the IAM was recently created


ajordens:
PTAL @cfieber 

Should have probably done this in the first place.

cfieber:
LGTM

PR #886, opened 2016-08-12
provider/google: Server group caching agents should be authoritative …

duftler:
…w.r.t. applications.


duftler:
@lwander please review.

lwander:
I don't think this is the right fix, since both of those caching agents will encounter the same set of applications. It's probably best to keep the INFORMATIVE, and then write a separate AUTHORITATIVE caching agent for applications only.

lwander:
We talked out of band, the fact that each agent prefixes their keys with the agent name means this should still work

lwander:
LGTM

PR #887, opened 2016-08-12
provider/google: Avoid NPE when enable a server group for an L7 load …

duftler:
…balancer that has zero backends.


duftler:
@jtk54 please review.

jtk54:
LGTM

PR #888, opened 2016-08-12
provider/google: Fix disable of SVGs with same name in different regions

jtk54:
Also fixes disabled state for the same case. @duftler or @lwander please review.


duftler:
LGTM.

PR #889, opened 2016-08-12
provider/google: Avoid NPE when deleting a server group with zero ins…

duftler:
…tances that is associated with an L4 load balancer.


duftler:
@jtk54 please review.

jtk54:
LGTM

PR #891, opened 2016-08-12
provider/google updated gceVCpuMaxList

danielpeach:
@duftler please review

- updated gceVCpuMaxList.
- defaults to 16 cpus if the provided location is not on the list.


duftler:
LGTM.

PR #893, opened 2016-08-12
Added credentials alias on openstack operations.

varikin:
Deck wants to send the account name as 'credentials' which has to be
overriden for every openstack operation done in deck. Switching to the
standard pattern.


varikin:
@spinnaker/reviewers PTAL

duftler:
LGTM.

PR #897, opened 2016-08-15
provider/cf: Add 'disk limit' as an option for deployment

gregturn:
Setting 'disk limit' when deploying to Cloud Foundry apps is also a critically needed feature.


gregturn:
@spinnaker/reviewers PTAL

duftler:
LGTM. Needs a rebase.

gregturn:
Rebased

PR #899, opened 2016-08-15
provider/cf: Clean up artifact downloads after upload

gregturn:
Files are pulled down and copied into temp files. Unfortunately, temp files aren't cleaned up until the JVM shuts down, something that doesn't happen often when run on a platform. This patch deletes the temp files after they've been uploaded.


gregturn:
@spinnaker/reviewers PTAL

gregturn:
Rebased

jtk54:
LGTM

PR #900, opened 2016-08-15
provider/cf: Find buildInfo information with safe navigation

gregturn:
Looking up buildInfo for a deployment needs safe navigation.


gregturn:
@spinnaker/reviewers PTAL

duftler:
LGTM.

PR #903, opened 2016-08-15
provider/google: Report L7 health check per backend service.

jtk54:
Used to report one per LB (incorrectly), now reports per backend service. @duftler or @lwander please review.


lwander:
LGTM

PR #907, opened 2016-08-16
provider/openstack: allow custom heat templates to be loaded

drmaas:


drmaas:
@derekolk @varikin @spinnaker/reviewers

clouddriver-web/config/clouddriver.yml
```diff
@@ -267,6 +267,7 @@ openstack:
       endpoint: ${OS_AUTH_URL:''}
       regions: ${OS_REGIONS:''}
       insecure: ${OS_INSECURE:false}
+      heatTemplatePath: /Users/z001mj7/.spinnaker
```
varikin: Probably not the best default. In fact, do you want to not have a default at all?

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/ops/servergroup/DeployOpenstackAtomicOperation.groovy
```diff
@@ -177,4 +184,20 @@ class DeployOpenstackAtomicOperation implements AtomicOperation<DeploymentResult
...
   }
 
+  /**
+   * Return the file contents of a template, either from the account config location or from the classpath.
+   * @param filename
+   * @return
+   */
+  String getTemplateFile(String filename) {
+    String template = ""
+    String tmplDir = description.credentials.credentials.heatTemplateLocation
```
varikin: This will read the path off the config option and then slurp in the file that already exists on disk, correct? Just want to ensure that the specified path isn't being passed in through the request.

clouddriver-web/config/clouddriver.yml
drmaas: oops lol

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/ops/servergroup/DeployOpenstackAtomicOperation.groovy
drmaas: it reads it from config, that is the description.credentials.credentials object.

lwander:
LGTM once the Derek & John are happy

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/ops/servergroup/DeployOpenstackAtomicOperation.groovy
```diff
@@ -177,4 +184,20 @@ class DeployOpenstackAtomicOperation implements AtomicOperation<DeploymentResult
     result
   }
 
+  /**
+   * Return the file contents of a template, either from the account config location or from the classpath.
+   * @param filename
+   * @return
+   */
```
derekolk: It feels like this should be cached ... I don't think we should be reading the file in on each server group creation.

varikin: We could, but how frequently does that happen?

drmaas: We could cache in simple map. Loading the template is going to be a small fraction of the time it takes to create the server group, though.

drmaas:
@derekolk

derekolk:
LGTM 👍

PR #909, opened 2016-08-16
Kubernetes : Return only image:tag for clusters and projects

bfarrell:
Update images by removing registry based on conversation in [deck #2530](https://github.com/spinnaker/deck/pull/2530).


bfarrell:
@lwander Here is the code for the updated list.

lwander:
LGTM

PR #915, opened 2016-08-18
provider/google: Associate SVG with all L7s that use the backend serv…

jtk54:
…ices it joins. @duftler @lwander please review. The load balancer name resolution is a bit hairy -- should we  just not support shared backend services?

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
@@ -822,6 +822,53 @@ class GCEUtil {
   }
 
   /**
+   * Resolve the L7 load balancer names that need added to the instance metadata.
+   *
+   * @param backendServiceNames - Backend service names explicitly included in the request.
+   * @param compute
+   * @param project
+   * @return List of L7 load balancer names to put into the instance metadata.
+   */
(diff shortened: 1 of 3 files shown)


lwander:
LGTM - defering to @duftler about supporting shared backend services

jtk54:
Shared backend services are supported in Pantheon, and resolving the LB names in this PR generates `N+2` calls for `N` load balancers, which isn't _too_ bad.

duftler:
A belated LGTM. Given that L7 load balancers cannot be created directly within the Spinnaker ui yet, and the alternative methods for creating them do support sharing backend services, seems like the right call to support them here. And as you pointed out, the cost isn't high.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -822,6 +822,53 @@ class GCEUtil {
...
+   * @param backendServiceNames - Backend service names explicitly included in the request.
+   * @param compute
+   * @param project
+   * @return List of L7 load balancer names to put into the instance metadata.
+   */
+  static List<String> resolveHttpLoadBalancerNamesMetadata(List<String> backendServiceNames, Compute compute, String project) {
+    def loadBalancerNames = []
+    def projectUrlMaps = compute.urlMaps().list(project).execute().getItems()
+    def servicesByUrlMap = [:]
+    projectUrlMaps.each { UrlMap urlMap ->
```
duftler: This could be done cleaner with a `.collectEntries {}`.

PR #917, opened 2016-08-19
Added on demand caching for security groups in openstack

varikin:


varikin:
@spinnaker/reviewers PTAL

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/provider/agent/OpenstackSecurityGroupCachingAgent.groovy
```diff
@@ -58,10 +83,80 @@ class OpenstackSecurityGroupCachingAgent extends AbstractOpenstackCachingAgent {
...
+      buildCacheResult(new CacheResultBuilder(startTime: Long.MAX_VALUE), securityGroups)
+    }
+
+    String namespace = SECURITY_GROUPS.ns
+    String resolvedKey = null
+    try {
+      resolvedKey = resolveKey(providerCache, namespace, key)
+      processOnDemandCache(cacheResult, objectMapper, metricsSupport, providerCache, resolvedKey)
+    } catch(UnresolvableKeyException e) {
+      log.info("Security group ${name} is not resolvable", e)
```
lwander: Wouldn't this cause the cache update to fail? Or really just never call `processOnDemandCache`?

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/provider/agent/OpenstackSecurityGroupCachingAgent.groovy
```diff
@@ -58,10 +83,80 @@ class OpenstackSecurityGroupCachingAgent extends AbstractOpenstackCachingAgent {
...
+         */
+        List<SecGroupExtension> groups = clientProvider.getSecurityGroups(region).findAll { it.name == name }
+        if (groups.size() == 1) {
+          group = groups.first()
+        } else {
+          log.warn("Failed to find unique security group with name ${name} in region ${region}")
+        }
+      } catch (OpenstackProviderException e) {
+        //Do nothing ... Exception is thrown if a security group isn't found
+        log.debug("Unable to find security group to add to OnDemand cache", e)
```
lwander: I think in this case you want to explicitly evict the SG from the cache.

varikin: It attempts to later on. The key includes name (not unique in OS) and id, but the request from Orca only includes a name so at this time, I can't construct the actual key. Later on, the key is created using '*' for the id and then `resolveKey` attempts to find a single key that matches on name, region, and account for a currently cached security group. As long as that key is resolved, when it calls `buildOnDemandCache`, the object will be evicted due to not having an object to add to the cache.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/provider/agent/OpenstackSecurityGroupCachingAgent.groovy
varikin: Yes. At this point, we never found, by name, the security group to be processed in Openstack and we couldn't find cached security group with this name, region, and account.

varikin:
@lwander is there more you want to discuss with this?

lwander:
LGTM, needs a rebase

varikin:
Rebased

PR #919, opened 2016-08-22
provider/openstack: Refactored upsert load balancer to use lbaasv2

derekolk:
closes https://github.com/spinnaker/spinnaker/issues/977


clouddriver-openstack/clouddriver-openstack.gradle
```diff
@@ -14,8 +14,13 @@
  * limitations under the License.
  */
 
+repositories {
+  maven { url "https://oss.sonatype.org/content/repositories/snapshots" }
```
cfieber: ok with this in the short-term but as soon as possible it would be great to get this out of here

derekolk: Created a ticket for this https://github.com/spinnaker/spinnaker/issues/1077

derekolk:
@spinnaker/reviewers

lwander:
A bit too big to review in detail, but this is local to the openstack provider. LGTM

PR #920, opened 2016-08-22
provider/kubernetes: Begin work on caching Replica Sets

lwander:
@duftler 

The goal is to allow users of Kubernetes to smoothly transition to replica sets. The next step will be to surface server groups backed by replica sets in the UI, which requires some more frontend work before it's feasible. Once everything is surfaced & cached properly, I'll switch the atomic operations to operate on replica sets over replication controllers, completing the migration.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
```diff
@@ -148,18 +149,24 @@ class KubernetesApiAdaptor {
...
-      client.replicationControllers().inNamespace(namespace).list().items
+  List<ReplicaSet> getReplicaSets(String namespace) {
+    atomicWrapper("Get Replica Sets", namespace) { KubernetesClient client ->
+      client.extensions().replicaSets().inNamespace(namespace).list().items
     }
   }
 
-  List<Pod> getReplicationControllerPods(String namespace, String replicationControllerName) {
-    atomicWrapper("Get Replication Controller Pods for $replicationControllerName", namespace) { KubernetesClient client ->
+  List<Pod> getReplicaSetPods(String namespace, String replicationControllerName) {
```
duftler: Should the other replication controller names here change as well (e.g. `replicationControllerName`)?

duftler:

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
lwander: Ah yeah that would read nicer

PR #921, opened 2016-08-22
provider/kubernetes: Unscheduled pods are "Down"

lwander:
Written to address https://github.com/spinnaker/spinnaker/issues/1049

Unscheduled pods are now considered down, causing pipelines/stages depending on the deployment to hang until the pods are scheduled. Unfortunately the API doesn't seem to list the reason why a container couldn't be scheduled, but at least this behavior is now more correct.

@duftler 

gjp4bsnbwtd


duftler:
LGTM.

PR #923, opened 2016-08-22
provider/kubernetes: Render replica sets

lwander:
drqke2hymxr

Turns out this was simpler than I was afraid it might be, since all UI details are stored in that DeployDescription object. 

@duftler

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesProviderUtils.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesProviderUtils.groovy
@@ -23,7 +23,10 @@ import com.netflix.spinnaker.cats.cache.CacheFilter
 import com.netflix.spinnaker.cats.cache.RelationshipCacheFilter
 import com.netflix.spinnaker.clouddriver.kubernetes.cache.Keys
 import com.netflix.spinnaker.clouddriver.kubernetes.model.KubernetesInstance
+import com.netflix.spinnaker.clouddriver.kubernetes.model.KubernetesServerGroup
 import io.fabric8.kubernetes.api.model.Pod
+import io.fabric8.kubernetes.api.model.ReplicationController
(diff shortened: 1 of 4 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesProviderUtils.groovy
```diff
@@ -65,4 +68,30 @@ class KubernetesProviderUtils {
...
+    if (replicationController) {
+      serverGroup = new KubernetesServerGroup(replicationController, instances, parse.account)
+    } else if (replicaSet) {
+      serverGroup = new KubernetesServerGroup(replicaSet, instances, parse.account)
+    } else {
+      throw new IllegalStateException("Expected either a ReplicaSet or ReplicationController") // Placate the linter
+    }
+
+    return serverGroup
+
```
duftler: Extraneous blank line.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesProviderUtils.groovy
```diff
@@ -65,4 +68,30 @@ class KubernetesProviderUtils {
...
+  static KubernetesServerGroup serverGroupFromCacheData(ObjectMapper objectMapper, CacheData cacheData, Set<KubernetesInstance> instances) {
+    ReplicationController replicationController
+    ReplicaSet replicaSet
+    if (cacheData.attributes.replicationController) {
+      replicationController = objectMapper.convertValue(cacheData.attributes.replicationController, ReplicationController)
+    } else if (cacheData.attributes.replicaSet) {
+      replicaSet = objectMapper.convertValue(cacheData.attributes.replicaSet, ReplicaSet)
+    } else {
+      throw new IllegalStateException("Expected either a ReplicaSet or ReplicationController")
```
duftler: Just a nit, but the order of the resource types in the error message don't match the order of the comparisons. Same thing down below.

duftler:
LGTM after addressing comments.

PR #943, opened 2016-08-25
Initial support for AWS Application Load Balancers

cfieber:
Right now it is treating the ALB the same as we treat an ELB - you request it by the name of the load balancer, and your server group is attached to all the target groups for that load balancer.

LoadBalancer names are not unique across classic and application load balancers, so currently this will attach both the classic and the application load balancer that match the requested name - not really sure that is the best strategy. We could fail the deploy if this was the case, we could just use the classic LB if it matches on name (I guess more backwards compatible?)

I could also see us having to evolve the deploy description if we want to support fancier ALB use cases (only attach my server group to a specific target group, not to every target group in the ALB). I'm sure at some point that will be a request...


cfieber:
WCGW!

PR #945, opened 2016-08-25
provider/kubernetes: Stop caching job objects

lwander:
The motivation is that "jobs" are retrieved only from orca during stage-completion checks. Since "jobs" are now Spinnaker instances, they can still be surfaced using the search provider - but are really only surfaced in pipelines. @duftler


duftler:
LGTM.

PR #946, opened 2016-08-25
provider/google: Added toString for GStrings when serializing

nwwebb:
GStrings are stored in an odd format when put in a GCS bucket, converting them to strings solves this problem. @lwander @duftler


lwander:
Ah, the classic GString problem. LGTM

PR #947, opened 2016-08-25
Merge L7 and L4 operations into same code path.

jtk54:
@duftler or @lwander please review.


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/converters/UpsertGoogleLoadBalancerAtomicOperationConverter.groovy
```diff
@@ -28,6 +30,19 @@ import org.springframework.stereotype.Component
 @Component("upsertGoogleLoadBalancerDescription")
 class UpsertGoogleLoadBalancerAtomicOperationConverter extends AbstractAtomicOperationsCredentialsSupport {
   AtomicOperation convertOperation(Map input) {
+    UpsertGoogleLoadBalancerDescription description = convertDescription(input)
+    switch (description.loadBalancerType) {
+      case  GoogleLoadBalancerType.NETWORK:
```
duftler: Extraneous space after `case`.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/converters/UpsertGoogleLoadBalancerAtomicOperationConverter.groovy
```diff
@@ -28,6 +30,19 @@ import org.springframework.stereotype.Component
...
+        break
+      case GoogleLoadBalancerType.HTTP:
+        return new CreateGoogleHttpLoadBalancerAtomicOperation(description)
+        break
+      default:
+        // TODO(jacobkiefer): This is for backwards compatibility for L4 Upsert.
+        return new UpsertGoogleLoadBalancerAtomicOperation(description)
+        break
+    }
     new UpsertGoogleLoadBalancerAtomicOperation(convertDescription(input))
```
duftler: Isn't this line unreachable now?

clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/deploy/validators/UpsertGoogleLoadBalancerDescriptionValidatorSpec.groovy
```diff
@@ -126,4 +155,200 @@ class UpsertGoogleLoadBalancerDescriptionValidatorSpec extends Specification {
       1 * errors.rejectValue('loadBalancerName', _)
       1 * errors.rejectValue('region', _)
   }
+
+  void "pass validation with proper http description inputs"() {
+    setup:
+    def input = [
```
duftler: Up to this point, this file has the content of the `setup`/`when`/`then` blocks indented.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/description/UpsertGoogleLoadBalancerDescription.groovy
```diff
@@ -26,6 +29,8 @@ class UpsertGoogleLoadBalancerDescription extends AbstractGoogleCredentialsDescr
   String portRange
   String region
   String accountName
+  GoogleHttpLoadBalancer googleHttpLoadBalancer
```
duftler: Feels a bit weird to have this as an attribute since it contains some of the same attributes that are already siblings of `googleHttpLoadBalancer`. Haven't thought very deeply about it, but would have expected this description to have as top-level elements the full set of load balancer attributes; and then some subset would apply to each type of load balancer.

jtk54: The L7 description needs the layout of the backend services in the URL map, which is encoded in the `GoogleHttpLoadBalancer`. We could unpack that into its constituent `healthChecks`, `backendServices` and `hostRules` etc, but we'd be duplicating the structure of the `GoogleHttpLoadBalancer` object.

duftler: Maybe you can have some kind of shared base class between GoogleHttpLoadBalancer and a GoogleLoadBalancer (really 'network load balancer')? Something like both GoogleHttpLoadBalancer and GoogleNetworkLoadBalancer extending GoogleLoadBalancer, and GoogleLoadBalancer itself is abstract?
Just feels hinky to have both `description.portRange` and `description. googleHttpLoadBalancer.portRange` (same for the other similar fields).

jtk54: Yeah that seems cleaner -- maybe we can use `GoogleLoadBalancerView` for that, but just push the common fields into one object. Got it.

clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/deploy/converters/UpsertGoogleLoadBalancerAtomicOperationConverterUnitSpec.groovy
```diff
@@ -81,4 +88,79 @@ class UpsertGoogleLoadBalancerAtomicOperationConverterUnitSpec extends Specifica
...
+                "healthCheck": hc,
+              ]
+            ]
+          ]
+        ]
+      ]
+    ]
+
+    when:
+    def description = converter.convertDescription(input)
```
lwander: I think all other google unit tests indent after the `when:` and `then:` style blocks

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/description/UpsertGoogleLoadBalancerDescription.groovy
jtk54: I actually went with the first option. Copying the object structure is easier than refactoring the description for the L4 operation.

duftler:
Ok, sgtm.
Can always revisit this decision when working on the next load balancer type if you like.

duftler:
LGTM.

PR #948, opened 2016-08-25
return cloud provider types for synthetically generated applications

tomaslin:
If there are multiple cloud providers configured, we currently don't indicate which cloud providers are used for synthetic applications. This change makes the create / clone dialogs work correctly based on the clusters that caused the synthetic application to be created


clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/controllers/ApplicationControllerSpec.groovy
```diff
@@ -23,6 +23,7 @@ import com.netflix.spinnaker.clouddriver.model.ServerGroup
 import com.netflix.spinnaker.clouddriver.aws.model.AmazonApplication
 import spock.lang.Shared
 import spock.lang.Specification
+import spock.lang.Unroll
 
 import java.lang.Void as Should
```
cfieber: did somebody use the tabs-to-separate-words formatting style here?

cfieber:
giv'er

PR #953, opened 2016-08-26
provider/kubernetes: Allow envvars to come from k8s sources

lwander:
Now env vars can come from both secrets and config maps (which will be configurable in the "properties" stage) @duftler


duftler:
LGTM.

PR #955, opened 2016-08-26
(consul) fix health check reporting for disabled instances

lwander:
@duftler

### DIFF
--- a/clouddriver-consul/src/main/groovy/com/netflix/spinnaker/clouddriver/consul/api/v1/model/CheckResult.groovy
+++ b/clouddriver-consul/src/main/groovy/com/netflix/spinnaker/clouddriver/consul/api/v1/model/CheckResult.groovy
@@ -18,30 +18,32 @@ package com.netflix.spinnaker.clouddriver.consul.api.v1.model
 
 import com.fasterxml.jackson.annotation.JsonProperty
 
+// JsonProperty seems to misbehave here. The field names are capitalized, but their json counterparts are lower-case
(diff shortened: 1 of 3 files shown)


clouddriver-consul/src/main/groovy/com/netflix/spinnaker/clouddriver/consul/api/v1/model/CheckResult.groovy
```diff
@@ -19,29 +19,29 @@ package com.netflix.spinnaker.clouddriver.consul.api.v1.model
 import com.fasterxml.jackson.annotation.JsonProperty
 
 class CheckResult {
-  @JsonProperty("Node")
-  String node
+  @JsonProperty("node")
```
duftler: Can you use http://fasterxml.github.io/jackson-annotations/javadoc/2.0.0/com/fasterxml/jackson/annotation/JsonSetter.html instead of capitalizing the field names?

lwander: No combination of JsonSetter & Getter seems to fix this. I'll drop a comment in explaining why they are capitlized. You'll see however that you can still access these fields in their lower-case form: https://github.com/spinnaker/clouddriver/pull/955/commits/a444bc3dfc02dba5e108bc4f2272eb8f04b11b7e#diff-dd9c15219119c26655d8f40b3f9e0e0dR30

duftler:
LGTM.

PR #959, opened 2016-08-29
provider/google: Fix L7 clones when L4 SSL LB exists in project.

jtk54:
@duftler or @lwander please review.


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -867,10 +867,10 @@ class GCEUtil {
           proxy = compute.targetHttpsProxies().get(project, getLocalName(fr.target)).execute()
           break
         default:
-          throw new GoogleOperationException("Illegal proxy type for ${fr.target}.")
+          proxy = null
```
duftler: Looks like `proxy` is already `null` here.

duftler:

PR #960, opened 2016-08-29
(kubernetes) Job pod container names can be empty

lwander:
@danielpeach FYI (in fact, they shouldn't ever need to be specified from the UI).


duftler:
LGTM.

PR #961, opened 2016-08-29
(kubernetes) slight improvments to health widget contents

lwander:
@duftler. The goal is to better explain why certain kubernetes-level events fail (scheduling decisions, probe failures, etc...) but it looks like this information isn't exposed at the resource level, so I will have to dig a little deeper.


duftler:
LGTM

PR #963, opened 2016-08-29
(kubernetes) [] is groovy true in null-safe check

lwander:
So this is pretty bizarre - 

```
[]?.get(0)
```

will fail with an `IndexOutOfBounds` exception, which you wouldn't expect, since `[]` is groovy-false.


lwander:
@danielpeach @duftler

duftler:
Pretty sure the issue is just that `[]` is not null, so `.get(0)` is used to try to fetch whatever is at index sub `0`. Since nothing is, it throws that exception. If you use `.getAt(0)`, you'll get `null` instead of that exception.

lwander:
Yeah, I had always assumed that 

```
x?.y
``` 

was desugared into

```
if (x) {
  y
}
```

rather than 

```
if (x != null) {
  y
}
```

PR #968, opened 2016-08-30
(kubernetes) log pod events

lwander:
Capture pod events, cache them, and send to the UI.

@duftler PTAL, @danielpeach FYI


duftler:
LGTM.

PR #969, opened 2016-08-30
Renamed lb method to algorithm

varikin:
The field was named different between upsert and get.


varikin:
@spinnaker/reviewers PTAL

lwander:
LGTM

PR #971, opened 2016-08-30
(docker): stop returning empty images for missing accounts

lwander:
@duftler


lwander:
When changing docker configuration, keys that haven't been evicted yet would be pulled out of the cache, but weren't resolvable into an existing account. This would cause the image to appear as `{}` to deck, resulting in UI errors.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/provider/agent/DockerRegistryImageCachingAgent.groovy
```diff
@@ -88,12 +88,13 @@ class DockerRegistryImageCachingAgent implements CachingAgent, AccountAware {
           return [:]
       }
       DockerRegistryTags tags
+      tags = null
```
duftler: Combine with previous line?

duftler:

PR #972, opened 2016-08-30
(kubernetes) fix application delete

lwander:
This is an artifact of having a job and servergroup caching agent adding applications at the same time. Changing the agent to authoritative now allows applications to be deleted again (since jobs are no longer cached).


lwander:
@ttomsu

ttomsu:
LGTM

PR #974, opened 2016-08-30
(kubernetes) handle kubernetes-level errors when running jobs

lwander:
There are error conditions surfaced when a container is set to "waiting" that will cause a pipeline to hang indefinitely that depends on a job. These surface those errors: https://github.com/kubernetes/kubernetes/blob/de92db35206c986c8313400910e87773df3144a2/pkg/kubelet/images/types.go


lwander:
@duftler PTAL @skim1420  FYI

duftler:
LGTM.

PR #976, opened 2016-08-31
(kubernetes) report server group events

lwander:
I also dropped the `Reason` enum because there were too many undocumented reasons for it to make sense to keep around, when all we were interested in was the string.

@duftler


duftler:
LGTM, but not to (Metal) Travis.

PR #975, opened 2016-08-31
(google) Speed up unit tests by mocking poller clock and decreasing r…

duftler:
…etry interval.

PR #977, opened 2016-08-31
adds slowTest build flag to log slow tests

cfieber:
use with `-PslowTest` or `-PslowTest=<threshold millis>` (if unspecified the threshold is 250ms)

any test that runs for longer than the provided threshold will be logged with its execution timing

if the flag is not supplied, test timing is not logged


cfieber:
Much thanks to @duftler for already fixing the worst offenders that this found in #975

PR #980, opened 2016-09-01
Support rudimentary templating in default IAM roles

ajordens:
- Currently only handles {[application}}


ajordens:
Machine was barfing ... I'll fix these tests.

PR #983, opened 2016-09-02
support setting the job type to experiment with batch titus jobs

dzapata:


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/model/TitusServerGroup.groovy
```diff
@@ -54,6 +54,7 @@ class TitusServerGroup implements ServerGroup, Serializable {
   TitusServerGroup(Job job, String account, String region) {
     id = job.id
     name = job.name
+    type = job.type
```
tomaslin: this is probably not necessary and incorrect since we don't index batch jobs

dzapata: Good catch, fixed.

tomaslin:
LGTM, please merge down your commits

tomaslin:
closing because operation won't complete without further changes in clouddriver.

PR #985, opened 2016-09-02
 provider/openstack: allow providing user data to server group create/clone

drmaas:


drmaas:
@derekolk @varikin @spinnaker/reviewers

drmaas:
This is a prerequisite for external load balancers with consul.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/ops/servergroup/DeployOpenstackAtomicOperation.groovy
```diff
@@ -147,9 +148,22 @@ class DeployOpenstackAtomicOperation implements AtomicOperation<DeploymentResult
       Subnet subnet = provider.getSubnet(description.region, subnetId)
       task.updateStatus BASE_PHASE, "Found network id $subnet.networkId from subnet $subnetId."
 
+      String userData
+      if (description.userData) {
+        if (description.userData.startsWith("http")) {
+          task.updateStatus BASE_PHASE, "Resolving user data from url $description.userData..."
+          userData = description.userData.toURL()?.text
```
lwander: What's the value in doing this?

lwander:
Why is this needed for consul support?

drmaas:
I was assuming the user needed to provide the script to boot consul, based on how consul was baked into the image. Are you saying that the startup script will be auto-generated based on clouddriver.yml config?

User data can also be used to execute other arbitrary startup functions, so I don' think the need goes away, even if the answer to the above is an auto-generated consul startup script.

lwander:
Ideally the image should take care of running consul, but providing a user-script is definitely one way of doing this. It's probably better practice to make an upstart entry for the consul agent to handle crashes & ensure the agent is always running. But Spinnaker shouldn't take a hard stance on how to run Consul, so this is good to add.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/ops/servergroup/DeployOpenstackAtomicOperation.groovy
drmaas: Fetching scripts from an object store, e.g. http://docs.openstack.org/developer/swift

lwander:
LGTM

PR #992, opened 2016-09-05
(kubernetes) specify image pull policy

lwander:
@duftler


duftler:
LGTM.
As discussed offline, please clean up the enum-handling here and elsewhere so we can avoid having to keep multiple/external string<->constant mappings in sync.

PR #994, opened 2016-09-06
Different Ebs volume sizes per capacity

tomaslin:


cfieber:
worth noting that the sizes picked for `m4` ebs match the ephemeral size that was available on the `m3` equivalent instance

cfieber:
LGTM

PR #996, opened 2016-09-06
provider/google: More detail in safeRetry reporting.

jtk54:
@duftler @lwander please review. Instead of more reporting in specific cases, I made the final `OperationException` more verbose with the final encountered exception.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
@@ -24,6 +24,7 @@ import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.api.client.http.HttpHeaders
 import com.google.api.client.http.HttpRequest
 import com.google.api.client.http.HttpRequestInitializer
+import com.google.api.client.http.HttpResponseException
 import com.google.api.services.compute.Compute
 import com.google.api.services.compute.model.*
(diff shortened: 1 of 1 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -961,11 +962,23 @@ class GCEUtil {
...
+      def lastSeenError = lastSeenException?.getDetails()?.getErrors()[0] ?: null
+      if (lastSeenException && lastSeenError) {
+        throw new GoogleOperationException("Failed to $action $resource after #$tries."
+        + " Last seen exception has status code ${lastSeenException.getStatusCode()} with error message ${lastSeenError.getMessage()}"
+        + " and reason ${lastSeenError.getReason()}.")
+      } else if (lastSeenException) {
+        throw new GoogleOperationException("Failed to $action $resource after #$tries."
+          + " Last seen exception has status code ${lastSeenException.getStatusCode()} with message ${lastSeenException.getMessage()}.")
+      } else {
+        throw new GoogleOperationException("Failed to $action $resource after #$tries.")
```
lwander: Is this reachable? If not this should be a different kind of exception or an assertion as a sanity check for you.

jtk54: This is currently reachable if we timeout 10 times. It would be better to also catch the retry exceptions and report them as well.

lwander:
LGTM

PR #997, opened 2016-09-06
(kubernetes) batch & frontload event lookup

lwander:
@duftler I saw the servergroup & pod refresh cycles taking abnormally long (1 minute +), and realized it was due to the events being loaded for each resource individually. This approach loads the events all at once, greatly speeding up the refresh cycles.


duftler:
LGTM.

PR #1002, opened 2016-09-08
provider/google: Added UUID in path matcher name.

jtk54:
@duftler please review.


duftler:
Can you please add some context describing _why_ this change is being made, as opposed to just what the actual change is?

jtk54:
Previously, the names of the path matchers occasionally collided because `currentTimeMillis` reported the same time in different loops. The `UUID` method is a bit more resilient.

duftler:
LGTM.

PR #1003, opened 2016-09-09
adds caching agent for ELBv2 health state

cfieber:
adds healthState attribute to the per-loadbalancer detail in the health payload

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/edda/InstanceLoadBalancerState.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/edda/InstanceLoadBalancerState.groovy
@@ -16,21 +16,59 @@
 
 package com.netflix.spinnaker.clouddriver.aws.model.edda
 
+import com.netflix.spinnaker.clouddriver.model.HealthState
 import groovy.transform.CompileStatic
 import groovy.transform.EqualsAndHashCode
-import groovy.transform.Immutable
 
-@Immutable
 @CompileStatic
 @EqualsAndHashCode(cache = true)
 class InstanceLoadBalancerState {
-  String instanceId
-  String loadBalancerName
-  String state
-  String reasonCode
-  String description
+  final String instanceId
+  final String loadBalancerType
+  final String loadBalancerName
+  final String state
+  final String reasonCode
+  final String description
+
+  final HealthState healthState
+
+  InstanceLoadBalancerState(String instanceId, String loadBalancerType, String loadBalancerName, String state, String reasonCode, String description) {
+    this.instanceId = instanceId
+    this.loadBalancerType = loadBalancerType
+    this.loadBalancerName = loadBalancerName
+    this.state = state
+    this.reasonCode = reasonCode
+    this.description = description
+    this.healthState = deriveHealthState()
+  }
 
   static List<InstanceLoadBalancerState> fromLoadBalancerInstanceState(LoadBalancerInstanceState lbis) {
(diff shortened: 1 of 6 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/edda/InstanceLoadBalancerState.groovy
```diff
@@ -16,21 +16,59 @@
...
-  String reasonCode
-  String description
+  final String instanceId
+  final String loadBalancerType
+  final String loadBalancerName
+  final String state
+  final String reasonCode
+  final String description
+
+  final HealthState healthState
```
cfieber: this is now part of this object representation so the front end can use it instead of interpreting `state` (InService on classic load balancers, vs healthy on ELBv2)

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonLoadBalancerV2InstanceStateCachingAgent.groovy
```diff
@@ -0,0 +1,142 @@
...
+
+    log.info("Caching ${lbHealths.size()} items in ${agentType}")
+
+    return new DefaultCacheResult(
+      (HEALTH.ns): lbHealths,
+      (INSTANCES.ns): instances)
+  }
+
+  static Optional<String> extractLoadBalancerName(String loadBalancerArn) {
+    def m = ALB_ARN_PATTERN.matcher(loadBalancerArn)
```
cfieber: this is a bit dirty... but saves actually having to describe load balancers

cfieber:

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/edda/InstanceLoadBalancerState.groovy
```diff
@@ -16,21 +16,59 @@
...
+        * Instance is not currently registered with the LoadBalancer.
+     */
+    if (state == 'InService' || state == 'healthy') {
+      return HealthState.Up
+    }
+
+    if (state == 'initial' || state == 'Instance registration is still in progress') {
+      return HealthState.Starting
+    }
+    if (state == 'unused' || state == 'draining' || state == 'Instance is not currently registered with the LoadBalancer.') {
```
ajordens: Is `Instance is not ...` a legitimate `state` ... same with the other states.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/edda/InstanceLoadBalancerState.groovy
```diff
@@ -16,21 +16,59 @@
...
+      return HealthState.OutOfService
+    }
+    if (description == 'Instance is in the EC2 Availability Zone for which LoadBalancer is not configured to route traffic to.') {
```
ajordens: Any particular reason for handling this even when it resolves to the default of `HealthState.Down`?

ajordens:
LGTM

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/edda/InstanceLoadBalancerState.groovy
cfieber: ah I did some `#fail` when I was relocating this and making it non static, nice catch, thanks!

PR #1005, opened 2016-09-09
provider/google: Null guard for L7 pathRules in caching agent.

jtk54:
@duftler please review. GCP can return `null` for the `pathRule` in a `pathMatcher`. The frontend expects a list of some sort (even if empty), so I just added a null guard defaulting to `[]`.


duftler:
LGTM.

PR #1006, opened 2016-09-09
provider/google: Properly dedupe backend services in upsert.

jtk54:
@duftler please review. When a backend service has a backend associated, `unique()` equality check doesn't work properly because of the backend list. This `unique` checks everything but the backends, which aren't managed by L7 upsert.


duftler:
LGTM.

jtk54:
Took an alternate route to get to the same place -- changed the equals function in `GoogleBackendService`. The list comparison is not reliable anyway since it's address-based.

duftler:
ok, sgtm

PR #1008, opened 2016-09-09
Add implementation support for selecting an existing virtual network …

duftler:
…and subnet when creating a load balancer (ApplicationGateway) or server group.


duftler:
Cherry-picks commit from https://github.com/spinnaker/clouddriver/pull/1004.

duftler:
Closes https://github.com/spinnaker/clouddriver/pull/1004.

PR #1010, opened 2016-09-12
Lookup owner of AMI in AllowLaunch operation.

varikin:
If the AMI is baked with a different account that the one being used to
deploy, that owner account needs to be used to update the permissions on
that AMI.

### DIFF
--- a/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AllowLaunchAtomicOperationUnitSpec.groovy
+++ b/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AllowLaunchAtomicOperationUnitSpec.groovy
@@ -68,7 +68,7 @@ class AllowLaunchAtomicOperationUnitSpec extends Specification {
         assert dir.filters.first().name == 'name'
         assert dir.filters.first().values == ['super-awesome-ami']
 
-        new DescribeImagesResult().withImages(new Image().withImageId('ami-12345'))
+        new DescribeImagesResult().withImages(new Image().withImageId('ami-12345').withOwnerId('67890'))
     }
     1 * ec2.modifyImageAttribute(_)
   }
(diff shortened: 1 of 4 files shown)


varikin:
@spinnaker/reviewers PTAL

We ran into an issue with multi-region bakes where the account used that bakes is different than the ones clouddriver is used to deploy.

edwinavalos:
To add on to this, we were able to accomplish multi-region through multiple-account baking. But we ran into the issue that Spinnaker couldn't tell if we baked the second region with the same account as first "home" region. It ends up using the "home" region credentials which can't modify certain rights that only the originating account can.

I think our use case might be a little outside the original intention of 'multi-region bake' as we've had to do a lot of working to get multi-account baking to work, but in the end this should offer value to others that end up going down our path with one account per region.

ajordens:
PTAL @cfieber

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AllowLaunchAtomicOperationUnitSpec.groovy
```diff
@@ -162,13 +165,57 @@ class AllowLaunchAtomicOperationUnitSpec extends Specification {
...
     with(sourceAmazonEc2) {
-      1 * describeImages(_) >> new DescribeImagesResult().withImages(new Image().withImageId("ami-123456"))
+      1 * describeImages(_) >> new DescribeImagesResult().withImages(new Image().withImageId("ami-123456").withOwnerId(testCredentials.accountId))
       1 * modifyImageAttribute(_)
     }
 
     0 * _
   }
 
+  void "should lookup owner account of resolved ami and "() {
```
ajordens: (minor) "resolved ami and"... ? Looks cut off.

varikin: Not sure what I was thinking to finish that with

varikin:
Rebased

cfieber:
if you give it one more rebase we'll fire ze mergemissiles

varikin:
Rebased again.

PR #1013, opened 2016-09-12
provider/google: Refactor safeRetry and handle socket timeout.

jtk54:
@duftler @lwander please review. Refactored `safeRetry` into a generic, type-safe class since `getOperation` actually returns something instead of `void`. Added socket timeout handling instead of `GoogleOperationTimedOutException` since the latter is not something the platform throws.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GoogleOperationPoller.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GoogleOperationPoller.groovy
@@ -92,7 +92,7 @@ class GoogleOperationPoller {
     The timeoutSeconds parameter is really treated as a lower-bound. We will poll until the operation reaches a DONE
     state or until <em>at least</em> that many seconds have passed.
    */
-  private Operation waitForOperation(Closure getOperation, long timeoutSeconds) {
+  private Operation waitForOperation(Closure<Operation> getOperation, long timeoutSeconds) {
     int totalTimePollingSeconds = 0
     boolean timeoutExceeded = false
 
@@ -105,7 +105,8 @@ class GoogleOperationPoller {
 
       totalTimePollingSeconds += pollInterval
 
-      Operation operation = getOperation()
+      def retry = new SafeRetry<Operation>()
+      Operation operation = retry.doRetry(getOperation, "wait", "operation", null, null, [], [])
 
       if (operation.getStatus() == "DONE") {
         return operation
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/SafeRetry.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/SafeRetry.groovy
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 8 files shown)


lwander:
Good stuff, LGTM

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GoogleOperationPoller.groovy
```diff
@@ -105,7 +105,8 @@ class GoogleOperationPoller {
 
       totalTimePollingSeconds += pollInterval
 
-      Operation operation = getOperation()
+      def retry = new SafeRetry<Operation>()
+      Operation operation = retry.doRetry(getOperation, "wait", "operation", null, null, [], [])
```
duftler: Nice.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/SafeRetry.groovy
```diff
@@ -0,0 +1,104 @@
...
+      int tries = 1
+      Exception lastSeenException = null
+      while (tries < 10) { // Retry 10 times.
+        try {
+          tries++
+          sleep(SAFE_RETRY_INTERVAL_MILLIS) // Sleep for some interval between attempts.
+          log.warn "$action $resource attempt #$tries..."
+          return operation()
+          log.warn "$action $resource attempt #$tries succeeded"
+          return
```
duftler: Is this `return` statement (and the logging call above it) reachable?

Looks like you probably want to capture the result instead of just using `return operation()` and then return the captured result here.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GoogleOperationPoller.groovy
duftler: Hm, looks strange to call a static method via an instance reference.
Why not drop the `static` from the `doRetry()` method?

clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/DestroyGoogleServerGroupAtomicOperationUnitSpec.groovy
```diff
@@ -281,8 +282,8 @@ class DestroyGoogleServerGroupAtomicOperationUnitSpec extends Specification {
   }
 
   @Unroll
-  void "should retry http backend deletion on 400, 412, succeed on 404"() {
-    // Note: Implicitly tests GCEUtil.safeRetry
+  void "should retry http backend deletion on 400, 412, sock timeout, succeed on 404"() {
```
duftler: Laundry day?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GoogleOperationPoller.groovy
jtk54: Dropped static, now call sites that don't return anything use `SafeRetry<Void>`.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/SafeRetry.groovy
```diff
@@ -0,0 +1,102 @@
...
+   * @param resource - Resource we are operating on.
+   * @param task - Spinnaker task. Can be null.
+   * @param phase
+   * @param retryCodes - GoogleJsonResponseException codes we retry on.
+   * @param successfulErrorCodes - GoogleJsonException codes we treat as success.
+   *
+   * @return Object of type T returned from the operation.
+   */
+  public T doRetry(Closure<T> operation,
+                          String action,
```
duftler: Indentation is off here.

duftler:
LGTM after addressing final comment.

PR #1014, opened 2016-09-13
updates amazon security group caching for better interaction of on demand and edda caching

cfieber:
for an eddaEnabled amazon account, only use the edda data if the lastModified time is newer than the last time a force cache refresh happened

we only have to track lastModified time since an on_demand and a read from edda are otherwise the same - they load the full data set for the account/region

@spinnaker/netflix-reviewers


ajordens:
LGTM

PR #1016, opened 2016-09-13
Don't print stacktrace when no application is found.

varikin:
When a new application is created via Deck (to Gate -> Orca -> Front50), a GET
request for the application is made to Clouddriver and it fails because
clouddriver doesn't know about this application yet. Clouddriver only
learns about applications through having a server group in that
application. So, keeping the logging, not the stacktrace.


varikin:
@spinnaker/reviewers PTAL

varikin:
@spinnaker/reviewers PTAL?

ajordens:
Seems reasonable to me, thanks for the fix.

PR #1020, opened 2016-09-14
(kubernetes) Translate objects before loading them into the cache

lwander:
In an effort to make the clusters tab more responsive I do the translation of kubernetes resource -> spinnaker resource before the instance and server group object enters the cache. This is at the expense of more overall CPU usage.


duftler:
LGTM.

PR #1022, opened 2016-09-15
Do not bother marking `Terminating` instances as UP/OUT_OF_SERVICE in…

ajordens:
… discovery


ajordens:
PTAL @cfieber ... tested this out with:

- resize asg to 20
- disable asg
- resize asg to 15
- immediately enable

The enable repeated failed prior to this change

PR #1023, opened 2016-09-15
Bug fix; don't overwrite the user's selection for vnet/subnet

milismsft:
Bug fix; we end-up overwriting the user's selection for vnet/subnet with default values after the validation check. This causes the deployment to fail later on if the resources corresponding to those default values do not exist.


milismsft:
@spinnaker/netflix-reviewers @duftler PTAL and merge if looks good.

duftler:
LGTM.

PR #1030, opened 2016-09-19
Removed time parsing workaround for old Openstack version.

varikin:
Since Kilo isn't supported anymore, we don't need to try parsing the
time with a timezone and fallback to without the timezone anymore.


varikin:
@spinnaker/reviewers PTAL

PR #1033, opened 2016-09-19
(kubernetes) horizontal pod autoscaling

lwander:
This is missing semantics for "resize" and "clone", those are up next.

@duftler


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesAutoscalerStatus.groovy
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright 2016 Netflix, Inc.
```
duftler: s/Netflix/Google

PR #1036, opened 2016-09-20
provider/google: Implemented ssl cert caching agent.

jtk54:
@duftler please review. @danielpeach FYI. This caches `sslCertificates` so we can suggest them in L7 upsert.


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/cache/Keys.groovy
```diff
@@ -215,6 +216,10 @@ class Keys {
     Names names = Names.parseName(managedInstanceGroupName)
     "$GoogleCloudProvider.GCE:${Namespace.SERVER_GROUPS}:${names.cluster}:${account}:${region}:${names.group}${zone ? ":$zone" : ""}"
   }
+  static String getSslCertificateKey(String account,
```
duftler: I think you'll also need to add support for the new key in `parse()`.

PR #1039, opened 2016-09-20
(google) Add additional retry behavior around compute calls.

duftler:
(google) Reduce number of live compute calls.


duftler:
@jtk54 please review.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -225,14 +224,23 @@ class GCEUtil {
...
+    def forwardingRules = new SafeRetry<List<ForwardingRule>>().doRetry(
+      { return compute.forwardingRules().list(projectName, region).execute().items },
+      "list",
+      "regional forwarding rules",
+      task,
+      phase,
+      RETRY_ERROR_CODES,
+      []
+    )
+    def foundForwardingRuleNames = forwardingRules.findAll {
```
jtk54: This filter returns the actual forwarding rules, not just the names. The variable name is a bit misleading.

duftler: Good point. I've changed it back.

PR #1042, opened 2016-09-21
provider/google: Cache search result hydrator for health checks.

jtk54:
We should suggest `HttpHealthCheck`s in L7 upsert (among other operations). This adds the JSON serialized object to the `/search` result for health checks. @duftler please review. @danielpeach FYI.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/GoogleInfrastructureProvider.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/GoogleInfrastructureProvider.groovy
@@ -62,17 +62,6 @@ class GoogleInfrastructureProvider extends AgentSchedulerAware implements Search
     return Keys.parse(key)
   }
 
-  private static class HttpHealthCheckResultHydrator implements SearchableProvider.SearchResultHydrator {
-
-    @Override
-    Map<String, String> hydrateResult(Cache cacheView, Map<String, String> result, String id) {
-      CacheData healthCheckCacheData = cacheView.get(HTTP_HEALTH_CHECKS.ns, id)
-      return result + [
-        httpHealthCheck: JsonOutput.toJson(healthCheckCacheData.attributes.httpHealthCheck)
-      ]
-    }
-  }
-
   private static class BackendServiceResultHydrator implements SearchableProvider.SearchResultHydrator {
 
     @Override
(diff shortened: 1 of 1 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/GoogleInfrastructureProvider.groovy
```diff
@@ -51,15 +52,27 @@ class GoogleInfrastructureProvider extends AgentSchedulerAware implements Search
...
     return Keys.parse(key)
   }
 
+  private static class HttpHealthCheckResultHydrator implements SearchableProvider.SearchResultHydrator {
+
+    @Override
+    Map<String, String> hydrateResult(Cache cacheView, Map<String, String> result, String id) {
+      CacheData healthCheckCacheData = cacheView.get(HTTP_HEALTH_CHECKS.ns, id)
+      return result + [
+        httpHealthCheck: JsonOutput.toJson(healthCheckCacheData.attributes.httpHealthCheck)
```
duftler: Is this treating the value as a string instead of as a json object?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/GoogleInfrastructureProvider.groovy
```diff
@@ -51,15 +52,27 @@ class GoogleInfrastructureProvider extends AgentSchedulerAware implements Search
...
+    (HTTP_HEALTH_CHECKS.ns): new HttpHealthCheckResultHydrator(),
+    (INSTANCES.ns): new InstanceSearchResultHydrator()
   ]
 
   @Override
   Map<String, String> parseKey(String key) {
     return Keys.parse(key)
   }
 
+  private static class HttpHealthCheckResultHydrator implements SearchableProvider.SearchResultHydrator {
```
duftler: Could sort the class definitions to match the order in which they are referenced in `searchResultHydrators`.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/GoogleInfrastructureProvider.groovy
jtk54: Yes, the return type is `Map<String, String>`, so it's a serialized JSON string.

duftler: Ah, ok

PR #1049, opened 2016-09-22
Migrate from v1 -> v2 Front50 API

ajordens:
- Also pass `pageSize` down to Front50


ajordens:

PR #1051, opened 2016-09-23
Filter server groups by cloud provider

ajordens:


ajordens:
Worth supporting a `cloudProvider` filter @cfieber?

cfieber:
Would be nice to not filter post query, could save a pretty significant amount of work

Also could consider an expand param on /applications/:app/clusters/:account

ajordens:
That's fair .... figured as much but opt'd for the fewer lines of code in the first commit.

I've updated to filter `clusterProviders` if a `cloudProvider` was specified.

I'm less inclined on `/clusters/:account?expand=true` as a solution to this "issue" (although supporting ?expand more generally makes sense to me).

PR #1053, opened 2016-09-23
provider/google: Replaced String LB type with Enum in View.

jtk54:
There were some nasty comparisons casting `Enum`s to `String`s, in some of the load balancing code, so I cleaned that up a bit. I tested this with all permutations of L4/L7 with all the operations. @duftler please review.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController.groovy
@@ -40,8 +40,6 @@ class GoogleLoadBalancerController {
   @Autowired
   GoogleLoadBalancerProvider googleLoadBalancerProvider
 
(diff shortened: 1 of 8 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController.groovy
```diff
@@ -69,8 +67,8 @@ class GoogleLoadBalancerController {
             account: view.account,
             region: view.region,
             name: view.name,
-            loadBalancerType: loadBalancerType,
-            backendServices: loadBalancerType == HTTP ? backendServices.unique() as List<String> : null
+            loadBalancerType: loadBalancerType as String,
```
duftler: Any reason not to also make `GoogleLoadBalancerSummary.loadBalancerType` the `enum` type instead of a `String`?

jtk54: Nope, good call.

PR #1060, opened 2016-09-26
provider/google: Represent L4/L7 single port range as a single port.

jtk54:
L4/L7 port ranges were represented previously as a "range" of the same port, e.g. 8080-8080. This PR expresses that string as a single port. Additionally fixes the `instancePort` detail for Https to be the named port 'http'. @duftler @danielpeach please review.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController.groovy
@@ -104,17 +104,21 @@ class GoogleLoadBalancerController {
       }
     }
 
+    def isHttp = view.loadBalancerType == GoogleLoadBalancerType.HTTP
     [new GoogleLoadBalancerDetails(loadBalancerName: view.name,
+                                   loadBalancerType: view.loadBalancerType,
                                    createdTime: view.createdTime,
                                    dnsname: view.ipAddress,
                                    ipAddress: view.ipAddress,
                                    healthCheck: view.healthCheck ?: null,
                                    backendServiceHealthChecks: backendServiceHealthChecks ?: null,
                                    listenerDescriptions: [[
-                                       listener: new ListenerDescription(instancePort: view.portRange,
-                                                                         loadBalancerPort: view.portRange,
-                                                                         instanceProtocol: view.ipProtocol,
-                                                                         protocol: view.ipProtocol)
+                                       listener: new ListenerDescription(
+                                         instancePort: isHttp ? 'http' : Utils.derivePortOrPortRange(view.portRange), // Https forwards traffic to a 'named' port.
+                                         loadBalancerPort: Utils.derivePortOrPortRange(view.portRange),
+                                         instanceProtocol: view.ipProtocol,
+                                         protocol: view.ipProtocol
+                                       )
                                    ]])]
   }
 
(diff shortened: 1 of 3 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/callbacks/Utils.groovy
```diff
@@ -36,6 +36,26 @@ class Utils {
...
+   * Return a single port string if a port range refers to a single port (e.g. 80-80).
+   *
+   * @param portRange - Port range to parse.
+   * @return - Single port if the ports in the port range are the same.
+   */
+  static String getPortFromPortRange(String portRange) {
+    if (!portRange || !portRange.contains('-')) {
+      return portRange
+    }
+    def dashIndex = portRange.indexOf('-')
```
duftler: Can use `split()` or `tokenize()` here instead of doing it by hand.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/callbacks/Utils.groovy
```diff
@@ -36,6 +36,26 @@ class Utils {
...
+    def beginRange = portRange.substring(0, dashIndex)
+    def endRange = portRange.substring(dashIndex + 1)
+
+    if (beginRange != endRange) {
```
duftler: Could use `if ... else` here or ternary operator.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/callbacks/Utils.groovy
```diff
@@ -36,6 +36,26 @@ class Utils {
     }
   }
 
+  /**
+   * Return a single port string if a port range refers to a single port (e.g. 80-80).
+   *
+   * @param portRange - Port range to parse.
+   * @return - Single port if the ports in the port range are the same.
+   */
+  static String getPortFromPortRange(String portRange) {
```
duftler: This might be more explicitly named something like `derivePortOrPortRangeDescription`.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleLoadBalancerController.groovy
```diff
@@ -104,17 +104,21 @@ class GoogleLoadBalancerController {
...
                                    ipAddress: view.ipAddress,
                                    healthCheck: view.healthCheck ?: null,
                                    backendServiceHealthChecks: backendServiceHealthChecks ?: null,
                                    listenerDescriptions: [[
-                                       listener: new ListenerDescription(instancePort: view.portRange,
-                                                                         loadBalancerPort: view.portRange,
-                                                                         instanceProtocol: view.ipProtocol,
-                                                                         protocol: view.ipProtocol)
+                                       listener: new ListenerDescription(
+                                         instancePort: isHttp ? 'http' : Utils.getPortFromPortRange(view.portRange), // Https forwards traffic to a 'named' port.
```
duftler: Did you test that this has no negative implications on the Edit Load Balancer capability in the ui?

jtk54: The listener ports are specified in the edit modal (and actually are pretty fixed), so there is no 'reading' from the old description. As far as I can tell, there's no negative effects in the UI - I ran a few successful operations.

duftler: Please test with L4 as well.

jtk54: Tested out L4, same thing applies.

PR #1063, opened 2016-09-26
Support multiple Eureka hosts using the same AWS account ID

robertsigler:
Spinnaker allows you to configure multiple credentials/accounts in clouddriver that correspond to the same AWS account. However, if you use Eureka discovery to determine their health, instances will only show up as healthy in one credential/account because clouddriver uses only the AWS account ID to determine the credential name. This pull request will try to use the AWS account ID and eureka host URL in order to determine the credential name when possible--defaulting to simply matching AWS account ID when no such AWS & Eureka match can be found. This will allow users to successfully deploy to multiple 'environments' or accounts even when they only have one AWS account ID.

**Update: this merge request now uses attempts to match AWS account ID and account name, instead of eureka host URL, when retrieving the credential name.**

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/AwsProvider.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/AwsProvider.groovy
@@ -111,19 +111,30 @@ class AwsProvider extends AgentSchedulerAware implements SearchableProvider, Eur
 
   @Override
   String getInstanceKey(Map<String, Object> attributes, String region) {
-    if (getCredentialName(attributes.accountId) == null) {
+    def credentialName = getCredentialName(attributes.accountId, attributes.allowMultipleEurekaPerAccount, attributes.eurekaAccountName)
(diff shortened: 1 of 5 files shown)


tomaslin:
I don't like that we're matching eureka host URL given than the eureka configuration has a name field - `eureka.provider.accounts.name`. We should use that to match the configuration instead of using the name field. 

This also breaks the current titus integration so changes would have to be made in the titus cloud provider as well.

robertsigler:
I've updated my branch to use the `eureka.provider.accounts.name` for matching.

Can you clarify how this breaks the titus integration? I don't use it so I'm still trying to work out how my changes will affect titus.

tomaslin:
By broken in titus I mean that this feature won't work there. Maybe it's ok since this situation doesn't apply. 

It would be nice if this was explicitly featured flagged under `allowMultipleEurekaPerAccount` in configuration.

robertsigler:
Okay, I've updated the merge request such that the old behavior is used unless `eureka.providers.allowMultipleEurekaPerAccount` is true. Let me know if there are any further improvements you would like to see.

tomaslin:
LGTM

PR #1064, opened 2016-09-26
(kubernetes) register non-ready pods as down to stall pipelines

lwander:
@duftler 

I was incorrectly considering non-ready pods as "Starting" causing pipelines to succeed when health checks were failing.


duftler:
LGTM.

PR #1068, opened 2016-09-27
provider/google: Cached urlMapName for L7 LBs.

jtk54:
We want to be able to group the L7 LBs by `urlMapName` in the UI, so we cache it with this PR. I'll follow up soon with another PR that requires a `urlMapName` in L7 upsert. @duftler @danielpeach please review.


duftler:
LGTM.

PR #1070, opened 2016-09-27
provider/google: Support urlMapName in L7 upsert description.

jtk54:
To wire up more than one listener to a urlMap, we need an explicit `urlMapName` in the L7 upsert description. I tested this PR with create, upsert with one listener, and then another upsert that adds a listener to the same urlMap. We'll need to change L7 delete to deal with all the possible cases this enables. @duftler or @lwander or @danielpeach please review.


danielpeach:
LGTM

PR #1071, opened 2016-09-27
Secures the /search endpoint by inspecting the results for 'applicati…

ttomsu:
…on' and 'account' attributes.

@duftler @jtk54 PTAL

Closes https://github.com/spinnaker/fiat/issues/65


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/SearchController.groovy
```diff
@@ -42,6 +43,7 @@ class SearchController {
    * @param filters (optional) a map of key-value pairs to further filter the keys
    * @return a list {@link SearchResultSet)s
    */
+  @PreAuthorize("@fiatPermissionEvaluator.storeWholePermission()")
```
jtk54: Why do this on both the endpoint and the `search` function? Are both needed?

ttomsu: You're right. Removed.

PR #1074, opened 2016-09-28
Ensure that the regular caching cycle sets a ttl for OnDemand records

ajordens:
Previously the OnDemand records were loaded, incremented and re-saved w/o a TTL.

### DIFF
--- a/cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/DefaultCacheData.java
+++ b/cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/DefaultCacheData.java
@@ -16,9 +16,11 @@
 
 package com.netflix.spinnaker.cats.cache;
 
+import java.time.Clock;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Map;
+import java.util.HashMap;
+import java.util.Optional;
 
 /**
  * An immutable CacheData.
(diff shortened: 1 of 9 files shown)


ajordens:
Have a slight modification that should make this a little more generic.

cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/DefaultCacheData.java
```diff
@@ -34,10 +36,27 @@ public DefaultCacheData(String id, Map<String, Object> attributes, Map<String, C
...
+    public DefaultCacheData(String id, int ttlSeconds, Map<String, Object> attributes, Map<String, Collection<String>> relationships, Clock clock) {
+        // ensure attributes is non-null and mutable given that `cacheExpiry` will be added
+        attributes = attributes == null ? new HashMap<>() : new HashMap<>(attributes);
+
         this.id = id;
-        this.ttlSeconds = ttlSeconds;
         this.attributes = attributes;
         this.relationships = relationships;
+
+        if (ttlSeconds > 0) {
```
ajordens: Rather than modifying `RedisCache` such that it could fetch `ttl` for each key (performance? complication?) ... I opt'd to store it as an additional attribute and make `DefaultCacheData` smart enough to calculate the `ttl` on-demand.

ajordens:
PTAL @cfieber 

This PR aims to maintain the ttl for an OnDemand key despite it being loaded/stored during the cache refresh cycle.

https://github.com/spinnaker/clouddriver/blob/master/clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/provider/agent/ClusterCachingAgent.groovy#L262

Currently the `ttl` is lost when ^^ occurs. When an OnDemand key has no ttl, there is possibility that it won't ever be evicted and eventually fill up the OnDemand collection.

PR #1075, opened 2016-09-28
provider/google: Added urlMapName to L7 LB view.

jtk54:
Needed to add `urlMapName` to L7 LB view for frontend to receive it through Gate. @danielpeach please review.


danielpeach:
LGTM

PR #1079, opened 2016-09-29
provider/google: Added support for L7 session affinity cookie ttl.

jtk54:
We added support earlier for setting session affinity for an L7 backend service, but not the cookie TTL in the case of `GENERATED_COOKIE` session affinity. @danielpeach @duftler please review.


danielpeach:
LGTM

PR #1080, opened 2016-09-29
Adapt cloud driver to handle front50 new project API

jeyrschabu:
- Added a temp try catch to handle the search to return Set<Map>


jeyrschabu:
@ajordens PTAL. Another PR coming up for the search fix

ajordens:
LGTM

PR #1081, opened 2016-09-29
Failure of a single SearchProvider should not fail all searches

jeyrschabu:
- Added an exception handler for a provider search that returns an empty result set in case of a failure


jeyrschabu:
@ajordens PTAL

ajordens:
LGTM

ajordens:
Needs a rebase tho

PR #1083, opened 2016-09-30
provider/google: Added ability to remove listeners in L7 upsert.

jtk54:
We added the ability to add multiple listeners to one url map earlier this week - this is the symmetric 'delete' operation for listeners in upsert. This also adds a utility function to delete global listeners. I'm planning on using that in a future PR changing L7 delete. @duftler @danielpeach please review.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
@@ -892,6 +892,60 @@ class GCEUtil {
     return loadBalancerNames
   }
 
+  /**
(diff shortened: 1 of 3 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -892,6 +892,59 @@ class GCEUtil {
     return loadBalancerNames
   }
 
+  /**
+   * Deletes an L7 global listener, i.e. a global forwarding rule and its target proxy.
+   * @param compute
+   * @param forwardingRuleName - Name of global forwarding rule to delete (along with its target proxy).
```
duftler: Missing the `project` parameter.

PR #1095, opened 2016-10-03
provider/google: Guard against reusing listeners for different url maps.

jtk54:
@duftler @danielpeach please review. If we allow the user to add multiple listeners and specify the names manually, we need to guard against reusing a listener to try and point to two different url maps. There should be a symmetric frontend change preventing this type of situation also.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/loadbalancer/UpsertGoogleHttpLoadBalancerAtomicOperation.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/loadbalancer/UpsertGoogleHttpLoadBalancerAtomicOperation.groovy
@@ -163,6 +163,12 @@ class UpsertGoogleHttpLoadBalancerAtomicOperation extends UpsertGoogleLoadBalanc
           break
       }
       targetProxyExists = existingProxy as Boolean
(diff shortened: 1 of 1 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/loadbalancer/UpsertGoogleHttpLoadBalancerAtomicOperation.groovy
```diff
@@ -163,6 +163,9 @@ class UpsertGoogleHttpLoadBalancerAtomicOperation extends UpsertGoogleLoadBalanc
           break
       }
       targetProxyExists = existingProxy as Boolean
+      if (existingProxy && GCEUtil.getLocalName(existingProxy.getUrlMap()) != description.urlMapName) {
+        throw new IllegalStateException("Listener with name ${existingRule.getName()} already exists and points to a different url map than ${description.urlMapName}")
```
danielpeach: It might be helpful to specify the name of the url map that the listener is pointing to .

jtk54: Done.

danielpeach:
LGTM

PR #1098, opened 2016-10-03
(openstack) add healthClass to fix deploys which only consider platform health

drmaas:
This will allow enable/disable/create/destroy to pass health checks and not hang for this scenario.


drmaas:
@spinnaker/reviewers

jcwest:
LGTM. Same changes I made on my system that resolved the issue.

PR #1099, opened 2016-10-03
provider/google: Update url map every L7 upsert operation.

jtk54:
We previously checked if a url map needed updated by checking host rules + path matchers, but not if only the host rules changed with no path matchers involved. @duftler @danielpeach please review.


jtk54:
Don't merge yet, I'm getting rid of this function. It's _much_ simpler to just run the url map update call each time rather than try this complex logic and get it wrong if we need to update. It also adds a negligible amount of time to this operation, which is already the fastest L7 operation.

jtk54:
Ping @duftler and @danielpeach -- I tested and it's working fine.

PR #1105, opened 2016-10-04
don't reconcile classic link until autoscaler has done so first.

cfieber:
Adds a delay before reconciling an instance to allow the autoscaler to apply any groups from the launch configuration before we attempt to add any from our conventions.


ajordens:
LGTM

PR #1108, opened 2016-10-05
(Openstack) Do not base64 encode userdata.

varikin:
Somewhere between Mitaka and cloud-init on the base image doesn't like
the base64 encoded userdata. It may be a configuration issue with
Mitaka or cloud-init (tested with v0.7.3) in the base image. We may need
to rethink this if there are problems in other environments.

Fixes spinnaker/spinnaker#1193


varikin:
@spinnaker/reviewers PTAL

PR #1110, opened 2016-10-05
Removes all LocalRedisCheck tests that run against any Redis server h…

ttomsu:
…osted on the default port. Instead, uses EmbeddedRedis.

Fixes https://github.com/spinnaker/spinnaker/issues/1202

@duftler @ajordens @cfieber @tomaslin PTAL


ttomsu:
@jtk54, @lwander  too

jtk54:
LGTM

lwander:
LGTM

PR #1114, opened 2016-10-06
Periodically cleanup the onDemand:members sets if the referenced keys…

ajordens:
… no longer exist

- keys are being TTL'd out so this just keeps the onDemand:members  set in-sync
- runs as part of a new core provider

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/agent/CleanupPendingOnDemandCachesAgent.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/agent/CleanupPendingOnDemandCachesAgent.java
@@ -1,0 +1,122 @@
+package com.netflix.spinnaker.clouddriver.core.agent;
+
+import com.netflix.spinnaker.cats.agent.RunnableAgent;
+import com.netflix.spinnaker.cats.module.CatsModule;
+import com.netflix.spinnaker.cats.provider.Provider;
+import com.netflix.spinnaker.cats.redis.JedisSource;
+import com.netflix.spinnaker.clouddriver.cache.CustomScheduledAgent;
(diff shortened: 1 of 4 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/agent/CleanupPendingOnDemandCachesAgent.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  @Override
+  public String getProviderName() {
+    return CoreProvider.PROVIDER_NAME;
+  }
+
+  @Override
+  public void run() {
+    getCatsModule().getProviderRegistry().getProviders().forEach(provider -> {
```
ajordens: Specifically this aims to cleanup the onDemand:members sets for Aws/AwsInfrastructure/Titus.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/agent/CleanupPendingOnDemandCachesAgent.java
```diff
@@ -0,0 +1,117 @@
...
+
+  public long getPollIntervalMillis() {
+    return pollIntervalMillis;
+  }
+
+  public long getTimeoutMillis() {
+    return timeoutMillis;
+  }
+
+  private Set<String> scanMembers(String setKey) {
```
ajordens: This was lifted from `RedisCache`.

ajordens:
PTAL @cfieber -- I've had this sitting locally for a little while but it provides a bit of house keeping around the `onDemand:members` sets now that the underlying keys TTL out after 10 minutes.

Again this only handles scenarios where the normal evict after edda read process doesn't kick in.

cfieber:
LGTM

PR #1117, opened 2016-10-06
Remove fallback for Project API

jeyrschabu:
- removed the try and catch to handle the new project API.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/services/Front50Service.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/services/Front50Service.groovy
@@ -16,8 +16,6 @@
 
 package com.netflix.spinnaker.clouddriver.core.services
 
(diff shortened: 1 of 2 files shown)


jeyrschabu:
This is to close the loop on the project API. @ajordens PTAL

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/services/Front50Service.groovy
```diff
@@ -35,9 +33,6 @@ interface Front50Service {
   Map getProject(@Path('project') String project)
 
   @GET('/v2/projects/search')
```
ajordens: Maybe I'll have Rob (when he starts next week) support searching via `/v2/projects` with query parameters vs. `/v2/projects/search`. That would make it a _little_ more consistent with how application search is done.

ajordens:
Needs a quick rebase

PR #1121, opened 2016-10-10
provider/google: Implemented deploy support for ILBs.

jtk54:
@duftler please review. Implements support for associating created and cloned server groups with internal load balancers. This reports health correctly, but doesn't report the server groups as enabled correctly -- I'll follow up shortly with another PR addressing that as well as load balancer associations in the server group caching agents.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
@@ -234,12 +234,16 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
     // Update the instance metadata for ILBs and queue up region backend service calls.
     List<BackendService> regionBackendServicesToUpdate = []
     if (internalLoadBalancers) {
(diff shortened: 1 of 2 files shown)


jtk54:
@danielpeach and/or @lwander.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -226,7 +231,38 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
...
+
+      ilbNames.each { String ilbName ->
+        if (!(ilbName in existingRegionalLbs))  {
+          existingRegionalLbs << ilbName
+        }
+      }
+      instanceMetadata[GoogleServerGroup.View.REGIONAL_LOAD_BALANCER_NAMES] = existingRegionalLbs.join(",")
+
+      ilbServices.each { String backendServiceName ->
+        BackendService backendService = compute.regionBackendServices().get(project, region, backendServiceName).execute()
```
lwander: Should this be wrapped in retry-logic?

jtk54: `GET`s aren't as prone to retry-able failures as `UPDATE`s are. The Http(s) LBs don't use retries here, so I won't here until it becomes an issue.

PR #1125, opened 2016-10-10
Cache public AMIs longer than private

AMeng:
This is meant to address https://github.com/spinnaker/spinnaker/issues/1099
There was work done in https://github.com/spinnaker/clouddriver/pull/1093 but it was closed.

Cache private and public AMIs separately. This allows us to cache public images for much longer.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ImageCachingAgent.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ImageCachingAgent.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.clouddriver.aws.provider.agent
 
+import com.amazonaws.services.ec2.model.DescribeImagesRequest
 import com.amazonaws.services.ec2.model.Image
 import com.fasterxml.jackson.core.type.TypeReference
 import com.fasterxml.jackson.databind.ObjectMapper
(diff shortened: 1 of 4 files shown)


gbougeard:
Is it possible to fully disable public images caching?

AMeng:
It is definitely possible, and I would be fine with that approach. We'll have to wait and see what the maintainers think is appropriate.

gbougeard:
ping @cfieber

AMeng:
@spinnaker/reviewers

cfieber:
this makes sense to me but we'll need to make a couple modifications to allow it to work if edda is backing the API request (it ignores the filter parameters).. will try to take a look at it today

AMeng:
Any updates on this? I'm glad to help wherever I can.

cfieber:
right now if an account is eddaEnabled=true (we do caching reads from edda instead of direct to AWS) the `describeFoo` requests will return the entire dataset ignoring any filters.

Instead of trying to fix that at our JDK dynamic proxy layer to pass filter parameters to edda, I'd be ok if this PR was just updated to do a post-filtering of the describe results to apply the same filtering if the account.isEddaEnabled 

I think if that was in, we would be good to go on this PR

gbougeard:
@cfieber Aren't edda and clouddriver redundant? I mean, don't they do the same thing?
In the spinnaker docs there's no mention of edda so maybe it would be nice if there was a Best practices/architectures part explaining if edda is needed or not?

AMeng:
@cfieber Added manual filtering for Edda. Added some tests as well.

AMeng:
Any objections to merging this?

cfieber:
testing this out now with our configs / edda, sorry for the delay

cfieber:
LGTM - if you rebase this on master and I will merge

PR #1130, opened 2016-10-12
Add the assume role'd `accountId` to an AmazonServiceException

ajordens:
This supports using the `accountId` in metrics generated by the
`InstrumentedRetryPolicy` in `kork`.


ajordens:
@cfieber - I'm trying to access to the `accountId` from the RetryPolicy such that it could be included in the metrics that are being generated from kork.

I didn't come across an easy way of doing this, do you know of a more suitable alternative?

cfieber:
ship it

PR #1132, opened 2016-10-12
Docker Registry - make docker registry informative rather than author…

tomaslin:
…ative to prevent images from being deleted


lwander:
LGTM

PR #1135, opened 2016-10-12
(Openstack) Honor the insecure flag for all connections.

varikin:
Was only skipping SSL verification on authentication. Now doing so on
all calls.


varikin:
@spinnaker/reviewers PTAL

PR #1142, opened 2016-10-13
(openstack) allow floating ip allocation in auto scaling groups

drmaas:
Also fixes the naming for the member resource template.

@spinnaker/reviewers

### DIFF
--- a/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/description/servergroup/ServerGroupParameters.groovy
+++ b/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/description/servergroup/ServerGroupParameters.groovy
@@ -51,11 +51,12 @@ class ServerGroupParameters {
   String sourceUserData
   Map<String, String> tags
   String resourceFilename
+  String floatingNetworkId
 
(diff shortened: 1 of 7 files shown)


lwander:
Small nitpick, could you make the commit titles more descriptive and keep the issue numbers in the message body?

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/description/servergroup/ServerGroupParameters.groovy
```diff
@@ -80,6 +81,10 @@ class ServerGroupParameters {
       user_data            : rawUserData ?: null,
       resource_filename    : resourceFilename ?: ServerGroupConstants.SUBTEMPLATE_FILE
     ]
+    if (floatingNetworkId) {
+      params << [floating_network_id: floatingNetworkId]
```
lwander: Why not just add this inside the map the same way you do for `rawUserData`?

drmaas: I found that if the key is present, openstack rejects the stack creation due to an unexpected (even if null) parameter.

PR #1152, opened 2016-10-17
provider/google: Fixed ILB ports in LB details.

jtk54:
@duftler @danielpeach please review. ILBs don't use `portRange` like everyone else does, they have a list of port (strings) as `ports`. I changed the details to serialize the ports as a comma-delimited list for ILBs, since Google Cloud Console reports the port list this way as well.


danielpeach:
LGTM

PR #1153, opened 2016-10-18
(docker) create builder for credentials object

lwander:
I was originally going to allow docker credentials to be initialized from the dockerconfigFile (see it's a new property) in this PR, but realized the current docker credentials initializer is a bit of a mess.

@duftler @tomaslin PTAL

### DIFF
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/config/DockerRegistryConfigurationProperties.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/config/DockerRegistryConfigurationProperties.groovy
@@ -29,8 +29,10 @@ class DockerRegistryConfigurationProperties {
     String username
     // Docker registry password.
     String password
-    // File containing docker password.
+    // File containing docker password in plaintext.
     String passwordFile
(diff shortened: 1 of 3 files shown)


clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/config/DockerRegistryConfigurationProperties.groovy
```diff
@@ -29,8 +29,10 @@ class DockerRegistryConfigurationProperties {
     String username
     // Docker registry password.
     String password
-    // File containing docker password.
+    // File containing docker password in plaintext.
     String passwordFile
+    // File containing docker's auth config (managed by docker-cli). Typically `~/.docker/config.json`.
+    String dockerconfigFile
```
duftler: Is this `dockerconfigFile` instead of `dockerConfigFile` to be consistent with how it is referred to elsewhere?

lwander: I wanted to be consistent with `kubeconfig` which is how it's capitalized in all the docs. I'll make sure `dockerconfig` is consistent here.

PR #1155, opened 2016-10-19
provider/google: Implemented new health check caching agent.

jtk54:
Provides support for '/search' as well. The new agent caches all types of GCP load balancers. I left the old agent for backwards-compatibility purposes, and will remove when the frontend cuts over. @duftler @danielpeach please review.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/cache/Keys.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/cache/Keys.groovy
@@ -26,6 +26,7 @@ class Keys {
     APPLICATIONS,
     BACKEND_SERVICES,
     CLUSTERS,
+    HEALTH_CHECKS,
     HTTP_HEALTH_CHECKS,
     IMAGES,
     INSTANCES,
@@ -85,10 +86,16 @@ class Keys {
             detail     : names.detail
         ]
         break
-      case Namespace.HTTP_HEALTH_CHECKS.ns:
+      case Namespace.HEALTH_CHECKS.ns:
         result << [
             account: parts[2],
             name   : parts[3],
+        ]
+        break
+      case Namespace.HTTP_HEALTH_CHECKS.ns:
+        result << [
+          account: parts[2],
+          name   : parts[3],
         ]
         break
       case Namespace.IMAGES.ns:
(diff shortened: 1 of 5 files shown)


danielpeach:
LGTM

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/GoogleInfrastructureProvider.groovy
```diff
@@ -71,18 +74,19 @@ class GoogleInfrastructureProvider extends AgentSchedulerAware implements Search
       return result + [
           healthCheckLink: backendService.attributes.healthCheckLink as String,
           sessionAffinity: backendService.attributes.sessionAffinity as String,
-          affinityCookieTtlSec: backendService.attributes.affinityCookieTtlSec as String
+          affinityCookieTtlSec: backendService.attributes?.affinityCookieTtlSec as String,
```
duftler: Would have already NPE'd on the previous line; can't see reason to use `?.` here and below.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleHealthCheckCachingAgent.groovy
```diff
@@ -0,0 +1,146 @@
...
+      def newHC = new GoogleHealthCheck(
+        name: hc.getName(),
+        checkIntervalSec: hc.getCheckIntervalSec(),
+        timeoutSec: hc.getTimeoutSec(),
+        healthyThreshold: hc.getHealthyThreshold(),
+        unhealthyThreshold: hc.getUnhealthyThreshold()
+      )
+
+      switch(hc.getType()) {
+        case 'HTTP':
```
duftler: Please add some comments in here about why you have to query for http health checks both here and above.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/cache/Keys.groovy
```diff
@@ -85,12 +86,18 @@ class Keys {
             detail     : names.detail
         ]
         break
-      case Namespace.HTTP_HEALTH_CHECKS.ns:
+      case Namespace.HEALTH_CHECKS.ns:
```
duftler: Seems like it might be more future-proof to include the 'type' of the health check as a key segment to avoid potential naming collisions.

PR #1157, opened 2016-10-19
provider/google: Adds session affinities for ILBs.

jtk54:
ILBs have more options for session affinities in backend services than external load balancers, so I added them. @duftler @danielpeach please review.


danielpeach:
LGTM

PR #1167, opened 2016-10-24
provider/google: Prevent name/key collisions in health check agent.

jtk54:
Also provide context which health check 'family' a health check belongs to. There are three health check 'families': `compute#healthCheck`, `compute#httpHealthCheck`, and `compute#httpsHealthCheck`. Each has its own set of endpoints. @duftler @danielpeach please review -- also Dan we should think about how to cut the frontend over to use this new search provider.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/cache/Keys.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/cache/Keys.groovy
@@ -89,7 +89,8 @@ class Keys {
       case Namespace.HEALTH_CHECKS.ns:
         result << [
             account: parts[2],
-            name   : parts[3],
+            kind   : parts[3],
+            name   : parts[4],
         ]
         break
       case Namespace.HTTP_HEALTH_CHECKS.ns:
@@ -194,8 +195,9 @@ class Keys {
   }
 
   static String getHealthCheckKey(String account,
+                                  String kind,
                                   String healthCheckName) {
-    "$GoogleCloudProvider.GCE:${Namespace.HEALTH_CHECKS}:${account}:${healthCheckName}"
+    "$GoogleCloudProvider.GCE:${Namespace.HEALTH_CHECKS}:${account}:${kind}:${healthCheckName}"
   }
 
   static String getHttpHealthCheckKey(String account,
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleHealthCheck.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleHealthCheck.groovy
@@ -21,7 +21,6 @@ import groovy.transform.Canonical
 
 @Canonical
 class GoogleHealthCheck {
-
   String name
   String requestPath
   int port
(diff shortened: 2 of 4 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/cache/Keys.groovy
```diff
@@ -88,8 +88,9 @@ class Keys {
         break
       case Namespace.HEALTH_CHECKS.ns:
         result << [
-            account: parts[2],
-            name   : parts[3],
+            account         : parts[2],
+            loadBalancerType: parts[3],
```
duftler: Isn't this actually the 'kind', and not the 'type' or 'endpoint'?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/cache/Keys.groovy
```diff
@@ -194,8 +195,9 @@ class Keys {
   }
 
   static String getHealthCheckKey(String account,
+                                  String loadBalancerEndpoint,
                                   String healthCheckName) {
-    "$GoogleCloudProvider.GCE:${Namespace.HEALTH_CHECKS}:${account}:${healthCheckName}"
+    "$GoogleCloudProvider.GCE:${Namespace.HEALTH_CHECKS}:${account}:${loadBalancerEndpoint}:${healthCheckName}"
```
duftler: Same question here.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/cache/Keys.groovy
duftler: 'Kind' without the `compute#` prefix...

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleHealthCheck.groovy
```diff
@@ -34,7 +37,32 @@ class GoogleHealthCheck {
   int healthyThreshold
 
   /**
-   * Name of the GCP certificate, if HTTPS.
+   * Specifies the GCP endpoint 'family' this health check originated from.
+   *
+   * There are currently three different sets of health check endpoints:
```
duftler: That seems perfectly simple enough.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/cache/Keys.groovy
duftler: As discussed offline, let's just drop the `compute#` prefix altogether and unify `kind` and `endpoint`.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/GoogleInfrastructureProvider.groovy
```diff
@@ -74,7 +74,7 @@ class GoogleInfrastructureProvider extends AgentSchedulerAware implements Search
       return result + [
           healthCheckLink: backendService.attributes.healthCheckLink as String,
           sessionAffinity: backendService.attributes.sessionAffinity as String,
-          affinityCookieTtlSec: backendService.attributes?.affinityCookieTtlSec as String,
+          affinityCookieTtlSec: backendService.attributes.affinityCookieTtlSec as String,
           region: GCEUtil.getLocalName(backendService.attributes?.region as String)
```
duftler: Don't need `?.` on this one either.

PR #1168, opened 2016-10-24
provider/google: Prevent NPE in backend service caching agent.

jtk54:
@duftler please review.


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleBackendServiceCachingAgent.groovy
```diff
@@ -55,7 +55,7 @@ class GoogleBackendServiceCachingAgent extends AbstractGoogleCachingAgent {
 
   List<BackendService> loadBackendServices() {
     List<BackendService> ret = []
-    ret.addAll(compute.backendServices().list(project).execute().items as List)
+    ret.addAll(compute.backendServices().list(project).execute().items as List ?: [])
```
duftler: Feels weird to have `.addAll([])`. Can you capture the results and call `addAll()` only if there is something to add?

jtk54: Done, same as the one below it.

PR #1170, opened 2016-10-26
Simplify spring Configuration classes.

cfieber:
This will be broken pending release of spinnaker/kork#62

Eliminates use of `@Value` in `@Configuration` classes.. creates `@ConfigurationProperties` objects instead.
Eliminates most use of `@Autowired` into `@Configuration` classes, instead passes dependencies around the `@Bean` method parameters

Other cleanups include:
- Consolodate duplication of user agent (via reading the gradle produced META-INF file for version) into `clouddriver-core`
- Clean up titus bootstrapping to eliminate some nasty circular dependencies that were being resolved via `@Value` and a SpEL expression
- Removes `ConversionService` bean now that properties are injected via `@ConfigurationProperties`
- Pushed some configuration upstream to kork around retrofit logging and okhttp connection pools

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ application.yml
 *.hprof
 README.html
 *~
+es-tmp/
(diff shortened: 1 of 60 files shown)


robzienert:
Yes please!

PR #1174, opened 2016-10-26
Errorconfig handles XSS vulnerabilites by removing requested paths

lwander:
Also cleaned out some unused imports 

@duftler PTAL


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy
```diff
@@ -48,6 +47,7 @@ import java.security.Security
 @Configuration
 @Import([
   WebConfig,
+  ErrorConfiguration,
```
ttomsu: Hmm... the @ComponentScan below should have picked this up.

PR #1179, opened 2016-10-28
provider/google: Separate Redis keyspaces for backendServices and reg…

jtk54:
…ionBackendServices.

Also added a 'kind' attribute to GoogleBackendService to distinguish between the API families. @duftler @danielpeach please review.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/loadbalancing/GoogleBackendService.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/loadbalancing/GoogleBackendService.groovy
@@ -23,9 +23,30 @@ import groovy.transform.EqualsAndHashCode
 @EqualsAndHashCode(excludes="backends")
 class GoogleBackendService {
   String name
+  String region
+
+  /**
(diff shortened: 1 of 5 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/loadbalancing/GoogleBackendService.groovy
```diff
@@ -23,9 +23,30 @@ import groovy.transform.EqualsAndHashCode
...
   GoogleHealthCheck healthCheck
   List<GoogleLoadBalancedBackend> backends
   GoogleSessionAffinity sessionAffinity
   Integer affinityCookieTtlSec
   GoogleLoadBalancingScheme loadBalancingScheme
+
+  // Note: This enum has non-standard style constants because we use these constants as strings directly
+  // in the redis cache keys for backend services, where we want to avoid underscores and camelcase is the norm.
+  static enum BackendServiceKind {
+    backendService,
```
duftler: Would these be better-named as `globalBackendService` and `regionalBackendService`? I know the platform api uses `regionBackendService`, but it still sounds funny.

PR #1180, opened 2016-10-28
(docker) reload passwordfile everytime credentials are loaded

lwander:
This will allow us to regenerate docker password files using `gcloud auth print-access-token` every 30 mins rather than require the IAM API to be enabled to create a service account key. 

@duftler PTAL

@skim1420 FYI


clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerTokenService.groovy
```diff
@@ -27,24 +26,53 @@ import retrofit.http.Path
...
+    if (!(username || password || passwordFile)) {
+      return null
+    }
+
+    def resolvedPassword = null
+
+    if (password) {
+      resolvedPassword = password
+    } else if (passwordFile) {
+      resolvedPassword = new BufferedReader(new FileReader(passwordFile))
```
duftler: Is that all you need here? You don't need to call `getText()` or something similar on the `BufferedReader`?

clouddriver-docker/src/test/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerTokenServiceSpec.groovy
```diff
@@ -33,7 +33,7 @@ class DockerBearerTokenServiceSpec extends Specification {
   DockerBearerTokenService tokenService
```
duftler: Probably wouldn't hurt to have a test that verifies the behavior of the construct that expects a `passwordFile`.

PR #1181, opened 2016-10-28
Preliminary tag support (WIP)

ajordens:


ajordens:
Outstanding:
- `partial` support
- re-indexing support (source from `front50` and re-index in `elasticsearch`)
- tag searching

gbougeard:
Does it mean elasticsearch will be mandatory/embedded ?

ajordens:
It's possible ES will be an optional dependency, if it's not available than the general purpose tagging operations will be disabled.

We have a number of use cases that ES (and `front50` for durable/permanent storage) would seem to fit with and will be doing some experimentation over the next couple weeks.

ajordens:
Supports queries like:

- `https://localhost:8081/tags?tag:owner=foobar`
- `https://localhost:8081/tags/aws:cluster:front50-main:mgmt:*`
- `https://localhost:8081/tags?cloudProvider=aws&entityType=cluster`
- `https://localhost:8081/tags?idPrefix=aws:cluster:front50*`

clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/EntityRefIdBuilder.java
```diff
@@ -0,0 +1,47 @@
...
+
+package com.netflix.spinnaker.clouddriver.elasticsearch;
+
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class EntityRefIdBuilder {
+  public static EntityRefId buildId(String cloudProvider, String entityType, String entityId, String account, String region) {
```
ajordens: After some back and forth ... decided to go with an explicit `id` generation scheme of `{{cloudProvider}}:{{entityType}}:{{entityId}}:{{account}}:{{region}}` instead of trying to leverage the existing cloudprovider-specific `Keys` generators.

I didn't see anything significant in the existing the existing `Keys` implementations that wasn't compatible w/ `cloudProvider:entityType:entityId:account:region`. 

`account` and `region` can be optional.

clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
```diff
@@ -0,0 +1,167 @@
...
+
+  private BoolQueryBuilder applyTagsToBuilder(BoolQueryBuilder queryBuilder, Map<String, Object> tags) {
+    if (tags == null) {
+      return queryBuilder;
+    }
+
+    for (Map.Entry<String, Object> entry : tags.entrySet()) {
+      // restrict to specific tags (optional)
+      if (entry.getValue().equals("*")) {
+        queryBuilder = queryBuilder.must(QueryBuilders.existsQuery("tags." + entry.getKey()));
```
ajordens: Treating `*` as convention for "don't care what the tag value is, just that the tag exists".

ajordens:
PTAL @spinnaker/netflix-reviewers (@cfieber especially)

FYI @spinnaker/google-reviewers 

Probably a bit light on documentation but for now I'm interested in getting this landed. I expect the only consumer (_for the next few weeks_) will be a Netflix-specific auth extension (tagging iam roles).

Expectation is that `elasticsearch` and the tagging operation will be off by default, and they'll be no impact to startup with current configuration.

There are few obvious next steps needed:
- support for `DeleteEntityTags` (right now I just `UpsertEntityTags` with an empty tag set)
- support for partial updates (_requires groovy scripting to be enabled on the elasticsearch cluster_)
- support for re-indexing (ie. delete the elasticsearch index and re-index off of what's stored in `front50`)
- performance testing

robzienert:
LTGM

PR #1183, opened 2016-10-31
(kubernetes) ~20x faster application load

lwander:
The translation of the cache entries to server group objects went from ~11 seconds to ~.5 seconds for a single application with around 60 server groups.

This was done by batching up the request to load all instances associated with every server group in a given cluster, rather than make one call to the cache per server group.

@duftler  PTAL


duftler:
Nice.
Probably worth capturing the before/after results and diff'ing to ensure no change.

lwander:
Diffs are identical

PR #1184, opened 2016-10-31
(google) ~6x faster application loads

lwander:
By frontloading most of the instance & security group loading, I was able to significantly cut down application load times. Here are the benchmarks (where `load(N) = s` means `N` server groups, `s` milliseconds)

**Before**
```
load(17) = 8240
load(17) = 8619
load(17) = 8412
load(17) = 8428
```

**After**
```
load(17) = 1310
load(17) = 1211
load(17) = 1214
load(17) = 1198
```

I've also captured diffs before & after with multiple server groups, load balancers & security groups w/o any differences.


lwander:
@duftler

PR #1193, opened 2016-11-02
(docker) warn against trailing & leading whitespace in password

lwander:
https://github.com/spinnaker/spinnaker/issues/1265

### DIFF
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerTokenService.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerTokenService.groovy
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.clouddriver.docker.registry.api.v2.auth
 
 import com.netflix.spinnaker.clouddriver.docker.registry.api.v2.exception.DockerRegistryAuthenticationException
+import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import retrofit.RestAdapter
 import retrofit.client.Header
@@ -25,6 +26,7 @@ import retrofit.http.Headers
 import retrofit.http.Path
 import retrofit.http.Query
 
+@Slf4j
 class DockerBearerTokenService {
   Map<String, TokenService> realmToService
   Map<String, DockerBearerToken> cachedTokens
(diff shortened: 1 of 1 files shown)


lwander:
@duftler PTAL

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerTokenService.groovy
```diff
@@ -67,6 +69,17 @@ class DockerBearerTokenService {
       resolvedPassword = "" // I'm assuming it's ok to have an empty password if the username is specified
     }
 
+    if (resolvedPassword?.length() > 0) {
+      def message = "Your registry password has %s whitespace, if this is not intentional authentication will fail."
+      if (resolvedPassword.charAt(0).isWhitespace()) {
```
duftler: Can just use `[0]` instead of `charAt(0)`.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerTokenService.groovy
```diff
@@ -67,6 +69,17 @@ class DockerBearerTokenService {
...
+        log.warn sprintf(message, ["leading"])
+      }
+
+      if (resolvedPassword.charAt(resolvedPassword.length() - 1).isWhitespace()) {
```
duftler: Same here.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerTokenService.groovy
```diff
@@ -67,6 +69,17 @@ class DockerBearerTokenService {
       resolvedPassword = "" // I'm assuming it's ok to have an empty password if the username is specified
     }
 
+    if (resolvedPassword?.length() > 0) {
+      def message = "Your registry password has %s whitespace, if this is not intentional authentication will fail."
```
duftler: Maybe s/not intentional/unintentional?

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerTokenService.groovy
lwander: We talked about this offline - but `[0]` seems to return a `String` rather than a `char` likely due to the fact that groovy supports that fancy range syntax.

PR #1196, opened 2016-11-02
Initial support for AWS autoscaling lifecycle hooks

robzienert:
Allows templated addition of account-wide autoscaling lifecycle hooks. If `defaultLifecycleHooks` are defined, or account-level `lifecycleHooks` is defined, any server group created within that account will have lifecycle hooks created and applied to it.

There's currently no way to delete lifecycle hooks from server groups outside of the deletion of the ASG.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
@@ -16,8 +16,9 @@
 
 package com.netflix.spinnaker.clouddriver.aws.deploy.description
 
-import com.netflix.spinnaker.clouddriver.deploy.DeployDescription
+import com.netflix.spinnaker.clouddriver.aws.model.AmazonAsgLifecycleHook
 import com.netflix.spinnaker.clouddriver.aws.model.AmazonBlockDevice
+import com.netflix.spinnaker.clouddriver.deploy.DeployDescription
 import groovy.transform.AutoClone
 import groovy.transform.Canonical
 
@@ -63,11 +64,13 @@ class BasicAmazonDeployDescription extends AbstractAmazonCredentialsDescription 
 
   boolean ignoreSequence
   boolean startDisabled
+  boolean includeAccountLifecycleHooks = true
 
   List<AmazonBlockDevice> blockDevices
   Boolean useAmiBlockDeviceMappings
   List<String> loadBalancers
   List<String> securityGroups
+  List<AmazonAsgLifecycleHook> lifecycleHooks = []
   Map<String, List<String>> availabilityZones = [:]
   Capacity capacity = new Capacity()
   Source source = new Source()
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
@@ -24,22 +24,23 @@ import com.google.common.annotations.VisibleForTesting
 import com.netflix.frigga.Names
 import com.netflix.spinnaker.clouddriver.aws.AwsConfiguration
 import com.netflix.spinnaker.clouddriver.aws.AwsConfiguration.DeployDefaults
+import com.netflix.spinnaker.clouddriver.aws.deploy.AmiIdResolver
+import com.netflix.spinnaker.clouddriver.aws.deploy.AutoScalingWorker
 import com.netflix.spinnaker.clouddriver.aws.deploy.BlockDeviceConfig
+import com.netflix.spinnaker.clouddriver.aws.deploy.ResolvedAmiResult
+import com.netflix.spinnaker.clouddriver.aws.deploy.description.BasicAmazonDeployDescription
 import com.netflix.spinnaker.clouddriver.aws.deploy.ops.loadbalancer.LoadBalancerLookupHelper
+import com.netflix.spinnaker.clouddriver.aws.deploy.ops.loadbalancer.UpsertAmazonLoadBalancerResult
+import com.netflix.spinnaker.clouddriver.aws.model.AmazonAsgLifecycleHook
+import com.netflix.spinnaker.clouddriver.aws.model.AmazonBlockDevice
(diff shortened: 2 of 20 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ClusterCachingAgent.groovy
```diff
@@ -351,6 +355,18 @@ class ClusterCachingAgent implements CachingAgent, OnDemandAgent, AccountAware,
       .groupBy { it.autoScalingGroupName }
   }
 
+  private Map<String, List<Map>> loadLifecycleHooks(AmazonClients clients, List<String> autoScalingGroupNames) {
+    autoScalingGroupNames.collect { loadLifecycleHooks(clients, it) }*.groupBy { it.autoScalingGroupName } as Map<String, List<Map>>
```
cfieber: I'm a bit concerned about adding this in our caching run as it is going to be one API call per ASG (the rest of the describe calls we make are one per collection). I suspect this will trip us into throttling territory. Unfortunately not sure there is an alternative way to look these up in bulk.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/UpsertAsgLifecycleHookAtomicOperation.groovy
```diff
@@ -0,0 +1,60 @@
...
+
+  UpsertAsgLifecycleHookAtomicOperation(UpsertAsgLifecycleHookDescription description) {
+    this.description = description
+  }
+
+  @Override
+  Void operate(List priorOutputs) {
+    final lifecycleHookName = description.name ?: "${description.serverGroupName}-lifecycle-${idGenerator.nextId()}"
+    final request = new PutLifecycleHookRequest(
+      lifecycleHookName: lifecycleHookName,
```
cfieber: does this request modify an existing hook with this name if its already present? (Upsert naming would imply so)

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/UpsertAsgLifecycleHookAtomicOperation.groovy
```diff
@@ -0,0 +1,60 @@
...
+package com.netflix.spinnaker.clouddriver.aws.deploy.ops
+
+import com.amazonaws.services.autoscaling.model.PutLifecycleHookRequest
+import com.netflix.spinnaker.clouddriver.aws.deploy.description.UpsertAsgLifecycleHookDescription
+import com.netflix.spinnaker.clouddriver.aws.security.AmazonClientProvider
+import com.netflix.spinnaker.clouddriver.aws.services.IdGenerator
+import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation
+import org.springframework.beans.factory.annotation.Autowired
+
+class UpsertAsgLifecycleHookAtomicOperation implements AtomicOperation<Void> {
```
cfieber: Feels like we should have the corresponding delete or detach operation to go with this one

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ClusterCachingAgent.groovy
robzienert: Yeah, there's no method of looking them up in bulk. This was extra code that's not actually necessary for this PR, I was adding it so we could expose it via other means later on. Agree, it'd be a lot of extra load on the API. I'll remove.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/UpsertAsgLifecycleHookAtomicOperation.groovy
robzienert: It does, yes.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/UpsertAsgLifecycleHookAtomicOperation.groovy
robzienert: Sure, we can do that; expose it through an API endpoint? I was just expecting that we'd want to keep the hooks around for the duration of the ASG. AWS will cascade deletions of lifecycle hooks when the ASG is deleted.

cfieber: ehhh, if the deletes are going to happen automatically, lets leave out the operation until someone actually asks for it

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
```diff
@@ -265,6 +265,8 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
       copyScalingPoliciesAndScheduledActions(
         task, sourceRegionScopedProvider, description.credentials, description.source.asgName, region, asgName
       )
+
+      createLifecycleHooks(task, regionScopedProvider, account, asgName)
```
robzienert: I've updated the codebase to support both adding deploy-time lifecycle hooks, as well as optionally opt-ing out of account lifecycle hooks.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
```diff
@@ -63,11 +64,13 @@ class BasicAmazonDeployDescription extends AbstractAmazonCredentialsDescription
 
   boolean ignoreSequence
   boolean startDisabled
+  boolean includeAccountLifecycleHooks
```
cfieber: should this have a default (`true` I'd guess)?

robzienert:
All comments addressed.

PR #1202, opened 2016-11-03
(Openstack) Added user data file to use with every deploy.

varikin:
Provides common user data that every Openstack deployment will use. The
user data provided per app will be appended to this userdata.

Ideally, this will be replaced/moved into a core location where it is
available for all providers. See spinnaker/spinnaker#1274


varikin:
@spinnaker/reviewers PTAL

PR #1207, opened 2016-11-04
(kubernetes) fix clone from non-default namespace

lwander:
@duflter, I found this bug while refactoring the kubernetes caching agents. Orca strips the namespace field, and clones out of the default namespace fail in clouddriver.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/servergroup/CloneKubernetesAtomicOperation.groovy
```diff
@@ -47,11 +47,13 @@ class CloneKubernetesAtomicOperation implements AtomicOperation<DeploymentResult
   */
   @Override
   DeploymentResult operate(List priorOutputs) {
-    CloneKubernetesAtomicOperationDescription newDescription = cloneAndOverrideDescription()
+    description.source.namespace = description.source.region
```
duftler: Should this only override the namespace if it's not explicitly set? Wouldn't that behavior be expected if the calls didn't only originate from orca?

lwander: Ah good call

PR #1211, opened 2016-11-07
(google) fixed spring bean loading

lwander:
Some recent changes to the credential initializer were preventing the provider from starting.

@duftler PTAL
@cfieber FYI


cfieber:
Thanks and sorry I missed that

lwander:
np

PR #1218, opened 2016-11-08
provider/google: Health check 'target' display fix.

jtk54:
@duftler @danielpeach please review. `target` was not rendering properly for load balancers not using the `GoogleHealthCheckView` object and using the raw `GoogleHealthCheck` object (i.e. LBs where the health check is reported through a `BackendService`).  I moved the `target` calculation into the actual model object rather than in the `View`, so that it's present regardless if you use the `View` or not.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleHealthCheck.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleHealthCheck.groovy
@@ -57,6 +57,42 @@ class GoogleHealthCheck {
     new View()
   }
 
+  /**
+   * Health check endpoint of the form '{PROTO}:{PORT}{PATH}'.
+   */
+  String getTarget() {
+    switch (healthCheckType) {
+      case HealthCheckType.HTTP:
+        return this.port ?
+          "HTTP:${this.port}${this.requestPath ?: '/'}" :
+          null
+        break
+      case HealthCheckType.HTTPS:
+        break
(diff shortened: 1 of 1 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleHealthCheck.groovy
```diff
@@ -57,6 +57,47 @@ class GoogleHealthCheck {
     new View()
   }
 
+  /**
+   * Health check endpoint of the form '{PROTO}:{PORT}{PATH}'.
+   */
+  String getTarget() {
+    switch (healthCheckType) {
+      case HealthCheckType.HTTP:
+        println ",, getting http target"
```
duftler: Several extraneous `println`s here.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleHealthCheck.groovy
```diff
@@ -57,6 +57,47 @@ class GoogleHealthCheck {
...
+        return this.port ?
+          "HTTP:${this.port}${this.requestPath ?: '/'}" :
+          null
+        break
+      case HealthCheckType.HTTPS:
+        println ",, getting https target"
+        break
+        return this.port ?
+          "HTTP:${this.port}${this.requestPath ?: '/'}" :
```
danielpeach: HTTPS?

PR #1224, opened 2016-11-09
provider/google: Support CONNECTION balancing mode in SSL LB.

jtk54:
@duftler @danielpeach please review. From the [docs](https://cloud.google.com/compute/docs/load-balancing/tcp-ssl/), SSL LBs can support `UTILIZATION` and `CONNECTION` balancing modes. We had `UTILIZATION`, so this adds `CONNECTION`. We'll have to add support to the server group wizard on the frontend as well.


danielpeach:
LGTM

PR #1227, opened 2016-11-09
Introduces shim interface for making load balancer controller interfa…

ttomsu:
…ce explicit.

I've implemented the interface for Google and AWS load balancers as a sanity check. I'll then move on to the remaining providers one by one, asking owners to confirm the output before and after is the same.

@duftler @jtk54 @cfieber @ajordens PTAL

@lwander @varikin @scotmoor @gregturn Brace yourselves, PRs are coming.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/LoadBalancerProviderTempShim.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/LoadBalancerProviderTempShim.groovy
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.model
+
+import com.fasterxml.jackson.annotation.JsonProperty
+
+/**
+ * This is a temporary class for consolidating the load balancer controllers for each cloud provider.
+ * Each cloud provider-specific controller will implement this interface (it already does
+ * implicitly, this interface just makes it explicit). Then, this interface will be merged into the
+ * LoadBalancerProvider interface while each controller will merge with its
+ * <Cloud>LoadBalancerProvider.
+ */
+interface LoadBalancerProviderTempShim {
+
+  List<Item> list()
+
+  Item get(String name)
+
+  List<Details> byAccountAndRegionAndName(String account, String region, String name)
+
+  // Some providers call this a "Summary", which I think is semantically different from what it is:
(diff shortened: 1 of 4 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/LoadBalancerProviderTempShim.groovy
```diff
@@ -0,0 +1,61 @@
...
+
+package com.netflix.spinnaker.clouddriver.model
+
+import com.fasterxml.jackson.annotation.JsonProperty
+
+/**
+ * This is a temporary class for consolidating the load balancer controllers for each cloud provider.
+ * Each cloud provider-specific controller will implement this interface (it already does
+ * implicitly, this interface just makes it explicit). Then, this interface will be merged into the
+ * LoadBalancerProvider interface while each controller will merge with it's
```
duftler: s/it's/its

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/LoadBalancerProviderTempShim.groovy
```diff
@@ -0,0 +1,61 @@
...
+  // a details view object, grouped by account, then region.
+  interface Item {
+    String getName()
+
+    @JsonProperty("accounts")
+    List<ByAccount> getByAccounts()
+  }
+
+  interface ByAccount {
+    String getName()
```
lwander: Is this name the account name, or the load balancer name?

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/LoadBalancerController.groovy
```diff
@@ -26,19 +26,18 @@ import org.springframework.web.bind.annotation.RequestMethod
...
   Set<LoadBalancer> list(@PathVariable String application) {
-    ((List<LoadBalancer>) loadBalancerProviders.findResults {
+    loadBalancerProviders.findResults {
       it.getApplicationLoadBalancers(application)
-    })
-      .flatten()
-      .sort { a, b -> a.name.toLowerCase() <=> b.name.toLowerCase() }
+    }
+    .flatten()
+    .sort { a, b -> a.name.toLowerCase() <=> b.name.toLowerCase() } as Set<LoadBalancer>
```
lwander: Does this have to be coerced/return a Set? That means each call requires hashing every load balancer object to return an unordered list after we've sorted our results.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/LoadBalancerProviderTempShim.groovy
lwander: Looks like account, but could perhaps be made clearer with a comment

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/LoadBalancerController.groovy
ttomsu: Not sure why it is returned as as a Set. I hope not for deduplication purposes, because I'm now making it return a List

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/LoadBalancerProviderTempShim.groovy
ttomsu: English are hard.

PR #1233, opened 2016-11-10
Azure load balancer temp shim impl

ttomsu:
@scotmoor @rguthriemsft PTAL

tag: https://github.com/spinnaker/fiat/issues/99


ttomsu:
Obsoleted by https://github.com/spinnaker/clouddriver/pull/1246

PR #1234, opened 2016-11-10
Openstack load balancer controller temp shim impl

ttomsu:
@varikin PTAL

tag: https://github.com/spinnaker/fiat/issues/99


ttomsu:
The Summary object changes will likely cause Deck changes as it's a change to the model.

Also check out https://github.com/spinnaker/clouddriver/blob/9e597d0197ddb112ecbd0e996649e98586feb3da/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/LoadBalancerProviderTempShim.groovy#L36-L37

I think the original Amazon model incorrectly named it an `AmazonLoadBalancerSummary`, and that mistake has propagated to each provider.

PR #1236, opened 2016-11-11
provider/google: Interpret RUNNING instance health as up, not unknown.

jtk54:
@duftler @danielpeach @ttomsu thoughts? Currently, creating a single server group fails to complete because the instance is never 'UP', even though the platform says it's 'RUNNING'. So far, the only negative effect I've seen from the fix is that disabled server groups on an LB are green, as in 'UP' even though the LB health checks return 'UNKNOWN' because of the platform health.

Example:
up

Anyone have issues with this? It seems more broken to have operations failing than technically incorrect health states. Note that instances that are checked with a LB health check and fail will still be reported 'DOWN' via [this function](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleInstance.groovy#L130).


lwander:
This is solved by "consider only cloud provider health" in the application edit screen.

jtk54:
Yep, closing.

PR #1239, opened 2016-11-14
fix AWS cn-north-1

qqshfox:
to use inside ec2 in China region, must set `SPINNAKER_AWS_DEFAULT_REGION` to `cn-north-1`
to use outside ec2 (e.g. docker-compose) for China region, must set both `SPINNAKER_AWS_DEFAULT_REGION` and `AWS_REGION` to `cn-north-1`


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
```diff
@@ -64,7 +64,12 @@
  */
 public class AmazonClientProvider {
 
-  public static final String DEFAULT_REGION = null;
+  public static final String DEFAULT_REGION;
+
+  static {
+    final Region currentRegion = Regions.getCurrentRegion();
+    DEFAULT_REGION = currentRegion == null ? null : currentRegion.getName();
```
lwander: Groovy allows you to shorten this to `currentRegion?.name`

qqshfox: @lwander Yes, I think so too. But this is Java code actually.

qqshfox:
This PR does not work for running outside AWS in China region.

According to the AWS SDK:

> public static Region getCurrentRegion()
> Returns a Region object representing the region the application is running in, when running in EC2. If this method is called from a non-EC2 environment, it will return null.

The `DEFAULT_REGION` is always `null`.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
lwander: Ah, whoops, sorry :|

qqshfox:
Update the code to work with both global and china region

robzienert:
@qqshfox There's some additional work needed for this PR:

Similar partition work for the CN region is necessary here: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/ARN.java
... And here: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AssumeRoleAmazonCredentials.java#L37

That said, you're on the right track.

qqshfox:
Rebased with the master branch. Changed `ARN.java` according to @robzienert .

https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AssumeRoleAmazonCredentials.java#L37 remaining untouched since the parameter is a string, not a regex. I don't how to make it work with both AWS Global and AWS China.

robzienert:
@qqshfox Without `AssumeRoleAmazonCredentials` updated, major parts of clouddriver probably won't function correctly (or at all). I think certain we can get away with something like this:


```diff
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AssumeRoleAmazonCredentials.java
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AssumeRoleAmazonCredentials.java
@@ -32,9 +32,13 @@ public class AssumeRoleAmazonCredentials extends AmazonCredentials {
     static final String DEFAULT_SESSION_NAME = "Spinnaker";

     static AWSCredentialsProvider createSTSCredentialsProvider(AWSCredentialsProvider credentialsProvider, String accountId, String assumeRole, String sessionName) {
+        String assumeRoleValue = Objects.requireNonNull(assumeRole, "assumeRole");
+        if (!assumeRoleValue.startsWith("arn:")) {
+          assumeRoleValue = String.format("arn:aws:iam::%s:%s", Objects.requireNonNull(accountId, "accountId"), Objects.requireNonNull(assumeRole, "assumeRole"));
+        }
         return credentialsProvider == null ? null : new NetflixSTSAssumeRoleSessionCredentialsProvider(
           credentialsProvider,
-          String.format("arn:aws:iam::%s:%s", Objects.requireNonNull(accountId, "accountId"), Objects.requireNonNull(assumeRole, "assumeRole")),
+          assumeRoleValue,
           Objects.requireNonNull(sessionName, "sessionName"),
           accountId
         );
```

In doing so, your clouddriver config for the CN regions, instead of defining your `assumeRole` value as `role/myAssumedRole`, would be the fully qualified ARN: `arn:aws-cn:iam:{{accountId}}:role/myAssumedRole`. For example:

```yaml
aws:
  defaultAssumeRole: role/myAssumedRole
  accounts:
  - name: myAccountWithCN
    accountId: 123456789012
    assumeRole: arn:aws-cn:iam:{{accountId}}:role/myAssumedRole
    regions:
    - name: cn-north-1
  - name: myAccountWithoutCN
    accountId: 123456789012
```

The downside to this, (@cfieber keep me honest here?) is that you'd only be able to have the CN region in this clouddriver account. If your AWS account has other regions Spinnaker needs to manage, I believe you would need to add another clouddriver account with the other regions (as I outlined above).

qqshfox:
@robzienert Thanks for your help. I modified `AssumeRoleAmazonCredentials` according to your suggestion.

> If your AWS account has other regions Spinnaker needs to manage, I believe you would need to add another clouddriver account with the other regions (as I outlined above)

Since AWS Beijing is an isolated region, I don't think this is a problem.

> Q: Can I use my existing AWS Inc. Account to access the China (Beijing) Region? 
> Customers who wish to use the new Beijing Region are required to sign up for a separate set of account credentials unique to the Beijing Region. Customers with existing AWS credentials will not be able to access resources in the Beijing Region.

robzienert:
Thanks @qqshfox for the contribution!

PR #1240, opened 2016-11-14
(Openstack) Allow overriding floatingNetworkId when cloning.

varikin:
When cloning a stack, a null floatingNetworkId means don't set a fip on
each instance, so can't clone the previous setting if not set in the new
one.


varikin:
@spinnaker/reviewers PTAL

PR #1242, opened 2016-11-15
Allow for Descriptions that do not contain `credentials`

ajordens:
- Supports the case where EntityTags may target all accounts (ie. `account: "*"`)

### DIFF
--- a/clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/converters/UpsertEntityTagsAtomicOperationConverter.java
+++ b/clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/converters/UpsertEntityTagsAtomicOperationConverter.java
@@ -59,8 +59,6 @@ public class UpsertEntityTagsAtomicOperationConverter extends AbstractAtomicOper
   }
 
   public UpsertEntityTagsDescription convertDescription(Map input) {
-    UpsertEntityTagsDescription description = objectMapper.convertValue(input, UpsertEntityTagsDescription.class);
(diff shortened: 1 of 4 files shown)


ajordens:
PTAL @ttomsu / @spinnaker/netflix-reviewers 

This supports some legitimate use-cases around tagging entities across all accounts (_the tag id syntax already supports this but the `clouddriver` operation would reject it because a specific account was not provided_).

Not quite sure what the auth story is around global tags ... I suspect we'll do something programmatically (and Netflix-specific) to start.

If nothing else, the operations could still be restricted by application.

clouddriver-security/src/main/groovy/com/netflix/spinnaker/clouddriver/security/resources/Global.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.security.resources;
+
+/**
+ * Marker interface indicating that a description is global and may not have account-level credentials specified.
+ */
+public interface Global {
```
ttomsu: `Global` isn't really descriptive of what it is - a credential-less operation? I'm not familiar enough with what you're working on to know what kind of auth impact it will have.

ajordens:
You're right, Global is/was a poor name.

I'll come with something a little better.

Not quite sure what I was thinking end-of-day.

In a nutshell, just looking to support clouddriver that are
non-credentialed.

ajordens:
Back story on what we're doing ... 

We're planning on using `iamrole` tags to denote what applications (and subsequently groups when we have `fiat` in place) are able to launch with a particular role.

PR #1249, opened 2016-11-16
(kubernetes) terminate security context builder

lwander:
@danielpeach


lwander:
Required by https://github.com/spinnaker/deck/pull/3007

PR #1254, opened 2016-11-17
fixed bug in determining if subnet needed to be removed from Azure or…

scotmoor:
… not when destroying the server group


scotmoor:

PR #1255, opened 2016-11-17
Avoid attempting to register terminated instances in load balancer

jeyrschabu:
- Updated to get current instance state by looking instances up by tag (In this case the asg name)
- This fixes an issue where an error would occur because we tried to register a terminated instance in the load balancer


jeyrschabu:

ajordens:
I believe you'll need to handle the potential for paged results in the
DescribeInstances call.

Also, I think you could use InstanceStateName rather than comparing against
the literal strings?

I don't think "rebooting" is an actual state is it?

(On phone and not laptop so going partially from memory)

jeyrschabu:
updated good sir. @ajordens PTAL

PR #1257, opened 2016-11-17
(docker) ignore 401 on docker health check when unauthenticated

lwander:
@skim1420, @tomaslin FYI

Steven noticed this morning that unauthenticated calls to the /v2/ endpoint of quay were returning a 401. I'm guessing this is a recent change since the Kubernetes-Spinnaker deployment has relied on quay working without authentication for some time to display the Spinnaker quay images. This is OK according to the spec I realize now: https://docs.docker.com/registry/spec/api/#/api-version-check, but frustrating, since unauthenticated clients can't properly check if their registry supports the v2 endpoint. This endpoint is only used to determine if the docker provider is healthy in clouddriver, and only affects (helps) users who don't authenticate against their registry.


tomaslin:
LGTM

PR #1258, opened 2016-11-18
titus - improve ondemand behavior so that we don't load all the jobs …

tomaslin:
…in a region when only one asg is being asked for


tomaslin:
@cfieber FYI

PR #1259, opened 2016-11-18
Gracefully skip reservation report in regions with errors

robzienert:
Saw an issue today where if an EC2 client for a single account+region fails, the entire reservation report will fail. It would be better to suppress & log the failure for the account & region combination and continue with the rest rather than being all-or-nothing.


cfieber:
I'm not actually sure this is a good idea because if we failed to describe an account/region, then all the running instances in that region would potentially appear as available and cause someone consuming the report to act as if there was excess capacity.

I could get behind something that doesn't blow up the whole report if a region fails, but any one failure in a region should mark that region's data as bad.

robzienert:
Oh, yeah. That makes sense. I'll revisit this tomorrow with a little.nore thought put into it.

PR #1263, opened 2016-11-22
Load Balancer Controller Overhaul Part 2: Controller Harder

ttomsu:
The follow up to all of that jazz with the LB temp shim. 

Implementations were copy/pasted from the Controller --> Provider, and the temp shim has been merged into the provider. Each step along the way is a different commit, with the last two being the end goal: a single, secured LB controller.

PTAL: @duftler @lwander @ajordens @cfieber 
FYI: @varikin @scotmoor @rguthriemsft @gregturn 

tags: 
* Fixes https://github.com/spinnaker/fiat/issues/99 
* https://github.com/spinnaker/spinnaker/issues/807 
* https://github.com/spinnaker/spinnaker/issues/755


duftler:
LGTM for GCE.
Will need to be sync'ed with these changes: https://github.com/spinnaker/clouddriver/pull/1268

duftler:
Ping @ttomsu .

ttomsu:
@danielpeach This PR will break you. I've added the methods you'll need to implement to conform to the interface.

duftler:
image

PR #1271, opened 2016-11-28
provider/google: Added logging to LB cache retrieval.

jtk54:
@duftler please review. This adds some logging to the LB cache retrieval since there seems to be a few intermittent scenarios where the LB cache data is corrupted. This will give us a bit more insight into what is going wrong.


jtk54:
Added logging for the raw cache data also so we have an before and after view.

PR #1274, opened 2016-11-29
provider/google: Handle partial LB cache data in fetch calls.

jtk54:
@duftler please review. The LB cache call [here](https://www.google.com/url?q=https://github.com/spinnaker/clouddriver/blob/master/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/view/GoogleLoadBalancerProvider.groovy%23L48&sa=D&usg=AFQjCNGxYKGnvVXnp4-FgSiHCw2HX4xT3A) loops over all the cached LBs, which can be malformed in funky timing cases (e.g. if one of the GCP calls in the caching agent hangs, etc). This seems to be the only place this isn't handled properly.

We have been experiencing very intermittent issues where the load balancer we are operating on is well-formed, but another in the cache occasionally seems to be malformed, causing failures. This will give us some resiliency to these timing issues so completely unrelated operations don't fail for bad LB cache entries.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/callbacks/Utils.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/callbacks/Utils.groovy
@@ -20,10 +20,12 @@ import com.google.api.services.compute.model.*
 import com.netflix.spinnaker.clouddriver.google.deploy.GCEUtil
 import com.netflix.spinnaker.clouddriver.google.model.GoogleServerGroup
 import com.netflix.spinnaker.clouddriver.google.model.loadbalancing.*
+import groovy.util.logging.Slf4j
 import org.springframework.util.ClassUtils
 
 import java.text.SimpleDateFormat
 
+@Slf4j
 class Utils {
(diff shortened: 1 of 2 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/callbacks/Utils.groovy
```diff
@@ -277,7 +277,7 @@ class Utils {
   static boolean determineInternalLoadBalancerDisabledState(GoogleInternalLoadBalancer loadBalancer,
                                                             GoogleServerGroup serverGroup) {
     def regionalLoadBalancersFromMetadata = serverGroup.asg.get(GoogleServerGroup.View.REGIONAL_LOAD_BALANCER_NAMES)
-    List<GoogleLoadBalancedBackend> serviceBackends = loadBalancer.backendService.backends
+    List<GoogleLoadBalancedBackend> serviceBackends = loadBalancer?.backendService?.backends
```
duftler: Would be a good idea to log the fact that we encountered a malformed load balancer representation here and below.

PR #1276, opened 2016-11-30
provider/appengine server group scaling policy

danielpeach:
@lwander or @duftler please review

- Adds scaling policy details to server group model. 
- Maps scaling policies to server group capacity
- Instance status is either Up or OutOfService, depending on if the server group is attached to a load balancer.

### DIFF
--- a/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/model/AppEngineModelUtil.groovy
+++ b/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/model/AppEngineModelUtil.groovy
@@ -16,10 +16,38 @@
 
 package com.netflix.spinnaker.clouddriver.appengine.model
 
+import com.google.api.services.appengine.v1.model.Service
+import com.google.api.services.appengine.v1.model.Version
+import com.netflix.spinnaker.clouddriver.model.HealthState
+
 import java.text.SimpleDateFormat
 
 class AppEngineModelUtil {
+  private static final dateFormats = ["yyyy-MM-dd'T'HH:mm:ss.SSSSSS'Z'", "yyyy-MM-dd'T'HH:mm:ss'Z'"]
+    .collect { new SimpleDateFormat(it) }
+
   static Long translateTime(String time) {
-    time ? (new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSSSS'Z'").parse(time)).getTime() : 0
+    for (SimpleDateFormat dateFormat: dateFormats) {
+      try {
+        return dateFormat.parse(time).getTime()
+      } catch (e) { }
+    }
+
+    null
+  }
+
+  static AppEngineScalingPolicy getScalingPolicy(Version version) {
+    if (version.getAutomaticScaling()) {
+      return new AppEngineScalingPolicy(version.getAutomaticScaling())
+    } else if (version.getBasicScaling()) {
+      return new AppEngineScalingPolicy(version.getBasicScaling())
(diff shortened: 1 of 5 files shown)


clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/model/AppEngineModelUtil.groovy
```diff
@@ -16,10 +16,37 @@
...
+    .collect { new SimpleDateFormat(it) }
+
   static Long translateTime(String time) {
-    time ? (new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSSSS'Z'").parse(time)).getTime() : 0
+    for (SimpleDateFormat dateFormat: dateFormats) {
+      try {
+        return dateFormat.parse(time).getTime()
+      } catch (e) { }
+    }
+    return 0
```
lwander: Since you're returning `Long` rather than `long`, why not return `null`?

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/model/AppEngineServerGroup.groovy
```diff
@@ -48,13 +52,16 @@ class AppEngineServerGroup implements ServerGroup, Serializable {
...
+    this.servingStatus = version.getServingStatus() ? ServingStatus.valueOf(version.getServingStatus()) : null
+    this.env = version.getEnv() ? Environment.valueOf(version.getEnv().toUpperCase()) : null
   }
 
   @Override
   ServerGroup.InstanceCounts getInstanceCounts() {
     new ServerGroup.InstanceCounts(
       down: 0,
-      outOfService: 0,
+      outOfService: (Integer) instances?.count { it.healthState == HealthState.OutOfService },
```
lwander: Why allow this one to be `null` but instead have `up` be `0`?

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/model/AppEngineServerGroup.groovy
```diff
@@ -64,8 +71,36 @@ class AppEngineServerGroup implements ServerGroup, Serializable {
...
+
+    switch (scalingPolicy?.type) {
+      case ScalingPolicyType.AUTOMATIC:
+        /*
+        * For the flexible environment, a version using automatic scaling can be stopped.
+        * A stopped version scales down to zero instances and ignores its scaling policy.
+        * */
+        def min = Math.min(instanceCount, scalingPolicy.minTotalInstances)
+        def desired = servingStatus == ServingStatus.SERVING ? scalingPolicy.maxTotalInstances : 0
+        return new ServerGroup.Capacity(min: min ?: 0,
```
lwander: What's the point of the `?:` here? It seems like `min` can't be null since `instanceCount` will never be `null`.

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/model/AppEngineServerGroup.groovy
```diff
@@ -64,8 +71,36 @@ class AppEngineServerGroup implements ServerGroup, Serializable {
...
+                                        max: scalingPolicy.maxTotalInstances ?: 0,
+                                        desired: desired ?: 0)
```
lwander: Same question here about `?:`

PR #1281, opened 2016-12-06
(kubernetes) Pod container conditions may be empty

lwander:
@varikin FYI


lwander:
Thanks for finding this @varikin

PR #1283, opened 2016-12-06
(docker) paginate tag list

lwander:
@duftler PTAL
@skim1420 FYI


clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -275,6 +278,26 @@ class DockerRegistryClient {
     return catalog
   }
 
+  public DockerRegistryTags getTags(String repository, String path = null) {
+    def response = request({
+      path ? registryService.get(path, tokenService.basicAuthHeader, clouddriverUserAgentApplicationName) :
```
duftler: boo

PR #1284, opened 2016-12-07
Configurable Service Limits

andrewbackes:
* Generic Service Limit Configuration
* Configurable rates for:
	* EC2InstanceHealth
	* ELBInstanceHealth
	* ASGDescribeLaunchConfigurations


andrewbackes:
@cfieber Could you take a look at this for me?

cfieber:
this is a good start.

I'm working on some thoughts around taking this a bit further (I would like to support a hierarchy of global defaults, provider defaults, account defaults, agent defaults, agent+account defaults) - I will provide a concrete example of what I think that would look like.

specifically to what is in this PR, I think the rateLimiter stuff as is won't have a huge effect - most of the API calls take significant fractions of a second to execute so any rateLimit greater than 1 probably isn't really throttling the API request rate.

I think we'll get the biggest win out of making the caching agent intervals explicitly configurable (will provide an example of that) so that we can tune exactly how often each type of cache ends up getting refreshed (via [CustomScheduledAgent](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CustomScheduledAgent.groovy) interface)

andrewbackes:
Superseded by https://github.com/spinnaker/clouddriver/pull/1291

PR #1287, opened 2016-12-09
Ensure there are security groups before a load balancer update

jeyrschabu:
- Added a check to make sure we have security groups before applying them to the load balancer


jeyrschabu:
@anotherchrisberry @robzienert

jeyrschabu:
merging this down to stop the chex blast

PR #1291, opened 2016-12-09
adds support for fine grained service limits.

cfieber:
a service limit is a key to Double value pair, for example a rateLimit against an API.

limits can be configured at a fine grained level (implementation and account) and fall back through
several levels of defaults (implementation default, account default, cloud provider default, global default).

configuration looks like:

````yaml
serviceLimits:
  defaults:
    rateLimit: 10

  cloudProviderOverrides:
    aws:
      rateLimit: 15

  accountOverrides:
    test:
      rateLimit: 5
    prod:
      rateLimit: 100

  implementationLimits:

    AmazonEC2:
      defaults:
        rateLimit: 200
      accountOverrides:
        prod:
          rateLimit: 500

    AmazonElasticLoadBalancing:
      defaults:
        rateLimit: 10
````

In this example:
* requesting the rateLimit for AmazonElasticLoadBalancing for any account would return 10
* requesting the rateLimit for AmazonEC2 for the prod account would return 500, and any other account 200
* requesting the rateLimit for AmazonCloudWatch for the prod account would return 100, test 5, any other account 15

This configuration should extend to support per caching-agent level configuration (pollInterval) and can be applied in other cloud provider api clients as well.

The initial application of this configuration is to configure a rateLimit (maximum requests per second) for the various Amazon API clients returned by AmazonClientProvider. If unconfigured a default of 10 requests per second is used.

The rate limits in AmazonClientProvider are scoped to a client type (e.g. AmazonEC2) in a specific account and region.
The rate limits are applied globally to all clients requested by AmazonClientProvider and enforced by a request handler that acquires an object from a guava RateLimiter before request execution.

Additionally, this change updates AmazonClientProvider to keep a cache of API clients it has created (again per client type, account, and region) instead of always creating a new SDK client for each call.

closes #1284 

@andrewbackes PTAL (In playing with my prototype implementation and working through some scenarios I ended up getting this close enough that I just polished it off)
@spinnaker/google-reviewers FYI (config mechanism may be useful in the near term if you have any similar API throttling issues to contend with, and I expect to extend this into the caching agent scheduler as a followup)

I'm still doing some testing around this, so not up for merge yet

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/AwsConfiguration.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/AwsConfiguration.groovy
@@ -20,6 +20,7 @@ import com.amazonaws.retry.RetryPolicy.BackoffStrategy
 import com.amazonaws.retry.RetryPolicy.RetryCondition
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.awsobjectmapper.AmazonObjectMapper
+import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.cats.agent.Agent
 import com.netflix.spinnaker.cats.provider.ProviderSynchronizerTypeWrapper
 import com.netflix.spinnaker.clouddriver.aws.agent.CleanupAlarmsAgent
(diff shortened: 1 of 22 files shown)


venturaville:
I like the idea of it.  Most of the rate limiting problems I ran into were limited to particular part of the API

cfieber:
@ewiseblatt rate limiting is applied to all AWS SDK clients created in AmazonClientProvider ( https://github.com/spinnaker/clouddriver/pull/1291/files#diff-7f74150448d21001ad0596173027f1cdR486 )

agree that it is not perfect from a multi node perspective, but I think we can mostly ignore that problem because we coordinate workers across the nodes so there should only ever be one running instance of a particular caching agent that is doing the particular rate limited API calls. if that turns out not to be the case we could look at a redis-backed rate limiter to allocate tickets across multiple processes

ewiseblatt:
ok. thanks. I missed the "beforeRequest" acquisition in my first pass.

regarding multi node, i was just pointing that out to make sure you were
aware in case it was relevant.

jlarocque88:
Hi,

Was just wondering if this might get merged soon, I think it may help with a problem I'm having of ExceedingThrottle.

Thanks.

PR #1295, opened 2016-12-12
(kubernetes) assorted cache fixes

lwander:
* autoscalers weren't showing up
 * buildInfo now shows what created a replicaset
 * hashcode implementation for load balancers, server groups, & security groups were incorrect, leading to omitted results.

@duftler PTAL

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesServerGroup.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesServerGroup.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.clouddriver.kubernetes.model
 
+import com.netflix.frigga.Names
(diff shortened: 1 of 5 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesServerGroup.groovy
```diff
@@ -58,6 +59,22 @@ class KubernetesServerGroup implements ServerGroup, Serializable {
...
   List<KubernetesEvent> events
 
+  Map<String, Object> getBuildInfo() {
+    def imageList = []
+    def buildInfo = [:]
+    for (def container : this.deployDescription.containers) {
+      imageList.add(KubernetesUtil.getImageIdWithoutRegistry(container.imageDescription))
+    }
+
+    buildInfo += [images: imageList]
```
duftler: I think this might read easier as `buildInfo.images = ...`.

lwander: Agreed

PR #1296, opened 2016-12-12
add key vault and default resource group to config

scotmoor:


scotmoor:
@rguthriemsft PTAL

PR #1298, opened 2016-12-13
use objectMapper to handle deserialization of Azureinstance class fro…

scotmoor:
…m cache


scotmoor:
@anotherchrisberry @duftler @rguthriemsft PTAL

PR #1303, opened 2016-12-14
Failures in one region should not impact overall reservation report g…

ajordens:
…eneration

- Emits a metric on every error
- Still tries to describe every account/region even if there was a previous error in that region
- Includes a first level `errorsByRegion` attribute in the reservation report


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservationReportCachingAgent.groovy
```diff
@@ -268,7 +279,13 @@ class ReservationReportCachingAgent implements CachingAgent, CustomScheduledAgen
         def describeInstancesRequest = new DescribeInstancesRequest().withMaxResults(500)
         def allowedStates = ["pending", "running"] as Set<String>
         while (true) {
-          def result = amazonEC2.describeInstances(describeInstancesRequest)
+          DescribeInstancesResult result
+          try {
```
cfieber: is there more that could fail on a region than just this call? should the whole flatMap per region be try-catch'd

ajordens:
Broadened the scope of the try/catch, we haven't seen an error outside of the AWS calls but it seems more appropriate nevertheless.

Anything else @cfieber?

cfieber:
LGTM

PR #1305, opened 2016-12-14
provider/appengine load balancer caching agent

danielpeach:

### DIFF
--- a/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AppEngineLoadBalancerCachingAgent.groovy
+++ b/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AppEngineLoadBalancerCachingAgent.groovy
@@ -1,0 +1,239 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.appengine.provider.agent
+
+import com.fasterxml.jackson.core.type.TypeReference
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.google.api.services.appengine.v1.model.Service
+import com.netflix.spectator.api.Registry
+import com.netflix.spinnaker.cats.agent.AgentDataType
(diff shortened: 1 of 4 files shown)


clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AppEngineLoadBalancerCachingAgent.groovy
```diff
@@ -0,0 +1,237 @@
...
+
+    def jsonResult = objectMapper.writeValueAsString(result.cacheResults)
+    def loadBalancerKey = Keys.getLoadBalancerKey(accountName, loadBalancerName)
+    if (result.cacheResults.values().flatten().isEmpty()) {
+      providerCache.evictDeletedItems(ON_DEMAND.ns, [loadBalancerKey])
+    } else {
+      metricsSupport.onDemandStore {
+        def cacheData = new DefaultCacheData(
+          loadBalancerKey,
+          10 * 60, // ttl is 60 minutes.
```
duftler: Could use `TimeUnit.MINUTES.toSeconds(10)` instead to improve readability.
Also, I think the ttl you have here is 10 minutes, not 60 minutes. Which did you intend?

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AppEngineLoadBalancerCachingAgent.groovy
```diff
@@ -0,0 +1,237 @@
...
+
+        providerCache.putCacheData(ON_DEMAND.ns, cacheData)
+      }
+    }
+
+    Map<String, Collection<String>> evictions = service ? [:] : [(LOAD_BALANCERS.ns): [loadBalancerKey]]
+
+    log.info "On demand cache refresh (data: ${data}) succeeded."
+
+    new OnDemandAgent.OnDemandResult(
```
duftler: Let's use explicit `return` statements for methods that aren't simple one-liners.

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AppEngineLoadBalancerCachingAgent.groovy
```diff
@@ -0,0 +1,239 @@
...
+
+    def jsonResult = objectMapper.writeValueAsString(result.cacheResults)
+    def loadBalancerKey = Keys.getLoadBalancerKey(accountName, loadBalancerName)
+    if (result.cacheResults.values().flatten().isEmpty()) {
+      providerCache.evictDeletedItems(ON_DEMAND.ns, [loadBalancerKey])
+    } else {
+      metricsSupport.onDemandStore {
+        def cacheData = new DefaultCacheData(
+          loadBalancerKey,
+          TimeUnit.MINUTES.toSeconds(10) as Integer,
```
duftler: Not a big deal, but you'll incur some extra conversion cost here. The `DefaultCacheData` constructor takes an `int`. If you cast it to an `Integer`, java will first autobox it for you into that `Integer`, then it will unbox it for you into the `int` the constructor requires. Should be able to just use `as int`.

PR #1308, opened 2016-12-14
(kubernetes) fix issues discovered upgrading to 1.4.

lwander:
This includes a change in the label semantics, behavior of IntOrString, as well as the annotations keymap.

@duftler PTAL, @gardleopard FYI

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
@@ -808,11 +808,19 @@ class KubernetesApiConverter {
         builder = builder.withNewRollingUpdate()
 
         if (rollingUpdate.maxSurge) {
-          builder = builder.withNewMaxSurge(rollingUpdate.maxSurge)
+          def maxSurge = rollingUpdate.maxSurge
+          if (maxSurge.isInteger()) {
+            maxSurge = maxSurge as int
+          }
+          builder = builder.withNewMaxSurge(maxSurge)
         }
 
         if (rollingUpdate.maxUnavailable) {
-          builder = builder.withNewMaxUnavailable(rollingUpdate.maxUnavailable)
+          def maxUnavailable = rollingUpdate.maxUnavailable
+          if (maxUnavailable.isInteger()) {
+            maxUnavailable = maxUnavailable as int
+          }
+          builder = builder.withNewMaxUnavailable(maxUnavailable)
         }
 
         builder = builder.endRollingUpdate()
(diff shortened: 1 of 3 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/servergroup/DestroyKubernetesAtomicOperation.groovy
```diff
@@ -63,6 +63,8 @@ class DestroyKubernetesAtomicOperation implements AtomicOperation<Void> {
     if (deployment && replicaSet) {
       task.updateStatus BASE_PHASE, "Checking if deployment ${deploymentName} needs to be destroyed..."
       // If we selected to delete the replica set in the currently active deployment, this will delete everything owned by the deployment.
+      print ",, d ${credentials.apiAdaptor.getDeploymentRevision(deployment)}"
```
duftler: Extraneous `print` statements.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/servergroup/DeployKubernetesAtomicOperation.groovy
```diff
@@ -101,11 +101,6 @@ class DeployKubernetesAtomicOperation implements AtomicOperation<DeploymentResul
       }
     }
 
-
-    task.updateStatus BASE_PHASE, "Sending replica set spec to the Kubernetes master."
```
duftler: Why were these removed? Did some other log msg take their place?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
```diff
@@ -808,11 +808,19 @@ class KubernetesApiConverter {
         builder = builder.withNewRollingUpdate()
 
         if (rollingUpdate.maxSurge) {
-          builder = builder.withNewMaxSurge(rollingUpdate.maxSurge)
+          def maxSurge = rollingUpdate.maxSurge
```
duftler: I didn't chase down the api docs, but I'm guessing int is preferred, but either int or string is ok?
Maybe add a comment explaining why the extra complication here, including what exception types are possible (guessing NPE and NFE).

Or you could pull this logic out into a method that returns an `IntOrString` and use it here and below.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/servergroup/DeployKubernetesAtomicOperation.groovy
lwander: No I should move them

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
lwander: I'd prefer to pull these into separate methods, but the builder is overloaded to take an int or a string, and the builder changes type with each invocation depending on how nested you are in the build so you can't pass that to a helper.

I'll clean up the exception handler for clarity

duftler: Makes sense.
There is apparently also a groovy `isInteger()` method you can use to avoid the try/catch technique entirely.

lwander: Ah much nicer

PR #1314, opened 2016-12-15
Support for specifying a default block device mapping for unknown ins…

ajordens:
…tance types


ajordens:

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/AwsConfiguration.groovy
```diff
@@ -179,6 +180,7 @@ class AwsConfiguration {
     int maxSecurityGroups = 5
     ReconcileMode reconcileClassicLinkSecurityGroups = ReconcileMode.NONE
     List<String> reconcileClassicLinkAccounts = []
+    AmazonBlockDevice unknownInstanceTypeBlockDevice
```
cfieber: worth setting this to a default value instead of null?

cfieber:
just the one question (could go either way on) otherwise LGTM

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/AwsConfiguration.groovy
ajordens: was debating that myself ... original intention was the preserve the previous behavior (_which only would have been encountered had someone hand modified their pipeline JSON or made API calls directly_).

I think a reasonable default is fair, I'll go with a 20GB volume.

PR #1324, opened 2016-12-17
Add docker/kubernetes default config

EricJizbaMSFT:


EricJizbaMSFT:
@rguthriemsft @scotmoor @lwander @duftler PTAL

EricJizbaMSFT:
This gets used when I'm debugging clouddriver in a dev environment

lwander:
I see - my issue with this is that this change means that this config for Kubernetes will always be shipped with every installation of clouddriver, regardless of whether it's in a dev or prod environment. Now every installation has another layer of config to resolve and to potentially override, which is problematic for fields like "password" vs "passwordFile" where at most one should be specified, now forcing the user to manually set whatever they aren't providing as null/empty.

EricJizbaMSFT:
I see how this could cause issues down the line and there are other methods of setting these in a dev environment, so I'll close this

PR #1331, opened 2016-12-20
Stubs out KubernetesLoadBalancer to implement LoadBalancerProvider.It…

ttomsu:
…em, allowing it to pass through the auth filters. This prevents HTTP 500 errors that were happening, but doesn't actually lock down the kubernetes load balancer endpoint.

The proper fix for this is to redo the Kubernetes Deck module to utilize the same structure of responses all of the other providers supply. Namely, the `account -> region -> summary` breakdown.

@lwander PTAL
@danielpeach FYI


lwander:
Validated

PR #1338, opened 2016-12-21
Revised entity tags model

ajordens:
- Tags are now stored as List<Map> rather than Map


ajordens:
FYI @anotherchrisberry

PR #1343, opened 2016-12-23
Updating to use pagination-enabled project search endpoint from front50

jeyrschabu:
- Going to cover the current usage of the search endpoint in front50 (searches against name & apps)


jeyrschabu:
@ajordens PTAL

ajordens:
On my phone but this LGTM.

PR #1348, opened 2016-12-30
Making the max security groups allowed per instance configurable at a…

jeyrschabu:
…ccount level

-(aws) added a new field to account maxSecurityGroupsPerInstance to check for the max allowed number of sgs in VPC


jeyrschabu:
@ajordens PTAL. If we believe this brings value, the next step would be to identify which accounts will need to be configured accordingly.
This will solve an issue with instances failing to launch due to the sec groups limit. The UI shows the "Wait For Up Instances" hangs till the timeout is reached

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
```diff
@@ -81,6 +81,7 @@ class AutoScalingWorker {
   private int desiredInstances
 
   private RegionScopedProviderFactory.RegionScopedProvider regionScopedProvider
+  private Integer maxSecurityGroupsPerInstance
```
ajordens: `maxSecurityGroups` vs `maxSecurityGroupsPerInstance` as the config property name?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
```diff
@@ -203,6 +204,9 @@ class AutoScalingWorker {
     // Favor subnetIds over availability zones
     def subnetIds = subnetIds?.join(',')
     if (subnetIds) {
+      if (securityGroups && maxSecurityGroupsPerInstance && securityGroups.size() > maxSecurityGroupsPerInstance) {
+        throw new IllegalArgumentException("An instance can have at most ${maxSecurityGroupsPerInstance} security groups")
```
ajordens: Not sure if calling out `instance` in the error message would be confusing to users given that it's happening in the context of them creating a server group?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
```diff
@@ -256,7 +256,9 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
         regionScopedProvider: regionScopedProvider,
         base64UserData: description.base64UserData,
         legacyUdf: description.legacyUdf,
-        tags: description.tags)
+        tags: description.tags,
+        maxSecurityGroupsPerInstance: account.getMaxSecurityGroupsPerInstance() ?: deployDefaults.maxSecurityGroups
```
ajordens: Caveat that if this rolls out we'll need to track down each account and make sure `maxSecurityGroupsPerInstance` is set right otherwise we might have grief with the default of 5.

jeyrschabu:
This has been solved by surfacing notices via tags

PR #1350, opened 2017-01-03
provider/appengine cleanup api utils

danielpeach:
@lwander or @duftler 

Cleans up Appengine api `queryAllVersions` utility so that it uses a batch request instead of making separate calls for each service.

### DIFF
--- a/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/AppEngineUtils.groovy
+++ b/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/AppEngineUtils.groovy
@@ -16,8 +16,12 @@
 
 package com.netflix.spinnaker.clouddriver.appengine.deploy
 
+import com.google.api.client.googleapis.batch.BatchRequest
+import com.google.api.client.http.HttpHeaders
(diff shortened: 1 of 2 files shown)


clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/AppEngineUtils.groovy
```diff
@@ -29,17 +33,31 @@ class AppEngineUtils {
...
     def services = queryAllServices(project, credentials, task, phase)
 
-    services.collect { service ->
-      queryVersionsForService(project, parseResourceName(service.getName()), credentials, task, phase)
+    BatchRequest batch = credentials.appengine.batch()
+    def allVersions = []
+
+    services.each { service ->
+      def callback = new AppEngineCallback<ListVersionsResponse>()
+        .success { ListVersionsResponse versionsResponse, HttpHeaders responseHeaders ->
```
lwander: What's the default failure behavior?

danielpeach: `AppEngineCallback` logs the error if a failure closure isn't provided.

PR #1351, opened 2017-01-03
provider/appengine retry on 409 for start, stop, and load balancer-re…

danielpeach:
…lated operations


danielpeach:
@duftler I've re-worked this.

Also added two missing file headers.

PR #1352, opened 2017-01-04
(titus) low hanging fruit. fix string comparison on the job type

jeyrschabu:


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/RegionScopedTitusClient.java
```diff
@@ -180,7 +180,7 @@ public String submitJob(SubmitJobRequest submitJobRequest) {
         if (jobDescription.getUser() == null) {
             jobDescription.setUser("spinnaker");
         }
-        if (jobDescription.getJobGroupSequence() == null && jobDescription.getType()!="batch") {
+        if (jobDescription.getJobGroupSequence() == null && jobDescription.getType().equals("batch")) {
```
tomaslin: this is not the same logic

tomaslin: needs to be !jobDescription.getType( ... or change batch to service

jeyrschabu: that snuck in there. fixing it

jeyrschabu: good catch!

jeyrschabu:
@tomaslin PTAL

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/RegionScopedTitusClient.java
```diff
@@ -180,7 +180,7 @@ public String submitJob(SubmitJobRequest submitJobRequest) {
         if (jobDescription.getUser() == null) {
             jobDescription.setUser("spinnaker");
         }
-        if (jobDescription.getJobGroupSequence() == null && jobDescription.getType()!="batch") {
+        if (jobDescription.getJobGroupSequence() == null && !jobDescription.getType().equals("batch")) {
```
tomaslin: I would change the check to be .equals("service")

PR #1353, opened 2017-01-04
(google-common) move safe retry to google-common

danielpeach:
@jtk54 or @duftler please review.

Moves `SafeRetry` logic to a base class in `google-common` so that https://github.com/spinnaker/clouddriver/pull/1351 can be rewritten.

### DIFF
--- a/clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/deploy/GoogleCommonSafeRetry.groovy
+++ b/clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/deploy/GoogleCommonSafeRetry.groovy
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/deploy/GoogleCommonSafeRetry.groovy
```diff
@@ -0,0 +1,113 @@
...
+   * @return Object returned from the operation.
+   */
+  public Object doRetry(Closure operation,
+                        String action,
+                        String resource,
+                        Task task,
+                        String phase,
+                        List<Integer> retryCodes,
+                        List<Integer> successfulErrorCodes,
+                        Class ProviderOperationException,
```
duftler: The logic around throwing this exception type might read easier if you drop the reflection in favor of having some abstract exception factory method defined on the base class.

You wouldn't need to pass the `Class` arg and you could just call `throw providerOperationException('my message...')` or something similar.

danielpeach: Yep, that is better.

PR #1354, opened 2017-01-04
(google) Add support for tagging GCE images.

duftler:


duftler:
@lwander or @danielpeach please review.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/UpsertGoogleImageTagsAtomicOperation.groovy
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * curl -X POST -H "Content-Type: application/json" -d '[ { "upsertImageTags": { "imageName": "my-image-123", "tags": { "some-key-1": "some-val-2" }, "credentials": "my-account-name" } } ]' localhost:7002/gce/ops
+   */
+  @Override
+  Void operate(List priorOutputs) {
+    task.updateStatus BASE_PHASE, "Initializing upsert of image tags for $description.imageName..."
+
+    def credentials = description.credentials
+    def compute = credentials.compute
+    def project = credentials.project
```
lwander: So this means you can't edit tags on an image not belonging to the project your service account is associated with?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/UpsertGoogleImageTagsAtomicOperation.groovy
```diff
@@ -0,0 +1,99 @@
...
+      def imageSelfLinkTokens = image.getSelfLink().split("/")
+      def imageProject = imageSelfLinkTokens[imageSelfLinkTokens.length - 4]
+
+      Map<String, String> originalLabels = image.getLabels()
+
+      if (originalLabels == null) {
+        originalLabels = [:]
+      }
+
+      Map<String, String> newLabels = originalLabels + tags
```
lwander: Is it impossible to remove labels?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/UpsertGoogleImageTagsAtomicOperation.groovy
duftler: You can (provided you have the proper permissions). The `project` above is just passed so it can be added to the list of projects to search for the requested image. The project containing the resolved image is the one used when calling `.setLabels()` down below.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/UpsertGoogleImageTagsAtomicOperation.groovy
duftler: It is with this particular operation. We could I suppose treat an empty value as indicating that a label is to be removed, but this would deviate from the behavior of the GCE api itself.

I think it's best to handle this use case if we eventually add a deleteLabels operation. For now, this is consistent with how the aws operation functions.

PR #1355, opened 2017-01-04
For AWS health indicator, don't explicitly set a region.

cfieber:
Fixes spinnaker/spinnaker#1357

Prefers SDK current region determination methods from the AmazonClientProvider to
work around issues where connectivity to us-east-1 is not available.


robzienert:
👍

PR #1356, opened 2017-01-04
enable filtering out of specific instance types from the instanceTypes endpoint

cfieber:
example configuration:

````yaml
aws:
  instanceTypes:
    excluded:
      # exclude all m1.large
      - name: m1.large
      # exclude x1.32xlarge in us-east-1 only
      - name: x1.32xlarge
        regions:
          - us-east-1
````


robzienert:
👍

PR #1357, opened 2017-01-04
adding null checks in kubernetes provider

emjburns:
Kubernetes provider is failing on my juju charmed instance of kubernetes (the canonical distribution of kubernetes) because of missing null checks. This seems to fix my problem.


emjburns:
Need to fix my commit stream...will open again

PR #1358, opened 2017-01-04
Adding null checks in kubernetes provider for caching.

emjburns:
Kubernetes provider is failing on my juju charmed instance of kubernetes (the canonical distribution of kubernetes) because of missing null checks. This seems to fix my problem.

@lwander @varikin  take a look?

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesServerGroupCachingAgent.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesServerGroupCachingAgent.groovy
@@ -336,7 +336,7 @@ class KubernetesServerGroupCachingAgent extends KubernetesCachingAgent implement
           relationships[Keys.Namespace.LOAD_BALANCERS.ns].addAll(loadBalancerKeys)
         }
 
-        pods.forEach { pod ->
+        pods?.forEach { pod ->
           def key = Keys.getInstanceKey(accountName, pod.metadata.namespace, pod.metadata.name)
           instanceKeys << key
           cachedInstances[key].with {
@@ -347,7 +347,7 @@ class KubernetesServerGroupCachingAgent extends KubernetesCachingAgent implement
           }
         }
 
-        loadBalancerKeys.forEach { loadBalancerKey ->
+        loadBalancerKeys?.forEach { loadBalancerKey ->
           cachedLoadBalancers[loadBalancerKey].with {
             relationships[Keys.Namespace.SERVER_GROUPS.ns].add(serverGroupKey)
             relationships[Keys.Namespace.INSTANCES.ns].addAll(instanceKeys)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesServerGroupCachingAgent.groovy
```diff
@@ -336,7 +336,7 @@ class KubernetesServerGroupCachingAgent extends KubernetesCachingAgent implement
           relationships[Keys.Namespace.LOAD_BALANCERS.ns].addAll(loadBalancerKeys)
         }
 
-        pods.forEach { pod ->
+        if (pods != null) pods.forEach { pod ->
```
duftler: Would be more idiomatic to instead switch from the java 8 `.forEach` construct to groovy's `.each` construct. (Not sure if there is a reason this wasn't already using `.each`...) You'll want to verify this, but I believe it will not throw NPE's even if applied to a `null` reference.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesServerGroupCachingAgent.groovy
```diff
@@ -347,14 +347,14 @@ class KubernetesServerGroupCachingAgent extends KubernetesCachingAgent implement
...
-        loadBalancerKeys.forEach { loadBalancerKey ->
+        if (loadBalancerKeys != null) loadBalancerKeys.forEach { loadBalancerKey ->
           cachedLoadBalancers[loadBalancerKey].with {
             relationships[Keys.Namespace.SERVER_GROUPS.ns].add(serverGroupKey)
             relationships[Keys.Namespace.INSTANCES.ns].addAll(instanceKeys)
           }
         }
 
-        cachedServerGroups[serverGroupKey].with {
+        if (cachedServerGroups != null) cachedServerGroups[serverGroupKey].with {
```
duftler: Could use `cachedServerGroups?.get(serverGroupKey).with` here instead.
Would want to also check if `.with` NPE's if applied to `null` reference.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesServerGroupCachingAgent.groovy
lwander: Or just write `pods?.forEach { pod ->`

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesServerGroupCachingAgent.groovy
```diff
@@ -347,14 +347,14 @@ class KubernetesServerGroupCachingAgent extends KubernetesCachingAgent implement
           }
         }
 
-        loadBalancerKeys.forEach { loadBalancerKey ->
+        if (loadBalancerKeys != null) loadBalancerKeys.forEach { loadBalancerKey ->
```
lwander: Same here as above

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesServerGroupCachingAgent.groovy
emjburns: @lwander changing to your syntax for both, works for me.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesServerGroupCachingAgent.groovy
emjburns: going to not change anything here since the bug seems to be fixed w/o it. will change if I encounter a bug in the future with this line.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesServerGroupCachingAgent.groovy
emjburns: Changed.

edwinavalos:
Thanks for getting these done so promptly! We're running into these as well now that I check our currently GKE -> Clouddriver integration

PR #1359, opened 2017-01-04
Allow Spinnaker to authenticate with k8s service accounts

lwander:
The service account's cert is always placed in `KUBERNETES_SERVICE_ACCOUNT_CA_CRT_PATH`, and the token in `KUBERNETES_SERVICE_ACCOUNT_TOKEN_PATH` by Kubernetes in every pod. If the user specifies `serviceAccount: true` in their clouddriver config, Spinnaker will read the cert & token out of those paths for authentication.

@duftler @danielpeach  PTAL

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesConfigParser.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesConfigParser.groovy
@@ -23,9 +23,45 @@ import io.fabric8.kubernetes.api.model.NamedContext
 import io.fabric8.kubernetes.client.Config
 import io.fabric8.kubernetes.client.internal.KubeConfigUtils
 
-class KubernetesConfigParser {
-  static Config parse(String kubeconfigFile, String context, String cluster, String user, List<String> namespaces) {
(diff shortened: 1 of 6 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesConfigParser.groovy
```diff
@@ -23,9 +23,45 @@ import io.fabric8.kubernetes.api.model.NamedContext
...
+      throw new IllegalStateException("Could not find CA cert for service account at $Config.KUBERNETES_SERVICE_ACCOUNT_CA_CRT_PATH")
+    }
+
+    try {
+      String serviceTokenCandidate = new String(Files.readAllBytes(new File(Config.KUBERNETES_SERVICE_ACCOUNT_TOKEN_PATH).toPath()))
+      if (serviceTokenCandidate != null) {
+        String error = "Configured service account doesn't have access. Service account may have been revoked."
+        config.setOauthToken(serviceTokenCandidate)
+        config.getErrorMessages().put(401, "Unauthorized! " + error)
+        config.getErrorMessages().put(403, "Forbidden!" + error)
```
danielpeach: I think you want an extra space after `Forbidden!` here.

PR #1364, opened 2017-01-07
(titus) Adjusting the retry settings for Titus calls

jeyrschabu:
- Current sleep = 1000 * 2 ^ 3 => 1000 * 8, Changing to: sleep = 100 * 2 ^ 8 => 100 * 256
- First step at better handling of titus failures
- This is the current standard for most of our aws calls


jeyrschabu:

tomaslin:
LGTM

PR #1365, opened 2017-01-10
LaunchFailureNotificationAgent should tag any server group that fails…

ajordens:
… to launch

Relies on elastic search and appropriate notification hooks being registered
server groups are created.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/LaunchFailureNotificationAgent.java
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/LaunchFailureNotificationAgent.java
@@ -32,11 +32,13 @@ import com.amazonaws.services.sqs.model.ReceiveMessageRequest;
 import com.amazonaws.services.sqs.model.ReceiveMessageResult;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spinnaker.cats.agent.RunnableAgent;
+import com.netflix.spinnaker.clouddriver.aws.AmazonCloudProvider;
 import com.netflix.spinnaker.clouddriver.aws.provider.AwsProvider;
 import com.netflix.spinnaker.clouddriver.aws.security.AmazonClientProvider;
 import com.netflix.spinnaker.clouddriver.aws.security.NetflixAmazonCredentials;
 import com.netflix.spinnaker.clouddriver.security.AccountCredentials;
 import com.netflix.spinnaker.clouddriver.security.AccountCredentialsProvider;
+import com.netflix.spinnaker.clouddriver.tags.ServerGroupTagger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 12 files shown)


ajordens:
image

PTAL @anotherchrisberry or @robzienert

ajordens:
@anotherchrisberry It might be worthwhile to show a timestamp alongside the alert or notice? 

Also ... not currently doing anything to automatically remove a tag if the underlying problem goes away. I have ideas on how this could be accomplished but would like to get this first bit out and running in a couple select accounts.

clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/ElasticSearchServerGroupTagger.java
```diff
@@ -0,0 +1,101 @@
...
+      entityTagsProvider,
+      upsertEntityTagsDescription(cloudProvider, accountId, region, serverGroupName, key, value)
+    );
+
+    try {
+      TaskRepository.threadLocalTask.set(new DefaultTask(this.getClass().getSimpleName()));
+      upsertEntityTagsAtomicOperation.operate(Collections.emptyList());
+    } finally {
+      TaskRepository.threadLocalTask.set(null);
+    }
```
robzienert: Is this a lock of some kind?

ajordens:
Not a lock but rather a reference to the underlying persisted clouddriver task.

In this situation I'm running an AtomicOperation but outside the scope of a persisted task so this is a workaround to requiring an actual task.

clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/ElasticSearchServerGroupTagger.java
ajordens: Not a lock but rather a reference to the underlying persisted clouddriver task.

In this situation we're running an AtomicOperation outside the scope of a persisted task and this is a workaround to requiring an actual task.

The usage of said Task is ...

`getTask().updateStatus(BASE_PHASE, format("Tagged %s in Front50", durableEntityTags.getId()));`

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/LaunchFailureNotificationAgent.java
```diff
@@ -146,13 +151,29 @@ public void run() {
...
+
+    String region = sqsMatcher.group(1);
+    String accountId = sqsMatcher.group(2);
+
+    serverGroupTagger.alert(
+      AmazonCloudProvider.ID,
+      accountId,
+      region,
+      notificationMessage.autoScalingGroupName,
+      notificationMessage.event,
```
anotherchrisberry: What does the `event` here look like? is it a UUID or a static string?

If it's dynamic, it might make life easier when handling cleanup if it's static, e.g. `launch_failure`.

anotherchrisberry:
@ajordens agree on adding a timestamp (have it locally, will get it pushed this afternoon).

LGTM

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/LaunchFailureNotificationAgent.java
ajordens: It's static ...`autoscaling:ec2_instance_launch_error`

anotherchrisberry: 👍

PR #1366, opened 2017-01-10
Added instrumentation on [google] APIs duration.

ewiseblatt:
@duftler 

The tests dont run yet, but before investing the effort in getting them working I want to run this by you because it is a big change.

This instruments most of the compute calls in clouddriver to give latency metrics.
I thought I might be able to create an interceptor in spring. This would have been pretty simple (just a few paragraphs of code) but the API has so many levels of factories that the proxies would have to create other proxies and there numerous final classes and methods so you get lots of warnings and not everything needs to be proxied so the runtime is a mess and feels scary.

The other approach is to figure out how to use the pre/post interceptors in the API itself. This would be less impactful in the code since only the compute factory would need to change. However I couldnt do much with tags there. I think it might be interesting to have tags for zones and regions to isolate latency if it really does matter since different regions are going to have different loads.

It could also be that we dont need any of this and just use the built-in platform metrics. The API panel in the developer's console already shows latency. However that is aggregated for the whole project, not just spinnaker. So not only do you not have the region/zone breakdown, but it isnt just Spinnaker so it might make using the instrumentation to debug issues harder since it is adding noise into the data.

To make the tests easier, I might see if I can accomodate a null Registry and not capture metrics in that case. Or maybe add some helper class or base to setup a global registry. At some point a test can test the metric writing, but not all the tests need to do that.

The changes in tests is old from my first stab at this PR. I havent looked to see if they are current or not so just ignore the tests for your first pass.

I can show you dashboard data if you need.

The gist is that most of the execute methods are instrumented with a timer (count/time).
In addition the polling wait operations are instrumented separately. As a rule of thumb, I do not instrument the internal get calls for those since I dont think that latency matters since you are already waiting on the effects.

The batch operations are counted separate. I added a context to each batch to indicate what it was for. and a counter for the size of the batch in addition to the timer for the batch.execute. I dont know if this is helpful or not, but could facilitate debugging and review.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/GoogleExecutor.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/GoogleExecutor.groovy
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.clouddriver.google
+
+import com.google.api.client.googleapis.batch.BatchRequest
+import com.google.api.client.googleapis.services.AbstractGoogleClientRequest;
+import com.netflix.spectator.api.Id;
+import com.netflix.spectator.api.Registry;
+
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+
+
+public class GoogleExecutor {
+  public static final  String TAG_SCOPE = GoogleExecutorTraits.TAG_SCOPE
+  public static final  String TAG_REGION = GoogleExecutorTraits.TAG_REGION
+  public static final  String TAG_ZONE = GoogleExecutorTraits.TAG_ZONE
+  public static final  String TAG_BATCH_CONTEXT = "context"
+  public static final  String SCOPE_BATCH = "batch"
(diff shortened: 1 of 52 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/GoogleExecutor.groovy
```diff
@@ -0,0 +1,89 @@
...
+import com.google.api.client.googleapis.services.AbstractGoogleClientRequest;
+import com.netflix.spectator.api.Id;
+import com.netflix.spectator.api.Registry;
+
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+
+
+public class GoogleExecutor {
+  public static final  String TAG_SCOPE = GoogleExecutorTraits.TAG_SCOPE
```
duftler: I think you can just `implements GoogleExecutorTraits` to get all of these members.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -458,26 +478,32 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
 
   private Closure updateRegionBackendServices(Compute compute, String project, String region, String backendServiceName, BackendService backendService) {
     return {
-      BackendService serviceToUpdate = compute.regionBackendServices().get(project, region, backendServiceName).execute()
+      BackendService serviceToUpdate = GoogleExecutor.timeExecute(
+          compute.regionBackendServices().get(project, region, backendServiceName),
+          "compute.regionBackendServices.get", TAG_SCOPE, SCOPE_REGIONAL, TAG_REGION, region)
```
duftler: From browsing around here, it kind of looks like we could use some reflection to figure out an appropriate class name instead of passing the string each time. Haven't dug deeply yet though.

ewiseblatt: I originally wanted to do that, but it didnt look like the Java API provided the ability to do that (at least I did not see a way browsing the Javadoc). I'd prefer to do it that way. Maybe even break out the resource from the method so it is easier to filter dashboards to only look at particular resources (e.g. backendservice calls but not image calls)

PR #1368, opened 2017-01-10
(kubernetes) Update percent enable disable semantics

lwander:
This fix is important, because if we prematurely switch a replica set's
behavior to spawn disabled pods when owned by an autoscaler, the autoscaler
becomes useless, since it cannot create pods to reduce load on the pods still
serving traffic.

@duftler or @jtk54 or @danielpeach PTAL


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/servergroup/AbstractEnableDisableKubernetesAtomicOperation.groovy
```diff
@@ -52,56 +52,62 @@ abstract class AbstractEnableDisableKubernetesAtomicOperation implements AtomicO
...
 
     task.updateStatus basePhase, "Finding requisite server group..."
 
     def replicationController = credentials.apiAdaptor.getReplicationController(namespace, description.serverGroupName)
     def replicaSet = credentials.apiAdaptor.getReplicaSet(namespace, description.serverGroupName)
 
-    task.updateStatus basePhase, "Getting list of attached services..."
+    // If we edit the spec when disabling less than 100% of pods, we won't be able to handle autoscaling
+    // actively correctly.
+    if (description.desiredPercentage == 100 || action == "true") {
```
duftler: Can also drop `description.` here.

PR #1369, opened 2017-01-10
An AtomicOperation can now return an <optional> set of OperationEvents

ajordens:
The `DeleteServerGroupEventHandler` will delete all tags associated with
a deleted server group (if ElasticSearch is enabled).

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AtomicOperation.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AtomicOperation.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Netflix, Inc.
+ * Copyright 2017 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AtomicOperation.java
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * This method will initiate the operation's work. In this, operation's can get a handle on prior output results
+   * from the requiremed method argument.
+   *
+   * @param priorOutputs
+   * @return parameterized type
+   */
+  R operate(List priorOutputs);
+
+  default Collection<OperationEvent> getEvents() {
```
ajordens: Converted this to a Java8 interface to get `default`.

ajordens:
Short-term motivation is to support tag cleanup whenever a server group is deleted (_and possibly created_).

Figured more structured events around operations _might_ be useful?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/events/DeleteServerGroupEvent.java
```diff
@@ -0,0 +1,74 @@
...
+package com.netflix.spinnaker.clouddriver.orchestration.events;
+
+public class DeleteServerGroupEvent implements OperationEvent {
+  private final OperationEvent.Type type = OperationEvent.Type.SERVER_GROUP;
+  private final OperationEvent.Action action = OperationEvent.Action.DELETE;
+
+  private final String cloudProvider;
+
+  private final String accountId;
+  private final String region;
```
cfieber: worth checking whether this is a good enough assumption on the set of coordinates that identify a server group across all providers (thinking of the GCE zone/region shenanigoats that are all over orca)?

PR #1376, opened 2017-01-12
provider/appengine use service account email in deploy operation

danielpeach:

Slightly reworks how the deploy operation works. Rather than activating a service account for every deploy operation, service accounts are activated on start-up. The deploy command runs just using the service account email and project name.

### DIFF
--- a/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/security/AppEngineCredentialsInitializer.groovy
+++ b/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/security/AppEngineCredentialsInitializer.groovy
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.clouddriver.appengine.security
 
 import com.netflix.spinnaker.cats.module.CatsModule
+import com.netflix.spinnaker.clouddriver.appengine.AppEngineJobExecutor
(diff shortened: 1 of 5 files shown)


clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/security/AppEngineCredentialsInitializer.groovy
```diff
@@ -68,8 +78,11 @@ class AppEngineCredentialsInitializer implements CredentialsInitializerSynchroni
           .applicationName(clouddriverUserAgentApplicationName)
           .jsonPath(managedAccount.jsonPath)
           .requiredGroupMembership(managedAccount.requiredGroupMembership)
+          .serviceAccountEmail(managedAccount.serviceAccountEmail)
           .build()
 
+        jobExecutor.runCommand(["gcloud", "auth", "activate-service-account", "--key-file", managedAccount.jsonPath])
```
lwander: Previously, wasn't it possible to start this using just application default credentials?

danielpeach:
I think this should handle application default credentials properly now. In the case that we're using application default credentials, the account will have to have both the project name and default service account email - but we can just write those to spinnaker-local.yaml.

danielpeach:
@lwander updated to pull down project and service account email from metadata.

PR #1379, opened 2017-01-17
Adds checking of account from server group and load balancer list() c…

ttomsu:
…alls

@duftler @jtk54 PTAL

Fixes https://github.com/spinnaker/fiat/issues/133

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/AuthorizationSupport.groovy
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/AuthorizationSupport.groovy
@@ -53,6 +53,30 @@ class AuthorizationSupport {
     return true
   }
 
+  /**
+   *   Used for filtering result lists by searching for "account"-like properties.
+   */
+  boolean filterForAccounts(List items) {
+    if (!items) {
+      return true
+    }
+
+    Authentication auth = SecurityContextHolder.context.authentication;
+
+    new ArrayList<>(items).each { Object item ->
+      Map propertySource = item.properties
+      if (item instanceof Map) {
+        propertySource = item
+      }
+      String account = propertySource.account ?: propertySource.accountName
+
+      if (account && !permissionEvaluator.hasPermission(auth, account, 'ACCOUNT', 'READ')) {
+        items.remove(item)
+      }
+    }
+    return true
+  }
+
   boolean filterLoadBalancerProviderItems(List<LoadBalancerProvider.Item> lbItems) {
     if (!lbItems) {
       return true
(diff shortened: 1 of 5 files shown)


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/AuthorizationSupport.groovy
```diff
@@ -53,6 +53,30 @@ class AuthorizationSupport {
...
+  boolean filterForAccounts(List items) {
+    if (!items) {
+      return true
+    }
+
+    Authentication auth = SecurityContextHolder.context.authentication;
+
+    new ArrayList<>(items).each { Object item ->
+      Map propertySource = item.properties
+      if (item instanceof Map) {
```
jtk54: Seems like there might be more types that would cause trouble here, like `List` or `Array`. If you never call it with those types this is fine. Maybe just add something to the docstring so that doesn't happen in the __future__.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/AuthorizationSupport.groovy
```diff
@@ -53,6 +53,30 @@ class AuthorizationSupport {
...
+        propertySource = item
+      }
+      String account = propertySource.account ?: propertySource.accountName
+
+      if (account && !permissionEvaluator.hasPermission(auth, account, 'ACCOUNT', 'READ')) {
+        items.remove(item)
+      }
+    }
+    return true
```
jtk54: All branches of this function return `true` I think. Is this just to satisfy the `PostAuthorize` annotation?

ttomsu: Yep - it's mostly acting as a filter, even if all items are filtered out. Returning `false` would indicate the user doesn't have access to the resource, resulting in an HTTP 403, rather than an HTTP 200 with an empty list response.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/AuthorizationSupport.groovy
ttomsu: The `items` parameter is a `List`, and it would be a very weird design for a REST method to return a list of lists without another kind of structure in between (say, a Map). In this case, comments can do nothing to correct bad design.

jtk54: Ok SGTM.

PR #1380, opened 2017-01-17
Add Spring Cloud Config Server

gregturn:
* Spring Cloud Config Server makes it possible to externalize and centralize property files with a single github repo.
* Make Spring Cloud Config Server an optional way to retrieve property settings.
* By default, it's disabled. Enable with spring.cloud.config.enabled=true

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/configuration/CustomConfigClientConfiguration.groovy
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/configuration/CustomConfigClientConfiguration.groovy
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2017 Pivotal, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.configuration
+
+import com.netflix.spinnaker.config.OkHttpClientConfiguration
+import com.netflix.spinnaker.okhttp.OkHttpClientConfigurationProperties
+import groovy.util.logging.Slf4j
+import org.apache.http.client.HttpClient
+import org.apache.http.impl.client.HttpClients
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.cloud.config.client.ConfigClientProperties
+import org.springframework.cloud.config.client.ConfigServicePropertySourceLocator
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.core.env.Environment
+import org.springframework.http.client.ClientHttpRequestFactory
+import org.springframework.http.client.HttpComponentsClientHttpRequestFactory
+import org.springframework.web.client.RestTemplate
+/**
+ * Configures Spring Cloud Config Client to speak to a Config Server over an SSL encrypted channel.
+ *
+ * NOTE: By default, spring.cloud.config.enabled=true activates Config Client, but WITHOUT SSL. TO
+ *       have Config Client with SSL, use config-client.ssl.enabled=true instead.
+ *
+ * For property settings, see commented out settings in bootstrap.yml.
+ *
+ * NOTE: This is independent of serving requests via SSL.
+ */
+@Slf4j
+@Configuration
+@ConditionalOnProperty('okHttpClient.keyStore')
+class CustomConfigClientConfiguration {
+
+	private final OkHttpClientConfigurationProperties okHttpClientConfigurationProperties
+	private final OkHttpClientConfiguration okHttpClientConfiguration
+
+	@Autowired
+	public CustomConfigClientConfiguration(OkHttpClientConfigurationProperties okHttpClientConfigurationProperties,
+										   OkHttpClientConfiguration okHttpClientConfiguration) {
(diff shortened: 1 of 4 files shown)


gregturn:
Depends on https://github.com/spinnaker/gate/pull/326

gregturn:

This is disabled by default. Enable by **spring.cloud.config.enabled=true**. Defaults to localhost:8084/config. To override in production, set **spring.cloud.config.uri=<gate>/config**.

This is part two. The same change can be applied to each module, to allow them to pull config settings from gate.

gregturn:
@cfieber I've updated this PR to go along with defaulting to front50's URL.

gregturn:
This PR now depends on https://github.com/spinnaker/front50/pull/183

gregturn:
@cfieber This PR now includes the option to use SSL.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/configuration/ConfigClientSslProperties.groovy
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.boot.context.properties.NestedConfigurationProperty
+
+/**
+ * Use the following properties to enable SSL and configure a certificate store.
+ *
+ * See bootstrap.yml for commented example.
+ */
+@ConfigurationProperties(prefix = "config-client")
+class ConfigClientSslProperties {
```
cfieber: I'd actually prefer (though it isn't exactly accurately named) to see the use of https://github.com/spinnaker/kork/blob/master/kork-web/src/main/groovy/com/netflix/spinnaker/okhttp/OkHttpClientConfigurationProperties.groovy bean for the SSL configuration. If someone is using SSL they will have configured that bean, and it would suck to have to duplicate the same set of configuration into this bean

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/configuration/CustomConfigClientConfiguration.groovy
```diff
@@ -0,0 +1,107 @@
...
+		configServicePropertySourceLocator
+	}
+
+	@Bean
+	RestTemplate restTemplate(ClientHttpRequestFactory clientHttpRequestFactory) {
+		new RestTemplate(clientHttpRequestFactory)
+	}
+
+	@Bean
+	ClientHttpRequestFactory clientHttpRequestFactory(HttpClient httpClient) {
```
cfieber: using http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/http/client/OkHttpClientHttpRequestFactory.html would probably be better (and make my previous comment feel much less dirty)

gregturn: One problem: OkHttpClientHttpRequestFactory isn't available until Spring 4.2.

cfieber: looks like that needs us to get upgraded past spring 4.1.x 

so I'd recommend bringing in OkHttpClientConfigurationProperties for now and configuring the HttpClient with it instead of the config-server bean, and when we get upgraded we can switch out the HttpComponentsCHRF to the OkHttp flavor

gregturn:
Done. If it looks good, I can squash.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/configuration/CustomConfigClientConfiguration.groovy
```diff
@@ -0,0 +1,107 @@
...
+		def keyStoreFile = new File(configClientSslProperties.ssl.keyStore)
+
+		if (keyStoreFile.exists()) {
+			trustStore.load(keyStoreFile.newInputStream(), configClientSslProperties.ssl.keyStorePassword.toCharArray())
+			keyManagerFactory.init(trustStore, configClientSslProperties.ssl.keyStorePassword.toCharArray())
+		} else {
+			throw new RuntimeException("Cannot find resource: ${keyStoreFile.absolutePath}")
+		}
+
+		SSLContext sslContext = SSLContexts.custom().loadTrustMaterial(trustStore, new TrustSelfSignedStrategy()).build()
```
cfieber: have a look at https://github.com/spinnaker/kork/blob/master/kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttpClientConfiguration.groovy#L75 for how the SSL context init happens

that properly configures both the keystore and truststore so there is explicit trust of both the client and server

gregturn: ;)

cfieber: not sure what `;)` means in the context of my comment

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/configuration/CustomConfigClientConfiguration.groovy
```diff
@@ -0,0 +1,109 @@
...
+		KeyStore trustStore = KeyStore.getInstance(okHttpClientConfigurationProperties.keyStoreType)
+
+		if (okHttpClientConfigurationProperties.keyStore.exists()) {
+			trustStore.load(okHttpClientConfigurationProperties.keyStore.newInputStream(), okHttpClientConfigurationProperties.keyStorePassword.toCharArray())
+			keyManagerFactory.init(trustStore, okHttpClientConfigurationProperties.keyStorePassword.toCharArray())
+		} else {
+			throw new RuntimeException("Cannot find resource: ${okHttpClientConfigurationProperties.keyStore.absolutePath}")
+		}
+
+		SSLContext sslContext = SSLContexts.custom().loadTrustMaterial(trustStore, new TrustSelfSignedStrategy()).build()
```
cfieber: this still needs to be addressed as per my other comment

gregturn:
@cfieber okay, how about this for using okHttp properties AND the code for creating an sslFactory?

PR #1381, opened 2017-01-18
(titus)  Setting interestingHealthProvider (if present) as a titus job label

jeyrschabu:
- This label will be used for discovery purposes

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
@@ -46,6 +46,8 @@ class TitusDeployHandler implements DeployHandler<TitusDeployDescription> {
 
   private static final String BASE_PHASE = "DEPLOY"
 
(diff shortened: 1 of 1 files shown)


jeyrschabu:

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
```diff
@@ -77,6 +79,10 @@ class TitusDeployHandler implements DeployHandler<TitusDeployDescription> {
       if (!description.env) description.env = [:]
       if (!description.labels) description.labels = [:]
 
+      if (description.labels.containsKey(INTERESTING_HEALTH_PROVIDER_NAMES) && "Titus" in description.labels.get(INTERESTING_HEALTH_PROVIDER_NAMES)) {
```
ajordens: Not really sure what this is doing ... seems like it's overriding the `interestingHealthProviderNames` label value? 

I don't believe it's necessary, is it?

jeyrschabu: As currently coded, it's necessary. But will revisit the Orca side in the am. Might end up not needing the explicit set here afterall

PR #1384, opened 2017-01-20
Delete the launch failure tag if a server group is no longer having t…

ajordens:
…roubles

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/LaunchFailureNotificationCleanupAgent.java
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/LaunchFailureNotificationCleanupAgent.java
@@ -1,0 +1,163 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.aws.lifecycle;
+
+import com.amazonaws.AmazonServiceException;
+import com.amazonaws.services.autoscaling.AmazonAutoScaling;
+import com.amazonaws.services.autoscaling.model.Activity;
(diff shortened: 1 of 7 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/LaunchFailureNotificationCleanupAgent.java
```diff
@@ -0,0 +1,179 @@
...
+
+    taggedEntities.forEach(entityTags -> {
+      EntityTags.EntityRef entityRef = entityTags.getEntityRef();
+
+      AmazonAutoScaling amazonAutoScaling = getAmazonAutoScalingClient(
+        amazonAutoScalingClients, entityRef.getAccount(), entityRef.getRegion()
+      );
+
+      try {
+        if (hasLaunchFailures(amazonAutoScaling, entityTags)) {
```
ajordens: @cfieber Wasn't sure if it was worth an additional `RateLimiter` here if there's fear that fetching scaling activities will blow a limit.

I don't really expect there to ever be _many_ server groups in this state.

(_for example_) after a couple weeks of being live there were ~30-40 server groups that got cleaned up.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/LaunchFailureNotificationCleanupAgent.java
```diff
@@ -0,0 +1,179 @@
...
+        if (amazonServiceException.getErrorMessage().toLowerCase().contains("name not found")) {
+          return false;
+        }
+      }
+
+      throw e;
+    }
+  }
+
+  private AmazonAutoScaling getAmazonAutoScalingClient(Map<String, AmazonAutoScaling> cachedClients,
```
cfieber: no need to cache the clients AmazonClientProvider already keeps them in a guava loadingcache

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/LaunchFailureNotificationCleanupAgent.java
cfieber: it will be default rate limited to 10rps

ajordens: True ... wasn't sure if 10rps was good/bad for this particular call.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/LaunchFailureNotificationCleanupAgent.java
ajordens: Hah ok.

cfieber:
LGTM

PR #1385, opened 2017-01-20
refactors UserDataProvider interface to accept full LaunchConfigurationSettings

cfieber:
Assuming someone had actually implemented a UserDataProvider in their own codebase, this change would be compatible (introduced new preferred method with a default impl that delegates to the deprecated method)

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/UserDataProvider.java
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/UserDataProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Netflix, Inc.
+ * Copyright 2017 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/UserDataProvider.java
```diff
@@ -0,0 +1,23 @@
+package com.netflix.spinnaker.clouddriver.aws.deploy.userdata;
```
robzienert: Are we off the headers now?

robzienert:
👍

ajordens:
LGTM

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/UserDataProvider.java
cfieber: ah - didn't notice that. was an intellij convert to java guess it lost the header (I'll add it back)

PR #1389, opened 2017-01-24
(titus)  Setting interestingHealthProvider (if present) as a titus job label

jeyrschabu:
- This label will be used for discovery purposes

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/description/TitusDeployDescription.groovy
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/description/TitusDeployDescription.groovy
@@ -44,6 +44,7 @@ class TitusDeployDescription extends AbstractTitusCredentialsDescription impleme
   int retries
   int runtimeLimitSecs
   Boolean useApplicationDefaultSecurityGroup = true
+  String interestingHealthProviderNames
 
   @Canonical
   static class Capacity {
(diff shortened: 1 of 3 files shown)


jeyrschabu:
- Grabbing interestingProviderNames from the request and explicitly set it as a label
- interestingProviderNames: "Titus,..." (comma-separated for lists) is how the user should set it. Orca will extract that correctly as a list

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/description/TitusDeployDescription.groovy
```diff
@@ -44,6 +44,7 @@ class TitusDeployDescription extends AbstractTitusCredentialsDescription impleme
   int retries
   int runtimeLimitSecs
   Boolean useApplicationDefaultSecurityGroup = true
+  String interestingHealthProviderNames
```
tomaslin: Isn't this a list?

tomaslin: nvm, saw the comment above. LGTM

jeyrschabu: It's a comma-separated String and is processed like shown in here:  https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/test/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/TitusInterestingHealthProviderNamesSupplierSpec.groovy#L77

PR #1392, opened 2017-01-25
(google) Coerce maxUnavailable attributes from doubles to ints.

duftler:


duftler:
@danielpeach please review.

PR #1393, opened 2017-01-25
provider/appengine validation for gradual traffic migration

danielpeach:
@jtk54 or @lwander or @duftler please review.

@jtk54 this is the backend half of the validation you were looking for here https://github.com/spinnaker/deck/pull/3211

### DIFF
--- a/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/validators/StandardAppEngineAttributeValidator.groovy
+++ b/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/validators/StandardAppEngineAttributeValidator.groovy
@@ -16,11 +16,7 @@
 
 package com.netflix.spinnaker.clouddriver.appengine.deploy.validators
 
-import com.netflix.spinnaker.clouddriver.appengine.deploy.exception.AppEngineIllegalArgumentExeception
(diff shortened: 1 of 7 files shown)


clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/validators/StandardAppEngineAttributeValidator.groovy
```diff
@@ -261,4 +257,54 @@ class StandardAppEngineAttributeValidator {
       return true
     }
   }
+
+  def validateShardBy(AppEngineTrafficSplit split, Boolean migrateTraffic, String attribute) {
+    if (!split) {
+      return true
+    }
+
+    def shardByExists = { validateNotEmpty(split.shardBy, attribute) }
```
lwander: Why is this a closure?

danielpeach: Oh, I think there used to be more logic in there, so I can just `validateNotEmpty` directly. But I want `validateNotEmpty` to run only in certain cases, since there are times when I don't care if `split.shardBy` is empty, and in those cases I don't want `validateNotEmpty` to add error messages. If that makes sense.

PR #1394, opened 2017-01-25
Added traits to facilitate instrumentation of Google APIs

ewiseblatt:
Instruments one operation as example of how it will be used.

@duftler 
This is independent of earlier PR, which I'm looking at now.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/GoogleExecutorTraits.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/GoogleExecutorTraits.groovy
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.clouddriver.google
+
+import com.google.api.client.googleapis.batch.BatchRequest
+import com.google.api.client.googleapis.services.AbstractGoogleClientRequest;
+import com.google.api.client.http.HttpResponseException
+import com.netflix.spectator.api.Clock
+import com.netflix.spectator.api.Id
+import com.netflix.spectator.api.Registry
+
+import java.util.concurrent.TimeUnit;
+
+
+trait GoogleExecutorTraits {
+  final String TAG_BATCH_CONTEXT = "context"
+  final String TAG_REGION = "region"
+  final String TAG_SCOPE = "scope"
+  final String TAG_ZONE = "zone"
+  final String SCOPE_BATCH = "batch"
+  final String SCOPE_GLOBAL = "global"
+  final String SCOPE_REGIONAL = "regional"
+  final String SCOPE_ZONAL = "zonal"
+
+  abstract Registry getRegistry()
+
+
+  public <T> T timeExecuteBatch(BatchRequest batch, String batchContext, String... tags) throws IOException {
+     def success = "false"
+     Registry registry = getRegistry()
+     Clock clock = registry.clock()
+     long startTime = clock.monotonicTime()
+
+     try {
+       batch.execute()
+       success = "true"
+     } finally {
+       def tagDetails = [(TAG_BATCH_CONTEXT): batchContext, "success": success]
+       long nanos = clock.monotonicTime() - startTime
+       registry.timer(registry.createId("google.batchExecute", tags).withTags(tagDetails)).record(nanos, TimeUnit.NANOSECONDS)
+       registry.counter(registry.createId("google.batchSize", tags).withTags(tagDetails)).increment(batch.size())
+     }
+  }
+
+  public <T> T timeExecute(AbstractGoogleClientRequest<T> request, String api, String... tags) throws IOException {
+     def success = "false"
+     T result
+     Registry registry = getRegistry()
+     Clock clock = registry.clock()
+     long startTime = clock.monotonicTime()
+
+     try {
+       result = request.execute()
+       success = "true"
+     } finally {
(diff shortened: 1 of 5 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/GoogleExecutorTraits.groovy
```diff
@@ -0,0 +1,86 @@
...
+
+     try {
+       T result = request.execute()
+       long nanos = clock.monotonicTime() - startTime
+       registry.timer(id.withTag("success", "true")).record(nanos, TimeUnit.NANOSECONDS)
+       return result
+     } catch (IOException ioex) {
+       registry.timer(id.withTag("success", "false")).record(clock.monotonicTime() - startTime, TimeUnit.NANOSECONDS)
+       throw ioex;
+     } catch (Exception ex) {
```
duftler: If you are catching both IOException and Exception I suppose you could just catch Throwable instead.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/GoogleExecutorTraits.groovy
```diff
@@ -0,0 +1,86 @@
...
+       registry.timer(id.withTag("success", "true")).record(nanos, TimeUnit.NANOSECONDS)
+       registry.counter(sizeId.withTag("success", "true")).increment(batch.size())
+     } catch (IOException ioex) {
+       registry.timer(id.withTag("success", "false")).record(clock.monotonicTime() - startTime, TimeUnit.NANOSECONDS)
+       registry.counter(sizeId.withTag("success", "false")).increment(batch.size())
+       throw ioex
+     } catch (Exception ex) {
+       registry.timer(id.withTag("success", "false")).record(clock.monotonicTime() - startTime, TimeUnit.NANOSECONDS)
+       registry.counter(sizeId.withTag("success", "false")).increment(batch.size())
+       throw new IllegalStateException(ex);
```
duftler: Why IllegalStateException?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/ResizeGoogleServerGroupAtomicOperation.groovy
```diff
@@ -63,17 +62,21 @@ class ResizeGoogleServerGroupAtomicOperation implements AtomicOperation<Void> {
     if (isRegional) {
       def instanceGroupManagers = compute.regionInstanceGroupManagers()
 
-      instanceGroupManagers.resize(project,
-                                   region,
-                                   serverGroupName,
-                                   targetSize).execute()
+      timeExecute(
+          instanceGroupManagers.resize(project,
```
duftler: I think the formatting might look better if it were indented only 2 spaces and the closing parens after `...TAG_REGION, region` were on the next line.
But I don't feel too strongly about it if you prefer it this way.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/GoogleExecutorTraits.groovy
ewiseblatt: I'm not sure. I'd seen that pattern elsewhere, but could have been in front50. I can take it out.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/GoogleExecutorTraits.groovy
ewiseblatt: The IOException is handled different from the others. It is passed through. The caller might grab the HTTP response from HTTP-like exceptions. My thought here was that API calls are going to fail IO or HTTP exceptions (which are derived from IO) and get reported back. All the other exceptions and errors are more likely internal to Spinnaker implementation/configuration rather than user requests, resource state, and permissions.

duftler: Sorry, that was a typo on my part. I intended to write: "If you are catching both `Exception` and `Error` I suppose you could just catch Throwable instead.".

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/GoogleExecutorTraits.groovy
duftler: If you've seen a similar pattern elsewhere in the neighborhood, I can live with it. Maybe `RuntimeException` would work too. I think you just want any unchecked exception here, and `RuntimeException` is the most general of those.

ewiseblatt:
FYI I changed things around to clean up and simplify as I mentioned off line yesterday.
The diff is "Refined and simplified" above. I'm going to submit since you already approved and saw the gist of the changes I planned on making.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/GoogleExecutorTraits.groovy
```diff
@@ -0,0 +1,76 @@
...
+     def success = "false"
+     Registry registry = getRegistry()
+     Clock clock = registry.clock()
+     long startTime = clock.monotonicTime()
+
+     try {
+       batch.execute()
+       success = "true"
+     } finally {
+       def tagDetails = [TAG_BATCH_CONTEXT: batchContext, "success": success]
```
duftler: Pretty sure you'll need to use `[(TAG_BATCH_CONTEXT): batchContext...]` here.

```
def testKey = 'some-key'
def testMapBad = [testKey: 'some-value']
def testMapGood = [(testKey): 'some-value']
System.out.println("testMapBad=$testMapBad")
System.out.println("testMapGood=$testMapGood")
```

Yields:
```
testMapBad=[testKey:some-value]
testMapGood=[some-key:some-value]
```

duftler: Also, the quotes can be dropped from any literal map keys like `"success"`.

ewiseblatt: Interesting (though not in a good way). That also explains why I dont need the quotes on success. Thanks for catching that.

I've added in batching here, but I havent ported the batch operations so havent tested this.

PR #1400, opened 2017-01-26
(aws)(titus) - change discovery integration to issue DELETE instead o…

tomaslin:
…f PUT for setting discovery status to be up and provides the default to be OUT_OF_SERVICE


tomaslin:
@cfieber @jeyrschabu @ajordens PTAL

cfieber:
:shipit:

PR #1402, opened 2017-01-27
Adding instance termination lifecycle to pull apps out of Eureka faster

robzienert:
Porting the code that I added into Eureka to update application service states on termination faster. The agent will create an SQS queue in a single account and every region that spinnaker knows about.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/InstanceTerminationLifecycleAgent.java
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/InstanceTerminationLifecycleAgent.java
@@ -1,0 +1,214 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.clouddriver.aws.lifecycle;
+
+import com.amazonaws.auth.policy.Condition;
+import com.amazonaws.auth.policy.Policy;
+import com.amazonaws.auth.policy.Principal;
+import com.amazonaws.auth.policy.Resource;
+import com.amazonaws.auth.policy.Statement;
+import com.amazonaws.auth.policy.actions.SQSActions;
+import com.amazonaws.services.sqs.AmazonSQS;
+import com.amazonaws.services.sqs.model.Message;
+import com.amazonaws.services.sqs.model.ReceiptHandleIsInvalidException;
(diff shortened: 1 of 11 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/InstanceTerminationLifecycleAgent.java
```diff
@@ -0,0 +1,220 @@
...
+      .filter(c -> c.getAccountId().equals(accountId))
+      .findFirst()
+      .orElseThrow((Supplier<RuntimeException>) () -> {
+        return new RuntimeException(String.format("Unable to find AmazonCredentials by id (id: %s)", accountId));
+      });
+  }
+
+  private static String ensureQueueExists(AmazonSQS amazonSQS, ARN queueARN, String sourceARN, Set<String> allAccountIds) {
+    String queueUrl;
+    try {
```
jeyrschabu: I believe the createQueue(name) will return the queue URL if it already exists. http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/InstanceTerminationLifecycleAgent.java
```diff
@@ -0,0 +1,220 @@
...
+          }
+        } catch (IOException e) {
+          log.error("Unable to convert NotificationMessage (body: {})", message.getBody(), e);
+        }
+
+        deleteMessage(amazonSQS, queueId, message);
+        messagesProcessed.incrementAndGet();
+      });
+
+      if (receiveMessageResult.getMessages().isEmpty()) {
```
jeyrschabu: IMO this would be more readable up top

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/InstanceTerminationLifecycleAgent.java
```diff
@@ -0,0 +1,220 @@
...
+              handleMessage(lifecycleMessage, TaskRepository.threadLocalTask.get());
+            } finally {
+              TaskRepository.threadLocalTask.set(originalTask);
+            }
+          }
+        } catch (IOException e) {
+          log.error("Unable to convert NotificationMessage (body: {})", message.getBody(), e);
+        }
+
+        deleteMessage(amazonSQS, queueId, message);
```
jeyrschabu: Do we want to delete the message if we fail to process it ?

robzienert: We don't want to keep processing messages that fail, otherwise we'll likely fall behind during larger / concurrent deployments. Letting the message go unprocessed isn't necessarily bad, since the default behavior will just be that the instance falls out of discovery a little slower (which is what the current behavior is).

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/InstanceTerminationLifecycleAgent.java
robzienert: You're right, I'll update. Not sure why I opted for the longer form.

PR #1405, opened 2017-01-31
(google) Avoid NPE when auto healer is specified but max unavailable …

duftler:
…is not.


duftler:
@danielpeach or @lwander please review.

PR #1424, opened 2017-02-13
fix(provider/google): Don't retry on 'successful' codes in SafeRetry.

jtk54:
`SafeRetry` used to retry once even if the initial HTTP status code was in the "don't retry" list. Now it doesn't.


duftler:
What about a code that is not in success or retry set? Shouldn't that fall right away?

jtk54:
A code not in the success or retry set is re-thrown on L115.

jtk54:
Or are you saying to just pull that bit out of the initial `while` as well?

jtk54:
Updated to fail fast on a code not in the success set nor the retry set. I kept the failsafe in the `while` just for sanity.

PR #1430, opened 2017-02-15
Delete ignored messages (like test notifications, etc)

robzienert:
We need to ensure that unwanted (or unknown) messages are removed from the queue so they don't get reprocessed.

@spinnaker/netflix-reviewers ptal


jrsquared:
:shipit:

PR #1436, opened 2017-02-17
Propagate 401 error for basic auth docker registry

EricJizbaMSFT:
There are two cases where a DockerRegistryAuthenticationException stating "Docker registry must support 'Bearer' authentication." was propagated instead of a 401, causing issues:
1. checkV2Availability always fails for a registry that only supports basic auth
2. If user specifies an incorrect password, the error thrown is the DockerRegistryAuthenticationException instead of the more accurate "401 Unauthorized" error

### DIFF
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
@@ -351,9 +351,31 @@ class DockerRegistryClient {
         }
       } catch (RetrofitError error) {
         if (error.response?.status == 401) {
-          dockerToken = tokenService.getToken(target, error.response.headers)
-          token = "Bearer ${dockerToken.bearer_token ?: dockerToken.token}"
-          response = withToken(token)
+          String authenticateHeader = null
+
+          error.response.headers.forEach { header ->
+            if (header.name.equalsIgnoreCase("www-authenticate")) {
+              authenticateHeader = header.value
+            }
+          }
+
+          if (!authenticateHeader) {
+            throw error
+          }
+
+          String bearerPrefix = "bearer "
+          String basicPrefix = "basic "
+          if (bearerPrefix.equalsIgnoreCase(authenticateHeader.substring(0, bearerPrefix.length()))) {
+            // If we got a 401 and the request requires bearer auth, get a new token and try again
+            dockerToken = tokenService.getToken(target, authenticateHeader.substring(bearerPrefix.length()))
+            token = "Bearer ${dockerToken.bearer_token ?: dockerToken.token}"
+            response = withToken(token)
+          } else if (basicPrefix.equalsIgnoreCase(authenticateHeader.substring(0, basicPrefix.length()))) {
+            // If we got a 401 and the request requires basic auth, there's no point in trying again
+            throw error
+          } else {
+            throw new DockerRegistryAuthenticationException("Docker registry must support 'Bearer' or 'Basic' authentication.")
+          }
         } else {
           throw error
         }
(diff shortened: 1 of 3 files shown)


EricJizbaMSFT:
@lwander PTAL. Here is a more detailed example for scenario 1 above:

For a basic auth registry, here is the current flow for checkV2Availability:

1. Call 'doCheckV2Availability()' without passing 'basicAuthHeader'
1. Receive 401 Unauthorized
1. Try to get a bearer token
1. Fail to parse www-authenticate because it's something like "Basic realm='Azure Container Registry'"
1. Throw DockerRegistryAuthenticationException "Docker registry must support 'Bearer' authentication."
1. Done

Here's the new flow:

1. Call 'doCheckV2Availability()' without passing 'basicAuthHeader'
1. Receive 401 Unauthorized
1. Propagate 401 error (aka don't try to get bearer token)
1. Catch 401 and call 'doCheckV2Availability()', but pass basicAuthHeader this time
1. Succeed (assuming the user name and password are correct)

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -351,9 +351,31 @@ class DockerRegistryClient {
         }
       } catch (RetrofitError error) {
         if (error.response?.status == 401) {
-          dockerToken = tokenService.getToken(target, error.response.headers)
-          token = "Bearer ${dockerToken.bearer_token ?: dockerToken.token}"
-          response = withToken(token)
+          String authenticateHeader = null
+
+          error.response.headers.forEach { header ->
+            if (header.name.equalsIgnoreCase("www-authenticate")) {
```
lwander: This bearer token specific code should probably live in the `DockerBearerTokenService`

lwander:
I'm recommending that old PR since it seems like that would have handled your existing flow prior the change it introduced, but it seemed to upset some registries.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
EricJizbaMSFT: It's not bearer token specific, though. It's the same logic for basic auth. The only 'bearer' logic is checking whether or not it's bearer

lwander: Good point - sounds good.

EricJizbaMSFT:
Haha I assumed there was some reasoning behind 'doCheckV2Availability', so I left that logic alone. I'm fine changing that back, but that only fixes scenario 1 above. You would still get the DockerRegistryAuthenticationException instead of a 401 Unauthorized if you pass incorrect credentials to a basic auth registry

lwander:
Which registries have you tested this against? I'd like to make sure a few have been tried before merging this.

EricJizbaMSFT:
I tested these scenarios:
1. Index.docker.io for a private repo with _in_correct creds (returns 401 as expected)
1. index.docker.io for a private repo with correct creds (succeeds)
1. azure container registry with _in_correct creds (returns 401 as expected)
1. azure container registry with correct creds (succeeds)

In the 'succeeds' scenarios, I verified against checkV2Availability and I created a test cluster using an image from that repo.

The unit test in DockerRegistryClientSpec covers index.docker.io for public repo (library/ubuntu)

lwander:
Sweet, I'll try quay & GCR

lwander:
@tomaslin could you double check this works with your internal registry?

EricJizbaMSFT:
@lwander @tomaslin any update on this?

lwander:
I want to hold off on merging until we know this won't break Netflix's container deployments.

tomaslin:
Sorry, was out of the office this week. Tested it with our registries and it's fine. LGTM

PR #1439, opened 2017-02-18
(k8s) SG is enabled when no LBs are attached

lwander:
@duftler FYI


lwander:
This also lists replica sets with no instances as disabled, since this is the more k8s-conventional way of looking at a running application.

PR #1440, opened 2017-02-19
(kubernetes) Complete enable/disable early when there are no pods to operate on

lwander:
@duftler 

In Kubernetes 1.5 pods are returned as a null object rather than an empty array.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/servergroup/AbstractEnableDisableKubernetesAtomicOperation.groovy
```diff
@@ -102,6 +102,11 @@ abstract class AbstractEnableDisableKubernetesAtomicOperation implements AtomicO
       }
     }
 
+    if (!pods || pods.size() == 0) {
```
duftler: I think `!pods` is sufficient here since it'll return `true` if it's `null` or empty.

lwander: Ah right, groovy

PR #1443, opened 2017-02-20
(google) Log error when caching target pool instance health instead o…

duftler:
…f throwing exception.


duftler:
@lwander please review.

lwander:
LGTM

PR #1445, opened 2017-02-20
support region-scoped amazon instance reservations

robzienert:
@ajordens PTAL


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReport.groovy
```diff
@@ -56,6 +56,8 @@ class AmazonReservationReport implements ReservationReport {
   static class OverallReservationDetail {
     String availabilityZone
     String instanceType
+    // Support for region-scoped RIs
+    String awsRegion
```
ajordens: Can this be `region` for consistency?

Otherwise LGTM

Should be able to spot check against a report from `main` once this is running in `prestaging`.

PR #1447, opened 2017-02-21
added gcloud install to Dockerfile

danielpeach:
@ttomsu please review.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,4 +11,17 @@ RUN GRADLE_USER_HOME=cache ./gradlew buildDeb -x test && \
   cd .. && \
   rm -rf workdir
 
+RUN apt-get -y update && \
+   apt-get -y install python2.7 wget unzip ca-certificates && \
+   wget -nv https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.zip && \
+   mkdir -p /builder && \
+   unzip -qq google-cloud-sdk.zip -d /builder && \
+   rm google-cloud-sdk.zip && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -11,4 +11,17 @@ RUN GRADLE_USER_HOME=cache ./gradlew buildDeb -x test && \
   cd .. && \
   rm -rf workdir
 
+RUN apt-get -y update && \
```
ttomsu: I suggest merging this into the previous `RUN` command. If you download it to `workdir`, and unzip it to `workdir/gcloud`, you won't have worry about removing the `.zip` file or its inflated contents, because it will get wiped away with `rm -rf  workdir`. As it is currently, you've left behind the `/builder` dir.

Dockerfile
```diff
@@ -11,4 +11,17 @@ RUN GRADLE_USER_HOME=cache ./gradlew buildDeb -x test && \
...
+   wget -nv https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.zip && \
+   mkdir -p /builder && \
+   unzip -qq google-cloud-sdk.zip -d /builder && \
+   rm google-cloud-sdk.zip && \
+   CLOUDSDK_PYTHON="python2.7" /builder/google-cloud-sdk/install.sh --usage-reporting=false \
+       --bash-completion=false \
+       --disable-installation-options && \
+   rm -rf ~/.config/gcloud
+
+ENV PATH "/usr/local/gcloud/google-cloud-sdk/bin:$PATH"
```
ttomsu: If possible, I'd prefer to keep it in `/opt/gcloud`, to keep everything of concern within the `/opt` directory.

PR #1452, opened 2017-02-22
Adding metrics for termination lifecycle lag

robzienert:
Discovery is receiving the termination events, but I'd like to know how long it's actually taking for all of this to happen so we can continue the termination process as soon as possible without losing the benefits.


ajordens:
👍 LGTM

PR #1457, opened 2017-02-23
feat(provider/aws): Add ALB Support

jrsquared:


jrsquared:
Converted `AtomicOperation.groovy` to Java since the new Java classes were having a hard time with const values.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -0,0 +1,265 @@
...
+          .withHealthyThresholdCount(targetGroup.healthyThreshold)
+          .withUnhealthyThresholdCount(targetGroup.unhealthyThreshold)
+          .withMatcher(new Matcher().withHttpCode(targetGroup.healthCheckMatcher))
+        )
+        task.updateStatus BASE_PHASE, "Target group created in ${loadBalancerName} (${targetGroup.name}:${targetGroup.port}:${targetGroup.protocol})."
+        createdTargetGroup = createTargetGroupResult.getTargetGroups().get(0)
+      } catch (AmazonServiceException e) {
+        def exceptionMessage = "Failed to create target group ${targetGroup.name} for ${loadBalancerName} - reason: ${e.errorMessage}."
+        task.updateStatus BASE_PHASE, exceptionMessage
+        amazonErrors << exceptionMessage
```
robzienert: I had to add some new logic (for classic ELB) to rollback removed listeners when creates fail. You'll likely want to do the same for the target groups as well.

jrsquared: Added rollback for removed listeners to this as well. Was a little more work since listener actions are by ARN and we don't pass the ARN around, I had to recreate the listener objects.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/DeleteAmazonLoadBalancerV2AtomicOperation.groovy
```diff
@@ -0,0 +1,73 @@
...
+import com.netflix.spinnaker.clouddriver.aws.deploy.description.DeleteAmazonLoadBalancerDescription
+import com.netflix.spinnaker.clouddriver.aws.security.AmazonClientProvider
+import com.netflix.spinnaker.clouddriver.data.task.Task
+import com.netflix.spinnaker.clouddriver.data.task.TaskRepository
+import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation
+import org.springframework.beans.factory.annotation.Autowired
+
+class DeleteAmazonLoadBalancerV2AtomicOperation implements AtomicOperation<Void> {
+
+  private static final String BASE_PHASE = "DELETE_ELB"
```
robzienert: Wouldn't naming this `DELETE_ELBV2` make more sense?

robzienert:
Finally made it through - do like the Java conversions. Everything looks good outside of that question (although totally optional imo). 👍

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/DeleteAmazonLoadBalancerV2AtomicOperation.groovy
jrsquared: Yes. Yes it would.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -0,0 +1,284 @@
...
+        loadBalancing.deleteListener(new DeleteListenerRequest().withListenerArn(it.listenerArn))
+        task.updateStatus BASE_PHASE, "Listener removed from ${loadBalancerName} (${it.port}:${it.protocol})."
+      }
+
+      // No unique identifier on listeners to know when to update vs create, so just remove ones that don't match exactly
+      def listenersToCreate = listeners.findAll { listener ->
+        existingListeners.find({ listener.compare(it, listenerToActions.get(listener)) }) == null
+      }
+
+      def createListener = { UpsertAmazonLoadBalancerV2Description.Listener listener, boolean isRollback ->
```
chongkong: It seems `isRollback` is not used? is there any intention for this?

jrsquared: Yep, to drive the log messages; fixed.

KristianOellegaard:
Will it work like the classic load balancer - or will different spinnaker apps be able to register with individual target groups and mount apps on e.g. /app1, /app2?

jrsquared:
The end goal will definitely be to allow different clusters register with individual target groups within an ALB.  If we allow clusters across applications to mount in the same load balancer is still under discussion.

robumberger:
I started looking at adding support for ALB a few weeks ago and was delighted to see this PR sitting there. What's holding up this PR? If additional testing or development needs to be done, I can help.

bradleyjames:
Would it possible to separate the conversion from to java from the ALB support?  If it's too painful I get it but it would make it simpler to review and digest.

cdoughty-r7:
We tested this branch and it appears to be solving our problems with ALBs.  We are able to now use an ALB with multiple target groups by creating a "targetGroups" array in the pipeline JSON and we're now able to get multiple ec2 instances routing through a single ALB.  

We'd love to get this merged asap so we can use this.  Please let us know if we can help in any way. Thanks for creating this!

jrsquared:
@cdoughty-r7 Did you test registering and deregistering instances from target groups as well?

rdickey-r7:
@jrsquared in what ways, exactly?  Do you have a list of test cases that you need confirmed?  IE: what buttons we should click and what to expect.  Answering for @cdoughty-r7 because I was part of the referenced "we".

jrsquared:
@rdickey-r7 There is no UI for it yet, so no buttons to be clicked. That will come soon! I was just curious if you were attaching instances/ASGs to the target groups via the pipeline JSON as well, or if you manually added those instances to the target groups via the AWS console.

rdickey-r7:
Oh!  Yeah, we created a pipeline and edited the json to add the target group:
pipeline_configure

Then we executed it and it rolled the asg as we expected:
pipeline_deploy

However now that I really look at it closely, our setup is that we have one ALB and three pipelines each with their own target group/asg.  It appears that the ASG is getting created with all three target groups in its list, as if it's getting the list from somewhere other than the json we specified because as you can see in the above json, we only specified one.  But here in the aws console, there's three:
aws_asg

We're still testing with this app so I don't know 100% if it's working like we want it but that seems like it could be an issue.

jrsquared:
@rdickey-r7 Ah, not sure if it's the only problem, but the pipeline should have _either_ `loadBalancers` or `targetGroups`. I think it's looking at the loadBalancers list, and if one of those is an ALB, grabbing all the targetGroups.  That was an old quick hack to make sure spinnaker didn't just blow up with ALBs. Good to know, I should probably add an exception if an ALB is in that list, because you can't attach ASGs to ALBs, just ELBs or Target Groups.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/converters/DeleteAmazonLoadBalancerAtomicOperationConverter.groovy
```diff
@@ -28,8 +31,22 @@ import org.springframework.stereotype.Component
...
+    if (description.loadBalancerType == null) {
+      description.loadBalancerType = AmazonLoadBalancerType.ELB_CLASSIC
+    }
+
+    switch (description.loadBalancerType) {
+      case AmazonLoadBalancerType.ALB:
+        return new DeleteAmazonLoadBalancerV2AtomicOperation(description)
+        break
+      case AmazonLoadBalancerType.ELB_CLASSIC:
+      default:
```
cfieber: I'd probably just throw an exception on the default case otherwise if some new type of load balancer gets added it won't silently slip through here and end up handled as a classic ELB

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/UpsertAmazonLoadBalancerClassicDescription.java
```diff
@@ -14,35 +14,30 @@
...
-  Integer healthTimeout = 5
-  Integer unhealthyThreshold = 2
-  Integer healthyThreshold = 10
-  Boolean crossZoneBalancing = Boolean.TRUE
+import java.util.List;
 
-  static class Listener {
-    enum ListenerType {
+public class UpsertAmazonLoadBalancerClassicDescription extends UpsertAmazonLoadBalancerDescription {
+  public List<Listener> listeners;
```
cfieber: these should be private with getter/setter (what the equivalent groovy was generating).

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/UpsertAmazonLoadBalancerDescription.java
```diff
@@ -14,22 +14,21 @@
...
-    return true
-  }
+public class UpsertAmazonLoadBalancerDescription extends AbstractAmazonCredentialsDescription {
+  public AmazonLoadBalancerType loadBalancerType = AmazonLoadBalancerType.ELB_CLASSIC;
 
-  @Override
-  String getPhaseName() {
-    return "REGISTER_INSTANCES_WITH_LB"
-  }
+  public String clusterName;
```
cfieber: same comment as above wrt the Field + accessor methods

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
```diff
@@ -116,9 +117,14 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
 
       def regionScopedProvider = regionScopedProviderFactory.forRegion(description.credentials, region)
 
-      def loadBalancers = lookupHelper().getLoadBalancersByName(regionScopedProvider, description.loadBalancers)
+      def loadBalancers = loadBalancerLookupHelper().getLoadBalancersByName(regionScopedProvider, description.loadBalancers)
```
cfieber: where did we land on handling the case of someone passing an ALB name as the load balancer name?

it looks like if we were going to continue to let that work that the targetGroupArns that get set below needs to be a union of loadBalancers.targetGroupArns + targetGroups.targetGroupARNs

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AbstractInstanceLoadBalancerRegistrationAtomicOperation.groovy
```diff
@@ -34,8 +31,7 @@ import groovy.transform.PackageScope
 import org.springframework.beans.factory.annotation.Autowired
 
 abstract class AbstractInstanceLoadBalancerRegistrationAtomicOperation implements AtomicOperation<Void> {
```
cfieber: I'm not sure partitioning this by the type of load balancer is the correct thing given how we use it.  We typically have a specific instance and say 'add to load balancer' or 'remove from load balancer' and expect that it will fully enable or disable that instance.

With this change I think we would end up making the caller (either deck or orca) know that it has to call an add/remove from load balancer as well as an add/remove from target group

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/AmazonClientInvocationHandler.java
```diff
@@ -274,6 +274,16 @@ public DescribeLoadBalancersResult describeLoadBalancers(DescribeLoadBalancersRe
         describe(request, "loadBalancerNames", "loadBalancers", LoadBalancerDescription.class));
   }
 
+  public com.amazonaws.services.elasticloadbalancingv2.model.DescribeLoadBalancersResult describeLoadBalancersV2() {
```
cfieber: I don't think this is necessary - this class is a proxy handler so it looks for the same method signature on the underlying SDK class (I don't think the elbv2 api call is describeLoadBalancersV2?)

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
cfieber: I see that loadBalancers.targetGroupArns is gone in this change set though

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AbstractInstanceLoadBalancerRegistrationAtomicOperation.groovy
jrsquared: Since you can't 'add to load balancer' in the ALB land, and you have to add to a 'target group', does it make sense to leave them the same?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
jrsquared: kill it with fire

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/converters/DeleteAmazonLoadBalancerAtomicOperationConverter.groovy
jrsquared: Per conversation, moved the check in the validator so it can bail early.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/AmazonClientInvocationHandler.java
jrsquared: Yep, removed.

cdoughty-r7:
I should state here that we have _not_ been able to validate this further after our discussion above with @rdickey-r7  where it was found we were using loadBalancers array and targetGroups array at the same time in our pipeline JSON.  

When we remove loadBalancers, the launched ec2 instances never attach to the ALB and fail to start from the ALB health check for the associated target group.  

We are still debugging this to see if the problem is on our side, but I wanted to at least state that there might be an issue here.

cdoughty-r7:
Ok, things are looking good still on this branch.  We had some confusion because we think at some point our code got reverted and we started using master again without intending to. 

We pulled this branch, rebuilt, and tested a pipeline with modified JSON to contain a "targetGroups" array with only one target group (from an ALB with multiple target groups), and the autoscaling group was created with the single specified targetGroup and  the ec2 instance successfully passed ALB health checks. 

We tested this with loadBalancers set to an empty array in the JSON and removing the loadBalancers array entirely from the JSON, and both passed similarly.

jrsquared:
Awesome! We have a little bit of work to do to make sure we aren't breaking current users of ALBs internally, then we can get this merged... thank you for kicking the tires!

PR #1458, opened 2017-02-23
Copies K8S load balancer lables over upon update

ttomsu:
@lwander PTAL


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/loadbalancer/UpsertKubernetesLoadBalancerAtomicOperation.groovy
```diff
@@ -63,6 +63,10 @@ class UpsertKubernetesLoadBalancerAtomicOperation implements AtomicOperation<Map
 
     serviceBuilder = serviceBuilder.withNewMetadata().withName(name)
 
+    def labels = existingService.metadata?.labels ?: [:]
```
lwander: `existingService` may be null here. Switch to `existingService?.metadata?.labels`

PR #1461, opened 2017-02-23
(appengine) allow config files to be specified in deploy operation

danielpeach:

Adds the ability to include the contents of a config yaml file in the upsert description.


clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperation.groovy
```diff
@@ -117,9 +122,37 @@ class DeployAppengineAtomicOperation implements AtomicOperation<DeploymentResult
 
     task.updateStatus BASE_PHASE, "Deploying version $versionName..."
     jobExecutor.runCommand(deployCommand)
+    deleteFiles(writtenFullConfigFilePaths)
```
lwander: Any chance `jobExecutor` will throw an exception? If so, `deleteFiles` should probably live in a `finally` block so we don't leave old configs dangling.

Also, doesn't `runCommand` run asynchronously? If so, it's possible the files will be deleted before the command is able to read them.

danielpeach: Yes to the first point - good suggestion.

I just verified that it is not asynchronous, so the deletion should be fine.

PR #1462, opened 2017-02-24
(openstack) add accound credentials builder

emjburns:
Adding a builder so that it is easy to instantiate credentials from Halyard without having to specify the arguments in the proper order.

@lwander @varikin @drmaas please take a look.


emjburns:
@lwander I'd like to leave it the way it is. This builder will be available for a refactor in the future, and I will take a deeper look in a bit (I have to investigate the openstack account options for an issue I'm having).

lwander:
Sounds good, I'll merge once the build finishes

PR #1466, opened 2017-02-27
(google) Be more tolerant of missing server group <-> load balancer r…

duftler:
…elationships.


duftler:
@lwander please review.

PR #1467, opened 2017-02-27
(appengine) change model for upsert load balancer description

danielpeach:


clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/converters/UpsertAppengineLoadBalancerAtomicOperationConverter.groovy
```diff
@@ -32,6 +33,13 @@ class UpsertAppengineLoadBalancerAtomicOperationConverter extends AbstractAtomic
   }
 
   UpsertAppengineLoadBalancerDescription convertDescription(Map input) {
-    AppengineAtomicOperationConverterHelper.convertDescription(input, this, UpsertAppengineLoadBalancerDescription)
+    UpsertAppengineLoadBalancerDescription description = AppengineAtomicOperationConverterHelper
+      .convertDescription(input, this, UpsertAppengineLoadBalancerDescription) as UpsertAppengineLoadBalancerDescription
+
+    if (description.splitDescription) {
+      description.split = AppengineModelUtil.convertTrafficSplitDescriptionToTrafficSplit(description.splitDescription)
```
lwander: Just wondering, why is this needed? Why not supply clouddriver with the right format?

danielpeach: We talked about this offline, and I have mixed feelings about it. It's ugly to make this transformation here, but since enable, disable, upsertLoadBalancer, and the App Engine API all represent a load balancer's allocations as a map, I think this solution limits the total number of transformations.

PR #1468, opened 2017-02-27
(openstack) bugfix: make credential build method public

emjburns:
Fixing accidental private method. @lwander PTAL


clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/security/OpenstackNamedAccountCredentials.groovy
```diff
@@ -201,7 +201,7 @@ class OpenstackNamedAccountCredentials implements AccountCredentials<OpenstackCr
       return this
     }
 
-    private OpenstackCredentials build() {
+    OpenstackCredentials build() {
```
lwander: This is package-private now, you probably want `public OpenstackCredentials build()`

emjburns: Good call thanks. Updated.

PR #1470, opened 2017-02-27
fix(aws): Tightening direct SQS IAM perms in termination events

robzienert:
Sets direct SQS lifecycle event messages to only allow from the
configured lifecycle hook source ARNs, which is more correct
behavior than accepting anything from autoscaling.

Also added more info into the discovery log message in case we need to search lag for a specific server group, etc.


jeyrschabu:
LGTM

PR #1472, opened 2017-02-28
Addresses an issue wherein Spinnaker is incorrectly cleaning up tags …

ajordens:
…when server groups are created

BasicAmazonDeployHandler is a singleton and _cannot_ have local state.

This change moves the event state from BasicAmazonDeployHandler to the
underlying DeployDescription (which is _not_ a singleton).


robzienert:
👍 Basically what I was doing.

PR #1474, opened 2017-02-28
(bugfix) change openstack credentials builder return type

emjburns:
@lwander PTAL


lwander:
Unit tests are failing

PR #1475, opened 2017-02-28
Adds configuration option to regionally disable edda

cfieber:
This is a bit messier than I hoped, having to wire EddaTimeoutConfig into a few spots where behaviour changes if edda is enabled. It didn't feel like sticking that right into the credentials object would be right either..

There are a few ignored checks for eddaEnabled in caching agents where there is already a reasonable fallback if we don't get a lastModified timestamp back from edda

### DIFF
--- a/clouddriver-docker/src/test/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClientSpec.groovy
+++ b/clouddriver-docker/src/test/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClientSpec.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.clouddriver.docker.registry.api.v2.client
 
+import spock.lang.Ignore
 import spock.lang.Shared
 import spock.lang.Specification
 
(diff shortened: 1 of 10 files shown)


clouddriver-docker/src/test/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClientSpec.groovy
```diff
@@ -26,6 +27,7 @@ import java.util.concurrent.TimeUnit
  * with an exception indicating a network or HTTP error, or will fail to load data
  * from dockerhub.
  */
+@Ignore
```
cfieber: unit tests with network dependencies ...

tomaslin:
LGTM

robzienert:
👍

clouddriver-docker/src/test/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClientSpec.groovy
robzienert: Would prefer to see the ignore include a body as well so we know why later on.

jordiclariana:
hi @cfieber. We have some timeout problems and we think this setting can help us. I guess these settings (eddaTimeoutConfig) are configurable via `clouddriver.yml`. If so, can you post an example on how to configure `eddaTimeoutConfig.disabledRegions`?

Thanks.

PR #1488, opened 2017-03-03
fix(aws) : region default resolution does not have to be statically initialized

cfieber:
This undoes some changes that were introduced in #1239 

The DEFAULT_REGION constant in AmazonClientProvider is there as an indicator that the region should be resolved to the current region, and that happens in AwsClientSupplier via SpinnakerAwsRegionProvider.

@qqshfox Can you confirm that this still works for you, provided you are running on an EC2 instance within cn-north-1 (to get the region from instance metadata) or you have set the AWS_REGION environment variable to cn-north-1 ?


cfieber:
Additional motivation for this: AmazonClientProvider is heavily used in unit tests and the current static initializer takes a while to time out attempting to hit the instance metadata endpoint.

PR #1491, opened 2017-03-05
perf(aws): configurable termination lifecycle agent poll time

robzienert:
Allows users to adjust the termination lifecycle agent poll time. We should see a huge gain in performance for our average hooks reducing this from the default.


jrsquared:
👍

PR #1493, opened 2017-03-07
perf(aws): refactor termination lifecycle agent to run as background thread

robzienert:
We've seen the agent style deployment fall behind while running
as an agent (running on a single instance in the cluster). This
refactor will run the termination lifecycle listener as a
background thread on every clouddriver instance, long-polling SQS
as quickly as possible.


jeyrschabu:
LGTM

PR #1494, opened 2017-03-07
chore(eureka): enhancing logging on eureka service failures

robzienert:
The new termination lifecycle worker is performing better, but there's odd behavior in our interactions with discovery, which leads to me to believe we may be getting throttled. These changes just add some more logging to help track down specifically what's going on so we have a better idea of what to tackle next.


clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/deploy/ops/AbstractEurekaSupport.groovy
```diff
@@ -164,6 +164,8 @@ abstract class AbstractEurekaSupport {
           throw re
         }
 
+        AbstractEurekaSupport.log.debug("[$phaseName] - Failed calling external service", re)
```
ajordens: Worth logging in the other two catch blocks?

PR #1500, opened 2017-03-07
fix(aws): Allow deploys to continue if lifecycles fail to create

robzienert:
Prevents the case where an incorrectly configured lifecycle blocks the progress of a deploy operation.

@cfieber PTAL


ajordens:
LGTM

PR #1502, opened 2017-03-07
perf(aws): termination lifecycle eureka perf improvements

robzienert:
A couple items in here:

1. Allow setting specific eureka maximum retry settings when calling from termination lifecycle worker (reducing defaults from 10 retries to 3).
2. Updating worker to only run on caching instances
3. Short-circuit eureka calls if the application name cannot be found from an ASG and if the ASG/instance no longer exists. This can happen when processing stale messages. Rather than waste cycles attempting to remove it from discovery, move on with life.

@ajordens PTAL


robzienert:

jrsquared:
👀

ajordens:
LGTM

PR #1504, opened 2017-03-08
feat(provider/kubernetes): Add hostNetwork option for pod creation.  …

chlung:
feat(provider/kubernetes): Add hostNetwork option for pod creation.

This is backend changes which will allow user to modify pipeline JSON file and add under stages -> clusters -> "hostNetowrk": true

When test it, the hostIP and podIP will match.  It still require UI work.  Per fabric8io suggestion, spinnaker should use io.fabric8:kubernetes-client jar not io.fabric8:kubernetes-api jar.  I don't observe any issue while testing, but it does need to further review.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
@@ -610,6 +610,8 @@ class KubernetesApiConverter {
       fromVolume(it)
     } ?: []
 
+    deployDescription.hostNetwork = replicaSet?.spec?.template?.spec?.hostNetwork
+
     deployDescription.containers = replicaSet?.spec?.template?.spec?.containers?.collect {
       fromContainer(it)
     } ?: []
@@ -640,6 +642,7 @@ class KubernetesApiConverter {
       .withNewScaleRef()
       .withKind(KubernetesUtil.SERVER_GROUP_KIND)
       .withName(description.serverGroupName)
+      .withHostName(description.hostName)
       .endScaleRef()
       .endSpec().build()
   }
@@ -895,6 +898,8 @@ class KubernetesApiConverter {
 
       podTemplateSpecBuilder = podTemplateSpecBuilder.withVolumes(volumeSources)
     }
+
+    podTemplateSpecBuilder = podTemplateSpecBuilder.withHostNetwork(description.hostNetwork)
 
     def containers = description.containers.collect { container ->
       toContainer(container)
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
@@ -642,7 +642,6 @@ class KubernetesApiConverter {
       .withNewScaleRef()
       .withKind(KubernetesUtil.SERVER_GROUP_KIND)
       .withName(description.serverGroupName)
-      .withHostName(description.hostName)
       .endScaleRef()
       .endSpec().build()
   }
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/description/servergroup/DeployKubernetesAtomicOperationDescription.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/description/servergroup/DeployKubernetesAtomicOperationDescription.groovy
@@ -32,6 +32,7 @@ class DeployKubernetesAtomicOperationDescription extends KubernetesAtomicOperati
   String namespace
   String restartPolicy
   Integer targetSize
+  Boolean hostNetwork
   List<String> loadBalancers
   List<String> securityGroups
   List<KubernetesContainerDescription> containers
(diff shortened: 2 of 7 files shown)


chlung:
@lwander Can you review this changes.  We can discuss about UI.  Thx.

Note:  There is an open issue related this.

lwander:
Great start, you are only missing a change in this function as well: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy#L866

chlung:
@lwander I see my changes in my own branch.  https://github.com/chlung/clouddriver/blob/feature/hostnetwork/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy#L902

It does not get merge into master for some reasons.  Do I need to resubmit all my changes with a new branch?  Thanks.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
```diff
@@ -640,6 +642,7 @@ class KubernetesApiConverter {
       .withNewScaleRef()
       .withKind(KubernetesUtil.SERVER_GROUP_KIND)
       .withName(description.serverGroupName)
+      .withHostName(description.hostName)
```
lwander: Is setting the host name what you intend for the autoscaler?

chlung:
@lwander Thanks.  I will update my code.

chlung:
@lwander I am re-thinking your question.  If autoscaling for multiple pods, should hostnetwork is applicable.  Is it correct?  I am new in Kubernetes as well.  Can you suggest?  Thx.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/description/servergroup/DeployKubernetesAtomicOperationDescription.groovy
```diff
@@ -32,6 +32,7 @@ class DeployKubernetesAtomicOperationDescription extends KubernetesAtomicOperati
   String namespace
   String restartPolicy
   Integer targetSize
+  Boolean hostNetwork
```
lwander: This should be in the `KubernetesSecurityContext` class.

lwander:
Since the `hostNetwork` is applied to each pod spec, you don't have to worry about the autoscaler applying it as well. I can't find any docs on what `hostName` does for the autoscaler, and it doesn't seem to be related to the pod spec `hostNetwork`. I would drop that.

chlung:
@lwander I updated the code.  Please review to see if it is ready to merge into master.  Also, I have question about UI.  Are you planning to have UI to support this extra field?  Thanks a lot.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/description/servergroup/DeployKubernetesAtomicOperationDescription.groovy
lwander: This comment still needs addressing.

lwander:
There's still one unaddressed comment.

You'll need to supply the UI changes if you want them soon. It shouldn't much work, here is the commit adding the other security options in the UI: https://github.com/spinnaker/deck/commit/29d107482b343338b0c2e83e3cc2d0f5b4a19b8a

chlung:
@lwander It is using in https://github.com/chlung/clouddriver/blob/feature/hostnetwork/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy#L902

static PodTemplateSpec toPodTemplateSpec(DeployKubernetesAtomicOperationDescription description, String name) {

Do you think that I need to remove it?  Thx.

lwander:
You just need to move it into the nested security class.

chlung:
From https://godoc.org/k8s.io/kubernetes/pkg/api/v1#PodSpec, it is under PodSpec.  Do you think that it will work if I move under SecurityContext.  Thx.

lwander:
Oh you're totally right, my bad. Let me run this locally to make sure the API change doesn't break anything unexpected.

chlung:
@lwander Great.  For hostNetwork UI changes, my team will discuss if we can wait or not.  I would like to hear your suggestion about where this hostNetwork option should be add.  Is it under Server Group?  Should we have a new session or use any of existing section, such as Advanced Settings?  Many thanks for you help.

cbuckley01:
It appears this change has broken Kubernates jobs:
{"exception":{"cause":"MissingPropertyException","message":"No such property: hostNetwork for class: com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.job.RunKubernetesJobDescription","operation":"RunKubernetesJobAtomicOperation","type":"EXCEPTION"},

chlung:
@cbuckley01 Do you have an example how to configure spinnaker to run a Kubernetes job.  So, I can verify if there is a bug and fix it.

cbuckley01:
Certainly, in a pipeline add a 'Run Job Stage'.  Using the same docker image etc.. that are deploying fine within clusters, the pipeline/stage should error out with the message above

chlung:
@cbuckley01  I miss the declaration part.  I don't think that I can use the same branch.  I will create a new pull quest and fix it.  Thanks for reporting this issue.  I follow your instruction to create pipeline. The pod is created, but the task never finishes.  How can I make the task completed.

cbuckley01:
just use a command of 'echo foo' or something with a 0 exit status

chlung:
@cbuckley01 Here is the new pull request - https://github.com/spinnaker/clouddriver/pull/1518.

PR #1505, opened 2017-03-09
fix(groovy): Groovyc making the static property private to child classes

robzienert:
@robfletcher PTAL


robfletcher:
Yep, if you don't do that in Groovy what you actually have is ```private static final DIMENSION_NAME_FOR_ASG = "AutoScalingGroupName";

public static getDIMENSION_NAME_FOR_ASG() {
  return DIMENSION_NAME_FOR_ASG;
}``` which seems to confuse IntelliJ and the compiler at odd times. Constants are literally the only time you should use the `public` keyword in Groovy, I think.

PR #1506, opened 2017-03-10
feat(aws): Adding conditional scaling policy copy method

robzienert:
Some components in Netflix code need to conditionally filter out scaling policies that are copied to a new ASG. For OSS, this has no effect.


jrsquared:
LGTM

PR #1507, opened 2017-03-10
chore(fiat): Bump Fiat API, add @EnableFiatAutoConfig annotation

ttomsu:
@cfieber please take for a spin


ttomsu:
This worked in the various other components. Merging

PR #1508, opened 2017-03-10
fix(google): Made instrumentation a little more consistent

ewiseblatt:
@jtk54

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/loadbalancer/DeleteGoogleSslLoadBalancerAtomicOperation.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/loadbalancer/DeleteGoogleSslLoadBalancerAtomicOperation.groovy
@@ -68,7 +68,7 @@ class DeleteGoogleSslLoadBalancerAtomicOperation extends DeleteGoogleLoadBalance
 
     List<ForwardingRule> projectForwardingRules = timeExecute(
         compute.globalForwardingRules().list(project),
-        "compute.globalForwardingRules",
+        "compute.globalForwardingRules.list",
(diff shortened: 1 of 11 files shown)


ewiseblatt:
Can you look at the last delta I did?
I wasnt able to submit this earlier because Travis never completed for some reason.
While trying to retrigger it I surveyed all the call sites and found another cluster of changes.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/loadbalancer/DeleteGoogleSslLoadBalancerAtomicOperation.groovy
```diff
@@ -173,7 +173,7 @@ class DeleteGoogleSslLoadBalancerAtomicOperation extends DeleteGoogleLoadBalance
         "Health check $healthCheckName",
         project,
         task,
-        BASE_PHASE,
+        [action: 'delete', operation: 'compute.healthChecks.delete', phase: BASE_PHASE, (TAG_SCOPE): SCOPE_GLOBAL],
```
jtk54: If these things are all of the same form, using an object here could be helpful.

ewiseblatt: I was tempted to do a big refactoring here but I dont want to break it. Instead I saved the fight for another day when I have more time. There are wrappers upon wrappers here and redundant parameters to go around. I have some thoughts about how to approach this
but probably not the time for a few weeks.

jtk54: SGTM.

PR #1509, opened 2017-03-10
fix(elasticsearch): Fix NPE generated when an entity tag type is null

robzienert:
Encountered an issue in test where entity tags was returning a value with a null value type. Filtering out these values.

@ajordens FYI


jeyrschabu:
LGTM

PR #1515, opened 2017-03-15
feat(web): ServerGroup view model post processor interface

robzienert:
Allows extension of mutating ServerGroup API response data. This is
needed to prevent Deck incorrectly rendering some internal features.


jrsquared:
👍

PR #1517, opened 2017-03-16
perf(aws): Simplifying termination lifecyle to not lookup app info first

robzienert:
After talking to cloudperf, we're mostly interested in processing messages
as quick as possible. We don't really care if discovery knows about the
application or instance, so we'll only retry on network errors, otherwise,
we'll just move onto the next message.


jrsquared:
👍

ajordens:
Suspect you're looking to run this on every instance to achieve max throughput? (saw you removed the condition)

I don't have a concern with that ... LGTM

robzienert:
@ajordens Argh, good catch. I was only removing that for local testing.

PR #1518, opened 2017-03-17
fix(kubenetesjob): Fixed missing updated hostNetwork option for job.

chlung:
Fixed missing hostNetwork filed for job which caused missing property exception during "Run a Job" deployment.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/description/job/RunKubernetesJobDescription.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/description/job/RunKubernetesJobDescription.groovy
@@ -30,6 +30,7 @@ class RunKubernetesJobDescription extends KubernetesAtomicOperationDescription {
   String stack
   String freeFormDetails
   String namespace
+  Boolean hostNetwork=false
   KubernetesContainerDescription container
   List<KubernetesVolumeSource> volumeSources
 }


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/description/job/RunKubernetesJobDescription.groovy
```diff
@@ -30,6 +30,7 @@ class RunKubernetesJobDescription extends KubernetesAtomicOperationDescription {
   String stack
   String freeFormDetails
   String namespace
+  Boolean hostNetwork=false
```
lwander: Is there a reason why you picked a nullable bool rather than `boolean`?

chlung:
@lwander From PodSpec, the default value is false and this is why I set the default to false instead of null.
HostNetwork bool `json:"hostNetwork,omitempty" protobuf:"varint,11,opt,name=hostNetwork"`
    // Use the host's pid namespace.
    // Optional: Default to false.

lwander:
Yup that makes sense, but I was asking because usually when you use a nullable type it's because `null` is an option. If it isn't and only `true`or `false` make sense, that should be a `boolean` instead.

chlung:
I am following fabric8 api.

public PodSpec(Long activeDeadlineSeconds,
               List<Container> containers,
               String dnsPolicy,
               Boolean hostIPC,
               Boolean hostNetwork,  ==> use Boolean not bool
...

lwander:
Sounds good

chlung:
@lwander Thanks.

PR #1519, opened 2017-03-20
feat(elasticsearch): EntityTags metadata endpoint

ajordens:
Reports on any differences between Front50 and Elasticsearch.

### DIFF
--- a/clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
+++ b/clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
@@ -26,12 +26,15 @@ import com.netflix.spinnaker.config.ElasticSearchConfigProperties;
 import io.searchbox.client.JestClient;
 import io.searchbox.client.JestResult;
 import io.searchbox.core.Bulk;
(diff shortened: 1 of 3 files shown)


ajordens:
PTAL @anotherchrisberry.

image

Putting together a little blog post of what we've done with `EntityTags` and this sort of a metadata endpoint came to mind as useful.

I haven't dug but I suspect we need some retries around the ElasticSearch operations. The bits that are in elasticsearch but not front50 all appear to be tests (`cbtest*` server groups).

ajordens:
Don't quite know why I went with snake case keys (_kid is really into snakes I guess_) ... happy to change (_it's an internal-only admin api and I thought it was easy to read_).

clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
```diff
@@ -342,6 +403,57 @@ private QueryBuilder applyTagsToBuilder(String namespace, Map<String, Object> ta
     }
   }
 
+  private List<EntityTags> fetchAll(String type, int scrollSize, String scrollTime) {
```
ajordens: @anotherchrisberry I believe we chatted about the elasticsearch scroll apis ... this seems to do the trick.

PR #1520, opened 2017-03-20
Fix Eureka integration (delete override) 

kirillsablin:
Currently clouddriver sends requests like 
DELETE /apps/{application}/{instanceId}/status?value=OUT_OF_SERVICE
to enable instance in eureka. 
Even though it seems like correct behaviour (remove previous disable state) but in reality such query deletes override and sets current status to OUT_OF_SERVICE effectively doing nothing with observable state.

To see real behaviour go to com.netflix.eureka.registry.AbstractInstanceRegistry::deleteStatusOverride

                    info.setOverriddenStatus(InstanceStatus.UNKNOWN);
                    info.setStatus(newStatus);

### DIFF
--- a/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/EnableAsgAtomicOperationUnitSpec.groovy
+++ b/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/EnableAsgAtomicOperationUnitSpec.groovy
@@ -84,7 +84,7 @@ class EnableAsgAtomicOperationUnitSpec extends EnableDisableAtomicOperationUnitS
                 app: "asg1"
             ]
         ]
-    1 * eureka.resetInstanceStatus('asg1', 'i1', AbstractEurekaSupport.DiscoveryStatus.Disable.value)
+    1 * eureka.resetInstanceStatus('asg1', 'i1', AbstractEurekaSupport.DiscoveryStatus.Enable.value)
   }
 
   def 'should skip discovery if not enabled for account'() {
--- a/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/discovery/DiscoverySupportUnitSpec.groovy
+++ b/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/discovery/DiscoverySupportUnitSpec.groovy
@@ -181,7 +181,7 @@ class DiscoverySupportUnitSpec extends Specification {
 
     0 * task.fail()
     instanceIds.each {
-      1 * eureka.resetInstanceStatus(appName, it, AbstractEurekaSupport.DiscoveryStatus.Disable.value) >> response(200)
+      1 * eureka.resetInstanceStatus(appName, it, AbstractEurekaSupport.DiscoveryStatus.Enable.value) >> response(200)
     }
 
     where:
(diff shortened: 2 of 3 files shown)


cfieber:
The current implementation is correct - we clear the explicit OUT_OF_SERVICE override and put a temporary OOS status in place, but that temp status is not an override so the next time the instance heartbeats it will reflect the instances status. This is essentially to prevent an OOS -> unknown -> instance's status and instead clients will just see OOS -> instance's status

kirillsablin:
According to 
com.netflix.eureka.resources.InstanceResource::renewLease
status param is completely ignored here (only overriddenstatus is used).
Also according to Eureka documentation overridenstatus is not required, so many Eureka clients will not add overriddenstatus to heartbeat (in fact Spring Cloud doesn't send it)

tomaslin:
I think you're confusing updateStatus with resetStatus.

resetInstanceStatus is a DELETE command. It removes the override so it the cloud level heartbeat status is passed through. This is different than an updateStatus because the update actually sets an override. 

This change was done by request of the Eureka team at Netflix. Your PR will actually cause a regression. Closing.

kirillsablin:
Could you please explain how instance should become UP after 
DELETE /apps/{application}/{instanceId}/status?value=OUT_OF_SERVICE. 
For me it seems that heartbeat call will not move it to UP status ( I can't find this logic in code).
May be there are some differences between published Eureka code and internal? 
I would appreciate any help/suggestion on this topic 

Thanks for your time!

tomaslin:
The value being passed in to the delete endpoint is the default value to set if the discovery record. If the record for the instance has a status it will return this instead of the default.

The instances need to heartbeat their instance status to discovery. The expected behavior is -> discovery record has override -> spinnaker removes override and says 'if there is no status, return OUT_OF_SERVICE until we know a status` -> instance heartbeats with UP / Down / etc -> instance status is returned. 

Basically the discovery instance will look at the status field whenever the overriden status is set to unknown.

ekesken:
but this is not the case if there is an OUT-OF-SERVICE status in lease, as you can see here:

https://github.com/Netflix/eureka/blob/master/eureka-core/src/main/java/com/netflix/eureka/registry/rule/LeaseExistsRule.java#L31

indeed, in current version the only way for status to be overridden by an heartbeat is setting it to UNKNOWN, as you can see here:

https://github.com/Netflix/eureka/blob/master/eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java#L362

and I can't see the usage of status parameter in renewRelease method as @kirillsablin mentioned:

https://github.com/Netflix/eureka/blob/master/eureka-core/src/main/java/com/netflix/eureka/resources/InstanceResource.java#L106

may there be differences between published eureka and internal one used in Netflix?

tomaslin:
This is possible.
 
This is the expected behavior and I can verify with our eureka integration:

Disabled in Discovery:

<overriddenStatus>OUT-OF-SERVICE</overriddenStatus>
<status>UP</status>

When I mark instances as enabled

<overriddenStatus>UNKNOWN</overriddenStatus>
<status>UP</status>

Are you seeing that the OverridenStatus remains OUT-OF-SERVICE after the enable call?

It might be helpful to post the entire discovery record for that instance in this issue as this might help debugging this (i.e, misconfigured lease setting, etc )

kirillsablin:
Problem is that, when clouddriver does 

DELETE /apps/{application}/{instanceId}/status?value=OUT_OF_SERVICE

Instance status become 
<overriddenStatus>UNKNOWN</overriddenStatus> <status>OUT_OF_SERVICE</status>

And it keeps this status (OUT_OF_SERVICE) even after heartbeats

kirillsablin:
This is full instance info 

```
{
  "instance": {
    "hostName": "ip-10-159-2-173.eu-west-2.compute.internal",
    "app": "APPLICATION-NAME",
    "ipAddr": "10.59.2.173",
    "status": "OUT_OF_SERVICE",
    "overriddenstatus": "UNKNOWN",
    "port": {
      "$": 8081,
      "@enabled": "true"
    },
    "securePort": {
      "$": 443,
      "@enabled": "false"
    },
    "countryId": 1,
    "dataCenterInfo": {
      "@class": "com.netflix.appinfo.AmazonInfo",
      "name": "Amazon",
      "metadata": {
        "accountId": "748575466205",
        "local-hostname": "ip-10-159-2-173.eu-west-2.compute.internal",
        "instance-id": "i-0ed66d4cbd3b39cee",
        "local-ipv4": "10.59.2.173",
        "instance-type": "t2.small",
        "vpc-id": "vpc-12312313",
        "ami-id": "ami-12312312",
        "mac": "0a:b9:c5:cf:17:83",
        "availability-zone": "eu-west-2b"
      }
    },
    "leaseInfo": {
      "renewalIntervalInSecs": 30,
      "durationInSecs": 90,
      "registrationTimestamp": 1490172897743,
      "lastRenewalTimestamp": 1490370333578,
      "evictionTimestamp": 0,
      "serviceUpTimestamp": 1490172897743
    },
    "metadata": {
      "@class": "java.util.Collections$EmptyMap"
    },
    "homePageUrl": "http://ip-10-159-2-173.eu-west-2.compute.internal:8081/",
    "statusPageUrl": "http://ip-10-159-2-173.eu-west-2.compute.internal:8081/info",
    "healthCheckUrl": "http://ip-10-159-2-173.eu-west-2.compute.internal:8081/health",
    "vipAddress": "application-name",
    "secureVipAddress": "application-name",
    "isCoordinatingDiscoveryServer": "false",
    "lastUpdatedTimestamp": "1490368843063",
    "lastDirtyTimestamp": "1490269248659",
    "actionType": "MODIFIED"
  }
}

```

qiangdavidliu:
Hi @kirillsablin from your full info, it looks like you are not providing an AWS asgName. The LeaseExistsRule you linked is only activated for a legacy case of when an asgName is not available within an EC2 environment. What sort of client are you using?

kirillsablin:
Hello, @qiangdavidliu we use client from Spring Cloud (spring-cloud-netflix-eureka-client:1.2.0)

qiangdavidliu:
The spring clients should expose ways to add the asgName, so you could try to make sure that field is properly populated.

PR #1524, opened 2017-03-21
Reducing the amount of times we repopulate the docker registry secret…

brujoand:
Currently we are seeing problems related to the aggressive repopulation of k8s secrets in our clusters. The result is that accounts will be unavailable in Spinnaker whenever a repopulation has failed, which it does many times each day. Probably due to a race condition between the CloudDriver instances. 

This is an attempt to reduce the number of repopulations. If we need to repopulate this often, I'd argue that a more explicit approach outside of `getNamespaces` would be more appropriate, and we would also need some sort of locking when acting upon a namespace. WDYT?


brujoand:
Need to reiterate this before submitting.

PR #1525, opened 2017-03-21
fix(kubernetes): Fix multi-instance clouddriver secret churn

lwander:
Previously when new clouddriver instances would startup, they would delete & recreate every docker registry secret. If this coincided with a deploy, it would fail because it wouldn't have the necessary image pull secrets. This fixes this by 
1. using edit rather than destroy + recreate to update.
2. not replacing secrets that don't have changes.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiAdaptor.groovy
@@ -361,9 +361,9 @@ class KubernetesApiAdaptor {
     }
   }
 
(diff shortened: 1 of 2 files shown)


chlung:
@lwander Does this fix allow to pull docker image using previously configured docker credential?  We can use Kubernetes to remove obsolete docker credential if we don't need it anymore?  Thx

lwander:
As long as the docker registry is still configured in spinnaker, it will make sure it's credentials are registered with Kubernetes.

chlung:
How about wipe out the cache?  Will it clean up the kubernetes secret?  Thx

lwander:
The secrets are stored in Kubernetes, so you'd have to wipe them out there

brujoand:
@lwander @duftler Any chance of getting this merged? We're being hit quit a bit by this so it would be great to get out of the way.

duftler:
This needs a rebase and can then be merged.

PR #1526, opened 2017-03-21
fix(docker): Prefer repository name over image name when loading tags

lwander:
It turns out some docker registries (e.g. artifactory) will respond to a load tags call with an image name that doesn't fully qualify that docker image. To compensate for this we augment the load tags call with the real repository name.


chlung:
@duftler Do you mind to review this pull request? Our team needs this fix.  Thx.

chlung:
Thanks everyone.

PR #1531, opened 2017-03-23
feat(titus): allow secure signature to be optional

tomaslin:


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
```diff
@@ -89,6 +89,10 @@ class TitusDeployHandler implements DeployHandler<TitusDeployDescription> {
         description.labels.put("interestingHealthProviderNames", description.interestingHealthProviderNames.join(",") )
       }
 
+      if (description.addSecureSignature == false) {
```
cfieber: does this need a null check? I see there is a default value on the field but it is a big b boolean so..

tomaslin:
We could just have it depend on label and not a field on the actual job description itself. Does feel slightly more fragile. 

This way potential customizers can just depend on their own labels and there is a contract between the two.

tomaslin:
closing -> just doing this via labels

PR #1532, opened 2017-03-25
feat(provider/oraclebmcs): Add initial scaffolding for Oracle BMCS pr…

simonlord:
Based on the discussions during the demo we gave last Wednesday here is the first PR for an Oracle BMCS provider. This is the smallest PR i felt i could create. 

It is essentially: 
- Gradle build script
- Credentials/config loading
- A no-op atomic operation (required by FeatureController - otherwise it won't start with just oraclebmcs enabled). This'll be removed as soon as the first real atomic operation gets added
- A health indicator
- Integration into clouddriver-web
- Added Oracle BMCS section to clouddriver.yaml

It passes './gradlew test', and when enabled with a valid account it will run the health indicator.

The only bit i know you're not going to like is the way we're adding the Oracle BMCS SDK dependency. At the moment it's only available via a zip file on the github releases area of the github repo. I didn't want to add the jars to git, so instead we're downloading on the fly, unzipping and adding to the gradle deps. This should be in a public maven repo soon.

We'll get a demo account setup for you to able to run this as soon as possible.

### DIFF
--- a/clouddriver-oracle-bmcs/clouddriver-oracle-bmcs.gradle
+++ b/clouddriver-oracle-bmcs/clouddriver-oracle-bmcs.gradle
@@ -1,0 +1,59 @@
+// Oracle BMCS SDK isn't published to any maven repo (yet!), so we manually download, unpack and add to compile/runtime deps
+def bmcsSDK = project.file('bmcs-sdk/lib/oracle-bmc-java-sdk-full-1.2.2.jar')
+def sdkDest = "bmcs-sdk"
+
+if (bmcsSDK.exists() == false) {
+  def sdkArchive = File.createTempFile("bmcsSDK","archive.zip")
+  sdkArchive << new URL("https://github.com/oracle/bmcs-java-sdk/releases/download/v1.2.2/oracle-bmcs-java-sdk.zip").openStream()
+
+  copy {
+    from zipTree(sdkArchive)
+    into sdkDest
+    include "**/*.jar"
+    exclude "**/*-sources.jar"
+    exclude "**/*-javadoc.jar"
+
+    // Scary but works. I think clouddriver deps in general need cleaning at some point
+    // Even without the oracle bmc sdk 3rd party deps there's still multiple javax.inject and commons-X JARs
+    exclude "**/*jackson*.jar"
+    exclude "**/*jersey*.jar"
+    exclude "**/hk2*.jar"
+    exclude "**/*guava*.jar"
+    exclude "**/commons*.jar"
+    exclude "**/javax*.jar"
+    exclude "**/aopalliance*.jar"
+    exclude "**/javassist*.jar"
+    exclude "**/slf*.jar"
+    exclude "**/osgi*.jar"
+    exclude "**/validation*.jar"
+    exclude "**/jsr305*.jar"
+    exclude "**/bcprov*.jar"
+
+  }
+}
+
+def bmcsSDKDeps = project.fileTree('bmcs-sdk/third-party/lib')
+
+task cleanSDK << {
+  project.file(sdkDest).deleteDir()
+}
+clean.dependsOn cleanSDK
+
+
+
+dependencies {
+  compile project(":clouddriver-core")
+  compile spinnaker.dependency('frigga')
+  compile spinnaker.dependency('bootActuator')
+  compile spinnaker.dependency('bootWeb')
+  compile files(bmcsSDK)
+  runtime files(bmcsSDKDeps)
+}
+
+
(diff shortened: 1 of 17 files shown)


clouddriver-oracle-bmcs/src/main/groovy/com/netflix/spinnaker/clouddriver/oraclebmcs/security/OracleBMCSCredentials.groovy
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (c) 2017 Oracle America, Inc.
+ *
+ * The contents of this file are subject to the Apache License Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * If a copy of the Apache License Version 2.0 was not distributed with this file,
+ * You can obtain one at https://www.apache.org/licenses/LICENSE-2.0.html
+ */
+package com.netflix.spinnaker.clouddriver.oraclebmcs.security
+
+class OracleBMCSCredentials {
```
lwander: Is this going to be extended in a future PR?

clouddriver-web/config/clouddriver.yml
```diff
@@ -266,6 +266,12 @@ cf:
       artifactUsername: ${cf.repo.username}
       artifactPassword: ${cf.repo.password}
 
+oraclebmcs:
+  enabled: ${ORACLEBMCS_ENABLED:false}
+  accounts:
```
lwander: This config file is a bit of a mixed bag, but I'd recommend not adding any accounts here since it could lead to spring-config confusion down the road. Arrays across multiple spring profiles are overwritten entirely instead of merged, and it's not always obvious what the profile-load-order is. The other provider (minus CF) use this file to define provider-wide defaults, which are a little easier to reason about and change less frequently than the set of enabled accounts.

clouddriver-oracle-bmcs/src/main/groovy/com/netflix/spinnaker/clouddriver/oraclebmcs/security/OracleBMCSNamedAccountCredentials.groovy
```diff
@@ -0,0 +1,109 @@
...
+      this.compartmentId = compartmentID
+      return this
+    }
+
+    OracleBMCSNamedAccountCredentials build() {
+      return new OracleBMCSNamedAccountCredentials(this.name, this.environment, this.accountType, this.requiredGroupMembership, this.compartmentId)
+    }
+  }
+
+  class OracleBMCSRegion {
```
lwander: Just curious, how do regions work in a bare-metal cloud?

clouddriver-oracle-bmcs/clouddriver-oracle-bmcs.gradle
```diff
@@ -0,0 +1,52 @@
+// Oracle BMCS SDK isn't published to any maven repo (yet!), so we manually download, unpack and add to compile/runtime deps
+def bmcsSDK = project.file('bmcs-sdk/lib/oracle-bmc-java-sdk-full-1.2.2.jar')
```
duftler: Can this also be removed as a result of a `./gradlew clean`?

simonlord: Good catch! Added a cleanSDK task and added it as a dependency for the clean task.

clouddriver-oracle-bmcs/src/main/groovy/com/netflix/spinnaker/clouddriver/oraclebmcs/security/OracleBMCSNamedAccountCredentials.groovy
simonlord: Same as other clouds really. Currently we have 1 Region with 3 Availability Domains. More info here https://docs.us-phoenix-1.oraclecloud.com/Content/General/Concepts/regions.htm

simonlord:
So I removed the OracleBMCSCredentials class as we don't actually use it for anything as we don't need to keep credentials in memory. However, that means i've had to update OracleBMCSNameAccountCredentials to implement AccountCredentials<Object>. Is that ok?

clouddriver-web/config/clouddriver.yml
simonlord: That's good to know. I've removed the accounts now.

lwander:
I can't think of a reason why not

PR #1535, opened 2017-03-28
fix(provider/google): Removed static function keyword in L7 upsert.

jtk54:
@varikin FYI.


jtk54:
Closes https://github.com/spinnaker/spinnaker/issues/1517

PR #1547, opened 2017-03-31
fix(core): Add NoopAtomicOperationConverter.

duftler:


duftler:
@lwander @jtk54 @danielpeach please review.

duftler:
This change will allow clouddriver to start up with no providers enabled, or with providers that define zero atomic operation converters. Will also emit a warning on startup.

PR #1549, opened 2017-03-31
fix(provider/google): Add pagination support for retrieving image res…

duftler:
…ources.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
@@ -82,32 +82,51 @@ class GCEUtil {
 
     def imageProjects = [projectName] + credentials?.imageProjects + baseImageProjects - null
     def sourceImage = null
+    // We want predictable iteration order that matches the order of insertion.
+    def imageProjectToNextPageTokenMap = new LinkedHashMap<String, String>()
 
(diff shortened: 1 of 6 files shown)


duftler:
@lwander @danielpeach or @jtk54 please review.

duftler:
fyi @edwinavalos @varikin

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -82,32 +82,51 @@ class GCEUtil {
...
-          for (def image : imageList.items) {
-            if (image.name == imageName) {
-              sourceImage = image
-            }
+    while (!sourceImage && imageProjectToNextPageTokenMap) {
+      def imageListBatch = buildBatchRequest(compute, clouddriverUserAgentApplicationName)
+      def imageListCallback = new JsonBatchCallback<ImageList>() {
+        @Override
+        void onFailure(GoogleJsonError e, HttpHeaders responseHeaders) throws IOException {
+          updateStatusAndThrowNotFoundException("Error locating $imageName in these projects: $imageProjects: $e.message", task, phase)
```
danielpeach: It feels a little odd to throw (rather than just log) the error here, since you might find the image in a different image project.

duftler: These aren't not found exceptions, only things like network or server-side errors.

PR #1550, opened 2017-03-31
feat(provider/google): Added caching for external IP addresses.

jtk54:
Caching IP addresses to suggest for forwarding rule IPs in LB upserts. The support for setting IPs exists in the LB upsert operations, but there is some frontend work to actually invoke that path. Currently we leave the IPs blank and use dynamic ephemeral IPs.

@varikin This is the first bit of https://github.com/spinnaker/spinnaker/issues/1519.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/GoogleInfrastructureProvider.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/GoogleInfrastructureProvider.groovy
@@ -34,6 +34,7 @@ class GoogleInfrastructureProvider extends AgentSchedulerAware implements Search
   final String providerName = GoogleInfrastructureProvider.name
 
   final Set<String> defaultCaches = [
+      ADDRESSES.ns,
       APPLICATIONS.ns,
       BACKEND_SERVICES.ns,
       CLUSTERS.ns,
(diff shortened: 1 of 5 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/GoogleInfrastructureProvider.groovy
```diff
@@ -55,6 +56,7 @@ class GoogleInfrastructureProvider extends AgentSchedulerAware implements Search
   ]
 
   final Map<SearchableResource, SearchableProvider.SearchResultHydrator> searchResultHydrators = [
+      (new GoogleSearchableResource(ADDRESSES.ns)): new AddressResultHydrator(),
```
danielpeach: Indentation looks off here.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleRegionalAddressCachingAgent.groovy
```diff
@@ -0,0 +1,86 @@
...
+    super(clouddriverUserAgentApplicationName,
+          credentials,
+          objectMapper,
+          registry)
+    this.region = region
+  }
+
+  @Override
+  CacheResult loadData(ProviderCache providerCache) {
+    List<Address> httpHealthCheckList = loadAddresses()
```
danielpeach: `addresses` would probably be a better variable name.

PR #1555, opened 2017-04-04
refactor(provider/google): Rely on GCE server-side filtering when querying image by name.

duftler:


duftler:
@ewiseblatt please review.

duftler:
fyi @varikin @edwinavalos This will avoid having to fetch all the images.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -80,54 +80,33 @@ class GCEUtil {
...
       }
 
-      imageProjectToNextPageTokenMap.each { imageProject, pageToken ->
-        def imagesList = compute.images().list(imageProject)
-
-        if (pageToken) {
-          imagesList = imagesList.setPageToken(pageToken)
+      @Override
+      void onSuccess(ImageList imageList, HttpHeaders responseHeaders) throws IOException {
+        if (!sourceImage && imageList.items) {
```
duftler: It's not random. We take the first one found by looking in: managed project, (optional) image projects, public cloud projects. And each project can only have at most one with that name.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -80,54 +80,33 @@ class GCEUtil {
...
-            imageProjectToNextPageTokenMap[imageProject] = imageList.nextPageToken
-          } else {
-            imageProjectToNextPageTokenMap.remove(imageProject)
-          }
-        }
+    def imageListBatch = buildBatchRequest(compute, clouddriverUserAgentApplicationName)
+    def imageListCallback = new JsonBatchCallback<ImageList>() {
+      @Override
+      void onFailure(GoogleJsonError e, HttpHeaders responseHeaders) throws IOException {
+        updateStatusAndThrowNotFoundException("Error locating $imageName in these projects: $imageProjects: $e.message", task, phase)
```
duftler: 404 won't ever be caught here, even if the image does not exist (true, throwing 'not found' doesn't seem perfect here).

What I want to avoid is getting the wrong image due to something like a network error. Just logging all of the errors wouldn't be right, but throwing on all of them is also not right either.

Dan also raised the idea of not throwing the error and just logging it. The order of precedence does take into account the order of the image projects. First, the managed project, then any image projects, and lastly the public base image projects.

Would you be ok with this going through as-is, since it's an optimization of the existing logic, and then I'll open a new bug to apply retry logic here and give more thought to the error-handling?

ewiseblatt: yea, as I noted offline, I'm ok as is. Just wanted to bring this to your attention.
I was assuming projects was essentially a set and not a list so adding or removing a project (even in another execution) would change the order even if the added/removed project did not contain the image in question. That would situation would be really nasty to debug.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
ewiseblatt: So why !sourceImage in the guard? Isnt !sourceImage an implicit invariant?

ewiseblatt:
LGTM

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
duftler: Projects is a set.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
duftler: `imageProjects` is also a list. Unless you saw it specified as a set somewhere and I've missed it?

PR #1560, opened 2017-04-06
fix(openstack): Fix to have the OpenstackImageV1Provider return all images when requested.

danveloper:
This fix changes the behavior of the `OpenstackImageV1Provider` to return all images from Openstack rather than just the first page.

This is necessary for Openstack environments with greater than 25 images.


duftler:
Needs a rebase.

duftler:
Can you please squash and rebase instead of merging master?
Then this is good to go.

PR #1562, opened 2017-04-07
feat(docker registry): endpoint to get tags sorted by date created

tomaslin:


clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/security/DockerRegistryNamedAccountCredentials.groovy
```diff
@@ -254,6 +254,13 @@ class DockerRegistryNamedAccountCredentials implements AccountCredentials<Docker
     return this.credentials?.client?.basicAuth ?: ""
   }
 
+  @JsonIgnore
+  List<String> getTags(String repository) {
+    return credentials.client.getTags(repository).tags.sort{ tag ->
+      credentials.client.getCreationDate(repository, tag)
```
lwander: This could cause problems with some of the public registries. I feature flagged the digest downloads because the 'get manifest' endpoint was surprisingly flaky. Could you put this sort behind that same digest/manifest feature flag?

tomaslin:
the flag is different than digest since we don't want to track all digests but be able to sort tags on demand

lwander:
SGTM

PR #1566, opened 2017-04-12
fix(provider/oraclebmcs): Handle missing instances

simonlord:
Fix handling when requested instance doesn't exist anymore. Groovy's .first() doesn't like empty collections, so i've added a check. Also added test case.

### DIFF
--- a/clouddriver-oracle-bmcs/src/main/groovy/com/netflix/spinnaker/clouddriver/oraclebmcs/provider/view/OracleBMCSInstanceProvider.groovy
+++ b/clouddriver-oracle-bmcs/src/main/groovy/com/netflix/spinnaker/clouddriver/oraclebmcs/provider/view/OracleBMCSInstanceProvider.groovy
@@ -51,8 +51,8 @@ class OracleBMCSInstanceProvider implements InstanceProvider<OracleBMCSInstance>
     if (identifiers.size() > 1) {
       log.warn("There should be at most one identifier!")
     }
-    Set<OracleBMCSInstance> instances = loadInstances(identifiers)
-    return instances.first()
+    Set<OracleBMCSInstance> instances = loadInstances(identifiers) ?: null
+    return instances?.first()
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


clouddriver-oracle-bmcs/src/main/groovy/com/netflix/spinnaker/clouddriver/oraclebmcs/provider/view/OracleBMCSInstanceProvider.groovy
```diff
@@ -52,7 +52,7 @@ class OracleBMCSInstanceProvider implements InstanceProvider<OracleBMCSInstance>
       log.warn("There should be at most one identifier!")
     }
     Set<OracleBMCSInstance> instances = loadInstances(identifiers)
-    return instances.first()
+    return instances?.isEmpty() ? null: instances.first()
```
duftler: Would be more idiomatic to use: `return instances?.first()`

That would also avoid an NPE when `instances` is `null` (which I'm guessing it can't be, but still seems safer that way).

simonlord: The problem is that `[].first()` will still throw a NoSuchElementException, which is what i'm trying to safeguard against. It is a shame IMO that the .first() groovy addon method doesn't just return null when the collection is empty.

duftler: Good point.
How about:
```
Set<OracleBMCSInstance> instances = loadInstances(identifiers) ?: null
return instances?.first()
```

simonlord: Yes, that's nice - i'll add that now

PR #1564, opened 2017-04-07
feat(appengine): upsert autoscaling policy

danielpeach:
@lwander or @duftler or @jtk54 please review.

PR #1568, opened 2017-04-12
fix(appengine): capacity values for automatic scaling

danielpeach:

A change in https://github.com/spinnaker/clouddriver/pull/1564 caused the capacity values to sometimes be `null`, which caused an NPE in Orca, which broke several operations. This fixes that.

### DIFF
--- a/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/model/AppengineServerGroup.groovy
+++ b/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/model/AppengineServerGroup.groovy
@@ -88,10 +88,8 @@ class AppengineServerGroup implements ServerGroup, Serializable {
         * For the flexible environment, a version using automatic scaling can be stopped.
         * A stopped version scales down to zero instances and ignores its scaling policy.
         * */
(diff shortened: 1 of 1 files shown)


clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/model/AppengineServerGroup.groovy
```diff
@@ -114,6 +112,26 @@ class AppengineServerGroup implements ServerGroup, Serializable {
     }
   }
 
+  Integer computeMinForAutomaticScaling(AppengineScalingPolicy scalingPolicy) {
+    Integer instanceCount = instances?.size() ?: 0
+    if (servingStatus == ServingStatus.SERVING) {
+      def candidateMinValues = [scalingPolicy.minIdleInstances, scalingPolicy.minTotalInstances, instanceCount].findAll { it != null }
```
danielpeach: The compiler doesn't like `[...] - null` here. Not sure why.

lwander: groovy

PR #1572, opened 2017-04-12
feat(provider/kubernetes): Specify termination grace period seconds

parajanoff:
Adds clouddriver support for https://github.com/spinnaker/spinnaker/issues/1107

@spinnaker/reviewers


lwander:
Thanks for this @parajanoff :)

PR #1573, opened 2017-04-12
fix(provider/kubernetes): Get node selector from ReplicaSet/Replicati…

parajanoff:
The Server Group Information panel (Deck) contains section that is responsible for displaying Node Selectors settings, but it is always hidden as node selector is not being retrieved from the ReplicaSet/ReplicationController. This fixes the issue https://github.com/spinnaker/spinnaker/issues/1264

@spinnaker/reviewers


parajanoff:
@lwander I've updated the branch to resolve the conflicts. 

This is a small one. Could you please check if convenient? Thanks!

lwander:
Thanks :)

PR #1588, opened 2017-04-22
Ensure that `EntityTags.tags` is always a modifiable list

ajordens:
ElasticSearchServerGroupTagger supplies a singletonList(entityTag) which
causes an issue when updated tags are merged over existing tags.

Opt'd for a generic fix in `EntityTags.setTags()` rather than a point
fix in `ElasticSearchServerGroupTagger` to prevent a future recurrence.


ajordens:
FYI @anotherchrisberry ... noticed this issue when merging existing tags over updated tags.

Symptom was server groups not getting tagged as expected with launch failures.

Quite possible/Almost certainly this was always broken but made worse when the `spinnaker:metadata` tag was applied everywhere (_which means that we almost always have existing tags to merge_).

https://github.com/spinnaker/clouddriver/blob/master/clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/ElasticSearchServerGroupTagger.java#L169

PR #1593, opened 2017-04-24
fix(provider/azure): Default environment and accountType to accountName

EricJizbaMSFT:
I was testing out some account behavior in deck (primary vs. secondary accounts) when I noticed Azure behaved differently than all other providers. Now the primary Azure credential will show up as a primary account by default.


EricJizbaMSFT:
@rguthriemsft @dcaro @clguimanMSFT PTAL

rguthriemsft:
LGTM

EricJizbaMSFT:
@spinnaker/reviewers PTAL

PR #1597, opened 2017-04-25
fix(core/cache): Allow clouddriver to be started without redis by mak…

duftler:
…ing CoreProvider conditionally constructed.


duftler:
@ajordens @cfieber or @robzienert please review.

duftler:
This is consistent with: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/RedisCacheConfig.groovy#L48

duftler:
Without this change, it tries to construct `CoreProvider` but fails since `JedisSource` is guarded by the presence of `redis.connection`.

PR #1599, opened 2017-04-25
feat(core/model): Add category to EntityTag model

christopherthielen:
This allows a tag to be assigned a category. Alerts/notifications will then be grouped by category in deck.


ajordens:
LGTM

PR #1600, opened 2017-04-25
chore(build): fix build config for oracle bmcs cloudprovider

cfieber:
moves logic out of inline gradle script and into tasks wired into the task graph

The previous config suffered from an evaluation order problem where the SDK unzipping would happen as the build script was interpreted while the clean task would happen during task execution so doing a `./gradlew clean build` would extract the SDK then clean it up then try to do the build

@simonlord PTAL

### DIFF
--- a/clouddriver-oracle-bmcs/clouddriver-oracle-bmcs.gradle
+++ b/clouddriver-oracle-bmcs/clouddriver-oracle-bmcs.gradle
@@ -1,55 +1,67 @@
-// Oracle BMCS SDK isn't published to any maven repo (yet!), so we manually download, unpack and add to compile/runtime deps
-def bmcsSDK = project.file('bmcs-sdk/lib/oracle-bmc-java-sdk-full-1.2.2.jar')
-def sdkDest = "bmcs-sdk"
 
-if (bmcsSDK.exists() == false) {
-  def sdkArchive = File.createTempFile("bmcsSDK","archive.zip")
-  sdkArchive << new URL("https://github.com/oracle/bmcs-java-sdk/releases/download/v1.2.2/oracle-bmcs-java-sdk.zip").openStream()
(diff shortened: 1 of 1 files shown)


simonlord:
Thanks @cfieber this looks great. I should have foreseen this problem.

simonlord:
Just a couple thoughts/questions:

- Does this download the sdk on every build? 
- Does it handle the sdk version changing in the url and notice it needs to download a newer version? I guess if it is downloaded every time a build occurs then this is fine.

cfieber:
The download and unpack tasks will noop provided the inputs and outputs are the same:
```
:clouddriver-oracle-bmcs:fetchSdk UP-TO-DATE
:clouddriver-oracle-bmcs:unpackSdk UP-TO-DATE
:clouddriver-oracle-bmcs:compileJava UP-TO-DATE
```
If the input URL changes it would re-run the download task and re-extract the SDK (it might do something bad if you changed the URL and didn't run a `clean` though in that I don't think it cleans the destination directory when extracting the zip - would be worth checking out the behaviour there)

I was thinking a bit more about this in light of spinnaker/spinnaker-dependencies#84 and I think that until this artifact is going to land in a maven repository for consumption (any ETA on that) then this probably needs to go a step further to shade or embed the SDK dependency into clouddriver-oracle-bmcs because otherwise any consumer of that library (halyard in this case) is going to have to re-do this dependency download business.

Any thoughts on how you want to proceed there?

cfieber:
changed the Copy task to a Sync task to address the issue of the zip file contents changing

simonlord:
@cfieber I'm afraid we can't do that due to legal ramifications. I guess for the time being it just means you won't be able to access the SDK classes via the `compile spinnaker.dependency(clouddriverOracleBmcs)`. I'll chase up the contact i have on how the maven release is coming along.

PR #1602, opened 2017-04-26
fix(provider/oraclebmcs): Add downloadSDK as a task

jervi:
It was not possible to run `gradlew clean build` because the SDK was downloaded before the clean task was executed

FYI @simonlord


simonlord:
Thanks for this @jervi I should of thought of that.

jervi:
Just noticed PR #1600. This should probably be closed as a duplicate.

PR #1605, opened 2017-04-27
fix(provider/google): Support on-demand cache updates for GCE Network Load Balancers.

duftler:
fix(provider/google): Support on-demand cache updates for GCE Network Load Balancers.


duftler:
Bunch of refactoring here as well to be explicit about commonality across GCE load balancers and their caching agents.

duftler:
@lwander @jtk54 ready for review. PTAL.

duftler:
Feel free to review when you have time, but please don't merge yet. Doing some additional testing and seeing a couple of weird things in the ui.

duftler:
Ok, resolved the issue I was seeing so deleting the prior comment.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleSslLoadBalancerCachingAgent.groovy
```diff
@@ -227,7 +191,7 @@ class GoogleSslLoadBalancerCachingAgent extends AbstractGoogleCachingAgent imple
         healthCheckRequest: healthCheckRequest,
         groupHealthRequest: groupHealthRequest,
         subject: googleLoadBalancer.name,
-        failedSubjects: failedLoadBalancers,
+        failedSubjects: failedSubjects,
```
jtk54: I think `failedSubjects`should be `failedLoadBalancers` here. Seems like it's not consistent what's passed into the constructors as `failedSubjects`.

duftler: `failedLoadBalancers` used to be an instance field within this caching agent. It isn't any longer, but `failedSubjects` holds the same reference since `TargetSslProxyCallback ` is a `FailedSubjectChronicler`.

PR #1607, opened 2017-04-29
fix(provider/google): Support on-demand cache updates for GCE Security Groups (Firewalls).

duftler:
Just the 2nd commit needs to be reviewed. The first commit is from the earlier pending PR.


duftler:
Need to fix a couple of things. Closing for now.

PR #1612, opened 2017-05-02
feat(provider/openstack): consul enable/disable

emjburns:
Adding consul enable/disable for openstack provider modeled after the google implementation (https://github.com/spinnaker/clouddriver/blob/master/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/AbstractEnableDisableAtomicOperation.groovy#L83).

Adding try/catch for getHealths in OpenStack caching.

@lwander PTAL

### DIFF
--- a/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/ops/servergroup/AbstractEnableDisableOpenstackAtomicOperation.groovy
+++ b/clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/ops/servergroup/AbstractEnableDisableOpenstackAtomicOperation.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.clouddriver.openstack.deploy.ops.servergroup
 
+import com.netflix.spinnaker.clouddriver.consul.deploy.ops.EnableDisableConsulInstance
 import com.netflix.spinnaker.clouddriver.data.task.Task
 import com.netflix.spinnaker.clouddriver.data.task.TaskRepository
 import com.netflix.spinnaker.clouddriver.openstack.client.BlockingStatusChecker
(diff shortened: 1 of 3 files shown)


clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/ops/servergroup/AbstractEnableDisableOpenstackAtomicOperation.groovy
```diff
@@ -54,7 +59,29 @@ abstract class AbstractEnableDisableOpenstackAtomicOperation implements AtomicOp
...
+
+    if (credentials.credentials.consulConfig?.enabled) {
+      task.updateStatus phaseName, "$gerund server group in Consul..."
+
+      List<String> instanceIds = provider.getInstanceIdsForStack(description.region, description.serverGroupName)
+      instanceIds.each { String instanceId ->
+        Server instance = provider.getServerInstance(description.region, instanceId)
+        try {
+          EnableDisableConsulInstance.operate(credentials.credentials.consulConfig,
+            instance.name,
```
lwander: Does openstack resolve instance IPs based on name the same way GCP does?

emjburns: If you have resolvable hostnames, openstack can resolve IP based on "name". So, I think yes. We don't have resolvable hostname right now, so we're putting in a hack to get it to resolve the same way I'd assume GCP does it.

lwander: Sounds good

PR #1613, opened 2017-05-02
fix(kubernetes): handle images without registry

danielpeach:
@lwander please review.

@ethanfrogers FYI

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/KubernetesUtil.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/KubernetesUtil.groovy
@@ -109,7 +109,11 @@ class KubernetesUtil {
   }
 
   static String getImageId(String registry, String repository, String tag) {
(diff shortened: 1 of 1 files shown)


lwander:
Just curious, what's prompting this?

ethanfrogers:
@lwander 
image

lwander:
How did that happen? There isn't a clear standard (AFAIK) around resolving that registry. Docker CLI will assume `index.docker.io` and K8S will assume `gcr.io`, which is why I'm nervous about assuming some default.

ethanfrogers:
@danielpeach and I have been working through an issue in Deck that prevents images like `nginx:latest` or `ethanfroger/spindemo:1.0.0` from being populated when Cloning a ServerGroup or using it as template. I have preexisting Deployments that use this image syntax(?) that aren't being populated. We believe we've fixed the problem in Deck but when we both cloned an SG with that style image it resulted in the above image.

lwander:
Ok makes sense - but are those `nginx:latest` images coming from Spinnaker, or other sources?

ethanfrogers:
They're preexisting Deployments we have in Kubernetes.

lwander:
Ah fair, LGTM!

PR #1618, opened 2017-05-03
fix(provider/openstack) adds name nullcheck to filtering images

emjburns:
Fixes images not loading when one of them has no name.


emjburns:
@lwander PTAL, thanks!

PR #1622, opened 2017-05-10
fix(provider/google): Wait for forwarding rule creation in LB upserts.

jtk54:
Orca's orchestration for upserting a Google load balancer does not contain a task to wait for the state of the platform to show that a load balancer was created (for good reason, that would be a complicated operation). Instead, Orca waits for Clouddriver to execute this operation and do a force cache refresh. We should wait for the whole load balancer to be created in the platform before we exit this upsert operation, so we wait for the forwarding rule to be created before continuing so we _know_ the state of the platform when we do a force cache refresh.


jtk54:
The *more* correct fix for this issue is a more widespread orca change to wait for the platform state (for each platform), but that would be quite a lot of work and require core Orca changes.

duftler:
...and that the unit tests pass.

jtk54:
I'm going to start a Spock band. It's going to be called "Mocky and the Counting Calls".

PR #1626, opened 2017-05-12
Fix issue 1632: not able to find AMI with encrypted snapshots in target account

jcusa:
This is a proposed fix for issue: [https://github.com/spinnaker/spinnaker/issues/1632](url)

During the allow atomic launch process, it was only looking for AMI using the source account. Normally it works fine as the AMIs can be shared between accounts. However, when using EBS backed AMI, AWS doesn't allow sharing. The proposed fix is to also make an attempt to look for the AMI in the target account where we are doing the deployment (as in the EBS case, the AMI will have to be in the same account where the deployment happens anyway, due to the limitation of not allowing sharing). Also, for EBS AMI, AWS gives error when you make any permission modifications on the AMI, even to add the same accountId as the owner of AMI.  Since the launch will happen in the same account, there's no need to do the permission update either anyway. Thus, added condition to just return AMI when ownerId and target accounts are the same.


jcusa:
checking failures

PR #1628, opened 2017-05-13
Fix issue 1632: not able to find AMI in target account during deploy

jcusa:
This is a proposed fix for [https://github.com/spinnaker/spinnaker/issues/1632](url)
During the allow atomic launch step, it was only looking for AMI using the source account. Normally it works fine as the AMIs can be shared between accounts. However, when using EBS backed AMI, AWS doesn't allow sharing. The proposed fix is to also make an attempt to look for the AMI in the target account where we are doing the deployment (as in the EBS case, the AMI will have to be in the same account where the deployment happens anyway, due to the limitation of not allowing sharing). If the AMI is found there, we can return with it as the rest is no-op.


robzienert:
Sharing EBS-backed AMIs across accounts is allowed (Netflix does it this way, for example), but the work itself seems valid - may as well skip the op if it doesn't need to happen.

👍

PR #1629, opened 2017-05-13
feat(amazon): A 'v3' reservation report that attempts to use regional…

ajordens:
… reserved instances to cover zonal shortfalls

It accomplishes this by:
- normalizing all regional reserved instances to `xlarge`
- using some number of `xlarge` reservations to cover each zonal shortfall

Shortfalls are covered in descending order:
- us-west-2c > us-west-2b : us-west-2a
- m4.16xlarge > m4.8xlarge > m4.xlarge > m4.large

Relies on the fact that AWS credits regional reserved instances towards a shortfall of any instance type.

Looking for feedback from @robzienert

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReport.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReport.groovy
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.clouddriver.aws.model
 import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.fasterxml.jackson.annotation.JsonPropertyOrder
+import com.fasterxml.jackson.annotation.JsonView
 import com.netflix.spinnaker.clouddriver.model.ReservationReport
 
 import java.util.concurrent.ConcurrentHashMap
@@ -26,12 +27,22 @@ import java.util.concurrent.atomic.AtomicInteger
 
 @JsonPropertyOrder(["reservations", "start", "end", "accounts", "type"])
 class AmazonReservationReport implements ReservationReport {
+  @JsonView(Views.V1.class)
   Date start
+
+  @JsonView(Views.V1.class)
   Date end
+
+  @JsonView(Views.V1.class)
   String type = "aws"
 
+  @JsonView(Views.V1.class)
   Collection<Map> accounts = []
+
+  @JsonView(Views.V1.class)
   Collection<OverallReservationDetail> reservations = []
+
+  @JsonView(Views.V1.class)
   Map<String, Collection<String>> errorsByRegion = [:]
 
   static enum OperatingSystemType {
@@ -52,31 +63,88 @@ class AmazonReservationReport implements ReservationReport {
     }
   }
 
+  static class Views {
+    static class V1 {
+
+    }
+
+    static class V2 extends V1 {
+
+    }
+
+    static class V3 extends V2 {
+
+    }
+  }
+
   @JsonPropertyOrder(["availabilityZone", "region", "availabilityZoneId", "instanceType", "os", "totalReserved", "totalUsed", "totalSurplus", "details", "accounts"])
   static class OverallReservationDetail {
     String availabilityZone
+
+    @JsonView(Views.V1.class)
     String instanceType
+
     // Support for region-scoped RIs
     String region
+
+    @JsonView(Views.V1.class)
     OperatingSystemType os
+
+    @JsonView(Views.V1.class)
     AtomicInteger totalReserved = new AtomicInteger(0)
+
+    @JsonView(Views.V1.class)
     AtomicInteger totalUsed = new AtomicInteger(0)
 
+    AtomicInteger totalAllocated = new AtomicInteger(0)
+
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    @JsonView(Views.V1.class)
     Map<String, AccountReservationDetail> accounts = new ConcurrentHashMap<>()
 
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    @JsonView(Views.V3.class)
+    List<Allocation> allocations = []
+
     @JsonProperty
+    @JsonView(Views.V3.class)
+    String id() {
+      return "${region}:${availabilityZoneId()}:${instanceType}:${os.name.toLowerCase()}"
+    }
+
+    @JsonProperty
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    @JsonView(Views.V3.class)
+    Integer totalAllocated() {
+      return totalAllocated.intValue()
+    }
+
+    @JsonProperty
+    @JsonView(Views.V1.class)
     int totalSurplus() {
-      return (totalReserved.intValue() - totalUsed.intValue())
+      return (totalReserved.intValue() + totalAllocated.intValue() - totalUsed.intValue())
     }
 
     @JsonProperty
+    @JsonView(Views.V1.class)
     String region() {
-      return availabilityZone == null ? region : availabilityZone[0..-2]
+      return (availabilityZone == null || availabilityZone == '*') ? region : availabilityZone[0..-2]
     }
 
     @JsonProperty
+    @JsonView(Views.V1.class)
     String availabilityZoneId() {
-      return availabilityZone == null ? '*' : availabilityZone[-1..-1]
+      return (availabilityZone == null || availabilityZone == '*') ? '*' : availabilityZone[-1..-1]
+    }
+
+    @JsonProperty
+    @JsonView(Views.V1.class)
+    String availabilityZone() {
+      return availabilityZone == null ? '*' : availabilityZone
+    }
+
+    String instanceFamily() {
+      return instanceType.substring(0, instanceType.indexOf('.'))
     }
 
     AccountReservationDetail getAccount(String accountName) {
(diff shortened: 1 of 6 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReport.groovy
```diff
@@ -61,22 +61,50 @@ class AmazonReservationReport implements ReservationReport {
...
 
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
     Map<String, AccountReservationDetail> accounts = new ConcurrentHashMap<>()
 
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    List<Borrow> borrows = []
+
+    @JsonProperty
+    String id() {
+      return "${region}:${availabilityZoneId()}:${instanceType}:${os.name.toLowerCase()}"
```
ajordens: This new `id` attribute will appear in the existing `v1` and `v2` reports. 

It seems useful to me so I didn't go out of my way (_like I did for `totalBorrowed`_) to make it conditionally appear.

I don't believe existing callers have explicit bindings that will fail on a new attribute, but I'm not confident. This is probably worth a quick follow up.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReport.groovy
```diff
@@ -61,22 +61,50 @@ class AmazonReservationReport implements ReservationReport {
...
 
     @JsonProperty
     String availabilityZoneId() {
-      return availabilityZone == null ? '*' : availabilityZone[-1..-1]
+      return (availabilityZone == null || availabilityZone == '*') ? '*' : availabilityZone[-1..-1]
+    }
+
+    @JsonProperty
+    String availabilityZone() {
+      return availabilityZone == null ? '*' : availabilityZone
```
ajordens: This was pointed out as missing for regional reservations.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReport.groovy
```diff
@@ -112,4 +139,69 @@ class AmazonReservationReport implements ReservationReport {
       return (reservedVpc.intValue() - usedVpc.intValue())
     }
   }
+
+  static class Borrow {
```
ajordens: Not sold on the name `Borrow` (I've been back and forth between Borrow, Allocation, RegionalAllocation, etc.). Happy to hear feedback either way.

The intention is to share details around how instance types are borrowed/allocated from regional reservations to cover shortfalls.

```
"borrows": [
          {
            "index": 1,
            "source": "us-west-2:*:m4.xlarge:linux",
            "sourceInstanceQuantity": 40,
            "sourceInstanceType": "m4.xlarge",
            "target": "us-west-2:c:m4.4xlarge:linux",
            "targetInstanceQuantity": 10,
            "targetInstanceType": "m4.4xlarge"
          }
        ]
```

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReport.groovy
```diff
@@ -112,4 +139,69 @@ class AmazonReservationReport implements ReservationReport {
...
+    String target
+    String targetInstanceType
+    int targetInstanceQuantity
+
+    int index
+  }
+
+  static class DescendingOverallReservationDetailComparator implements Comparator<OverallReservationDetail> {
+    @Override
+    int compare(OverallReservationDetail a, OverallReservationDetail b) {
```
ajordens: Expanded for readability more than anything ... made explicitly descending.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReportBuilder.groovy
```diff
@@ -0,0 +1,233 @@
...
+  /**
+   * This version (v3) of the reservation report:
+   *
+   * - normalizes all regional reservations to `xlarge` instance types
+   * - attempts to cover all shortfalls with regional reservations (converting `xlarge` to the necessary instance types)
+   */
+  class V3 implements AmazonReservationReportBuilder {
+    // relative multipliers for instance types < xlarge where the baseline is xlarge
+    // (would be nice if AWS had an API to derive this from)
+    Map<String, Double> instanceTypeMultipliers = [
```
ajordens: Pulled this from the AWS EC2 Instances site ... worth double checking for accuracy. 

I'm not aware of another source for this information (other than perhaps https://aws.amazon.com/blogs/aws/new-aws-price-list-api/).

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReportBuilder.groovy
```diff
@@ -0,0 +1,229 @@
...
+      )
+    }
+
+    private List<OverallReservationDetail> aggregateRegionalReservations(List<OverallReservationDetail> reservations) {
+      def regionalReservations = filterRegionalReservations(reservations)
+
+      regionalReservations.groupBy { "${it.region}-${it.instanceFamily()}-${it.os}" }.collect {
+        def overallRegionalReservationDetail = new OverallReservationDetail(
+          availabilityZone: "*",
+          instanceType: it.value[0].instanceFamily() + ".xlarge",
```
ajordens: I did not explicitly do `.Fxlarge` ... wasn't convinced it was necessary. 

I'm not aware of the entire backstory so happy to see this changed if there is a legitimate reason to do so (`.Fxlarge` seems equivalent to `availabilityZone == '*' && instanceType == 'xlarge'`).

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReportBuilder.groovy
```diff
@@ -0,0 +1,229 @@
...
+      }
+    }
+
+    /**
+     * @return the multiplier of a given instanceType related to an 'xlarge', or 0 if instanceType is not supported
+     */
+    private double getMultiplier(String instanceType) {
+      if (!instanceTypeMultipliers.containsKey(instanceType) && !instanceType.endsWith("xlarge")) {
+        // not an xlarge instance type (or no explicit multiplier if < xlarge)
+        return 0
```
ajordens: could be null? should be null? could be Optional?

ajordens:
FYI @robzienert

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReport.groovy
```diff
@@ -61,22 +61,49 @@ class AmazonReservationReport implements ReservationReport {
...
+    String id() {
+      return "${region}:${availabilityZoneId()}:${instanceType}:${os.name.toLowerCase()}"
+    }
+
+    @JsonProperty
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    Integer totalBorrowed() {
+      // returning null will cause this attribute to not be serialized
+      // (will only appear in v3 report for instance types that have related borrows
+      return totalBorrowed.intValue() ?: null
```
ajordens: This might be better off as a jackson `View` that excludes `totalBorrowed` from the v1/v2 reports.

I'll investigate that.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReportBuilder.groovy
robzienert: Missing t2.nano size.

robzienert: IIRC, we had originally talked about making sub-xlarge continually half-as-small, e.g. `medium: 0.25, small: 0.125, etc...` for ease. That said, I think using the pricing list API would be a good thing to explore after a first pass - I'm not excited about keeping a static size list.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReportBuilder.groovy
robzienert: I need to double back on why we chose `Flarge`. I seem to recall there being a good reason, but I can't remember it.

ajordens:
I'm not excited either.

I think we were using the pricing json back in the early days and got burnt
when the structure changed on us.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReportBuilder.groovy
robzienert: I think 0 is fine, but we should definitely be logging this scenario.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReportBuilder.groovy
```diff
@@ -0,0 +1,229 @@
...
+      )
+    }
+
+    private void coverShortfall(AtomicInteger borrowIndex,
+                                OverallReservationDetail regional,
+                                OverallReservationDetail shortfall) {
+      def multiplier = getMultiplier(shortfall.instanceType)
+      if (!multiplier) {
+        // instance type is unsupported (some unknown variant smaller than an xlarge)
+        return
```
robzienert: Again, I think we should log this.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReport.groovy
robzienert: Agree - allocation seems a better choice to me, since regional RIs are just a financial concept as opposed to true AZ capacity guarantees. The way I think of it, Spinnaker is allocating these reservations to the instance types, rather than an external system performing a borrow on an offering.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReport.groovy
robzienert: FWIW, I don't recall anyone raising an alarm when I added `region`.

ajordens:
Sounds good to me. I'll switch back to Allocation/allocations.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReportBuilder.groovy
ajordens: Sure ... I think I was basing the multipliers off of the # of vCPU but that may have been incorrect. 

Not sure how AWS does the crediting for the < xlarge types?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonReservationReportBuilder.groovy
ajordens: Logs added on the caller side of `getMultiplier()`

ajordens:
@robzienert - I made the switch to `fxlarge` for the aggregated regional reservations. I think that makes more sense than `rii` but it's an easy enough thing to change.

How's the rest of it look? Thoughts on getting it merged down and pushed?

As it currently stands (no `View` magic), it will add a new `id` and `totalAllocated` attribute to v1/v2 reports. If nobody complained when `region` was added, these two shouldn't cause any problems.

ajordens:
Added a standalone commit that introduces a few jackson `View`s. 

The new `id` and `totalAllocated` attributes will only be serialized in V3. 

This is probably unnecessary for current consumers but felt it might be a worthwhile attempt at derisking future changes to these reports.

robzienert:
👍 I think this works, I'm fine with `fxlarge` over `rri`. Probably best to get it in, then we can iterate further if necessary - but it looks accurate.

PR #1631, opened 2017-05-15
refactor(kubernetes): Update fabric8 client to 2.3.1 for StatefulSet

dsimonto:
StatefulSet model is not available in the fabric8 client currently used.  This upgrades fabric8 to the latest client.  Also had to import the Job model class from a different location.

### DIFF
--- a/clouddriver-kubernetes/clouddriver-kubernetes.gradle
+++ b/clouddriver-kubernetes/clouddriver-kubernetes.gradle
@@ -6,5 +6,5 @@ dependencies {
   compile spinnaker.dependency('bootWeb')
 
   // TODO (move to https://github.com/spinnaker/spinnaker-dependencies/blob/master/src/spinnaker-dependencies.yml)
-  compile 'io.fabric8:kubernetes-client:1.4.35'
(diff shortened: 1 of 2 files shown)


dsimonto:
@lwander please review.  This sets the stage for adding StatefulSet to the mix.

lwander:
Have you run through the various operations locally, and if so, against what k8s cluster versions?

dsimonto:
I ran the clouddriver unit tests which all passed.  I also ran these ops against Minikube 0.14.0 (kubernetes 1.5)  from Spinnaker: create/delete load balancer, create/destroy server group, resize server group, terminate instance of server group.  I did not try anything with Jobs (I don't have a Job container).

Does the integration test verify these operations?

lwander:
Awesome, we'll run the integration test tonight after it's merged.

lwander:
It verifies them, but it's part of a larger suite that takes a while to run, and we're trying to keep those clean/passing for the next two weeks while we practice our release process. A sanity check before merging goes a long way while we try to get this process in place.

PR #1634, opened 2017-05-17
feat(dcos): Add DC/OS provider

willgorman:
First of all, apologies for the size of this PR.  Regarding the commit history, I'm happy to do whatever would be preferred if you don't want to keep it as is.  We can either break it down into a few commits for groups of changes or just a single commit.  For context, here are [some notes on the relationship between DC/OS and Spinnaker concepts](https://github.com/cerner/spinnaker/blob/integration/DCOS%20and%20Clouddriver%20Concepts.md)


duftler:
Glad to see support for DC/OS added to Spinnaker.

The shortest path to getting this reviewed and merged is to break the PR into the smallest meaningful pieces.

This typically (for a new provider integration) looks like:
Clouddriver:
- PR containing support for enabling a new provider, with zero operations supported; just support populating the credentials
- Series of PR's containing support for 1 or 2 operations end-to-end (the converter, description, validator, unit tests)

Orca:
- PR containing small bits of provider-specific scaffolding like ServerGroupCreator

Deck:
- Crazy PR's in whatever framework/language we are using at the instant you submit the PR's (if you don't hit the 'Create Pull Request' button really quickly you'll have to try to lead the target a bit)

The Oracle guys are very far along in getting their support merged in, so that's a good path to follow. I'll grab some links in a moment.

duftler:
This query yields a solid, recent path to follow in terms of decomposing this PR: https://github.com/spinnaker/clouddriver/pulls?utf8=%E2%9C%93&q=is%3Apr%20is%3Aclosed%20author%3Asimonlord

duftler:
Note the 'Noop' operation in this PR: https://github.com/spinnaker/clouddriver/pull/1532

You should not need to do that anymore when enabling your new provider with zero operations defined.

willgorman:
Sounds good, thanks.  We'll work on getting this split up.

willgorman:
Going to cancel this PR since it looks like I can't change the source branch from our fork.

PR #1635, opened 2017-05-17
refactor(aws): s/totalAllocated/totalRegionalReserved and s/allocatio…

ajordens:
…ns/regionalReservedAllocations


ajordens:
PTAL @robzienert ... naming is hard.

PR #1638, opened 2017-05-19
feat(provider/aws): Add caching support for ALBs, Target Groups, Listeners, TargetHealth

jrsquared:
Add caching support for ALBs, Target Groups, Listeners, TargetHealth so we can view albs and target groups.

Depends on https://github.com/spinnaker/deck/pull/3757


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/data/Keys.groovy
```diff
@@ -101,16 +107,32 @@ class Keys {
...
+          key += ":${loadBalancerArn}"
+        }
       }
     }
     return key
   }
 
+  static String getTargetGroupKey(String targetGroupName, String account, String region, String vpcId) {
+    String key = "${ID}:${Namespace.TARGET_GROUPS}:${account}:${region}:${targetGroupName}"
+    if (vpcId) {
```
cfieber: vpcId should always be present on a target group

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonLoadBalancer.groovy
```diff
@@ -57,6 +59,6 @@ class AmazonLoadBalancer implements LoadBalancer {
 
   @Override
   int hashCode() {
-    getAccount().hashCode() + getName().hashCode() + getType().hashCode() + getServerGroups().hashCode() + getRegion().hashCode()
+    getAccount().hashCode() + getName().hashCode() + getType().hashCode() + getServerGroups().hashCode() + getRegion().hashCode() + getTargetGroups().hashCode()
```
cfieber: not sure who did the initial hashCode on here but it is kind of bad - probably worth letting the IDE auto-generate one

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonClassicLoadBalancerCachingAgent.groovy
```diff
@@ -19,116 +19,32 @@ package com.netflix.spinnaker.clouddriver.aws.provider.agent
...
-      this(id);
-      this.attributes.putAll(attributes);
-      this.relationships.putAll(relationships);
-    }
-  }
-
-  @Override
-  boolean handles(OnDemandAgent.OnDemandType type, String cloudProvider) {
-    type == OnDemandAgent.OnDemandType.LoadBalancer && cloudProvider == amazonCloudProvider.id
+class AmazonClassicLoadBalancerCachingAgent extends AbstractAmazonLoadBalancerCachingAgent {
```
cfieber: FWIW this change will mean we need to do a redis replacement roll in since the class name is previously in the key

cfieber: I actually think it would be worth not changing the classname - (we've already got some other less that spectacularly named caching agents in clouddriver-aws) - just to keep the roll in/roll back story simpler

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/data/Keys.groovy
```diff
@@ -101,16 +107,32 @@ class Keys {
...
+    return getLoadBalancerKey(loadBalancerName, account, region, vpcId, loadBalancerType, null)
+  }
+
+  static String getLoadBalancerKey(String loadBalancerName, String account, String region, String vpcId, String loadBalancerType, String loadBalancerArn) {
     String key = "${ID}:${Namespace.LOAD_BALANCERS}:${account}:${region}:${loadBalancerName}"
     if (vpcId) {
       key += ":${vpcId}"
       if (loadBalancerType && loadBalancerType != 'classic') {
         key += ":${loadBalancerType}"
+        if (loadBalancerArn != null) {
```
cfieber: is it legitimate to have a non-classic ELB without an ARN?

is there something that needs the ARN to be adde to the key? (can't we infer it by load balancer name)

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonLoadBalancer.groovy
jrsquared: Ha, removed.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonClassicLoadBalancerCachingAgent.groovy
jrsquared: Haha, alright, makes sense.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/data/Keys.groovy
jrsquared: 👍

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/data/Keys.groovy
jrsquared: No. This is more a guard against someone doing something stupid later on trying to create a key without the arn. I'll remove it though. Better to blow up in testing I guess.

jrsquared: There was something, but looks like I refactored whatever it was? Sure, that's it. I'll remove it.

PR #1643, opened 2017-05-22
feat(aws): Optionally delete ASG with autoscaling instance terminations

robzienert:
In some orchestrations, we need to be able to delete instances that will
allow the autoscaling lifecycle hooks to be triggered.

By default, termination logic will be unchanged from its current behavior.


anotherchrisberry:
This has been open a month - any concerns or should we rebase/merge it?

robzienert:
I just started reconsidering how I'm doing all of this. I think the real
fix for canary terminations will actually need to be more orchestrated.
Would like to leave open a bit longer while I deliberate.

PR #1649, opened 2017-05-26
fix(provider/amazon): Fix isInternal flag on load balancer description

jrsquared:
Fixes https://github.com/spinnaker/spinnaker/issues/1692


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/UpsertAmazonLoadBalancerDescription.java
```diff
@@ -65,11 +65,11 @@ public void setVpcId(String vpcId) {
     this.vpcId = vpcId;
   }
 
-  public Boolean getInternal() {
+  public Boolean isInternal() {
```
cfieber: this maybe should really be `getIsInternal` or `isIsInternal` ... since the actual attribute in the JSON payload is `"isInternal": true|false`

jrsquared: Fixed.

PR #1653, opened 2017-05-30
fix(provider/amazon):  add gov-cloud special case to arn patterns

rogerswingle:
GovCloud ARNs don't fit the standard model, much in the same way that the china region does.

This PR adds support for GovCloud Style Arns to fix spinnaker/spinnaker/issues/1698


rogerswingle:
@spinnaker/reviewers, is there anything else I need to do or provide here?

jrsquared:
LGTM

st33v:
hey... does anyone have spinnaker deploying to govcloud? seems like it might be hitting 'non-gov' for the token service.

PR #1655, opened 2017-05-31
fix(provider/amazon): Fix upsert ALB when target group does not exist

jrsquared:


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -84,7 +84,14 @@ class LoadBalancerV2UpsertHandler {
     task.updateStatus BASE_PHASE, "Security groups updated on ${loadBalancerName}."
 
     // Get existing target groups so we can reconcile
-    def existingTargetGroups = loadBalancing.describeTargetGroups(new DescribeTargetGroupsRequest().withNames(targetGroups.collect { it.name }))?.targetGroups
+    def existingTargetGroups
+    try {
+      existingTargetGroups = loadBalancing.describeTargetGroups(new DescribeTargetGroupsRequest().withNames(targetGroups.collect { it.name }))?.targetGroups
```
cfieber: worth validating here that targetGroups is non-empty? this might fall into the same situation where it returns the entire list if that was the case

jrsquared: Ah, good call. Better safe than 🇨🇦 .

PR #1659, opened 2017-06-01
fix(aws: enable/disable) faster instance lookup for enable/disable

cfieber:
Reworks instance lookup logic in #1255 to prefer a query against known instance ids instead of a query with a tag filter.

The current implementation can time out with large instance counts in a region

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AbstractEnableDisableAtomicOperation.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AbstractEnableDisableAtomicOperation.groovy
@@ -108,25 +108,29 @@ abstract class AbstractEnableDisableAtomicOperation implements AtomicOperation<V
         asgService.resumeProcesses(serverGroupName, AutoScalingProcessType.getDisableProcesses())
       }
 
-      List<String> instanceIds = new ArrayList<>()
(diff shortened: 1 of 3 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AbstractEnableDisableAtomicOperation.groovy
```diff
@@ -108,26 +108,30 @@ abstract class AbstractEnableDisableAtomicOperation implements AtomicOperation<V
...
-        describeInstancesRequest.setNextToken(describeInstancesResult.getNextToken())
-        describeInstancesResult = regionScopedProvider.amazonEC2.describeInstances(describeInstancesRequest)
-        reservations += describeInstancesResult.getReservations()
-      }
-
-      for (Reservation reservation : reservations) {
-        instanceIds += reservation.getInstances().findAll {
-          [ InstanceStateName.Running, InstanceStateName.Pending ].contains(InstanceStateName.fromValue(it.getState().getName()))
-        }*.instanceId
+      List<String> instanceIds = asg.instances.findAll {
```
jeyrschabu: Do we need to filter the list of instances since we are doing it after the describe-instances?

cfieber: I figured if we know an instance will not be in a good state may as well not query for it

PR #1667, opened 2017-06-06
feat(request buffer): optional buffer to distribute api requests

cfieber:
When enabled submitted API requests are first queued by application. Requests are popped
one at a time from each application queue and submitted for execution. This should mitigate
one particular application user hammering the API and degrading service for others.

In addition the requests are executed on a fixed size thread pool so no matter how many requests
are submitted only a bounded number will be executed at any time.

The request thread is blocked until the request is serviced from the threadpool or times out.


robzienert:
👍

PR #1668, opened 2017-06-06
fix(provider/amazon): Disable alb caching via edda by default

jrsquared:


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/EddaTimeoutConfig.java
```diff
@@ -36,14 +36,16 @@
   private final int connectTimeout;
   private final int socketTimeout;
   private final Set<String> disabledRegions;
+  private Boolean albEnabled;
```
cfieber: I'd make this little-b boolean - should never be null at this point

PR #1672, opened 2017-06-07
fix(provider/google): Don't NPE when machine types aren't returned.

duftler:


duftler:
@danielpeach Tested the ui against this and it correctly prevents us from choosing a static instance type. But I can still configure a custom type, which I think is wrong (my thinking is if no static instance types exist in a zone, we shouldn't allow provisioning of instances at all). WDYT?

PR #1675, opened 2017-06-07
feat(provider/amazon): Support caching of Rules and support CRUD of rules

jrsquared:


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/AmazonClientInvocationHandler.java
```diff
@@ -104,7 +105,11 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
       return thisMethod.invoke(this, args);
     } catch (NoSuchMethodException e) {
       wasDelegated = true;
-      return method.invoke(delegate, args);
+      try {
```
jrsquared: Needed this to rethrow the error so the caching agent can handle the originating error

cfieber: as we discussed I think this was not our typical code path before now but this makes sense to me

PR #1677, opened 2017-06-08
fix(provider/google): Don't NPE when machine types aren't returned. (…

duftler:
…#1672)

Cherry-pick GCE NPE fix (https://github.com/spinnaker/clouddriver/pull/1672) into 1.0.x.


jtk54:
LGTM.

PR #1683, opened 2017-06-11
feat(provider/kubernetes): Run Job Pod Labels

ethanfrogers:
Provide a way to define a set of labels on a Run Job pod. Useful for
being able to pin point specific pods for logging purposes. Adds a Map
for labels in the Run Job description and applies them if they're set.
Does not impact current without a `labels` field.

Part of [#1683](https://github.com/spinnaker/spinnaker/issues/1683). Deck PR coming soon.

@lwander PTAL


ethanfrogers:
Rebased and ready to go!

PR #1688, opened 2017-06-13
fix(requestQueue): differentiate timeout in queue vs in processing.

cfieber:
Allows a separate (shorter) timeout for how long a message can queue up before being rejected.

Requests that time out from the request queue fail with HTTP 429

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/requestqueue/QueuedRequestException.java
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/requestqueue/QueuedRequestException.java
@@ -14,12 +14,14 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.clouddriver.requestqueue.pooled;
+package com.netflix.spinnaker.clouddriver.requestqueue;
 
-import com.netflix.spinnaker.clouddriver.requestqueue.QueuedRequestException;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
 
-class PromiseTimeoutException extends QueuedRequestException {
-  PromiseTimeoutException() {
+@ResponseStatus(value = HttpStatus.TOO_MANY_REQUESTS)
+public class QueuedRequestException extends RuntimeException {
+  public QueuedRequestException() {
     super();
   }
 }
--- a/clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/requestqueue/pooled/PooledRequestQueueSpec.groovy
+++ b/clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/requestqueue/pooled/PooledRequestQueueSpec.groovy
@@ -19,12 +19,14 @@ package com.netflix.spinnaker.clouddriver.requestqueue.pooled
 import com.netflix.spectator.api.NoopRegistry
 import spock.lang.Specification
 
-import java.util.concurrent.TimeoutException
+import java.util.concurrent.Callable
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.atomic.AtomicBoolean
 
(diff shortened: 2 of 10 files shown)


clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/requestqueue/pooled/PooledRequestQueueSpec.groovy
```diff
@@ -35,12 +37,51 @@ class PooledRequestQueueSpec extends Specification {
...
     when:
-    Long result = queue.execute("foo", { Thread.sleep(20); return 12345L })
+    queue.execute("foo", { Thread.sleep(20); return 12345L })
 
     then:
     thrown(PromiseTimeoutException)
   }
+
+
+  def "should time out if request does not start in time"() {
```
cfieber: not super proud of this one

jrsquared: not super proud of the extra whitespace to pad your LOC?

robfletcher: How long will that take to run? Seems like the executor could evict that 40 second sleep thread in cleanup so it's not hanging around.

clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/requestqueue/pooled/PooledRequestQueueSpec.groovy
```diff
@@ -35,12 +37,51 @@ class PooledRequestQueueSpec extends Specification {
...
+    given: "a queue with one worker thread"
+    def queue = new PooledRequestQueue(new NoopRegistry(), 10, 10, 1)
+    AtomicBoolean itRan = new AtomicBoolean(false)
+    def didItRun = new Callable() {
```
robfletcher:
```def didItRun = {->
  itRan.set(true)
}```
Should work, no? Closure implements Callable

File: clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/requestqueue/pooled/PooledRequestQueueSpec.groovy (hunk 1 above)
Comment (cfieber) [inline clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/requestqueue/pooled/PooledRequestQueueSpec.groovy]: its a 40 millisecond sleep thread that the test ensures has actually started before exercising the specific test case so I don't think there is a scenario where it ends up hanging around ever

File: clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/requestqueue/pooled/PooledRequestQueueSpec.groovy (hunk 2 above)
Comment (cfieber) [inline clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/requestqueue/pooled/PooledRequestQueueSpec.groovy]: fixed

**Comment by cfieber** [comment]
@robfletcher updated - PTALA

File: clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/requestqueue/QueuedRequestException.java [hunk 3]
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.requestqueue;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(value = HttpStatus.TOO_MANY_REQUESTS)
+public class QueuedRequestException extends RuntimeException {
```
cfieber: needed to move this to a RuntimeException instead of a subclass of TimeoutException which is checked in order to get the spring mvc exception handling stuff to work

PR #1692, opened 2017-06-15
fix(provider/kubernetes): Update HPA endpoint for k8s 1.6

AMeng:
Fixes https://github.com/spinnaker/spinnaker/issues/1651

This updates kubernentes-client with this fix: https://github.com/fabric8io/kubernetes-client/pull/780

Changes between the two APIs:
> * Field CPUUtilization which was a nested structure CPUTargetUtilization in HorizontalPodAutoscalerSpec was replaced by TargetCPUUtilizationPercentage which is an integer.
> * ScaleRef of type SubresourceReference in HorizontalPodAutoscalerSpec which referred to scale subresource of the resource being scaled was replaced by ScaleTargetRef which points just to the resource being scaled.

This is necessary for supporting Kubernetes 1.6, but breaks support for Kubernetes 1.1. I'm not sure what the officially supported versions are.

### DIFF
--- a/clouddriver-kubernetes/clouddriver-kubernetes.gradle
+++ b/clouddriver-kubernetes/clouddriver-kubernetes.gradle
@@ -6,5 +6,5 @@ dependencies {
   compile spinnaker.dependency('bootWeb')
 
   // TODO (move to https://github.com/spinnaker/spinnaker-dependencies/blob/master/src/spinnaker-dependencies.yml)
-  compile 'io.fabric8:kubernetes-client:2.3.1'
+  compile 'io.fabric8:kubernetes-client:2.4.1'
 }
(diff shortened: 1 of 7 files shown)


ethanfrogers:
Does this also support clusters that still use the old API? Not sure what K8S release changed the API but I've seen reports that it may be pre 1.5.5.

AMeng:
> Does this also support clusters that still use the old API?

No. However, the new endpoint was actually added in K8S 1.2. The old endpoint was finally removed in 1.6.

See: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md#other-notable-improvements

AMeng:
@spinnaker/reviewers

ethanfrogers:
@lwander are you waiting on the release of that other client library to tackle the HPA issue?

lwander:
Oh this is great - sorry I missed this. Have you checked to see if this works with kubernetes versions prior to 1.6?

AMeng:
> Have you checked to see if this works with kubernetes versions prior to 1.6?

I have not, but according to the docs, this endpoint was added in 1.2.

lwander:
Ok - I'll pull it down and see how it does against prior cluster versions

lwander:
Looks good!

ethanfrogers:
image

PR #1701, opened 2017-06-20
fix(kubernetes): allow configMaps to have no items

dblackdblack:
Having a configMap volume source with zero key/path pairs is allowed by kubernetes, but is rejected by the clouddriver pipeline validator. Having no items in a configMap already works fine if you manually create a server group, presumably b/c server group validation is implemented separately. However if you take this manually created server group and import it as your server group template in a pipeline, the resulting pipeline does not execute b/c of the strict validation that is loosened by this PR.

Addresses https://github.com/spinnaker/spinnaker/issues/1622


dblackdblack:
@lwander do you think this needs tests?  It is simply removing an overzealous validation step.  I tested this change by hand on our staging spinnaker server.

lwander:
No test needed IMO - thanks for the fix!

PR #1710, opened 2017-06-23
fix(provider/google): Filter LBs by name in Google LB provider.

jtk54:
We don't actually filter to get the `View` by name, so in the case where two forwarding rules have similar names, this may find the wrong one since `getApplicationLoadBalancers` searches for `name*`.


jtk54:
Closing because the gnarly changes in https://github.com/spinnaker/deck/pull/3859 aren't necessary.

PR #1711, opened 2017-06-23
feat(provider/kubernetes): Configure Service Acct

ethanfrogers:
Allows Service Account to be configured for both Server Group and Run
Job Pipeline stage. I also added it to the `fromReplicationController`
in the event that someone is using ReplicationControllers pre-Spinnaker.

@danielpeach PTAL
@lwander FYI


ethanfrogers:
@danielpeach from my testing, `CloneKubernetesAtomicOperation` does not need to be updated. I cloned a Server Group and it picked up the correct `serviceAccountName`.

As discussed on Slack, I think we can go ahead and merge this and come back to populating the fields/verifying the service account exists. That would also be great to do for ConfigMaps/Secrets as well so knocking it out at one time would be better I think.

PR #1714, opened 2017-06-26
feat(dynomite): Add initial Dynomite support

robzienert:
Adds support for Dynomite support as an alternative backing store to Redis. Since this PR is pretty large, I wanted to put it out there now that things are working locally. The major blocking things for merge:

Major things of note:

1. Redis impl now uses string protocol over binary commands for hashing operations. This may cause a temporary performance regressions on initial caching agents in deployments that had hashing enabled, but would auto-resolve as hash tables are updated with string values.
2. Dyno team is working on Lua evaluation support, which would allow us to use `ClusteredSortAgentScheduler`, until then, I've implemented a janky fork of `ClusteredAgentScheduler` that is non-atomic, which should suffice for initial pre-production testing.

@spinnaker/reviewers PTAL


cats/cats-dynomite/src/main/java/com/netflix/spinnaker/cats/dynomite/cluster/DynoClusteredAgentScheduler.java
```diff
@@ -0,0 +1,216 @@
...
+        AgentIntervalProvider.Interval interval = intervalProvider.getInterval(a.getValue().getAgent());
+        if (acquireRunKey(agentType, interval.getTimeout())) {
+          acquired.put(agentType, System.currentTimeMillis() + interval.getInterval());
+        }
+      });
+    return acquired;
+  }
+
+  private boolean acquireRunKey(String agentType, long timeout) {
+    // This isn't safe, but it's the best one can do without `String set(String key, String value, String nxxx, String expx, long time).`
```
robzienert: https://github.com/Netflix/dyno/pull/177 😍

cfieber: looks like `setnx` is currently supported so you could switch to that in the interim

cats/cats-dynomite/src/main/java/com/netflix/spinnaker/cats/dynomite/cache/DynomiteCache.java
```diff
@@ -0,0 +1,180 @@
...
+      redisClientDelegate.withCommandsClient(client -> {
+        for (List<String> idPart : Iterables.partition(idSet, options.getMaxSaddSize())) {
+          final String[] ids = idPart.toArray(new String[idPart.size()]);
+          client.sadd(allOfTypeReindex(type), ids);
+          saddOperations.incrementAndGet();
+          client.sadd(allOfTypeId(type), ids);
+          saddOperations.incrementAndGet();
+        }
+
+        for (List<String> keys : Lists.partition(keysToSet, options.getMaxMsetSize())) {
```
cfieber: since these are just doing `.set` there isn't really any reason to partition

cfieber:
There is also the `JedisTaskRepository` that needs updating for dyno friendlyness. It has mild use of pipelining that could just be removed altogether and I suspect otherwise would pretty much just work aside from updating it to use the delegate instead of the JedisPool

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/JedisClientDelegate.java
```diff
@@ -0,0 +1,110 @@
...
+import redis.clients.jedis.commands.RedisPipeline;
+
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+public class JedisClientDelegate implements RedisClientDelegate {
+
+  JedisSource jedisSource;
+
+  public JedisClientDelegate(JedisSource jedisSource) {
```
cfieber: if this is the only place still using JedisSource you could probably just switch this to Pool<Jedis> here

robzienert:
This could use another round of eyeballs.

robzienert:
Closing, will target feature branch.

PR #1717, opened 2017-06-27
feat(provider/dcos): Add destroy, disable, and resize server group

willgorman:
Adds the destroy, disable, and resize server group operations for DC/OS.


willgorman:
@duftler @robzienert Would it be possible to have someone take a look at this?

robzienert:
👍

PR #1718, opened 2017-06-28
fix(aws/serverGroup): add build info

icfantv:
* the REST call for retrieving individual server groups was not adding
the `buildInfo` property to the `ServerGroup` object returned (but it
was still present in the `tags` array).  like the REST call for
retrieving all the server groups for an applications, this PR adds the
build information to the individual server group.


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonClusterProvider.groovy
```diff
@@ -228,32 +229,11 @@ class AmazonClusterProvider implements ClusterProvider<AmazonCluster> {
...
-          }
-
-        }
-      }
 
       serverGroupIds.each { serverGroupId ->
         def serverGroup = serverGroups[serverGroupId]
         serverGroup.image = image.attributes
-        serverGroup.buildInfo = buildInfo
+        serverGroup.buildInfo = getBuildInfoFromImage(image)
```
icfantv: this retains the original behavior of setting the `buildInfo` to `null` if any of the conditions in the original code are met such that the value to be set is `null`.

PR #1723, opened 2017-07-05
feat(provider/google): Add support for useSourceCapacity on Deploy.

duftler:


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -362,6 +362,20 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
       }
     }
 
+    if (description.source?.useSourceCapacity && description?.source?.region && description.source?.serverGroupName) {
```
danielpeach: I would make the null checks on these consistent (you have one `description?`).

duftler: Fixed.

PR #1729, opened 2017-07-09
feat(provider/kubernetes): Support EBS Volumes

ethanfrogers:
Adds support for AWS Elastic Block Store volumes. Can serve as a good
example for further supporting other volume source types.

Note: EBS Volumes cannot be mounted ReadOnly even though the client library says it can be. When I attempted to add support for ReadOnly, Kubernetes failed to mount the volume with an error saying as much. Thus, it was left out.

Partially addresses spinnaker/spinnaker#1826

@lwander PTAL

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
@@ -154,6 +154,17 @@ class KubernetesApiConverter {
         volume.configMap = res.build()
         break
 
+      case KubernetesVolumeSourceType.AwsElasticBlockStore:
+        def res = new AWSElasticBlockStoreVolumeSourceBuilder().withVolumeID(volumeSource.awsElasticBlockStore.volumeId)
+        res = res.withFsType(volumeSource.awsElasticBlockStore.fsType)
+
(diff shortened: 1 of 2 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/description/servergroup/DeployKubernetesAtomicOperationDescription.groovy
```diff
@@ -269,6 +273,15 @@ class KubernetesConfigMapVolumeSource {
 
 @AutoClone
 @Canonical
+class KubernetesAwsElasticBlockStoreVolumeSource {
+  String volumeId
+  String fsType
+  Integer partition
+
```
lwander: Extra newline

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/api/KubernetesApiConverter.groovy
```diff
@@ -153,6 +153,16 @@ class KubernetesApiConverter {
         res = res.withItems(items)
         volume.configMap = res.build()
         break
+      case KubernetesVolumeSourceType.AwsElasticBlockStore:
```
lwander: Missing newline after above `break`.

ethanfrogers:
@lwander nits un-nitted.

PR #1731, opened 2017-07-11
feat(provider/google): Adding functionality for selecting certain regions for indexing.

dbakshiyev:
Specifying which regions you want indexed reduces the number of calls the caching agents have to make to the API.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/config/GoogleConfigurationProperties.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/config/GoogleConfigurationProperties.groovy
@@ -30,6 +30,9 @@ class GoogleConfigurationProperties {
     List<String> imageProjects
     ConsulConfig consul
     String userDataFile
+    // Takes a list of regions you want indexed, per-account. Will default to the value in
+    // defaultRegions if left unspecified. An empty list will index no regions.
+    List<String> regions
   }
 
   List<ManagedAccount> accounts = []
@@ -37,4 +40,7 @@ class GoogleConfigurationProperties {
   int asyncOperationMaxPollingIntervalSeconds = ASYNC_OPERATION_MAX_POLLING_INTERVAL_SECONDS
   List<String> baseImageProjects
   long maxMIGPageSize = 50
+  // Takes a list of regions you want indexed. Will default to indexing all regions if left
+  // unspecified. An empty list will index no regions.
+  List<String> defaultRegions
 }
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleBackendServiceCachingAgent.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleBackendServiceCachingAgent.groovy
@@ -67,14 +67,12 @@ class GoogleBackendServiceCachingAgent extends AbstractGoogleCachingAgent {
     if (globalBackendServices) {
       ret.addAll(globalBackendServices.collect { toGoogleBackendService(it, GoogleBackendService.BackendServiceKind.globalBackendService) })
     }
-    timeExecute(compute.regions().list(project),
-                "compute.regions.list",
-                TAG_SCOPE, SCOPE_GLOBAL
-    ).items.each { Region region ->
+
+    credentials.regions.collect { it.name }.each { String region ->
       def regionBackendServices = timeExecute(
-          compute.regionBackendServices().list(project, region.getName()),
+          compute.regionBackendServices().list(project, region),
           "compute.regionBackendServices.list",
-          TAG_SCOPE, SCOPE_REGIONAL, TAG_REGION, region.getName()
+          TAG_SCOPE, SCOPE_REGIONAL, TAG_REGION, region
       )?.items as List
       if (regionBackendServices) {
         ret.addAll(regionBackendServices.collect { toGoogleBackendService(it, GoogleBackendService.BackendServiceKind.regionBackendService) })
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.groovy
@@ -64,6 +64,7 @@ class GoogleNamedAccountCredentials implements AccountCredentials<GoogleCredenti
   final ConsulConfig consulConfig
   final Compute compute
   final String userDataFile
+  final List<String> regionsToManage
 
   static class Builder {
     String name
@@ -85,6 +86,7 @@ class GoogleNamedAccountCredentials implements AccountCredentials<GoogleCredenti
     Compute compute
     ConsulConfig consulConfig
     String userDataFile
+    List<String> regionsToManage
 
     /**
      * If true, overwrites any value in regionToZoneMap, locationToInstanceTypesMap and locationToCpuPlatformsMap with values from the platform.
@@ -184,6 +186,11 @@ class GoogleNamedAccountCredentials implements AccountCredentials<GoogleCredenti
       return this
     }
 
+    Builder regionsToManage(List<String> regionsToManage, List<String> defaultRegions) {
+      this.regionsToManage = (regionsToManage != null) ? regionsToManage : defaultRegions
+      return this
+    }
+
     @VisibleForTesting
     Builder credentials(GoogleCredentials credentials) {
       this.credentials = credentials
@@ -236,19 +243,25 @@ class GoogleNamedAccountCredentials implements AccountCredentials<GoogleCredenti
                                         instanceTypeDisks,
                                         consulConfig,
                                         compute,
-                                        userDataFile)
+                                        userDataFile,
+                                        regionsToManage)
 
     }
   }
 
   public List<Map> getRegions() {
     List<Map> regionList = []
-
-    regionToZonesMap.each { String region, List<String> zones ->
-      Map regionMap = new HashMap()
-      regionMap.put("name", region)
-      regionMap.put("zones", zones)
-      regionList.add(regionMap)
+    if (regionsToManage != null) {
+      if (!regionsToManage.isEmpty()) {
+        regionToZonesMap.findAll { regionsToManage.contains(it.getKey()) }.each {
+          String region, List<String> zones ->
+            regionList.add([name: region, zones: zones])
+        }
+      }
+    } else {
+      regionToZonesMap.each { String region, List<String> zones ->
+        regionList.add([name: region, zones: zones])
+      }
     }
     return regionList
   }
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleCredentialsInitializer.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleCredentialsInitializer.groovy
@@ -81,6 +81,7 @@ class GoogleCredentialsInitializer implements CredentialsInitializerSynchronizab
             .consulConfig(managedAccount.consul)
             .instanceTypeDisks(googleDeployDefaults.instanceTypeDisks)
             .userDataFile(managedAccount.userDataFile)
+            .regionsToManage(managedAccount.regions, googleConfigurationProperties.defaultRegions)
             .build()
 
         if (!managedAccount.project) {


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleBackendServiceCachingAgent.groovy
```diff
@@ -67,10 +67,12 @@ class GoogleBackendServiceCachingAgent extends AbstractGoogleCachingAgent {
     if (globalBackendServices) {
       ret.addAll(globalBackendServices.collect { toGoogleBackendService(it, GoogleBackendService.BackendServiceKind.globalBackendService) })
     }
+    def regions = credentials.regions.collect { it.name }
+
     timeExecute(compute.regions().list(project),
                 "compute.regions.list",
                 TAG_SCOPE, SCOPE_GLOBAL
-    ).items.each { Region region ->
+    ).items.findAll { regions.contains(it.getName()) }.each { Region region ->
```
danielpeach: Is there a reason why we have to make a live call to get all regions here (rather than just iterating over our list of regions)? If there is, please drop an explanatory comment.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.groovy
```diff
@@ -236,19 +243,27 @@ class GoogleNamedAccountCredentials implements AccountCredentials<GoogleCredenti
...
+      regionsToIndex.each { String region ->
+        Map regionMap = new HashMap()
+        regionMap.put("name", region)
+        regionList.add(regionMap)
+      }
+    } else {
+      regionToZonesMap.each { String region, List<String> zones ->
+        Map regionMap = new HashMap()
+        regionMap.put("name", region)
+        regionMap.put("zones", zones)
```
danielpeach: The `regionMap` in the `if` block is missing this key (I think it's only used by the UI, which is probably why you didn't see any issues).

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/config/GoogleConfigurationProperties.groovy
```diff
@@ -30,6 +30,7 @@ class GoogleConfigurationProperties {
     List<String> imageProjects
     ConsulConfig consul
     String userDataFile
+    List<String> regionsToIndex
```
danielpeach: Two things here:

1). I think the name of the property here should match what's used for AWS (just `region`). I think it's OK to reference it as `regionsToIndex` on the `credentials` object. 

2). Like AWS, there should be a way to have a set of default regions for all GCE accounts, and then be able to override that set of defaults per account. You can do this piece in a later PR.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleBackendServiceCachingAgent.groovy
jtk54: +1, seems like we can get rid of the live call as long as we compute the regions we care about up front in the credentials.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/config/GoogleConfigurationProperties.groovy
jtk54: Default should probably be all GCE regions (since that's the behavior now).

danielpeach:
A few other thoughts:

I think the UI is going to handle this change properly, but that's not certain. You should verify that the server group wizard and the network and internal load balancer wizards show only the restricted set of regions, and that they appropriately switch between different sets of regions depending on the selected account.

We should also prevent (or at least think about preventing / figure out what AWS is doing) users of the API from deploying a resource into a region with an account that isn't indexing that region.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.groovy
dbakshiyev: Fixed

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleBackendServiceCachingAgent.groovy
dbakshiyev: Fixed

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/config/GoogleConfigurationProperties.groovy
dbakshiyev: Changed property name to `regions`

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleBackendServiceCachingAgent.groovy
```diff
@@ -67,14 +67,13 @@ class GoogleBackendServiceCachingAgent extends AbstractGoogleCachingAgent {
     if (globalBackendServices) {
       ret.addAll(globalBackendServices.collect { toGoogleBackendService(it, GoogleBackendService.BackendServiceKind.globalBackendService) })
     }
-    timeExecute(compute.regions().list(project),
-                "compute.regions.list",
-                TAG_SCOPE, SCOPE_GLOBAL
-    ).items.each { Region region ->
+
+    def regions = credentials.regions.collect { it.name }
```
jtk54: Doesn't `credentials` have `regionsToIndex`, not `regions`? Or am I mistaken?

dbakshiyev: explained offline (credentials.regions returns a list of region and zone maps)

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleBackendServiceCachingAgent.groovy
```diff
@@ -67,14 +67,13 @@ class GoogleBackendServiceCachingAgent extends AbstractGoogleCachingAgent {
...
+    regions.each { Region region ->
```
duftler: If you've just collected the names of the regions on the line above, this iterator shouldn't define the type as `Region`, right?

duftler: Might also want to simplify to something like:
```
credentials.regions.collect { it.name }.each { String region ->
```

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/config/GoogleConfigurationProperties.groovy
```diff
@@ -30,11 +30,13 @@ class GoogleConfigurationProperties {
...
     String userDataFile
+    List<String> regions
   }
 
   List<ManagedAccount> accounts = []
   int asyncOperationTimeoutSecondsDefault = ASYNC_OPERATION_TIMEOUT_SECONDS_DEFAULT
   int asyncOperationMaxPollingIntervalSeconds = ASYNC_OPERATION_MAX_POLLING_INTERVAL_SECONDS
   List<String> baseImageProjects
   long maxMIGPageSize = 50
+  List<String> defaultRegions
```
duftler: The semantics of this attribute warrant a detailed comment. What does it mean if it is left unspecified? If it is declared to be an empty list?

duftler: Same for the `regions` attribute up above.

Also, if the `regions` and `defaultRegions` attributes have implications only for indexing, the names should probably change. Is it 'managedRegions', `regionsToIndex`, or something else?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleBackendServiceCachingAgent.groovy
dbakshiyev: fixed

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/config/GoogleConfigurationProperties.groovy
dbakshiyev: Added comments and updated behavior so that an empty list will index no regions. Didn't update the names because we decided this will have some affect on mutating calls as well.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.groovy
```diff
@@ -184,6 +186,12 @@ class GoogleNamedAccountCredentials implements AccountCredentials<GoogleCredenti
       return this
     }
 
+    Builder regionsToIndex(List<String> regionsToIndex, List<String> defaultRegions) {
+
```
duftler: Extraneous blank line.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.groovy
```diff
@@ -236,19 +244,31 @@ class GoogleNamedAccountCredentials implements AccountCredentials<GoogleCredenti
...
-    regionToZonesMap.each { String region, List<String> zones ->
-      Map regionMap = new HashMap()
-      regionMap.put("name", region)
-      regionMap.put("zones", zones)
-      regionList.add(regionMap)
+    if (regionsToIndex != null) {
+      if (!regionsToIndex.isEmpty()) {
+        regionToZonesMap.findAll { regionsToIndex.contains(it.getKey()) }.each {
+          String region, List<String> zones ->
+            Map regionMap = new HashMap()
```
duftler: Would be more idiomatic to use:
```
regionList.add([name: region, zones: zones])
```

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.groovy
```diff
@@ -64,6 +64,7 @@ class GoogleNamedAccountCredentials implements AccountCredentials<GoogleCredenti
   final ConsulConfig consulConfig
   final Compute compute
   final String userDataFile
+  final List<String> regionsToIndex
```
duftler: Should this still be `regionsToIndex`, since you're going to constrain mutating calls as well?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.groovy
dbakshiyev: Fixed

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.groovy
dbakshiyev: Done

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.groovy
dbakshiyev: Renamed to regionsToManage, what do you think?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.groovy
```diff
@@ -184,6 +186,11 @@ class GoogleNamedAccountCredentials implements AccountCredentials<GoogleCredenti
       return this
     }
 
+    Builder regionsToManage(List<String> regionsToManage, List<String> defaultRegions) {
+      this.regionsToManage = regionsToManage ?: defaultRegions
```
danielpeach: If `regionsToManage` is an empty list (i.e., don't index anything for this account), won't this default to `defaultRegions`? I think you need to explicitly check that `regionsToManage` is `null`.

dbakshiyev: Fixed

PR #1736, opened 2017-07-12
chore(dependencies): Bump spinnaker-dependencies to 0.103.0

ajordens:
The generic exception handling from `kork-web` allows us to remove all
custom @ExceptionHandler's and simplify how error api responses are created.

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ClusterController.groovy
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ClusterController.groovy
@@ -24,6 +24,7 @@ import com.netflix.spinnaker.clouddriver.model.ServerGroup
 import com.netflix.spinnaker.clouddriver.model.Summary
 import com.netflix.spinnaker.clouddriver.model.TargetServerGroup
 import com.netflix.spinnaker.clouddriver.requestqueue.RequestQueue
+import com.netflix.spinnaker.kork.web.exceptions.NotFoundException
 import groovy.transform.Canonical
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.context.MessageSource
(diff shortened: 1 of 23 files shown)


ajordens:
Basically killed off a bunch of custom `@ExceptionHandler's` and used a few generic exceptions.

Plan on doing a bit of testing against the API before merging down ... specifically around the description validations where we're returning a set of error messages.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ClusterController.groovy
```diff
@@ -100,7 +101,7 @@ class ClusterController {
     }
     clusters.removeAll([null])
     if (!clusters) {
-      throw new ClusterNotFoundException(cluster: name)
+      throw new NotFoundException("Cluster not found (application: ${application}, account: ${account}, name: ${name})")
```
ajordens: Couldn't really think of a good reason to use explicit exceptions (vs the generic `NotFoundException`) inside of a controller.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ClusterController.groovy
```diff
@@ -238,78 +239,10 @@ class ClusterController {
...
-  @ResponseStatus(HttpStatus.NOT_FOUND)
-  Map handleClusterNotFoundException(AccountClustersNotFoundException ex) {
-    def message = messageSource.getMessage("account.clusters.not.found", [ex.application, ex.account] as String[], "account.clusters.not.found", LocaleContextHolder.locale)
-    [error: "account.clusters.not.found", message: message, status: HttpStatus.NOT_FOUND]
-  }
-
-  @ExceptionHandler
-  @ResponseStatus(HttpStatus.NOT_FOUND)
-  Map handleClusterNotFoundException(ClusterNotFoundException ex) {
-    def message = messageSource.getMessage("cluster.not.found", [ex.cluster] as String[], "cluster.not.found", LocaleContextHolder.locale)
```
ajordens: I can't see us ever internationalizing error messages ... use of `messageSource` seems overly complicated and unnecessary. Pretty sure we don't even get a good error message out of it right now.

Agree?

ajordens:
Ran some tests directly against the API ... believe that errors are reported in a compatible way with what was there previously.

PR #1740, opened 2017-07-13
fix(provider/google): Use port from frontend when upserting SSL Proxy

sayap:
This addresses the following issue found in review [1]:
> Changing the port on an SSL load balancer in the UI has no effect

[1] https://github.com/spinnaker/deck/pull/3894#pullrequestreview-49475804


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/loadbalancer/UpsertGoogleSslLoadBalancerAtomicOperation.groovy
```diff
@@ -287,7 +287,7 @@ class UpsertGoogleSslLoadBalancerAtomicOperation extends UpsertGoogleLoadBalance
         loadBalancingScheme: 'EXTERNAL',
         IPProtocol: description.ipProtocol,
         IPAddress: description.ipAddress,
-        portRange: description.certificate ? "443" : description.portRange,
```
danielpeach: I verified that the UI (as it exists now) sends `description.portRange == "443"`, rather than relying on this ternary, so this is OK.

sayap:
Thanks. Just rebased.

PR #1741, opened 2017-07-13
fix(aws): Short-circuit enable/disable when describe instance fails

robzienert:
Fixes a potential infinite loop that can be caused when a server group is disabled, then an instance is deleted. AWS will not update the list of known instances and will continually return an instance ID that we should describe which no longer exists.

This will short-circuit after 10 total failures.

@ajordens PTAL


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AbstractEnableDisableAtomicOperation.groovy
```diff
@@ -129,7 +130,12 @@ abstract class AbstractEnableDisableAtomicOperation implements AtomicOperation<V
 
             describeInstancesRequest.setNextToken(describeInstancesResult.nextToken)
           } catch(Exception e) {
```
ajordens: IMO I'd also handle the known instance does not exist exception explicitly ...

ajordens: The other thing would be doing a DescribeInstances w/ a filter of the asg being xyz ... vs fetching the ASG and doing `.instances`

PR #1743, opened 2017-07-14
fix(provider/kubernetes): Allow None Cluster IP

ethanfrogers:
Kubernetes allows creation of "headless" services by setting
`spec.clusterIP` to `None`. This allows `None` as the only non-ip value
that could be accepted.

Addresses spinnaker/spinnaker#1836

@lwander PTAL

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/validators/loadbalancer/UpsertKubernetesLoadBalancerAtomicOperationValidator.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/validators/loadbalancer/UpsertKubernetesLoadBalancerAtomicOperationValidator.groovy
@@ -58,7 +58,9 @@ class UpsertKubernetesLoadBalancerAtomicOperationValidator extends DescriptionVa
       helper.validateIpv4(ip, "externalIps[$idx]")
     }
 
-    description.clusterIp ? helper.validateIpv4(description.clusterIp, "clusterIp")  : null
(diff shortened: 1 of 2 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/validators/loadbalancer/UpsertKubernetesLoadBalancerAtomicOperationValidator.groovy
```diff
@@ -58,7 +58,9 @@ class UpsertKubernetesLoadBalancerAtomicOperationValidator extends DescriptionVa
       helper.validateIpv4(ip, "externalIps[$idx]")
     }
 
-    description.clusterIp ? helper.validateIpv4(description.clusterIp, "clusterIp")  : null
+    if (description.clusterIp && description.clusterIp != "None") {
```
lwander: Do you know if kubernetes' "None" is case-sensitive?

ethanfrogers: From my interpretation of the docs, yes.
- "Valid values are "None", empty string (""), or a valid IP address."

https://kubernetes.io/docs/api-reference/v1.6/#servicespec-v1-core

lwander: Neat, thanks!

PR #1746, opened 2017-07-16
fix(provider/kubernetes): Cross Account Clone

ethanfrogers:
Fixes cloning Server Groups across accounts. The previous implementation
was relying on the credentials of the target account which meant
cloning server groups that didn't exist in the target account failed.
This adds a `sourceCredentials` field to the Clone description that is
used to verify the source server group against the correct account.

Fixes spinnaker/spinnaker#1829

@lwander PTAL

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/converters/servergroup/CloneKubernetesAtomicOperationConverter.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/converters/servergroup/CloneKubernetesAtomicOperationConverter.groovy
@@ -23,6 +23,7 @@ import com.netflix.spinnaker.clouddriver.kubernetes.deploy.ops.servergroup.Clone
 import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation
 import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperations
 import com.netflix.spinnaker.clouddriver.security.AbstractAtomicOperationsCredentialsSupport
+import com.netflix.spinnaker.clouddriver.kubernetes.security.KubernetesNamedAccountCredentials
 import org.springframework.stereotype.Component
 
 @KubernetesOperation(AtomicOperations.CLONE_SERVER_GROUP)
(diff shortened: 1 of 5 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/converters/servergroup/CloneKubernetesAtomicOperationConverter.groovy
```diff
@@ -33,6 +34,9 @@ class CloneKubernetesAtomicOperationConverter extends AbstractAtomicOperationsCr
   }
 
   CloneKubernetesAtomicOperationDescription convertDescription(Map input) {
-    KubernetesAtomicOperationConverterHelper.convertDescription(input, this, CloneKubernetesAtomicOperationDescription)
+    def converted = KubernetesAtomicOperationConverterHelper.convertDescription(input, this, CloneKubernetesAtomicOperationDescription)
+    converted.sourceCredentials = (KubernetesNamedAccountCredentials) this.getCredentialsObject(converted.source.account)
```
lwander: Just wondering, why isn't this converted in `convertDescription`? If it was, we wouldn't need this extra handling

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/servergroup/CloneKubernetesAtomicOperation.groovy
```diff
@@ -72,7 +72,7 @@ class CloneKubernetesAtomicOperation implements AtomicOperation<DeploymentResult
 
     task.updateStatus BASE_PHASE, "Reading ancestor server group ${description.source.serverGroupName}..."
 
-    def credentials = description.credentials.credentials
+    def credentials = description.sourceCredentials.credentials
```
lwander: Related to above, what's preventing this from being `description.source.account`?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/converters/servergroup/CloneKubernetesAtomicOperationConverter.groovy
ethanfrogers: @lwander are you referring to `KubernetesAtomicOperationConverterHelper.convertDescription`? I handled it here because it's specific to the Clone description. I can move it, what did you have in mind?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/servergroup/CloneKubernetesAtomicOperation.groovy
ethanfrogers: `description.source.account` is `String` denoting the account name.

lwander:
Talked offline, keeping the conversion here makes the most sense, in particular since credential objects are typically not loaded within an operation.

PR #1753, opened 2017-07-21
feat(dynomite): Adds support for Dynomite as a persistence backend

robzienert:
It's been running in test all week, and we vetted the redis codepath yesterday during our fun.

I squashed all of the `dynomite` branch commits into this single one.

@spinnaker/reviewers PTAL


cats/cats-dynomite/src/main/java/com/netflix/spinnaker/cats/dynomite/cluster/DynoClusteredAgentScheduler.java
```diff
@@ -0,0 +1,217 @@
...
+
+  @Override
+  public void run() {
+    if (!nodeStatusProvider.isNodeEnabled()) {
+      return;
+    }
+    try {
+      runAgents();
+    } catch (Throwable t) {
+      t.printStackTrace(System.err);
```
ajordens: This really what you want?

cats/cats-dynomite/src/main/java/com/netflix/spinnaker/cats/dynomite/cluster/DynoClusteredAgentScheduler.java
```diff
@@ -0,0 +1,217 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Temporary clustered agent scheduler while we're waiting for Dyno client support of evalsha and loadscript.
+ *
+ * Shares a similiar strategy as ClusteredAgentScheduler, but doesn't use Lua, is slower and unsafe (does not use atomic
```
ajordens: Can you elaborate on this ... a risk that the same agent could run multiple times?

cats/cats-dynomite/src/main/java/com/netflix/spinnaker/cats/dynomite/cluster/DynoClusteredAgentScheduler.java
```diff
@@ -0,0 +1,217 @@
...
+      });
+    return acquired;
+  }
+
+  private boolean acquireRunKey(String agentType, long timeout) {
+    // This isn't safe, but it's the best one can do without `String set(String key, String value, String nxxx, String expx, long time).`
+    return redisClientDelegate.withCommandsClient(client -> {
+      String response = client.get(agentType);
+      if (response == null) {
+        if (client.setnx(agentType, nodeIdentity.getNodeIdentity()) == 1) {
```
ajordens: probably worth a comment on what it's actually doing and how it's different than non-dyno locking we do.

is the issue that it's two separate calls and someone could be sneaky in the middle?

cats/cats-dynomite/src/test/groovy/com/netflix/spinnaker/cat/dynomite/cache/DynomiteCacheSpec.groovy
```diff
@@ -0,0 +1,264 @@
...
+import com.netflix.spinnaker.cats.redis.cache.RedisCacheOptions
+import com.netflix.spinnaker.kork.jedis.EmbeddedRedis
+import redis.clients.jedis.Jedis
+import spock.lang.AutoCleanup
+import spock.lang.IgnoreIf
+import spock.lang.Shared
+import spock.lang.Unroll
+
+// TODO rz - Figure out how to get dyno client to connect w/ an embedded redis
+@IgnoreIf({ System.getProperty("dyno.address") == null })
```
ajordens: if not embedded I take it you've got a local dyno to test against?

cats/cats-dynomite/src/test/groovy/com/netflix/spinnaker/cat/dynomite/cache/DynomiteCacheSpec.groovy
```diff
@@ -0,0 +1,264 @@
...
+      this.key2 = key2
+    }
+  }
+
+  private void initLocalDynoClusterClient() {
+    directClient = new Jedis('192.168.99.100', 22122)
+    directClient.flushAll()
+    directClient.close()
+
+    // This setup assumes that you're running a single-node Dynomite cluster via the Docker image.
```
ajordens: worth a link to dynomite image (and anything special to run it)?

robzienert:
yolo

cats/cats-dynomite/src/main/java/com/netflix/spinnaker/cats/dynomite/cluster/DynoClusteredAgentScheduler.java
robzienert: This is actually a bit dated, now that we're doing `setnx` we should be a lot safer. This is our deployment-specific, but since we're running consistency of `ONE` in dyno_clouddriver, there is a chance for running multiple times if things are just right.

cats/cats-dynomite/src/main/java/com/netflix/spinnaker/cats/dynomite/cluster/DynoClusteredAgentScheduler.java
robzienert: Mmm, nope. I'll add another PR to fix that.

cats/cats-dynomite/src/main/java/com/netflix/spinnaker/cats/dynomite/cluster/DynoClusteredAgentScheduler.java
robzienert: That's right, there's a bit of sneakiness that can happen, in addition to the consistency thing I mentioned above. CDE recommends moving things like this, as well as the task repository stuff into its own cluster so it can run in quorum. However, once Lua is added (coming this quarter), we can switch to the other agent and we'll hash to the same server, having the same atomic guarantees for locks as we have in Redis today.

cats/cats-dynomite/src/test/groovy/com/netflix/spinnaker/cat/dynomite/cache/DynomiteCacheSpec.groovy
robzienert: That's right. I use the official dyno docker image.

cats/cats-dynomite/src/test/groovy/com/netflix/spinnaker/cat/dynomite/cache/DynomiteCacheSpec.groovy
robzienert: https://github.com/Netflix/dynomite/blob/dev/docker/HOWTO.md

Their official image didn't seem to start up correctly in Kitematic, so I built it locally. Running it, you just need to forward the 4 ports that it exposes (again, if Kitematic).

PR #1768, opened 2017-07-28
feat(provider/google): Adding region and zone filter checks to validators.

dbakshiyev:


duftler:
Nice.
Feels like we would want a different validation msg when the region/zone is not configured, as opposed to referring to a region/zone that does not exist. The former is 'unconfigured' (or something like that), as opposed to 'invalid'. Mentioning this since the Spinnaker user who encounters these messages may not be the same as the operator who configures Spinnaker itself.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/validators/StandardGceAttributeValidator.groovy
```diff
@@ -173,15 +173,15 @@ class StandardGceAttributeValidator {
 
   def validateRegion(String region, GoogleNamedAccountCredentials credentials) {
     if (validateNotEmpty(region, "region") && credentials) {
-      if (!(region in credentials.regionToZonesMap)) {
+      if (!(region in credentials.regions.collect { it.name })) {
```
duftler: We can use the groovy spread-dot operator here and below:

`credentials.regions*.name`

dbakshiyev: Done

dbakshiyev:
Added an 'unconfigured' validation msg for when the region/zone is not one of those configured.

PR #1777, opened 2017-08-07
feat(kubernetes): Provides ability to drive pod cache TTL via metadata attribute

danveloper:
This PR provides the ability to control the CATS cache TTL of a `KubernetesInstance` by adding a metadata annotation, `cache.spinnaker.io/ttl`, to the `Pod`.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesInstanceCachingAgent.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesInstanceCachingAgent.groovy
@@ -34,7 +34,7 @@ import static com.netflix.spinnaker.cats.agent.AgentDataType.Authority.AUTHORITA
 
 @Slf4j
 class KubernetesInstanceCachingAgent extends KubernetesCachingAgent {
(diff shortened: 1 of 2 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/agent/KubernetesInstanceCachingAgent.groovy
```diff
@@ -34,6 +34,8 @@ import static com.netflix.spinnaker.cats.agent.AgentDataType.Authority.AUTHORITA
 
 @Slf4j
 class KubernetesInstanceCachingAgent extends KubernetesCachingAgent {
+  static final String CACHE_TTL_ANNOTATION = "spinnaker.cache-ttl"
```
lwander: A small nit, but I'd like to make the annotations match k8s' designated annotations: https://kubernetes.io/docs/reference/labels-annotations-taints/

I think something like "cache.spinnaker.io/ttl" could make sense

danveloper: Ah! Thanks -- I wasn't super sure on this. Thanks for the pointer!

PR #1788, opened 2017-08-11
feat(cache): Make cache polling interval configurable

danveloper:
Allows the agent scheduler polling interval to be configurable.

```
cache:
  pollTime: 60
```


lwander:
I thought this was already configurable via
```
redis.poll.[intervalSeconds/timeoutSeconds]
```

PR #1794, opened 2017-08-15
feat(provider/ecs): foundation of the ECS cloud provider

BrunoCarrier:
feat(provider/ecs): foundation of the ECS cloud provider


BrunoCarrier:
@robzienert @tomaslin

clouddriver-ecs/src/main/java/com/netflix/spinnakerclouddriver/aws/ecs/EcsCloudProvider.java
```diff
@@ -0,0 +1,34 @@
+package com.netflix.spinnakerclouddriver.aws.ecs;
+
+import org.springframework.stereotype.Component;
+import com.netflix.spinnaker.clouddriver.core.CloudProvider;
+
+import java.lang.annotation.Annotation;
+
+//@Component
```
BrunoCarrier: Commenting this out prevents ECS to appear as a cloud provider at this time - is this sensical?  I can see arguments for both options, but would rather go for the same approach as you've used in the past while implementing new cloud providers @tomaslin @robzienert

BrunoCarrier: @tomaslin So you'd rather have the importing of EcsConfiguration removed, as well as the project dependency in clouddriver-web.gradle?  Or did I misunderstand what you wrote?

tomaslin: If you don't want this exposed, then yes. But I'm not sure why @Component is commented out above.

BrunoCarrier: My thinking was that ECS would appear as a cloud provider if @component is not commented out, and might give false hopes.  Perhaps that assertion of mine was incorrect, and leaving @component in would not mislead anyone?

robzienert: I would prefer to keep the module itself totally intact, but remove it from the `clouddriver-web` dependencies as Tomas suggested. This will reduce cognitive load of what needs to happen when we're all finally comfortable with the state of the module, it also means you can do local development more easily.

robzienert:
Also noticed one more thing: The package names / filepaths are incorrect:

```
clouddriver-ecs/src/main/java/com/netflix/spinnakerclouddriver/aws/ecs/EcsCloudProvider.java
```

Should be...

```
clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/aws/ecs/EcsCloudProvider.java
```

BrunoCarrier:
@robzienert @tomaslin 

Nice one, didn't notice the mistake in the package naming.  I corrected it

I removed ECS from main.groovy and clouddriver-web.gradle, as per the discussion.  The @component annotation is no longer commented out, as per the discussion as well

settings.gradle already had the "clouddriver-ecs" module, so I didn't modify that: https://github.com/spinnaker/clouddriver/blob/master/settings.gradle#L23

Does the PR now match your vision of it?

tomaslin:
The modules wouldn't have any effect unless ecs.enabled was set to true, given that the default would be false and the cloud provider wouldn't load. I would also just use ecs instead of aws/ecs to be consistent with appengine

BrunoCarrier:
@tomaslin @robzienert as per the slack discussion:
:32] 
1- leave @component enabled
2- import EcsConfiguration, which wont be enabled due to `ecs.enabled`
3- have clouddriver-web depend on the ECS module

BrunoCarrier:
@tomaslin @robzienert As per our discussion, I think we are good to go.  Is there anything missing?

PR #1796, opened 2017-08-15
fix(entityTags): gracefully handle missing account

icfantv:
* add error handling around when the entity ref account name is not
found in the account credentials provider.

### DIFF
--- a/clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/ops/BulkUpsertEntityTagsAtomicOperation.java
+++ b/clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/ops/BulkUpsertEntityTagsAtomicOperation.java
@@ -153,14 +153,14 @@ public class BulkUpsertEntityTagsAtomicOperation implements AtomicOperation<Bulk
 
     if (entityRefAccount != null && entityRefAccountId == null) {
       // add `accountId` if not explicitly provided
-      AccountCredentials accountCredentials = accountCredentialsProvider.getCredentials(entityRefAccount);
+      AccountCredentials accountCredentials = lookupAccountCredentialsByAccountIdOrName(accountCredentialsProvider, entityRefAccount, "accountName");
       entityRefAccountId = accountCredentials.getAccountId();
       entityRef.setAccountId(entityRefAccountId);
     }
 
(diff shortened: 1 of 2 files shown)


clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/ops/BulkUpsertEntityTagsAtomicOperation.java
```diff
@@ -246,8 +246,16 @@ private static void addTagMetadata(Date now, EntityTags entityTags) {
     entityTags.getTags().forEach(tag -> entityTags.putEntityTagMetadata(tagMetadata(tag.getName(), now)));
   }
 
-  private static AccountCredentials lookupAccountCredentials(AccountCredentialsProvider accountCredentialsProvider,
-                                                             String entityRefAccountId) {
+  private static AccountCredentials lookupAccountCredentialsByAccountName(AccountCredentialsProvider accountCredentialsProvider,
```
icfantv: I wasn't sure whether to do this this way or to simply use the `getCredentials` method and then handle a `null` returned - throwing an exception there.

ajordens: I dig it ... 

I'd also be ok with

```
private static AccountCredentials lookupAccountCredentials(AccountCredentialsProvider accountCredentialsProvider,
                                                             String entityRefAccountIdOrName) {
    return accountCredentialsProvider.getAll().stream()
      .filter(c -> entityRefAccountIdOrName.equals(c.getAccountId()) || entityRefAccountIdOrName.equals(c.getName()))
      .findFirst()
      .orElseThrow(() -> new IllegalArgumentException("No credentials found for account '" + entityRefAccountIdOrName + "'"));
  }
```

Just one method that takes a param representing either an account name or id and does the match ... slightly less duplicated code. It's a `private` method so the context is all kept within this class anyhow.

Otherwise LGTM

icfantv: Yea, agreed.  I did like the explicit error message, so I'll add a parameter to specify that.

PR #1814, opened 2017-08-21
fix(provider/gce): Avoid casting deserialized network to Network object.

duftler:

PR #1815, opened 2017-08-21
fix(provider/gce): Avoid casting deserialized network to Network obje…

ttomsu:
…ct. (#1814)


jtk54:
Cherry pick of #1814 into release-1.2.x -- for posterity.

PR #1817, opened 2017-08-21
feat(provider/kubernetes): Add option to use source capacity

AMeng:
Fixes https://github.com/spinnaker/spinnaker/issues/1657

Add an option to use source capacity when deploying a new k8s replica set. This feature exists for other cloud providers.

Follow-up work will be required in Deck to expose this option.


lwander:
Thanks!

AMeng:
@axot I never did the work to expose this option in Deck, so its possible that it isn't in the UI. You can still use the feature with the API or by editing the pipeline JSON in the UI.

PR #1820, opened 2017-08-22
feat(core): Version provider accounts/operations

lwander:
This is intended for the v2 k8s provider.

The idea is that accounts can expose a "version" which converters,
validators, and (in the future) caching agents will use to determine
when to be run. This is also useful if we want to start moving features
into release tracks, as we can flag operations with a "beta" version
that requires accounts to be marked with that version.

Sorry for the half-finished k8s work, I wanted to submit this before the
PR got too big

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AnnotationsBasedAtomicOperationsRegistry.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AnnotationsBasedAtomicOperationsRegistry.groovy
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.clouddriver.orchestration
 import com.netflix.spinnaker.clouddriver.core.CloudProvider
 import com.netflix.spinnaker.clouddriver.deploy.DescriptionValidator
 import com.netflix.spinnaker.clouddriver.exceptions.CloudProviderNotFoundException
+import com.netflix.spinnaker.clouddriver.security.ProviderVersion
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.NoSuchBeanDefinitionException
 import org.springframework.beans.factory.annotation.Autowired
@@ -32,10 +33,10 @@ class AnnotationsBasedAtomicOperationsRegistry extends ApplicationContextAtomicO
   List<CloudProvider> cloudProviders
 
   @Override
-  AtomicOperationConverter getAtomicOperationConverter(String description, String cloudProvider) {
+  AtomicOperationConverter getAtomicOperationConverter(String description, String cloudProvider, ProviderVersion version) {
     // Legacy naming convention which is not generic and description name is specific to cloud provider
     try {
-      AtomicOperationConverter converter = super.getAtomicOperationConverter(description, cloudProvider)
+      AtomicOperationConverter converter = super.getAtomicOperationConverter(description, cloudProvider, version)
       if (converter) return converter
     } catch (NoSuchBeanDefinitionException e) {
       /**
@@ -57,16 +58,18 @@ class AnnotationsBasedAtomicOperationsRegistry extends ApplicationContextAtomicO
       value instanceof AtomicOperationConverter
     }.values().toList()
 
+    converters = VersionedOperationHelper.findVersionMatches(version, converters)
+
     if (!converters) {
       throw new AtomicOperationConverterNotFoundException(
-        "No atomic operation converter found for description '${description}' and cloud provider '${cloudProvider}'"
+          "No atomic operation converter found for description '${description}' and cloud provider '${cloudProvider}' at version '${version}'"
       )
     }
 
     if (converters.size() > 1) {
       throw new RuntimeException(
         "More than one (${converters.size()}) atomic operation converters found for description '${description}' and cloud provider " +
-          "'${cloudProvider}'"
+          "'${cloudProvider}' at version '${version}'"
       )
     }
 
@@ -74,10 +77,10 @@ class AnnotationsBasedAtomicOperationsRegistry extends ApplicationContextAtomicO
   }
 
   @Override
-  DescriptionValidator getAtomicOperationDescriptionValidator(String validator, String cloudProvider) {
+  DescriptionValidator getAtomicOperationDescriptionValidator(String validator, String cloudProvider, ProviderVersion version) {
     // Legacy naming convention which is not generic and validator name is specific to cloud provider
     try {
-      DescriptionValidator descriptionValidator = super.getAtomicOperationDescriptionValidator(validator, cloudProvider)
+      DescriptionValidator descriptionValidator = super.getAtomicOperationDescriptionValidator(validator, cloudProvider, version)
       if (descriptionValidator) {
         return descriptionValidator
       }
@@ -91,6 +94,8 @@ class AnnotationsBasedAtomicOperationsRegistry extends ApplicationContextAtomicO
       DescriptionValidator.getValidatorName(value.getClass().getAnnotation(providerAnnotationType).value()) == validator &&
       value instanceof DescriptionValidator
     }.values().toList()
+
+    validators = VersionedOperationHelper.findVersionMatches(version, validators)
 
     return validators ? (DescriptionValidator) validators[0] : null
   }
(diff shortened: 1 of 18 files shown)


lwander:
spinnaker/spinnaker#1902

lwander:
https://github.com/spinnaker/spinnaker/issues/1916

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AnnotationsBasedAtomicOperationsRegistry.groovy
```diff
@@ -57,27 +58,29 @@ class AnnotationsBasedAtomicOperationsRegistry extends ApplicationContextAtomicO
       value instanceof AtomicOperationConverter
     }.values().toList()
 
+    converters = converters.findAll { AtomicOperationConverter cv -> cv.acceptsVersion(version) }
+
     if (!converters) {
       throw new AtomicOperationConverterNotFoundException(
-        "No atomic operation converter found for description '${description}' and cloud provider '${cloudProvider}'"
+          "No atomic operation converter found for description '${description}' and cloud provider '${cloudProvider}' at version '${version}'"
```
duftler: Accidentally indented?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AnnotationsBasedAtomicOperationsRegistry.groovy
```diff
@@ -57,27 +58,29 @@ class AnnotationsBasedAtomicOperationsRegistry extends ApplicationContextAtomicO
       value instanceof AtomicOperationConverter
     }.values().toList()
 
+    converters = converters.findAll { AtomicOperationConverter cv -> cv.acceptsVersion(version) }
```
duftler: Why not pull this out into a helper function?

`converters = VersionHelper.findVersionMatches(converters)`
`validators = VersionHelper.findVersionMatches(validators)`

Or something injected if you don't want to use the static (to ease testing).

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/ApplicationContextAtomicOperationsRegistry.groovy
```diff
@@ -31,12 +32,22 @@ class ApplicationContextAtomicOperationsRegistry implements AtomicOperationsRegi
   ApplicationContext applicationContext
 
   @Override
-  AtomicOperationConverter getAtomicOperationConverter(String description, String cloudProvider) {
-    (AtomicOperationConverter) applicationContext.getBean(description)
+  AtomicOperationConverter getAtomicOperationConverter(String description, String cloudProvider, ProviderVersion version) {
+    def result = (AtomicOperationConverter) applicationContext.getBean(description)
+    if (!result.acceptsVersion(version)) {
```
duftler: Maybe:

`VersionHelper.verifyMatch(result, version, description)`
`VersionHelper.verifyMatch(result, version, validator)`

duftler: Since the helper function would directly throw the exception, could even do this:

`return VersionHelper.verifyMatch(applicationContext.getBean(description), version, description)`
`return VersionHelper.verifyMatch(applicationContext.getBean(validator), version, validator)`

Thought it would feel weird to pass the `applicationContext`, so just inlined that call above.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/VersionedCloudProviderOperation.groovy
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+
+package com.netflix.spinnaker.clouddriver.orchestration
+
+import com.netflix.spinnaker.clouddriver.security.ProviderVersion
+
+trait VersionedCloudProviderOperation {
+  /**
+   * Various converters can satisfy different provider's versions. This converter will only be applicable to accounts
```
duftler: Applies to more than just `Converters`, right?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/VersionedCloudProviderOperation.groovy
```diff
@@ -0,0 +1,32 @@
...
+   * at this version.
+   *
+   * @return true i.f.f. this converter works accounts at this version
```
duftler: `works accounts` is missing a word.

duftler:
I like the approach. Would it be possible to break out the k8s v2/manifest work into a follow-in PR? I think if you can do that it will make life easier for future readers trying to figure out how the versioning works by reading this PR.
The versioning logic and the k8s v2/manifest work all LGTM.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AnnotationsBasedAtomicOperationsRegistry.groovy
lwander: I think if you add a newline to break up a line you typically indent twice, that's what intellij did for me when I made the edit here. Does that sound right?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AnnotationsBasedAtomicOperationsRegistry.groovy
lwander: Oh good idea, I know that in the future if we start to lean on this feature we'll want to define how to resolve a match when more than one version can be used (e.g. v1 can be used in place of v1beta1 safely). Putting that in a helper now will make adding that later easier.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/ApplicationContextAtomicOperationsRegistry.groovy
lwander: Why move this into the version helper? Would the individual converters/validators still implement `acceptsVersion()`? Or is the idea to move the logic someplace else?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/VersionedCloudProviderOperation.groovy
lwander: Yes, whoops

lwander:
Will do

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AnnotationsBasedAtomicOperationsRegistry.groovy
duftler: It caught my eye because it doesn't match the style just below.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/ApplicationContextAtomicOperationsRegistry.groovy
duftler: I think they can still implement `acceptsVersion()`, it was just to reduce boilerplate code at the call-sites. The exception messages and everything else match.

lwander: SGTM

lwander: The messages differ though, so the helper would be taking four parameters which feels weird. I feel like moving it into another class to save a few lines of code breaks up the readability too much. I'm going to leave it as-is for now, let me know if you have strong feelings.

lwander: Messages differ beyond the current template *

lwander:
Remove all non-provider-version kubernetes code.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/ApplicationContextAtomicOperationsRegistry.groovy
duftler: ok, sgtm

PR #1821, opened 2017-08-22
perf(core): Optimize `ClusterController.getServerGroup` when region =…

ajordens:
… null

Added ?expand=<boolean> request parameters to every ClusterController
endpoint.


ajordens:
I will need to make a similar change to the internal `TitusClusterProvider`

PR #1822, opened 2017-08-22
perf(core): Align the results from /clusters/{cluster}/serverGroups a…

ajordens:
…nd /serverGroups/

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonClusterProvider.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonClusterProvider.groovy
@@ -86,15 +86,12 @@ class AmazonClusterProvider implements ClusterProvider<AmazonCluster> {
     String imageId = launchConfigs?.attributes?.get('imageId')
     CacheData imageConfigs = imageId ? cacheView.get(IMAGES.ns, Keys.getImageKey(imageId, account, region)) : null
 
-    def serverGroup = new AmazonServerGroup(name: name, region: region)
+    def serverGroup = new AmazonServerGroup(serverGroupData.attributes)
     serverGroup.accountName = account
-    serverGroup.zones = serverGroupData.attributes["zones"]
     serverGroup.launchConfig = launchConfigs ? launchConfigs.attributes : null
     serverGroup.image = imageConfigs ? imageConfigs.attributes : null
     serverGroup.buildInfo = imageConfigs ? getBuildInfoFromImage(imageConfigs) : null
-    serverGroup.asg = asg
-    serverGroup.scalingPolicies = serverGroupData.attributes["scalingPolicies"]
-    serverGroup.scheduledActions = serverGroupData.attributes["scheduledActions"]
+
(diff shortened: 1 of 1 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonClusterProvider.groovy
```diff
@@ -86,15 +86,12 @@ class AmazonClusterProvider implements ClusterProvider<AmazonCluster> {
     String imageId = launchConfigs?.attributes?.get('imageId')
     CacheData imageConfigs = imageId ? cacheView.get(IMAGES.ns, Keys.getImageKey(imageId, account, region)) : null
 
-    def serverGroup = new AmazonServerGroup(name: name, region: region)
+    def serverGroup = new AmazonServerGroup(serverGroupData.attributes)
     serverGroup.accountName = account
```
ajordens: The result from `/clusters/{cluster}/serverGroups/{serverGroup}` does not contain `accountName` ... but that should be the only difference.

Previously we had a number of null attributes being returned from `/serverGroups/{serverGroup}` that were there when fetched via the cluster.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonClusterProvider.groovy
```diff
@@ -148,6 +145,9 @@ class AmazonClusterProvider implements ClusterProvider<AmazonCluster> {
       loadBalancers = translateLoadBalancers(allLoadBalancers)
       targetGroups = translateTargetGroups(allTargetGroups)
       serverGroups = translateServerGroups(allServerGroups)
+    } else {
+      Collection<CacheData> allServerGroups = resolveRelationshipDataForCollection(clusterData, SERVER_GROUPS.ns, RelationshipCacheFilter.include(LAUNCH_CONFIGS.ns))
```
ajordens:
```
def sortedServerGroups = getServerGroups(application, account, clusterName, cloudProvider, null /* region */, false).findAll {
      def scopeMatch = it.region == scope || it.zones.contains(scope)
```

If `LAUNCH_CONFIGS` were not included, `it.zones` was null and the previously working check would NPE.

The big perf overhead comes with loading prematurely instances so I'd expect including `LAUNCH_CONFIGS` to not be material.

ajordens:
```

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonClusterProvider.groovy
cfieber: curious why including LAUNCH_CONFIGS made zones appear

cfieber: I think its useful to include launch config anyway because that brings in buildInfo (or maybe we need to also include IMAGE relationship to get that)...

PR #1825, opened 2017-08-23
fix(provider/google): Handling edge cases with user data files better.

varikin:
This incorporates 3 changes/improvements for handling user data files. It
allows blank lines in the user data file. It also interprets the lack of an
equal sign as a key with no value. Finally, it allows values to contain equal
signs, in other words, it only splits on the first equal sign.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/GoogleUserDataProvider.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/GoogleUserDataProvider.groovy
@@ -23,7 +23,6 @@ import groovy.transform.PackageScope
 import groovy.util.logging.Slf4j
 import org.springframework.stereotype.Component
 
-
 /**
  * Provides the common user data from a local file to be applied to all Google deployments.
  *
@@ -41,10 +40,11 @@ public class GoogleUserDataProvider {
                   BasicGoogleDeployDescription description,
                   GoogleNamedAccountCredentials credentials, String customUserData) {
     String userDataFile = credentials.getUserDataFile()
-    String rawUserData = getFileContents(userDataFile)
-    String commonUserData = replaceTokens(rawUserData, description, serverGroupName, instanceTemplateName)
+    List<String> rawUserData = getFileContents(userDataFile)
+    List<String> cleanedUserData = cleanUpUserData(rawUserData)
+    List<String> commonUserData = replaceTokens(cleanedUserData, description, serverGroupName, instanceTemplateName)
 
-    Map userDataMap = stringToUserDataMap(commonUserData)
+    Map<String, String> userDataMap = convertToMap(commonUserData)
 
     if (customUserData) {
       def customUserDataMap = ["customUserData": customUserData] << stringToUserDataMap(customUserData)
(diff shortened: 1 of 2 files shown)


varikin:
@dbakshiyev @duftler here is my update to the user data file stuff. I hope my commit/pr message has enough context. The major issue was that blank lines in the udf file were turned into key/value pairs of empty strings which GCE does not like. Beyond that, I cleaned up some edge cases and refactored some things so I could test the filtering more easily.

varikin:
Hmm...And I meant to make the type in the commit message by fix, not feature.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -375,8 +375,7 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
     }
 
     List<InstanceGroupManagerAutoHealingPolicy> autoHealingPolicy =
-      description.autoHealingPolicy?.healthCheck
-      ? [new InstanceGroupManagerAutoHealingPolicy(
+      description.autoHealingPolicy?.healthCheck ? [new InstanceGroupManagerAutoHealingPolicy(
```
duftler: Is this reformatting intentional? Don't see a code change for this line.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/GoogleUserDataProvider.groovy
```diff
@@ -92,40 +101,45 @@ public class GoogleUserDataProvider {
...
-      .replace('%%region%%', description.region ?: '')
-      .replace('%%stack%%', description.stack ?: '')
-      .replace('%%group%%', names.group ?: '')
-      .replace('%%cluster%%', names.cluster ?: '')
-      .replace('%%detail%%', names.detail ?: '')
-      .replace('%%launchconfig%%', instanceTemplateName ?: '')
-
-    return result
+    return rawUserData.collect{userData ->
+      return userData.replace('%%account%%', description.accountName)
```
duftler: I think it reads slightly better with all of the `.replace()`'s aligned.

duftler:
Could also update the spinnaker [readme](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/UserData.md) on userdata to describe the edge cases mentioned in this pr. Seems like that shouldn't exist just under clouddriver-aws, so we should probably move that elsewhere.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
varikin: Intellij was complaining about the `?` on the second line

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/GoogleUserDataProvider.groovy
varikin: I can do that.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
duftler: Yeah. The folks here make fun of me for formatting it this way. I prefer the original style, but can live with it either way. I do typically try to avoid reformatting lines that don't have code changes unless it really affects readability.

duftler:
Thanks for digging into this.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
varikin: I can adjust it. I might have some intellij setting that was flagging it as an error and so wouldn't compile.

varikin:
@duftler would you prefer updating the readme or updating spinnaker.io documentation? I added a section when doing the [Openstack work to include it](https://www.spinnaker.io/setup/providers/openstack/#common-userdata). I can update and move that to a common area in the docs if you would prefer that.

duftler:
@varikin I think the best approach would be to do what you proposed: update the spinnaker.io/openstack/userdata section and move it to a common area (and probably list what providers have support for this feature).

duftler:
Going to cherrypick this commit into a new branch so I can merge it without you having to rebase again.

duftler:
Probably also want to add some text describing the 3 edge cases covered in this pr.

duftler:
Closing this PR since the commit was cherrypicked in #1852.

PR #1826, opened 2017-08-23
feat(provider/ecs): Introduced atomic operation stubs

BrunoCarrier:
@robzienert  @tomaslin @ajordens @cfieber 
I created stubs for most server group atomic operations.  Although the number of files is quite high for a pull request, there is no code logic apart from the design of atomic operations, which is meant to mimic the design of the AWS module.

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/converters/CloneServiceAtomicOperationConverter.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/converters/CloneServiceAtomicOperationConverter.java
@@ -1,0 +1,32 @@
+package com.netflix.spinnaker.clouddriver.ecs.deploy.converters;
+
+import com.netflix.spinnaker.clouddriver.deploy.DeployAtomicOperation;
+import com.netflix.spinnaker.clouddriver.ecs.EcsOperation;
+import com.netflix.spinnaker.clouddriver.ecs.deploy.description.BasicEcsDeployDescription;
+import com.netflix.spinnaker.clouddriver.ecs.deploy.description.CloneServiceDescription;
+import com.netflix.spinnaker.clouddriver.ecs.deploy.ops.CloneServiceAtomicOperation;
+import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation;
+import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperations;
+import com.netflix.spinnaker.clouddriver.security.AbstractAtomicOperationsCredentialsSupport;
+import org.springframework.stereotype.Component;
+
+import java.util.Map;
+
+@EcsOperation(AtomicOperations.CLONE_SERVER_GROUP)
+@Component("ecsCloneLastService")
+public class CloneServiceAtomicOperationConverter extends AbstractAtomicOperationsCredentialsSupport {
+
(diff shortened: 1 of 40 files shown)


clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/converters/UpdateServiceConfigAtomicOperationConverter.java
```diff
@@ -0,0 +1,32 @@
...
+import com.netflix.spinnaker.clouddriver.ecs.deploy.ops.UpdateServiceConfigAtomicOperation;
+import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation;
+import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperations;
+import com.netflix.spinnaker.clouddriver.security.AbstractAtomicOperationsCredentialsSupport;
+import org.springframework.stereotype.Component;
+
+import java.util.Map;
+
+@EcsOperation(AtomicOperations.UPDATE_LAUNCH_CONFIG)
+@Component("ecsUpdateLaunchConfig")
```
robzienert: From what I understand of ECS, wouldn't this actually be something more like `ecsUpdateTaskDefinition`?

BrunoCarrier: Hi @robzienert , thanks for reviewing :)

You are kind of right.  Services launch tasks using task definitions.  However, you cannot modify a task definition - you can only make a new revision.  Therefore you need to create a new task definition revision, and then tell your service to use that new task definition revision.  

Also, some parameters would clearly fall on the service rather than the task definition.  For instance, the number of tasks to run is at the service level, while the task definition defines what resources a task may reserve

Does this make more sense now?

robzienert: I suppose so - but launch configurations are also immutable. I'm just concerned about conflation of an actual AWS resource name with what sounds to be an composite resource that you're creating. So, ECS Launch Config, effectively, would touch both a Service _and_ Task Definitions?

BrunoCarrier: Yes @robzienert, it would be a composite resource.  What's your guidance here?

robzienert: The name feels weird, but I'm not dogmatic about changing it if you think it correctly describes its purpose.

BrunoCarrier: @robzienert I don't really mind what the name is.  If it wasn't clear to you, it is likely to violate the POLA in others as well.  I'm updating it to `UpdateServiceAndTaskConfigAtomicOperation`.  Not sexy, but at least it's clear

BrunoCarrier:
Is there any pending change required on this PR?

PR #1832, opened 2017-08-24
feat(dynomite): hashtags & dynomite pipelines

robzienert:
The latest round of dyno changes dramatically improved perf across the board, but it's still slow enough to rustle up some race conditions.

The dyno client currently doesn't support hashtags, but I wanted to put this out here so I don't forget about the work. This will further cut down the RTT dynomite latency. All keys for a provider's caching agent will get pinned to a single node, allowing most ops to get pipelined in `DynomiteCache`.


cats/cats-dynomite/src/main/java/com/netflix/spinnaker/cats/dynomite/cache/DynomiteCache.java
```diff
@@ -345,10 +369,49 @@ private int getHashExpiry() {
   }
 
   private String itemId(String type, String id) {
-    return String.format("%s:%s:item:%s", prefix, type, id);
```
robzienert: This weirdness will go away.

PR #1835, opened 2017-08-24
feat(provider/kubernetes): V2 operation init

lwander:
Continuation of https://github.com/spinnaker/clouddriver/pull/1820

spinnaker/spinnaker#1902
spinnaker/spinnaker#1916
spinnaker/spinnaker#1922

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/converter/KubernetesDeployManifestConverter.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/converter/KubernetesDeployManifestConverter.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.clouddriver.kubernetes.v2.converter;
+
+import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation;
+import com.netflix.spinnaker.clouddriver.security.AbstractAtomicOperationsCredentialsSupport;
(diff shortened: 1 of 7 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesApiVersion.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ */
+
+package com.netflix.spinnaker.clouddriver.kubernetes.v2.description;
+
+import java.util.Arrays;
+
+public enum KubernetesApiVersion {
+  V1("v1"),
+  EXTENSIONS_V1BETA1("extensions/v1beta1");
```
jtk54: Does this correspond to v2 of the Kubernetes provider? Will there be a mapping defined between Kubernetes API version and provider version?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/converter/KubernetesDeployManifestConverter.java
```diff
@@ -0,0 +1,41 @@
...
+import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation;
+import com.netflix.spinnaker.clouddriver.security.AbstractAtomicOperationsCredentialsSupport;
+import com.netflix.spinnaker.clouddriver.security.ProviderVersion;
+
+import java.util.Map;
+
+public class KubernetesDeployManifestConverter extends AbstractAtomicOperationsCredentialsSupport {
+  @Override
+  public AtomicOperation convertOperation(Map input) {
+    throw new RuntimeException("Not implemented");
```
duftler: Could throw this: https://docs.oracle.com/javase/7/docs/api/java/lang/UnsupportedOperationException.html

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesApiVersion.java
lwander: No these are Kubernetes-only release tracks. The client library is broken down by operations on each of these, so we need to be able to pull the right release track from a manifest to know how to update/deploy/destroy it.

We want to support all of these, so no mapping is really needed. What is important will be to connect Spinnaker versions to Kubernetes versions (both separate from provider version and kubernetes api version entirely).

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/converter/KubernetesDeployManifestConverter.java
lwander: SGTM

PR #1836, opened 2017-08-25
fix(provider/dcos): Fully load server group details before sorting

willgorman:
Delay sorting of server groups in `ClusterController#getTargetServerGroup` until they are fully loaded.  Fixes spinnaker/spinnaker#1927.

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ClusterController.groovy
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ClusterController.groovy
@@ -201,7 +201,7 @@ class ClusterController {
     }
 
     // load all server groups w/o instance details (this is reasonably efficient)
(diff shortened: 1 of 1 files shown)


willgorman:
@ajordens Looks like #1842 should make this unnecessary?

ajordens:
Correct, that should render this unnecessary.

At some point I suspect there's value in doing a review of non-AWS cloud providers and understanding what needs to happen to each in order to support loading "minimal" clusters.

A "minimal" cluster should include server groups with (at least?) the following details present:
- disabled state
- created timestamps
- regions/zones

The current behavior loads fully expanded clusters (including server groups w/ full instance details) and then filters server groups by name/region. This gets quite slow when you have clusters w/ multiple server groups and _many_ instances.

I've fixed this for Amazon but in-hindsight should have left the other providers to their respective authors (hence #1842).

ajordens:
Feel free to close this after you've confirmed that dc/os is back to working as expected.

PR #1843, opened 2017-08-25
fix(authz): Ensure operation checker does not NPE on null annotations.

duftler:


duftler:
Without this change, Clouddriver will not start up with zero cloud providers configured.

PR #1848, opened 2017-08-28
feat(provider/ecs): Made EcsCloudProvider public

BrunoCarrier:
feat(provider/ecs): Made EcsCloudProvider public


robzienert:
I think PRs of this size can be rolled up into subsequent PRs, just because it takes forever to build and release clouddriver.

PR #1856, opened 2017-08-29
feat(provider/kubernetes): V2 deploy other resources

lwander:


lwander:
spinnaker/spinnaker#1902
spinnaker/spinnaker#1916
spinnaker/spinnaker#1922

PR #1857, opened 2017-08-29
fix(discovery): skip terminated instances when enabling a disabled asg

asher:
When enabling & scaling up a disabled ASG - if instances are no longer running when trying to mark them up in discovery, log and ignore instead of failing the operation.


ajordens:
LGTM

PR #1864, opened 2017-08-30
refactor(provider/kubernetes): Pull v1 cache code into separate class

lwander:
Some of the loops being broken from .each to for() is due to groovy weirdness (language bugs)


lwander:
spinnaker/spinnaker#1902
spinnaker/spinnaker#1919

lwander:
spinnaker/spinnaker#1923

PR #1866, opened 2017-08-31
feat(provider/kubernetes): Add spectator to v2 client

lwander:


lwander:
spinnaker/spinnaker#1902
spinnaker/spinnaker#1916

PR #1874, opened 2017-09-04
fix(provider/amazon): Handle missing IAM permissions in caching agent

jervi:
Clouddriver will crash if the spinnaker role is missing permission for `ListServerCertificates`. This fix will log a warning with the affected account ID to make it easier to fix.
The warning will only be logged once every ~~hour~~ ten minutes for each affected account.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonCertificateCachingAgent.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonCertificateCachingAgent.groovy
@@ -16,10 +16,17 @@
 
 package com.netflix.spinnaker.clouddriver.aws.provider.agent
 
+import com.amazonaws.services.identitymanagement.model.AmazonIdentityManagementException
 import com.amazonaws.services.identitymanagement.model.ListServerCertificatesRequest
 import com.amazonaws.services.identitymanagement.model.ServerCertificateMetadata
 import com.fasterxml.jackson.databind.ObjectMapper
-import com.netflix.spinnaker.cats.agent.*
+import com.netflix.spectator.api.Id
+import com.netflix.spectator.api.Registry
+import com.netflix.spinnaker.cats.agent.AccountAware
+import com.netflix.spinnaker.cats.agent.AgentDataType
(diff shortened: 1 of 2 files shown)


gardleopard:
@spinnaker/reviewers PTAL

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonCertificateCachingAgent.groovy
```diff
@@ -76,29 +85,40 @@ class AmazonCertificateCachingAgent implements CachingAgent, AccountAware {
...
+      while (true) {
+        try {
+          def resp = iam.listServerCertificates(listServerCertificatesRequest)
+          iamCertificates.addAll(resp.serverCertificateMetadataList)
+          if (resp.marker) {
+            listServerCertificatesRequest.withMarker(resp.marker)
+          } else {
+            break
+          }
+        } catch (AmazonServiceException e) {
```
cfieber: as this stands, any exception calling the aws api is going to result in clearing the certificate list for an hour (on this instance)

I'd suggest: 
- trim the time down to something smaller (10 minutes maybe?)
- be more specific in the check on the AmazonServiceException to ensure it is actually a permission denied and not some other random service failure
- add a counter metric (`Registry.counter`) so that these can be surfaced to an alerting system that is not log based

jervi: Thanks @cfieber. I've taken a shot at implementing all three suggestions. I'm unsure of the best practices for naming and tagging the metrics counter, so suggestions are more than welcome 👍

jervi: @cfieber Can you please have another look at this?

costimuraru:
👍

robzienert:
@jervi Re-visiting this, I forgot to merge the last time I updated the branch. Still needed?

jervi:
We're running a custom clouddriver build with this patch applied. We find it handy, so I think it should be merged, yes. I'll rebase it.

PR #1876, opened 2017-09-05
feat(provider/kubernetes): Infer relationships from ownerReference

lwander:
The `ownerReferences` are a list of version/kind/name resources that created a particular resource. An example would be a replica set reference within a pod. This simplifies manually constructing relationships in the cache.


lwander:
spinnaker/spinnaker#1902
spinnaker/spinnaker#1919

PR #1879, opened 2017-09-06
feat(provider/kubernetes): Cache pods

lwander:
Moves shared caching logic into super class KubernetesV2CachingAgent


lwander:
spinnaker/spinnaker#1902
spinnaker/spinnaker#1919

PR #1880, opened 2017-09-06
feat(provider/kubernetes): Annotate templates as well

lwander:
spinnaker/spinnaker#1902
spinnaker/spinnaker#1919

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesManifest.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesManifest.java
@@ -28,6 +28,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 public class KubernetesManifest extends HashMap<String, Object> {
   private static <T> T getRequiredField(KubernetesManifest manifest, String field) {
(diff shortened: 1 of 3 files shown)


lwander:
spinnaker/spinnaker#1902
spinnaker/spinnaker#1919

lwander:
https://github.com/spinnaker/spinnaker/issues/1922

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesManifest.java
```diff
@@ -102,6 +103,32 @@ public void setNamespace(String namespace) {
...
+    Map<String, Object> template = (Map<String, Object>) spec.get("template");
+    if (!template.containsKey("metadata")) {
+      return Optional.empty();
+    }
+
+    Map<String, Object> metadata = (Map<String, Object>) template.get("metadata");
+    Map<String, String> result = (Map<String, String>) metadata.get("annotations");
+    if (result == null) {
+      result = new HashMap<>();
+      metadata.put("annotations", result);
```
danielpeach: It feels like you should do this outside of this method if it turns out `result == null`, but it took so much work to get here that it's probably not worth it.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesManifestAnnotater.java
```diff
@@ -62,7 +63,15 @@ private static void storeAnnotation(Map<String, String> annotations, String key,
 
   public static void annotateManifestWithRelationships(KubernetesManifest manifest, KubernetesManifestSpinnakerRelationships relationships) {
     Map<String, String> annotations = manifest.getAnnotations();
+    annotateManifestWithRelationships(annotations, relationships);
 
+    manifest.getSpecTemplateAnnotations().flatMap((a) -> {
```
danielpeach: Don't you usually omit the parens if there's only one parameter here? Just wondering.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesManifest.java
lwander: That's true, and it's also annoying to have your new map's edits not be automatically reflected in the manifest object.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesManifestAnnotater.java
lwander: Good point

danielpeach: Actually why is this a `flatMap` and not an `ifPresent` or something?

lwander: Because this lets you avoid using a control flow structure. The fewer `if` statements the better

PR #1888, opened 2017-09-12
 fix(provider/aws): ensure STSAssumeRoleSessionCredentialsProvider ge…

eisig:
NetflixSTSAssumeRoleSessionCredentialsProvider would always using the default sts endpoint. It would not work in cn-north-1.
It work when using the builder style.
I open a issue.[1945](https://github.com/spinnaker/spinnaker/issues/1949)


robzienert:
Nice catch - thanks!

PR #1890, opened 2017-09-13
feat(provider/docker): add catalogFile option

ethanfrogers:
adds the `catalogFile` option to the Docker Registry provider. This is
an attempt to make it easier to use registries that don't support
`v2/_catalog` (which happens to be a lot). This file can be managed by
an external process. Each caching cycle, Clouddriver reads this file to
determine the list of repositories it should cache tags for.

the format of the `configFile` is:
```
{
  "repositories":[
    "library/nginx",
    "library/busybox"
  ]
}
```

@lwander PTAL. This is the first iteration/POC for spinnaker/spinnaker#1909. We can implement more remote options (S3/GCS) when there's more precedence for it in Clouddriver.

### DIFF
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
@@ -51,6 +51,7 @@ class DockerRegistryClient {
     File dockerconfigFile
     long clientTimeoutMillis
     int paginateSize
+    String catalogFile
 
     Builder address(String address) {
       this.address = address
@@ -92,16 +93,22 @@ class DockerRegistryClient {
       return this
     }
 
+    Builder catalogFile(String catalogFile) {
+      this.catalogFile = catalogFile
+      return this
+    }
+
     DockerRegistryClient build() {
+
       if (password && passwordFile) {
         throw new IllegalArgumentException('Error, at most one of "password", "passwordFile", or "dockerconfigFile" can be specified')
       }
       if (password) {
-        return new DockerRegistryClient(address, email, username, password, clientTimeoutMillis, paginateSize)
+        return new DockerRegistryClient(address, email, username, password, clientTimeoutMillis, paginateSize, catalogFile)
       } else if (passwordFile) {
-        return new DockerRegistryClient(address, email, username, passwordFile, clientTimeoutMillis, paginateSize)
+        return new DockerRegistryClient(address, email, username, passwordFile, clientTimeoutMillis, paginateSize, catalogFile)
       } else {
-        return new DockerRegistryClient(address, clientTimeoutMillis, paginateSize)
+        return new DockerRegistryClient(address, clientTimeoutMillis, paginateSize, catalogFile)
       }
     }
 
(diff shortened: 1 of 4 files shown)


clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -270,6 +279,17 @@ class DockerRegistryClient {
    * don't want you to download their whole catalog (it's potentially a lot of data).
    */
   public DockerRegistryCatalog getCatalog(String path = null) {
+    if(catalogFile) {
```
lwander: Missing space before `(`

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/config/DockerRegistryConfigurationProperties.groovy
```diff
@@ -53,6 +53,8 @@ class DockerRegistryConfigurationProperties {
     // or <library> for repositories like 'ubuntu'.
     List<String> repositories
     List<String> skip
+    // a file listing all targetd repositories
```
lwander: s/targetd/targeted

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -270,6 +279,17 @@ class DockerRegistryClient {
...
+      log.info("Using catalog list at $catalogFile")
+      try {
+        def userDefinedCatalog = new BufferedReader(new FileReader(catalogFile)).getText()
+        return (DockerRegistryCatalog) new Gson().fromJson(userDefinedCatalog, DockerRegistryCatalog.class)
+      } catch(Exception e) {
```
lwander: Missing space before `(`

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -270,6 +279,17 @@ class DockerRegistryClient {
...
+        log.warn("Error encountered reading catalogFile" + e.getMessage())
+        return new DockerRegistryCatalog(repositories: [])
```
lwander: It might be best to fail to add the account here, this way the logs have a big stack trace with the exception in it rather than a single warning line

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/config/DockerRegistryConfigurationProperties.groovy
```diff
@@ -53,6 +53,8 @@ class DockerRegistryConfigurationProperties {
...
+    File catalogFile
```
lwander: How are `File` classes serialized/deserialized? This might be better suited as a `String` and then converted

ethanfrogers: Would we want to switch it up for `passwordFile` as well?

PR #1895, opened 2017-09-14
(feat/cats) Add support for an error interval for clustered agent sch…

rfarrjr:
…edulers

The following change introduces a new errorInterval used by the ClusteredSortAgentScheduler and ClusteredAgentScheduler.  This interval is used when the Agent throws an exception on a run.

Once this PR is accepted an released a future PR will follow for fiat to leverage this functionality.

This PR is the result of previous discussion with @cfieber.


rfarrjr:
Thanks.  I can look over the Dynomite scheduler today.   Would you prefer I add it to this PR or create a separate one?

robzienert:
Same one is fine

rfarrjr:
here it is, was't too bad

rfarrjr:
ping @spinnaker/reviewers

PR #1897, opened 2017-09-15
fix(provider/kubernetes): Add missing credential props

lwander:
In the de-groovification these two were lost. Will also merge into 1.3 once closed.


lwander:
Broke docker image rendering by removing the registries they belonged to. Broke fiat permissions by not returning any.

PR #1901, opened 2017-09-16
refactor(provider/kubernetes): v2 include api version in resource id

lwander:
Since in many cases these identifiers will be our only way to lookup objects, we need to disambiguate across api versions.


lwander:
spinnaker/spinnaker#1902

PR #1903, opened 2017-09-17
fix(provider/docker): fix client constructor

ethanfrogers:
Bug introduced in `f70d6ea`. I updated the constructor to include
`catalogFile` but didn't update the constructor for those using
`password` and `passwordFile` configurations. This fixes that potential
issue.

@lwander PTAL.


lwander:
Silly groovy w/o proper static errors.

PR #1905, opened 2017-09-18
feat(provider/kubernetes): v2 version deployed manifests as resources

lwander:
What's happening here is that every k8s resource is either "versioned" or "unversioned" which determines  how Spinnaker should create it. If it's versioned (e.g. replica set, configmap, secret, etc...) Spinnaker will get a new versioned name (`my-app-vasdf84` - will be suffixed with `vNNN` in the future) and deploy that. If it's unversioned (e.g. deployment, statefulset, service) it will be created as-is w/o modification to the name, forcing Spinnaker to patch the resource.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,7 @@ allprojects {
   apply plugin: 'groovy'
 
   ext {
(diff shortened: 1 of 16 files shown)


lwander:
spinnaker/spinnaker#1902

lwander:
ping

lwander:
https://github.com/spinnaker/spinnaker/issues/1922

PR #1906, opened 2017-09-19
feat(provider/kubernetes): Add proper support for leveraging source capacity

Z3r0Sum:
Fully addresses: spinnaker/spinnaker#1657

  Also accounts for the possibility that the user might want to set
  their own targetSize while ignoring the source capacity.

  Deck work is still needed to expose "useSourceCapacity" in the UI.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/ops/servergroup/DeployKubernetesAtomicOperation.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/ops/servergroup/DeployKubernetesAtomicOperation.groovy
@@ -72,7 +72,20 @@ class DeployKubernetesAtomicOperation implements AtomicOperation<DeploymentResul
 
     def credentials = description.credentials.credentials
 
-    def namespace = KubernetesUtil.validateNamespace(credentials, description.namespace)
+    /*
+     * Prefer the source namespace when it is available
+     * Fall back on the current namespace and use 'default' if both are not set.
+     */
+    def namespaceToValidate
+    if (description.source?.namespace) {
+      namespaceToValidate = description.source.namespace
+    }
+    else if (description.namespace) {
+      namespaceToValidate = description.namespace
+    }
+    else { namespaceToValidate = "default" }
+
+    def namespace = KubernetesUtil.validateNamespace(credentials, namespaceToValidate)
     description.imagePullSecrets = credentials.imagePullSecrets[namespace]
 
     def serverGroupNameResolver = new KubernetesServerGroupNameResolver(namespace, credentials)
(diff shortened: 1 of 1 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/servergroup/DeployKubernetesAtomicOperation.groovy
```diff
@@ -93,31 +93,40 @@ class DeployKubernetesAtomicOperation implements AtomicOperation<DeploymentResul
...
+    def replicaSet
 
     if (description.source?.useSourceCapacity) {
-      task.updateStatus BASE_PHASE, "Reading ancestor server group ${description.source.serverGroupName}..."
-      def ancestorServerGroup = credentials.apiAdaptor.getReplicationController(description.source.namespace, description.source.serverGroupName)
+      task.updateStatus BASE_PHASE, "Searching for ancestor server group ${description.source.serverGroupName}..."
+      def ancestorServerGroup = credentials.apiAdaptor.getReplicationController(namespace, description.source.serverGroupName)
       if (!ancestorServerGroup) {
-        ancestorServerGroup = credentials.apiAdaptor.getReplicaSet(description.source.namespace, description.source.serverGroupName)
+        ancestorServerGroup = credentials.apiAdaptor.getReplicaSet(namespace, description.source.serverGroupName)
```
ethanfrogers: Is it possible that the source might exist in some other namespace? It seems like that's something that should be supported.

Z3r0Sum: Thanks for pointing this out.

lwander:
Sorry thought I'd submitted my review!

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/servergroup/DeployKubernetesAtomicOperation.groovy
Z3r0Sum: @lwander what about the case where it might not be set you because the Server Group is dynamic?  That was causing issues for me when testing a workflow we have...

lwander: I would default to the current namespace, but rely on source when it's available

Z3r0Sum: Sounds good.

Z3r0Sum:
I added some changes that should accommodate all the concerns above - let me know if you think anything else needs tweaking.  Thanks in advance! ping: @lwander & @ethanfrogers

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/deploy/ops/servergroup/DeployKubernetesAtomicOperation.groovy
```diff
@@ -93,31 +106,40 @@ class DeployKubernetesAtomicOperation implements AtomicOperation<DeploymentResul
...
-        throw new KubernetesResourceNotFoundException("Source server group $description.source.serverGroupName does not exist.")
+        throw new KubernetesResourceNotFoundException("Source Server Group: $description.source.serverGroupName does not exist in Namespace: ${namespace}!")
       }
+      task.updateStatus BASE_PHASE, "Ancestor Server Group Located: ${ancestorServerGroup}"
 
       description.targetSize = ancestorServerGroup.spec?.replicas
+      task.updateStatus BASE_PHASE, "Building replica set..."
+      replicaSet = KubernetesApiConverter.toReplicaSet(new ReplicaSetBuilder(), description, replicaSetName)
+      if (hasDeployment) {
+        replicaSet.spec.replicas = 0
```
lwander: :+1:

PR #1915, opened 2017-09-20
feat(moniker): Adds moniker to all clusters & server groups

lwander:
@andrewbackes fyi


lwander:
spinnaker/spinnaker#1902
spinnaker/spinnaker#1918
spinnaker/spinnaker#1914

lwander:
cmqtgk12ghm

lwander:
It'll be up to individual providers to wean off of the `default` implementation and use the name registry for a provider/account/resource specific strategy to derive a moniker. Examples of this will come in future k8s v2 prs.

PR #1920, opened 2017-09-21
fix(provider/kubernetes): Advanced targetSize use case.

Z3r0Sum:
The elvis operator before was not evaluating truthiness correctly for a value
of 0. A User could and may want to deploy an update ReplicaSet of 0 and Scale
it later. Or perhaps it's bundled in a Spinnaker Pipeline with multiple Server
Groups and one of the Server Groups is scaled to 0. If the User is implementing
'useSourceCapacity' we would want that 0 value respected and not overwritten
with a value of 1.

The proposal is that the desired capacity trumps targetSize.  It's
reasonable to assume that someone who wants a targetSize of 0 would also
have a desired capacity of 0.  However, we shouldn't use the targetSize
if the desired capacity is greater than its value.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/api/KubernetesApiConverter.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/api/KubernetesApiConverter.groovy
@@ -810,7 +810,13 @@ class KubernetesApiConverter {
                                  DeployKubernetesAtomicOperationDescription description,
                                  String replicaSetName) {
 
-    def targetSize = description.targetSize ?: description.capacity?.desired
+    def targetSize
+    if (description.targetSize == 0) {
+      targetSize = description.targetSize
+    }
+    else {
+      targetSize = description.targetSize ?: description.capacity?.desired
+    }
 
     return serverGroupBuilder.withNewMetadata()
       .withName(replicaSetName)
@@ -830,8 +836,15 @@ class KubernetesApiConverter {
   static DeploymentFluentImpl toDeployment(DeploymentFluentImpl serverGroupBuilder,
                                         DeployKubernetesAtomicOperationDescription description,
                                         String replicaSetName) {
+
     def parsedName = Names.parseName(replicaSetName)
-    def targetSize = description.targetSize ?: description.capacity?.desired
+    def targetSize
+    if (description.targetSize == 0) {
+      targetSize = description.targetSize
+    }
+    else {
+      targetSize = description.targetSize ?: description.capacity?.desired
+    }
 
     def builder = serverGroupBuilder.withNewMetadata()
       .withName(parsedName.cluster)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/api/KubernetesApiConverter.groovy
```diff
@@ -809,8 +809,20 @@ class KubernetesApiConverter {
...
+    /*
+    The elvis operator before was not evaluating truthiness correctly for a value of 0.
+    A User could and may want to deploy an update ReplicaSet of 0 and Scale it later.  Or perhaps it's
+    bundled in a Spinnaker Pipeline with multiple Server Groups and one of the Server Groups is scaled to 0.
+    If the User is implementing 'useSourceCapacity' we would want that 0 value respected and not overwritten with a
+    value of 1, which will happen if the targetSize is represented by a null value at this juncture in the code.
+
+    Validation around targetSize already happens at the Create Server Group level.  This is just properly casting the
+    description.targetSize.
+     */
```
ethanfrogers: I think this comment should just be part of the commit message. It's really only relevant with comparing the diff. You could just briefly mention the intentions of the 3 lines here.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/api/KubernetesApiConverter.groovy
```diff
@@ -830,8 +842,21 @@ class KubernetesApiConverter {
...
```
ethanfrogers: Same here.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/api/KubernetesApiConverter.groovy
```diff
@@ -809,8 +809,20 @@ class KubernetesApiConverter {
...
+    Integer targetSize = description.targetSize.toInteger()
+    if (targetSize < 0) {
```
ethanfrogers: What happens if `targetSize` is null? Does `description.targetSize.toInteger()` return `-1` in that case? Just want to make sure we preserve the original behavior as well.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/api/KubernetesApiConverter.groovy
```diff
@@ -830,8 +842,21 @@ class KubernetesApiConverter {
...
+    If the User is implementing 'useSourceCapacity' we would want that 0 value respected and not overwritten with a
+    value of 1, which will happen if the targetSize is represented by a null value at this juncture in the code.
+
+    Validation around targetSize already happens at the Create Server Group level.  This is just properly casting the
+    description.targetSize.
+     */
+    Integer targetSize = description.targetSize.toInteger()
     def parsedName = Names.parseName(replicaSetName)
-    def targetSize = description.targetSize ?: description.capacity?.desired
+    if (targetSize < 0) {
```
ethanfrogers: Same as above.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/api/KubernetesApiConverter.groovy
Z3r0Sum: targetSize cannot be null - that's part of the Server Group Validation.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/api/KubernetesApiConverter.groovy
```diff
@@ -809,8 +809,17 @@ class KubernetesApiConverter {
...
-    def targetSize = description.targetSize ?: description.capacity?.desired
+    /*
+    The elvis operator before was not evaluating truthiness correctly for a value of 0.
+    A User could and may want to deploy an update ReplicaSet of 0 and Scale it later.  Or perhaps it's
+    bundled in a Spinnaker Pipeline with multiple Server Groups and one of the Server Groups is scaled to 0.
+    If the User is implementing 'useSourceCapacity' we would want that 0 value respected and not overwritten with a
+    value of 1, which will happen if the targetSize is represented by a null value at this juncture in the code.
+     */
+    Integer targetSize = description.targetSize.toInteger()
+    if (targetSize < 0) {
```
ethanfrogers: Since `targetSize` can never be `null` or `< 0`, I don't think `description.capacity` will ever be considered.

Z3r0Sum: Yeah, I'll need to reason about on this... this will need to change.

Z3r0Sum: I'm thinking the right way to approach this would be the following:

Defer to the desired source capacity, so if the targetSize is less than it - the desired source capacity wins.  This still covers the use case of this PR as you wouldn't set a desired source capacity of > 0 if you wanted a targetSize of 0 or you wanted the resource at 0.  Essentially, desired source capacity always trumps targetSize.  Thoughts?

Z3r0Sum:
Scenarios Tested:

`useSourceCapacity: false`

- [ ]  targetSize < desired capacity --- desired capacity value is used
- [ ] targetSize > desired capacity -- desired capacity value is still used (it should be brought in line with targetSize) -- I'm not sure what the correct philosophical debate is here, but maybe we should set the desired capacity equal to the targetSize if the targetSize is greater than the desired capacity.

`useSourceCapacity: true`

- [ ] targetSize < desired capacity --- Does not matter, original Replicas Count is respected.
- [ ] targetSize > desired capacity --- Does not matter, original Replicas Count is respected.

Z3r0Sum:
@lwander I'd appreciate you weighing in on this as you probably have some opinions on what should get precedence...

Z3r0Sum:
It seems like the desired source capacity is always null and just sets the targetSize, is this assumption correct?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/api/KubernetesApiConverter.groovy
```diff
@@ -810,7 +810,10 @@ class KubernetesApiConverter {
                                  DeployKubernetesAtomicOperationDescription description,
                                  String replicaSetName) {
 
-    def targetSize = description.targetSize ?: description.capacity?.desired
+    def targetSize
+    if (description.targetSize == 0) {
+      targetSize = description.targetSize
+    } else { targetSize = description.targetSize ?: description.capacity?.desired }
```
lwander: Could you put this onto a newline?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/api/KubernetesApiConverter.groovy
```diff
@@ -830,8 +833,12 @@ class KubernetesApiConverter {
   static DeploymentFluentImpl toDeployment(DeploymentFluentImpl serverGroupBuilder,
                                         DeployKubernetesAtomicOperationDescription description,
                                         String replicaSetName) {
+
     def parsedName = Names.parseName(replicaSetName)
-    def targetSize = description.targetSize ?: description.capacity?.desired
+    def targetSize
+    if (description.targetSize == 0) {
+      targetSize = description.targetSize
+    } else { targetSize = description.targetSize ?: description.capacity?.desired }
```
lwander: Same here.

PR #1922, opened 2017-09-25
fix(rollbacks): support for tolerating some instance failures

asher:
[AWS] Allow rollbacks to proceed when < 100% of instances in the rollback asg report healthy.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AbstractEnableDisableAtomicOperation.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AbstractEnableDisableAtomicOperation.groovy
@@ -150,6 +150,7 @@ abstract class AbstractEnableDisableAtomicOperation implements AtomicOperation<V
         task.updateStatus phaseName, "Only disabling instances $instanceIds on ASG $serverGroupName with percentage ${description.desiredPercentage}"
       }
 
+      // ELB/ALB registration
       if (disable) {
         changeRegistrationOfInstancesWithLoadBalancer(asg.loadBalancerNames, instanceIds) { String loadBalancerName, List<Instance> instances ->
           try {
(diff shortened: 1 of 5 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AbstractEnableDisableAtomicOperation.groovy
```diff
@@ -186,8 +188,10 @@ abstract class AbstractEnableDisableAtomicOperation implements AtomicOperation<V
             credentials: credentials,
             region: region,
             asgName: serverGroupName,
-            instanceIds: instanceIds
+            instanceIds: instanceIds,
+            targetHealthyDeployPercentage: description.targetHealthyDeployPercentage ?: 100
         )
+        // @TODO pass targetHealthy% here
```
cfieber: what was the TODO here?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AbstractEnableDisableAtomicOperation.groovy
```diff
@@ -186,8 +188,10 @@ abstract class AbstractEnableDisableAtomicOperation implements AtomicOperation<V
             credentials: credentials,
             region: region,
             asgName: serverGroupName,
-            instanceIds: instanceIds
+            instanceIds: instanceIds,
+            targetHealthyDeployPercentage: description.targetHealthyDeployPercentage ?: 100
```
cfieber: I think (not totally sure) that the groovy truthiness here would convert an explicitly supplied value of `0` to `100`

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AbstractEnableDisableAtomicOperation.groovy
asher: It was resolved by adding targetHealthyDeployPercentage to EnableDisableAsgDescription && AbstractRegionAsgInstanceIdsDescription, removing the TODO.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AbstractEnableDisableAtomicOperation.groovy
asher: Verified that you're right - going to change this to the more explicit `description.targetHealthyDeployPercentage != null ? description.targetHealthyDeployPercentage : 100`

PR #1924, opened 2017-09-25
fix(aws) : task stuck on wait for up instances

njfix6:
We are having trouble with tasks "waiting for up instances" in aws.
I think it is because this part of the code is getting throttled,
but im not possitive of the fix. This code it meant to catch the
throttling error, and stop the current task that is ruuning,
so the user can restart the task if possible.


njfix6:
I've added a test for this now.

tomaslin:
This PR makes no sense to me. There is no task associated with the caching agent. I don't understand what this is doing at all.

tomaslin:
Caching agents run periodically to get the state of the cloud. WaitForUpInstances is an orca task that polls the state of clouddriver. There is no clouddriver task associated with this polling operation, so this solution is not going to work at all.

njfix6:
@tomaslin Sorry for the confusion, the task is probably/definitely not the right way to handle this. The source of this problem is that when `amazonClientProvider.getAmazonElasticLoadBalancing(account, region)` is called too many times, it get throttled by AWS, causing the state to get stuck. Is there a different way we can handle this throttling so it could catch the error, or possibly even retry?

PR #1925, opened 2017-09-25
feat(provider/kubernetes): Build v2 creds based on context & kubeconf…

lwander:
…igFile


lwander:
spinnaker/spinnaker#1902

PR #1927, opened 2017-09-25
feat(provider/kubernetes): v2 cache network policies

lwander:


lwander:
spinnaker/spinnaker#1902
spinnaker/spinnaker#1919

PR #1933, opened 2017-09-26
fix(provider/kubernetes): fix k8s client configuration and image id p…

chlung:
About this PR,

Fix the problem when context does not present in k8s configuration file and incorrect image id assignment.

…roblem.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesApiClientConfig.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesApiClientConfig.groovy
@@ -55,17 +55,7 @@ public class KubernetesApiClientConfig extends Config {
     } catch (FileNotFoundException e) {
       throw new RuntimeException("Unable to create credentials from kubeconfig file: " + e, e)
     } catch (Exception e2) {
-    }
-
-    InputStream is = new FileInputStream(kubeconfigFile)
-    Reader input = new InputStreamReader(is)
-    Yaml yaml = new Yaml(new SafeConstructor())
-    Object config = yaml.load(input)
-    Map<String, Object> configMap = (Map<String, Object>)config
-
-    //TODO: Need to validate cluster and user when client library exposes these api.
-    if (StringUtils.isEmpty(context) && !configMap.get("current-context")) {
-      throw new RuntimeException("Missing required field ${context} in kubeconfig file and clouddriver configuration.")
+      throw new RuntimeException("Missing required field(s) in kubeconfig file or/and clouddriver configuration.")
     }
 
     if (!StringUtils.isEmpty(context)) {
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesApiClientConfig.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesApiClientConfig.groovy
@@ -44,7 +44,6 @@ public class KubernetesApiClientConfig extends Config {
   }
 
   public ApiClient getApiCient() throws Exception {
-    def kubeConfig = KubeConfigUtils.parseConfig(new File(kubeconfigFile))
     KubeConfig kubeconfig
 
     try {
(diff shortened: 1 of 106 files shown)


chlung:
@lwander Please review.  Thanks.

ethanfrogers:
@lwander @chlung I wonder if this is really an issue with Spinnaker. It seems like the client library is just assuming that `current-context` exists. If the client libraries across various languages require it, maybe we should say that the provided Kubernetes config is invalid. If it isn't, which it wasn't with the Fabric8 library, the client library should be fixed accordingly.

chlung:
@lwander  and @ethanfrogers.  Without these required fields, the kubectl won't work.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesApiClientConfig.groovy
```diff
@@ -22,18 +22,48 @@ import io.kubernetes.client.util.KubeConfig;
...
+      configMap.put("clusters", cluster)
+    }
+
+    if (this.user) {
+      configMap.put("users", user)
+    }
+
+    yaml = new Yaml(new SafeConstructor());
+    String out = yaml.dump(configMap);
+    InputStream kubeconfigIS = new ByteArrayInputStream(out.getBytes());
```
lwander: What is the goal here? You don't need to recreate a kubeconfig. If a user has specified a context, it's assumed it exists in their kubeconfig, not the one they are creating.

chlung: If the value is presenting is clouddriver-local.yml.  It should be allow to override.  Is it? or we just need to update if the value does not present in the config file.

lwander:
That's a good point @chlung, also this is a regression, so even if it wasn't needed it's still breaking users (like yourself) without warning.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesApiClientConfig.groovy
lwander: Yes it should override `current-context`. That doesn't mean we need to create a new kubeconfig object, though.

chlung:
I can throw some error message if these fields are missing in both config and clouddriver-local.yml.

chlung:
@lwander https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/util/KubeConfig.java#L101 will still throw NPE if "current-context" does not present.  So, we must re-create the kubeconfig with all required data before call or allow call setContext() later if null.  Please review and let me your suggestion.

lwander:
I don't see how this will work - if the kubeconfig you create doesn't have any credentials beyond the name of the context, cluster & user you add how will it authenticate?

chlung:
Then, we might consider to call setContext() later if the initial context is null.

lwander:
I would just follow the pattern in the code I sent you via slack

Here is the relevant section: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java#L127-L146

chlung:
@lwander I patched the changes per our discussion.  Please merge this PR after client library patched and apply to clouddriver-kubernetes; otherwise, the reported issue won't fix.  thanks.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesApiClientConfig.groovy
```diff
@@ -22,18 +22,65 @@ import io.kubernetes.client.util.KubeConfig;
...
+
+    try {
+      if (StringUtils.isEmpty(kubeconfigFile)) {
+        kubeconfig = KubeConfig.loadDefaultKubeConfig()
+      } else {
+        kubeconfig = KubeConfig.loadKubeConfig(new FileReader(kubeconfigFile))
+      }
+    } catch (FileNotFoundException e) {
+      throw new RuntimeException("Unable to create credentials from kubeconfig file: " + e, e)
+    } catch (Exception e2) {
```
lwander: You should probably handle this as well ( you can just throw an exception )

ethanfrogers:
Out of curiosity, what's the patch that needs to be made to the client library?

lwander:
Client library has the same issue where if `current-context` isn't set, it will crash. We could get around it here by loading the kubeconfig by hand & setting it if unset, but that would inconsistent with how kubectl handles it (by creating the default localhost:8080 context).

chlung:
@lwander I updated the code per your feedback.  Please review.  thanks.

chlung:
@ethanfrogers thanks

PR #1935, opened 2017-09-26
feat(appengine): Deploy from Google Cloud Storage [accounts].

ewiseblatt:
@danielpeach please review
@duftler FYI

This works with the existing citest for gcs buckets (without passing a storage account).
I have another change for the test which does pass a storage account in with the deployment request.

I havent yet tested the tar file thing, but did test the extraction of tar files from buckets so believe it will work.

My only technical concern is that I do not preserve any file properties other than timestamps from GCS (e.g. mode bits, owner/group, etc). I dont think this matters for appengine, and that's the only scope at this time.

Note also that I load credentials on demand rather than on startup like other places seem to. It seems simpler this way. I dont think performance is a concenr, but for better or worse it means that a misconfiguration will be caught during request processing rather than startup (but on the positive side, failure will be scoped only to individual accounts rather than the whole server).

### DIFF
--- a/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/description/DeployAppengineDescription.groovy
+++ b/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/description/DeployAppengineDescription.groovy
@@ -29,6 +29,7 @@ class DeployAppengineDescription extends AbstractAppengineCredentialsDescription
   String stack
   String freeFormDetails
   String repositoryUrl
+  String storageAccountName  // for GCS repositories only
   AppengineGitCredentialType gitCredentialType
   String branch
   String applicationDirectoryRoot
--- a/clouddriver-appengine/clouddriver-appengine.gradle
+++ b/clouddriver-appengine/clouddriver-appengine.gradle
@@ -8,4 +8,7 @@ dependencies {
   // TODO(dpeach): move to spinnaker/spinnaker-dependencies.
   compile "com.google.apis:google-api-services-appengine:v1-rev4-1.22.0"
   compile "org.eclipse.jgit:org.eclipse.jgit:4.5.0.201609210915-r"
+
+  compile spinnaker.dependency("googleStorage")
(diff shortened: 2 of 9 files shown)


clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/config/AppengineConfigurationProperties.groovy
```diff
@@ -35,6 +35,7 @@ class AppengineConfigurationProperties {
 
     String serviceAccountEmail
     String localRepositoryDirectory
+    String gcsAccount
```
danielpeach: Is this used anywhere?

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/description/DeployAppengineDescription.groovy
```diff
@@ -29,6 +29,7 @@ class DeployAppengineDescription extends AbstractAppengineCredentialsDescription
   String stack
   String freeFormDetails
   String repositoryUrl
+  String storageAccountName  // for GCS repositories only
```
danielpeach: How does a Clouddriver client know of a `storageAccountName`?

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/config/AppengineConfigurationProperties.groovy
ewiseblatt: No. I meant to make it the default account but am taking it out entirely for now.

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/description/DeployAppengineDescription.groovy
ewiseblatt: Good point. I'll add a method to the appengine controller in a followup PR.

PR #1940, opened 2017-09-27
feat(moniker): Adds moniker to pendingOnDemandResults

lwander:
See https://github.com/spinnaker/clouddriver/compare/master...lwander:moniker-on-demand?expand=1#diff-793a249c0692198d2557378d889e7285 for the "correct" long-term implementation, the "deriveMoniker" is more of a patch to get this working without proper support in each cloud provider.

@andrewbackes

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/OnDemandAgent.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/OnDemandAgent.java
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.clouddriver.cache;
 
 import com.netflix.spinnaker.cats.agent.CacheResult;
 import com.netflix.spinnaker.cats.provider.ProviderCache;
+import com.netflix.spinnaker.moniker.Moniker;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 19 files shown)


lwander:
@cfieber FYI

lwander:
spinnaker/spinnaker#1902
spinnaker/spinnaker#1914
spinnaker/spinnaker#1969

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/OnDemandAgent.java
```diff
@@ -18,13 +18,17 @@
...
+import com.netflix.spinnaker.moniker.Moniker;
+import lombok.extern.slf4j.Slf4j;
 
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 
+import static jdk.nashorn.internal.runtime.regexp.joni.Config.log;
```
duftler: Long trip for a `log`.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/OnDemandAgent.java
```diff
@@ -65,6 +69,28 @@ public OnDemandResult(String sourceAgentType, CacheResult cacheResult, Map<Strin
     }
   }
 
+  /*
+   * WARNING: this is an interim solution while cloud providers write their own ways to derive monikers.
```
duftler: Worth logging an actual warning?

lwander: I think so, when we migrate each cloud provider off. Before then it will just generate noise.

PR #1941, opened 2017-09-28
feat(moniker): Add monikers to security groups & load balancers

lwander:


lwander:
spinnaker/spinnaker#1902
spinnaker/spinnaker#1914
spinnaker/spinnaker#1969

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/securitygroups/SecurityGroupRule.groovy
```diff
@@ -17,9 +17,7 @@
 package com.netflix.spinnaker.clouddriver.model.securitygroups
 
 import com.netflix.spinnaker.clouddriver.model.SecurityGroup
-import groovy.transform.Immutable
 
-@Immutable
```
lwander: Dropped because this wasn't being enforced. It was falsely assuming that `SecurityGroup` was immutable which wasn't made apparent to the runtime until `SecurityGroup` was converted to Java.

PR #1948, opened 2017-09-29
fix(provider/gce): Paginate server groups when calculating next name.

duftler:


edwinavalos:
screen shot 2017-09-29 at 6 10 56 pm

Looks good on our end, my debug code finds all the server groups and takenSlots as expected as well.

PR #1951, opened 2017-10-01
feat(search): cull missing ASGs before returning

icfantv:
* removes ASGs that show up in the search results from the redis cache
but don't actually exist.


icfantv:
@cfieber or @robzienert PTAL.  primarily concerned about whether or not there are performance ramifications.

PR #1952, opened 2017-10-02
feat(search): allow bulk retrieval of ASGs

icfantv:
* add ability to bulk retrieve ASGs from a list of comma-delimited
strings of pipe-delimited basic ASG info: application, account, region,
and ASG name.

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
@@ -28,17 +28,20 @@ import com.netflix.spinnaker.clouddriver.model.view.ServerGroupViewModelPostProc
 import com.netflix.spinnaker.clouddriver.requestqueue.RequestQueue
 import com.netflix.spinnaker.kork.web.exceptions.NotFoundException
 import com.netflix.spinnaker.moniker.Moniker
+import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.context.MessageSource
 import org.springframework.security.access.prepost.PostAuthorize
 import org.springframework.security.access.prepost.PostFilter
 import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.web.bind.annotation.PathVariable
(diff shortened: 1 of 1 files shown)


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -76,6 +79,30 @@ class ServerGroupController {
     serverGroup
   }
 
+  @RequestMapping(value = "/serverGroupsDetailsForList", method = RequestMethod.POST)
```
icfantv: I don't love the name or that I have to use a POST, but I really don't want to have to batch tiny GET requests from deck since, e.g., when we do INT stack searches, we could be passing 100+ strings of decent length.  Open to better suggestions here.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -76,6 +79,30 @@ class ServerGroupController {
...
+  List<ServerGroup> getServerGroupsDetailsForList(@RequestBody String serverGroupList) {
+
+    List<ServerGroup> results = [];
+    serverGroupList.split(',').collect {
```
icfantv: I feel like there's a better way to create the map from the pipe-delimited values, but I played around with `inject` far too long and wasn't terribly successful.

icfantv:
@cfieber or @robzienert PTAL.  Open to suggestions on this.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -76,6 +79,30 @@ class ServerGroupController {
...
+      List<String> items = it.split("\\|")
+      return ["application": items[0], "account": items[1], "region": items[2], "name": items[3]]
+    }.every {
+        def map = it
+        def matches = (Set<ServerGroup>) clusterProviders.findResults {
```
icfantv: This will likely need Fiat - if so, I'll get with google.

PR #1957, opened 2017-10-03
feat(appengine): Add AppengineStorageController to list accounts

ewiseblatt:
@danielpeach

### DIFF
--- a/clouddriver-appengine/src/main/java/com/netflix/spinnaker/clouddriver/appengine/storage/controllers/AppengineStorageController.java
+++ b/clouddriver-appengine/src/main/java/com/netflix/spinnaker/clouddriver/appengine/storage/controllers/AppengineStorageController.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


clouddriver-appengine/src/main/java/com/netflix/spinnaker/clouddriver/appengine/storage/controllers/AppengineStorageController.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RestController;
+
+import javax.servlet.http.HttpServletRequest;
+import java.util.ArrayList;
+import java.util.List;
+
+@Slf4j
+@RestController
+@RequestMapping("/appengine/storage")
```
lwander: Will this be moved out of the appengine namespace?

ewiseblatt:
Eventually, but not initially. I dont have a handle on it.
There are "DataProviders' in clouddriver. Only S3 is supported. Seems like
it shouldbe related but currently isnt. So keeping this inside appengine
for now.

lwander:
After talking to @ajordens the idea was to replace the data providers with these storage accounts once we're ready

ewiseblatt:
I dont quite understand the data provider so will need to discuss it more
with Tomas. I had an initial exchange with him a couple weeks ago.

ewiseblatt:
FWIW, the issue isnt so much how it works, rather than relating to the
intent and philosophy behind it so I can approach it in some expected way
(or know that this makes sense). Especially to match the intuition of how
usage is expected to evolve.

PR #1962, opened 2017-10-04
chore(provider/kubernetes): Bump client to 1.0.0-beta1

lwander:


lwander:
@skim1420 1.4.3 should be ready now.

PR #1967, opened 2017-10-05
feat(moniker): Add getServerGroup endpoint that does not require application name.

andrewbackes:
In order for Orca to generate Clouddriver operations it needs to pull information about server groups. Since we no longer want to deduce the application name from the serverGroupName using frigga, one of two changes will need to be made:

1. Add an application name parameter to all of the methods in the oortHelper and oortService in Orca. Then follow all of those method calls and pass application name as function parameters all the way up stream to the stage data where we should find a moniker. This would result in hundreds of modifications.

2. Add an endpoint in Clouddriver that only requires serverGroupName, account, and region.

Now that we have the moniker as part of the server group in Clouddriver, option (2) seemed much more feasible.


andrewbackes:
I have other threads of work to work on until @ttomsu gets back. I am going to assume this is an acceptable approach and move on for the time being.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -59,12 +59,27 @@ class ServerGroupController {
 
   @PreAuthorize("hasPermission(#application, 'APPLICATION', 'READ') and hasPermission(#account, 'ACCOUNT', 'READ')")
   @RequestMapping(value = "/applications/{application}/serverGroups/{account}/{region}/{name:.+}", method = RequestMethod.GET)
-  ServerGroup getServerGroup(@PathVariable String application, // needed for @PreAuthorize
+  ServerGroup getServerGroupByApplication(@PathVariable String application, // needed for @PreAuthorize
```
ttomsu: Param formatting is off here.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -59,12 +59,27 @@ class ServerGroupController {
...
+  @PreAuthorize("hasPermission(#account, 'ACCOUNT', 'READ')")
+  @PostAuthorize("hasPermission(returnObject?.moniker?.app, 'APPLICATION', 'READ')")
+  @RequestMapping(value = "/serverGroups/{account}/{region}/{name:.+}", method = RequestMethod.GET)
+  ServerGroup getServerGroupByName(@PathVariable String account,
+                                   @PathVariable String region,
+                                   @PathVariable String name) {
+    getServerGroup(account, region, name)
+  }
+
+  private getServerGroup(String account,
```
ttomsu: Formatting here too.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -59,12 +59,27 @@ class ServerGroupController {
...
                              @PathVariable String account,
                              @PathVariable String region,
                              @PathVariable String name) {
+    getServerGroup(account, region, name)
+  }
+
+  @PreAuthorize("hasPermission(#account, 'ACCOUNT', 'READ')")
+  @PostAuthorize("hasPermission(returnObject?.moniker?.app, 'APPLICATION', 'READ')")
+  @RequestMapping(value = "/serverGroups/{account}/{region}/{name:.+}", method = RequestMethod.GET)
+  ServerGroup getServerGroupByName(@PathVariable String account,
```
ttomsu: The distinction in method names between `getServerGroupByApplication` and `getServerGroupByName` is not clear to me. Perhaps `getServerGroupByMoniker`?

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -59,12 +59,27 @@ class ServerGroupController {
...
+  ServerGroup getServerGroupByApplication(@PathVariable String application, // needed for @PreAuthorize
                              @PathVariable String account,
                              @PathVariable String region,
                              @PathVariable String name) {
+    getServerGroup(account, region, name)
+  }
+
+  @PreAuthorize("hasPermission(#account, 'ACCOUNT', 'READ')")
+  @PostAuthorize("hasPermission(returnObject?.moniker?.app, 'APPLICATION', 'READ')")
+  @RequestMapping(value = "/serverGroups/{account}/{region}/{name:.+}", method = RequestMethod.GET)
```
ttomsu: The mixing of `/serverGroups` and `/applications/...` endpoints is starting to have a bad smell. Can you file an issue to have this cleaned up and reference it in a TODO here?

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
andrewbackes: Good catch. Fixed.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
andrewbackes: - Created issue here: https://github.com/spinnaker/spinnaker/issues/2023
- Added a comment here: https://github.com/spinnaker/clouddriver/pull/1967/files#diff-d8e4115f4785725b1e418532c640c617R62

PR #1970, opened 2017-10-06
fix(provider/gae): Set path correct when application root non-empty.

jtk54:


clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/gcsClient/AppengineGcsRepositoryClient.groovy
```diff
@@ -50,7 +50,7 @@ class AppengineGcsRepositoryClient implements AppengineRepositoryClient {
 
     def fullPath = repositoryUrl.substring(gsPrefix.length())
     if (applicationDirectoryRoot) {
-      fullPath + '/' + applicationDirectoryRoot
+      fullPath = fullPath + '/' + applicationDirectoryRoot
```
duftler: Could be:
`fullPath += '/' + applicationDirectoryRoot`

or:
`fullPath += "/$applicationDirectoryRoot"`

PR #1971, opened 2017-10-06
feat(provider/kuberentes): Application provider

lwander:


lwander:
spinnaker/spinnaker#1902
spinnaker/spinnaker#1995

PR #1972, opened 2017-10-06
fix(provider/aws): fix issue if VPCZoneIdentifier contains deleted subnet

asher:


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ClusterCachingAgent.groovy
```diff
@@ -686,9 +686,9 @@ class ClusterCachingAgent implements CachingAgent, OnDemandAgent, AccountAware,
       String vpcId = null
       if (asg.getVPCZoneIdentifier()) {
         String[] subnets = asg.getVPCZoneIdentifier().split(',')
-        Set<String> vpcIds = subnets.collect { subnetMap[it] }
+        Set<String> vpcIds = subnets.collect { subnetMap[it] }.findAll { it != null }
```
cfieber: could replace this with `subnets.findResults { subnetMap[it] }`

PR #1985, opened 2017-10-11
fix(core): Support the eviction of stale cache key identifiers

ajordens:
This PR attempts to clean up situations where the `:members` set for a
given cached type contains keys that no longer exist.

In particular, a `CachingAgent` can optionally return a pattern glob for
each authoritative type.

Any `:members` key matching the pattern glob BUT not included in the
most recent caching cycle will be evicted.

### DIFF
--- a/cats/cats-core/src/test/groovy/com/netflix/spinnaker/cats/agent/CacheExecutionSpec.groovy
+++ b/cats/cats-core/src/test/groovy/com/netflix/spinnaker/cats/agent/CacheExecutionSpec.groovy
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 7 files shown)


cats/cats-core/src/test/groovy/com/netflix/spinnaker/cats/agent/CacheExecutionSpec.groovy
```diff
@@ -0,0 +1,63 @@
...
+import spock.lang.Specification
+
+import static com.netflix.spinnaker.cats.agent.AgentDataType.Authority.AUTHORITATIVE;
+
+class CacheExecutionSpec extends Specification {
+  def providerRegistry = Mock(ProviderRegistry)
+  def cachingAgent = Mock(CachingAgent)
+  def providerCache = Mock(ProviderCache)
+
+  void "should evict keys that were NOT cached by the responsible agent"() {
```
ajordens: Few more tests are coming.

ajordens:
Thoughts @cfieber?

I've jammed keys into various `:members` that satisfy an agents key pattern and verified that they get evicted.

Tried to come up with something that wouldn't require modifications to individual caching agents but came up short. Figured this was a minimal change and agents could opt themselves in.

ajordens:
Memory is fuzzy around relationship sets and whether we also have the occasional stale key problem there as well.

cfieber:
This won't affect the clusters members set, we don't currently have anything that is set as AUTHORITATIVE on cluster 

We could theoretically make a cluster indexing agent that just crawls all server groups in cache to build the authoritative clusters list ..

lwander:
Just wondering (I'm in the same boat), do groups with nothing set as AUTHORITATIVE ever have entries evicted?

ajordens:
@lwander Nothing specific in this PR about cleaning up sets w/o an AUTHORITATIVE owner. 

`cluster` is an interesting one that might warrant some attention, I don't recall offhand if we've had issues with stale data there lately.

PR #1990, opened 2017-10-12
feat(provider/gce): Replace instanceTemplate get with list call.

jtk54:
Do one upfront list call per server group caching cycle for instance templates as opposed to a get call per server group.


lwander:
Neat

duftler:
Very nice.

PR #1994, opened 2017-10-12
fix(aws): Enable stale key evictions for amazon load balancers

ajordens:
Key structures vary between classic and application/networking load
balancers.


ajordens:
FYI @cfieber

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonLoadBalancerCachingAgent.groovy
```diff
@@ -48,6 +48,13 @@ class AmazonLoadBalancerCachingAgent extends AbstractAmazonLoadBalancerCachingAg
   }
 
   @Override
+  Optional<Map<String, String>> getCacheKeyPatterns() {
+    return [
+      (LOAD_BALANCERS.ns): Keys.getLoadBalancerKey('*', account.name, region, 'vpc-????????', null)
```
ajordens: key structure differs depending on whether or not a load balancer type is present so need to fix the vpc length vs '*' globing.

classic load balancers do not include a type in their keys

PR #1997, opened 2017-10-13
refactor(aws): Simplified ELB security group autocreation

jeyrschabu:
- updated code that allows ingress from elb group to app security group

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/UpsertAmazonLoadBalancerAtomicOperation.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/UpsertAmazonLoadBalancerAtomicOperation.groovy
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.clouddriver.aws.deploy.ops.loadbalancer
 
 import com.amazonaws.AmazonServiceException
 import com.amazonaws.services.ec2.model.IpPermission
+import com.amazonaws.services.ec2.model.SecurityGroup
 import com.amazonaws.services.ec2.model.UserIdGroupPair
 import com.amazonaws.services.elasticloadbalancing.model.ConfigureHealthCheckRequest
 import com.amazonaws.services.elasticloadbalancing.model.CrossZoneLoadBalancing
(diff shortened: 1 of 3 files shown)


jeyrschabu:
@ajordens PTAL (cc @cfieber)

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/UpsertAmazonLoadBalancerAtomicOperation.groovy
```diff
@@ -192,62 +191,88 @@ class UpsertAmazonLoadBalancerAtomicOperation implements AtomicOperation<UpsertA
...
+    }
 
-    IngressSecurityGroupDescription applicationLoadBalancerSecurityGroupDescription = new IngressSecurityGroupDescription(
-      applicationLoadBalancerSecurityGroupId,
-      applicationLoadBalancerSecurityGroupName,
-      description.credentialAccount,
-      description.vpcId
-    )
+    // include health check endpoint
+    Matcher portMatcher = HEALTHCHECK_PORT_PATTERN.matcher(description.healthCheck)
```
ajordens: Is `description.healthCheck` the right attribute to check?

```
        "healthCheck": "HTTP:7001/healthcheck",
        "healthCheckProtocol": "HTTP",
        "healthCheckPort": 7001,
```

Could we just check `healthCheckPort`? Alternatively if we're checking `.healthCheck`, is there any implication if the value was `HTTP:7001/v2/healthcheck`?

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/UpsertAmazonLoadBalancerClassicAtomicOperationSpec.groovy
```diff
@@ -86,21 +86,57 @@ class UpsertAmazonLoadBalancerClassicAtomicOperationSpec extends Specification {
   def regionScopedProviderFactory = Stub(RegionScopedProviderFactory) {
     forRegion(_, "us-east-1") >> regionScopedProvider
   }
+
+  def securityGroupLookup = Mock(SecurityGroupLookup)
```
ajordens: Out of curiosity, is there a test when the application security group does not exist? Looking at the code we'll create it right?

ajordens:
Couple minor comments but the fix to how we're determining application names looks sensible to me (caveat Frigga vs. Moniker).

@lwander Any moniker-related concerns here? Not sure what the overarching goal is around people potentially creating load balancers that don't fit the existing naming patterns (fair bit of fallout if-so I suspect).

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/UpsertAmazonLoadBalancerAtomicOperation.groovy
jeyrschabu: For simplicity, I will use `healthCheckPort` here

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/UpsertAmazonLoadBalancerClassicAtomicOperationSpec.groovy
jeyrschabu: Yes. we will create it. I will add a test for that as well

jeyrschabu:
Updated

PR #1999, opened 2017-10-14
feat(provider/kubernetes): v2 delete operation

lwander:


lwander:
spinnaker/spinnaker#1902

PR #2017, opened 2017-10-18
fix(provider/kubernetes): allows dots in secretNames

polarbizzle:
We have some secrets in our cluster that have dots in the name and have deployed manually using them as volume sources. The next logical step was to set up a spinnaker pipeline but doing so resulted in:

    deployKubernetesAtomicOperationDescription.volumeSources[1].secret.secretName.invalid (Must match ^[a-z0-9]+([-a-z0-9]*[a-z0-9])?$)

Upon further investigation, the kubernetes api allows dots in secret names:
https://github.com/kubernetes/kubernetes/blob/e5ac41331119bb8c67004b8efaa0c3d942b31d3a/pkg/api/validation/validation.go#L160-L162
https://github.com/kubernetes/kubernetes/blob/8d7d7a5e0d4d7e75f5a860574346944b8cc0fc43/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L126-L142
https://github.com/kubernetes/kubernetes/blob/e5ac41331119bb8c67004b8efaa0c3d942b31d3a/pkg/api/validation/validation.go#L1488

It seems that spinnaker should line up with the kubernetes api in the validation it does on secret names, so I brewed up this code.


polarbizzle:
Rebased 🌈 Thanks for having a look!

PR #2019, opened 2017-10-19
perf(cluster): Avoid prematurely fetching full server groups

ajordens:
This is an optimization for fetching target server groups by `CURRENT`,
`PREVIOUS`, `OLDEST` and `FAIL`.

For these strategies, the unexpanded server group is sufficient to
narrow down to a match.

Once a match is found, _only_ that server group will be expanded.

This only works for cluster providers that support loading both
expanded and unexpanded server groups (ie. aws). The behavior for other
providers should be unchanged.


ajordens:
FYI @spinnaker/google-reviewers  / @duftler, I don't expect this to affect any cloudprovider other than `aws` in the OSS release.

ajordens:
I compared outputs from both old / new on a large cluster to observe both (a) big speed increase and (b) no difference in responses

PR #2023, opened 2017-10-20
fix(provider/gce): Fix broken deploys with namedPorts.

jtk54:


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -1072,6 +1072,15 @@ class GCEUtil {
     )
   }
 
+  static void updateMetadatWithLoadBalancingPolicy(GoogleHttpLoadBalancingPolicy policy, Map instanceMetadata, ObjectMapper objectMapper) {
```
danielpeach: Missing an 'a' ('updateMetadata').

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -425,7 +415,13 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
         }
         namedPorts = sourceServerGroup?.namedPorts?.collect { name, port -> new NamedPort(name: name, port: port) }
       } else {
-        namedPorts = description?.loadBalancingPolicy?.namedPorts
+        def loadBalancingPolicy = description?.loadBalancingPolicy
+        if (loadBalancingPolicy?.namedPorts)  {
```
danielpeach: Right here I'm worried you could get into a weird state if `namedPorts` is an empty list and `listeningPort` is still defined on the loadBalancingPolicy.

I would do an explicit `if (loadBalancingPolicy?.namedPorts != null) { ... }`.

danielpeach: If someone sets `namedPorts` to an empty list (that would be strange), and they haven't removed `loadBalancingPolicy.listeningPort` from their deploy description, they'll hit the `else if` below, won't they? That would be unexpected, though maybe unlikely. I would do a strict null check here anyways.

PR #2025, opened 2017-10-20
fix(aws): Only autocreate app elb security group on Create

jeyrschabu:
- Updated to only auto create <appname>-elb group for new ELBs


jeyrschabu:
@emjburns PTAL

PR #2027, opened 2017-10-24
fix(provider/gce): Paginate instance template list in svg caching.

jtk54:


jtk54:
Refactored into one static method.

PR #2029, opened 2017-10-25
[FIX]kubernetes collectJob status  ClassCastException

binjianwu:
I have operation spinnaker with kubernetes provider,but there throw expection:
```
Exception ( Wait On Job Completion )
Cannot cast object 'com.netflix.spinnaker.clouddriver.kubernetes.v1.security.KubernetesV1Credentials@45ff365d' 
with class 'com.netflix.spinnaker.clouddriver.kubernetes.v1.security.KubernetesV1Credentials' 
to class 'com.netflix.spinnaker.clouddriver.kubernetes.security.KubernetesNamedAccountCredentials'
```
then , I found the question on the code at `com.netflix.spinnaker.clouddriver.kubernetes.v1.provider.view.KubernetesJobProvider` . In this `collectJob()` method ,the `if (!credentials instanceof KubernetesNamedAccountCredentials)` always is false.  and `accountCredentialsProvider.getCredentials(account).credentials` is can not cast to `KubernetesNamedAccountCredentials ` .
So, I made a little change here。

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/view/KubernetesJobProvider.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/view/KubernetesJobProvider.groovy
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.clouddriver.kubernetes.v1.provider.view
 
 import com.netflix.spinnaker.clouddriver.kubernetes.v1.model.KubernetesJobStatus
 import com.netflix.spinnaker.clouddriver.kubernetes.security.KubernetesNamedAccountCredentials
+import com.netflix.spinnaker.clouddriver.kubernetes.v1.security.KubernetesV1Credentials
 import com.netflix.spinnaker.clouddriver.model.JobProvider
 import com.netflix.spinnaker.clouddriver.model.JobState
 import com.netflix.spinnaker.clouddriver.security.AccountCredentialsProvider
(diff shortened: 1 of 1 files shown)


carllhw:
@danielpeach I have the same problem.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/view/KubernetesJobProvider.groovy
```diff
@@ -34,11 +34,11 @@ class KubernetesJobProvider implements JobProvider<KubernetesJobStatus> {
...
   KubernetesJobStatus collectJob(String account, String location, String id) {
-    def credentials = accountCredentialsProvider.getCredentials(account).credentials
-    if (!credentials instanceof KubernetesNamedAccountCredentials) {
+    def credentials = accountCredentialsProvider.getCredentials(account)
+    if (!(credentials instanceof KubernetesNamedAccountCredentials)) {
       return null
     }
-    def trueCredentials = credentials as KubernetesNamedAccountCredentials
+    def trueCredentials = (credentials as KubernetesNamedAccountCredentials).credentials
     def status = new KubernetesJobStatus(trueCredentials.apiAdaptor.getPod(location, id), account)
```
danielpeach: You could reach this line with a set of `KubernetesV2Credentials`, and this would fail unexpectedly (since that class has no `apiAdaptor`). I think the proper check is 
```
if (!(credentials?.credentials instanceof KubernetesV1Credentials)) { 
  return null
}
```

danielpeach:
Thanks.

Could you please use the standard commit message style? It allows us to autogenerate changelogs. 

e.g., `fix(provider/kubernetes): collectJob status ClassCastException`.

PR #2030, opened 2017-10-25
feat(provider/google): Support UHCs in L7 load balancers.

jtk54:
Reprising unified health check support for L7. The only other place UHCs are not supported yet are Network LBs, but I need to research whether they should support UHCs or not.


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/loadbalancer/UpsertGoogleHttpLoadBalancerAtomicOperation.groovy
```diff
@@ -227,6 +229,7 @@ class UpsertGoogleHttpLoadBalancerAtomicOperation extends UpsertGoogleLoadBalanc
         "Please specify one object per named backend service.")
     }
 
+    List<String> legacyHealthCheckNames = legacyHealthChecks.collect { it.getName() }
```
danielpeach: Could be `legacyHealthChecks*.name`

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/loadbalancer/UpsertGoogleHttpLoadBalancerAtomicOperation.groovy
```diff
@@ -255,37 +259,21 @@ class UpsertGoogleHttpLoadBalancerAtomicOperation extends UpsertGoogleLoadBalanc
 
       if (!healthCheckExistsSet.contains(healthCheck.name)) {
```
danielpeach: If I'm converting a legacy health check to a UHC - that's an `insert`, and I'm going down this path, right?

danielpeach: (It's not so clear to me how `healthCheckExistsSet` would not include my legacy health check, which is why I feel like I'm missing something).

jtk54: `healthCheckExistsSet` is populated from the list of UHCs (Added on [this line](https://github.com/jtk54/clouddriver/blob/eb182ef4fef937b290b547ad73c489e54b0db808/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/loadbalancer/UpsertGoogleHttpLoadBalancerAtomicOperation.groovy#L219)). The idea is that if there is no UHC of that name (whether it's a legacy health check or just missing), we'll `insert` a UHC and point to it from the LB config instead of the legacy health check.

jtk54: So yes to your initial question.

PR #2031, opened 2017-10-26
feat(provider/kubernetes): use kubectl to handle deploy apply

lwander:
Since kubectl uses a three-way merge strategy to decide how to handle deploys that isn't documented, the raw API call to handle create/deploy isn't enough.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubectlJobExecutor.java
```diff
@@ -0,0 +1,138 @@
...
+import org.springframework.stereotype.Component;
+
+import java.io.ByteArrayInputStream;
+import java.util.ArrayList;
+import java.util.List;
+
+@Component
+@Slf4j
+public class KubectlJobExecutor {
+  @Value("${kubernetes.kubectl.poll.minSleepMillis:100}")
```
danielpeach: Should top-level kubernetes config values be namespaced by the provider version? Or is this meant as a job executor for all provider versions (e.g., `deployManifest` could be overloaded to allow for other versions' credentials)?

lwander: I figure this can be shared between providers once needed

PR #2032, opened 2017-10-26
feat(core): Manifest provider

lwander:


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ManifestController.java
```diff
@@ -0,0 +1,76 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Slf4j
+@RestController
+@RequestMapping("/applications/{application}/manifests")
```
jtk54: Does this path variable need a `:.+`?

lwander: Yes

PR #2033, opened 2017-10-27
feat(provider/kubernetes): v2 manifest provider

lwander:


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/Manifest.java
```diff
@@ -17,8 +17,8 @@
 
 package com.netflix.spinnaker.clouddriver.model;
 
-public interface Manifest<S> {
-  S getStatus();
-  boolean isReady();
+public interface Manifest {
+  boolean isStable();
```
danielpeach: What does this signify?

lwander: That the manifest has reached it desired state - not implemented yet - that'll come later.

In k8s we can't track an operation through to an action that k8s takes. All we know is if k8s is happy with the state it's in. We'll be using this in orca after a deploy, update, resize, etc... to decide if the next step is ready. It'll roll up info like health & such. I'll add comments as this code matures.

lwander:
spinnaker/spinnaker#1902

lwander:
spinnaker/spinnaker#2018

PR #2034, opened 2017-10-27
feat(provider/ecs): ECS Account Credentials

dkirillov:
feat(provider/ecs): ECS Account Credentials

Added EcsCredentialsInitializer and supporting classes to enable the ECS module to have an Amazon ECS account.
ECS account depdends on a corresponding AWS account being configured and enabled.

@cfieber @ajordens @BrunoCarrier


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonCredentials.java
```diff
@@ -39,7 +39,8 @@
  *
  */
 public class AmazonCredentials implements AccountCredentials<AWSCredentials> {
-    private static final String CLOUD_PROVIDER = "aws";
+    // TODO: CLOUD_PROVIDER was originally private static final, this needs discussion. It was done to allow AmazonCredentials to serve different cloud providers that use Amazon services, for instance ECS.
+    public String CLOUD_PROVIDER = "aws";
```
robzienert: Not really stoked on this change, as it breaks the immutability of the class. Have you considered an `EcsCredentials` class that extends `AmazonCredentials` (or `NetflixAmazonCredentials`) that changes this?

BrunoCarrier: Extending `AmazonCredentials` wouldn't allow us to modify the `CLOUD_PROVIDER` field, so that's not feasible.  

An alternative could be to move all current `AmazonCredentials` to a `BaseAmazonCredentials` class, except for the `CLOUD_PROVIDER` field.  Then `AmazonCredentials` would have as its only code `private static final String CLOUD_PROVIDER = "aws";` and `EcsCredentials` would have only `private static final String CLOUD_PROVIDER = "ecs";`

I think that'd be a great way to ensure immutability on your end, and let us avoid code duplication.  What do you think @robzienert ?

robzienert: `getCloudProvider()` is a method that can be overwritten, which returns this (private) constant. Couldn't that method just be overridden instead, or perhaps the constructor modified to take an `AmazonCloudProvider` enum that has either `AMAZON` or `ECS` as values?

dkirillov: I'll make the constructor take in an enum.
There will then be two constructors - one that takes in an enum and another that is currently there and which will passing in `AMAZON` to the overloaded one.

dkirillov: @robzienert On second thought, lets go with extending and overwriting as was the first suggestion.

The quick reaction to pick enum, mainly came from the thought of possibly another amazon cloud provider coming along after ECS - they would just need to add an enum.

After trying to pass an enum to an overloaded constructor first `AmazonCredentials`, then realizing I would need to modify `NetflixAmazonCredentials` as well - leading me to make a few constructor... At that point I wasn't comfortable making those changes - potentially causing a fire somewhere off in the distance.

dkirillov:
I've made the `NetflixECSCredentials` (which extends `NetflixAmazonCredentials`).

@robzienert please have a look and see if it's fine.

Thank you! :D

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/security/EcsAccountMaker.java
```diff
@@ -0,0 +1,92 @@
...
+import com.netflix.spinnaker.clouddriver.aws.security.NetflixAmazonCredentials;
+import com.netflix.spinnaker.clouddriver.aws.security.NetflixAssumeRoleAmazonCredentials;
+import com.netflix.spinnaker.clouddriver.aws.security.config.CredentialsConfig;
+import com.netflix.spinnaker.fiat.model.Authorization;
+import com.netflix.spinnaker.fiat.model.resources.Permissions;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class EcsAccountMaker {
```
robzienert: Pedantic, but maybe `EcsAccountBuilder` instead, and then `build()` as the method?

dkirillov: Done.

PR #2035, opened 2017-10-27
chore(depdencies): Update spinnaker-dependencies version.

jtk54:
I mistyped the spinnaker-depencies version when I released it, so it's v0.118.9.


jtk54:
I can't see any LBs in Spinnaker when I pick up this version of spinnaker-dependencies, so I'm closing this for now.

PR #2036, opened 2017-10-27
chore(dependencies): Bump spinnaker-dependencies version and pin google

jtk54:
api client lib.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,7 @@ allprojects {
   apply plugin: 'groovy'
 
   ext {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -108,6 +108,7 @@ subprojects { project ->
             it.useVersion '4.3'
           }
         }
+        force 'com.google.api-client:google-api-client:1.22.0'
```
duftler: @danielpeach I believe when you rollback the GAE lib version you can remove this line.

@jtk54 Is this correct?

jtk54: Yes, that's correct.

PR #2042, opened 2017-10-30
feat(cats): Add keyspaces support to dynomite backend

robzienert:
Dynomite only supports db0, so we can't flip keys in the same dynomite cluster. This PR allows us to replicate the a/b flips performed in Redis with a single Dynomite cluster. That said, when performing a flip, we'll double our write overhead for the duration of the caching agents' red/black.

cc @asher @cfieber @ajordens


asher:
lgtm

PR #2048, opened 2017-10-31
fix(provider/gce): Fix GCE destroy backend calls on LB cache misses.

jtk54:
The GCE calls on cache misses were passing incorrectly typed arguments to the 'delete backends' closure, causing havoc.


jtk54:
@edwinavalos FYI.

PR #2052, opened 2017-11-01
fix(provider/kubernetes): allow cache data to be serailized/deserialized

lwander:


ajordens:
What is different about k8s that needs this?

That retrofitDefault group dependency doesn’t seem like the right one to add to cats-core (cats-core is basically dep free and retrofitDefault must be more than bare Jackson?)

lwander:
Every other provider creates a custom `MutableCacheData` class that extends `CacheData` just for this purpose. This is the start to consolidating all those different implementations. 

Yeah let me find a better dependency

ajordens:
`jackson` is passable, but you really only need jackson-annotations.

`compile "com.fasterxml.jackson.core:jackson-annotations:${spinnaker.version('jackson')}"`

For whatever reason spinnaker-dependencies (wrongly!) has:

```
jacksonAnnotations: "com.fasterxml.jackson.core:jackson-databind:${versions.jackson}"
```

Otherwise this seems reasonable. Worth noting that the groovy `MutableCacheData` has `[:].withDefault { [] as Set }` so this won't be a simple drop-in replacement.

lwander:
Neat, switched to the `jacksonAnnotations`. Should I open a spinnaker-dependencies PR to fix that package name? I'm asking because it might break some things in weird ways.

Yeah good point on the `.withDefault`

ajordens:
Sure, let's fix `spinnaker-dependencies` ... I think it's just a typo as we also have `jacksonDatabind` referencing the exact same artifact. 

Otherwise, let's ship it.

cfieber:
the .withDefault bit can probably be worked around via java 8 map createIfAbsent

lwander:
https://github.com/spinnaker/spinnaker-dependencies/pull/131

PR #2055, opened 2017-11-01
fix(aws): Flagging elb security group autocreation (defaults to off)

jeyrschabu:
- made auto creation of <appname>-elb group for new ELBs configurable
- defaulting to off


jeyrschabu:
@cfieber PTAL (cc: @ewiseblatt )

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/UpsertAmazonLoadBalancerAtomicOperation.groovy
```diff
@@ -71,6 +73,9 @@ class UpsertAmazonLoadBalancerAtomicOperation implements AtomicOperation<UpsertA
   @Autowired
   SecurityGroupLookupFactory securityGroupLookupFactory
 
+  @Value("\${aws.autoCreateELBGroup:false}")
```
cfieber: could you add this flag to the AwsConfiguration.DeployDefaults class then autowire that bean in -- that is where the flag for the application security group creation is (the config namespace would then be `aws.defaults.`)

jeyrschabu:
@ewiseblatt this will take care of that. the autocreation creates a <appname>-elb group and a <appname> group if it doesnt exist.

ewiseblatt:
ok. I assume this is good since cameron is happy. If not I'll ping you
again.

PR #2058, opened 2017-11-01
fix(provider/google): Cache and propagate GCE numeric instance id.

duftler:


edwinavalos:
I think this breaks instance rebooting (or something related to this PR), haven't tracked it down completely yet, but in 1.701.0 we can't reboot instances:

screen shot 2017-11-05 at 6 16 28 pm

PR #2061, opened 2017-11-01
fix(gce/defaults): instanceType vs. instance-type

loomisdf:
I noticed that in the defaults there is both `instanceType` and `instance-type`. PR is for making them consistent. Unless there's a specific reason for the difference.


duftler:
Needs a rebase and then we can merge.

loomisdf:
@duftler should be good to go now

PR #2062, opened 2017-11-01
fix(amazon/alb): Reorder/refactor alb upsert to handle updating port and protocol

jrsquared:


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -177,153 +104,224 @@ class LoadBalancerV2UpsertHandler {
...
+        task.updateStatus BASE_PHASE, "Failed to delete listener ${it.listenerArn}. Listener could not be found. ${e.errorMessage}"
       }
+    }
+  }
+
+  static void updateLoadBalancer(AmazonElasticLoadBalancing loadBalancing, LoadBalancer loadBalancer,
+                                        Collection<String> securityGroups,
+                                        List<UpsertAmazonLoadBalancerV2Description.TargetGroup> targetGroups,
+                                        List<UpsertAmazonLoadBalancerV2Description.Listener> listeners) {
+    def amazonErrors = []
```
cfieber: I noticed that the operation keeps on going and just accumulates error messages as it goes.

I wonder if there are some points it should bail out if it has encountered failures?  Perhaps not deleting anything if the creation phases failed?

jrsquared: I do check if listeners were all created on line 318 before removing the old ones, but good call, I'll check if there's more places to safeguard.

PR #2064, opened 2017-11-02
perf(cluster): Optimize target/LARGEST lookups

ajordens:
If `ClusterProvider.supportsMinimalClusters()` is true, we can now sort and
filter on the set of unexpanded server groups.

Only the actual LARGEST server group will be expanded.

This offers significant savings for target/LARGEST lookups on
clusters with > 1 server group, particularly those containing _many_
instances.

A target/LARGEST lookup commonly occurs when doing a clone operation
with `useSourceCapacity` set to `true`.

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ClusterController.groovy
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ClusterController.groovy
@@ -234,10 +234,6 @@ class ClusterController {
       )
     }
 
-    def expandServerGroups = { List<ServerGroup> serverGroups ->
-      return sortedServerGroups.collect { expandServerGroup(it) }
-    }
-
     switch (tsg) {
       case TargetServerGroup.CURRENT:
         return expandServerGroup(sortedServerGroups.get(0))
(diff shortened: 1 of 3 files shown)


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ClusterController.groovy
```diff
@@ -254,10 +250,10 @@ class ClusterController {
         return expandServerGroup(sortedServerGroups.last())
       case TargetServerGroup.LARGEST:
         // Choose the server group with the most instances, falling back to newest in the case of a tie.
-        return expandServerGroups(sortedServerGroups).sort { lhs, rhs ->
+        return expandServerGroup(sortedServerGroups.sort { lhs, rhs ->
```
ajordens: I've been wanting a way to avoid having to pre-expand server groups in order to sort by instance count.

Upon inspection of the Amazon and Titus caching agents, we are explicitly serializing instance details into `attributes` (and not just relationships). A spot check of other `*ServerGroupCachingAgents` makes me think that others are doing something similar.

Although they aren't "real" instances (missing health amongst other things), it's sufficient for the `AmazonClusterProvider` to use when the `INSTANCES` relationship is not populated.

Is this crazy? What say you @cfieber?

This only kicks in when `ClusterProvider. supportsMinimalClusters()` is `true` (Amazon-only right now with Titus support out for PR).

ajordens:
Decent improvements fetching `LARGEST` from a cluster with 25 server groups.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ClusterController.groovy
cfieber: it seems reasonable to me - would be sort of nice if we had a generic ClusterProvider spec that asserted the expectations around supportsMinimalClusters but not sure what exactly that might look like

PR #2067, opened 2017-11-02
feat(provider/kubernetes): automatically configure caching agents

lwander:
This allows us to not hard-code a list of caching agent types, and have them be discovered based on the "Handler" for the resource that owns them.


lwander:
spinnaker/spinnaker#1902

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2CachingAgentDispatcher.java
```diff
@@ -44,17 +48,52 @@
...
+                    );
+                  } catch (NoSuchMethodException e) {
+                    log.warn("Missing canonical constructor", e);
+                    return null;
+                  }
+                }
+            )
+            .map(c -> {
+                  try {
+                    return (KubernetesV2CachingAgent) c.newInstance(
```
danielpeach: Won't you hit an NPE on `c.newInstance` if you really returned `null` in the last `.map`? ...or do I not know Java?

lwander: No I think you know more than I do - will fix in the morning

PR #2068, opened 2017-11-02
feat(provider/kubernetes): check controller type for disable/enable operation

chlung:
About this PR,

Add code to check the controller type and determine if it needs to skip disable/enable server group operation.  The reason is to work around orca logic will always call disable/enable server group first before destroy server group.

Thanks.


chlung:
@lwander please review.  thanks.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/ops/servergroup/AbstractEnableDisableKubernetesAtomicOperation.groovy
```diff
@@ -151,4 +158,22 @@ abstract class AbstractEnableDisableKubernetesAtomicOperation implements AtomicO
 
     null // Return nothing from void
   }
+
+  Boolean checkControllerType() {
```
lwander: This should probably be something more descriptive, like "supportsEnableDisable"

chlung:
@lwander updated per your feedback.  please review.  thanks.

PR #2071, opened 2017-11-03
fix(provider/openstack): Fix type error in LoadBalancerSummary

johnrengelman:
Fixes the `Exception: upsertLoadBalancer failed: Unable to process request` error when creating a LoadBalancer even though the LB is created in Openstack.


clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/provider/view/OpenstackLoadBalancerProvider.groovy
```diff
@@ -216,25 +216,25 @@ class OpenstackLoadBalancerProvider implements LoadBalancerProvider<OpenstackLoa
...
+    List<OpenstackLoadBalancerByRegion> getByRegions() {
       mappedRegions.values() as List
     }
   }
 
-  static class OpenstackLoadBalancerAccountRegion implements LoadBalancerProvider.ByRegion {
+  static class OpenstackLoadBalancerByRegion implements LoadBalancerProvider.ByRegion {
     String name
-    List<OpenstackLoadBalancerSummary> loadBalancers
+    List<OpenstackLoadBalancerDetail> loadBalancers
```
johnrengelman: This line is the real fix. Everything else is because of the class rename.

PR #2074, opened 2017-11-03
feat(provider/kubernetes): v2 statefulset support

lwander:
7f72wfbqzyx

Automatically shows up as a server group since it's registered as one


lwander:
spinnaker/spinnaker#1902

PR #2076, opened 2017-11-03
fix(cats/redis): Don't hash ttl'd keys

lwander:
The issue here is that when a key with a ttl is hashed, the hash
persists forever, even after the key disappears. The obvious fix would
be to associate the hash with a ttl as well, but as far as I can tell,
there is no good way to ensure that the hash will always expire
when/before the key does without introducing some subtle race.

There's not a lot of keys that do have a ttl, so I don't see this having
a large performance impact. It feels like the cleanest/most contained
way to deal with the problem described above.


lwander:
I can add it to the dyno cache if that makes sense as a part of this PR

lwander:
https://github.com/spinnaker/clouddriver/pull/2080

PR #2081, opened 2017-11-04
fix(provider/amazon): Commentary on when includePartialInstances = false

ajordens:


ajordens:
Also verifying that tests still get run for PR builds

PR #2084, opened 2017-11-06
fix(provider/gce): Distinguish instance id from gceId.

duftler:


duftler:
Looks like my change here (https://github.com/spinnaker/clouddriver/pull/2058) caused instance-focused operations to start using the gce numeric id instead of the instance name.

edwinavalos:
I'm testing this morning against our lab environment as well.

edwinavalos:
We're good on our side, test pipelines still work and individual instance actions now work.

PR #2091, opened 2017-11-07
feat(provider/ecs): ECS Cluster caching classes and tests.

dkirillov:
feat(provider/ecs): ECS Cluster caching classes and tests.

Adds an `EcsClusterCachingAgent` which is responsible for caching ECS clusters, along with `EcsClusterCacheClient`.
Tests included!

@cfieber @ajordens @robzienert  @BrunoCarrier


clouddriver-ecs/src/test/java/com/netflix/spinnaker/clouddriver/ecs/cache/EcsClusterCacheClientTest.java
```diff
@@ -0,0 +1,57 @@
...
+import spock.lang.Subject;
+
+import java.util.Collections;
+import java.util.Map;
+
+import static com.netflix.spinnaker.clouddriver.ecs.cache.Keys.Namespace.ECS_CLUSTERS;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.when;
+
+public class EcsClusterCacheClientTest extends CommonCacheClient {
```
robzienert: Would tend to recommend Spock over straight up junit & mockito. It's more consistent with the rest of the project and will cut down your boilerplate. While we don't encourage usage of Groovy in `main`, we're totally down with continued Groovy use in `test`.

dkirillov: @robzienert can the tests be removed _for now_, and be added in a separate pull.
Currently have them all in java/junit.

robzienert: We can merge as-is with the tests as they are (if the PR is ready), then double back for Spock... but definitely don't want to add code without tests at all.

dkirillov: Sounds good 👍

PR #2092, opened 2017-11-07
feat(provider/amazon): block device config for c5 instance types

asher:


ajordens:
The powers that be were happy with this?

asher:
@ajordens yep, checked with the base ami / perf folks

PR #2093, opened 2017-11-07
feat(provider/kubernetes): v2 server group manager

lwander:
spinnaker/spinnaker#1902

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupManagerController.java
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupManagerController.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.clouddriver.controllers;
+
+import com.netflix.spinnaker.clouddriver.model.ServerGroupManager;
+import com.netflix.spinnaker.clouddriver.model.ServerGroupManagerProvider;
+import com.netflix.spinnaker.clouddriver.requestqueue.RequestQueue;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.access.prepost.PostAuthorize;
+import org.springframework.security.access.prepost.PostFilter;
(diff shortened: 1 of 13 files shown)


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupManagerController.java
```diff
@@ -0,0 +1,71 @@
...
+  @PreAuthorize("hasPermission(#application, 'APPLICATION', 'READ')")
+  @PostFilter("hasPermission(filterObject.account, 'ACCOUNT', 'READ')" )
+  @RequestMapping(method = RequestMethod.GET)
+  Set<ServerGroupManager> getForApplication(@PathVariable String application) {
+    return serverGroupManagerProviders.stream()
+        .map(provider -> {
+          try {
+            return requestQueue.execute(application, () -> provider.getServerGroupManagersByApplication(application));
+          } catch (Throwable t) {
+            log.warn("Failed to read manifest " , t);
```
danielpeach: This feels kubernetes-specific.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupManagerController.java
```diff
@@ -0,0 +1,71 @@
...
+  @Autowired
+  public ServerGroupManagerController(List<ServerGroupManagerProvider> serverGroupManagerProviders, RequestQueue requestQueue) {
+    this.serverGroupManagerProviders = serverGroupManagerProviders;
+    this.requestQueue = requestQueue;
+  }
+
+  @PreAuthorize("hasPermission(#application, 'APPLICATION', 'READ')")
+  @PostFilter("hasPermission(filterObject.account, 'ACCOUNT', 'READ')" )
+  @RequestMapping(method = RequestMethod.GET)
+  Set<ServerGroupManager> getForApplication(@PathVariable String application) {
```
danielpeach: A server group manager's server groups might not all live in the same application, right? I wonder how the UI should show that...

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupManagerController.java
lwander: Ah this was copied out of the "ManifestController", where that warning actually makes sense

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupManagerController.java
lwander: That's fine, if you're on a certain "cluster" tab you would only see the server groups in that app. Clicking on the server group manager could show you the list of owned server groups (even ones in different apps) in the details panel the same way we do for load balancers.
ygauhxgpjwl

lwander: Ah are you worried about server groups who's manager isn't in the current app? It's probably safe to ignore those (at least for now).

lwander: We could also annotate server groups with their managers, and if one points to a manager not in this app, we can add a little button like the load balancer button, but this feels like an unlikely use-case.

PR #2096, opened 2017-11-08
feat(provider/amazon): Optionally tag server groups w/ app/stack/details

ajordens:
Set `aws.defaults.addAppStackDetailTags: true` to enable.

When enabled, all created server groups will be tagged with:
- `spinnaker:application`
- `spinnaker:stack`
- `spinnaker:details`

The tag will only be created if the value is non-null.

By default, these tags will be propagated to launched instances.


jrsquared:
👍

PR #2097, opened 2017-11-08
feat(provider/kuberentes): v2 security group provider

lwander:


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2SecurityGroup.java
```diff
@@ -0,0 +1,160 @@
...
+    if (manifest == null) {
+      log.warn("Cache data {} inserted without a manifest", cd.getId());
+      return null;
+    }
+
+    Set<Rule> inboundRules = new HashSet<>();
+    Set<Rule> outboundRules = new HashSet<>();
+
+    if (manifest.getKind() != KubernetesKind.NETWORK_POLICY) {
+      log.warn("Unknown security group kind " + manifest.getKind());
```
jtk54: If this `warn` is hit, this function still returns a security group with no inbound or outbound rules. Is that OK or should these be more than warnings?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2SecurityGroup.java
```diff
@@ -0,0 +1,160 @@
...
+        .setProtocol(policyPort.getProtocol())
+        .setPortRanges(new TreeSet<>(Collections.singletonList(new StringPortRange(port))));
+  }
+
+  @Data
+  @NoArgsConstructor
+  @AllArgsConstructor
+  @Builder
+  private static class KubernetesV2SecurityGroupSummary implements SecurityGroupSummary {
+    String name;
```
jtk54: Access modifiers here.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2SecurityGroup.java
lwander: Yup, this warning is a signal that the logic needs to be implemented. Returning `null` would be surprising since at least an object with a `manifest` is expected to be returned.

PR #2104, opened 2017-11-10
feat(provider/kubernetes): Add kubectl to Dockerfile.

andrewbackes:
The new k8s-v2 provider requires `kubectl`


andrewbackes:
@lwander

PR #2112, opened 2017-11-13
perf(provider/gce): s/get/list/ in ILB caching agent.

jtk54:
A few small fixes/guards to the L7 caching agent as well.


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleHttpLoadBalancerCachingAgent.groovy
```diff
@@ -335,7 +335,7 @@ class GoogleHttpLoadBalancerCachingAgent extends AbstractGoogleLoadBalancerCachi
 
       // Process queued backend services.
       queuedServices?.each { backendServiceName ->
-        BackendService service = projectBackendServices.find { bs -> Utils.getLocalName(bs.getName()) == backendServiceName }
+        BackendService service = projectBackendServices?.find { bs -> Utils.getLocalName(bs.getName()) == backendServiceName }
```
lwander: Does `getLocalName` crash if passed `null`?

jtk54: No, it returns null.

PR #2114, opened 2017-11-13
feat(core/search): Add parameter for fetching a batch of servergroups by name

christopherthielen:


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -160,9 +160,35 @@ class ServerGroupController {
   @PostFilter("hasPermission(filterObject?.application, 'APPLICATION', 'READ')")
   @PostAuthorize("@authorizationSupport.filterForAccounts(returnObject)")
   @RequestMapping(value = "/serverGroups", method = RequestMethod.GET)
-  List getServerGroupsByApplications(@RequestParam(value = 'applications') List<String> applications,
-                                     @RequestParam(required = false, value = 'cloudProvider') String cloudProvider) {
-    applications.collectMany { summaryList(it, cloudProvider) }
+  List getServerGroups(@RequestParam(required = false, value = 'applications') List<String> applications,
```
christopherthielen: Hoping that renaming endpoint methods doesn't cause any problems...

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -160,9 +160,35 @@ class ServerGroupController {
   @PostFilter("hasPermission(filterObject?.application, 'APPLICATION', 'READ')")
   @PostAuthorize("@authorizationSupport.filterForAccounts(returnObject)")
   @RequestMapping(value = "/serverGroups", method = RequestMethod.GET)
```
christopherthielen: I added server group names as a parameter here because `/serverGroups?serverGroupNames=foo,bar,baz` seemed the most appropriate url

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -160,9 +160,35 @@ class ServerGroupController {
...
+
+  private List<ServerGroupViewModel> getServerGroupsForNames(List<String> serverGroupNames) {
+    return serverGroupNames.collect {
+      def segments = it.split(':')
+      if (segments.size() != 3) {
+        throw new IllegalArgumentException("Expected serverGroupNames in the format [account1:region1:name1,account2:region2:name2] but got " + serverGroupNames)
+      }
+
+      def (account, region, name) = segments;
+      def serverGroup = getServerGroup(account, region, name)
```
christopherthielen: If a single server group is not found, the entire call fails with a 404.  Does this seem appropriate, or should the results be returned without missing data?

ajordens: I don't think I would expect this to throw a 404 ... more generally I wouldn't expect any root endpoint (ie. `/serverGroups`) that returns a collection and takes filter query parameters to 404. 

There's a whole other discussion / argument to be had about our inconsistent APIs and the pain associated with _only_ returning a collection vs. an envelope. The latter would allow us to return both the collection of matched server groups and a set of missing server groups if desired.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -160,9 +160,35 @@ class ServerGroupController {
...
+
+  private List<ServerGroupViewModel> getServerGroupsForApplications(List<String> applications, String cloudProvider) {
+    return applications.collectMany { summaryList(it, cloudProvider) }
+  }
+
+  private List<ServerGroupViewModel> getServerGroupsForNames(List<String> serverGroupNames) {
+    return serverGroupNames.collect {
+      def segments = it.split(':')
+      if (segments.size() != 3) {
+        throw new IllegalArgumentException("Expected serverGroupNames in the format [account1:region1:name1,account2:region2:name2] but got " + serverGroupNames)
```
ajordens: Maybe move this outside the loop and split every `serverGroupName`, and throw an exception containing only the ones that failed to `split`.

Error message is also misleading, at this point you're not dealing with a query parameter so no need to comma-separate ... might be better as some variant of:

```Expected server group names like <account>:<region>:<name> but got ${invalidServerGroupNames.join(", ")```

File: clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy (hunk 3 above)
Comment (ajordens) [reply clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy]: Not sure how many server groups are likely to be in this request ... might be an inefficient lookup if there are _lots_.

Comment (christopherthielen) [reply clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy]: >  I wouldn't expect any root endpoint (ie. /serverGroups) that returns a collection and takes filter query parameters to 404

Sure.  I was thinking of this less as a filter query (tell me what's there that satisfies this query) and more as a batched "get by id" (I know these things exist, give me them).  For my use case, the former is actually more convenient.  I will address this.

> might be an inefficient lookup if there are lots

For search v2, this will be small batches (25 right now).  Is there a more efficient mechanism I could be using?

Comment (ajordens) [reply clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy]: Let's see how it goes ... I'm not sure ... especially if the server groups are somewhat spread across region / application.

File: clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy (hunk 1 above)
Comment (christopherthielen) [reply clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy]: Hmm... I feel like `ids` is also misleading, but maybe less so?

Comment (ajordens) [reply clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy]: IMO it's less misleading ... but you're right, hard to say.

Comment (ajordens) [reply clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy]: FWIW for entity tags we did

```
https://api....net/tags?entityId=clouddriver-main-readonly-elb,clouddriver-prestaging-readonly-elb,clouddriver-redis...
```

In this case `entityId` is one of the identity fields on an entity tag. I bring this up as perhaps there's some precedence for going singular `id` vs plural `ids` even if the `id` happens to be comma-separated.

Not sure if we have any other patterns for this in our APIs.

Comment (christopherthielen) [reply clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy]: Since the other param is `applications` I think `ids` should be used here.  seem ok?

**Comment by christopherthielen** [comment]
updated with requested changes

File: clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy (hunk 1 above)
Comment (ajordens) [reply clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy]: 👍

File: clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy [hunk 5]
```diff
@@ -160,9 +162,43 @@ class ServerGroupController {
...
+      return getServerGroupsForIds(ids)
+    }
+  }
+
+  private List<ServerGroupViewModel> getServerGroupsForApplications(List<String> applications, String cloudProvider) {
+    return applications.collectMany { summaryList(it, cloudProvider) }
+  }
+
+  private List<ServerGroupViewModel> getServerGroupsForIds(List<String> serverGroupIds) {
+    String[][] allIdTokens = serverGroupIds.collect { it.split(':') }
```
ajordens: Would have expected this to be `List<String[]> allIdTokens` but whatever works I suppose.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -160,9 +162,43 @@ class ServerGroupController {
...
+
+    def invalidIds = allIdTokens.findAll { it.size() != 3 }
+    if (!invalidIds.empty) {
```
ajordens: Can just be `if(invalidIds)`

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -160,9 +162,43 @@ class ServerGroupController {
...
+        invalidIds.collect { it.join(':') }.join(', '))
+    }
+
+    allIdTokens.collect { String[] idTokens ->
+      def (account, region, name) = idTokens
+      try {
+        def serverGroup = getServerGroup(account, region, name)
+        return new ServerGroupViewModel(serverGroup, serverGroup.moniker.cluster, account)
+      } catch (ignored) {
+        log.info("Couldn't get server group ${idTokens.join(':')}")
```
ajordens: Seems like a strange thing to actually ignore?

`log.error("...", e)` (unless there's a good reason to ignore and not `.error` it

PR #2119, opened 2017-11-14
feat(provider/kubernetes): scale manifest

lwander:


lwander:
spinnaker/spinnaker#1902

PR #2120, opened 2017-11-15
feat(provider/kubernetes): pause rollout

lwander:


lwander:
spinnaker/spinnaker#1902

PR #2122, opened 2017-11-15
fix(docker): Use wget instead of curl

andrewbackes:
Alpine does not include curl. Need to use wget instead.

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -11,7 +11,7 @@ RUN apk --no-cache add --update bash wget unzip python=2.7.12-r0 && \
   CLOUDSDK_PYTHON="python2.7" /opt/google-cloud-sdk/install.sh --usage-reporting=false --bash-completion=false --disable-installation-options && \
   rm -rf ~/.config/gcloud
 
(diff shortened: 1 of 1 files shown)


Dockerfile.slim
```diff
@@ -11,7 +11,7 @@ RUN apk --no-cache add --update bash wget unzip python=2.7.12-r0 && \
   CLOUDSDK_PYTHON="python2.7" /opt/google-cloud-sdk/install.sh --usage-reporting=false --bash-completion=false --disable-installation-options && \
   rm -rf ~/.config/gcloud
 
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl && \
+RUN wget https://storage.googleapis.com/kubernetes-release/release/stable.txt && wget https://storage.googleapis.com/kubernetes-release/release/$(cat stable.txt)/bin/linux/amd64/kubectl && \
```
lwander: Should we run `rm stable.txt` as another RUN step?

andrewbackes: yup. good call

Dockerfile.slim
```diff
@@ -11,7 +11,8 @@ RUN apk --no-cache add --update bash wget unzip python=2.7.12-r0 && \
...
+  rm stable.txt && \
```
andrewbackes: Added it here

PR #2134, opened 2017-11-16
fix(provider/kubernetes): fix security group loading by name

lwander:


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/SecurityGroupController.groovy
```diff
@@ -158,9 +158,10 @@ class SecurityGroupController {
                     @PathVariable String region,
                     @PathVariable String securityGroupName,
                     @RequestParam(value = "vpcId", required = false) String vpcId) {
-    def securityGroup = securityGroupProviders.find { secGrpProv ->
-      secGrpProv.cloudProvider == cloudProvider
-    }.get(account, region, securityGroupName, vpcId)
+    def securityGroup = securityGroupProviders.collect { secGrpProv ->
+      secGrpProv.cloudProvider == cloudProvider ? secGrpProv.get(account, region, securityGroupName, vpcId) : null
+    }.findAll({ it })
```
danielpeach: Isn't your `collect` + `findAll` a  `findResults`?

lwander: Ahhh I was doing `findAll` thinking it was supposed to be `findResults`, this is better

PR #2139, opened 2017-11-16
fix(provider/gce): Tolerate named port values as Doubles. (#2137)

jtk54:
Pipelines are being persisted with 'port' values of NamedPorts as
Doubles. This is an issue. This change tolerates that behavior in deploys
since users will have pipelines with Double ports in the wild.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/converters/BasicGoogleDeployAtomicOperationConverter.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/converters/BasicGoogleDeployAtomicOperationConverter.groovy
@@ -22,16 +22,31 @@ import com.netflix.spinnaker.clouddriver.google.deploy.description.BasicGoogleDe
 import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation
 import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperations
 import com.netflix.spinnaker.clouddriver.security.AbstractAtomicOperationsCredentialsSupport
+import groovy.util.logging.Slf4j
 import org.springframework.stereotype.Component
 
 @GoogleOperation(AtomicOperations.CREATE_SERVER_GROUP)
 @Component("basicGoogleDeployDescription")
+@Slf4j
 class BasicGoogleDeployAtomicOperationConverter extends AbstractAtomicOperationsCredentialsSupport {
   AtomicOperation convertOperation(Map input) {
     new DeployAtomicOperation(convertDescription(input))
   }
 
   BasicGoogleDeployDescription convertDescription(Map input) {
+    // Note: Deploy descriptions are stored in pipelines such that namedPort ports are cast to doubles. This is a bug.
+    // As a result, pipelines in the wild have namedPort ports stored as doubles.
+    // If a pipeline containing a deploy is executed with a trigger (e.g. a cron trigger), the description converter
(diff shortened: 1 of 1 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/converters/BasicGoogleDeployAtomicOperationConverter.groovy
```diff
@@ -22,16 +22,33 @@ import com.netflix.spinnaker.clouddriver.google.deploy.description.BasicGoogleDe
...
   }
 
   BasicGoogleDeployDescription convertDescription(Map input) {
+    // Note: Deploy descriptions are stored in pipelines such that namedPort ports are cast to doubles. This is a bug.
+    // As a result, pipelines in the wild have namedPort ports stored as doubles.
+    // If a pipeline containing a deploy is executed with a trigger (e.g. a cron trigger), the description converter
+    // chokes and the pipeline fails. The next block handles double ports for namedPorts gracefully.
+    def namedPorts = input?.loadBalancingPolicy?.namedPorts
+    if (namedPorts && namedPorts.any { np -> np.port instanceof Double }) {
+      log.warn("Deploy description contained named ports with ports of type Double." +
```
ewiseblatt: Do you really need a warning here?
Unless you can coerce it to being fixed, the warnings seem like noise.

For example, if you were to write a maintenance script or something that can rewrite the pipelines then the warning makes more sense. Otherwise it seems like noise, especially if anything in existence on or before this next release is going to trigger the warning.

Also, is it possible to "fix" this where teh loadBalancingPolicy namedPorts are loaded?

jtk54: > Unless you can coerce it to being fixed, the warnings seem like noise.

We're coercing an incorrect parameter to be correct within Clouddriver, so we're just warning the user that the pipeline config Clouddriver received had to be translated. This appears to only happen on cron triggers, not everything.

A script won't fix the pipeline config permanently, since any update to the UI will overwrite the fixed config with the incorrect config again until we fix the save.

> Also, is it possible to "fix" this where the loadBalancingPolicy namedPorts are loaded?

It is, I'm looking through Echo's cron code now and Orca's pipeline save code. Clouddriver should still be tolerant of this behavior since there are pipelines in the wild that will be configured and saved before the fix.

lwander: I think 'warn' is good here - while it's not user-actionable yet, it will be once jake finds the root cause in orca or echo. At that point, these are really pointing out that either 1. the user needs to do something or 2. something is really wrong again.

ewiseblatt: Right. Fix the cause, write a script people can run to back fix it, *then* add the warning.
A warning you cant control isnt that helpful, and you are recovering from the issue where the warning is so there's no value to bringing attempt to it. It only serves as a distraction or cause for stress. If you really want to add logging volume, make this a lower level log, then bump it up after you have a corrective action to take.

jtk54:
Ping.

jtk54:
```
> Failed to apply plugin [class 'nebula.plugin.bintray.NebulaBintrayPublishingPlugin']
   > Walk failure.
```

ewiseblatt:
lgtm

PR #2143, opened 2017-11-16
feat(provider/ecs): ECS Container Instance caching classes and tests.

dkirillov:
feat(provider/ecs): ECS Container Instance caching classes and tests.

Adds an ContainerInstanceCachingAgent which is responsible for caching ECS container instance information, along with ContainerInstanceCacheClient.


dkirillov:

PR #2144, opened 2017-11-16
feat(provider/ecs): ECS Service caching classes and tests.

dkirillov:
feat(provider/ecs): ECS Service caching classes and tests.

Adds an ServiceCachingAgent which is responsible for caching ECS service information, along with ServiceCacheClient.


clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/ServiceCacheClient.java
```diff
@@ -0,0 +1,79 @@
...
+    service.setClusterName((String) attributes.get("clusterName"));
+    service.setClusterArn((String) attributes.get("clusterArn"));
+    service.setRoleArn((String) attributes.get("roleArn"));
+    service.setTaskDefinition((String) attributes.get("taskDefinition"));
+    service.setDesiredCount((Integer) attributes.get("desiredCount"));
+    service.setMaximumPercent((Integer) attributes.get("maximumPercent"));
+    service.setMinimumHealthyPercent((Integer) attributes.get("minimumHealthyPercent"));
+
+    if (attributes.containsKey("loadBalancers")) {
+      ObjectMapper mapper = new ObjectMapper();
```
robzienert: It'd seem to make more sense to me to inject the object mapper, rather than create a new one.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/ServiceCachingAgent.java
```diff
@@ -0,0 +1,147 @@
...
+  ));
+  private final Logger log = LoggerFactory.getLogger(getClass());
+
+  public ServiceCachingAgent(String accountName, String region, AmazonClientProvider amazonClientProvider, AWSCredentialsProvider awsCredentialsProvider, Registry registry) {
+    super(accountName, region, amazonClientProvider, awsCredentialsProvider, registry);
+  }
+
+  public static Map<String, Object> convertServiceToAttributes(String accountName, String region, Service service) {
+    Map<String, Object> attributes = new HashMap<>();
+    String applicationName = service.getServiceName().contains("-") ? StringUtils.substringBefore(service.getServiceName(), "-") : service.getServiceName();
```
robzienert: I believe this should be using Moniker instead.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/ServiceCacheClient.java
dkirillov: Done.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/ServiceCachingAgent.java
dkirillov: Could you please clarify this a bit.

`service` is an object from the AWS library. The service name is generally: `appname-stack-detail-version` (stack/detail being optional of course). To get the application name from the service I would still need to parse it out of the service name.

So, a bit confused by what you're suggesting.

robzienert: I'm wondering if the ECS module needs to be updated to use Moniker, rather than directly using the Frigga naming convention? If this `app-stack-detail-version` is enforced by AWS, then I guess there's nothing we can do and it's coincidental. :)

dkirillov: I don't think AWS enforces it :(

Could you please provide some implementation details for Moniker, or an example/lines of code. Admittedly, I think I missed it entirely.

robzienert: Here's an example: https://github.com/spinnaker/clouddriver/blob/4fab668aef7564e8940c62b8044df84927d1239d/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/LoadBalancer.java#L44

@lwander - do you have better implementation examples in CD for Moniker?

lwander: If you want to be able to allow more flexible naming (have users define their own naming policies) you should take a look at https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/names/KubernetesManifestNamer.java

If not, then there is no need to do anything special

dkirillov: Thank you guys for the examples!  The ECS module will need to be updated to use Moniker.

For now, can we please stick to the Frigga, and have an action to switch to Moniker right after we have a working cloud provider?

robzienert: sure.

PR #2146, opened 2017-11-16
feat(provider/ecs): ECS Task caching classes and tests.

dkirillov:
feat(provider/ecs): ECS Task caching classes and tests.

Adds an TaskCachingAgent which is responsible for caching ECS task information, along with TaskCacheClient.


clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/TaskCacheClient.java
```diff
@@ -0,0 +1,71 @@
...
+    task.setTaskArn((String) attributes.get("taskArn"));
+    task.setClusterArn((String) attributes.get("clusterArn"));
+    task.setContainerInstanceArn((String) attributes.get("containerInstanceArn"));
+    task.setGroup((String) attributes.get("group"));
+    task.setLastStatus((String) attributes.get("lastStatus"));
+    task.setDesiredStatus((String) attributes.get("desiredStatus"));
+    task.setStartedAt((Long) attributes.get("startedAt"));
+
+    if (attributes.containsKey("containers")) {
+      ObjectMapper mapper = new ObjectMapper();
```
robzienert: Again, not sure we want to be creating object mappers when we can inject one.

dkirillov: Done.

PR #2151, opened 2017-11-17
feat(clouddriver/gce) - Add 'required' flag to account configs.

johnrengelman:
This prevents clouddriver from loading in the case that it cannot reach an account when it starts. Non-required accounts
that are unreachable are simply ignored from the configuration and never re-attempted.

This is behavior we saw in a clouddriver that had a network issue reaching Google when it started. It had 3 accounts but only 1 was subsequently added to the accountRepository and was served up by the `/credentials` endpoint.

This caused undesirable behavior where, depending on the clouddriver instance that gate or fiat routed to, the number of accounts would change in the UI.


johnrengelman:
Added configuration to sample `clouddriver.yml`

PR #2153, opened 2017-11-17
fix(provider/gce): Tolerate failed backend service getHealth() calls.

jtk54:
Currently, if the backend service getHealth() calls fail, we fail to cache the whole load balancer. This was fine using backend service get()s, since the platform seemed to always return fully-formed and ready backend services. Since we are using one backend service list() at the beginning of the cache cycle, it seems like resources _can be partially-formed_.

Rather than fail the whole LB if a getHealth() fails, we should be tolerant of that failure, mark the health state as unknown, log it, and retry next cycle. The health states are the leaves of the LB configuration, and an unknown health state is an acceptable configuration to Spinnaker (blue health check).


jtk54:
Ping.

PR #2155, opened 2017-11-18
fix(clouddriver/log): Append to clouddriver.log instead of overwrite

kristi:
Append to clouddriver.log so that restarting clouddriver doesn't make the existing log lines disappear.


kristi:
@spinnaker/reviewers

lwander:
Seems reasonable, although usually we recommend exporting logs to some log retention system to avoid eating up the local disk as well as provide more stable storage. WDYT @cfieber?

PR #2158, opened 2017-11-19
feat(provider/kuberentes): send deployed artifact to orca

lwander:


lwander:
https://github.com/spinnaker/spinnaker/issues/2158

PR #2163, opened 2017-11-20
fix(provider/gce): Break out labels.

duftler:


duftler:
I can fake the region by doing a `starts-with` on the zone (which is implicitly available), but would rather be explicit here.

And the `spinnaker-` prefix doesn't bother me too much since we'll likely end up having a bunch of sibling labels for the Moniker stuff (e.g. `spinnaker-app`, `spinnaker-cluster`, ...).

ewiseblatt:
I think you are doing the right thing here based on what you have to work
with.
I was just offering out of band subliminal commentary for future
conversations with related people.

PR #2170, opened 2017-11-21
bug(provider/openstack) - Display VIP for LBs that do not have FIP

johnrengelman:
Currently displays `null` instead.

image


johnrengelman:
@spinnaker-reviewers

johnrengelman:
@edwinavalos rebased and squashed

PR #2173, opened 2017-11-21
fix(provider/kubernetes): fix v1 credentials init

ethanfrogers:
without depending on docker credentials, initialization fails because
KubernetesV1Credentials depends on them being present. this leads to
inability to deploy images properly.

@lwander PTAL. if you can think of another solution, let me know. `KubernetesV1Credentials` _does_ have a hard dependency on `dockerRegistryNamedAccountCredentials`. The credential initialization _does not_ get fixed on the next cycle.


ethanfrogers:
spinnaker/clouddriver#2106

lwander:
This doesn't quite work - k8s v2 doesn't require docker

lwander:
Let me look at why the initialization doesn't get resolved in the next cycle...

ethanfrogers:
Okay, let me know if I can help. This is a blocker for me rolling out the work I've been doing.

PR #2175, opened 2017-11-21
refactor(entitytags): Support for alerts/notices on any entity type

ajordens:


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/LaunchFailureNotificationCleanupAgent.java
```diff
@@ -116,6 +115,7 @@ public void run() {
           AmazonCloudProvider.ID,
           entityRef.getAccountId(),
           entityRef.getRegion(),
+          EntityTagger.ENTITY_TYPE_SERVER_GROUP,
```
ajordens: we now need to supply the `entityType` (previously it was always explicitly `servergroup`).

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/tags/EntityTagger.groovy
```diff
@@ -0,0 +1,63 @@
...
+
+package com.netflix.spinnaker.clouddriver.tags
+
+import com.netflix.spinnaker.clouddriver.model.EntityTags;
+
+interface EntityTagger {
+  public static final String ENTITY_TYPE_SERVER_GROUP = "servergroup"
+  public static final String ENTITY_TYPE_CLUSTER = "cluster"
+
+  void alert(String cloudProvider,
```
ajordens: The parameters are a bit out of control ... next time this is touched I'll introduce some sort of Builder to better encapsulate the different bits.

clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/ops/BulkUpsertEntityTagsAtomicOperation.java
```diff
@@ -243,9 +243,15 @@ private static void replaceTagContents(EntityTags currentTags, EntityTags entity
...
   }
 
-  private static EntityTags.EntityTagMetadata tagMetadata(String tagName, Date now) {
+  private static EntityTags.EntityTagMetadata tagMetadata(EntityTags.EntityTag entityTag, Date now) {
     String user = AuthenticatedRequest.getSpinnakerUser().orElse("unknown");
 
+    String tagName = entityTag.getName();
+    if (entityTag.getTimestamp() != null) {
+      // entity tag has an explicit timestamp, favor it for last modified date
+      now = new Date(entityTag.getTimestamp());
```
ajordens: if we're creating an entity tag for some external event, this lets us use the timestamp of the external event rather than `now`.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/tags/EntityTagger.groovy
```diff
@@ -0,0 +1,63 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.tags
+
+import com.netflix.spinnaker.clouddriver.model.EntityTags;
+
+interface EntityTagger {
+  public static final String ENTITY_TYPE_SERVER_GROUP = "servergroup"
+  public static final String ENTITY_TYPE_CLUSTER = "cluster"
```
robzienert: Any reason you didn't go for an enum instead?

ajordens: No reason ... the previous interface had a String constant that's all.

PR #2177, opened 2017-11-21
feat(cats): Support for selectively enabling agents

ajordens:
This is an assist to local development when you do not want _all_
agents to run by default.

```
redis:
  agent:
    enabledPattern: .*reservation.*
```

By default `enabledPattern` is `.*` and all agents will be scheduled.


cfieber:
I mean you could just get a better laptop, but in the meantime I guess this will do

PR #2178, opened 2017-11-21
fix(provider/kubernetes): Use setApiKey when using token from service account

viglesiasce:
This fixes the use of service accounts for getting credentials when Spinnaker is deployed inside a Kubernetes cluster.

Fix found via https://github.com/kubernetes-client/java/issues/74.
Issue found when testing https://github.com/kubernetes/charts/pull/2616


lwander:
Awesome, thanks Vic

PR #2180, opened 2017-11-22
fix(provider/aws): Reservation caching agent, now with less RxJava

ajordens:


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservationReportCachingAgent.groovy
```diff
@@ -178,15 +178,22 @@ class ReservationReportCachingAgent implements CachingAgent, CustomScheduledAgen
...
-    })
-      .observeOn(scheduler)
-      .toList()
-      .toBlocking()
-      .single()
+
+    Map<NetflixAmazonCredentials, Future> tasks = accounts.collectEntries { NetflixAmazonCredentials credential ->
+      [
+        (credential) : reservationReportPool.submit {
+          extractReservations(reservations, errorsByRegion, credential)
```
cfieber: my rx is bad enough that I can't tell whether the previous version would have used a thread per region in each credential or not but either way I don't think it matters.

PR #2181, opened 2017-11-22
feat(provider/aws): Better handling of permission errors in AWS

jervi:
* Adds an `enabled` flag to accounts (credentials) (defaults to true)
* If access to an AWS account fails due to permission errors in the health check, the account will be disabled, but the health check status will still be UP
* Conversely, the account will be re-enabled if it passes the health check
* The `/credentials` endpoint will not include disabled accounts
* A new `/credentials/disabled` endpoint will list disabled accounts

---

#### OK, a bit of rationale for this PR:

We are managing more than 100 AWS accounts in Spinnaker, and many of these are managed by the teams themselves. If this is a good idea or not is another discussion, but it is what it is.
There have been a couple of occasions where a team has tampered with the permissions of the account, or it has been removed altogether. Within two minutes, all our Clouddriver instances will then be marked as DOWN, and will after a while be replaced by new instances that will never be healthy because of the misconfigured account.

It is less than ideal that a team can, unknowingly, bring down all our clouddriver instances by misconfiguring the Spinnaker permissions; alas this PR.

In the next iteration, I will add a feature to Deck that queries the `/credentials/disabled` endpoint and warns of any disabled accounts.

I hope this looks OK, but I'm all ears if you have suggestions or disagree with our approach 🙂


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/CredentialsController.groovy
```diff
@@ -59,6 +59,11 @@ class CredentialsController {
     return accountDetail
   }
 
+  @RequestMapping(value = "/disabled", method = RequestMethod.GET)
```
ajordens: Not a huge fan of a `/disabled` endpoint.

It should be possible to iteratively make callers of `/credentials` aware of the `enabled` attribute?

jervi: Sure. I just figured this would be less work. But you mean to always include all accounts? Or to make a query param or something at the `/credentials` endpoint?

ajordens:
I was thinking just return all

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/CredentialsController.groovy
jervi: I've removed it

jervi:
I'm closing this, as it seems like #2245 will solve this issue.

PR #2182, opened 2017-11-22
feat(xenial_debians): Add systemd service configuration.

brandonnelson3:
This PR adds a systemd service configuration for CloudDriver. It is coupled with https://github.com/spinnaker/spinnaker/pull/2116, and similar changes in each repository adding a similar config.


clouddriver-web/lib/systemd/system/clouddriver.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Spinnaker CloudDriver
```
duftler: s/CloudDriver/Clouddriver

PR #2188, opened 2017-11-27
fix(provider/kubernetes): fix registry init

ethanfrogers:
if using the `namespaces` key for kubernetes accounts, reconfigure
registries using those namespaces.

@lwander PTAL


ethanfrogers:
it looks like this puts clouddriver into a weird state where it starts looping endlessly and never becomes healthy. i might revert this in the morning as it doesn't appear to be what was intended.

PR #2191, opened 2017-11-28
fix(provider/kubernetes): fix registry init 2

ethanfrogers:
the previous fix caused sever performance issues. this fix will only
configure hard coded registries once, only reconfiguring failed
registries on the next loop, if any.

@lwander PTAL. My Java may be poor here, but it gets the idea across.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/security/KubernetesV1Credentials.java
```diff
@@ -131,8 +131,11 @@ private void configureDockerRegistries() {
 
   public List<String> getDeclaredNamespaces() {
     if (namespaces != null && !namespaces.isEmpty()) {
+      List<String> toInitialize = new ArrayList<String>(namespaces);
```
lwander: The `ArrayList` can omit the type parameter (`ArrayList<>(namespaces)`)

ethanfrogers: 👍

ethanfrogers:
:/ it looks like this didn't quite work as expected. if the list of namespaces passed to `reconfigureRegistries` is an empty list, the `registry.setNamespaces` is also set to an empty list. this causes issues in Deck with the Server Group UI. will issue a fix ASAP.

PR #2193, opened 2017-11-29
fix(google): Retry on all 5xx errors from the platform.

ttomsu:


ttomsu:
The surrounding code counts retries

PR #2201, opened 2017-11-30
feat(provider/gce): Support for RMIG zone selection.

jtk54:
fyi @duftler as well.

Added a metadata entry to handle the case where users just want to use the default zones in a region since there is no flag indicating that in the platform. Removed the (possibly) brittle logic calculating default zones since the platform reports the zones it will use regardless of whether or not a user chose specific zones.


jtk54:
UI is here: https://github.com/spinnaker/deck/pull/4548.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/CopyLastGoogleServerGroupAtomicOperation.groovy
```diff
@@ -128,6 +129,11 @@ class CopyLastGoogleServerGroupAtomicOperation extends GoogleAtomicOperation<Dep
         description.targetSize != null
         ? description.targetSize
         : ancestorServerGroup.capacity.desired
+    newDescription.distributionPolicy =
+      description.distributionPolicy != null ?
```
danielpeach: Can't this also be an elvis?

jtk54: I think we want to explicitly null check since this is initialized with an empty list of zones in the UI.

PR #2203, opened 2017-11-30
fix(amazon/loadBalancers): Be smarter about updating ALB listeners

jrsquared:


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -205,6 +208,58 @@ class LoadBalancerV2UpsertHandler {
...
+    // Compare the old rules; if any are different, just replace them all.
+    boolean rulesSame = existingRules.size() == newRules.size() &&
+      containsAllRules(existingRules, newRules) &&
+      containsAllRules(newRules, existingRules)
+
+    if (!rulesSame) {
+      existingRules.each { rule ->
+        try {
+          loadBalancing.deleteRule(new DeleteRuleRequest(ruleArn: rule.ruleArn))
+        } catch (AmazonServiceException ignore) {}
```
robzienert: Makes sense to swallow the exception, but would like to at least have a comment in the catch body explaining why we don't care.

PR #2217, opened 2017-12-08
feat(provider/appengine): allow gcloud version to be configurable

danielpeach:
https://github.com/spinnaker/spinnaker/issues/2130 

Needs Halyard support.

### DIFF
--- a/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/config/AppengineConfigurationProperties.groovy
+++ b/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/config/AppengineConfigurationProperties.groovy
@@ -42,6 +42,7 @@ class AppengineConfigurationProperties {
     String sshPrivateKeyPassphrase
     String sshKnownHostsFilePath
     boolean sshTrustUnknownHosts
+    GcloudReleaseTrack gcloudReleaseTrack
 
     void initialize(AppengineJobExecutor jobExecutor) {
       if (this.jsonPath) {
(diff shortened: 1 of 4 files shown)


clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/config/AppengineConfigurationProperties.groovy
```diff
@@ -84,6 +85,11 @@ class AppengineConfigurationProperties {
     static String responseToString(Response response) {
       new String(((TypedByteArray) response.body).bytes)
     }
+
+    static enum GcloudVersion {
```
jtk54: If you manage applications that require different versions of the library, do you need multiple clouddrivers that are configured differently?

E.g. one python application `gcloud app deploy ...` and one java application `gcloud beta app deploy ...`?

danielpeach: Good question - it's per account, so you wouldn't need different Clouddrivers. An account maps to an App Engine application (a single service or small set of services), which is why this feels like the right granularity to me.

jtk54: SGTM.

lwander:
Doesn't gcloud have a proper version? What about gcloudReleaseTrack?

danielpeach:
Yep, that is better.

PR #2218, opened 2017-12-09
feat(artifacts): sets up artifact module

lwander:
My plan was to reuse the artifact work in appengine, but it would have
been too messy to refactor. It didn't expose any artifact-centric interface or
allow easy integration with other artifact sources without a fair bit of
extra work, so I rewrote in a separate module.

The appengine GCS integration was left unchanged (with
the small change of moving the storage utils into a 'AritifactUtils'
class).


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy
```diff
@@ -55,16 +56,18 @@ import java.security.Security
   DeployConfiguration,
   RetrofitConfig,
   AwsConfiguration,
+  ArtifactConfiguration,
```
jtk54: This is in the list twice.

lwander: Whoops thanks

PR #2219, opened 2017-12-10
feat(provider/kubernetes): init containers

ethanfrogers:
adds support for initContainer property of the Pod spec

@lwander can i get some feedback on this? I went with a separate `KubernetesInitContainerDescription` because they don't support readiness/liveness probes or lifecycle hooks. i'm also not totally stoked on the `toInitContainer` method creating a new container and passing it on. I'd rather cast to a container if possible. Any thoughts?

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/api/KubernetesApiConverter.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/api/KubernetesApiConverter.groovy
@@ -693,6 +693,10 @@ class KubernetesApiConverter {
       fromContainer(it)
     } ?: []
 
+    deployDescription.initContainers = replicaSet?.spec?.template?.spec?.initContainers?.collect {
+      fromContainer(it)
+    } ?: []
+
     deployDescription.terminationGracePeriodSeconds = replicaSet?.spec?.template?.spec?.terminationGracePeriodSeconds
     deployDescription.serviceAccountName = replicaSet?.spec?.template?.spec?.serviceAccountName
 
@@ -746,6 +750,10 @@ class KubernetesApiConverter {
     } ?: []
 
     deployDescription.containers = replicationController?.spec?.template?.spec?.containers?.collect {
+      fromContainer(it)
+    } ?: []
+
+    deployDescription.initContainers = replicationController?.spec?.template?.spec?.initContainers?.collect {
       fromContainer(it)
     } ?: []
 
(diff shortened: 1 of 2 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/api/KubernetesApiConverter.groovy
```diff
@@ -240,6 +240,23 @@ class KubernetesApiConverter {
     return handlerBuilder.build()
   }
 
+  static Container toInitContainer(KubernetesInitContainerDescription initContainer) {
+    return toContainer(new KubernetesContainerDescription(
```
lwander: Is an init container always a strict subset of a regular container in terms of the properties it supports?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/description/servergroup/DeployKubernetesAtomicOperationDescription.groovy
```diff
@@ -113,6 +114,28 @@ class KubernetesContainerDescription {
   KubernetesSecurityContext securityContext
 }
 
+@AutoClone
+@Canonical
+class KubernetesInitContainerDescription {
```
lwander: The the above answer is 'yes', maybe we can reflect that here by having the regular container definition inherit from here.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/api/KubernetesApiConverter.groovy
ethanfrogers: It is. See `initContainers` section [here](https://kubernetes.io/docs/api-reference/v1.8/#podspec-v1-core).

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/description/servergroup/DeployKubernetesAtomicOperationDescription.groovy
ethanfrogers: Yeah, that would be good. Do you think having something like this would okay?

```
class KubernetesBaseContainerDescription {
...
}

class KubernetesInitContainerDescription extends KubernetesBaseContainerDescription{}
class KubernetesContainerDescription extends KubernetesBaseContainerDescription{
// extra properies
}
```

ethanfrogers: or maybe the extra layer isn't necessary...

ethanfrogers:
@lwander this is ready to go!

PR #2220, opened 2017-12-11
chore(dependencies): Bump spinnaker-dependencies version.

jtk54:
Picks up the new compute library versions.


jtk54:
FYI @lwander too.

PR #2223, opened 2017-12-12
feat(provider/kubernetes): deploy from artifact

lwander:
With this you can now deploy from manifest & docker changes in the same pipeline
db2wsdewfss


lwander:
https://github.com/spinnaker/spinnaker/issues/2160

PR #2231, opened 2017-12-14
feat(artifacts): Download artifacts from GitHub

andrewbackes:
Add GitHub as a resource to download artifacts from.

@lwander


lwander:
https://github.com/spinnaker/spinnaker/issues/2161

PR #2241, opened 2017-12-19
feat(provider/appengine): enable flex deployments

sbwsg:
Allow App Engine deployments with a built container image URL and app.yaml.

### DIFF
--- a/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperation.groovy
+++ b/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperation.groovy
@@ -53,21 +53,32 @@ class DeployAppengineAtomicOperation implements AtomicOperation<DeploymentResult
 
   DeployAppengineDescription description
   boolean usesGcs
+  boolean containerDeployment
 
   DeployAppengineAtomicOperation(DeployAppengineDescription description) {
     this.description = description
-    this.usesGcs = description.repositoryUrl.startsWith("gs://")
+    this.containerDeployment = description.containerImageUrl?.trim()
+    this.usesGcs = !this.containerDeployment && description.repositoryUrl.startsWith("gs://")
   }
 
   /**
    * curl -X POST -H "Content-Type: application/json" -d '[ { "createServerGroup": { "application": "myapp", "stack": "stack", "freeFormDetails": "details", "repositoryUrl": "https://github.com/organization/project.git", "branch": "feature-branch", "credentials": "my-appengine-account", "configFilepaths": ["app.yaml"] } } ]' "http://localhost:7002/appengine/ops"
    * curl -X POST -H "Content-Type: application/json" -d '[ { "createServerGroup": { "application": "myapp", "stack": "stack", "freeFormDetails": "details", "repositoryUrl": "https://github.com/organization/project.git", "branch": "feature-branch", "credentials": "my-appengine-account", "configFilepaths": ["app.yaml"], "promote": true, "stopPreviousVersion": true } } ]' "http://localhost:7002/appengine/ops"
    * curl -X POST -H "Content-Type: application/json" -d '[ { "createServerGroup": { "application": "myapp", "stack": "stack", "freeFormDetails": "details", "repositoryUrl": "https://github.com/organization/project.git", "branch": "feature-branch", "credentials": "my-appengine-account", "configFilepaths": ["runtime: python27\napi_version: 1\nthreadsafe: true\nmanual_scaling:\n  instances: 5\ninbound_services:\n - warmup\nhandlers:\n - url: /.*\n   script: main.app"],} } ]' "http://localhost:7002/appengine/ops"
+   * curl -X POST -H "Content-Type: application/json" -d '[ { "createServerGroup": { "application": "myapp", "stack": "stack", "freeFormDetails": "details", "credentials": "my-appengine-account", "containerImageUrl": "gcr.io/my-project/my-image:my-tag", "configFiles": ["env: flex\nruntime: custom\nmanual_scaling:\n  instances: 1\nresources:\n  cpu: 1\n  memory_gb: 0.5\n  disk_size_gb: 10"] } } ]' "http://localhost:7002/appengine/ops"
    */
   @Override
   DeploymentResult operate(List priorOutputs) {
     def  baseDir = description.credentials.localRepositoryDirectory
-    def  directoryPath = getFullDirectoryPath(baseDir, description.repositoryUrl)
+
+    String directoryId
+    if (containerDeployment) {
+      directoryId = description.containerImageUrl
+    } else {
+      directoryId = description.repositoryUrl
+    }
+
+    def directoryPath = getFullDirectoryPath(baseDir, directoryId)
 
     /*
     * We can't allow concurrent deploy operations on the same local repository.
@@ -77,12 +88,28 @@ class DeployAppengineAtomicOperation implements AtomicOperation<DeploymentResult
     return AppengineMutexRepository.atomicWrapper(directoryPath, {
       task.updateStatus BASE_PHASE, "Initializing creation of version..."
       def result = new DeploymentResult()
-      def newVersionName = deploy(cloneOrUpdateLocalRepository(directoryPath, 1))
+      String newVersionName
+      if (containerDeployment) {
+        createEmptyDirectory(directoryPath)
+        newVersionName = deploy(directoryPath)
+      } else {
+        newVersionName = deploy(cloneOrUpdateLocalRepository(directoryPath, 1))
+      }
       def region = description.credentials.region
       result.serverGroupNames = Arrays.asList("$region:$newVersionName".toString())
       result.serverGroupNameByRegion[region] = newVersionName
       return result
     })
+  }
+
+  void createEmptyDirectory(String path) {
+    File directory = new File(path)
+    if (directory.exists()) {
+      directory.deleteDir()
+    }
+    if (!directory.mkdirs()) {
+      throw new AppengineOperationException("Failed to create directory: $path")
+    }
   }
 
   String cloneOrUpdateLocalRepository(String directoryPath, Integer retryCount) {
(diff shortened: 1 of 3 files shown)


clouddriver-appengine/src/test/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperationSpec.groovy
```diff
@@ -0,0 +1,48 @@
...
+        env: flex
+        runtime: python27
+        manual_scaling:
+          instances: 1
+        resources:
+          cpu: 1
+          memory_gb: 0.5
+          disk_size_gb: 10
+        '''
+
```
danielpeach: Extraneous space

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperation.groovy
```diff
@@ -53,21 +53,32 @@ class DeployAppengineAtomicOperation implements AtomicOperation<DeploymentResult
 
   DeployAppengineDescription description
   boolean usesGcs
+  boolean isContainerDeployment
```
lwander: I'd skip the `is` here - the auto generated getters will add `is` automatically (`op.isGeneratedDeployment()` in non-groovy code).

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperation.groovy
```diff
@@ -77,14 +88,30 @@ class DeployAppengineAtomicOperation implements AtomicOperation<DeploymentResult
     return AppengineMutexRepository.atomicWrapper(directoryPath, {
       task.updateStatus BASE_PHASE, "Initializing creation of version..."
       def result = new DeploymentResult()
-      def newVersionName = deploy(cloneOrUpdateLocalRepository(directoryPath, 1))
+      String newVersionName
+      if (this.isContainerDeployment) {
```
lwander: omit `this` here

clouddriver-appengine/src/test/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperationSpec.groovy
```diff
@@ -0,0 +1,48 @@
...
+      ]
+      List<String> results
```
danielpeach: Would just have `List<String> results = DeployAppengineAtomicOperation.updateConfigRuntime(configs)` unless Spock for some reason is unhappy with that.

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperation.groovy
```diff
@@ -53,21 +53,32 @@ class DeployAppengineAtomicOperation implements AtomicOperation<DeploymentResult
...
+   * curl -X POST -H "Content-Type: application/json" -d '[ { "createServerGroup": { "application": "myapp", "stack": "stack", "freeFormDetails": "details", "credentials": "my-appengine-account", "containerImageUrl": "gcr.io/my-project/my-image:my-tag", "configFiles": ["env: flex\nruntime: custom\nmanual_scaling:\n  instances: 1\nresources:\n  cpu: 1\n  memory_gb: 0.5\n  disk_size_gb: 10"] } } ]' "http://localhost:7002/appengine/ops"
    */
   @Override
   DeploymentResult operate(List priorOutputs) {
     def  baseDir = description.credentials.localRepositoryDirectory
-    def  directoryPath = getFullDirectoryPath(baseDir, description.repositoryUrl)
+
+    String directoryId
+    if (this.isContainerDeployment) {
```
lwander: You can omit `this` here, because there is no ambiguity without it.

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperation.groovy
```diff
@@ -171,6 +206,16 @@ class DeployAppengineAtomicOperation implements AtomicOperation<DeploymentResult
     return versionName
   }
 
+  static List<String> updateConfigRuntime(List<String> configFiles) {
+    if (!configFiles) {
+      return []
+    } else {
+      return configFiles.collect { configFile ->
+        return configFile.replaceAll(/runtime\s*\:\s*[a-zA-Z]+/, "runtime: custom")
```
lwander: Shouldn't the user be responsible for submitting the correct runtimes?

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperation.groovy
```diff
@@ -158,6 +190,9 @@ class DeployAppengineAtomicOperation implements AtomicOperation<DeploymentResult
     deployCommand << (description.stopPreviousVersion ? "--stop-previous-version": "--no-stop-previous-version")
     deployCommand << "--project=$project"
     deployCommand << "--account=$accountEmail"
+    if (this.isContainerDeployment) {
```
lwander: omit `this`

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperation.groovy
```diff
@@ -145,7 +172,12 @@ class DeployAppengineAtomicOperation implements AtomicOperation<DeploymentResult
                                                                          description.stack,
                                                                          description.freeFormDetails,
                                                                          false)
-    def writtenFullConfigFilePaths = writeConfigFiles(description.configFiles, repositoryPath, applicationDirectoryRoot)
+    def imageUrl = description.containerImageUrl
+    def configFiles = description.configFiles
+    if (this.isContainerDeployment) {
```
lwander: omit `this`

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperation.groovy
danielpeach: I would drop `this` unless it's disambiguating method params from class members (here and a few spots below).

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperation.groovy
danielpeach: Seems pretty unlikely, but an app.yaml can have arbitrary user-supplied values (e.g. env variables), so it's possible you could accidentally overwrite something here.

sbwsg: `gcloud app deploy` verifies that source code exists in the current directory and validates that it matches the app.yaml's runtime.  This validation runs even when deploying from an image URL, which is odd since in this case the local source code is never actually used for anything.  This seemed like the easiest workaround for gcloud's slightly errant behaviour.

With all that said, it doesn't really answer your question.  I guess we could allow the user to submit whatever runtime they want but could warn them that any runtime except custom will fail.  Or we could disallow form submission with any runtime other than custom.

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/validators/DeployAppengineDescriptionValidator.groovy
```diff
@@ -39,20 +39,22 @@ class DeployAppengineDescriptionValidator extends DescriptionValidator<DeployApp
...
+
+    if (!isContainerDeployment && !description.repositoryUrl.startsWith("gs://")) {
       if (!helper.validateGitCredentials(description.credentials.gitCredentials,
                                          description.gitCredentialType,
                                          description.credentials.name,
                                          "gitCredentialType")) {
          return
       }
       helper.validateNotEmpty(description.branch, "branch")
+      helper.validateNotEmpty(description.repositoryUrl, "repositoryUrl")
```
danielpeach: I think this should live outside this `if`, since the `repositoryUrl` validation won't happen in the case where `repositoryUrl.startsWith("gs://")`.

Also, in the case of `isContainerDeployment`, there should also be a check for `helper.validateNotEmpty(description.containerImageUrl)`.

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperation.groovy
sbwsg: OK, i've attempted to make the runtime regex a little more strict, and will only replace the first instance of the string.

Is there a more thorough approach or is this ok for now?

lwander: I would lean towards failing with incorrect runtimes - my (limited) understanding is that writing a custom runtime requires extra work, and switching it for a user without them knowing might cause strange breakages.

danielpeach: `runtime: custom` is essentially a confirmation to gcloud that the user wants to deploy their own container. The other runtimes are used when gcloud builds a new container - which is the old codepath, not this one. i.e., if the user supplied a container image, they definitely want `runtime: custom`.

danielpeach: (Though I think Scott is removing this anyway)

sbwsg: Removed app.yaml update code completely.  We can message and enforce better in deck for now.

PR #2249, opened 2017-12-21
fix(clusters): Return all matching server groups for a provider

lwander:


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ClusterController.groovy
```diff
@@ -155,24 +155,26 @@ class ClusterController {
                      @PathVariable String serverGroupName,
                      @RequestParam(value = "region", required = false) String region) {
     // we can optimize loads iff the cloud provider supports loading minimal clusters (ie. w/o instances)
-    def clusterProvider = clusterProviders.find { it.cloudProviderId == type }
-    if (!clusterProvider) {
+    def providers = clusterProviders.find { it.cloudProviderId == type }
```
jtk54: Meant to be a `findAll`?

lwander: Whoops yup

PR #2251, opened 2017-12-21
fix(provider/aws): handle spotPrice ==  in rollingpush

emjburns:
Correctly handling spotPrice in a rollingpush strategy. @duftler is there any reason why you left out validation logic in this class? Want to make sure I'm not breaking anything around spotPrice!

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ModifyAsgLaunchConfigurationOperation.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ModifyAsgLaunchConfigurationOperation.groovy
@@ -78,6 +78,10 @@ class ModifyAsgLaunchConfigurationOperation implements AtomicOperation<Void> {
     props = props + description.properties.findResults { k, v -> (v != null && settingsKeys.contains(k)) ? [k, v] : null }.collectEntries()
     props.remove('class')
 
+    if (props.spotPrice == "") {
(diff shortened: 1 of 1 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ModifyAsgLaunchConfigurationOperation.groovy
```diff
@@ -78,6 +78,10 @@ class ModifyAsgLaunchConfigurationOperation implements AtomicOperation<Void> {
     props = props + description.properties.findResults { k, v -> (v != null && settingsKeys.contains(k)) ? [k, v] : null }.collectEntries()
     props.remove('class')
 
+    if (props.spotPrice == "") {
+      // Handle the case where you previously set a spot price and want to unset it
```
ajordens: comment might be a tad clearer as

`spotPrice of "" indicates that it should be removed regardless of value on source launch configuration`

?

ajordens: otherwise looks mergeable to me

emjburns: Definitely better wording. Changing then merging.

PR #2257, opened 2017-12-28
bug(provider/openstack) - allow for binding to multiple loadbalancers…

johnrengelman:
… with the same port mappings.


johnrengelman:
@spinnaker/reviewers

PR #2264, opened 2018-01-02
feat(provider/ecs): ECS Loadbalancer cache client and tests.

dkirillov:
Adds an EcsLoadbalancerCacheClient which  allows providers/services to interface more easily with the already existing AWS loadbalancer cache.

### DIFF
--- a/clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/cache/EcsLoadbalancerCacheClientSpec.groovy
+++ b/clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/cache/EcsLoadbalancerCacheClientSpec.groovy
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2017 Lookout, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"):
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/cache/EcsLoadbalancerCacheClientSpec.groovy
```diff
@@ -0,0 +1,84 @@
...
+      vpcId: vpcId,
+      dnsname: 'dns-name',
+      createdTime: System.currentTimeMillis(),
+      subnets: [],
+      securityGroups: [],
+      targetGroups: [targetGroupName],
+      serverGroups: []
+    )
+
+    objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
robzienert: I feel like this should be added into the instantiation of the class at the top of the spec:

```
def objectMapper = new ObjectMapper().apply {
  configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
  return it
}
```

dkirillov: Done.

Couldn't use `apply` though, used `configure` directly which returns `ObjectMapper`.

PR #2266, opened 2018-01-04
fix(provider/aws): block device config for missing instance types

ajordens:
This should provide coverage for all currently available instance types.

Will look for better ways to manage this moving forward.


ajordens:
I have an uncommitted agent that checks the aws pricing "API" (https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonEC2/current/us-east-1/index.json) and compares the available instance types against `BlockDeviceConfig`. 

This PR addresses the deltas.

It's possible we could move away from `BlockDeviceConfig` and instead cache instance type details from the pricing api. It gets a little tricky with EBS-only instance types as we would need to apply sensible and consistent defaults.

For now, I plan to commit a Netflix-specific validation agent that will alert us when new instance types appear. 

Thoughts?

cfieber:
seems reasonable to me (alerting agent)

PR #2271, opened 2018-01-04
feat(provider/kubernetes): fail deploy if artifacts don't bind

lwander:
nsoqiftzmh6


lwander:
https://github.com/spinnaker/spinnaker/issues/2184

PR #2274, opened 2018-01-05
fix(provider/docker): update bearer_token to access_token

mike10010100:
Per https://docs.docker.com/registry/spec/auth/token/#requesting-a-token, the token response fields will either contain `token` or `access_token`.

In ACR, for example, only access_token is sent back as a response, meaning that the response from Spinnaker will fail to authenticate.

### DIFF
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerToken.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerToken.groovy
@@ -20,9 +20,9 @@ import groovy.transform.ToString
 
 @ToString(includeNames = true)
 class DockerBearerToken {
-  // One of token or bearer_token will be filled by the request.
+  // One of token or access_token will be filled by the request.
   String token
-  String bearer_token
+  String access_token
   int expires_in
   String issued_at
 }
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerToken.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerToken.groovy
@@ -20,9 +20,10 @@ import groovy.transform.ToString
 
 @ToString(includeNames = true)
 class DockerBearerToken {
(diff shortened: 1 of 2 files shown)


clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/auth/DockerBearerToken.groovy
```diff
@@ -20,9 +20,9 @@ import groovy.transform.ToString
 
 @ToString(includeNames = true)
 class DockerBearerToken {
-  // One of token or bearer_token will be filled by the request.
+  // One of token or access_token will be filled by the request.
   String token
-  String bearer_token
+  String access_token
```
lwander: While I'm not positive, there's a chance that `bearer_token` is used by some system (albeit incorrectly). Could you keep it in, and read it only when `token` & `access_token` are empty?

mike10010100: Certainly.

lwander: Thanks :)

mike10010100: Sorry, not particularly well versed in Groovy's Elvis operators, does the change make sense?

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -391,7 +391,7 @@ class DockerRegistryClient {
       DockerBearerToken dockerToken = tokenService.getToken(target)
       String token
       if (dockerToken) {
-        token = "Bearer ${dockerToken.access_token ?: dockerToken.token}"
+        token = "Bearer ${(dockerToken.bearer_token ?: dockerToken.token) ?: dockerToken.access_token}"
```
mike10010100: @lwander I'm not super familiar with Groovy's Elvis operators, does this make sense?

hafizullah:
The last build have failed, have this PR made into any of the releases? and I how can I consume it?

PR #2275, opened 2018-01-05
fix(provider/docker): update bearer_token to access_token (#2274)

mike10010100:
Per https://docs.docker.com/registry/spec/auth/token/#requesting-a-token, the token response fields will either contain `token` or `access_token`.

In ACR, for example, only access_token is sent back as a response, meaning that the response from Spinnaker will fail to authenticate.


lwander:
Thanks!

PR #2279, opened 2018-01-06
fix(provider/kubernetes): v2 Set default namespace dynamically

wjoel:
Set the default namespace according to the `kubectl` rules defined in
https://github.com/kubernetes/kubernetes/blob/bd4d511a40e142ee65edff3b286e57de502aa790/pkg/kubectl/cmd/util/factory_client_access.go#L127

In short, the default namespace is set in this way:

- If `/var/run/secrets/kubernetes.io/serviceaccount/namespace` exists,
  assume we are running in a Kubernetes container, and set the default
  namespace to the contents of this file.
- Set the default namespace to the namespace of current `kubectl`
  config context, if it exists. `kubectl config view` is used to get
  this namespace, which takes care of the complicated merges and
  overrides.
- Otherwise, set the default namespace to "default".

This fixes the issue of manifests without a namespace specification
being deployed in the namespace of clouddriver when Spinnaker has been
deployed in Kubernetes.

Previously manifests without a namespace were assumed to belong to the
"default" namespace, but clouddriver would deploy them in the
"spinnaker" namespace, where it is deployed by default by Halyard.
Deploying such manifests would cause Spinnaker to get stuck waiting
for a deployment to show up in the "default" namespace, since
clouddriver would instead have created a deployment in the "spinnaker"
namespace.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifest.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifest.java
@@ -86,7 +86,7 @@ public class KubernetesManifest extends HashMap<String, Object> {
   @JsonIgnore
   public String getNamespace() {
     String namespace = (String) getMetadata().get("namespace");
-    return StringUtils.isEmpty(namespace) ? "default" : namespace;
+    return StringUtils.isEmpty(namespace) ? "" : namespace;
   }
 
   @JsonIgnore
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
@@ -34,10 +34,12 @@ import org.apache.commons.lang3.StringUtils;
 import javax.validation.constraints.NotNull;
 import java.io.FileNotFoundException;
 import java.io.FileReader;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.*;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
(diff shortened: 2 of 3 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifest.java
```diff
@@ -81,7 +81,7 @@ public void setName(String name) {
   @JsonIgnore
   public String getNamespace() {
     String namespace = (String) getMetadata().get("namespace");
-    return StringUtils.isEmpty(namespace) ? "default" : namespace;
+    return StringUtils.isEmpty(namespace) ? "" : namespace;
```
lwander: I think some caching logic assumes `getNamespace` won't be empty - not entirely sure though. Did anything funny happen when you ran this?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -67,8 +69,21 @@
...
   private final String defaultNamespace = "default";
+  private final Path serviceAccountNamespacePath = Paths.get("/var/run/secrets/kubernetes.io/serviceaccount/namespace");
+  public String getDefaultNamespace() {
+    String namespace = defaultNamespace;
+    try {
+      Optional<String> serviceAccountNamespace = Files.lines(serviceAccountNamespacePath, StandardCharsets.UTF_8).findFirst();
+      namespace = serviceAccountNamespace.get();
+    } catch (IOException | NoSuchElementException e) {
+      try {
+        namespace = jobExecutor.defaultNamespace(this);
```
lwander: What happens when this is empty? (Seems like it can be [here](https://github.com/spinnaker/clouddriver/pull/2279/files#diff-65ee14f962603a2d9dccefbbc331e68cR94))

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifest.java
wjoel: Nope, but I only tested it on a very tiny Kubernetes cluster, so I might not have seen it. I'll try it out at work tomorrow and see how it goes.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
wjoel: Yes, good catch. I refactored this a bit too quickly. Fixed.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifest.java
wjoel: Nothing funny is happening with the cluster at work either, as far as I can tell. No surprises in any logs.

lwander: SGTM, let's merge and see what happens!

PR #2280, opened 2018-01-08
fix(jobs): Fix bottleneck around JobLocalExecutor.startJob

lwander:
The `startJob` method submits your request to a threadpool, which it
immediately starts polling. This isn't useful, and limits the number of
jobs that can be started to the number of threads supported by your
scheduler. As a result, submitting a lot of jobs at once causes
(needlessly) a lot of these jobs to timeout.

There is still a `sleep(500)` which doesn't seem to add anything in the
function, but didn't feel confident removing since other spots in
clouddriver make use of it.

@danielpeach I added you to that todo because I think appengine is 
the only other big consumer of this

https://github.com/spinnaker/spinnaker/issues/2232


lwander:
thanks - I'm going to wait on @duftler to be safe

lwander:
@duftler says "ok"

PR #2282, opened 2018-01-09
feat(artifacts): add artifact download endpoint

lwander:
https://github.com/spinnaker/spinnaker/issues/2235

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ArtifactController.java
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ArtifactController.java
@@ -18,13 +18,19 @@
 package com.netflix.spinnaker.clouddriver.controllers;
 
 import com.netflix.spinnaker.clouddriver.artifacts.ArtifactCredentialsRepository;
+import com.netflix.spinnaker.clouddriver.artifacts.ArtifactDownloader;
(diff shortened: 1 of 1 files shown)


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ArtifactController.java
```diff
@@ -48,4 +57,25 @@ public ArtifactController(Optional<ArtifactCredentialsRepository> artifactCreden
...
+    }
+
+    Artifact artifact = Artifact.builder()
+        .type(type)
+        .artifactAccount(artifactAccount)
+        .reference(reference)
+        .build();
+
+    try {
+      return IOUtils.toString(artifactDownloader.download(artifact));
```
ajordens: (just noticed this while cutting a release)

Thoughts on just sending the underlying stream all the way across to the caller without first converting to a `String`?

Similar to what was done here: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/DataController.groovy#L55

lwander: Oh sweet, didn't know that was possible - I'll switch that over.

PR #2285, opened 2018-01-10
feat(provider/aws) Fetch image information based on image id

PerGon:
This PR adds functionality to get an image based on the id of the image. Currently, only AWS provider is implemented.

Based on an image id (`ami-123321`), it would be possible to call `images/aws/ami-123321` and retrieve the information about such image - including the server groups and instances where the image might be used.

Usefulness of such endpoint:
- Cleaning up images. After using spinnaker for some time and baking a lot of images, these images start to pile up and increase costs and resource usage. Being able to know if a certain image is being used or not, is useful to make a decision on either delete it or not.
- Security. At some point, a certain image can be discovered to have been compromised - having a quick way to identify which server groups/instances are using such image might be useful.

For reference, here's a result of calling such endpoint:
```
{
  "artifactAccount" : null,
  "location" : "123456789/eu-west-1",
  "metadata" : {
    "architecture" : "x86_64",
    "blockDeviceMappings" : [ {
      "deviceName" : "/dev/sda1",
      "ebs" : {
        "deleteOnTermination" : true,
        "encrypted" : false,
        "snapshotId" : "snap-0ab7091111111e036",
        "volumeSize" : 8,
        "volumeType" : "gp2"
      }
    }, {
      "deviceName" : "/dev/sdb",
      "virtualName" : "ephemeral0"
    }, {
      "deviceName" : "/dev/sdc",
      "virtualName" : "ephemeral1"
    } ],
    "creationDate" : "2017-10-12T20:08:47.000Z",
    "enaSupport" : true,
    "hypervisor" : "xen",
    "imageId" : "ami-f531111c",
    "imageLocation" : "811111117767/my_app-amd64-20171012200143-xenial",
    "imageType" : "machine",
    "name" : "my_app-amd64-20171012200143-xenial",
    "ownerId" : "811111117767",
    "productCodes" : [ ],
    "public" : false,
    "rootDeviceName" : "/dev/sda1",
    "rootDeviceType" : "ebs",
    "serverGroups" : [ {
      "asg" : {
        "autoScalingGroupARN" : "arn:aws:autoscaling:eu-west-1:411111117094:autoScalingGroup:b23900e2-e127-1111-1111-38d16b970f81:autoScalingGroupName/my_app-v025",
        "autoScalingGroupName" : "my_app-v025",
        "availabilityZones" : [ "eu-west-1a" ],
        "createdTime" : 1507841610647,
        "defaultCooldown" : 10,
        "desiredCapacity" : 1,
        "enabledMetrics" : [ ],
        "healthCheckGracePeriod" : 600,
        "healthCheckType" : "EC2",
        "instances" : [ {
          "availabilityZone" : "eu-west-1a",
          "healthStatus" : "Healthy",
          "instanceId" : "i-04874711111111123",
          "launchConfigurationName" : "my_app-v025-10122017205329",
          "lifecycleState" : "InService",
          "protectedFromScaleIn" : false
        } ],
        "launchConfigurationName" : "my_app-v025-10122017205329",
        "loadBalancerNames" : [ "my_app-elb" ],
        "maxSize" : 1,
        "minSize" : 1,
        "newInstancesProtectedFromScaleIn" : false,
        "suspendedProcesses" : [ {
          "processName" : "AZRebalance",
          "suspensionReason" : "User suspended at 2017-10-12T20:53:30Z"
        } ],
        "tags" : [ ],
        "targetGroupARNs" : [ ],
        "terminationPolicies" : [ "Default" ],
        "vpczoneIdentifier" : "subnet-81111dd2"
      },
      "instances" : [ {
        "availabilityZone" : "eu-west-1a",
        "healthStatus" : "Healthy",
        "instanceId" : "i-04874711111111123",
        "launchConfigurationName" : "my_app-v025-10122017205329",
        "lifecycleState" : "InService",
        "protectedFromScaleIn" : false
      } ],
      "launchConfigName" : "my_app-v025-10122017205329",
      "name" : "my_app-v025",
      "region" : "eu-west-1",
      "scalingPolicies" : [ ],
      "scheduledActions" : [ ],
      "targetGroups" : [ ],
      "vpcId" : "vpc-c91111ad",
      "zones" : [ "eu-west-1a" ]
    } ],
    "sriovNetSupport" : "simple",
    "state" : "available",
    "tags" : [ {
      "key" : "build_host",
      "value" : "https://travis.company.io/"
    }, {
      "key" : "build_info_url",
      "value" : "https://travis.company.io/infrastructure/my_app/builds/2059959"
    }, {
      "key" : "appversion",
      "value" : "my_app-0.0.8-h92.b195a5f3d9dc6627e06f40779cdb3d3d36fe0ba0/infrastructure-my_app/92"
    } ],
    "virtualizationType" : "hvm"
  },
  "name" : "my_app-amd64-20171012200143-xenial",
  "provenance" : null,
  "reference" : "ami-f531111c",
  "type" : "aws/image",
  "uuid" : null,
  "version" : null
}
```

### DIFF
--- a/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonImageProviderSpec.groovy
+++ b/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonImageProviderSpec.groovy
@@ -16,11 +16,12 @@
 
 package com.netflix.spinnaker.clouddriver.aws.provider.view
 
+import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.cats.cache.Cache
 import com.netflix.spinnaker.cats.cache.CacheData
 import com.netflix.spinnaker.cats.cache.DefaultCacheData
+import com.netflix.spinnaker.clouddriver.aws.AwsConfiguration
 import com.netflix.spinnaker.clouddriver.aws.model.AmazonImage
-import com.netflix.spinnaker.kork.artifacts.model.Artifact
 import spock.lang.Specification
 import spock.lang.Subject
 
@@ -29,36 +30,34 @@ import com.netflix.spinnaker.clouddriver.aws.data.Keys
 
 class AmazonImageProviderSpec extends Specification {
   Cache cache = Mock(Cache)
+  AwsConfiguration.AmazonServerGroupProvider amazonServerGroupProvider = Mock(AwsConfiguration.AmazonServerGroupProvider)
+  ObjectMapper objectMapper = new ObjectMapper()
 
   @Subject
-  AmazonImageProvider provider = new AmazonImageProvider(cache)
+  AmazonImageProvider provider = new AmazonImageProvider(cache, amazonServerGroupProvider, objectMapper)
 
   void "should return one image"() {
-
     when:
     def result = provider.getImageById("ami-123321")
 
     then:
-    result == Optional.of(Artifact.builder()
-            .name("some_ami")
-            .type(AmazonImage.AMAZON_IMAGE_TYPE)
-            .location("123456789/eu-west-1")
-            .metadata("name": "some_ami", ownerId: '123456789', "serverGroups": [], "imageId": "ami-123321", "region": "eu-west-1", "account": "test_account")
-            .reference("ami-123321")
-            .build())
+    AmazonImage expectedImage = new AmazonImage()
+    expectedImage.setRegion("eu-west-1")
+    expectedImage.setId("ami-123321")
+    expectedImage.setImageId("ami-123321")
+    expectedImage.setOwnerId("1233211233231")
+    result == Optional.of(expectedImage)
 
     and:
     1 * cache.filterIdentifiers(IMAGES.ns, _ as String) >> [
-            "aws:images:test_account:eu-west-1:ami-123321"
+        "aws:images:test_account:eu-west-1:ami-123321"
     ]
 
     1 * cache.getAll(IMAGES.ns, ["aws:images:test_account:eu-west-1:ami-123321"]) >>
-            [imageCacheData('aws:images:test_account:eu-west-1:ami-123321', [
-                    account: 'test_account',
-                    region : 'eu-west-1',
-                    name   : 'some_ami',
-                    ownerId: '123456789',
-                    imageId: 'ami-123321'])]
+        [imageCacheData('aws:images:test_account:eu-west-1:ami-123321', [
+            ownerId: '1233211233231',
+            name   : 'some_ami',
+            imageId: 'ami-123321'])]
   }
 
   void "should not find any image"() {
--- a/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonImageProviderSpec.groovy
+++ b/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonImageProviderSpec.groovy
@@ -42,8 +42,8 @@ class AmazonImageProviderSpec extends Specification {
     result == Optional.of(Artifact.builder()
             .name("some_ami")
             .type(AmazonImage.AMAZON_IMAGE_TYPE)
-            .location("test_account/eu-west-1")
-            .metadata("name": "some_ami", "serverGroups": [], "imageId": "ami-123321", "region": "eu-west-1", "account": "test_account")
+            .location("123456789/eu-west-1")
+            .metadata("name": "some_ami", ownerId: '123456789', "serverGroups": [], "imageId": "ami-123321", "region": "eu-west-1", "account": "test_account")
             .reference("ami-123321")
             .build())
 
@@ -57,6 +57,7 @@ class AmazonImageProviderSpec extends Specification {
                     account: 'test_account',
                     region : 'eu-west-1',
                     name   : 'some_ami',
+                    ownerId: '123456789',
                     imageId: 'ami-123321'])]
   }
 
--- a/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonImageProviderSpec.groovy
+++ b/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonImageProviderSpec.groovy
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2018 Schibsted ASA.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonImageProviderSpec.groovy
```diff
@@ -0,0 +1,83 @@
...
+import com.netflix.spinnaker.cats.cache.DefaultCacheData
+import com.netflix.spinnaker.clouddriver.aws.model.AmazonImage
+import com.netflix.spinnaker.kork.artifacts.model.Artifact
+import spock.lang.Specification
+import spock.lang.Subject
+
+import static com.netflix.spinnaker.clouddriver.core.provider.agent.Namespace.IMAGES
+import com.netflix.spinnaker.clouddriver.aws.data.Keys
+
+class AmazonImageProviderSpec extends Specification {
```
robzienert: Please keep to the 2-space standard.

PerGon:
Thanks! Added licenses and better 2-space formatting. Will squash after I get more feedback :)

jeyrschabu:
Yes, this is relevant to the Swabbie effort.

@PerGon, we are currently working on a service that will automate the cleanup of unused resources and it sounds like it will be of good use to you.

PerGon:
Hi @jeyrschabu - thanks for the comment :) You said this is relevant for Swabbie - relevant in a way that it should be reviewed and merged or in a way that the functionality will leave in the Swabbie code?
Also, any additional information you can provide on the Swabbie effort? It's the first time I see it mentioned :)

jeyrschabu:
@PerGon it's relevant in the sense that adding this endpoint aligns with Swabbie. 

Swabbie is a service that will automate the cleanup of unsused resources based on rules. It's the successor to Janitor Monkey, and It will integrate with Spinnaker to orchestrate the deletion of resources. Still early in the process but I am planning to make the repo public for OSS contributions very soon.

jeyrschabu:
@robzienert lgtm

PerGon:
I've just added one commit that slightly changed the behavior.

At first, I thought that any redis key ending in `ami-123321` would have to be unique, but while testing this internally, I've realized that's not true.
When having multiple AWS accounts, AMIs are shared between accounts - this means the same AMI (e.g. `ami-123321`) will be present in multiple accounts and this will be reflected in redis since the key is something like this:
`aws:images:test_account:eu-west-1:ami-123321`
As I'm checking redis keys with the pattern `*.ami-123321`, this can retrieve multiple keys - one for each account where the ami is present/shared. 
The last commit takes that into consideration.

Can you have another look? :)

jeyrschabu:
At first glance, @PerGon I'm wondering if the endpoint should include the account you are interested in rather than plucking the first key.

PerGon:
(Oh, you just reminded me I needed to fix something :) check the last commit )

So, I think that having to pass the account in the endpoint, is not ideal. An image ID is a unique identifier - it only *belongs*  to one account. It might be shared with a bunch of other accounts (hence multiple entries in the cache), but the owner is only one.
So, if I have a certain image (e.g. `ami-123321`), I don't really care who owns it, but I want to have info about it, including knowing who is the owner :) If I have to pass the account in the endpoint, then, I might not be able to get info about the ami since I might not know the account.

Regarding the multiple entries in cache for the same ami id, we know that the ami id might be "present" in many accounts (when it's shared), but the owner of the ami is always the same (check the last commit :) ) - in fact, the image metadata for the multiple cache entries for the ami is always 
the same. The usefulness of getting all the cache entries for the ami, is that we can iterate on those entries and check if there is any server group or instances using it... if there is a server group / instance using an image, then it's not a candidate for deletion (as an example of why it's useful to have that additional info returned in the endpoint).

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ImageController.java
```diff
@@ -0,0 +1,54 @@
...
+
+@RestController
+@RequestMapping("/images")
+public class ImageController {
+
+  @Autowired
+  List<ImageProvider> imageProviders;
+
+  @RequestMapping(value = "/{provider}/{imageId}", method = RequestMethod.GET)
+  Artifact getImage(@PathVariable String provider, @PathVariable String imageId) {
```
robzienert: I don't think Artifact is the right thing here. We should be returning a native representation of the image, rather than jamming it into a class meant for passing artifact information along in triggers / executions.

robzienert:
Sorry to let this stagnate, one final feedback and then I think we're good here.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ImageController.java
PerGon: Thanks for the feedback!
Oh, I thought that Artifact was used in a more broader context :) I'll have a look and do the change as soon as possible.

PerGon: In order to use the image native representation, I've did a small refactor that can be found in here: https://github.com/spinnaker/clouddriver/pull/2373
`Image` does not exists at the core level. It does in the openstack module with a `TODO` to move to core - so I just executed the TODO :)

PerGon:
Oh, it took a while to get back to this... but I was waiting for a few other PR to get merged to finish this one. The other PR are: https://github.com/spinnaker/clouddriver/pull/2390 and https://github.com/spinnaker/clouddriver/pull/2373
The last commit replaces the initial approach (using the `Artifact` object) to use a native representation of an `Image`. It also uses the new relation between Images and Server Groups to get the Server Groups where the image is being used.
Please have a look. Thanks

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/CloudDriverConfig.groovy
```diff
@@ -207,6 +209,12 @@ class CloudDriverConfig {
     new NoopInstanceProvider()
   }
 
+  @Bean
+  @ConditionalOnMissingBean(ImageProvider)
+  ImageProvider noopImageIdProvider() {
+    new NoopImageProvider()
+  }
```
robzienert: Should name the bean consistently with the class or vice-versa.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ImageController.java
```diff
@@ -0,0 +1,55 @@
...
+  Image getImage(@PathVariable String provider, @PathVariable String imageId) {
+
+    List<ImageProvider> imageProviderList = imageProviders.stream()
+        .filter(imageProvider -> imageProvider.getCloudProvider().equals(provider))
+        .collect(Collectors.toList());
+
+    if (imageProviderList.isEmpty()) {
+      throw new UnsupportedOperationException("ImageProvider for provider " + provider + " not found.");
+    } else if (imageProviderList.size() > 1) {
+      throw new UnsupportedOperationException("Found multiple ImageProviders for provider " + provider + ". Multiple ImageProviders for a single provider are not supported.");
```
robzienert: I don't think this is an UnsupportedOperation so much as an IllegalState.

robzienert:
Thanks!

PR #2288, opened 2018-01-12
feat(provider/kubernetes): adds hook for named per-account configuration

danielpeach:


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/configuration/KubernetesAccountConfigurer.java
```diff
@@ -0,0 +1,44 @@
...
+  @Autowired
+  KubernetesAccountConfigurer(List<KubernetesAccountConfiguration> kubernetesAccountConfigurations) {
+    this.kubernetesAccountConfigurations = kubernetesAccountConfigurations;
+  }
+
+  public void configure(String configurationName, KubernetesNamedAccountCredentials credentials) {
+    Optional<KubernetesAccountConfiguration> configuration =
+      kubernetesAccountConfigurations.stream()
+        .filter(i -> i.handles(configurationName,  credentials.getProviderVersion()))
+        .findFirst();
```
lwander: You can just chain the `ifPresent` call here

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/configuration/KubernetesAccountConfiguration.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package com.netflix.spinnaker.clouddriver.kubernetes.v2.security.configuration;
+
+import com.netflix.spinnaker.clouddriver.kubernetes.security.KubernetesNamedAccountCredentials;
+import com.netflix.spinnaker.clouddriver.security.ProviderVersion;
+
+public interface KubernetesAccountConfiguration {
+  boolean handles(String configurationName, ProviderVersion providerVersion);
+  void configure(KubernetesNamedAccountCredentials credentials);
```
lwander: Seems that all that `configure` does it add a namer in the only implementation - is that the intent? Or will more configuration be added in the future? As it stands I would name this differently, and maybe even structure it differently (I don't see why this needs an interface)

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.groovy
```diff
@@ -32,6 +32,7 @@ class KubernetesConfigurationProperties {
     String cluster
     String oAuthServiceAccount
     List<String> oAuthScopes
+    String configuration
```
lwander: Not a big fan of this name - is this supposed to be the kind of moniker that's applied? If so, maybe call this "namingStrategy"? And have the default one for k8s v2 be "namingStrategy = annotations" or something.

lwander: Also it should probably be something that every account in clouddriver supplies, same way "permissions" is a field everywhere.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesNamedAccountCredentialsInitializer.groovy
```diff
@@ -99,6 +101,7 @@ class KubernetesNamedAccountCredentialsInitializer implements CredentialsInitial
           .debug(managedAccount.debug)
           .build()
 
+        kubernetesAccountConfigurer.configure(managedAccount.configuration, kubernetesAccount)
```
lwander: I guess I'm confused why this was moved out of the "build" into a separate "configure"

danielpeach:
@lwander

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.groovy
```diff
@@ -42,7 +42,8 @@ class KubernetesConfigurationProperties {
     List<LinkedDockerRegistryConfiguration> dockerRegistries
     List<String> requiredGroupMembership
     Permissions.Builder permissions = new Permissions.Builder()
-    Boolean debug = false;
+    String namingStrategy = "annotations"
```
lwander: Sorry let's call it "kubernetesAnnotations" to avoid overlap with other providers.

PR #2289, opened 2018-01-12
feat(provider/ecs): Ecs instance cache client

dkirillov:
Adds an EcsInstanceCacheClient which allows providers/services to interface more easily with the already existing AWS target group cache.


clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/EcsInstanceCacheClient.java
```diff
@@ -0,0 +1,60 @@
...
+
+import static com.netflix.spinnaker.clouddriver.core.provider.agent.Namespace.INSTANCES;
+
+@Component
+public class EcsInstanceCacheClient {
+
+  private final Cache cacheView;
+  private final ObjectMapper objectMapper;
+
+  public EcsInstanceCacheClient(Cache cacheView, @Qualifier("objectMapper") ObjectMapper objectMapper) {
```
robzienert: I don't think this `@Qualifier` is necessary, since spring will find the bean that matches the argument name if more than one exists.

dkirillov: You are correct. Thanks!

I've updated other already existing cache clients as well.

PR #2293, opened 2018-01-12
feat(provider/ecs): ECS Application Provider

dkirillov:
Adds EcsApplicationProvider, responsible for providing Spinnaker applications for the controller.


dkirillov:
The `TestCredential` class is the same as the one found in the AWS module.

If we can make that class public, I would be happy to remove the ECS counterpart.

PR #2303, opened 2018-01-17
fix(provider/kubernetes): v2 cache clusters for workloads only

lwander:
This ensures that only workload based resources are associated with a
spinnaker cluster, reducing 'application' screen clutter


lwander:
@danielpeach this will likely affect any custom builds fyi

andrewbackes:
This is great! Thanks Lars.

PR #2309, opened 2018-01-22
feat(provider/ecs): ECR Image Provider

dkirillov:
Adds an `EcrImageProvider` which is responsible for retrieving ECR image during the findImage stage. 

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcrImageProvider.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcrImageProvider.java
@@ -84,7 +84,7 @@ public class EcrImageProvider implements ImageRepositoryProvider {
     NetflixAmazonCredentials credentials = getCredentials(accountId);
 
     if (!isValidRegion(credentials, region)) {
-      throw new Error("The repository URI provided does not belong to a region that the credentials have access to or is not a valid region.");
+      throw new IllegalArgumentException("The repository URI provided does not belong to a region that the credentials have access to or the region is not valid.");
     }
 
     AmazonECR amazonECR = amazonClientProvider.getAmazonEcr(credentials.getName(), credentials.getCredentialsProvider(), region);
@@ -98,9 +98,9 @@ public class EcrImageProvider implements ImageRepositoryProvider {
       .collect(Collectors.toList());
 
     if (imagesWithThisIdentifier.size() > 1) {
-      throw new Error("More than 1 image has this " + (isTag ? "tag" : "digest") + "!  We can't handle this in the POC!");
+      throw new IllegalArgumentException("More than 1 image has this " + (isTag ? "tag" : "digest") + "!  This is currently not supported.");
     } else if (imagesWithThisIdentifier.size() == 0) {
-      throw new Error(String.format("No image with the " + (isTag ? "tag" : "digest") + " %s was found.", identifier));
+      throw new IllegalArgumentException(String.format("No image with the " + (isTag ? "tag" : "digest") + " %s was found.", identifier));
     }
 
     ImageDetail matchedImage = imagesWithThisIdentifier.get(0);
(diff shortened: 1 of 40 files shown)


clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcrImageProvider.java
```diff
@@ -0,0 +1,188 @@
...
+    String accountId = extractAwsAccountId(url);
+    String repository = extractEcrRepositoryName(url);
+    String identifier = extractEcrIdentifier(repository, url);
+    boolean isTag = !(identifier.startsWith("sha256:") && identifier.length() == ("sha256:".length() + 64));
+    String region = extractAwsRegion(url);
+
+    NetflixAmazonCredentials credentials = getCredentials(accountId);
+
+    if (!isValidRegion(credentials, region)) {
+      throw new Error("The repository URI provided does not belong to a region that the credentials have access to or is not a valid region.");
```
robzienert: `s/or is not a valid region/or the region is not valid`.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcrImageProvider.java
```diff
@@ -0,0 +1,188 @@
...
+
+  private String extractAwsRegion(String imageUrl) {
+    return extractString(REGION_PATTERN, imageUrl, 0,
+      "The repository URI provided does not contain a proper region.");
+  }
+
+  private String extractString(Pattern pattern, String imageUrl, int group, String error) {
+    Matcher matcher = pattern.matcher(imageUrl);
+    if (!matcher.find()) {
+      throw new Error(error);
```
robzienert: What is `Error`? Do you think we should have different exceptions based on the condition? I suspect if we need to do error handling, we wouldn't want to have a `Error`, as that'd be the same anti-pattern of doing a catch on `Exception` or `Throwable` -- too broad.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcrImageProvider.java
```diff
@@ -0,0 +1,188 @@
...
+
+  @Override
+  public List<EcsDockerImage> findImage(String url) {
+    // HTTP(S) part is not needed.
+    url = url.replace("http://", "").replace("https://", "");
+
+    String accountId = extractAwsAccountId(url);
+    String repository = extractEcrRepositoryName(url);
+    String identifier = extractEcrIdentifier(repository, url);
+    boolean isTag = !(identifier.startsWith("sha256:") && identifier.length() == ("sha256:".length() + 64));
```
robzienert: The digest is always a sha256?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcrImageProvider.java
dkirillov: Done.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcrImageProvider.java
dkirillov: Yes, the digest is always a sha256.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcrImageProvider.java
dkirillov: Changed it to `IllegalArgumentException` - it comes down to the passed in (ecr) `url` being improper.

PR #2311, opened 2018-01-23
fix(provider/aws): updating edda API base URL to use /api/v2

dosullivan:
Edda's API base URL seems to have /api/v2/ at the beginning now, rather than /rest/v2/. With the prior configuration, any calls to edda from clouddriver resulted in 404s. I think this covers what needs to be updated in clouddriver to match, but there could be other areas as well. 

FWIW, /api/v2 is shown in the Edda wiki, here: https://github.com/Netflix/edda/wiki/REST


robzienert:
Hmm. This seems to be a departure of the endpoints we have at Netflix and what's in OSS. Perhaps this PR should instead totally remove `/REST` and `/api`, then move this into the config?

cc @cfieber

robzienert:
Closing since this hasn't been updated since submission. Feel free to re-open if you update to address comments.

PR #2314, opened 2018-01-23
feat(provider/ecs): ECS Container Information Service

dkirillov:
Adds a `ContainerInformationService` which is responsible for providing ECS container-centric information to other providers.


BrunoCarrier:
LGTM +2

PR #2315, opened 2018-01-24
fix(provider/ecs): ECS Caching Agents - Account/Region awareness

dkirillov:
`getClusters(...)` method in `AbstractEcsCachingAgent` was returning back clusters for all regions and accounts - which caused issues for caching agents extend that calls and calling that method.

When registering with the scheduler, caching agents were imporperly named. This caused only one instance of each caching agent to be registered, even if there were several caching agents for different regions.

This PR fixes the above.

Issue was brought up by a member of the Spinnaker community - on slack, in the ecs channel - when he tried to use clouddriver, having an account with multiple regions. 

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/Keys.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/Keys.java
@@ -61,6 +61,10 @@ public class Keys implements KeyParser {
 
   @Override
   public Boolean canParseType(String type) {
+    return canParse(type);
+  }
+
+  private static Boolean canParse(String type){
     for (Namespace key : Namespace.values()) {
       if (key.toString().equals(type)) {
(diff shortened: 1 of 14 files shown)


clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/Keys.java
```diff
@@ -61,6 +61,10 @@ public String getCloudProvider() {
 
   @Override
   public Boolean canParseType(String type) {
+    return canParse(type);
+  }
+
+  private static Boolean canParse(String type){
```
robzienert: Please make best efforts to follow the google coding style guide (in the case of this PR, adequate whitespace between conditional expressions, parens, braces).

dkirillov: Done.

PR #2324, opened 2018-01-29
fix(gcs): Prevent NPE when gcs artifact account has no credentials path

sbwsg:


clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gcs/GcsArtifactCredentials.java
```diff
@@ -25,6 +25,7 @@
 import com.google.api.client.json.jackson2.JacksonFactory;
 import com.google.api.services.storage.Storage;
 import com.google.api.services.storage.StorageScopes;
+import org.apache.commons.lang.StringUtils;
```
duftler: Sort order is off here.

sbwsg:
This doesn't fix the specific bug I'm looking into unfortunately.  Just couldn't get clouddriver to start without adding this.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gcs/GcsArtifactCredentials.java
sbwsg: Ah, java noob question: Which part of the import name should I be sorting on?  I've just been doing alphabetical, so "org" comes after "com"

lwander: That's right, but you also have "com" entries below this.

sbwsg: doh

PR #2327, opened 2018-01-30
fix(rrb): Disable percentage calculation should use Math.ceil()

ajordens:
This aligns the calculation with what is also happening within `orca`.


ajordens:
FYI @lwander -- had some complaints about odd behavior when a server group had 3 instances and you were doing 33/66/100 ... the previous logic returned `0` for the 33% interval.

PR #2330, opened 2018-01-30
feat(provider/ecs): ECS Instance Provider

dkirillov:
Added `EcsInstanceProvider` which is responsible for providing an "instance" representation of an ECS Task - used for displaying the instances in server groups.

Updated `ContainerInstance` and `ContainerInstanceCacheClient` to have an `availabilityZone`.

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/view/EcsInstanceProvider.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/view/EcsInstanceProvider.java
@@ -88,10 +88,10 @@ public class EcsInstanceProvider implements InstanceProvider<EcsTask> {
   }
 
   private boolean isValidId(String id, String region) {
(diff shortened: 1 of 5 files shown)


clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/view/EcsInstanceProvider.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  @Override
+  public String getConsoleOutput(String account, String region, String id) {
+    return null;
+  }
+
+  private boolean isValidId(String id, String region) {
+    String id_regex = "[\\da-f]{8}-[\\da-f]{4}-[\\da-f]{4}-[\\da-f]{4}-[\\da-f]{12}";
+    String id_only = String.format("^%s$", id_regex);
```
robzienert: Please camelCase the things.

dkirillov: Done.

Think I was writing Python that day and it carried over :)

PR #2346, opened 2018-02-01
fix(provider/appengine): dont prepend "gs" to all repository URLs

sbwsg:
Incorrectly assumes that repositoryUrl always refers to a GS repo when it could be a git one instead.

Likely source of the integration test failures.


lwander:
Not given the way we setup gcs references

PR #2347, opened 2018-02-01
feat(provider/kubernetes): v2 hpa support

lwander:


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesHorizontalPodAutoscalerCachingAgent.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+
+import static com.netflix.spinnaker.cats.agent.AgentDataType.Authority.AUTHORITATIVE;
+import static com.netflix.spinnaker.cats.agent.AgentDataType.Authority.INFORMATIVE;
+
+@Slf4j
```
sbwsg: there aren't any `log...` calls in this class.  should this annotation remain despite that?

lwander: good catch

PR #2354, opened 2018-02-03
feat(provider/kubernetes): v2 per-account, custom resources

lwander:
This allows you to register (for now, only rudimentary) support for
custom resources, or resources not handled by Spinnaker.

So far, I haven't made `ReplicationControllers` explicitly deployable
using Spinnaker. With this patch, and the following config in my
clouddriver.yml, they could be deployed:

```yaml
kubernetes:
  accounts:
  - name: my-k8s-v2-account
    customResources:
    - kubernetesKind: ReplicationController
      spinnakerKind: ServerGroups
      versioned: true
```


lwander:
https://github.com/spinnaker/spinnaker/issues/2297

PR #2361, opened 2018-02-07
fix(eureka): Handle missing instances when determining what to disable

ajordens:
`getInstanceToModify()` is currently passed a set of instances that have
been fetched directly from AWS.

In the event of autoscaling, it's highly likely that some number of
these instances have not yet been cached by `clouddriver`.

This PR ignores any instances that have not been cached and excludes
them in the `desiredPercentage` calculation.

Previously this failed with:

```
NullPointerException; Message: Cannot invoke method flatten() on null object
```


ajordens:
I'm not completely sold on `AbstractEurekaSupport` doing `ServerGroup serverGroup = getCachedServerGroup(clusterProviders, account, region, asgName)` _but_ this avoids a definite NPE for any sort of autoscaling.

Seem reasonable @tomaslin?

tomaslin:
LGTM

PR #2363, opened 2018-02-08
feat(provider/kubernetes): support configmap & secret refs

lwander:


clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/deployer/KubernetesDeploymentHandlerSpec.groovy
```diff
@@ -141,4 +145,43 @@ spec:
...
+        .reference(target)
+        .build()
+
+    def result = handler.replaceArtifacts(stringToManifest(BASIC_DEPLOYMENT), [artifact])
+
+    then:
+    result.manifest.spec.template.spec.containers[0].envFrom[0].secretRef.name == target
+  }
+
+  void "check that secret volume replaced by the artifact replacer"() {
```
duftler: Is there an 'only' or 'not' missing from the test case name?

lwander: yes

PR #2364, opened 2018-02-08
fix(roles): Fix startup on unsatisfied dependency

lwander:
fyi @dkirillov (Also please stick to [commit message conventions](https://www.spinnaker.io/community/contributing/submitting/#commit-message-conventions))


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/RoleController.java
```diff
@@ -25,6 +25,7 @@
 import org.springframework.web.bind.annotation.RestController;
 
 import java.util.Collection;
+import java.util.Collections;
```
jtk54: This import occurs twice.

jtk54: ^ I need new glasses. LGTM.

dkirillov:
@lwander Sorry about that!

PR #2369, opened 2018-02-10
perf(provider/kubernetes): v2 small perf improvements

lwander:
Not sure if this addresses https://github.com/spinnaker/spinnaker/issues/2334, but it does cut down the number of kubectl calls to roughly 1/3 by reducing redundant namespace calls.


rossedman:
@lwander 👍

PR #2371, opened 2018-02-14
feat(aws): Allow self-referencing security group clone ops

robzienert:
Performs an upsert in a two-phase operation if the security group has self-referencing rules and does not already exist.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/securitygroup/SecurityGroupIngressConverter.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/securitygroup/SecurityGroupIngressConverter.groovy
@@ -33,7 +33,26 @@ class SecurityGroupIngressConverter {
   @Immutable
   static class ConvertedIngress {
     List<IpPermission> converted
-    List<SecurityGroupIngress> missingSecurityGroups
+    MissingSecurityGroups missingSecurityGroups
+  }
+
+  @Immutable
+  static class MissingSecurityGroups {
+    List<SecurityGroupIngress> all
+    List<SecurityGroupIngress> selfReferencing
+
+    boolean anyMissing(boolean ignoreSelfReferencing) {
+      if (all.isEmpty()) {
+        return false;
+      } else if (ignoreSelfReferencing) {
+        return all.size() > selfReferencing.size()
+      }
+      return true
+    }
+
+    boolean hasMissingNonSelfReferencingGroups() {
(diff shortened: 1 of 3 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/securitygroup/SecurityGroupIngressConverter.groovy
```diff
@@ -75,7 +85,10 @@ class SecurityGroupIngressConverter {
         ipPermissions.add(newIpPermission)
       }
     }
-    new ConvertedIngress(ipPermissions, missing)
+    new ConvertedIngress(ipPermissions, new MissingSecurityGroups(
+      all: missing,
+      selfReferencing: missing.findAll { it.name == description.name }
```
robzienert: Should maybe probably add more equality checks here?

ajordens: account/userId maybe?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/securitygroup/UpsertSecurityGroupAtomicOperation.groovy
```diff
@@ -112,4 +109,19 @@ class UpsertSecurityGroupAtomicOperation implements AtomicOperation<Void> {
     null
   }
 
+  private ConvertedIngress convertDescriptionToIngress(SecurityGroupLookup securityGroupLookup, UpsertSecurityGroupDescription description, boolean failOnMissingSelfReferencingRules) {
+    ConvertedIngress ipPermissionsFromDescription = SecurityGroupIngressConverter.
+      convertIngressToIpPermissions(securityGroupLookup, description)
+
+    if (ipPermissionsFromDescription.missingSecurityGroups.hasMissingNonSelfReferencingGroups() && !failOnMissingSelfReferencingRules) {
```
ajordens: Is this logic correct ... throw an exception if we have missing non-self referential groups and `failOnMissingSelfReferential == false`?

robzienert: Yeah, not sure where my head was at there. Updated.

PR #2377, opened 2018-02-16
refactor(provider/kubernetes): v2 rename deployer to handler

lwander:
& corrected import sorts


lwander:
This only renames the package fyi, the top-level class was already called a "handler"

PR #2380, opened 2018-02-18
feat(provider/kubernetes): v2 Add support for Kubernetes jobs

wjoel:
@lwander could you review this, please? Jobs are useful to run smoke tests as part of pipelines, and many other things.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesJobHandler.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesJobHandler.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2018 Joel Wilsson
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.clouddriver.kubernetes.v2.op.handler;
+
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.artifact.ArtifactReplacerFactory;
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.agent.KubernetesCacheDataConverter;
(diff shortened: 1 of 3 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesJobHandler.java
```diff
@@ -0,0 +1,95 @@
...
+    V1JobSpec spec = job.getSpec();
+    if (spec != null && spec.getCompletions() != null) {
+      completions = spec.getCompletions();
+    }
+    int succeeded = 0;
+    if (status.getSucceeded() != null) {
+      succeeded = status.getSucceeded();
+    }
+    if (succeeded < completions) {
+      return result.unstable("Waiting for jobs to finish");
```
lwander: Seems the only way this stage can fail is when the orca/deck configured time out is hit? It would be nice to somehow tie this into [active deadline seconds](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#job-termination-and-cleanup), or introduce a new concept of stability that could cause a stage to fail immediately. The latter could help when deploying individual pods that exit unsuccessfully.

wjoel: Yes, that's correct.

I had a version that tried to make use of active deadline. It would be nice, yes. Seems like orca is in charge of failing stages and there's no good way for clouddriver to signal a failure? My attempt at throwing an exception to abort did not get propagated to orca.

That's why I went looking for the timeout option instead, which kind of works but it's clumsy and sub-optimal.

```java
    if (succeeded < completions) {
      Long startTimeMillis = status.getStartTime().getMillis();
      Long deadlineSeconds = spec.getActiveDeadlineSeconds();
      if (startTimeMillis != null && deadlineSeconds != null &&
        startTimeMillis + Duration.ofSeconds(deadlineSeconds).toMillis() < System.currentTimeMillis()) {
        return result.unstable(
          "Job " + job.getMetadata().getName() + " in namespace " +
          job.getMetadata().getNamespace() + " exceeded deadline");
      }
      return result.unstable("Waiting for all jobs to finish");
    }
```

lwander: We could also try something like

`result.failed("Job timed out")` which orca treats as an immediate failure (regardless of timeout/retries).

wjoel: Sounds like a plan, would you like me to do that as part of this PR or in another one after this is merged?

lwander: Let's do it in a separate PR - it will require orca support too.

PR #2382, opened 2018-02-20
chore(*): Bump spinnaker-dependencies to 0.142.1

robzienert:
@duftler @lwander Trying to update to spinnaker-dependencies to 0.142.1, but one test is failing. I thrashed on it all day and I couldn't really figure out why the ObjectMapper started failing. My initially inclination was going to be to refactor `TestDefaults` to not use maps, but the actual objects, but don't want to go down this path if there's something simpler. Would any of you Googlers be able to pull this down and give it a pair of eyes?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,7 +39,7 @@ allprojects {
   apply plugin: 'groovy'
 
   ext {
-    spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '0.134.0'
+    spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '0.142.1'
   }
 
   def checkLocalVersions = [spinnakerDependenciesVersion: spinnakerDependenciesVersion]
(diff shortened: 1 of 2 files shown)


clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentialsSpec.groovy
```diff
@@ -49,6 +50,7 @@ class GoogleNamedAccountCredentialsSpec extends Specification implements TestDef
       map.region2 == ['z4', 'z5', 'z6']
   }
 
+  @Ignore("ObjectMapper is failing to convert INSTANCE_TYPE_LIST for some reason?")
```
duftler: It's not the ObjectMapper failing. The map the ObjectMapper is trying to convert is malformed. Each `machineTypes` key ends up mapping to `null` and the ObjectMapper then faithfully converts the map to a `MachineTypeAggregatedList`.

Remove the `static` from this line and it will pass: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/security/TestDefaults.groovy#L571

Not sure why it worked before or what changed, but seems like we shouldn't have been able to refer to the other static member defined on the trait in the first place.

PR #2388, opened 2018-02-21
fix(provider/gae): Fix artifact deploy switch statement.

jtk54:


jtk54:
Cherry pick of #2387.

PR #2390, opened 2018-02-23
feat(cache/aws) Add relation between server group->ami and ami->server group

PerGon:
This will add a relation between "Images" and "Server Groups" and the other way around.

It will be useful to check which server groups use the ami foobar, or which ami is being used by a certain server group.


PerGon:
@spinnaker/reviewers ptal, thanks!

emjburns:
@cfieber can you take a look at this? Thanks.

PerGon:
@spinnaker/reviewers

PR #2395, opened 2018-02-27
feat(artifacts): Add support for HTTP artifacts

wjoel:
Basic auth is supported just like for GitHub artifact accounts. I used "http/file" to match "github/file", but I can of course change it to "http/text" if that's preferred.

### DIFF
--- a/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/http/HttpArtifactAccount.java
+++ b/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/http/HttpArtifactAccount.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2018 Joel Wilsson
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/github/GitHubArtifactConfiguration.java
```diff
@@ -56,6 +57,7 @@ GitHubArtifactProviderProperties githubArtifactProviderProperties() {
   OkHttpClient okHttpClient;
 
   @Autowired
+  @Lazy
```
wjoel: Hmm, I've since removed the `ObjectMapper` from the other class, this was used to avoid a "bean is already under construction" error. Will try building and running without this now.

lwander:
https://github.com/spinnaker/spinnaker/issues/2475

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/github/GitHubArtifactConfiguration.java
wjoel: I learned that bean names are global (thanks @lwander), and the problem wasn't `ObjectMapper` anymore but `OkHttpClient`. Fixed.

wjoel:
There's some code duplication between this and the GitHub artifacts support, but since neither are likely to change much and the actual logic is very short, I think it's ok to leave that as a (possible) future refactoring exercise.

Perhaps for when/if we decide to support customization of headers to support custom authentication schemes.

lwander:
Agreed - if we need to revisit this it'll be quick & easy to dedup. Thanks!

PR #2410, opened 2018-03-06
feat(provider/kubernetes): v2 restrict namespaces for uploading manifests

gardleopard:
Solves https://github.com/spinnaker/spinnaker/issues/2325

Depends on #2409

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
@@ -115,6 +115,14 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
     return cachedDefaultNamespace;
   }
 
+  public List<String> getNamespaces() {
+    return namespaces;
+  }
+
+  public List<String> getOmitNamespaces() {
+    return omitNamespaces;
+  }
+
   public String lookupDefaultNamespace() {
     String namespace = defaultNamespace;
     try {
@@ -131,10 +139,6 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
       namespace = defaultNamespace;
     }
     return namespace;
-  }
-
-  public boolean containsNamespace(String namespace) {
-    return namespaces.contains(namespace) || lookupDefaultNamespace().equals(namespace);
   }
 
   @Getter
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
@@ -133,6 +133,10 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
     return namespace;
   }
 
+  public boolean containsNamespace(String namespace) {
+    return namespaces.contains(namespace) || lookupDefaultNamespace().equals(namespace);
+  }
+
   @Getter
   private final boolean debug;
 
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
@@ -59,7 +59,9 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
   private final Clock clock;
   private final String accountName;
   private final ObjectMapper mapper = new ObjectMapper();
+  @Getter
   private final List<String> namespaces;
+  @Getter
   private final List<String> omitNamespaces;
   private final List<KubernetesKind> kinds;
   private final List<KubernetesKind> omitKinds;
@@ -113,14 +115,6 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
     }
 
     return cachedDefaultNamespace;
-  }
-
-  public List<String> getNamespaces() {
-    return namespaces;
-  }
-
-  public List<String> getOmitNamespaces() {
-    return omitNamespaces;
   }
 
   public String lookupDefaultNamespace() {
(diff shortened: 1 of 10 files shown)


gardleopard:
I have tested this now and it works, except I think the default namespace handling is strange.
It gets to this line https://github.com/spinnaker/clouddriver/blob/0a288b7e21d7746cc91eac6f3dce27afca840dd4/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java#L127

I think maybe I cannot use that method and also that it needs to check for the ommitnamespaces and ignore all of this if no namespaces are explicitly defined.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/validator/KubernetesValidationUtil.java
```diff
@@ -93,6 +100,11 @@ public boolean validateV2Credentials(AccountCredentialsProvider provider, String
       return false;
     }
 
+    if (!((KubernetesV2Credentials)credentials.getCredentials()).containsNamespace(namespace)) {
```
lwander: missing space after first `)`

lwander:
Checking the omit namespace is definitely important - I don't think this should be the place to check the default namespace. I'm thinking some sort of validation should fail if someone specifies a "Default namespace" that they can't deploy to because it's not listed in the "namespaces", or omitted by the "omitNamespaces".

gardleopard:
@lwander I have changed the logic and added tests for this now. I think the tests clearly shows the behaviour now.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -115,6 +115,14 @@ public String getDefaultNamespace() {
     return cachedDefaultNamespace;
   }
 
+  public List<String> getNamespaces() {
```
lwander: I would just annotated the `namespaces` and `omitNamespaces` fields with `@Getter` like the properties above around line 77

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/validator/KubernetesValidationUtil.java
```diff
@@ -93,6 +100,25 @@ public boolean validateV2Credentials(AccountCredentialsProvider provider, String
...
+      return false;
+    }
+
+    return true;
+  }
+
+  protected boolean validateNamespace(String namespace, KubernetesV2Credentials credentials) {
+    final List<String> configuredNamespaces = credentials.getNamespaces();
+    if (configuredNamespaces != null && !configuredNamespaces.isEmpty() && !configuredNamespaces.contains(namespace)) {
+      reject( "wrongNamespace", "namespace");
```
lwander: Extra space before `"`

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/validator/KubernetesValidationUtil.java
```diff
@@ -93,6 +100,25 @@ public boolean validateV2Credentials(AccountCredentialsProvider provider, String
...
+      return false;
+    }
+
+    final List<String> omitNamespaces = credentials.getOmitNamespaces();
+    if (omitNamespaces != null && omitNamespaces.contains(namespace)) {
+      reject( "omittedNamespace", "namespace");
```
lwander: Extra space before `"`

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/validator/KubernetesValidationUtil.java
```diff
@@ -77,11 +79,16 @@ public boolean validateNotEmpty(String attribute, String value) {
...
   }
 
-  public boolean validateV2Credentials(AccountCredentialsProvider provider, String accountName) {
+  public boolean validateV2Credentials(AccountCredentialsProvider provider, String accountName, String namespace) {
+    log.info("Validating credentials for {} {}", accountName, namespace);
     if (!validateNotEmpty("account", accountName)) {
       return false;
     }
 
+    if (!validateNotEmpty("account", namespace)) {
```
lwander: Should this be `"namespace"`?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/validator/KubernetesValidationUtil.java
lwander: Also, should the string `"namespace`" be replaced with the actual requested namespace?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/validator/KubernetesValidationUtil.java
```diff
@@ -93,6 +100,25 @@ public boolean validateV2Credentials(AccountCredentialsProvider provider, String
...
     }
 
+    if (!validateNamespace(namespace, (KubernetesV2Credentials)credentials.getCredentials())) {
+      return false;
+    }
+
+    return true;
+  }
+
+  protected boolean validateNamespace(String namespace, KubernetesV2Credentials credentials) {
```
lwander: I'm thinking we should skip this (e.g. `return true`) when `namespace` is empty. If none is supplied, the default will be used which should be valid.

gardleopard: in line 88 we check if the namespace is empty, so it wont reach this part of the code if a namespace is not supplied.

lwander: Ah I missed that - thanks!

wjoel: Line 89 returns `false`, which is not correct for resources without a namespace. I've submitted https://github.com/spinnaker/clouddriver/pull/2422 to change this.

PR #2414, opened 2018-03-08
feat(kubernetes): allow cluster provider to return subclasses of Kube…

danielpeach:
…rnetesV2ServerGroup

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2ServerGroup.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2ServerGroup.java
@@ -25,6 +25,7 @@ import com.netflix.spinnaker.clouddriver.kubernetes.v2.artifact.ArtifactReplacer
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.artifact.ArtifactReplacerFactory;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.Keys;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.agent.KubernetesCacheDataConverter;
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.view.provider.data.KubernetesV2ServerGroupCacheData;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.KubernetesManifest;
 import com.netflix.spinnaker.clouddriver.model.HealthState;
 import com.netflix.spinnaker.clouddriver.model.Instance;
@@ -96,7 +97,7 @@ public class KubernetesV2ServerGroup extends ManifestBasedModel implements Serve
     return disabled;
   }
 
-  private KubernetesV2ServerGroup(KubernetesManifest manifest, String key, List<KubernetesV2Instance> instances, Set<String> loadBalancers) {
+  protected KubernetesV2ServerGroup(KubernetesManifest manifest, String key, List<KubernetesV2Instance> instances, Set<String> loadBalancers) {
     this.manifest = manifest;
     this.key = (Keys.InfrastructureCacheKey) Keys.parseKey(key).get();
     this.instances = new HashSet<>(instances);
@@ -118,15 +119,7 @@ public class KubernetesV2ServerGroup extends ManifestBasedModel implements Serve
         .build();
   }
 
-  public static KubernetesV2ServerGroup fromCacheData(CacheData cd) {
-    return fromCacheData(cd, new ArrayList<>(), new ArrayList<>());
-  }
-
-  public static KubernetesV2ServerGroup fromCacheData(CacheData cd, List<CacheData> instanceData) {
-    return fromCacheData(cd, instanceData, new ArrayList<>());
-  }
-
-  public static KubernetesV2ServerGroup fromCacheData(CacheData cd, List<CacheData> instanceData, List<CacheData> loadBalancerData) {
+  private static KubernetesV2ServerGroup fromCacheData(CacheData cd, List<CacheData> instanceData, List<CacheData> loadBalancerData) {
     if (cd == null) {
       return null;
     }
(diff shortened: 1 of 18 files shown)


lwander:
Can you explain why?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2ServerGroup.java
```diff
@@ -159,6 +160,10 @@ public static KubernetesV2ServerGroup fromCacheData(CacheData cd, List<CacheData
     return new KubernetesV2ServerGroup(manifest, cd.getId(), instances, loadBalancers);
   }
 
+  public static KubernetesV2ServerGroup fromCacheData(ServerGroupCacheData cacheData) {
```
lwander: If you're adding this to cleanup the number of parameters, can you apply it every place the fully expanded call gets used?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2ClusterProvider.java
```diff
@@ -284,4 +300,22 @@ private KubernetesV2Cluster translateCluster(CacheData clusterDatum,
 
     return new KubernetesV2Cluster(clusterDatum.getId(), serverGroups, loadBalancers);
   }
+
+  // Allows handlers to map cache data to a KubernetesV2ServerGroup subclass.
+  private KubernetesV2ServerGroup serverGroupFromCacheData(ServerGroupCacheData cacheData) {
+    KubernetesManifest manifest = KubernetesCacheDataConverter.getManifest(cacheData.getServerGroupData());
+    return kubernetesHandlers.stream()
+      .filter(h -> h.kind().toString().equals(manifest.getKind().toString()) && h instanceof HandlesCacheMapping)
+      .map(h -> ((HandlesCacheMapping<KubernetesV2ServerGroup, ServerGroupCacheData>)h).fromCacheData(cacheData))
```
lwander: Missing space between `)h`

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2ServerGroup.java
danielpeach: Sure, though it was more so that the interface could have one type parameter.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2ClusterProvider.java
```diff
@@ -284,4 +300,22 @@ private KubernetesV2Cluster translateCluster(CacheData clusterDatum,
 
     return new KubernetesV2Cluster(clusterDatum.getId(), serverGroups, loadBalancers);
   }
+
+  // Allows handlers to map cache data to a KubernetesV2ServerGroup subclass.
+  private KubernetesV2ServerGroup serverGroupFromCacheData(ServerGroupCacheData cacheData) {
+    KubernetesManifest manifest = KubernetesCacheDataConverter.getManifest(cacheData.getServerGroupData());
+    return kubernetesHandlers.stream()
+      .filter(h -> h.kind().toString().equals(manifest.getKind().toString()) && h instanceof HandlesCacheMapping)
+      .map(h -> ((HandlesCacheMapping<KubernetesV2ServerGroup, ServerGroupCacheData>) h).fromCacheData(cacheData))
```
lwander: Can you use the `kindMap` instead?

danielpeach:
@lwander please review. Updated server groups + server group managers.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2ClusterProvider.java
```diff
@@ -284,4 +300,16 @@ private KubernetesV2Cluster translateCluster(CacheData clusterDatum,
 
     return new KubernetesV2Cluster(clusterDatum.getId(), serverGroups, loadBalancers);
   }
+
+  @SuppressWarnings("unchecked")
+  private KubernetesV2ServerGroup serverGroupFromCacheData(KubernetesV2ServerGroupCacheData cacheData) {
```
lwander: Function name isn't consistent with sibling in "ServerGroupManager" code

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/CustomKubernetesHandlerFactory.java
```diff
@@ -33,7 +33,8 @@ public static KubernetesHandler create(KubernetesKind kubernetesKind, SpinnakerK
     return new Handler(kubernetesKind, spinnakerKind, versioned);
   }
 
-  private static class Handler extends KubernetesHandler {
+  // TODO(dpeach): come up with the right type parameters for KubernetesHandler here.
+  private static class Handler extends KubernetesHandler<Void, Void> {
```
lwander: I think Object, Object is a little friendlier

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesHandler.java
```diff
@@ -51,6 +57,19 @@
   abstract public SpinnakerKind spinnakerKind();
   abstract public Status status(KubernetesManifest manifest);
 
+  public ManifestBasedModel fromCacheData(KubernetesV2CacheData cacheData) {
```
lwander: Should this be `static`?

lwander: Ah I guess not because you plan to override it? It's a little strange that there's a non-static instance method that handles every type of cache data - but will be overridden to handle only one type of cache data...

lwander: If you had a `KubernetesServerGroupHandler` that implements this method and does only the server group bit, and is implemented by replica set, stateful set, daemon set, etc... and the same for server group managers, I think the pattern would be a little clearer.

danielpeach: Hmm... I don't think that would work for custom resources. The problem is that the default mapping has to be encoded with the `SpinnakerKind`.

PR #2417, opened 2018-03-09
feat(artifacts): Add a "no auth" http artifact account where needed

lwander:
@wjoel FYI


wjoel:
@lwander maybe we should just always include that account? That way, "no-auth-http-account" wouldn't disappear on users when they make an HTTP artifact account with authentication enabled (after they have started using "no-auth-http-account" in some pipelines).

PR #2419, opened 2018-03-11
feat(provider/kubernetes): v2 Add support for RBAC resource kinds

wjoel:
Adds support for ClusterRole, ClusterRoleBinding, Role, RoleBinding and ServiceAccount resource kinds.

Cluster-level resources like Role, ClusterRoleBinding, PersistentVolume, etc. do belong to any namespace, so we also add support for resource kinds without namespaces. Previously the "namespace" resource kind was the only kind which did not belong to a namespace.

While this PR adds a lot of lines, most of it is boilerplate code in the new caching agents and handlers.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesClusterRoleHandler.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesClusterRoleHandler.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2018 Joel Wilsson
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesClusterRoleHandler.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  @Override
+  public boolean versioned() {
+    return false;
+  }
+
+  @Override
+  public KubernetesSpinnakerKindMap.SpinnakerKind spinnakerKind() {
+    return KubernetesSpinnakerKindMap.SpinnakerKind.CONFIGS;
```
lwander: I don't think these should be "configs" -- while there is not "real" precedent they so far mean things that your application consumes/mounts. We could either leave these "unclassified", or create a new "kind" like "auth".

wjoel: Yes, I looked at `KubernetesPersistentVolumeHandler` and it's using "configs", so I picked that to be consistent... however "unclassified" seems like a better choice for both. I'll change this in a few hours.

wjoel: Done.

PR #2425, opened 2018-03-14
(DNMY) feat(provider/aws): Add X-Ray tracing support to SDK clients

robzienert:
Putzing around with adding [X-Ray support](https://docs.aws.amazon.com/xray/latest/devguide/aws-xray.html) for the AWS provider. Should we find value in this, it should work for our other services, too.

⚠️ Our clouddriver profile doesn't have xray perms, so I can't setup the daemon locally to test this out yet. Ticket is out, tho.


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/tracing/RedisDynamicSamplingStrategy.java
```diff
@@ -0,0 +1,169 @@
...
+ */
+public class RedisDynamicSamplingStrategy implements SamplingStrategy {
+
+  private final static Logger log = LoggerFactory.getLogger(RedisDynamicSamplingStrategy.class);
+
+  private final static String CLIENT_NAME = "tracing";
+  private final static String KEY = "{tracing:xray}:samplingRules";
+
+  private final RedisClientSelector redisClientSelector;
+  private final SamplingRuleManifest fallbackManifest;
```
robzienert: Rule manifests look like this: https://github.com/aws-samples/eb-java-scorekeep/blob/xray/src/main/resources/sampling-rules.json

dreynaud:
This should also install the X-Ray daemon, right?

robzienert:
I didn't want to include anything XRay related outside of `AwsXrayConfiguration`, so things are only wired up if the condition `@ConditionalOnExpression("${aws.enabled} && ${aws.xray.enabled}")` returns true. In this configuration class, the servlet filter is wired up, though.

dreynaud:
@robzienert gotchya, missed that. Good, go on then.

PR #2428, opened 2018-03-15
feat(reservations): V4 reservation report that proportionally covers shortfalls

ajordens:
The V3 reservation report uses regional reservations in a very specific
order to cover shortfalls.

It is problematic if the consumer does not look at all instance types in
a family when determining availability.

This PR uses a proportional approach to cover shortfalls. Each instance
type is allocated a portion of regional reservations in proportion to
how much of the overall shortfall it contributed.

ex)
m4.4xlarge has a shortfall equivalent to 20 fxlarge
m4.* has a total shortfall equivalent to 40 fxlarge

Total regional reservations equivalent to 30 fxlarge

m4.4xlarge would cover shortfall with 50% of the regional reservations
(20/40 * 30 == 15)


ajordens:
After validation, `v4` will replace `v3` ... I'll reconcile test cases and code when that happens.

PR #2430, opened 2018-03-15
feat(aws): add support for resizing ASGs with optional min/max/desired

dreynaud:
We need to replace ResizeAsgDescription.Capacity with something that has
optional (i.e. nullable) fields. ServerGroup.Capacity happens to have
exactly that.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperation.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperation.groovy
@@ -61,6 +61,12 @@ class ResizeAsgAtomicOperation implements AtomicOperation<Void> {
                          ServerGroup.Capacity capacity,
                          ResizeAsgDescription.Constraints constraints) {
     task.updateStatus PHASE, "Beginning resize of ${asgName} in ${region} to ${capacity}."
+
+    if (capacity.min == null && capacity.max == null && capacity.desired == null) {
+      task.updateStatus PHASE, "Skipping resize of ${asgName} in ${region}, at least one field in ${capacity} needs to be non-null"
+      return
+    }
+
     def autoScaling = amazonClientProvider.getAutoScaling(description.credentials, region, true)
     def describeAutoScalingGroups = autoScaling.describeAutoScalingGroups(
       new DescribeAutoScalingGroupsRequest().withAutoScalingGroupNames(asgName)
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperation.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperation.groovy
@@ -24,6 +24,7 @@ import com.netflix.spinnaker.clouddriver.aws.deploy.description.ResizeAsgDescrip
 import com.netflix.spinnaker.clouddriver.aws.security.AmazonClientProvider
 import com.netflix.spinnaker.clouddriver.data.task.Task
 import com.netflix.spinnaker.clouddriver.data.task.TaskRepository
+import com.netflix.spinnaker.clouddriver.model.ServerGroup
 import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation
 import org.springframework.beans.factory.annotation.Autowired
 
@@ -57,7 +58,7 @@ class ResizeAsgAtomicOperation implements AtomicOperation<Void> {
 
   private void resizeAsg(String asgName,
                          String region,
-                         ResizeAsgDescription.Capacity capacity,
+                         ServerGroup.Capacity capacity,
                          ResizeAsgDescription.Constraints constraints) {
     task.updateStatus PHASE, "Beginning resize of ${asgName} in ${region} to ${capacity}."
     def autoScaling = amazonClientProvider.getAutoScaling(description.credentials, region, true)
@@ -71,6 +72,7 @@ class ResizeAsgAtomicOperation implements AtomicOperation<Void> {
 
     validateConstraints(constraints, describeAutoScalingGroups.getAutoScalingGroups().get(0))
 
+    // min, max and desired may be null
     def request = new UpdateAutoScalingGroupRequest().withAutoScalingGroupName(asgName)
         .withMinSize(capacity.min)
         .withMaxSize(capacity.max)
--- a/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperationUnitSpec.groovy
+++ b/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperationUnitSpec.groovy
@@ -112,13 +112,13 @@ class ResizeAsgAtomicOperationUnitSpec extends Specification {
   }
 
   @Unroll
-  void "should support resize requests with only some fields to update (min: #minSize, max: #maxSize, desired: #desiredCapacity)"() {
+  void "should support resize requests with only some fields to update (min: #minSize, max: #maxSize, desired: #desired)"() {
     setup:
     def description = new ResizeAsgDescription(
       asgs: [[
                serverGroupName: "myasg-stack-v000",
                region         : "us-west-1",
-               capacity       : new ServerGroup.Capacity(min: minSize, max: maxSize, desired: desiredCapacity)
+               capacity       : new ServerGroup.Capacity(min: minSize, max: maxSize, desired: desired)
              ]]
     )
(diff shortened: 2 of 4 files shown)


clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperationUnitSpec.groovy
```diff
@@ -110,8 +109,38 @@ class ResizeAsgAtomicOperationUnitSpec extends Specification {
     capacity(0, 2, 3)    | 1          | 2          | 3              || true
   }
 
+  void "should support resize requests with only some fields to update"() {
+    def mockAutoScaling = Mock(AmazonAutoScaling)
```
ajordens: This should be in a `given:` block (or `setup:` as this specification appears to be using)

Looks like this setup is duplicated a bit, could extract to class-level:

```
class ResizeAsgAtomicOperationUnitSpec extends Specification {

  def amazonAutoScaling = Mock(AmazonAutoScaling)
  def amazonClientProvider = Mock(AmazonClientProvider) {
    _ * getAutoScaling(_, _, true) >> { return amazonAutoScaling }
  }

  def setupSpec() {
    TaskRepository.threadLocalTask.set(Mock(Task))
  }

  ...
```

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperationUnitSpec.groovy
```diff
@@ -110,8 +109,38 @@ class ResizeAsgAtomicOperationUnitSpec extends Specification {
...
+    def mockAmazonClientProvider = Mock(AmazonClientProvider)
+    mockAmazonClientProvider.getAutoScaling(_, _, true) >> mockAutoScaling
+    mockAutoScaling.describeAutoScalingGroups(_) >> new DescribeAutoScalingGroupsResult().withAutoScalingGroups(
```
ajordens: I'd probably put this in the `then:` block

```
1 * mockAutoScaling.describeAutoScalingGroups(_) >> { ... }
```

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperationUnitSpec.groovy
```diff
@@ -110,8 +109,38 @@ class ResizeAsgAtomicOperationUnitSpec extends Specification {
...
+      new AutoScalingGroup().withAutoScalingGroupName("myasg-stack-v0001")
+    )
+
+    def description = new ResizeAsgDescription(
+      asgs: [[
+               serverGroupName: "myasg-stack-v000",
+               region         : "us-west-1",
+               capacity       : new ServerGroup.Capacity(min: 42) // no max or desired specified
```
ajordens: I would suggest parameterizing this capacity and then using a `where:` to pass in a variety of combinations

```
sourceMin | sourceMax | sourceDesired || expectedMin || expectedMax || expectedDesired
42        | null      | null          || 42          || null        || null
42        | 40        | null          || 42          || null        || null

... etc.
```

The split between inputs and outputs might not be necessary (since they're expected to be identical) but you get the point.

Motivation being that we're actually asserting that it works for all combinations.

dreynaud: good idea. What do you think should happen when we have `(min: null, max: null, desired: null)`?

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperationUnitSpec.groovy
dreynaud: done in 6fbc3d7d7

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperationUnitSpec.groovy
dreynaud: good point, done in 6fbc3d7d7

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperation.groovy
```diff
@@ -57,9 +58,15 @@ class ResizeAsgAtomicOperation implements AtomicOperation<Void> {
 
   private void resizeAsg(String asgName,
                          String region,
-                         ResizeAsgDescription.Capacity capacity,
+                         ServerGroup.Capacity capacity,
                          ResizeAsgDescription.Constraints constraints) {
     task.updateStatus PHASE, "Beginning resize of ${asgName} in ${region} to ${capacity}."
+
+    if (capacity.min == null && capacity.max == null && capacity.desired == null) {
+      task.updateStatus PHASE, "Skipping resize of ${asgName} in ${region}, at least one field in ${capacity} needs to be non-null"
```
ajordens: 👍

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperationUnitSpec.groovy
```diff
@@ -110,8 +111,65 @@ class ResizeAsgAtomicOperationUnitSpec extends Specification {
...
   }
 
+  @Unroll
+  void "should support resize requests with only some fields to update (min: #minSize, max: #maxSize, desired: #desiredCapacity)"() {
+    setup:
+    def description = new ResizeAsgDescription(
+      asgs: [[
+               serverGroupName: "myasg-stack-v000",
+               region         : "us-west-1",
+               capacity       : new ServerGroup.Capacity(min: minSize, max: maxSize, desired: desiredCapacity)
```
ajordens: IMO I'd probably be consistent on the naming ... minSize vs maxSize vs desiredCapacity.

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperationUnitSpec.groovy
```diff
@@ -110,8 +111,65 @@ class ResizeAsgAtomicOperationUnitSpec extends Specification {
...
+    where:
+    minSize | maxSize | desiredCapacity
+    42      | null    | null
+    null    | 42      | null
+    null    | null    | 42
+    0       | null    | null
+    0       | 0       | 0    // not the same as (null, null, null)
+  }
+
+  void "skip resizing if all capacity fields are null"() {
```
ajordens: I think you could avoid the need for this whole test if you just added a `null | null | null` to the previous test. 

You could add a new column:

```
minSize | maxSize | desiredSize || expectedAutoScalingInvocations 
```

and then

```
expectedAutoScalingInvocations * mockAutoScaling. describeAutoScalingGroups(_) >> {
expectedAutoScalingInvocations * mockAutoScaling.updateAutoScalingGroup(_) >> {
...
}
```

The null/null/null test is certainly valid, this suggestion is just aimed at getting coverage without needing 20 or so additional lines of testing code.

ajordens:
Feel free to merge once the test spec is updated.

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperationUnitSpec.groovy
dreynaud: Ha, I know. I am sticking to the AWS terminology though

ajordens: Reasonable argument I suppose

ajordens: Reasonable argument I suppose

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ResizeAsgAtomicOperationUnitSpec.groovy
dreynaud: in bde71d874

PR #2437, opened 2018-03-20
fix(metrics/prometheus): dots are not allowed in labels in prometheus

gardleopard:
Putting account.name into account as account is used for account name elsewhere.
account.id will become account_id now.


ewiseblatt:
Perhaps we should change the prometheus exporter to transform the '.' names.
That would be a more general fix.
If we go with this, clear with @cfieber as this might affect their monitoring.

PR #2412, opened 2018-03-07
fix(artifacts): Make Docker image name matching more accurate

wjoel:
In particular, the previous pattern had issues with dashes in image names. Replace it with a pattern based on the Docker source code.

PR #2442, opened 2018-03-20
fix(artifacts): Fix broken Docker image name matching and add tests

wjoel:
The changes in https://github.com/spinnaker/clouddriver/pull/2412 used the wrong pattern, and didn't match the beginning and end of the string, so it allowed partial matches in invalid strings.

Adding some tests to make it less likely that this happens again.


wjoel:
I think the tests fail every once in a while due to the the parallel stuff that was enabled a while back. It gets stuck sometimes, at least on my machine, with one of the Java processes consuming 100% CPU. I'd like to do some additional verification before merging this anyway.

wjoel:
@lwander this should be good to go. The additional verification turned into https://github.com/spinnaker/clouddriver/pull/2452 and https://github.com/spinnaker/clouddriver/pull/2445

lwander:
Great work!

PR #2443, opened 2018-03-21
chore(credentials): enable verbose logging for the jsch client

dreynaud:


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/bastion/BastionCredentialsProvider.groovy
```diff
@@ -83,4 +85,23 @@ class BastionCredentialsProvider implements AWSCredentialsProvider {
...
+      new Integer(com.jcraft.jsch.Logger.ERROR), "JSCH ERROR: ",
+      new Integer(com.jcraft.jsch.Logger.FATAL), "JSCH FATAL: ",
+    )
+
+    boolean isEnabled(int level) {
+      return true
+    }
+
+    void log(int level, String message) {
+      System.err.print(name.get(level))
```
ajordens: System.err what what?

Seems like we could just `@Slf4j` and `log.debug` the message ... it doesn't seem all that important to worry about level.

ajordens:
we solved the problem ... do we actually want this?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/bastion/BastionCredentialsProvider.groovy
dreynaud: It did not actually help, let's close this

PR #2447, opened 2018-03-22
fix(ci/travis): Fix travis build timeout on prs

gardleopard:
https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received


robzienert:
I LIKE IT

cfieber:
Think we tracked this down to tooo many parallel workers in newer gradle for the travis ci nodes

PR #2452, opened 2018-03-23
fix(provider/kubernetes): v2 Fix artifact replacers for StatefulSets

wjoel:
The handler for statefulsets was essentially using a hard-coded `ArtifactReplacerFactory.dockerImageReplacer()`, and was missing replacers for configmaps, secrets, etc.


wjoel:
Similar to https://github.com/spinnaker/clouddriver/pull/2445. We should consider moving this to a base class or adding helper function to `ArtifactReplacer` or `ArtifactReplacerFactory` to register this standard set of replacers.

PR #2456, opened 2018-03-27
fix(provider/kubernetes): v2 Avoid NPE for unsupported k8s kinds

wjoel:
Show a nice error message from an `IllegalArgumentException` instead of an incomprehensible NPE stack trace when a user attempts to deploy an unsupported resource kind.

Instead of this:
unsupported-kubernetes-kind-npe

We show this:
unsupported-kubernetes-kind

(I will shortly submit a PR to fix the support for rolebindings)


lwander:
Nice

PR #2460, opened 2018-03-28
chore(build): add debug flag to clouddriver build

ezimanyi:
Starting clouddriver with /gradlew -DDEBUG=true now causes the JVM to
listen for a remote debugger on port 7102.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,5 +18,5 @@ To start the JVM in debug mode, set the Java system property `DEBUG=true`:
 ```
 
 The JVM will then listen for a debugger to be attached on port 7102.  The JVM will _not_ wait for
-the debugger to be attached before starting Clouddriver; the relevant JVM arguments can seen and be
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -9,3 +9,14 @@ This service is the main integration point for Spinnaker cloud providers like AW
...
+
+### Debugging
+
+To start the JVM in debug mode, set the Java system property `DEBUG=true`:
+```
+./gradlew -DDEBUG=true
+```
+
+The JVM will then listen for a debugger to be attached on port 7102.  The JVM will _not_ wait for
+the debugger to be attached before starting Clouddriver; the relevant JVM arguments can seen and be
```
sbwsg: lang nit: `arguments can _be_ seen and modified`

ezimanyi: Good catch; fixed

PR #2463, opened 2018-03-28
fix(aws): Make deploy atomic operation more resilient to AWS failure

robzienert:
* Adds some more logging
* Will attempt to continue deploy operation if the server group already exists and matches desired state

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AsgReferenceCopier.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AsgReferenceCopier.groovy
@@ -16,15 +16,20 @@
 package com.netflix.spinnaker.clouddriver.aws.deploy
 
 import com.amazonaws.services.autoscaling.AmazonAutoScaling
-import com.amazonaws.services.autoscaling.model.*
-import com.netflix.spinnaker.clouddriver.aws.deploy.scalingpolicy.ScalingPolicyCopier
+import com.amazonaws.services.autoscaling.model.AlreadyExistsException
+import com.amazonaws.services.autoscaling.model.DescribeScheduledActionsRequest
+import com.amazonaws.services.autoscaling.model.DescribeScheduledActionsResult
+import com.amazonaws.services.autoscaling.model.PutScheduledUpdateGroupActionRequest
+import com.amazonaws.services.autoscaling.model.ScheduledUpdateGroupAction
+import com.netflix.spinnaker.clouddriver.aws.model.AwsResultsRetriever
 import com.netflix.spinnaker.clouddriver.aws.security.AmazonClientProvider
 import com.netflix.spinnaker.clouddriver.aws.security.NetflixAmazonCredentials
-import com.netflix.spinnaker.clouddriver.data.task.Task
-import com.netflix.spinnaker.clouddriver.aws.model.AwsResultsRetriever
 import com.netflix.spinnaker.clouddriver.aws.services.IdGenerator
+import com.netflix.spinnaker.clouddriver.data.task.Task
 import groovy.transform.Canonical
+import groovy.util.logging.Slf4j
 
+@Slf4j
 @Canonical
 class AsgReferenceCopier {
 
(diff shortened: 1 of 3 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
```diff
@@ -147,8 +159,9 @@ class AutoScalingWorker {
 
     String launchConfigName = regionScopedProvider.getLaunchConfigurationBuilder().buildLaunchConfiguration(application, subnetType, settings, legacyUdf)
 
-    task.updateStatus AWS_PHASE, "Deploying ASG."
+    task.updateStatus AWS_PHASE, "Deploying ASG: $asgName"
 
+    // TODO rz - Make bester
```
emjburns: Remove or make bester.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AsgReferenceCopier.groovy
```diff
@@ -59,7 +64,13 @@ class AsgReferenceCopier {
       if (startTime?.time > System.currentTimeMillis()) {
         request.withStartTime(startTime)
       }
-      targetAutoScaling.putScheduledUpdateGroupAction(request)
+
+      try {
+        targetAutoScaling.putScheduledUpdateGroupAction(request)
+      } catch (AlreadyExistsException e) {
+        // This should never happen as the name is generated with a UUID.
+        log.info("Scheduled action already exists on ASG, continuing: $request")
```
ajordens: Maybe a warn?

For better or worse I try and follow a logging convention like `"Scheduled action already exists on ASG (request: {})", request)`

🤷‍♂️

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
```diff
@@ -232,17 +245,33 @@ class AutoScalingWorker {
...
+    Exception ex = retrySupport.retry({ ->
+      try {
+        autoScaling.createAutoScalingGroup(request)
+      } catch (AlreadyExistsException e) {
+        if (!shouldProceedWithExistingState(autoScaling, asgName, request)) {
+          return e
+        }
+        log.debug("Determined pre-existing ASG is desired state, continuing...", e)
+      }
+    }, 5, 1000, true)
```
ajordens: I normally don't do exponential backoff

dreynaud: why not?

robzienert: I think I'll keep the exp backoff, but having a knob exposed inside of kork to adjust the exponential base like Asher suggests is a great idea.

robzienert: Seen here: https://github.com/spinnaker/kork/pull/141

ajordens: The AWS SDK is already retrying exponentially, I'm still not convinced that we need to layer our own exponential retries on top of that even with the ability to manipulate the base.

🤷‍♂️

robzienert:
Updated the PR to not use exponential backoff. I also bumped retries from 5x -> 10x.

dreynaud:
I don't feel very good about these long linear retries. I am generally of
the opinion that if you didn't get a result after 3 attempts, you should
probably bail and have the next layer deal with it. E.g. we could have
clouddriver fail fast but signal that the error is retryable and orca could
re-attempt later. Am I making sense?

2018-04-04 10:55 GMT-07:00 Rob Zienert <notifications@github.com>:

ajordens:
I'm inclined to see this merge as it's an improvement over what's already in place.

I do agree with the general sentiment of pulling retries up to orca (retry for 2hrs > manual retry for a few minutes) _but_ not sure we can do this until we have idempotent requests to clouddriver.

Make sense?

robzienert:
I'm leaning towards agreement with Adam on this one. A personal objective for me is to get clouddriver operations to be idempotent, but I don't want that end goal to be a blocker on incremental improvement.

Very happy to talk offline about how we can move the ball towards that goal.

PR #2464, opened 2018-03-28
feat(provider/kubernetes): v2 cache & surface events per-resource

lwander:
https://github.com/spinnaker/spinnaker/issues/2587

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2Manifest.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2Manifest.java
@@ -26,7 +26,9 @@ import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
+import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 @Data
@@ -40,4 +42,5 @@ public class KubernetesV2Manifest implements Manifest {
   private KubernetesManifest manifest;
   private Status status;
   private Set<Artifact> artifacts = new HashSet<>();
+  private List<KubernetesManifest> events = new ArrayList<>();
 }
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2ManifestProvider.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2ManifestProvider.java
@@ -32,7 +32,12 @@ import org.apache.commons.lang3.tuple.Pair;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
+import java.util.Collections;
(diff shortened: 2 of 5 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2Manifest.java
```diff
@@ -40,4 +40,5 @@
   private KubernetesManifest manifest;
   private Status status;
   private Set<Artifact> artifacts = new HashSet<>();
+  private Set<KubernetesManifest> events = new HashSet<>();
```
wjoel: I would have expected this to be a list of `V1Event`. Shouldn't it return the events shown in `kubectl describe <kind> <name>`, similar to `kubectl get events --field-selector involvedObject.name=name-of-my-pod,involvedObject.kind=Pod`? I think that's what we'd like to show in the UI.

lwander: `V1Event` can be mapped to and from `KubernetesManifest`. I'm afraid of embedding those models in here -- we have no assurance that those will work across versions of future kubernetes clusters (what if `V2Event` gets introduced for all new resources starting at 1.10?)

lwander: Just to be clear, it is exactly storing the events you're pointing out in your sample `kubectl` commands.

wjoel: I see. It's a bit surprising (those are far from the kinds of manifests I'm used to) but it makes sense to keep the interface at that level. Just curious, then: why not a (sorted) list? We'll want to use the events sorted by time pretty much always, might as well generate a sorted list here?

lwander: Yup all resources follow that pattern.

Good point, I'll sort by `lastTimestamp`.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2ManifestProvider.java
```diff
@@ -73,6 +78,14 @@ public KubernetesV2Manifest getManifest(String account, String location, String
       return null;
     }
 
+    Function<KubernetesManifest, String> lastEventTimestamp = (m) -> (String) m.get("lastTimestamp");
```
wjoel: https://kubernetes.io/docs/reference/generated/federation/v1/definitions/#_v1_event lists this field as not required, so we should probably have a reasonable default.

lwander: Good point - I'll fall back to the "firstTimestamp", and then to "N/A" which'll sort higher than any timestamp

PR #2466, opened 2018-03-29
cleanup(aws): Avoid scheduling `ReconcileClassicLinkSecurityGroupsAgent`

ajordens:
Only schedule this agent for accounts where it is explicitly enabled.

This PR also cleans up some unnecessary usage of `e.printStackTrace()`.


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/OnDemandAgent.java
```diff
@@ -84,7 +88,7 @@ default Moniker convertOnDemandDetails(Map<String, String> details) {
           .sequence(Integer.valueOf(details.get("sequence")))
```
asher: While in here, also change `Integer.valueOf` to something null safe.

ajordens:
Oh ya, of course @asher! Thanks.

PR #2469, opened 2018-03-29
feat(provider/kubernetes): allow jobs & pods to "fail"

lwander:
@wjoel PTAL
bht6pbhfrhz


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesPodHandler.java
```diff
@@ -60,7 +63,34 @@ public SpinnakerKind spinnakerKind() {
...
+
+    // https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/
+    String phase = status.getPhase();
+
+    if (phase == null ) {
+      result.unstable("No phase reported yet")
+          .unavailable("No availability reported");
+    } else if (phase.equals("pending")) {
+      result.unstable("Pod is 'pending'")
+          .unavailable("Pod is still starting");
```
wjoel: "Pod is waiting to start" is how I'd put it. To me, `ContainerCreating` is the "still starting" phase, while `Pending` is "we haven't decided where to try starting this", but maybe that's just me.

lwander: No you're right, will change

PR #2482, opened 2018-04-04
fix(provider/openstack): corrects behavior of status checkers

brianberzins:
Co-authored-by: Gardnem6 <12258900+Gardnem6@users.noreply.github.com>
Co-authored-by: msilpala <msilpala@users.noreply.github.com>
Co-authored-by: esengstrom <37663918+esengstrom@users.noreply.github.com>

The underlying problem: we noticed a number of openstack operations failed/succeeded quickly on the openstack side, but those failures/successes were not always propagating through to spinnaker. Instead, spinnaker would continue to wait for deployments/deletes/resizes until they eventually timed out. The inconsistent behavior turned out to be whether or not the status was successful within the first check inside of a status-checking loop.

This looks like a large PR, but each of the following were done to fix the single aforementioned problem.

Use of the `BlockingStatusChecker` was inconsistent. In some cases it was assumed that the closure being called would be called within a loop, while other cases assumed it would be called only once. This pull request makes it so that all usages expect the closure to be called in the loop. Effectively, make the request to change once: then call the status checkers closure in a loop.

Moved the LoadBalancer checker into it's own class and created a new checker for Stack operations.

The openstack provider was inconsistent between whether or not calls to specific APIs should throw exceptions when a `null` is returned. This PR changes those behaviors to consistently allow `null` to be returned. When an exception should be thrown, it is now thrown from a separate `null` check.

Added a `StackConfig` analogous to the `LbaasConfig` to configure timeouts and polling intervals for operations to openstack "stack" APIs. Included default configuration for both. Reasonable defaults are included in case configuration files do not specify them.

A large number of test cases were updated to account for these changes. In most cases, the number of calls made to mocked objects changed based on the corrected looping logic. Additional tests were created for the extra logic of checking stack statuses.


brianberzins:
Additional comment about this. We've been running it in our production environment for almost two weeks. Other errors prevent us from being fully confident about the changes until recently.

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/config/OpenstackConfigurationProperties.groovy
```diff
@@ -35,14 +35,20 @@ class OpenstackConfigurationProperties {
...
 
   static class LbaasConfig {
-    Integer pollTimeout
-    Integer pollInterval
+    int pollTimeout = 60 // seconds
+    int pollInterval = 5 // seconds
+  }
+
+  static class StackConfig {
+    int pollTimeout = 600 // 10 minutes
```
lwander: I think `TimeUnit.MINUTES.toSeconds(10)` is maybe a little clearer

clouddriver-openstack/src/main/groovy/com/netflix/spinnaker/clouddriver/openstack/deploy/ops/servergroup/StackChecker.groovy
```diff
@@ -0,0 +1,40 @@
+package com.netflix.spinnaker.clouddriver.openstack.deploy.ops.servergroup
```
lwander: Missing a license header.

Also, new code should ideally be written in Java

msilpala: Oh? I missed the java preference. When did that happen?

brianberzins: This one might be weird. The rest of the clouddriver-openstack module is written in groovy. Is switching this one file over to Java what we want to do?

msilpala: Right at the top of the [Contributing doc](https://www.spinnaker.io/community/contributing/back-end-code/#java), FWIW. Wonder if it's been that way all along and we just missed it.

I'm _hoping_ we can let this one slide and promise to java in the future.

brianberzins: I also missed that other sub modules are mixing groovy/java source. If we're just talking about making it a java file inside the groovy source directory (as I saw in other sub modules) I have no problem with this.

brianberzins:
Updated PR based on comments, also re-squashed.

lwander:
Thanks! Looks like TravisCI timed out... kicking it off again

brianberzins:
and I'm out of date with master again...
update incoming

brianberzins:
rebasing again

msilpala:
@lwander

brianberzins:
@lwander Appreciate your timeliness and patience with getting this one merged! Thanks!

PR #2484, opened 2018-04-05
fix(core): Handle reading task results from previous redis delegate

ajordens:


ajordens:
```

ajordens:
Feels like this is the source of our problems when failing over.

PR #2488, opened 2018-04-05
fix(provider/gce): Support regional MIGs in validators.

jtk54:


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/validators/BasicGoogleDeployDescriptionValidator.groovy
```diff
@@ -46,7 +46,9 @@ class BasicGoogleDeployDescriptionValidator extends DescriptionValidator<BasicGo
                                 description.credentials)
 
     if (description.minCpuPlatform) {
-      helper.validateMinCpuPlatform(description.minCpuPlatform, description.zone, description.credentials)
+      helper.validateMinCpuPlatform(description.minCpuPlatform,
+                                    description.regional ? description.region: description.zone,
```
duftler: Missing a space just before the `:` here and below.

jtk54: Yep 2 for 2 :+1:

PR #2494, opened 2018-04-06
chore(build) single-thread ci builds

cfieber:
hoping for less/no cases of builds timing out with no output which may be
to too many things happening in parallel that none is making progress
fast enough


robzienert:
We should start doing builds on a cloud of X1's.

PR #2495, opened 2018-04-07
fix(openstack/javadoc): Add placeholder javadoc class comment to make…

duftler:
… gradle javadoc happy.

Otherwise, our build fails with:

```
:clouddriver-openstack:javadoc
javadoc: error - No public or protected classes found to document.
1 error
:clouddriver-openstack:javadoc FAILED
```


duftler:
fyi @lwander

PR #2500, opened 2018-04-10
fix(provider/ecs) Changes to allow ECS to work with AWS provider.

melbit-michaelw:
Fixes for [Setup issues for clouddriver-ecs](https://github.com/spinnaker/spinnaker/issues/2617).

Code now returns null if the account isn't an ECS account (i.e. no ECS Credentials exist), or if no server group is found (this is consistent with how the AWS provider seems to work).


robzienert:
This is a good spot. Throwing an exception is definitely the wrong behavior (whoops, shame on me for approving it!)

We _definitely_ do not want to swallow these exceptions without doing something, however. Let's get some logs in the catch statements so operators are notified of the issue (if it actually is an issue for someone).

melbit-michaelw:
@robzienert Anything further needed for this ?

BrunoCarrier:
Do other providers log when not finding anything for their provider, for a given application?  It seems that this would be overly verbose as not finding anything is actually a pretty standard case

btw, great job @melbit-michaelw !

melbit-michaelw:
Not sure if that comment was directed to me or not. As far as I could see from a quick look, the other providers don't log, but they also don't do their stuff via exception handling. Assuming there's a log.debug, I'm happy to lower the level of logging so only people who are actually after noise will get it, if that's a reasonable compromise ?

BrunoCarrier:
Totally agree on exceptions.  As for logging, it really isn't something relevant.  Imagine you had 6 cloud providers enabled and one provider per application, and 500 apps on Spinnaker.  The verbosity and usefulness of these log entries would be dubious

PR #2502, opened 2018-04-10
fix(google): Don't fetch image relationships

ezimanyi:
When listing the images for a Google account, we don't use any of the returned cache relationships, so skip fetching these from the cache entirely.


ezimanyi:
This should have a minor effect to improve the issue noted in https://github.com/spinnaker/spinnaker/issues/2585.  (It's addressing item (2) in https://github.com/spinnaker/spinnaker/issues/2585#issuecomment-379883858.)

PR #2508, opened 2018-04-12
fix(core): Optimize `CleanupPendingOnDemandCachesAgent`

ajordens:
Switched to using `exists` vs `get` with smaller partitions.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/agent/CleanupPendingOnDemandCachesAgent.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/agent/CleanupPendingOnDemandCachesAgent.java
@@ -16,6 +16,8 @@
 
 package com.netflix.spinnaker.clouddriver.core.agent;
 
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
 import com.netflix.spinnaker.cats.agent.RunnableAgent;
 import com.netflix.spinnaker.cats.module.CatsModule;
 import com.netflix.spinnaker.cats.provider.Provider;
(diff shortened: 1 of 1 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/agent/CleanupPendingOnDemandCachesAgent.java
```diff
@@ -88,33 +94,43 @@ void run(Collection<Provider> providers) {
...
        .collect(Collectors.toList());
 
-     List existingOnDemandKeys;
+     Map<String, Response<Boolean>> existingOnDemandKeys = new HashMap<>();
      if (redisClientDelegate.supportsMultiKeyPipelines()) {
-       existingOnDemandKeys = redisClientDelegate.withMultiKeyPipeline(pipeline -> {
-         onDemandKeys.forEach(k -> pipeline.get(provider.getProviderName() + ":onDemand:attributes:" + k));
-         return pipeline.syncAndReturnAll();
+       redisClientDelegate.withMultiKeyPipeline(pipeline -> {
+         for (List<String> partition : Iterables.partition(onDemandKeys, redisCacheOptions.getMaxDelSize())) {
```
ajordens: Could hardcode this to something small for the time being if we'd like -- it's not particularly critical

ajordens:
Does this seem better @robzienert @asher ?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/agent/CleanupPendingOnDemandCachesAgent.java
```diff
@@ -88,33 +94,43 @@ void run(Collection<Provider> providers) {
...
+           for (String id : partition) {
+             existingOnDemandKeys.put(id, pipeline.exists(provider.getProviderName() + ":onDemand:attributes:" + id));
```
ajordens: `exists` vs `get` must be better right?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/agent/CleanupPendingOnDemandCachesAgent.java
asher: We currently have `maxDelSize` set to 2 which makes me think we can do away with pipelineing here entirely for simplicity's sake, given that this cleanup agent isn't time sensitive. Totally fine to leave as-is though, should be a nice improvement.

ajordens: I didn't look close enough to see what the maxes were set to in the various profiles ... it looked like `maxPipelineSize` was pretty high which is why I err'd towards `maxDelSize`.

I'll stick with it as-is and see how it goes.

PR #2515, opened 2018-04-15
fix(provider/kubernetes): v2 Check status of workloads after update

wjoel:
Wait until the updated spec has been observed before trying to determine the status of the manifest.

This makes Spinnaker's results more consistent with the results from `kubectl rollout status`. See [kubernetes/pkg/kubectl/rollout_status.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/rollout_status.go)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesDeploymentHandler.java
```diff
@@ -119,14 +119,32 @@ private Status status(V1beta2Deployment deployment) {
       result.unavailable(available.getMessage());
     }
 
-    int desiredReplicas = deployment.getSpec().getReplicas();
+    Long generation = deployment.getMetadata().getGeneration();
+    Long observedGeneration = deployment.getStatus().getObservedGeneration();
+    if (observedGeneration == null || (generation != null && generation > observedGeneration)) {
```
lwander: It feels like this can be factored out by treating status as a `Map` or `ObjectMapping` it to something we know has an `ObservedGeneration` field. It'd be nice if each check for "stability" does this automatically.

wjoel: They don't check for `observedGeneration` being `nil` in rollout_status.go in the case of deployments (as they do for statefulset, and for `deployment.Spec.Replicas`), so it probably can't be null here. It's still a "probably" so I'd personally just keep the null check, but I can remove it if you prefer.

lwander: Sorry, I meant factoring the whole "wait until observed generation is at least generation" before proceeding into a separate shared method.

lwander: A static method seems reasonable -- `ObjectMapping` was a typo, `ObjectMapper` is what I meant :P 

There is even a utility: `KubernetesCacheDataConverter.convertToManifest(Object o)` to handle this for you.

I would also recommend against setting these intermediate states as "unstable", as something like "unknown" (with each `Condition` set to `null`) is arguably more accurate. I would say that setting any `Condition` means we have confidence in it.

wjoel: Thanks, I'll look into it!

Not sure if I follow the last bit, aren't we sure those are not stable states? Similar to how the current checks work, "Waiting for all replicas to be updated", "Waiting for all replicas to be available" etc. Are we any less sure of spec update or "Waiting for old replicas to finish termination" being unstable?

We should show the "Transitioning" banner with a message like "Waiting for old replicas to finish termination" in deck, I think?

lwander: The thing with the generation vs. observed generation is a client vs. server-side consistency issue. As long as the generation and observed generation are mismatched, we don't know what state the observed resource is in. It could be stable, unstable, failed, etc... Marking it as unstable feels like we'll introduce a lot of unnecessary flapping while we wait for the server's data to become consistent.

PR #2519, opened 2018-04-17
perf(titus): increasing page size for titus requests

emjburns:
In testing, we found that increasing page size dramatically reduced the titus caching time.


asher:
@emjburns @tomaslin we probably need to set `maxInboundMessageSize` for the OSS `GrpcChannelFactory`

tomaslin:
that's going to be a problem since it doesn't seem like this is supported by the default managedchannelbuilder.  https://grpc.io/grpc-java/javadoc/io/grpc/ManagedChannelBuilder.html

PR #2523, opened 2018-04-18
perf(caching): Support retrieving an individual on-demand record by id

ajordens:
Currently `orca` fetches all on-demand records and filters client-side
down to the specific record it is interested in.

This PR does two things:
- returns a set of typed identifiers as part of the original force cache refresh api
- allows for an `id` parameter to be supplied to the `/cache/{cloudProvider}/{type}` api

It currently supports a single `id` parameter as the expected use case
will be as part of the `ServerGroupForceCacheRefreshTask` which deals
with one server group at a time.


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy
```diff
@@ -126,4 +143,17 @@ class CatsOnDemandCacheUpdater implements OnDemandCacheUpdater {
       it.pendingOnDemandRequests(providerCache)
     }.flatten()
   }
+
+  @Override
+  Map pendingOnDemandRequest(OnDemandAgent.OnDemandType type, String cloudProvider, String id) {
```
ajordens: Not sure how we feel about returning `null` here or `Optional`.

I did it originally with `Optional` but it seemed a little clumsy to assemble everything downstream. 

Happy to revisit.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusClusterCachingAgent.groovy
```diff
@@ -375,16 +385,6 @@ class TitusClusterCachingAgent implements CachingAgent, CustomScheduledAgent, On
     }
   }
 
-
-  private void cacheTargetGroups(ServerGroupData data, Map<String, CacheData> targetGroups) {
```
ajordens: IntelliJ told me this wasn't being used

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/CacheController.groovy
```diff
@@ -36,19 +36,35 @@ class CacheController {
...
+      httpStatus
+    )
   }
 
   @RequestMapping(method = RequestMethod.GET, value = "/{cloudProvider}/{type}")
-  Collection<Map>  pendingOnDemands(@PathVariable String cloudProvider,
-                                    @PathVariable String type) {
+  Collection<Map> pendingOnDemands(@PathVariable String cloudProvider,
+                                   @PathVariable String type,
+                                   @RequestParam(name = "id", required = false) String id) {
```
ajordens: Any thought on this as a `RequestParam` vs `PathVariable`?

The latter would look something like `/{cloudProvider}/{type}/{id}`. I actually did this originally but figured it might be more annoying to deal with in `orca` (ie. empty collection vs. 404) and went back to it just being a filter. 

Happy to revisit.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy
dreynaud: null is fine IMO but I'd add `@Nullable`

ajordens: 👍

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusClusterCachingAgent.groovy
jrsquared: I would think it should be... defer to @tomaslin here.

PR #2524, opened 2018-04-18
fix(provider/kubernetes): harden check for default namespace

lwander:


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -118,13 +117,15 @@ public String lookupDefaultNamespace() {
...
       Optional<String> serviceAccountNamespace = Files.lines(serviceAccountNamespacePath, StandardCharsets.UTF_8).findFirst();
-      namespace = serviceAccountNamespace.get();
-    } catch (IOException | NoSuchElementException e) {
+      namespace = serviceAccountNamespace.orElse("");
+    } catch (IOException e) {
       try {
         namespace = jobExecutor.defaultNamespace(this);
       } catch (KubectlException ke) {
-        log.warn("Failure looking up desired namespace, defaulting to {}", namespace);
+        log.warn("Failure looking up desired namespace, defaulting to {}", namespace, e);
```
duftler: Is the `e` definitely a `Throwable` here (and below) and not just an additional optional argument? Or is the logger smart about always rendering an argument that is of type `Throwable`?

lwander: The logger is smart about taking the last argument and rendering a stack trace

lwander: * if it can

PR #2531, opened 2018-04-18
fix(docker): Do not cascade docker registry failures to health indicator

robzienert:
Making docker health indicator act like the AmazonHealthIndicator in that once it's healthy, it's always healthy. This will allow us to still gate the deployment of a clouddriver instance based on whether or not it can communicate with a docker registry, but not take clouddriver down if the docker registry goes unhealthy sometime afterwards.

Made the logic generic, so other health indicators can get the same sort of behavior.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/AlwaysUpHealthIndicator.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/AlwaysUpHealthIndicator.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/AlwaysUpHealthIndicator.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private static final Logger log = LoggerFactory.getLogger(AlwaysUpHealthIndicator.class);
+
+  protected final AtomicReference<Exception> lastException = new AtomicReference<>(null);
+  protected final AtomicReference<Boolean> hasInitialized = new AtomicReference<>(null);
+
+  private final AtomicLong errors;
+
+  public AlwaysUpHealthIndicator(Registry registry) {
+    this.errors = registry.gauge("health.amazon.errors", new AtomicLong(0));
```
cfieber: maybe stick a class tag on this gauge

cfieber: either that or the gauge name itself (probably not health.amazon at least)

robzienert: Oh, quite. A little too much copypasta in this place.

PR #2535, opened 2018-04-19
perf(provider/kubernetes): make one "large" caching agent

lwander:
It's worth pointing out that now it's more important to make use of the
"cacheThreads" flag in the account config simply because the default of
1 thread means at most one clouddriver can handle caching a single
account at any time.

Directly addresses: https://github.com/spinnaker/spinnaker/issues/2644
Is on the way to addressing: https://github.com/spinnaker/spinnaker/issues/2670


lwander:
@wjoel FYI

lwander:
This reduces the number of kubectl calls by a factor of roughly 25

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesCoreCachingAgent.java
```diff
@@ -0,0 +1,108 @@
...
+    types.add(AUTHORITATIVE.forType(Keys.LogicalKind.CLUSTERS.toString()));
+
+    types.addAll(primaryKinds().stream().map(k -> AUTHORITATIVE.forType(k.toString())).collect(Collectors.toList()));
+
+    return Collections.unmodifiableSet(new HashSet<>(types));
+  }
+
+  @Override
+  protected List<KubernetesKind> primaryKinds() {
+    return Arrays.asList(
```
danielpeach: Ideally this list would be generated such that any dynamically registered `KubernetesKind` would be included here as well.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesCoreCachingAgent.java
```diff
@@ -45,20 +47,23 @@
...
+    types.addAll(primaryKinds().stream().map(k -> AUTHORITATIVE.forType(k.toString())).collect(Collectors.toList()));
+
+    return Collections.unmodifiableSet(new HashSet<>(types));
   }
 
   @Override
-  protected KubernetesKind primaryKind() {
-    return KubernetesKind.ROLE;
+  protected List<KubernetesKind> primaryKinds() {
+    synchronized (KubernetesKind.getValues()) {
```
jtk54: Is there a reason this `synchronized` isn't scoped to the function?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesStatefulSetHandler.java
```diff
@@ -138,4 +146,29 @@ private Status status(V1beta2StatefulSet statefulSet) {
 
     return result;
   }
+
+  @Override
+  public void addRelationships(Map<KubernetesKind, List<KubernetesManifest>> allResources, Map<KubernetesManifest, List<KubernetesManifest>> relationshipMap) {
+    BiFunction<String, String, String> manifestName = (namespace, name) -> namespace + ":" + name;
```
jtk54: Used more than once or twice, might warrant its own util function.

jtk54:
Welp.

PR #2537, opened 2018-04-19
fix(titus): add retry to kill task, ignore NOT_FOUND

emjburns:
We're seeing occasional failures when a titus task cannot be found, and causes the operation to fail. If the task is terminate and shrink it's ok to ignore NOT_FOUND, but otherwise we should retry in case of transient failures.

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/v3client/RegionScopedV3TitusClient.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/v3client/RegionScopedV3TitusClient.java
@@ -21,6 +21,7 @@ import com.google.protobuf.Message;
 import com.google.protobuf.util.JsonFormat;
 import com.netflix.frigga.Names;
 import com.netflix.spectator.api.Registry;
+import com.netflix.spinnaker.clouddriver.titus.TitusException;
 import com.netflix.spinnaker.clouddriver.titus.client.TitusClient;
 import com.netflix.spinnaker.clouddriver.titus.client.TitusClientObjectMapper;
 import com.netflix.spinnaker.clouddriver.titus.client.TitusJobCustomizer;
@@ -213,12 +214,12 @@ public class RegionScopedV3TitusClient implements TitusClient {
           killTaskWithRetry(id, terminateTasksAndShrinkJob);
         } catch (Exception e) {
           failedTasks.add(id);
-          log.error("Failed to terminate and shrink titus task {}. Exception: {}", id, e);
+          log.error("Failed to terminate and shrink titus task {} in account {} and region {}", id, titusRegion.getAccount(), titusRegion.getName(), e);
         }
       }
     );
-    if (!failedTasks.isEmpty() && failedTasks.size() > 0) {
-      throw new RuntimeException("Failed to terminate and shrink titus tasks: " + StringUtils.join(failedTasks, ","));
+    if (!failedTasks.isEmpty()) {
+      throw new TitusException("Failed to terminate and shrink titus tasks: " + StringUtils.join(failedTasks, ","));
     }
   }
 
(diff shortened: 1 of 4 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/v3client/RegionScopedV3TitusClient.java
```diff
@@ -195,10 +198,24 @@ public void terminateJob(TerminateJobRequest terminateJobRequest) {
   @Override
   public void terminateTasksAndShrink(TerminateTasksAndShrinkJobRequest terminateTasksAndShrinkJob) {
     terminateTasksAndShrinkJob.getTaskIds().forEach(id ->
-      grpcBlockingStub.killTask(TaskKillRequest.newBuilder().setTaskId(id).setShrink(terminateTasksAndShrinkJob.isShrink()).build())
+      killTaskWithRetry(id, terminateTasksAndShrinkJob)
```
ajordens: might be worth doing some exception handling here such that the first failure doesn't stop at least attempting to destroy all the others.

emjburns: Ah great point. Will add that. Then, we can still fail at the end but only the ones with failures will have not been terminated.

tomaslin:
LGTM

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/v3client/RegionScopedV3TitusClient.java
```diff
@@ -197,9 +204,33 @@ public void terminateJob(TerminateJobRequest terminateJobRequest) {
...
   public void terminateTasksAndShrink(TerminateTasksAndShrinkJobRequest terminateTasksAndShrinkJob) {
-    terminateTasksAndShrinkJob.getTaskIds().forEach(id ->
-      grpcBlockingStub.killTask(TaskKillRequest.newBuilder().setTaskId(id).setShrink(terminateTasksAndShrinkJob.isShrink()).build())
+    List<String> failedTasks = new ArrayList<>();
+    terminateTasksAndShrinkJob.getTaskIds().forEach(id -> {
+        try {
+          killTaskWithRetry(id, terminateTasksAndShrinkJob);
+        } catch (Exception e) {
+          failedTasks.add(id);
+          log.error("Failed to terminate and shrink titus task {}. Exception: {}", id, e);
```
ajordens: I don't know that you need `. Exception: {}` ... could just `, e` and have the the stacktrace included as normal?

Might be worth including the region and account if we have access to it.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/v3client/RegionScopedV3TitusClient.java
```diff
@@ -197,9 +204,33 @@ public void terminateJob(TerminateJobRequest terminateJobRequest) {
...
+        }
+      }
     );
+    if (!failedTasks.isEmpty() && failedTasks.size() > 0) {
```
ajordens: if `failedTasks` isn't empty, it's going to have > 0 elements in it ... second part of this condition shouldn't be necessary?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/v3client/RegionScopedV3TitusClient.java
```diff
@@ -197,9 +204,33 @@ public void terminateJob(TerminateJobRequest terminateJobRequest) {
...
+      throw new RuntimeException("Failed to terminate and shrink titus tasks: " + StringUtils.join(failedTasks, ","));
```
ajordens: _might_ be worth a general purpose TitusException vs RuntimeException just as a matter of distinction.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/v3client/RegionScopedV3TitusClient.java
```diff
@@ -67,6 +71,9 @@
 
   private final JobManagementServiceGrpc.JobManagementServiceBlockingStub grpcBlockingStub;
 
+  @Autowired
```
ajordens: Can we move this into the constructor ... this class is constructor created so the `@Autowired` stands out as strange.

PR #2539, opened 2018-04-19
perf(aws): Optimize the retrieval of all security groups

ajordens:
When `includeRules` is `false`, the cache key for a security group
actually includes all relevant information.

We _do not_ need to retrieve all objects and can instead just parse
keys and manually construct an `AmazonSecurityGroup`.

There is _some_ chance that the set of identifiers includes some
security groups that do not actually exist. I believe this to be of
minimal risk and the performance benefits outweigh the potential for
issue.

The return types in `SecurityGroupProvider` were changed from `Set` to
`List`. It turns out that converting a list of 20k items to a set takes
a non-zero amount of time.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonSecurityGroupProvider.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonSecurityGroupProvider.groovy
@@ -64,27 +64,42 @@ class AmazonSecurityGroupProvider implements SecurityGroupProvider<AmazonSecurit
   }
 
   @Override
-  Set<AmazonSecurityGroup> getAll(boolean includeRules) {
-    getAllMatchingKeyPattern(Keys.getSecurityGroupKey('*', '*', '*', '*', '*'), includeRules)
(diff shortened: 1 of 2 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonSecurityGroupProvider.groovy
```diff
@@ -64,27 +64,42 @@ class AmazonSecurityGroupProvider implements SecurityGroupProvider<AmazonSecurit
   }
 
   @Override
-  Set<AmazonSecurityGroup> getAll(boolean includeRules) {
-    getAllMatchingKeyPattern(Keys.getSecurityGroupKey('*', '*', '*', '*', '*'), includeRules)
+  Collection<AmazonSecurityGroup> getAll(boolean includeRules) {
+    if (!includeRules) {
+      def identifiers = cacheView.getIdentifiers(SECURITY_GROUPS.ns)
```
ajordens: @anotherchrisberry actually did this as part of https://github.com/spinnaker/clouddriver/pull/2533 ... my contribution was testing and realization that `List` -> `Set` actually takes ~6s given our volume of data.

PR #2541, opened 2018-04-20
revert(titus): reverting #2534

emjburns:
Reverting to try and narrow down the performance impact on titus.


tomaslin:
ugh, fine

PR #2542, opened 2018-04-20
fix(titus): revert netty but keep titusRegion.url change

emjburns:
Reverting netty channel builder but keeping the property change of `.url`


tomaslin:
lgtm

PR #2544, opened 2018-04-20
feat(titus): adding feature flag in region scoped titus client

emjburns:
Titus Region Client can now behave differently in different Titus regions, so we can slowly roll out new client features.

Config option is at the region level, just like `apiVersion`, and is controlled with Boolean flag `experimentalFeaturesEnabled`

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/TitusRegion.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/TitusRegion.java
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.clouddriver.titus.client;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
@@ -30,7 +31,7 @@ public class TitusRegion {
   private final String applicationName;
   private final String url;
   private final int port;
-  private final Boolean experimentalFeaturesEnabled;
+  private final List<String> featureFlags;
 
   private <T> T notNull(T val, String name) {
     if (val == null) {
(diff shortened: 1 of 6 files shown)


tomaslin:
lgtm

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/TitusRegion.java
```diff
@@ -30,6 +30,7 @@
   private final String applicationName;
   private final String url;
   private final int port;
+  private final Boolean experimentalFeaturesEnabled;
```
tomaslin: Can we make this into a string or list called featureFlags? 

If this was a list, then we can enabled featureFlags like `skipJobIds,useInstanceCachingAgent`, etc vs having to have it be all or nothing

emjburns: Sure, I'll make that change. That's a great idea.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/v3client/RegionScopedV3TitusClient.java
```diff
@@ -96,6 +97,13 @@ public RegionScopedV3TitusClient(TitusRegion titusRegion,
 
     }
     this.grpcBlockingStub = JobManagementServiceGrpc.newBlockingStub(channelFactory.build(titusRegion, environment, eurekaName, DEFAULT_CONNECT_TIMEOUT, registry));
+
+    if (!titusRegion.getFeatureFlags().isEmpty()) {
+      log.info("Experimental Titus V3 client features {} enabled for account {} and region {}",
```
tomaslin: probably change the message to say feature flags

tomaslin: otherwise lgtm

PR #2557, opened 2018-04-24
fix(titus): fix load balancing

tomaslin:
handle attach and detach of load balancers during disable / enable operations


emjburns:
Looks fine to me

PR #2567, opened 2018-04-27
fix(provider/kubernetes): support key ref secrets & cms (#2566)

lwander:
cherry pick into 1.7


lwander:
Up to you @sbwsg -- I don't have a good sense for how many more fixes we expect to pick up in the next few days

PR #2574, opened 2018-04-27
feat(provider/kubernetes): v2 cluster is optional

lwander:


lwander:
closes: https://github.com/spinnaker/spinnaker/issues/2258

PR #2580, opened 2018-04-30
Titus: container attributes

tomaslin:


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
```diff
@@ -144,6 +144,12 @@ class TitusDeployHandler implements DeployHandler<TitusDeployDescription> {
           }
           sourceJob.labels.each { k, v -> description.labels.put(k, v) }
         }
+        if (!description.containerAttributes || description.containerAttributes.isEmpty()) {
+          if (!description.containerAttributes) {
+            description.containerAttributes = [:]
+          }
+          sourceJob.containerAttributes.each { k, v -> description.containerAttributes.put(k, v) }
```
emjburns: I think this should be outside the parent `if` so that it is run if there are containerAttributes (not just if it's an empty map)

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
```diff
@@ -175,6 +181,7 @@ class TitusDeployHandler implements DeployHandler<TitusDeployDescription> {
       task.updateStatus BASE_PHASE, "Resolved server group name to ${nextServerGroupName}"
 
       if (!description.env) description.env = [:]
+      if (!description.containerAttributes) description.containerAttributes = [:]
```
emjburns: This case seems to be handled above, this statement seems redundant

emjburns:
LGTM after comments

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
tomaslin: we're copying from the source asg here

PR #2584, opened 2018-05-01
feat(provider/dcos) Adding various features for upstream

TrevinTeacutter:
This includes support for Marathon 1.5, Marathon on Marathon, and a fix necessary for our JFrog artifactory to work with the docker registry catalog paging (that should be passive).


willgorman:
Can you split https://github.com/spinnaker/clouddriver/pull/2584/commits/187f795778e8f486e008218fe5d3a165732310a1 into two commits with commit messages following the convention?  

Looks like there are some test failures on https://github.com/spinnaker/clouddriver/pull/2584/commits/99732a80daebc2d129f87583dbaa7e8b637e7ed0 that need to be addressed.

willgorman:
@spinnaker/reviewers could I get another set of eyes on d0e284f since that's outside the dc/os provider?

willgorman:
Confirmed that the change in https://github.com/spinnaker/clouddriver/commit/d0e284f5eb570fe01cf44545a1e14b37dcc17363 to handle the paging links returned from Artifactory is passive with other registries (index.docker.io specifically).

Everything else for DC/OS looks good.

PR #2585, opened 2018-05-01
fix(provider/kubernetes): fix automatically assigned moniker cluster

lwander:
Closes https://github.com/spinnaker/spinnaker/issues/2730


ezimanyi:
LGTM...looks like there's a parameter `@Builder(toBuilder=true)` that lets you initialize a builder from an instance that might save some boilerplate (for when you move this to kork).

PR #2586, opened 2018-05-01
fix(provider/kubernetes): fix NPE in status check

lwander:


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesDeploymentHandler.java
```diff
@@ -139,16 +139,16 @@ private Status status(V1beta2Deployment deployment) {
...
     }
 
-    if (status.getAvailableReplicas() < existing) {
+    if (status.getAvailableReplicas() == null || status.getAvailableReplicas() < existing) {
       return result.unstable("Waiting for all replicas to be available");
     }
 
     existing = status.getReadyReplicas();
-    if (existing == null || desiredReplicas > existing) {
+    if (existing == null || desiredReplicas == null || desiredReplicas > existing) {
```
wjoel: I think we should skip this check if `replicas` is not included in the manifest.

lwander: Why is that? If it's not included we can still see how the existing replica count compares to the desired replica count, and if either don't exist something is wrong.

wjoel: With `status.getReadyReplicas() == 1` and `desiredReplicas = deployment.getSpec().getReplicas()` being `null`, say, it seems this would return `result.unstable("Waiting for all replicas to be ready");` forever. I think that's what I had in mind.

So I was thinking `if (existing == null || (desiredReplicas != null && desiredReplicas > existing))`, but maybe I'm not thinking this through correctly. It's been a long day.

lwander: Ah I see what you mean -- I don't think desired replicas could be empty on a running, stable manifest. I think even if a user can skip it Kubernetes will fill in some default once the manifest is accepted. 

Let me double check that before going one way or the other, though.

wjoel: But if that's true, isn't the null check you're adding here redundant?

lwander: I'm worried about the window between the master accepting the manifest and whatever changes it makes filling the logs with NPEs

wjoel: I think https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/rollout_status.go#L85 implies that our `desiredReplicas` can be null (so you're correct in adding the check) but it has an `&&` for the subsequent test.

lwander: I'm not sure how much good that does since the following clause ignores the null case: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/rollout_status.go#L88

wjoel: That line checks `deployment.Status.Replicas`, not `deployment.Spec.Replicas` ;)

lwander: whoops ;) switching over to a && now...

PR #2591, opened 2018-05-02
feat(artifacts): Add custom API endpoint for S3 client

bchbchbch:
Currently the S3 client does not support custom API endpoints. This PR adds `apiEndpoint` and `apiRegion` fields to `artifacts.s3.accounts`, and adds them to the S3 client.


lwander:
Could you make a symmetric halyard change?

PR #2594, opened 2018-05-03
feat(provider/kubernetes): allow user to set versioning behavior in annotations

lwander:
Closes https://github.com/spinnaker/spinnaker/issues/2351


lwander:
Docs: https://github.com/spinnaker/spinnaker.github.io/pull/726

PR #2598, opened 2018-05-03
Retry submits on titus

tomaslin:


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
```diff
@@ -290,8 +290,28 @@ class TitusDeployHandler implements DeployHandler<TitusDeployDescription> {
...
+          task.updateStatus BASE_PHASE, "Error encountered submitting job request to Titus ${e.status} ${e.message}"
+          if (e.status == Status.RESOURCE_EXHAUSTED || (e.status == Status.INVALID_ARGUMENT)) {
+            nextServerGroupName = resolveJobName(description, submitJobRequest, task, titusClient)
+            throw e;
+          }
+          if (e.status == Status.UNAVAILABLE) {
+            throw e;
+          }
+        }
+      }, 6, 1000, true)
```
ajordens: do you want to do exponential here? 

as written this would be 6 requests over ~30s ... is that better than 30 or 15 requests with 1s or 2s in between?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
```diff
@@ -290,8 +290,28 @@ class TitusDeployHandler implements DeployHandler<TitusDeployDescription> {
...
+
+      if (jobUri == null) {
+        throw new TitusException("Could not create job")
```
ajordens: any additional context we can include in this exception message?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/RegionScopedTitusClient.java
```diff
@@ -252,6 +252,10 @@ public TitusHealth getHealth() {
   }
 
   private List<Job> getJobs(JobQuery.Builder jobQuery) {
+    return getJobs(jobQuery, true);
+  }
+
+  private List<Job> getJobs(JobQuery.Builder jobQuery, boolean includeTasks) {
```
ajordens: Neato -- want to also add an `includeTasks` for `findJobByName`?

The current call in `TitusClusterCachingAgent` could be swapped over to make use of it.

Also happy to do this as a subsequent commit.

PR #2599, opened 2018-05-04
perf(aws): Optimize how ON_DEMAND records are fetched during caching cycles

ajordens:
Given that the cardinality of ON_DEMAND is quite small, it is _possibly_
more efficient to fetch the ON_DEMAND records for a given account +
region and perform an in-memory filter.

Previously we were doing an mget with many thousands of keys (one per
server group) when only a small number (dozens?) would ever exist in the
set.


ajordens:
I still need to actually test this ... consider it a request for feedback.

PR #2600, opened 2018-05-04
perf(provider/titus): do not load tasks when finding by name unless needed

tomaslin:


emjburns:
LGTM

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/RegionScopedTitusClient.java
```diff
@@ -114,15 +114,20 @@ public Job getJobAndAllRunningAndCompletedTasks(String jobId) {
...
       .putFilteringCriteria("attributes", "source:spinnaker,name:" + jobName)
       .putFilteringCriteria("attributes.op", "and");
-    List<Job> results = getJobs(jobQuery);
+    List<Job> results = getJobs(jobQuery, includeTasks);
     return results.isEmpty() ? null : results.get(0);
   }
 
+  @Override
+  public Job findJobByName(String jobName) {
+    return findJobByName(jobName, false);
```
ajordens: If the default is `false` -- you'll want to remove the `minimalOnDemand` feature flag from `TitusClusterCachingAgent`.

As it stands, with this change, accounts other than `titusprodvpc` and `titustestvpc` would be doing a proper cache refresh with a job sans tasks -- perhaps not ideal.

ajordens: I was going to remove the feature flag anyhow and make it the default -- so either do it as part of this PR -- or I can have a separate PR that just needs to be released side-by-side.

ajordens: https://github.com/spinnaker/clouddriver/pull/2602

PR #2602, opened 2018-05-04
chore(titus): Remove `legacyOnDemand` from `TitusClusterCachingAgent`

ajordens:
The new `minimalOnDemand` feature flag has been active for a week or so
and looks good.

This removes the legacy behavior.


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusClusterCachingAgent.groovy
```diff
@@ -154,11 +154,7 @@ class TitusClusterCachingAgent implements CachingAgent, CustomScheduledAgent, On
       titusClient.findJobByName(data.serverGroupName as String)
     }
 
-    if (titusRegion.featureFlags.contains("minimalOnDemand")) {
-      return minimalOnDemand(providerCache, job, data)
-    }
-
-    return legacyOnDemand(providerCache, job, data)
+    return minimalOnDemand(providerCache, job, data)
```
tomaslin: worth renaming to onDemand

PR #2614, opened 2018-05-08
perf(search): improves search performance

cfieber:
Within CatsSearchProvider only searches against
the specific caches for providers that host a
caching agent of the desired type.

Adds an existingIds to the Cache interface to
verify that key hits are still existing cache
items (previously would fully load each object
to verify existance)

Better search globbing if the SearchableProvider
defines a KeyParser.

Makes TitusCachingProvider a SearchableProvider.

depends on spinnaker/kork#152


build.gradle
```diff
@@ -40,7 +40,8 @@ allprojects {
   apply plugin: 'groovy'
 
   ext {
-    spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '0.151.0'
+    spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '0.154.0-SNAPSHOT'
```
cfieber: will fix once I've released the new kork

cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/CompositeCache.java
```diff
@@ -70,6 +70,21 @@ public CacheData get(String type, String id, CacheFilter cacheFilter) {
         return allItems.values();
     }
 
+    @Override
+    public Collection<String> existingIdentifiers(String type, Collection<String> ids) {
+        HashSet<String> identifiers = new HashSet<>(ids.size());
```
ajordens: wtb 2 space indent (or maybe cats is a 4 spacer and we should be consistent in this method)

cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/CompositeCache.java
```diff
@@ -70,6 +70,21 @@ public CacheData get(String type, String id, CacheFilter cacheFilter) {
...
+        HashSet<String> remainingIds = new HashSet<>(ids);
+        for (Cache cache : caches) {
+          Collection<String> existing = cache.existingIdentifiers(type, remainingIds);
+          identifiers.addAll(existing);
+          remainingIds.removeAll(existing);
```
ajordens: maybe a comment in here that we're basically short-circuiting once we've found a match for an `id` in one of the caches?

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cache/AbstractRedisCache.java
```diff
@@ -99,6 +102,23 @@ public CacheData get(String type, String id, CacheFilter cacheFilter) {
...
   }
 
+  @Override
+  public Collection<String> existingIdentifiers(String type, Collection<String> identifiers) {
+    final Map<String, Response<Boolean>> responses = new LinkedHashMap<>();
+    redisClientDelegate.withPipeline(p -> {
+      for (String id : identifiers) {
+        responses.put(id, p.exists(attributesId(type, id)));
+      }
+      redisClientDelegate.syncPipeline(p);
```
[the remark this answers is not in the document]
ajordens: I dig it -- probably same comment about including an inline comment

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/SearchableProvider.java
```diff
@@ -50,6 +54,42 @@
...
   Map<String, String> parseKey(String key);
 
+  default Optional<KeyParser> getKeyParser() {
+    return Optional.empty();
+  }
+  /**
+   *
+   */
+  default String buildSearchTerm(String type, String queryTerm) {
+    String prefix = getKeyParser().map(KeyParser::getCloudProvider).orElse("*");
```
ajordens: We have a legitimate `KeyParser` w/o a cloud provider I guess?

ajordens: IMO more inline comments in here too

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cache/AbstractRedisCache.java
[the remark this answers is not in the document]
cfieber: not really sure this one warrants a comment, its just the way to use withPipeline (previous alternative was doing something like ((Pipeline) pipeline).sync() which didn't play nice in an abstract user of RedisClientDelegate

cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/CompositeCache.java
cfieber: was consistent with the existing file, don't really want to push a reformat in this changeset

ajordens: that's fair ... it just stood out because the loop below is 2 space indented. meh.

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cache/AbstractRedisCache.java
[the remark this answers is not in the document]
ajordens: ya sorry ... don't think I was looking for a comment on `syncPipeline` specifically.

on a second read, the rest is alright --

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/SearchableProvider.java
cfieber: getKeyParser itself is optional, most cloud provider Keys don't currently implement that interface

ajordens: 👍

PR #2620, opened 2018-05-10
fix(titus): respect applicationDefaultSecurityGroup when cloning

tomaslin:
persists the decision to not create an applicationDefaultSecurityGroup in titus attribute


ajordens:
LGTM

PR #2629, opened 2018-05-14
provider/kubernetes: remove versioning of HorizontalPodAutoscalers

mattnworb:
When deploying updates to a HorizontalPodAutoscaler, versioning them
will create multiple instances of the HPA, all of which still try to
target/autoscale the scaleTargetRef. Since I don't believe there are
other resource types that refer to a HPA, it is not clear that this
needs to be versioned like ConfigMaps or Secrets.


davidxia:
cc @lwander

PR #2630, opened 2018-05-14
feat(artifacts): add gitlab artifact provider

ethanfrogers:
adds a gitlab artifact provider. follows most of the same logic in the
github provider. relies on the gitlab raw file api.

@lwander PTAL


ethanfrogers:
I just did this for fun because it looked easy. Let me know what you think. I'll add some docs in a little bit.

PR #2631, opened 2018-05-15
fix(artifacts): s3 improvements

ethanfrogers:
this commit fixes two issues with the s3 artifact provider. 1, it adds
support for configuring an s3 region. this is inline with how we already
handle it in front50, allowing for usage of buckets in separate regions.
also, when using `apiEndpoint` we not set `pathStyleAccessEnabled` which
enables support for things like Minio to act as a provider.

@lwander @benjaminws PTAL!


ethanfrogers:
also, @andrewbackes and @imosquera

ethanfrogers:
@benjaminws quick question: will you need to update your config to reflect the addition of `region` as a required config property? would it be helpful if i set some default value? now that i look at it, this isn't backwards compatible.

benjaminws:
Gah, that’s a great point. To be honest I thought the default was always
set to us-east-1 with the aws libs?

benjaminws:
Sounds good to me

ethanfrogers:
Fixed. If `region` is non-empty the region will be set. Easy fix. Thanks for the review!

PR #2632, opened 2018-05-15
fix(search) less redis interaction on instance cache

cfieber:
removes the existingIdentifiers call on the in-memory instance cache

as this cache is loaded per clouddriver instance it can balloon out to a
large number of exists calls


ajordens:
LGTM

PR #2636, opened 2018-05-17
feat(titus): split caching into two agents

emjburns:
(I'll update this PR with performance numbers, opening for feedback)

Only controller API call that produces different responses is this set:
```
https://localhost:8081/applications/emburnstest/clusters/titustestvpc/emburnstest-responses?expand=false
https://localhost:8081/applications/emburnstest/clusters/titustestvpc/emburnstest-responses?expand=true
```
In the current implementation, both return instance details. This PR changes the response so that when `expand=false` the `instances : []` list is empty. It is populated when `expand=true`.

Also, this PR fixes titus instance searching (it is broken in the current Titus implementation due to incorrect Key parsing)

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsSearchProvider.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsSearchProvider.groovy
@@ -16,7 +16,6 @@
 
 package com.netflix.spinnaker.clouddriver.cache
 
-import com.netflix.spinnaker.cats.agent.CachingAgent
 import com.netflix.spinnaker.cats.cache.Cache
 import com.netflix.spinnaker.cats.provider.ProviderRegistry
 import com.netflix.spinnaker.clouddriver.search.SearchProvider
@@ -114,28 +113,32 @@ class CatsSearchProvider implements SearchProvider, Runnable {
    */
   @Override
   void run() {
-    log.info("Refreshing Cached Identifiers (instances)")
-    def instanceIdentifiers = providers.findAll { provider ->
-      provider.supportsSearch('instances', Collections.emptyMap())
-    }.collect { provider ->
-      def cache = providerRegistry.getProviderCache(provider.getProviderName())
-      return cache.getIdentifiers("instances").findResults { key ->
-        def v = provider.parseKey(key)
-        if (v) {
-          v["_id"] = key
-        }
+    try {
+      log.info("Refreshing Cached Identifiers (instances)")
+      def instanceIdentifiers = providers.findAll { provider ->
+        provider.supportsSearch('instances', Collections.emptyMap())
+      }.collect { provider ->
+        def cache = providerRegistry.getProviderCache(provider.getProviderName())
+        return cache.getIdentifiers("instances").findResults { key ->
+          def v = provider.parseKey(key)
+          if (v) {
+            v["_id"] = key
+          }
 
-        return v?.collectEntries {
-          [it.key, it.value.toLowerCase()]
+          return v?.collectEntries {
+            [it.key, it.value.toLowerCase()]
+          }
         }
+      }.flatten()
+
+      if (instanceIdentifiers) {
+        cachedIdentifiersByType.set(["instances": instanceIdentifiers])
       }
-    }.flatten()
 
-    if (instanceIdentifiers) {
-      cachedIdentifiersByType.set(["instances": instanceIdentifiers])
+      log.info("Refreshed Cached Identifiers (found ${instanceIdentifiers.size()} instances)")
+    } catch (Exception e) {
+      log.error("Unable to refresh cached identifiers", e)
     }
-
-    log.info("Refreshed Cached Identifiers (found ${instanceIdentifiers.size()} instances)")
   }
 
   @Override
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsSearchProvider.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsSearchProvider.groovy
@@ -137,7 +137,7 @@ class CatsSearchProvider implements SearchProvider, Runnable {
 
       log.info("Refreshed Cached Identifiers (found ${instanceIdentifiers.size()} instances)")
     } catch (Exception e) {
-      log.error("Unable to refresh cached identifiers", e)
+      log.error("Unable to refresh cached identifiers (instances)", e)
     }
   }
 
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/provider/agent/Namespace.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/provider/agent/Namespace.groovy
@@ -38,7 +38,7 @@ enum Namespace {
   RESERVATION_REPORTS,
   RESERVED_INSTANCES
 
-  final String ns
+  public final String ns
   final Set<String> fields
 
   private Namespace(List<String> keyFields = []) {
@@ -48,5 +48,9 @@ enum Namespace {
 
   String toString() {
     ns
+  }
+
+  String getNs() {
+    return ns
   }
 }
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/Keys.groovy
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/Keys.groovy
@@ -18,8 +18,8 @@ package com.netflix.spinnaker.clouddriver.titus.caching
 
 import com.netflix.frigga.Names
 import com.netflix.spinnaker.clouddriver.cache.KeyParser
-import com.netflix.spinnaker.clouddriver.titus.TitusCloudProvider
 import org.springframework.stereotype.Component
+import com.netflix.spinnaker.clouddriver.titus.TitusCloudProvider
 
 @Component("titusKeyParser")
 class Keys implements KeyParser {
@@ -33,7 +33,7 @@ class Keys implements KeyParser {
     HEALTH,
     ON_DEMAND
 
-    final String ns
+    public final String ns
 
     private Namespace() {
       def parts = name().split('_')
@@ -66,6 +66,15 @@ class Keys implements KeyParser {
   }
 
 
+  static enum CachingSchema {
+    V1,
+    V2
+
+    String toLowerCase() {
+      return name().toLowerCase()
+    }
+  }
+
   static Map<String, String> parse(String key) {
     def parts = key.split(':')
 
@@ -77,37 +86,88 @@ class Keys implements KeyParser {
       return null
     }
 
-    def result = [provider: parts[0], type: parts[1]]
+    def result
+    result = [provider: parts[0], type: parts[1]]
 
-    switch (result.type) {
-      case Namespace.IMAGES.ns:
-        result << [account: parts[2], region: parts[3], imageId: parts[4]]
-        break
-      case Namespace.SERVER_GROUPS.ns:
-        def names = Names.parseName(parts[5])
-        result << [application: names.app.toLowerCase(), cluster: parts[2], account: parts[3], region: parts[4], serverGroup: parts[5], stack: names.stack, detail: names.detail, sequence: names.sequence?.toString()]
-        break
-      case Namespace.INSTANCES.ns:
-        result << [id: parts[2]]
-        break
-      case Namespace.CLUSTERS.ns:
-        def names = Names.parseName(parts[4])
-        result << [application: parts[2].toLowerCase(), account: parts[3], cluster: parts[4], stack: names.stack, detail: names.detail]
-        break
-      case Namespace.APPLICATIONS.ns:
-        result << [application: parts[2].toLowerCase()]
-        break
-      case Namespace.HEALTH.ns:
-        result << [id: parts[2], account: parts[3], region: parts[4], provider: parts[5]]
-        break
-      default:
-        return null
-        break
+    if (parts[2] == CachingSchema.V2.toString() || parts[2] == CachingSchema.V2) { // parsing for split caching
+      switch (result.type) {
+        case Namespace.IMAGES.ns:
+          result << [account: parts[3], region: parts[4], imageId: parts[5]]
+          break
+        case Namespace.SERVER_GROUPS.ns:
+          def names = Names.parseName(parts[6])
+          result << [application: names.app.toLowerCase(), cluster: parts[3], account: parts[4], region: parts[5], serverGroup: parts[6], stack: names.stack, detail: names.detail, sequence: names.sequence?.toString()]
+          break
+        case Namespace.INSTANCES.ns:
+          result << [id: parts[3], region: parts[4], instanceId: parts[5]]
+          break
+        case Namespace.CLUSTERS.ns:
+          def names = Names.parseName(parts[5])
+          result << [application: parts[3].toLowerCase(), account: parts[4], cluster: parts[5], stack: names.stack, detail: names.detail]
+          break
+        case Namespace.APPLICATIONS.ns:
+          result << [application: parts[3].toLowerCase()]
+          break
+        case Namespace.HEALTH.ns:
+          result << [id: parts[2], account: parts[2], provider: parts[3]]
+        default:
+          return null
+          break
+      }
+    } else { // original parsing
+      switch (result.type) {
+        case Namespace.IMAGES.ns:
+          result << [account: parts[2], region: parts[3], imageId: parts[4]]
+          break
+        case Namespace.SERVER_GROUPS.ns:
+          def names = Names.parseName(parts[5])
+          result << [application: names.app.toLowerCase(), cluster: parts[2], account: parts[3], region: parts[4], serverGroup: parts[5], stack: names.stack, detail: names.detail, sequence: names.sequence?.toString()]
+          break
+        case Namespace.INSTANCES.ns:
+            result << [id: parts[2], region: parts[3], instanceId: parts[5]]
+          break
+        case Namespace.CLUSTERS.ns:
+          def names = Names.parseName(parts[4])
+          result << [application: parts[2].toLowerCase(), account: parts[3], cluster: parts[4], stack: names.stack, detail: names.detail]
+          break
+        case Namespace.APPLICATIONS.ns:
+          result << [application: parts[2].toLowerCase()]
+          break
+        case Namespace.HEALTH.ns:
+          result << [id: parts[2], account: parts[3], region: parts[4], provider: parts[5]]
+          break
+        default:
+          return null
+          break
+      }
     }
 
-    result
+    return result
   }
 
+  // v2 keys for split provider
+  static String getImageV2Key(String imageId, String accountName, String region) {
+    "${TitusCloudProvider.ID}:${Namespace.IMAGES}:${CachingSchema.V2}:${accountName}:${region}:${imageId}"
+  }
+
+  static String getServerGroupV2Key(String cluster, String autoScalingGroupName, String account, String region) {
+    "${TitusCloudProvider.ID}:${Namespace.SERVER_GROUPS}:${CachingSchema.V2}:${cluster}:${account}:${region}:${autoScalingGroupName}"
+  }
+
+  static String getServerGroupV2Key(String serverGroupName, String account, String region) {
+    Names names = Names.parseName(serverGroupName)
+    return getServerGroupV2Key(names.cluster, names.group, account, region)
+  }
+
+  static String getInstanceV2Key(String instanceId, String account, String region) {
+    "${TitusCloudProvider.ID}:${Namespace.INSTANCES}:${CachingSchema.V2}:${account}:${region}:${instanceId}"
+  }
+
+  static String getClusterV2Key(String clusterName, String application, String account) {
+    "${TitusCloudProvider.ID}:${Namespace.CLUSTERS}:${CachingSchema.V2}:${application?.toLowerCase()}:${account}:${clusterName}"
+  }
+
+  // keys for normal provider
   static String getImageKey(String imageId, String account, String region) {
     "${TitusCloudProvider.ID}:${Namespace.IMAGES}:${account}:${region}:${imageId}"
   }
@@ -117,6 +177,7 @@ class Keys implements KeyParser {
     return getServerGroupKey(names.cluster, names.group, account, region)
   }
 
+  // stack here is Titus Stack, and it's not needed anymore because Titus ids are unique across all stacks
   static String getServerGroupKey(String cluster, String autoScalingGroupName, String account, String region) {
     "${TitusCloudProvider.ID}:${Namespace.SERVER_GROUPS}:${cluster}:${account}:${region}:${autoScalingGroupName}"
   }
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/Keys.groovy
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/Keys.groovy
@@ -80,7 +80,7 @@ class Keys implements KeyParser {
     def result
     result = [provider: parts[0], type: parts[1]]
 
-    if (parts[2] == CachingSchema.V2.toString() || parts[2] == CachingSchema.V2) { // parsing for split caching
+    if (parts[2] == CachingSchema.V2.toString()) { // parsing for split caching
       switch (result.type) {
         case Namespace.IMAGES.ns:
           result << [account: parts[3], region: parts[4], imageId: parts[5]]
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/Keys.groovy
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/Keys.groovy
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.clouddriver.titus.caching
 
 import com.netflix.frigga.Names
 import com.netflix.spinnaker.clouddriver.cache.KeyParser
+import com.netflix.spinnaker.clouddriver.titus.caching.utils.CachingSchema
 import org.springframework.stereotype.Component
 import com.netflix.spinnaker.clouddriver.titus.TitusCloudProvider
 
@@ -63,16 +64,6 @@ class Keys implements KeyParser {
   @Override
   Boolean canParseField(String field) {
     return false
-  }
-
-
-  static enum CachingSchema {
-    V1,
-    V2
-
-    String toLowerCase() {
-      return name().toLowerCase()
-    }
   }
 
   static Map<String, String> parse(String key) {
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/TitusCachingProviderConfig.groovy
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/TitusCachingProviderConfig.groovy
@@ -75,15 +75,12 @@ class TitusCachingProviderConfig {
           )
         } else { //use new split caching for this whole account
           agents << new TitusInstanceCachingAgent(
-            titusCloudProvider,
             titusClientProvider,
             account,
             region,
             objectMapper,
             registry,
-            awsLookupUtilProvider,
-            pollIntervalMillis,
-            timeOutMilis
+            awsLookupUtilProvider
           )
           agents << new TitusV2ClusterCachingAgent(
             titusCloudProvider,
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/TitusCachingProviderConfig.groovy
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/TitusCachingProviderConfig.groovy
@@ -25,6 +25,8 @@ import com.netflix.spinnaker.clouddriver.security.AccountCredentialsRepository
 import com.netflix.spinnaker.clouddriver.titus.TitusClientProvider
 import com.netflix.spinnaker.clouddriver.titus.TitusCloudProvider
 import com.netflix.spinnaker.clouddriver.titus.caching.agents.TitusClusterCachingAgent
+import com.netflix.spinnaker.clouddriver.titus.caching.agents.TitusInstanceCachingAgent
+import com.netflix.spinnaker.clouddriver.titus.caching.agents.TitusV2ClusterCachingAgent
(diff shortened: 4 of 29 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsSearchProvider.groovy
```diff
@@ -114,28 +113,32 @@ class CatsSearchProvider implements SearchProvider, Runnable {
...
+      if (instanceIdentifiers) {
+        cachedIdentifiersByType.set(["instances": instanceIdentifiers])
       }
-    }.flatten()
 
-    if (instanceIdentifiers) {
-      cachedIdentifiersByType.set(["instances": instanceIdentifiers])
+      log.info("Refreshed Cached Identifiers (found ${instanceIdentifiers.size()} instances)")
+    } catch (Exception e) {
+      log.error("Unable to refresh cached identifiers", e)
```
robzienert: You may want to include `(instances)` in the log message as you did at the beginning of the try/catch.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/provider/agent/Namespace.groovy
```diff
@@ -38,7 +38,7 @@ enum Namespace {
   RESERVATION_REPORTS,
   RESERVED_INSTANCES
 
-  final String ns
+  public final String ns
```
robzienert: In Groovy, members are public-by-default, so this isn't necessary.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/provider/agent/Namespace.groovy
```diff
@@ -49,4 +49,8 @@ enum Namespace {
   String toString() {
     ns
   }
+
+  String getNs() {
+    return ns
+  }
```
robzienert: ...And so this method is also unnecessary.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/TitusCachingProviderConfig.groovy
```diff
@@ -57,20 +59,45 @@ class TitusCachingProviderConfig {
...
-          titusClientProvider,
-          account,
-          region,
-          objectMapper,
-          registry,
-          awsLookupUtilProvider,
-          pollIntervalMillis,
-          timeOutMilis
-        )
+        if (account.splitCachingEnabled == null || !account.splitCachingEnabled) { //default case
```
robzienert: Because Groovy Truth, we can just do `if (!account.splitCachingEnabled)`.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusInstanceCachingAgent.java
```diff
@@ -0,0 +1,280 @@
...
+  private TitusLoadBalancerClient titusLoadBalancerClient;
+  private NetflixTitusCredentials account;
+  private TitusRegion region;
+  private ObjectMapper objectMapper;
+  private Provider<AwsLookupUtil> awsLookupUtil;
+  private long pollIntervalMillis;
+  private long timeoutMillis;
+  private Id metricId;
+  private Registry registry;
+
```
robzienert: Extra space.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusInstanceCachingAgent.java
```diff
@@ -0,0 +1,280 @@
...
+/**
+ * Caching agent for Titus tasks, mapping to the concept of Spinnaker Instances
+ * A Titus job has a set of Titus tasks.
+ */
+public class TitusInstanceCachingAgent implements CachingAgent {
+
+  private static final Logger log = LoggerFactory.getLogger(TitusInstanceCachingAgent.class);
+  private static final TypeReference<Map<String, Object>> ATTRIBUTES = new TypeReference<Map<String, Object>>() {};
+
+
```
robzienert: Extra space.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusInstanceCachingAgent.java
```diff
@@ -0,0 +1,280 @@
...
+    return false;
+  }
+
+  static class MutableCacheData implements CacheData {
+
+    final String id;
+    int ttlSeconds = -1;
+    final Map<String, Object> attributes = new HashMap<>();
+    final Map<String, Collection<String>> relationships = new HashMap<>();
+    public MutableCacheData(String id) {
```
robzienert: Add space.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusInstanceCachingAgent.java
```diff
@@ -0,0 +1,280 @@
...
+      this.id = id;
+    }
+    @Override
```
robzienert: Add space.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusInstanceCachingAgent.java
```diff
@@ -0,0 +1,280 @@
...
+  private Map<String, CacheData> createCache() {
+    return new HashMap<>();
+  }
+
+  @Override
+  public CacheResult loadData(ProviderCache providerCache) {
+    log.info("Describing items in {}", getAgentType());
+    Long startTime = System.currentTimeMillis();
+
+
```
robzienert: Extra space.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusInstanceCachingAgent.java
```diff
@@ -0,0 +1,280 @@
...
+      CacheData serverGroupCache = serverGroups.getOrDefault(data.serverGroup, new MutableCacheData(data.serverGroup));
+      serverGroupCache.getRelationships().computeIfAbsent(INSTANCES.ns, key -> new HashSet<>()).add(data.instanceId);
+      serverGroups.put(data.serverGroup, serverGroupCache);
+    }
+  }
+
+  private Map<String, String> getTitusHealth(Task task) {
+    TaskState taskState = task.getState();
+    HealthState healthState = HealthState.Unknown;
+    if (Arrays.asList(TaskState.STOPPED, TaskState.FAILED, TaskState.CRASHED, TaskState.FINISHED, TaskState.DEAD, TaskState.TERMINATING).contains(taskState)) {
```
robzienert: I'd move this into a `static final` on the class, rather than creating the array every time we invoke the method. Same goes for the one below.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
```diff
@@ -0,0 +1,636 @@
...
+  private ObjectMapper objectMapper;
+  private OnDemandMetricsSupport metricsSupport;
+  private Provider<AwsLookupUtil> awsLookupUtil;
+  private long pollIntervalMillis;
+  private long timeoutMillis;
+  private Registry registry;
+  private Id metricId;
+
+  public TitusV2ClusterCachingAgent(TitusCloudProvider titusCloudProvider,
+                           TitusClientProvider titusClientProvider,
```
robzienert: Some alignment here.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
```diff
@@ -0,0 +1,636 @@
...
+
+  @Override
+  public OnDemandResult handle(ProviderCache providerCache, Map<String, ?> data) {
+    Long startTime = System.currentTimeMillis();
+
+    if (!data.containsKey("serverGroupName") || !data.containsKey("account") || !data.containsKey("region")) {
+      return null;
+    }
+
+    if (account.getName() != data.get("account")) {
```
robzienert: Shouldn't this be `!account.getName().equals(data.get("account"))`?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
```diff
@@ -0,0 +1,636 @@
...
+    Map<String, Collection<String>> evictions = job == null
+      ? Collections.emptyMap()
+      : Collections.singletonMap(SERVER_GROUPS.ns, Collections.singletonList(serverGroupKey));
+
+    log.info("minimal onDemand cache refresh (data: {}, evictions: {})", data, evictions);
+    return new OnDemandResult(getOnDemandAgentType(), new DefaultCacheResult(cacheResults), evictions);
+  }
+
+
+
```
robzienert: Spaces.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
```diff
@@ -0,0 +1,636 @@
...
+      return attributes;
+    }
+
+    @Override
+    public Map<String, Collection<String>> getRelationships() {
+      return relationships;
+    }
+
+    @JsonCreator
+    public MutableCacheData(@JsonProperty("id") String id,
```
robzienert: Let's generally keep class formatting uniform; constructors at the top after properties.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
```diff
@@ -0,0 +1,636 @@
...
+    protected Map<String, Object> toMap() {
+      Map<String, String> status = new HashMap<>();
+      status.put("state", this.status.getState().name());
+      status.put("reason", this.status.getPendingReason());
+
+      Map<String, Object> result = new HashMap<>();
+      result.put("id", id);
+      result.put("status", status);
+
+      String scalingPolicy;
```
robzienert: Move this into the `try`, we don't do anything with it outside of that scope.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
```diff
@@ -250,22 +282,33 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
       serverGroup.awsAccount = awsLookupUtil.lookupAccount(serverGroupEntry.attributes.account, serverGroupEntry.attributes.region)?.awsAccount
       [(serverGroupEntry.id): serverGroup]
     }
-    serverGroups
+    return serverGroups
```
robzienert: +1 -- I don't like implicit returns.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusInstanceProvider.groovy
```diff
@@ -63,25 +68,50 @@ class TitusInstanceProvider implements InstanceProvider<TitusInstance> {
...
+        job = loadJob(instanceEntry)
+      } else {
+        log.error( "Task {} in {}:{} does not have a job", task.id, account, region)
+      }
+    } else {
+      // Instance is cached at the same time as job, V1 schema
+      objectMapper.convertValue(instanceEntry.attributes.job, Job)
+    }
+
     TitusInstance instance = new TitusInstance(job, task)
```
robzienert: Looks like we can get an NPE here with the new codepath. `@L91`, we do `Job job`, but only in a single branch in the following block of code do we assign it. Is this expected? I presume not, as at L101, you do a `objectMapper.convertValue` but do not assign its return value to anything.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/provider/agent/Namespace.groovy
emjburns: But when you access it from java, it needs the specific declaration of `public` :( w/o explicitly declaring it public it won't compile

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/provider/agent/Namespace.groovy
emjburns: Also java + groovy fun, the ecs provider breaks w/o this explicitly declared. Won't compile without this method :(

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusInstanceProvider.groovy
emjburns: Oops good catch. L101 should assign it. Fixing.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/Keys.groovy
```diff
@@ -77,37 +86,88 @@ class Keys implements KeyParser {
...
+      switch (result.type) {
+        case Namespace.IMAGES.ns:
+          result << [account: parts[3], region: parts[4], imageId: parts[5]]
+          break
+        case Namespace.SERVER_GROUPS.ns:
+          def names = Names.parseName(parts[6])
+          result << [application: names.app.toLowerCase(), cluster: parts[3], account: parts[4], region: parts[5], serverGroup: parts[6], stack: names.stack, detail: names.detail, sequence: names.sequence?.toString()]
+          break
+        case Namespace.INSTANCES.ns:
+          result << [id: parts[3], region: parts[4], instanceId: parts[5]]
```
emjburns: This fixed titus instance search (needed to return `instanceId`

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/Keys.groovy
```diff
@@ -77,37 +86,88 @@ class Keys implements KeyParser {
...
+      }
     }
 
-    result
+    return result
+  }
+
+  // v2 keys for split provider
+  static String getImageV2Key(String imageId, String accountName, String region) {
+    "${TitusCloudProvider.ID}:${Namespace.IMAGES}:${CachingSchema.V2}:${accountName}:${region}:${imageId}"
```
emjburns: Adding a "V2" into the key to that we can parse correctly. Needs to be in the 3rd spot because search assumes first two keys are `cloudProvider:namespace`

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/TitusCachingProviderConfig.groovy
```diff
@@ -57,20 +59,45 @@ class TitusCachingProviderConfig {
...
+            titusClientProvider,
+            account,
+            region,
+            objectMapper,
+            registry,
+            awsLookupUtilProvider,
+            pollIntervalMillis,
+            timeOutMilis
+          )
+        } else { //use new split caching for this whole account
```
emjburns: I've chosen to split the caching by account, not region, because in the controllers account is passed in and not region. I didn't want to support two key types w/in the same account for the transition because it would be a lot of extra code for little benefit.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusInstanceCachingAgent.java
```diff
@@ -0,0 +1,282 @@
...
+  private Map<String, CacheData> createCache() {
+    return new HashMap<>();
+  }
+
+  @Override
+  public CacheResult loadData(ProviderCache providerCache) {
+    log.info("Describing items in {}", getAgentType());
+    Long startTime = System.currentTimeMillis();
+
+    List<Task> tasks = titusClient.getAllTasks();
```
emjburns: This is the meaty fetch data call. Locally in us-east-1 this takes ~3-6s

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
```diff
@@ -0,0 +1,634 @@
...
+  @Override
+  public CacheResult loadData(ProviderCache providerCache) {
+    Long startTime = System.currentTimeMillis();
+
+    log.info("Describing items in {}", getAgentType());
+    List<CacheData> evictFromOnDemand = new ArrayList<>();
+    List<CacheData> keepInOnDemand = new ArrayList<>();
+
+
+    List<Job> jobs = titusClient.getAllJobsWithoutTasks();
```
emjburns: This is the meaty call to Titus. Running locally against us-east-1 this takes ~15s.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
```diff
@@ -0,0 +1,634 @@
...
+      .record(System.currentTimeMillis() - startScalingPolicyTime, MILLISECONDS);
+
+    Long startLoadBalancerTime = System.currentTimeMillis();
+    Map<String, List<String>> allLoadBalancers = titusLoadBalancerClient.getAllLoadBalancers();
+    PercentileTimer
+      .get(registry, metricId.withTag("operation", "getLoadBalancers"))
+      .record(System.currentTimeMillis() - startLoadBalancerTime, MILLISECONDS);
+
+    Long startJobIdsTime = System.currentTimeMillis();
+    Map<String, List<String>> taskAndJobIds = titusClient.getTaskIdsForJobIds();
```
emjburns: Since Titus tasks don't know their job ids, we need to grab them separately from Titus to build relationships. This call takes < 1s running locally against Titus us-east-1

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
```diff
@@ -279,9 +322,13 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
     healths.each { healthEntry ->
       def instanceId = healthKeysToInstance.get(healthEntry.id)
       healthEntry.attributes.remove('lastUpdatedTimestamp')
-      instances[instanceId].health << healthEntry.attributes
+      def health = healthEntry.attributes.collectEntries { key, value ->
```
emjburns: I'm not sure if changing this value to an explicit string here (instead of a boolean) will affect the UI. I am able to run tasks locally with Orca + Clouddriver (create/clone/distroy).

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusInstanceProvider.groovy
```diff
@@ -92,14 +122,33 @@ class TitusInstanceProvider implements InstanceProvider<TitusInstance> {
       healthKeys.unique().each { key ->
         def externalHealth = cacheView.getAll(HEALTH.ns, key)
         if (externalHealth) {
-          def health = externalHealth*.attributes
-          health.each { it.remove('lastUpdatedTimestamp') }
+          List<Map<String, Object>> health = externalHealth*.attributes
+          health.each {
+            it.remove('lastUpdatedTimestamp')
+            // groovy + java means that everything must be explicitly a string
+            it.allowMultipleEurekaPerAccount = it.allowMultipleEurekaPerAccount.toString()
```
emjburns: I'm not sure if changing this value to an explicit string here (instead of a boolean) will affect the UI. I am able to run tasks locally with Orca + Clouddriver (create/clone/distroy).

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/utils/AwsLookupUtil.groovy
```diff
@@ -27,23 +26,25 @@ import com.netflix.spinnaker.clouddriver.aws.provider.view.AmazonVpcProvider
...
 import com.netflix.spinnaker.clouddriver.titus.model.TitusInstance
-import com.netflix.spinnaker.clouddriver.titus.model.TitusSecurityGroup
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 
 import javax.annotation.PostConstruct
 
-import static com.netflix.spinnaker.clouddriver.aws.cache.Keys.Namespace.SECURITY_GROUPS
 import static com.netflix.spinnaker.clouddriver.core.provider.agent.Namespace.HEALTH
+import static com.netflix.spinnaker.clouddriver.titus.caching.Keys.CachingSchema.V1
```
tomaslin: Would prefer for the caching schema logic to be in its own utility class since this is going to be very likely very ephemeral. A titus region feature flag feels more appropiate

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/Job.java
```diff
@@ -23,192 +23,6 @@
 
 public class Job {
 
-  public static class TaskSummary {
```
tomaslin: nice

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
```diff
@@ -220,19 +238,33 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
...
+        objectMapper.convertValue(cacheData.attributes.task, Task)
+      }]
+    }
+
+    Map<String, TitusInstance> instances = translateInstances(allInstances, serverGroupData)
 
     Map<String, TitusServerGroup> serverGroups = serverGroupData.collectEntries { serverGroupEntry ->
       String json = objectMapper.writeValueAsString(serverGroupEntry.attributes.job)
       Job job = objectMapper.readValue(json, Job)
+      if (job.tasks == null || job.tasks.isEmpty()) {
```
tomaslin: not sure why this is needed

tomaslin: we seem to be resolving instances from relationships later on, if this is not needed it might save us a call to get all task ids when caching the job

emjburns: You're right, this is not needed. We still need to call task ids though to correctly populate the relationships I think

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/utils/AwsLookupUtil.groovy
emjburns: Moving caching schema logic to separate class. I chose to flag at account level and not region because of on how we query for things: account is often specified, whereas region is not. Seemed less error-prone to add an account level flag rather than adding a check to make sure that if the split flag is enabled in one region it's enabled everywhere. 

I'll remove the flag/config for it when we switch over completely.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/TitusCachingProviderConfig.groovy
cfieber: seems reasonable to me

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusInstanceCachingAgent.java
```diff
@@ -0,0 +1,282 @@
...
+
+  private static final java.util.Set<AgentDataType> types = Collections.unmodifiableSet(Stream.of(
+    AUTHORITATIVE.forType(INSTANCES.ns),
+    INFORMATIVE.forType(SERVER_GROUPS.ns)
+  ).collect(Collectors.toSet()));
+
+  private static final List<TaskState> DOWN_TASK_STATES = Arrays.asList(TaskState.STOPPED, TaskState.FAILED, TaskState.CRASHED, TaskState.FINISHED, TaskState.DEAD, TaskState.TERMINATING);
+  private static final List<TaskState> STARTING_TASK_STATES = Arrays.asList(TaskState.STARTING, TaskState.DISPATCHED, TaskState.PENDING, TaskState.QUEUED);
+
+  private TitusCloudProvider titusCloudProvider;
```
cfieber: this looks unused

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
```diff
@@ -0,0 +1,628 @@
...
+
+  static final java.util.Set<AgentDataType> types = Collections.unmodifiableSet(Stream.of(
+      AUTHORITATIVE.forType(SERVER_GROUPS.ns),
+      AUTHORITATIVE.forType(APPLICATIONS.ns),
+      INFORMATIVE.forType(IMAGES.ns),
+      INFORMATIVE.forType(CLUSTERS.ns),
+      INFORMATIVE.forType(TARGET_GROUPS.ns)
+    ).collect(Collectors.toSet()));
+
+  private TitusCloudProvider titusCloudProvider;
```
cfieber: should be able to delete/not inject this

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
```diff
@@ -0,0 +1,628 @@
...
+                                    Long timeoutMillis) {
+    this.account = account;
+    this.region = region;
+
+    this.titusCloudProvider = titusCloudProvider;
+    this.objectMapper = objectMapper;
+    this.metricsSupport = new OnDemandMetricsSupport(
+      registry,
+      this,
+      titusCloudProvider.getId() + ":" + OnDemandType.ServerGroup
```
cfieber: TitusCloudProvider.ID should be available as a constant here

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
```diff
@@ -0,0 +1,628 @@
...
+    return onDemandResult;
+  }
+
+  /**
+   * Avoid writing cache results to both ON_DEMAND and SERVER_GROUPS, etc.
+   *
+   * By writing a minimal record to ON_DEMAND only, we eliminate significant overhead (redis and network) at the cost
+   * of an increase in time before a change becomes visible in the UI.
+   *
+   * A change will not be visible until a caching cycle has completed.
```
cfieber: if we aren't going to see a change until a caching cycle completes, and we aren't reading through any intermediate caches, is there value in onDemand here?  cc: @ajordens think you touched it last

cfieber: if we are down to 15s for a caching agent run with the new split changes, maybe we don't need onDemand complexity?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
```diff
@@ -0,0 +1,628 @@
...
+      .map( job -> {
+        List<ScalingPolicyData> jobScalingPolicies = scalingPolicyResults.stream()
+          .filter( it -> it.getJobId().equalsIgnoreCase(job.getId()) && cacheablePolicyStates.contains(it.getPolicyState()))
+          .map( it -> new ScalingPolicyData(it.getId().getId(), it.getScalingPolicy(), it.getPolicyState()))
+          .collect(Collectors.toList());
+
+        List<String> jobLoadBalancers = allLoadBalancers.getOrDefault(job.getId(), Collections.emptyList());
+        return new ServerGroupData(job, jobScalingPolicies, jobLoadBalancers, taskAndJobIds.get(job.getId()), account.getName(), region.getName());
+      })
+      .collect(Collectors.toList());
```
cfieber: think you could drop the collect and just chain the following forEach off this stream

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
emjburns: I'll defer to @ajordens - this is just a cary over of his onDemand implementation.

ajordens: we still need it IMO

PR #2637, opened 2018-05-17
fix(provider/amazon): improve instancetype caching

cfieber:
Switches AWS Instance Type caching to process the amazon billing api
offerings file instead of making an API call to reserved instance
offerings.

Polls this endpoint on the regular caching interval, but only
processes it if it serves up a different etag value than the
previous cycle.

The billing endpoint is listed as us-east-1 but is actually fronted
by CloudFront so in theory shouldn't be an issue to hit from any
specific Spinnaker installation.


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
```diff
@@ -0,0 +1,196 @@
...
+
+  private static final TypeReference<Map<String, Object>> ATTRIBUTES = new TypeReference<Map<String, Object>>() {
+  };
+
+  // https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonEC2/current/us-west-2/index.json
+  private final String region;
+  private final AccountCredentialsRepository accountCredentialsRepository;
+  private final URI pricingUri;
+  private final HttpHost pricingHost;
+  private final HttpClient httpClient;
```
cfieber: This is all vanilla HttpClient because the cloudfront endpoint doesn't do a good job with content types and trying to retrofit it up would have just made all the methods have to return Response anyway

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
```diff
@@ -0,0 +1,196 @@
...
+    this.region = region;
+    this.accountCredentialsRepository = accountCredentialsRepository;
+    pricingHost = HttpHost.create("https://pricing.us-east-1.amazonaws.com");
+    pricingUri = URI.create("/offers/v1.0/aws/AmazonEC2/current/" + region + "/index.json");
+    this.httpClient = httpClient;
+  }
+
+  @Override
+  public Collection<AgentDataType> getProvidedDataTypes() {
+    return Collections.unmodifiableList(Arrays.asList(new AgentDataType(Keys.Namespace.INSTANCE_TYPES.getNs(), AgentDataType.Authority.AUTHORITATIVE), new AgentDataType(getAgentType(), AgentDataType.Authority.AUTHORITATIVE)));
```
robzienert: Do want 120 character wrapping.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
```diff
@@ -0,0 +1,196 @@
...
+        .map(AmazonCredentials.class::cast)
+        .filter(ac -> ac.getRegions().stream().anyMatch(r -> region.equals(r.getName())))
+        .map(AccountCredentials::getName)
+        .collect(Collectors.toSet());
+
+      if (matchingAccounts.isEmpty()) {
+        return new DefaultCacheResult(Collections.emptyMap());
+      }
+
+      CacheData metadata = providerCache.get(getAgentType(), "metadata", RelationshipCacheFilter.none());
```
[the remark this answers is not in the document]
robzienert: That is neat! It may be worth just adding a quick comment in the code to reflect this so future me doesn't forget this is what we're doing here.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/config/AwsInfrastructureProviderConfig.groovy
```diff
@@ -88,13 +88,17 @@ class AwsInfrastructureProviderConfig {
     def scheduledAccounts = ProviderUtils.getScheduledAccounts(awsInfrastructureProvider)
     def allAccounts = ProviderUtils.buildThreadSafeSetOfAccounts(accountCredentialsRepository, NetflixAmazonCredentials)
 
+    Set<String> regions = new HashSet<>();
     allAccounts.each { NetflixAmazonCredentials credentials ->
       for (AmazonCredentials.AWSRegion region : credentials.regions) {
         if (!scheduledAccounts.contains(credentials.name)) {
           def newlyAddedAgents = []
 
+          if (regions.add(region.name)) {
```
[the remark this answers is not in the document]
robzienert: Do like!

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
dreynaud: Could also use an okHttpClient for consistency's sake

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
```diff
@@ -0,0 +1,196 @@
...
+    public Map<String, Offering> products;
+  }
+
+  static class MetadataAttributes {
+    public String etag;
+    public Set<String> cachedInstanceTypes;
+  }
+
+
+  //visible for testing
```
ajordens: curious re: this vs. the ol `@VisibleForTesting` annotation?

clouddriver-aws/src/test/resources/com/netflix/spinnaker/clouddriver/aws/provider/agent/us-west-2.json
```diff
@@ -0,0 +1,3397 @@
+{
```
ajordens: this is how @cfieber wins the # of lines owned in clouddriver

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
```diff
@@ -0,0 +1,196 @@
...
+      MetadataAttributes metadataAttributes = null;
+      if (metadata != null) {
+        metadataAttributes = objectMapper.convertValue(metadata.getAttributes(), MetadataAttributes.class);
+      }
+
+      Set<String> instanceTypes = null;
+      if (metadataAttributes != null && metadataAttributes.etag != null && metadataAttributes.cachedInstanceTypes != null) {
+
+        //we have enough from a previous request to not re-request if the etag is unchanged..
+        HttpResponse headResponse = httpClient.execute(pricingHost, new HttpHead(pricingUri));
```
ajordens: out of curiosity, what are the default connect timeouts for stock `HttpClient` ... anything to be concerned with there?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
```diff
@@ -0,0 +1,196 @@
...
+          instanceTypeAttributes.put("account", account);
+          instanceTypeAttributes.put("region", region);
+          instanceTypeAttributes.put("name", instanceType);
+          instanceTypeData.add(new DefaultCacheData(Keys.getInstanceTypeKey(instanceType, region, account), instanceTypeAttributes, Collections.emptyMap()));
+        }
+      }
+
+      return new DefaultCacheResult(cacheResults, evictions);
+    } catch (Exception ex) {
+      throw new RuntimeException(ex);
```
dreynaud: do we really want to make every exception sneaky here?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
```diff
@@ -0,0 +1,196 @@
...
+        HttpResponse headResponse = httpClient.execute(pricingHost, new HttpHead(pricingUri));
+        EntityUtils.consumeQuietly(headResponse.getEntity());
+        if (headResponse.getStatusLine().getStatusCode() != 200) {
+          throw new Exception("failed to read instance type metadata for " + region + ": " + headResponse.getStatusLine().toString());
+        }
+
+        Optional<String> etag = getEtagHeader(headResponse);
+
+        if (etag.filter(metadataAttributes.etag::equals).isPresent()) {
+          instanceTypes = metadataAttributes.cachedInstanceTypes;
```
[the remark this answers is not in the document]
cfieber: I didn't want the TTL'd keys to just expire out from under this, so it's rewriting the data set each run. That said, its actually a pretty small set once it is down to just the instance types

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
```diff
@@ -0,0 +1,196 @@
...
+          metadataAttributes.etag = etag.get();
+          metadataAttributes.cachedInstanceTypes = new HashSet<>(instanceTypes);
+          metadata = new DefaultCacheData("metadata", objectMapper.convertValue(metadataAttributes, ATTRIBUTES), Collections.emptyMap());
+        } else {
+          metadata = null;
+        }
+      }
+      Map<String, Collection<String>> evictions = new HashMap<>();
+      Map<String, Collection<CacheData>> cacheResults = new HashMap<>();
+      List<CacheData> instanceTypeData = new LinkedList<>();
```
dreynaud: any particular reason to use a LinkedList over an ArrayList here? I could see an advantage if we were to remove random elements, but it seems we only ever `add` to the list

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
cfieber: not really a fan of importing a guava annotation when nothing else is guava'd and we don't really get much of anything tooling wise from that annotation

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
cfieber: it's effectively what most other caching agents to, just leveraging loosy groovy to do so.

The scheduler will log and reschedule the agent if it fails 

If there was a specifically handleable exception in here I'd be all for handling it (if find we need retries on the http call or anything)

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
```diff
@@ -0,0 +1,196 @@
...
+      cacheResults.put(Keys.Namespace.INSTANCE_TYPES.getNs(), instanceTypeData);
+      if (metadata != null) {
+        cacheResults.put(getAgentType(), Collections.singleton(metadata));
+      } else {
+        evictions.put(getAgentType(), Collections.singleton("metadata"));
+      }
+
+      for (String instanceType : instanceTypes) {
+        for (String account : matchingAccounts) {
+          Map<String, Object> instanceTypeAttributes = new HashMap<>();
```
dreynaud: If we have Guava around, I would do 

```
  Map<String, Object> instanceTypeAttributes = ImmutableMap.of(
    "account", account,
    "region", region,
    "name", instanceType);
```

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
cfieber: honestly I figured that it is just an append-only list of somewhat unknown size that is then only iterated over afterwords not random-accessed so linked list would make sense

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
cfieber: I'm always leery of the guava and it's fragility so I tend to avoid it unless I'm going to use a whole lot of it

PR #2643, opened 2018-05-21
fix(titus): on demand eviction

emjburns:
If `!job`, then this will throw an NPE, and we will never evict. We should instead use `serverGroupName` to construct the key so we can properly evict.


tomaslin:
LGTM

ajordens:
LGTM

PR #2645, opened 2018-05-22
chore(cache): debug logging for cache controller

danielpeach:


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy
```diff
@@ -61,7 +61,6 @@ class CatsOnDemandCacheUpdater implements OnDemandCacheUpdater {
   }
 
   OnDemandCacheResult handle(OnDemandAgent.OnDemandType type, Collection<OnDemandAgent> onDemandAgents, Map<String, ? extends Object> data) {
-    log.debug("Calling handle on data: {}, onDemandAgents: {}, type: {}", data, onDemandAgents, type)
```
lwander: Any reason to drop this already? I don't think it hurts to keep around a little longer as we get to the bottom of this...

danielpeach: Sure, added back.

PR #2651, opened 2018-05-23
fix(docker): Suppress unnecessary stacktraces when `getTags` returns 404

ajordens:


ajordens:
Just doing some cleaning up of what I'd consider unnecessary stacktraces ... sensible @tomaslin?

tomaslin:
makes sense, wonder if it makes sense to also just log out the message in line 102 since that is super verbose as well

tomaslin:
LGTM otherwise

ajordens:
At this point we're only seeing 404's, so I'm inclined to do a targeted fix. 

If (in the future) we see something else pop up in logs we can circle around.

PR #2660, opened 2018-05-24
fix(provider/kubernetes): default behavior shouldn't version volumes

lwander:
These are typically provisioned with a separate lifecycle, and are often used to persist data beyond the length of a pod's lifespan. They can still be versioned using the `strategy.spinnaker.io/versioned`
annotation, but it arguably is confusing as a default behavior

@imosquera @wjoel PTAL


wjoel:
TIL. That's confusing, yes. I agree that it should not be the default.

PR #2131, opened 2017-11-16
fix(provider/kubernetes): fix onDemand cache update in controller agent.

chlung:
About this PR,

Fix the statefulset and daemonset server group never invoke onDemand cache update.

Thanks.

We prefer small, well tested pull requests.

Please refer to [Contributing to Spinnaker](https://spinnaker.io/community/contributing/).

When filling out a pull request, please consider the following:

* Follow the commit message conventions [found here](http://www.spinnaker.io/v1.0/docs/how-to-submit-a-patch).
* Provide a descriptive summary for your changes.
* If it fixes a bug or resolves a feature request, be sure to link to that issue.
* Add inline code comments to changes that might not be obvious.
* Squash your commits as you keep adding changes.
* Add a comment to @spinnaker/reviewers for review if your issue has been outstanding for more than 3 days.

Note that we are unlikely to accept pull requests that add features without prior discussion. The best way to propose a feature is to open an issue first and discuss your ideas there before implementing them.

PR #2663, opened 2018-05-25
fix(caching): Fix on-demand cache result return status

ezimanyi:
When multiple caching agents handle an on-demand refresh, the return
status is currently based only on the result of the last agent.  Instead
we should return a PENDING result if *any* of the caching agents find
results, rather than only if that last one does.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy
@@ -84,12 +84,12 @@ class CatsOnDemandCacheUpdater implements OnDemandCacheUpdater {
             continue;
           }
           if (!agent.metricsSupport) {
-            hasOnDemandResults = false
             continue;
           }
           if (result.cacheResult) {
-            hasOnDemandResults = !(result.cacheResult.cacheResults ?: [:]).values().flatten().isEmpty() && !agentScheduler.atomic
-            if (hasOnDemandResults) {
+            boolean agentHasOnDemandResults = !(result.cacheResult.cacheResults ?: [:]).values().flatten().isEmpty() && !agentScheduler.atomic
+            if (agentHasOnDemandResults) {
+              hasOnDemandResults = true;
               result.cacheResult.cacheResults.each { k, v ->
                 if (v) {
                   cachedIdentifiersByType[k].addAll(v*.id)


ezimanyi:
Also removed the explicit `hasOnDemandResults = false` in the event that the agent doesn't have metrics support.  The behavior now is:
* `hasOnDemandResults` starts out `false`
* it becomes `true` if *any* agent either finds results or could not acquire a lock; we never explicitly set it to `false` again

Prior to this PR, it was possible for an earlier agent to set `hasOnDemandResults` to `true`, but have a later agent set it to `false` making us return `OnDemandCacheStatus.SUCCESSFUL` even though we should return `OnDemandCacheStatus.PENDING` because a cache found results.

ajordens:
Seems alright.

Can you offer an explanation as to why an `OnDemandAgent` would _not_ have `metricsSupport` ... that seems suspect to me and not what I would expect.

What is the scenario where you have multiple agents responding to a single cache refresh request? We've avoided that in AWS and Titus-land.

ezimanyi:
@ajordens : I'm not sure why an `OnDemandAgent` wouldn't have `metricsSupport`...the block to handle that was added in https://github.com/spinnaker/clouddriver/pull/2131 so maybe @lwander remembers why that change was made.  I removed the `hasOnDemandResults = false` because, in any case, it seemed incorrect that such an agent would overwrite the results of a prior valid agent.

The case of multiple agents responding is for GCE server groups...these server groups can be either zonal or regional, so both the zonal and regional caching agents will respond to the request.  Of course, only one will find results depending if it's zonal/regional, and currently the return status is depending on the order these two handlers are called.

Thanks for the advice on avoiding multiple agents responding to a request. I'll look to see if I can make those handlers more specific, and have the zonal/regional handlers decide if the request is for a zonal/regional server group before responding.  (I *think* it should always be possible based on the incoming data to know this, but will have to investigate first.)  But separate from that, it still seems like this will make the `CatsOnDemandCacheUpdater` more robust to multiple agents handling a request (even if that's maybe not the recommended situation).

ajordens:
Agree that it seems incorrect, 👍 on removing it.

Feel free to merge this at your convenience.

kcbaltz:
We're seeing an issue with the Force Cache Refresh step taking 1:50 - 2:00 minutes instead of 0-1 seconds and we think we've narrowed down the release that introduced this behavior.  I can't see anything in this code that would take that long to run, so I'm wondering if the change in return value might be causing orca to retry more often or something like that.  Maybe it was completing faster before even though all the agents hadn't been refreshed.  Does this seem right?

ezimanyi:
@kcbaltz : It does seem possible that prior to this change Orca was immediately returning success rather than waiting for the cache to actually refresh (which was the bug I was fixing).  In general for anything using asynchronous caching, it should take at least one caching cycle for the result to be cached so it's not surprising that it would take 1-2 minutes.

Which particular stage's cache refresh are you seeing this change in behavior on?  Also, this change has not made it into a released version of Spinnaker yet---are you running pre-release versions where you are seeing this behavior change?

kcbaltz:
@ezimanyi We're seeing the delay in both the Deploy and Disable stages at least.  
We're on the latest stable release (v1.17?) but we found deploys were taking a long time so we were experimenting with clouddriver versions.   We found a nightly clouddriver that ran faster but when we tried the latest nightly it was back to slow again.  I was asked to try to figure out what code change made the difference so we can make sure the next stable release fixes this performance issue. 

@lwander You're right, I misunderstood what folks were telling about how long it was taking before.  I guess it was up to 30 seconds, but nowhere close to 2 minutes.  

Thank you both for the quick reply.  I'm definitely new to this codebase and how our admins have set up Spinnaker/Kubernetes, etc.

PR #2666, opened 2018-05-29
perf(cats): Deserialize relationships into `Set` vs `List`

ajordens:


cats/cats-core/src/main/java/com/netflix/spinnaker/cats/provider/DefaultProviderCache.java
```diff
@@ -202,7 +202,7 @@ private boolean validType(String type) {
         if (relationship == null) {
             return new HashSet<>();
         }
-        return relationship;
+        return relationship instanceof Set ? (Set<String>) relationship : new HashSet<>(relationship);
```
ajordens: this is _probably_ not necessary as the deserialization is now to a `Set` but figured it was worth being explicit.

We have _many_ implementations of `CacheData` and I did not immediately want to go down the path of changing constructors to move away from `Collection<String>` for relationships.

Reasonable?

PR #2667, opened 2018-05-29
feat(build): separate release management opinions

cfieber:
see spinnaker/spinnaker#2844 for context
see spinnaker/spinnaker-gradle-project#39 for plugin changes

moves to publishing opinions via include file
removes spring boot plugin, was only used for bootRun,
switched to application plugin run configuration.

### DIFF
--- a/cats/cats.gradle
+++ b/cats/cats.gradle
@@ -1,4 +1,4 @@
-tasks.bintrayUpload.enabled = false
+tasks.findByName('bintrayUpload')?.enabled = false
 
 subprojects {
   apply plugin: 'findbugs'
--- a/clouddriver-oracle-bmcs/clouddriver-oracle-bmcs.gradle
+++ b/clouddriver-oracle-bmcs/clouddriver-oracle-bmcs.gradle
@@ -64,10 +64,12 @@ dependencies {
 }
 
 def allSourceSets = sourceSets
-license {
-  header project.file('oracle-source-header')
-  includes(["**/*.groovy", "**/*.java", "**/*.properties"])
-  strictCheck true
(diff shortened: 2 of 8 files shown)


cats/cats.gradle
```diff
@@ -1,4 +1,4 @@
-tasks.bintrayUpload.enabled = false
+tasks.findByName('bintrayUpload')?.enabled = false
```
cfieber: There is probably a better way to accomplish this, I'm not entirely sure why we were explicitly trying not to publish this - probably because it is an intermediate nested build. Maybe worth just flattening out the directory structure...

clouddriver-oracle-bmcs/clouddriver-oracle-bmcs.gradle
```diff
@@ -64,10 +64,12 @@ dependencies {
...
 
 def allSourceSets = sourceSets
-license {
-  header project.file('oracle-source-header')
-  includes(["**/*.groovy", "**/*.java", "**/*.properties"])
-  strictCheck true
-  skipExistingHeaders false
-  sourceSets = allSourceSets
+
+def licenseExtension = project.extensions.findByName('license')
```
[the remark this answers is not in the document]
robzienert: I think it should be - we'll be wanting the plugin across all of the projects anyway.

cats/cats.gradle
robfletcher: Yeah, I hate the way Gradle assumes any intermediate directory is its own sub-project.

ttomsu:
Made obsolete by https://github.com/spinnaker/clouddriver/pull/3035

PR #2668, opened 2018-05-29
fix(caching): add config for number of concurrent scheduled agents

emjburns:
(Please view with ["ignore whitespace"](https://blog.github.com/2011-10-21-github-secrets/#whitespace)).

We just saw clouddriver try and acquire all the locks, then die. This makes the max number of caching agents to schedule at once and the interval at which we schedule configurable. It defaults to the current behavior.

Set with:
```
redis:
  agent:
    maxConcurrentAgents: 100
    agentLockAcquisitionIntervalSeconds: 5
```

### DIFF
--- a/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredAgentScheduler.java
+++ b/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredAgentScheduler.java
@@ -29,11 +29,7 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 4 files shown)


cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredAgentScheduler.java
```diff
@@ -43,240 +39,260 @@
...
+  private final RedisClientDelegate redisClientDelegate;
+  private final NodeIdentity nodeIdentity;
+  private final AgentIntervalProvider intervalProvider;
+  private final ExecutorService agentExecutionPool;
+  private final Pattern enabledAgentPattern;
+
+  private final Map<String, AgentExecutionAction> agents = new ConcurrentHashMap<>();
+  private final Map<String, NextAttempt> activeAgents = new ConcurrentHashMap<>();
+  private final NodeStatusProvider nodeStatusProvider;
+  private Integer maxConcurrentAgents;
```
cfieber: `final`

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredAgentScheduler.java
```diff
@@ -43,240 +39,260 @@
...
+    Collections.shuffle(agentsEntrySet);
+    for (Map.Entry<String, AgentExecutionAction> agent : agentsEntrySet) {
+      if (!skip.contains(agent.getKey())) {
+        final String agentType = agent.getKey();
+        AgentIntervalProvider.Interval interval = intervalProvider.getInterval(agent.getValue().getAgent());
+        if (acquireRunKey(agentType, interval.getTimeout())) {
+          acquired.put(agentType, new NextAttempt(System.currentTimeMillis(), interval.getInterval(), interval.getErrorInterval()));
         }
+      }
+      if (acquired.keySet().size() >= availableAgents) {
```
cfieber: `acquired.size()` should work

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredAgentScheduler.java
```diff
@@ -43,240 +39,260 @@
...
+                                 String enabledAgentPattern,
+                                 Integer maxConcurrentAgents,
+                                 Integer agentLockAcquisitionIntervalSeconds) {
+    this.redisClientDelegate = redisClientDelegate;
+    this.nodeIdentity = nodeIdentity;
+    this.intervalProvider = intervalProvider;
+    this.nodeStatusProvider = nodeStatusProvider;
+    this.agentExecutionPool = agentExecutionPool;
+    this.enabledAgentPattern = Pattern.compile(enabledAgentPattern);
+    this.maxConcurrentAgents = maxConcurrentAgents == null ? Integer.MAX_VALUE : maxConcurrentAgents;
```
robzienert: Do we really want to have `Integer.MAX_VALUE` as the default? If we're going to have a default, I don't think it should be a value that we've demonstrated as being pathological.

emjburns: The goal is to keep the behavior the same. `Integer.MAX_VALUE` is definitely too high ... but it does keep the behavior the same. Thoughts on changing the behavior to something more sensible? 100?

asher: ➕

robzienert: That works for me.

PR #2671, opened 2018-05-30
fix(artifacts/gcs): support generation

lwander:


lwander:
@sbwsg yup it's a required parameter, we validate that further up the stack IIRC

lwander:
@jtk54 the context is that users should be able to refer to a GCS object by the exact revision used.

PR #2673, opened 2018-05-30
perf(aws): Batch up a few cases of `objectMapper.convertValue()`

ajordens:
This appears to have made some difference locally.


ajordens:
TBH I'm not sure how much impact this will have in production with all other caching agents running ... but it did make some difference running locally.

I spent some time pursuing the idea of not prematurely serializing to a `Map` but almost immediately ran into a couple spots where we're passing the serialized `attributes` along and expecting properties of the object to be present.

Potentially still something worth exploring in the future.

PR #2674, opened 2018-05-31
feat(artifacts/bitbucket): added bitbucket artifact support

nvndnl:


nvndnl:
@lwander

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactCredentials.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+    requestBuilder = builder;
+  }
+
+  private String credentialsFromFile(String filename) {
+    try {
+      String credentials = FileUtils.readFileToString(new File(filename));
+      return credentials.replace("\n", "");
+    } catch (IOException e) {
+      log.error("Could not read Bitbucket credentials file {}", filename);
```
lwander: If you add the exception `e` as a parameter: 

```java
  log.error("Could not read Bitbucket credentials file {}", filename, e);
```

It'll get printed to give the user some sense of what went wrong.

nvndnl:
fixed this one too

PR #2679, opened 2018-05-31
chore(dependencies): Update spinnaker-dependencies to 0.157.0

jrsquared:


cfieber:
we do a thing where we will build and commit the generated yml file as well since that is generated from a template and we want to have decent change history on the end state

PR #2682, opened 2018-06-01
fix(titus): correctly name metric

emjburns:


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
```diff
@@ -248,7 +248,7 @@ public CacheResult loadData(ProviderCache providerCache) {
 
     List<Job> jobs = titusClient.getAllJobsWithoutTasks();
     PercentileTimer
-      .get(registry, metricId.withTag("operation", "getAllTasks"))
+      .get(registry, metricId.withTag("operation", "getAllJobs"))
```
ajordens: Do you actually want these metric tags to line up with the actual titus client method names? Noticed they're not 1:1?

emjburns: That's probably a good idea.

PR #2688, opened 2018-06-04
feat(provider/kubernetes): max-version-history annotation

lwander:


lwander:
@danielpeach we could think about implementing with resource pointers for more fine-grained control down the road, but that is largely infeasible when dealing with large bundles of resources, especially in cases where not all resources are submitted to a deploy stage each time.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AtomicOperations.java
```diff
@@ -79,4 +79,7 @@
   public static final String PAUSE_ROLLOUT_MANIFEST = "pauseRolloutManifest";
   public static final String RESUME_ROLLOUT_MANIFEST = "resumeRolloutManifest";
   public static final String UNDO_ROLLOUT_MANIFEST = "undoRolloutManifest";
+
+  // Aritfact operations
```
sbwsg: `Aritfact` -> `Artifact`

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/artifact/KubernetesCleanupArtifactsOperation.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private List<Artifact> artifactsToDelete(KubernetesManifest manifest) {
+    Artifact artifact = KubernetesManifestAnnotater.getArtifact(manifest);
+    List<Artifact> artifacts = artifactProvider.getArtifacts(artifact.getType(), artifact.getName(), artifact.getLocation())
+        .stream()
+        .filter(a -> a.getMetadata() != null && accountName.equals(a.getMetadata().get("account")))
+        .collect(Collectors.toList());
+
+    KubernetesManifestStrategy strategy = KubernetesManifestAnnotater.getStrategy(manifest);
+    if (strategy.getMaxVersionHistory() == null) {
```
sbwsg: Minor nit: It doesn't look like we need to execute lines 98 - 102 if strategy.getMaxVersionHistory() is null.  This code could go at the start of the function to avoid unnecessary work.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AtomicOperations.java
lwander: :+1:

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/artifact/KubernetesCleanupArtifactsOperation.java
lwander: Good catch

PR #2690, opened 2018-06-05
fix(provider/kubernetes): add version to sequence in moniker

lwander:


lwander:
Closes https://github.com/spinnaker/spinnaker/issues/2867

lwander:
@dibyom

PR #2693, opened 2018-06-06
feat(provider/kubernetes): add CRD handler

lwander:
Forget to make these deployable after adding support for caching & deploying resources defined as CRDs.


jtk54:
@sbwsg 's [reaction when a PR opens](https://media.giphy.com/media/r3xE2dPNV9nGg/giphy.webp).

PR #2700, opened 2018-06-08
fix(startup): Fix startup warnings

ezimanyi:
Remove two warnings on startup about optional constructors being effectively required due to the lack of a default constructor.

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/EntityTagsController.java
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/EntityTagsController.java
@@ -42,7 +42,7 @@ public class EntityTagsController {
   private final MessageSource messageSource;
   private final EntityTagsProvider tagProvider;
 
-  @Autowired(required = false)
+  @Autowired
   public EntityTagsController(MessageSource messageSource,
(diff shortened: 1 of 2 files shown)


ezimanyi:
Just cleaning up a couple of warnings in the logs that were easy to resolve.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/EntityTagsController.java
```diff
@@ -42,7 +42,7 @@
   private final MessageSource messageSource;
   private final EntityTagsProvider tagProvider;
 
-  @Autowired(required = false)
+  @Autowired
```
lwander: Are we sure this will work? How It seems like if `MessageSource` can't be constructed this won't startup.

ezimanyi: This was already effectively required as there's no default constructor; I tested injecting an unconstructed bean into this constructor and the application failed to start even before removing `required=false`.  So this will not cause startup to fail in any cases where it isn't already.  (And I'm actually not sure you could ever have a situation where `messageSource` hasn't been constructed as spring provides a default `DelegatingMessageSource` to fall back on.)

PR #2702, opened 2018-06-11
Refactor oracle bmcs

guoyongzhang:
The name oraclebmcs is no longer used in Oracle product offering. To avoid confusion, spinnaker provider "oracebmcs" is replaced with the name "oracle". 
halyard has made this refactoring in https://github.com/spinnaker/halyard/pull/959.


robzienert:
@guoyongzhang Is this just renaming things, are there any backwards-incompatible changes that will require users to change pipeline configurations, etc?

robzienert:
@guoyongzhang How critical is this PR being merged quickly? For example, is there in-flight work that will soon be PR'ed that depends on this refactor?

markxnelson:
We do have work planned that will depend on this, but it is not imminent.  We are going to try to determine if anyone actually ever used this, we suspect not, because it is incomplete as-is, and we are working on finishing it.  For example it does not have server groups and it does not have any support for building images yet.  We certainly do not want to cause any user impact, and we are happy to work with you to minimize the impact, or align it with a release or whatever makes sense.  Can you confirm if this change as-is would require *all* users to do those things, or just people who actually used this provider?

lwander:
These changes (as far as I can tell) only impact users of this provider.

robzienert:
Cool dudes. Thanks for the patience.

PR #2703, opened 2018-06-11
feat(aws): Cluster autoscaling policy tagger

robzienert:
Allows us to record an inferred cluster-level desired state
of autoscaling policies based, and then report when policies
drift out of the desired state.

The concept is kinda gnarly, so there's a bunch of comments to go with.

### DIFF
--- a/clouddriver-elasticsearch-aws/src/main/kotlin/com/netflix/spinnaker/clouddriver/elasticsearch/aws/ClusterScalingPolicyTaggingAgent.kt
+++ b/clouddriver-elasticsearch-aws/src/main/kotlin/com/netflix/spinnaker/clouddriver/elasticsearch/aws/ClusterScalingPolicyTaggingAgent.kt
@@ -1,0 +1,327 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.elasticsearch.aws
+
+import com.amazonaws.services.autoscaling.model.AutoScalingGroup
+import com.amazonaws.services.autoscaling.model.ScalingPolicy
+import com.fasterxml.jackson.core.type.TypeReference
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.frigga.Names
+import com.netflix.spectator.api.Registry
+import com.netflix.spinnaker.cats.agent.RunnableAgent
+import com.netflix.spinnaker.clouddriver.aws.provider.AwsProvider
+import com.netflix.spinnaker.clouddriver.aws.security.AmazonClientProvider
+import com.netflix.spinnaker.clouddriver.aws.security.AmazonCredentials.AWSRegion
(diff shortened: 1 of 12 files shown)


clouddriver-elasticsearch-aws/src/main/kotlin/com/netflix/spinnaker/clouddriver/elasticsearch/aws/ClusterScalingPolicyTaggingAgent.kt
```diff
@@ -0,0 +1,327 @@
...
+    this, object : TypeReference<MutableMap<String, Any?>>() {}
+  ).apply {
+    remove("autoScalingGroupName")
+    remove("policyName")
+    remove("policyARN")
+    set("alarms", alarms.map { mapOf<String, Any?>() })
+  }
+
+/**
+ * This is kinda sad, but Entity Tags are designed to be a map, so if we're storing multiple values, the root-level
```
ajordens: This is likely correctable but just hasn't been all that pressing, potentially as simple as adding a `list` to the `EntityTagValueType` enum (and the associated serialization / deserialization).

```
static enum EntityTagValueType {
    literal, // number or string
    object   // map
  }
```

clouddriver-elasticsearch-aws/src/main/kotlin/com/netflix/spinnaker/clouddriver/elasticsearch/aws/ElasticSearchAmazonCachingAgentProvider.kt
```diff
@@ -67,6 +70,15 @@ open class ElasticSearchAmazonCachingAgentProvider(
         amazonClientProvider,
         credentials,
         elasticSearchClient
+      ),
+      ClusterScalingPolicyTaggingAgent(
```
ajordens: I would probably jam this agent into a different provider as the rest of the elastic search instance and server group caching agents are currently disabled in production (and we should be able to enable the scaling policy tagging independently)

ajordens:
Will take a closer read through the actual logic but at a glance it seems reasonable to me.

clouddriver-elasticsearch-aws/src/main/kotlin/com/netflix/spinnaker/config/ClusterScalingPolicyTaggingAgentProperties.kt
```diff
@@ -0,0 +1,26 @@
...
+package com.netflix.spinnaker.config
+
+import java.time.Duration
+
+data class ClusterScalingPolicyTaggingAgentProperties(
+  /**
+   * [oldServerGroupBoundaryMs] defines how long a server group must exist
+   * before the agent will consider a missing scaling policy as desired.
+   */
+  var oldServerGroupBoundaryMs: Long = Duration.ofMinutes(30).toMillis()
```
ajordens: Thoughts on whether this is too aggressive of a value? 

Perhaps I'm misunderstanding the comment, 30min seems a touch aggressive to be deemed the new norm if something goes sideways?

clouddriver-elasticsearch-aws/src/main/kotlin/com/netflix/spinnaker/clouddriver/elasticsearch/aws/ClusterScalingPolicyTaggingAgent.kt
robzienert: It's not that big of a deal to me, just wanted to add a comment so that future me doesn't forget why I'm doing it (and if it changes in the future, it'll be an easy call out to refactor).

clouddriver-elasticsearch-aws/src/main/kotlin/com/netflix/spinnaker/config/ClusterScalingPolicyTaggingAgentProperties.kt
robzienert: Hmm, you're probably right. I was operating under the idea that a deploy operation won't (typically) take very long, but you're right it doesn't protect us much if things go sideways. I'll set it to something more like 8 hours.

PR #2704, opened 2018-06-11
fix(titus): Ensure containerAttributes values are strings

robzienert:
Deploy operations explode if container attributes are received as non-strings.


emjburns:
Seems reasonable 👍

tomaslin:
gold star

PR #2705, opened 2018-06-11
fix(provider/dcos): Cache per cluster/service account pair.

mtweten:
Caches DC/OS data at the cluster level instead of account level for DC/OS accounts. This prevents duplicate caching and API calls when multiple Spinnaker accounts exist for the same DC/OS cluster.

### DIFF
--- a/clouddriver-dcos/src/main/groovy/com/netflix/spinnaker/clouddriver/dcos/provider/agent/DcosClusterAware.java
+++ b/clouddriver-dcos/src/main/groovy/com/netflix/spinnaker/clouddriver/dcos/provider/agent/DcosClusterAware.java
@@ -1,0 +1,13 @@
+package com.netflix.spinnaker.clouddriver.dcos.provider.agent;
+
+import com.netflix.spinnaker.clouddriver.dcos.security.DcosAccountCredentials;
+
+import java.util.Collection;
+
+public interface DcosClusterAware {
+  String getClusterName();
+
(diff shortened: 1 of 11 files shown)


clouddriver-dcos/src/main/groovy/com/netflix/spinnaker/clouddriver/dcos/provider/agent/DcosClusterAware.java
```diff
@@ -0,0 +1,13 @@
+package com.netflix.spinnaker.clouddriver.dcos.provider.agent;
```
lwander: This should have a copyright header

clouddriver-dcos/src/main/groovy/com/netflix/spinnaker/clouddriver/dcos/provider/agent/DcosServerGroupCachingAgent.groovy
```diff
@@ -374,14 +388,14 @@ class DcosServerGroupCachingAgent implements CachingAgent, AccountAware, OnDeman
     log.info("Caching ${cachedInstances.size()} instances in ${agentType}")
 
     new DefaultCacheResult([
-                             (Keys.Namespace.SERVER_GROUPS.ns) : cachedServerGroups.values(),
```
lwander: In general, to keep these more reviewable please separate formatting & functionality changes.

PR #2708, opened 2018-06-12
fix(entity_tagger): allow for no server group providers

nvndnl:
@lwander


lwander:
Looks like the tests need to be updated: https://travis-ci.org/spinnaker/clouddriver/builds/391431066#L1313

PR #2712, opened 2018-06-13
feat(provider/gce): Adds support for autoscaling policy mode in upsert.

jtk54:
Also adds mode handling for enable disable operations. Still needs a bit of work in Orca for capacity pinning on GCE deploys.


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/AbstractEnableDisableAtomicOperation.groovy
```diff
@@ -109,6 +111,13 @@ abstract class AbstractEnableDisableAtomicOperation extends GoogleAtomicOperatio
     }
 
     if (disable) {
+    } else {
```
danielpeach: Can this be removed?

jtk54:
Orca work is here: https://github.com/spinnaker/orca/pull/2284.

PR #2715, opened 2018-06-14
fix(provider/dcos): Check if Integer null before unboxing

rapaliwal:
Makes sure that a primitive integer is not assigned null in case input is not provided


clouddriver-dcos/src/main/groovy/com/netflix/spinnaker/clouddriver/dcos/deploy/util/mapper/DeployDcosServerGroupDescriptionToAppMapper.groovy
```diff
@@ -129,8 +129,8 @@ class DeployDcosServerGroupDescriptionToAppMapper {
           protocol = readinessCheck.protocol
           path = readinessCheck.path
           portName = readinessCheck.portName
-          intervalSeconds = readinessCheck.intervalSeconds
-          timeoutSeconds = readinessCheck.timeoutSeconds
+          intervalSeconds = readinessCheck.intervalSeconds ? readinessCheck.intervalSeconds : 0
```
lwander: Given that this is groovy, the "canonical way" to write this is

```groovy
intervalSeconds = readinessCheck.intervalSeconds ?: 0
```

PR #2719, opened 2018-06-15
fix(provider/amazon): ignore not found when already deleted

emjburns:
If we get this exception when it's already deleted, we should ignore it. That way pipelines don't fail because of this.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/DestroyAsgAtomicOperation.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/DestroyAsgAtomicOperation.groovy
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.clouddriver.aws.deploy.ops
 
 import com.amazonaws.AmazonClientException
+import com.amazonaws.services.autoscaling.model.AmazonAutoScalingException
 import com.amazonaws.services.autoscaling.model.AutoScalingGroup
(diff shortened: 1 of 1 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/DestroyAsgAtomicOperation.groovy
```diff
@@ -93,8 +94,17 @@ class DestroyAsgAtomicOperation implements AtomicOperation<Void> {
...
+      try {
+        autoScaling.deleteLaunchConfiguration(
+          new DeleteLaunchConfigurationRequest(launchConfigurationName: autoScalingGroup.launchConfigurationName)
+        )
+      } catch (AmazonAutoScalingException e) {
+        if (e.message.toLowerCase().contains("launch configuration name not found")) {
+          // Ignore, because we are trying to delete it.
+        } else {
+          throw e
+        }
```
robzienert: Personal preference here, but, I'd switch it to look like so:

```
try {
  autoScaling.delete...
} catch (AmazonAutoScalingException e) {
  // ignore not found exceptions
  if (!e.message.toLowerCase().contains("launch configuration name not found")) {
    throw e;
  }
}
```

emjburns: Sure, I'll switch.

PR #2722, opened 2018-06-19
feat(kubernetes): surface kind mapping to credentials API

sbwsg:
Expose the map from k8s resource kinds to Spinnaker kinds.  E.g. ReplicaSet -> ServerGroup; Deployment -> ServerGroupManager

This will be consumed by the UI to correctly route the user from a resource deployed during an execution to its details view.

Fixes https://github.com/spinnaker/spinnaker/issues/2496


sbwsg:
@lwander updated to include CRDs

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesNamedAccountCredentials.java
```diff
@@ -167,6 +171,20 @@ public Permissions getPermissions() {
     return permissions;
   }
 
+  public Map<String, String> getSpinnakerKindMap() {
+    if (kubernetesSpinnakerKindMap == null) {
+      return new HashMap<String, String>();
+    }
+    Map<String, String> kindMap = kubernetesSpinnakerKindMap.kubernetesToSpinnakerKindStringMap();
```
lwander: Maybe we should make a copy of this? I think by modifying it below we modify the global copy

lwander:
```java
Map<String, String> kindMap = new HashMap<>(kubernetesSpinnakerKindMap.kubernetesToSpinnakerKindStringMap());

sbwsg:
kindMap is now explicitly copied before CRD entries are added to it.

PR #2726, opened 2018-06-20
fix(provider/kubernetes): fix v2 error message

lwander:


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubectlJobExecutor.java
```diff
@@ -166,7 +166,13 @@ public String logs(KubernetesV2Credentials credentials, String namespace, String
     JobStatus status = backoffWait(jobId, credentials.isDebug());
 
     if (status.getResult() != JobStatus.Result.SUCCESS) {
-      throw new KubectlException("Failed to delete " + kind + "/" + name + " from " + namespace + ": " + status.getStdErr());
+      String id;
```
duftler: Don't like `?:`?

lwander: Not groovy

PR #2736, opened 2018-06-22
fix(provider/kubernetes): fix regex perf bug when parsing images

lwander:
Tests checking for the validity of an image were removed, since they are
only exercised on already deployed manifest objects. If kubernetes
accepts a docker image, we don't provide any value in refusing to parse
it. Also, parsing it turned out to be very expensive for certain image
configurations.


lwander:
Closes https://github.com/spinnaker/spinnaker/issues/2742

PR #2738, opened 2018-06-22
fix(aws): Equality of `RateLimitingRequestHandler` should not depend on counter

ajordens:
There is no guarantee that a `Counter` will be re-used across requests.

At Netflix, we're seeing an issue wherein the `AwsSdkClientSupplier`
has a zero hit rate due to the `Counter` reference changing between
requests to the same account/region.

This PR also introduces a TTL on the sdk clients cache (10min after access)
to offer future protection against poor hash/equals performance.


emjburns:
LGTM

PR #2746, opened 2018-06-27
fix(aws): don't respecifiy encryption if the snapshotId is provided

kevinawoo:
This should fix https://github.com/spinnaker/spinnaker/issues/2958


dthomson25:
LGMT

PR #2747, opened 2018-06-27
fix(aws): don't respecifiy encryption if the snapshotId is provided

kevinawoo:
same as https://github.com/spinnaker/clouddriver/pull/2746, but for `release-1.8.x`


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
```diff
@@ -483,7 +483,13 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
         device.size = volumeSize
         device.volumeType = volumeType
         device.snapshotId = snapshotId
-        device.encrypted = encrypted
+        if (snapshotId != null) {
```
ajordens: could this be

```
if (snapshotId == null) {
  // aws only allows specifying the encrypted flag when a snapshotId has not been provided 
  device.encrypted = encrypted
}
```

(this way you'd avoid having to write both the if/else)

kevinawoo: sure 👍 for cleanup

PR #2748, opened 2018-06-28
fix(provider/eureka): change Eureka timeout to 5 minutes

cfieber:


cfieber:
> clin [1:43 PM]
> ah, the timeout
> make sense
> LGTM

PR #2751, opened 2018-06-29
feat(provider/kubernetes): add ID field to k8s manifest objects

lwander:
@sbwsg pushed in case it helps you with https://github.com/spinnaker/spinnaker/issues/2979


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/Instance.java
```diff
@@ -14,58 +14,70 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.clouddriver.model
+package com.netflix.spinnaker.clouddriver.model;
```
lwander: Changed to java to make use of default interface methods

PR #2754, opened 2018-07-02
feat(provider/kubernetes): add UID field to LoadBalancerInstance

sbwsg:
As with https://github.com/spinnaker/clouddriver/pull/2751 this is to get unique IDs into Deck for Kubernetes instances.


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/LoadBalancerInstance.java
```diff
@@ -29,6 +30,21 @@
 @NoArgsConstructor
 public class LoadBalancerInstance {
   String id;
+  String uid;
   String zone;
   Map<String, Object> health;
+
+  public LoadBalancerInstance(String id, String zone, Map<String, Object> health) {
```
sbwsg: This form of the constructor is used by `TitusTargetGroupServerGroupProvider`

lwander: :+1:

PR #2756, opened 2018-07-03
feat(provider/kubernetes): make id field unique, add name field for full resource name

sbwsg:
DONT MERGE until a parallel PR in Deck has been approved (https://github.com/spinnaker/deck/pull/5506).

The `Instance` interface has an implicit constraint on it that the `name` field should return a globally unique ID for the resources it fronts.  This constraint is relied upon by Deck to disambiguate between instances in various UI contexts.  A non-unique string name can cause some strange UI behaviour.  See here for an example: https://github.com/spinnaker/spinnaker/issues/2979

Previously, a k8s v2 resource returned its "fullResourceName" as its `Instance.name` but this could lead to non-unique IDs being serialized by an InstanceViewModel or LoadBalancerInstance.  This PR makes the k8s v2 provider return a resource's uuid in place of its name when responding to requests for ServerGroup Instances or LoadBalancer Instances and also introduces a `name` field to those serialized responses so that the fullResourceName is still available to consumers of those APIs.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/Instance.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/Instance.java
@@ -26,18 +26,18 @@ import java.util.Map;
  */
 public interface Instance {
   /**
-   * The name of the instance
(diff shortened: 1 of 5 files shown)


erikmunson:
❤️thanks @sbwsg

PR #2758, opened 2018-07-04
fix(provider/titus): only fetch tasks if includeDetails is requested

tomaslin:


emjburns:
LGTM if you verified that the API responses are what they should be

PR #2759, opened 2018-07-04
feat(provider/appengine): Enables App Engine Standard Java 8.

woogie-wolgemuth:
This enables the App Engine provider to deploy Standard Java App Engine apps. This change only modifies the container, changing:

* The base image from JRE to JDK. `gcloud app deploy` always does inspection/compilation when deploying Java apps.
* The Google Cloud SDK install to include the Java App Engine components (also removing a deprecated flag in the install command).
* The current working directory. An issue with the Google Cloud SDK breaks Java App Engine deployments if the cwd is '/'.

This was tested against the App Engine Standard Java 8 "getting started" app "helloworld" from https://github.com/GoogleCloudPlatform/getting-started-java/tree/master/appengine-standard-java8/helloworld

Approach was discussed with @danielpeach offline.

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -1,4 +1,4 @@
-FROM openjdk:8-jre-alpine
+FROM openjdk:8-jdk-alpine
 
 MAINTAINER delivery-engineering@netflix.com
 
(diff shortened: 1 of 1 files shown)


Dockerfile.slim
```diff
@@ -1,4 +1,4 @@
-FROM openjdk:8-jre-alpine
+FROM openjdk:8-jdk-alpine
```
lwander: Is it possible to use `jre` in place of `jdk`? Reason I'm asking is this change makes the docker image quite a bit larger.

woogie-wolgemuth: Unfortunately, no - running `gcloud app deploy` for Java 8 Standard applications does rely on having a JDK available.

lwander: Oh well, LGTM

PR #2760, opened 2018-07-05
feat(kubernetes): Add namespace support in deploy k8s manifest stage

posox:
Issue: https://github.com/spinnaker/spinnaker/issues/2831


lwander:
Assuming we go down this path I think it'd be safer to overwrite `metadata.namespace` in the manifest object for every submitted manifest

posox:
@lwander no, we can use different namespaces in one manifest. For example, kube-dns and other system k8s objects can be created in `kube-system` namespace. Other objects in user namespace. Usually this namespace is hardcoded in `metadata.namespace` and we don't want to override it.

I can prepare patch which will be override `metadata.namespace` if this field is not defined in manifest. But I don't see big difference between overriding and using CLI option in `kubectl`.

astlock:
Hey guys. Any progress on this?

ezimanyi:
@posox : I'm cleaning up old PRs...looks like this hasn't had activity in a while.  Mind if I close it, and you can reopen if/when you revisit this?

ezimanyi:
Closing for now, but please feel free to re-open if you pick up this work again.  Happy to discuss this more as well.

PR #2765, opened 2018-07-10
fix(provider/titus): support specifying a digest as part of imageId

tomaslin:


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/JobDescription.java
```diff
@@ -451,7 +457,16 @@ public JobDescriptor getGrpcJobDescriptor() {
     }
 
     containerBuilder.setSecurityProfile(securityProfile);
-    containerBuilder.setImage(Image.newBuilder().setName(applicationName).setTag(version));
+
+    Image.Builder imageBuilder = Image.newBuilder();
+    imageBuilder.setName(applicationName);
+    if(digest!=null){
```
emjburns: Am I correct in assuming that you can have *either* a digest or a tag?

tomaslin: yes, they are exclusive

PR #2766, opened 2018-07-10
fix(titus): really ignore not found on terminate

emjburns:
Trying again to actually ignore 404 on terminate and shrink.

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/RegionScopedTitusClient.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/RegionScopedTitusClient.java
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.clouddriver.titus.client;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.protobuf.Empty;
 import com.netflix.frigga.Names;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.clouddriver.titus.TitusException;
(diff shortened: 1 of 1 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/RegionScopedTitusClient.java
```diff
@@ -226,17 +227,22 @@ public void terminateTasksAndShrink(TerminateTasksAndShrinkJobRequest terminateT
...
-        log.warn("Titus task {} not found, continuing with terminate tasks and shrink job request.", id);
-      } else {
-        retrySupport.retry(() ->
-            TitusClientAuthenticationUtil.attachCaller(grpcBlockingStub).killTask(TaskKillRequest.newBuilder().setTaskId(id).setShrink(terminateTasksAndShrinkJob.isShrink()).build())
-          , 2, 1000, false);
-      }
-    }
+      retrySupport.retry(() -> {
+        try {
+          return TitusClientAuthenticationUtil.attachCaller(grpcBlockingStub).killTask(
```
tomaslin: no return statement needed

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/RegionScopedTitusClient.java
```diff
@@ -226,17 +227,22 @@ public void terminateTasksAndShrink(TerminateTasksAndShrinkJobRequest terminateT
...
+            TaskKillRequest.newBuilder()
+              .setTaskId(id)
+              .setShrink(terminateTasksAndShrinkJob.isShrink())
+              .build()
+          );
+        } catch (io.grpc.StatusRuntimeException e) {
+          if (e.getStatus() == Status.NOT_FOUND) {
+            log.warn("Titus task {} not found, continuing with terminate tasks and shrink job request.", id);
+            return Empty.newBuilder().build();
```
tomaslin: don't need a return here either

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/RegionScopedTitusClient.java
emjburns: @tomaslin If I don't add both returns `retrySupport` yells at me, because the function it runs is supposed to return a value. Worth me changing the interface to add support for a function w/o a return type, or just keeping the return both places?

tomaslin: that's annoying. Yeah, either leave as it or change the signature works. LGTM otherwise

PR #2771, opened 2018-07-11
chore(Dockerfile): Update python version to match alpinelinux. (#2770)

jtk54:


jtk54:
Cherry pick of https://github.com/spinnaker/clouddriver/pull/2770.

PR #2772, opened 2018-07-11
feat(provider/kubernetes): Allow image selection by digest

ezimanyi:
In order to support deploying a container injected into a pipeline as an artifact, allow the Kubernetes provider to deploy images specified using a digest.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/KubernetesUtil.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/KubernetesUtil.groovy
@@ -63,32 +63,43 @@ class KubernetesUtil {
     rc.spec?.template?.spec?.imagePullSecrets?.collect({ it.name })
   }
 
-  static KubernetesImageDescription buildImageDescription(String image) {
-    def sIndex = image.indexOf('/')
-    def result = new KubernetesImageDescription()
-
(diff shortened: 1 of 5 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/KubernetesUtil.groovy
```diff
@@ -63,32 +63,43 @@ class KubernetesUtil {
...
     }
+    image
+  }
 
-    def cIndex = image.indexOf(':')
-
-    if (cIndex < 0) {
-      result.repository = image
+  private static String extractDigestOrTag(String image, KubernetesImageDescription description) {
+    def digestIndex = image.indexOf('@')
```
lwander: I think `.lastIndexOf('@')` is safer, as the image url could look like `username:password@example.org/repo/image@sha256:abababa`

lwander: Ah that makes the below check tricky... let me check the spec

lwander: Looks like the only occurrence of `@` is in the digest, so you're good to go: https://github.com/docker/distribution/blob/master/reference/regexp.go

lwander:
https://github.com/spinnaker/spinnaker/issues/2962

PR #2773, opened 2018-07-12
feat(provider/kubernetes): Parse container URI in clouddriver

ezimanyi:
Allow Orca to send just the container URI to clouddriver, so that the parsing logic in Clouddriver can be used to parse this into registry, repository, and tag, avoiding the need to re-implement this in Orca.


ezimanyi:
https://github.com/spinnaker/spinnaker/issues/2962

ezimanyi:
@lwander : One small update to the PR you just reviewed.  I realized that by having clouddriver accept just a URI and parse it into the container fields, I can avoid re-implementing the parsing logic in Orca.  This makes it trivial for Orca to just pass the URI along and let clouddriver handle the rest.

PR #2783, opened 2018-07-16
fix(provider/kubernetes): bind initContainers

ethanfrogers:
adds support for binding initContainer images in the
`dockerImageReplacer`.

fixes spinnaker/spinnaker#2535


ethanfrogers:
@lwander PTAL. i haven't tested it _yet_ but the tests pass and i tested the path [with this](https://jsonpath.herokuapp.com/). LMK what you think!

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/artifact/ArtifactReplacerFactory.java
```diff
@@ -66,6 +66,7 @@ public static Replacer dockerImageReplacer() {
         .build();
   }
 
+
```
lwander: Extra newline?

ethanfrogers:
Yep, will add a test before merging!

PR #2786, opened 2018-07-17
feat(provider/docker): Add metrics to docker registry provider to instrument number of calls made (#2963)

Remz-Jay:
This change introduces Spectator metrics for each request made by `DockerRegistryClient`'s HTTP Client.

In order for that to work, the global `Registry` had to be `@Autowired` in, and since both `DockerRegistryClient` and `DockerRegistryNamedAccountCredentials` are Builders, it had to be injected in `DockerRegistryCredentialsInitializer` and passed down from there.

Metrics are tagged by `host`, `path` ( and `successful` for the counter). This allows multi-cloud users to track their rate limits by top level repository.

Even though `DockerRegistryClientSpec` is `@Ignore`d, I refactored the tests a bit to be compatible with these changes, and added a new spec to verify the metrics/logs introduced by this change.

Internal review was done at https://github.com/bolcom/clouddriver/pull/1 .

Closes spinnaker/spinnaker#2963.


clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/metrics/UrlMetricsInstrumentation.groovy
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright 2018 Netflix, Inc.
```
lwander: I think this should be Bol rather than Netflix

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/metrics/UrlMetricsInstrumentation.groovy
```diff
@@ -0,0 +1,65 @@
...
+
+  private final Registry registry
+
+  private final Id timingId
+  private final Id counterId
+
+  @Autowired
+  UrlMetricsInstrumentation(Registry registry) {
+    this.registry = registry
+    timingId = registry.createId('UrlRequestTime').withTag('className', UrlMetricsInstrumentation.simpleName)
```
lwander: I could be wrong, but don't we usually write measurements with a leading lowercase character?

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/metrics/UrlMetricsInterceptor.groovy
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2018 Netflix, Inc.
```
lwander: Same copyright comment

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/metrics/UrlMetricsInstrumentation.groovy
Remz-Jay: Correct! I didn't even notice that. While I was changing the metric name, i also introduced two more changes to make it better:

- Most metrics under Clouddriver are namespaced (like `redis.command.*`, `jedis.pool.*`). Applied the same for url metrics; namespaced under `httpClient.url.*` (now with proper casing as well obviously).
- Updated the `className` shown for url metrics from`UrlMetricsInstrumentation` to `DockerRegistryClient`, which actually owns the `httpClient` that emits these metrics. Should make for more logical labeling.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/metrics/UrlMetricsInstrumentation.groovy
Remz-Jay: Thanks; updated both instances.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/metrics/UrlMetricsInstrumentation.groovy
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  Id getCounterId() {
+    return counterId
+  }
+
+  void interceptorCompleted(Response response, Instant start, Instant end) {
+    Duration duration = Duration.between(start, end)
+    URL url = response.request().url()
+    registry.timer(timingId.withTag('host', url.getHost()).withTag('path', url.getPath())).record(duration.toMillis(), TimeUnit.MILLISECONDS)
```
ewiseblatt: What are the path values? How many paths are we talking about? Regarding success, usually there is a categorization of errors so you know the types of failures. For that matter, if there is structure to the path then you might have that here instead of a flat path. And the path itself sounds suspicious since if this is an HTTP API then you can have different methods on the same path.

My concerns here are:
   1) Unbounded paths can strain the underlying metric stores and make it hard on the UX to make sense of
   2) Unstructured paths dont give much opportunity for aggregation
   3) Failures without the cause seems incomplete and forces more work to find the cause to determine if it is of interest or not

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/metrics/UrlMetricsInstrumentation.groovy
ewiseblatt: what is the value of className metadata on the metric?
why is this "UrlRequestTime" instead of something like "docker.registry.api"

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/metrics/UrlMetricsInstrumentation.groovy
```diff
@@ -0,0 +1,65 @@
...
+
+  private final Id timingId
+  private final Id counterId
+
+  @Autowired
+  UrlMetricsInstrumentation(Registry registry) {
+    this.registry = registry
+    timingId = registry.createId('UrlRequestTime').withTag('className', UrlMetricsInstrumentation.simpleName)
+    counterId = registry.createId('UrlCount').withTag('className', UrlMetricsInstrumentation.simpleName)
+  }
```
ewiseblatt: The spinnaker timer metrics have both a count and a time, so this UrlCount is superfluous.

In the timer, there is a "statistic" label which is either "count" or "totalTime". The "count" acts as a call count, the "totalTime" is the time counter. I dont like that it works like that but it is what it is. If you are using the monitoring daemon, it will split a timers into the two separate counters *__count and *__totalTime.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/metrics/UrlMetricsInstrumentation.groovy
```diff
@@ -0,0 +1,65 @@
...
+  Id getCounterId() {
+    return counterId
+  }
+
+  void interceptorCompleted(Response response, Instant start, Instant end) {
+    Duration duration = Duration.between(start, end)
+    URL url = response.request().url()
+    registry.timer(timingId.withTag('host', url.getHost()).withTag('path', url.getPath())).record(duration.toMillis(), TimeUnit.MILLISECONDS)
+    registry.counter(counterId.withTag('host', url.getHost()).withTag('path', url.getPath()).withTag('successful', response.isSuccessful())).increment()
+    log.info format("Received %d response for %s in %dms", response.code(), url, duration.toMillis())
```
ewiseblatt: Do you need to log these given you just instrumented it ?

ewiseblatt:
@Remz-Jay @lwander I just noticed this PR and have some concerns about it. It's fine in principal, but the actual metric structure warrants some discussion.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/metrics/UrlMetricsInstrumentation.groovy
Remz-Jay: The number of possible paths is limited by two factors:

1. The number of methods in the `DockerRegistryService` interface (See: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy#L181 )
2. The number of `dockerRegistry.repositories` entries in your CloudDriver configuration.

So it's not unbounded, but in fact quite comparable to the number of `agent` tags in `clouddriver:executionTime__totalTime` for instance (at least in our configuration), since the `@GET("/v2/{repository}/tags/list")` is the most commonly used query for reach repository entry. (There are only `GET` methods in this interface by the way)

Example, for a `clouddriver.yaml` with a single repository:
```
curl -s http://localhost:7002/spectator/metrics | jq '.metrics["httpClient.url.count"].values[].tags | map(select(.key | contains ("path"))) | .[] .value'
"/v2/bolcom-stg-spinnaker-6b3/provisioning-image/tags/list"
"/v2/"
"/v2/"

curl -s http://localhost:7002/spectator/metrics | jq '.metrics["httpClient.url.requestTime"].values[].tags | map(select(.key | contains ("path"))) | .[] .value'
"/v2/bolcom-stg-spinnaker-6b3/provisioning-image/tags/list"
"/v2/bolcom-stg-spinnaker-6b3/provisioning-image/tags/list"
"/v2/"
"/v2/"
```

That also brings me to the *why* part of these metrics; We want to see if that `/tags/list` call is in fact the URL that gets called, and how often, and if not, which others get called instead.

We've discussed aggregating this by repository instead of by URL called, but that would mean implementing a lot of logic into the interceptor that's tightly coupled to the RestAdapter interface (which is bound to change), so we thought it best to leave it as generic as possible.
We could make an exception for the `/tags/list` call, and tag that with the registry being used?

As for the "success" part: I based the result on `response.isSuccessful()` because we're mainly interested in seeing if the client encountered an error while communicating with the registry. This could obviously easily be switched to a tag using the `response.code()` field, giving more insight on the behavior of the remote registries, but would be slightly heavier on the metric stores. 

Let me know if and what you'd like to see changed based on this input, and I'll implement those changes.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/metrics/UrlMetricsInstrumentation.groovy
Remz-Jay: Good question!

We (I used to be in the team that manages the Metrics, Monitoring and Logging stores, back-ends and infrastructure) always tell developers *"If it's not an error; don't log it, but instrument it instead"*, and here I find myself logging something that's not an error.

To show why, here's the full logging output for a single `DockerRegistryImageCachingAgent` execution:

```
2018-07-26 12:24:53.790  INFO 52852 --- [ecutionAction-1] c.n.s.c.d.r.m.UrlMetricsInstrumentation  : Received 200 response for https://eu.gcr.io/v2/bolcom-stg-spinnaker-6b3/provisioning-image/tags/list in 252ms
2018-07-26 12:24:53.791  INFO 52852 --- [ecutionAction-1] .d.r.p.a.DockerRegistryImageCachingAgent : Describing items in my-docker-registry/DockerRegistryImageCachingAgent[1/1]
2018-07-26 12:24:53.791  INFO 52852 --- [ecutionAction-1] .d.r.p.a.DockerRegistryImageCachingAgent : Caching 1 tagged images in my-docker-registry/DockerRegistryImageCachingAgent[1/1]
2018-07-26 12:24:53.791  INFO 52852 --- [ecutionAction-1] .d.r.p.a.DockerRegistryImageCachingAgent : Caching 1 image ids in my-docker-registry/DockerRegistryImageCachingAgent[1/1]
2018-07-26 12:24:53.794  INFO 52852 --- [ecutionAction-1] c.n.s.c.cache.LoggingInstrumentation     : com.netflix.spinnaker.clouddriver.docker.registry.provider.DockerRegistryProvider:my-docker-registry/DockerRegistryImageCachingAgent[1/1] completed in 0.256s
```

5 lines of logs, of which (in my opinion) 1 (Describing items) is useless, 3 are moderately informative (they contain a single variable bit of information), and the one this PR adds is the most informative (since it adds 3 pieces of information, and provides valuable context to the total execution time, since apparently the HTTP call takes up almost all of the time taken).

I added it so we could get a quick glance in Kibana as well, giving more context to the logs around them. That said, I totally understand why we'd want to keep the logging down to a minimum and could live without this line. (Although I think other, less informative, log events should be removed instead)

Let me know if you want it to stay or go, and I'll adjust it accordingly.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/metrics/UrlMetricsInstrumentation.groovy
ewiseblatt: Again, I'm not familiar with the API or usage patterns, but looking at the interface in the link your provided has "/{path}" which is what concerns me. While you dont seem to mention it in your example, you wont be the only user and I'm not familiar with access patterns. I'm not against it if it makes sense, but I'm not in a position to know if it makes sense or not and if it doesnt, then I'd like to guard against it. So could we hold off on generalizing the metric. If you are also logging calls then you still have a means to see the details.

For tags/list you could have the number of calls with a label for the repository. I imagine that would make it easier for you to answer your questions because you can filter by repository rather than by url and be more explicit about the repository (which is more concise all around).

For success, the other APIs tend to have a true/false "success" label, a label with the individual code, and a "statusCode" label that partitions the responses into 2xx, 4xx, 5xx. Having a "statusCode" could be useful but maybe not for this particular case. You can leave out the individual codes for now if they arent needed. Easy enough to add later.

ewiseblatt: To clarify, I'm sure having {path} instrumentation would be useful. I just wish I knew more about the API.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/metrics/UrlMetricsInstrumentation.groovy
ewiseblatt: Use your discretion. As I noted above, if you are logging calls then you dont need to instrument the /{path} calls for now. So you should keep at least some of this.

Remz-Jay:
@ewiseblatt Based on your comments I've restructured the way the metrics are recorded in 72fe098.

An example for a single image:

```
curl -s http://localhost:7002/spectator/metrics | jq -c '.metrics["httpClient.url.requestTime"] | .values[] | [([(.tags[] | join(":"))] | join(" ")), (.values[].v|tostring)] | join(" = ")'
"className:DockerRegistryClient image:provisioning-image operation:tags/list registry:eu.gcr.io repository:bolcom-stg-spinnaker-6b3 statistic:count statusCode:2xx success:true = 27"
"className:DockerRegistryClient image:provisioning-image operation:tags/list registry:eu.gcr.io repository:bolcom-stg-spinnaker-6b3 statistic:totalTime statusCode:2xx success:true = 8248000000"
"className:DockerRegistryClient operation:other registry:eu.gcr.io statistic:count statusCode:2xx success:true = 3"
"className:DockerRegistryClient operation:other registry:eu.gcr.io statistic:totalTime statusCode:2xx success:true = 48000000"
"className:DockerRegistryClient operation:other registry:eu.gcr.io statistic:count statusCode:4xx success:false = 1"
"className:DockerRegistryClient operation:other registry:eu.gcr.io statistic:totalTime statusCode:4xx success:false = 17000000"
```

Calls to `/v2/` (like the ones recorded in the example) are now (along with all other stuff like `/{path}`) categorized under 'other', and details about them can be extracted from (the now even more relevant) line in the logs. 

The duplicate counter has been removed, along with some other minor changes; see the commit message for details.

pvdissel:
Any updates on this?
Ideas when this is going to be merged? Or is there anything that still needs to be done?

tomaslin:
can you pull master and resolve the merge conflict?

Remz-Jay:
@tomaslin : Rebased the branch on `master`, resolved the conflict and updated the (Ignored) `DockerRegistryClientSpec` to match the upstream changes (an extra `passwordCommand` was added that wasn't present in the tests yet).

Remz-Jay:
@lwander @ewiseblatt @tomaslin Hey guys, is there any particular reason why this PR wasn't accepted yet? (Anything I/we can do to fix?)

If you feel like talking about it; we're around at the Spinnaker Summit and up for a chat.

ezimanyi:
@ewiseblatt : Anything blocking merging this from your perspective (other than rebase and fix conflicts)?  I'm trying to clean up our old PRs and looks like the only thing outstanding is your comment above.

robzienert:
Over a year old - closing.

PR #2787, opened 2018-07-17
fix(provider/openstack): fix several issues in glance v2 support

posox:
* location value can be empty
* convert image properties to Map<String, String>

https://github.com/spinnaker/spinnaker/issues/3018


posox:
Tested with Openstack Ocata and Devstack cloud

posox:
@spinnaker/reviewers

posox:
Please review: @spinnaker/reviewers

PR #2788, opened 2018-07-17
fix(provider/kubernetes): set cluster type in k8s v2 provider

posox:
Return cluster type "kubernetes.v2" instead null

bug: https://github.com/spinnaker/spinnaker/issues/3056


posox:
Please review: @spinnaker/reviewers

posox:
@spinnaker/reviewers: please review / merge

PR #2791, opened 2018-07-18
feat(provider/kubernetes): adding heptio for seamless iam auth on aws

anshrma:


Dockerfile.slim
```diff
@@ -16,6 +19,12 @@ RUN wget https://storage.googleapis.com/kubernetes-release/release/stable.txt &&
   chmod +x kubectl && \
   mv ./kubectl /usr/local/bin/kubectl
 
+RUN curl -o heptio-authenticator-aws https://amazon-eks.s3-us-west-2.amazonaws.com/${KUBECTL_RELEASE}/${HEPTIO_BINARY_RELEASE_DATE}/bin/linux/amd64/heptio-authenticator-aws && \
```
ethanfrogers: @anshrma take a look at how we're getting the version for `kubectl` in the lines above. i think we want to replicate that (i.e. `cat stable.txt`) to get the version instead of defining `KUBECTL_RELEASE` for consistency.

anshrma: Hi @ethanfrogers - Unfortunately, there is no file having meta or file similar to stable.txt. Its hosted on S3 with two levels in folder structure (one being kubectl version and another one being release date).

ethanfrogers: ah, sorry, i just meant do the same thing that we do to get `kubectl` 

```
RUN wget https://storage.googleapis.com/kubernetes-release/release/stable.txt && https://amazon-eks.s3-us-west-2.amazonaws.com/$(cat stable.txt)/${HEPTIO_BINARY_RELEASE_DATE}/bin/linux/amd64/heptio-authenticator-aws && rm stable.txt...
```

anshrma: Got it. However, IMO, that might break in future.
Example :  kubectl version 11.0 gets released in stable.txt. AWS S3 URL to install heptio does not gets updated / need not be updated (likely).

Thoughts ?

ethanfrogers: That's a fair point. Sounds good to me. @sbwsg or @danielpeach thoughts?

sbwsg: The name `KUBECTL_RELEASE` becomes a bit confusing in this case since we now have different approaches of establishing the kubectl release happening in the same Dockerfile.

Would you be amenable to naming it `HEPTIO_KUBECTL_RELEASE` or something similar just to indicate the var's intention?  A comment to explain it would be useful as well.  Something like, `# Heptio releases aren't necessarily made in lockstep with kubectl stable releases so specify version here`.

Also, do we definitely need an environment variable for this?  Do we expect users to override it?

anshrma: Good point. ENV variables were carried over from Halyard, which followed a different convention. I dont foresee users overriding this. Good to coding it directly ?

sbwsg: I'm good with that.

sbwsg:
Thanks!

sbwsg:
Closes https://github.com/spinnaker/spinnaker/issues/2871

jamfdevops:
Any chance this could be moved up to a 1.8.x release?  I'm needing this to deploy our application to our Prod EKS cluster.

lwander:
We only merge fixes into existing release branches -- version `master-latest-unvalidated` (our nightly build) has this change, though.

PR #2793, opened 2018-07-18
fix(titus): Handle/report incomplete target group cache data

robzienert:
The NPE was caused by traversing a map-of-maps without actually being certain the data is there. Seems like the actual usage here is reasonable, but the cache data was corrupted / inconsistent in Redis. In this case, we'll just not report health data.

I took this as an opportunity to clean up the class a little as well, but the changes are in separate commits.

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusTargetGroupServerGroupProvider.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusTargetGroupServerGroupProvider.java
@@ -16,6 +16,8 @@
 
 package com.netflix.spinnaker.clouddriver.titus.caching.providers;
 
+import com.netflix.spectator.api.Id;
+import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.cats.cache.Cache;
 import com.netflix.spinnaker.cats.cache.CacheData;
 import com.netflix.spinnaker.clouddriver.aws.model.AmazonTargetGroup;
@@ -38,14 +40,22 @@ import static com.netflix.spinnaker.clouddriver.core.provider.agent.Namespace.*;
 public class TitusTargetGroupServerGroupProvider implements TargetGroupServerGroupProvider {
 
   private final Cache cacheView;
-
-  @Autowired
-  public TitusTargetGroupServerGroupProvider(Cache cacheView) {
-    this.cacheView = cacheView;
-  }
-
-  @Autowired
   AwsLookupUtil awsLookupUtil;
+  private final Registry registry;
+
+  private final Id inconsistentCacheId;
+
+  @Autowired
+  public TitusTargetGroupServerGroupProvider(Cache cacheView,
+                                             AwsLookupUtil awsLookupUtil,
+                                             Registry registry) {
+    this.cacheView = cacheView;
+    this.awsLookupUtil = awsLookupUtil;
+    this.registry = registry;
+
+    inconsistentCacheId = registry.createId("cache.inconsistentData")
+      .withTag("location", TitusTargetGroupServerGroupProvider.class.getSimpleName());
+  }
 
   @Override
   public Map<String, AmazonTargetGroup> getServerGroups(String applicationName,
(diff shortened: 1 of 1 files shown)


jrsquared:
🥈

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusTargetGroupServerGroupProvider.java
```diff
@@ -133,4 +125,75 @@ public TitusTargetGroupServerGroupProvider(Cache cacheView) {
...
+
+  private static Map<String, Object> getTargetGroupHealth(String instanceKey,
+                                                   Map<String, String> targetGroupDetails,
+                                                   CacheData healthData) {
+    try {
+      if (healthDataContainsTargetGroups(healthData)) {
+        Map targetGroupHealth = getTargetGroupHealthData(targetGroupDetails, healthData);
+
+        if (!targetGroupHealth.isEmpty()) {
+          Map<String, Object> health = new HashMap<>();
```
robzienert: Also a small bit of efficiency here. Doing a static initialization of a hashmap `e.g. new HashMap<>() { /* put commands */ };` creates a temporary class and doesn't buy us much terseness anyhow.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusTargetGroupServerGroupProvider.java
```diff
@@ -48,77 +49,68 @@ public TitusTargetGroupServerGroupProvider(Cache cacheView) {
...
-                  }
-                }
-              } catch (Exception e) {
-                log.error("failed to load health for " + instanceKey, e);
+
+              Optional<LoadBalancerInstance> instance = getInstanceHealth(instanceKey, instanceDetails, targetGroupDetails);
+              if (instance.isPresent()) {
+                targetGroupInstances.add(instance.get());
+              } else {
+                log.error(format(
```
ajordens: don't really need `String.format()` ... `"Detected potentially .. (targetGroup: {}, serverGroup: {})"` should suffice?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusTargetGroupServerGroupProvider.java
```diff
@@ -133,4 +125,75 @@ public TitusTargetGroupServerGroupProvider(Cache cacheView) {
...
+    } catch (NullPointerException e) {
+      return Optional.empty();
+    }
+
+    CacheData healthData = cacheView.get(HEALTH.ns, healthKey);
+
+    Map<String, Object> health = getTargetGroupHealth(instanceKey, targetGroupDetails, healthData);
+
+    return Optional.of(new LoadBalancerInstance(
+      ((Map) instanceDetails.get("task")).get("id").toString(),
```
ajordens: I guess it's safe to say an instance will always have a task will always have an id ... we weren't guarding against this previously and I can't think of how this wouldn't hold true

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusTargetGroupServerGroupProvider.java
ajordens: thinking out loud --- I wonder where else we could start looking for a signal that caches are potentially inconsistent in some way and emit a metric (short of a user telling us)

robzienert: This is a potential problem anywhere we're resolving relationship data between two types, right? If that assertion is correct, I'd tend to think the problem could pop up in quite a number of places, like in EC2 load balancer health checks.

ajordens: Yes, I believe that's true.

Something to keep on the back burner, it'd be nice to more readily detect when things are consistently inconsistent.

PR #2794, opened 2018-07-18
fix(provider/kubernetes): Fix broken dockerfile

ezimanyi:


ezimanyi:
@anshrma @sbwsg : FYI, https://github.com/spinnaker/clouddriver/pull/2791 is breaking the container builds as `curl` is not installed in the base image for `Dockerfile.slim`

sbwsg:
Doh!  I completely missed that.  Thanks @ezimanyi, retroactive lgtm

PR #2804, opened 2018-07-24
fix(provider/kubernetes): v2 check that artifact & cluster account match 

lwander:
cherry pick (#2799) into 1.8

When two artifacts with the same name & namespace are deployed to
different clusters, Spinnaker previously couldn't tell them apart when
deciding which to match & replace in a deployed manifest


lwander:
Integration tests passed on #2799

PR #2805, opened 2018-07-25
fix(google): Wait for firewall operations to complete

ezimanyi:
Currently inserting or deleting a firewall returns success as soon as the asynchronous operation is launched; clouddriver should instead wait for completion of the asynchronous operation before
returning success.


lwander:
I think this fix should ideally go in Orca, which is more tolerant of long running operations

jtk54:
We should also wait for the operation to complete in Clouddriver, right now we're just forking work in GCP and returning from Orca' view. I don't think there is a mechanism in Orca to wait on cloud provider ops without waiting on Clouddriver.

ezimanyi:
Made https://github.com/spinnaker/spinnaker/issues/3087 to come up with a mechanism for Orca to handle waiting for asynchronous operations, so we can move this logic to Orca.

PR #2806, opened 2018-07-25
fix(titus): Handle empty tasks when constructing `TitusJobStatus`

ajordens:


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/model/TitusJobStatus.groovy
```diff
@@ -55,23 +55,33 @@ class TitusJobStatus implements com.netflix.spinnaker.clouddriver.model.JobStatu
...
   }
 
-  JobState convertTaskStateToJobState(Job job, Task task) {
+  static JobState convertTaskStateToJobState(Job job, Task task) {
 
     if (job.getJobState() in ["Accepted", "KillInitiated"]) {
       return JobState.Running
+    } else if (task == null) {
+      // unexpected to have no task _and_ a job in an unexpected state
+      return JobState.Running
```
ajordens: not sure if this is the reasonable default or if it should just be `JobState.Failed` or something?

emjburns: I'll work with Corin to make sure this is the right state, but it looks reasonable to me.

PR #2811, opened 2018-07-26
fix(google): Properly handle failures caching HTTPS load balancers

ezimanyi:
When caching resources used by HTTPS load balancers, we don't pass along the list of failed load balancers, to a failure is not recorded and we can end up storing an incomplete load balancer to the cache.


ezimanyi:
I suspect this is contributing to this bug: https://github.com/spinnaker/spinnaker/issues/2780

In addition to passing along the `failedLoadBalancers` parameter, I also upgraded the severity of the warning when a global forwarding rule maps to a non-HTTP(S) back-end, as this is more than just informational---it's causing us to generate a partly-formed load balancer.

ezimanyi:
Wrong link...I suspect it's contributing to: https://github.com/spinnaker/spinnaker/issues/3003

PR #2812, opened 2018-07-26
feat(provider/kubernetes): Adds support for strategy.spinnaker.io/use…

gojenkuser:
… SourceCapacity annotation.

Adds support for strategy.spinnaker.io/useSourceCapacity annotation in manifests of type CanScale to use the deployed manifest replicas value. Allows teams to scale resources without having the cluster inadvertently scale during the next deployment.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesSourceCapacity.java
```diff
@@ -0,0 +1,40 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import java.lang.Double;
+
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.KubernetesManifest;
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.security.KubernetesV2Credentials;
+
+@Data
```
lwander: These three annotations aren't needed here -- they're only for classes you want to pass around as data, and this class only has one static method

lwander:
Nice work!

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesSourceCapacity.java
lwander: This also seems to have broken the build: https://travis-ci.org/spinnaker/clouddriver/builds/408619503#L750

PR #2814, opened 2018-07-26
chore(Dockerfile): Update python version to match alpinelinux. (#2770)

ezimanyi:


ezimanyi:
Cherry-pick https://github.com/spinnaker/clouddriver/pull/2770

PR #2817, opened 2018-07-27
fix(provider/docker): Clear docker token cache after 401

brantburnett:
If a 401 is received from the Docker registry, the cached token is
cleared so it isn't reused if a new token isn't acquired. Subsequent
requests will then get a new token, if necessary.

fixes spinnaker/spinnaker#2039


lwander:
Cherry picking into master (or usually from master into the release branch) is what we do.

PR #2818, opened 2018-07-27
fix(provider/docker): Clear docker token cache after 401

brantburnett:
If a 401 is received from the Docker registry, the cached token is
cleared so it isn't reused if a new token isn't acquired. Subsequent
requests will then get a new token, if necessary.

fixes spinnaker/spinnaker#2039


brantburnett:
@lwander looks like this fix didn't make it into 1.9?  Just want to make sure that was intentional.

lwander:
The fix you cherry picked in will be in the next build of 1.9

PR #2819, opened 2018-07-27
feat(builds): Introduces ability to only build specific cloud provide…

ttomsu:
…r platforms.

Some providers are more granular than others. This would allow a developer/release engineer to specify `./gradlew -PincludePlatforms=GCE,AWS,KUBERNETES build` with any permutation of platform dependencies. 

However, this doesn't quite work yet, due to entanglement of the clouddriver-security project and clouddriver-web previously being the centerpoint of how all providers come together. That work will come as a followup PR.


ewiseblatt:
@cfieber do you have thoughts?

settings.gradle
```diff
@@ -16,30 +16,93 @@
...
-  'clouddriver-oracle',
-  'clouddriver-titus',
-  'cats:cats-core',
-  'cats:cats-redis',
-  'cats:cats-dynomite',
-  'cats:cats-test'
+enum Platform {
+  ALL,
+  APPENGINE,
+  AWS, // Any thoughts on renaming this provider to EC2?
```
lwander: Renaming to EC2 in this context might make sense

ttomsu:
tag: https://github.com/spinnaker/spinnaker/issues/3114

settings.gradle
```diff
@@ -16,30 +16,93 @@
...
+    case Platform.ORACLE:
+      include 'clouddriver-oracle'
+      break
+    case Platform.TITUS:
+      includePlatform(Platform.AWS)
+      include 'clouddriver-titus'
+      break
+    case Platform.ALL: // fallthrough
+    default:
+      includePlatform(Platform.APPENGINE)
```
duftler: No way to do this without listing them all again?

duftler: Could do something like:

```
(Platform.values() - Platform.All).each { includePlatform(it) }​
```

The 2 lists are not short and are far apart. Would be easy for somebody to change one and neglect to change the other. The diff view would probably not even render both lists for the PR reviewer.

PR #2821, opened 2018-07-27
fix(perf): reverts DefaultProviderCache metric changes

cfieber:
Revert "fix(perf): fix performance regression due to default provider cache metrics causing extra redis reads. (#2816)"
This reverts commit 45642482c98d670df82b9853c63a9222f5aa5a65.

Revert "feat(cache): metrics on putCacheResult. (#2790)"
This reverts commit e78a1aebc28d3afadf358e7e69ed739c299e88dd.


cfieber:
was not happy with the extra complexity in that block of code after all was said and done

PR #2822, opened 2018-07-28
feat(aws): Adding a delete ami operation

jeyrschabu:
- added operation for deleting an AMI in account/region


jeyrschabu:
@lwander Yes it is for automating deletions with swabbie (https://github.com/spinnaker/swabbie).

PR #2820, opened 2018-07-27
fix(artifacts): fixes multiple http base providers

ethanfrogers:
when enabling multiple http based artifact providers (http + gitlab +
github) clouddriver filed to start stating that multiple beans were
found.

PR #2824, opened 2018-07-28
fix(artifacts): fixes multiple http base providers - 1.8

ethanfrogers:
when enabling multiple http based artifact providers (http + gitlab +
github) clouddriver filed to start stating that multiple beans were
found.


ethanfrogers:
@lwander same as #2820 but for `release-1.8.x`.

PR #2827, opened 2018-07-31
 fix(google): Guard against null backend in disabled load balancer check

ezimanyi:


ezimanyi:
* As TCP and Internal load balancers are required to have a backend, we should mark load balancers as failed if we can't find the back-end they referenced (likely due to a race condition where the back-end was deleted/changed between when we fetched the load balancer and when we looked up the back-ends).  This is how HTTP and HTTPS load balancers behave (the change to HTTPS having been recently done in https://github.com/spinnaker/clouddriver/pull/2811).
* We currently guard against a null back-end in all of the `determine{type}LoadBalancerDisabledState` functions except for HTTP.  Add in a null guard, just in case we still end up with a malformed load balancer, even after the caching changes.

These changes are to fix the intermittent issues seen in https://github.com/spinnaker/spinnaker/issues/3003

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleInternalLoadBalancerCachingAgent.groovy
```diff
@@ -182,7 +182,13 @@ class GoogleInternalLoadBalancerCachingAgent extends AbstractGoogleLoadBalancerC
 
       def backendServiceName = Utils.getLocalName(forwardingRule.backendService)
       BackendService backendService = projectRegionBackendServices?.find { BackendService bs -> bs.getName() == backendServiceName }
-      handleBackendService(backendService, newLoadBalancer, projectHttpHealthChecks, projectHttpsHealthChecks, projectHealthChecks, groupHealthRequest)
+      if (backendService == null) {
+        log.warn("Failed to read a component of subject ${newLoadBalancer.name}. Could not find BackendService ${backendServiceName}.\n"
```
lwander: Are there legitimate uses of GCP that can cause this?

lwander: In that case I would include in the warning that this is simply in a format not expected/deployed by Spinnaker

jtk54: SGTM.

ezimanyi: We only call `cacheRemainderOfLoadBalancerResourceGraph` if the fowarding rule has a non-null backend service (https://github.com/spinnaker/clouddriver/blob/52d79d1a01e8fb9cbd6c65eb55f82b955f89477b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleInternalLoadBalancerCachingAgent.groovy#L161) so the condition `backendService == null` means that the load balancer defined a back-end service but we couldn't find it.

For TCP load balancers, we're looking up a back-end on a Target TCP Proxy, which is required to have a back-end---so again, a failure to get the back-end means that it existed but we couldn't find it.

ezimanyi:
Made one small change after disucssing with @jtk54, which is to re-lower the level of a log statement in the HTTP caching agent, which I had raised in https://github.com/spinnaker/clouddriver/pull/2811, but which we should leave at `log` because it's possible to create valid infrastructure outside of Spinnaker that will trigger this warning, and we don't want to over-warn.

PR #2829, opened 2018-08-01
fix(ecs): specifiying resizeServerGroup collides with AWS

kevinawoo:
I'm testing this right now, let's not merge this! 🙏


bruno-carrier-lookout:
I'd think that removing that @Component annotation would have undesirable side effects.  If you believe this annotation collides with other components, then renaming it to something longer would probably be less risky

kevinawoo:
The annotation is found by [AnnotationsBasedAtomicOperationsRegistry.groovy#L39](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AnnotationsBasedAtomicOperationsRegistry.groovy#L37-L52), which is a fallback for legacy naming convention. It should instead look by `cloudprovider + description` combo further down.

One open question is if ECS provides the correctly named bean or not. This definitely screams unit tests lol.
Another question is what `description` gets passed in when doing an operation on a ECS cluster, and making sure all the names match. I believe they get passed in from deck.

bruno-carrier-lookout:
@kevinawoo perhaps this could be closed?

PR #2825, opened 2018-07-30
fix(provider/kubernetes): upgrade to latest version of spinnaker-dependencies to fix spinnaker/spinnaker#3082 (#2807)

lwander:

PR #2832, opened 2018-08-02
fix(cats): Remove unsupported annotations

wjoel:
Newer versions of spinnaker-dependencies do not include FindBugs.

clouddriver fails to build since upgrading spinnaker-dependencies to 0.158.0: https://github.com/spinnaker/clouddriver/pull/2732

https://github.com/spinnaker/clouddriver/pull/2825 fixed this for the 1.8 branch, and the discussion suggests dropping the annotations, so that's what this commit does.


lwander:
Weird how the PR to bump the dependencies didn't break

wjoel:
Yep. I'm surprised no one noticed this sooner, except for #2825.

PR #2857, opened 2018-08-10
fix(provider/kubernetes): speed up stratifying cache elements

lwander:
Cuts down cache time by ~1% -- not a huge improvement but was discovered
making other improvements


lwander:
Switching to streams was all that's required -- although I found out later this method wasn't very expensive to begin with.

PR #2862, opened 2018-08-10
fix(provider/kubernetes): only record app relationships for clusters

lwander:
Another effort at reducing the number of cache objects being deduped and
stored in memory during cache cycles.


ezimanyi:
@lwander : I think this commit (or one of the other ones this afternoon) broke our integration tests:
image

The kube_v2_cache_test is now failing consistently across all providers.

ezimanyi:
Oh, I think this is because a `SERVICE` `KubernetesKind` has `hasClusterRelationship` set to `false`; if I'm understanding that flag properly, it should probably be `true`.

PR #2865, opened 2018-08-10
fix(provider/kubernetes): Services have a cluster relationship

ezimanyi:


ezimanyi:
I think this is the correct way to fix the failing CI tests I noted in a comment on https://github.com/spinnaker/clouddriver/pull/2862.

lwander:
Nice! Thanks :)

PR #2869, opened 2018-08-13
fix(provider/kubernetes): make request timeout configurable

lwander:
Prevents caching agents from stacking up while failing to ever complete
when indexing unavailable clusters. Will need to be set by a
user/operator.

FYI @benjaminws


benjaminws:
This is awesome! Is there a default defined somewhere? Didn't see it in the diff.

lwander:
No I'm afraid to set one that will break existing users. Clouddriver property is per-account:

`kubectlRequestTimeoutSeconds: `

benjaminws:
@lwander makes sense, thanks for this!

PR #2871, opened 2018-08-13
feat(artifacts): expose types supported by a given artifact credential

sbwsg:
I initially planned to return only a single artifact type for a given credential but I think there are realistic potential use cases for a credential supporting multiple types: a github or gitlab commit object, an embedded artifact that isn't base64-encoded, gcs or s3 object metadata.


clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/config/ArtifactCredentials.java
```diff
@@ -21,9 +21,14 @@
...
+import java.util.List;
 
 public interface ArtifactCredentials {
   String getName();
-  boolean handlesType(String type);
+  List<String> getTypes();
   InputStream download(Artifact artifact) throws IOException;
+
+  default boolean handlesType(String type) {
+    return this.getTypes().stream().anyMatch(it -> it.equals(type));
```
lwander: You can omit `this` here

PR #2872, opened 2018-08-14
feat(provider/docker): adding passwordCommand option for retrieving d…

ethanfrogers:
…ocker credentials (#2837)

Allows users to use passwordCommand instead of password or passwordFile if their setup isn't conducive to the latter two options. This refreshes the credentials anytime the docker creds are used (every time it caches images), so an agent wasn't needed to re-initialize credentials constantly.


ethanfrogers:
cherry pick #2872 into `release-1.9.x`. @lwander @edwinavalos PTAL

ethanfrogers:
boo @spinnakerbot :(

lwander:
Features in release branches break our release process -- it's hard to rationalize this one as a fix for a bug

ethanfrogers:
yeah, you're right. i was hoping we might be able to get it in since `1.9` isn't out yet. but if you guys are trying to get 1.9 stabilized, we can close this.

lwander:
Sorry we have to close this, we're hoping to release 1.9 this afternoon -- we can try and find ways to make the process more tolerant of features in release branches in the future.

PR #2838, opened 2018-08-06
feat(provider/cf): add provider skeleton with credentials management

jkschneider:
Based on a conversation with @duftler, we're splitting up the Cloud Foundry provider submission into individual PRs that can be reviewed in smaller chunks.

This initial PR contains the skeleton of the provider, up to and including credentials.

```yml
cloudfoundry:
  enabled: true
  accounts:
    - name: pivotalwebservices
      user: MYUSER
      password: MYPASSWORD
      api: api.run.pivotal.io
```

cc / @ethanfrogers

PR #2875, opened 2018-08-14
feat(provider/cf): add model and cache keys

jkschneider:
Adds Cloud Foundry implementation of Spinnaker model objects.

Previous PRs in this sequence: #2838, #2828.

### DIFF
--- a/clouddriver-cloudfoundry/clouddriver-cloudfoundry.gradle
+++ b/clouddriver-cloudfoundry/clouddriver-cloudfoundry.gradle
@@ -1,3 +1,7 @@
+test {
+  useJUnitPlatform()
+}
+
 dependencies {
   compile project(":clouddriver-artifacts")
   compile project(":clouddriver-core")
@@ -10,7 +14,9 @@ dependencies {
   compile spinnaker.dependency('lombok')
 
   spinnaker.group('retrofitDefault')
-  spinnaker.group('test')
+
+  testCompile('org.junit.jupiter:junit-jupiter-api:5.2.0')
+  testRuntime('org.junit.jupiter:junit-jupiter-engine:5.2.0')
 
   compile "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${spinnaker.version("jackson")}"
 }
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/cache/Keys.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/cache/Keys.java
@@ -1,0 +1,158 @@
+/*
+ * Copyright 2018 Pivotal, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.cloudfoundry.cache;
+
+import com.netflix.spinnaker.clouddriver.cache.KeyParser;
+import com.netflix.spinnaker.clouddriver.cloudfoundry.model.CloudFoundryLoadBalancer;
+import org.springframework.stereotype.Component;
+
+import javax.annotation.Nullable;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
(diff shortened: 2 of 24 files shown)


clouddriver-cloudfoundry/clouddriver-cloudfoundry.gradle
```diff
@@ -10,7 +14,9 @@ dependencies {
   compile spinnaker.dependency('lombok')
 
   spinnaker.group('retrofitDefault')
-  spinnaker.group('test')
+
+  testCompile('org.junit.jupiter:junit-jupiter-api:5.2.0')
```
jkschneider: Is there a group similar to `spinnaker.group('test')` that we can create for this?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/Cluster.java
```diff
@@ -69,7 +69,7 @@ default Moniker getMoniker() {
    */
   @Empty
   @JsonSerialize(nullsUsing = NullCollectionSerializer.class)
-  Set<ServerGroup> getServerGroups();
+  Set<? extends ServerGroup> getServerGroups();
```
jkschneider: Allows us to use Lombok `@Getter` on model implementations while holding a more specific implementation of `ServerGroup`, e.g. `CloudFoundryServerGroup`. Much less casting when using the model object later.

Due to Groovy's relaxed type system, this likely never came up before.

jkschneider:
cc / @ethanfrogers

clouddriver-cloudfoundry/clouddriver-cloudfoundry.gradle
lwander: Not that I know of, but a new one would be welcome: github.com/spinnaker/spinnaker-dependencies

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/cache/Keys.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.Map;
+
+import static com.netflix.spinnaker.clouddriver.cloudfoundry.CloudFoundryCloudProvider.ID;
+import static com.netflix.spinnaker.clouddriver.cloudfoundry.cache.Keys.Namespace.*;
+
+@Component("CloudFoundryInfraKeys")
+public class Keys implements KeyParser {
+
+  @Nullable
+  public static Map<String, String> parse(String key) {
```
lwander: I've found it's been helpful to keep this as `Optional` for later usage, but up to you style wise

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/model/CloudFoundryLoadBalancer.java
```diff
@@ -0,0 +1,95 @@
...
+  private final Set<CloudFoundryServerGroup> mappedApps;
+
+  @JsonProperty
+  public String getName() {
+    return host + "." + domain + "." + getName() + (port == null ? "" : "-" + port) + (path == null ? "" : "/" + path);
+  }
+
+  @Override
+  public Moniker getMoniker() {
+    return EMPTY_MONIKER;
```
lwander: I would recommend setting this correctly unless there is an explicit need for keeping this empty (or you plan to implement this later).

jkschneider: Yes, there is an explicit need to keep it empty. Cloud Foundry load balancers are inseparable from public routes in a way that a moniker doesn't describe well.

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/cache/Keys.java
jkschneider: Thanks. I'm comfortable with the way it is if you are.

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/model/CloudFoundryLoadBalancer.java
jkschneider: In other words, it isn't possible to separate a CF load balancer into cluster, stack, detail, etc.

lwander: What about the 'app' field?

jkschneider: There is also no way to infer an app from a CF load balancer.

PR #2835, opened 2018-08-06
feat(provider/ecs): Added the ECS server group events endpoint

bruno-carrier-lookout:
feat(provider/ecs): Added the ECS server group events endpoint

@robzienert @dkirillov 

image

PR #2876, opened 2018-08-15
feat(provider/ecs): Adding server group events endpoint

robzienert:
Re-creating #2835 since, apparently, we weren't able to merge it.


robzienert:
Aaaand... can't do anything with this one either. Will submit off my own fork with an updated base branch.

PR #2879, opened 2018-08-15
feat(provider/aws) : Adding suffix to user agent

anshrma:
feat(provider/aws) : Adding suffix to user agent. This will give necessary telemetry information.


anshrma:
@spinnaker/netflix-reviewers - Can you please review this ?

cfieber:
you can set the property `aws.client.addSpinnakerUserToUserAgent: true`
The AWS SDK user agent will include for each request:
`spinnaker-user/<the username> spinnaker-executionId/<the execution id>`
Where `<the username>` is the username of the user logged in to spinnaker that initiated the operation that triggered the SDK call (or `unknown`)
and `<the execution id>` is the ID of the task or pipeline execution that triggered the SDK call (or `unknown`)

Does that cover what you are looking for out of this PR?

anshrma:
Unfortunately, no. It will be a overkill and not a recommended approach for telemetry. Moreoever, adding a suffix as done in PR will not cause any performance overheads.

PR #2883, opened 2018-08-16
fix(core): Fix dependency in Dockerfile

ezimanyi:


ezimanyi:
Clouddriver is not building because python `python=2.7.15-r0` can no longer be installed on the base alpine image; we need to install `python=2.7.15-r1`.  Given that we're not pinning the base image, I'm not sure the value of pinning the specific python version.  Instead, let's just install any version of python 2 greater than 2.7.9 (which is the declared dependency of the Google Cloud SDK).

PR #2889, opened 2018-08-20
fix(provider/aws) : adding useragent suffix

stewchen:


anshrma:
@stewchen / @cfieber - Changes look correct to me. Can we please merge this ?

PR #2890, opened 2018-08-20
fix(core): Fix dependency in Dockerfile

mrampton:
This was changed in master and causing issues in 1.8.


mrampton:
@lwander Could you take a look at this?

mrampton:
Sorry about that. Fixed -- mind taking another glance?

PR #2891, opened 2018-08-21
chore(provider/kubernetes): helpful startup & runtime logging

lwander:


danielpeach:
Could you please add a task status update at the very end of the operation? Sometimes it looks like Clouddriver crashes while submitting the manifest, but it's hard to tell without logging.

lwander:
Done

PR #2901, opened 2018-08-23
feat(build): Allow selecting specific cloud providers

dibyom:
This PR adds a gradle property called `cloudProviders` that can be used to only select a subset of all cloud providers e.g. `./gradlew -PcloudProviders=GCE`. By default, all providers are built.

This flag won't be useful until the rest of https://github.com/spinnaker/spinnaker/issues/3114#issuecomment-415542479 is complete.


gradle.properties
```diff
@@ -3,3 +3,5 @@ org.gradle.parallel=true
 jackson.version=2.9.2
 
 includePlatforms=ALL
+
+cloudProviders=ALL
```
ttomsu: `includedPlatforms` and `cloudProviders` have the same intent - to only include specific modules when building the code. Would you be opposed to combining them to the same variable?

build.gradle
```diff
@@ -14,6 +14,21 @@
  * limitations under the License.
  */
 
+enum Platform {
```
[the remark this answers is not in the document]
ttomsu: Lets discuss the progress I've made - I'm currently stuck, so this code duplication may be what we have to run with.

gradle.properties
dibyom: Changed it to only `includCloudProviders`. Unfortunately, we have to include all projects till we can detangle the tests in clouddriver-web

dibyom:
Cloud providers can be specified using `./gradlew -PincludeCloudProviders=aws,gce`. Defaults to all providers. 
@ttomsu PTAL
@robzienert @ajordens @cfieber let me know if you have any reservations on this approach.

ajordens:
No concerns on my part, looks like a reasonable and positive step forward. Thanks!

clouddriver-web/clouddriver-web.gradle
```diff
@@ -53,6 +40,16 @@ dependencies {
   testCompile ("org.springframework.boot:spring-boot-starter-test")
   //this brings in the jetty GzipFilter which boot will autoconfigure
   runtime 'org.eclipse.jetty:jetty-servlets:9.2.11.v20150529'
+
+  //Some controller tests rely on models from aws and gcp
+  testCompile project(':clouddriver-aws')
```
ttomsu: Blech. Lets get these out of the dependency tree!

PR #2904, opened 2018-08-23
fix(provider/ecs): Handle null startedAt field for tasks

clareliguori:
ECS tasks can have a null startedAt field for a short period -- this handles that gracefully


tomaslin:
LGTM

PR #2906, opened 2018-08-24
fix(core): use snakeyaml's SafeConstructor

lwander:


clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/ArtifactDownloader.java
```diff
@@ -40,7 +41,7 @@
   public ArtifactDownloader(ArtifactCredentialsRepository artifactCredentialsRepository, ObjectMapper objectMapper) {
     this.artifactCredentialsRepository = artifactCredentialsRepository;
     this.objectMapper = objectMapper;
-    this.yamlParser = new Yaml();
+    this.yamlParser = new Yaml(new SafeConstructor());
```
duftler: Can't we have `Yaml` created elsewhere and autowired like the other args?

PR #2909, opened 2018-08-25
feat(provider/ecs): Add support for Fargate launch type

clareliguori:
Adds support for the Fargate launch type for ECS provider

Notes:
* Uses the Spinnaker-managed role for the ECS account as the ECS execution role (same as how it used for the ECS service role)
* Sets the task's hard limits for CPU/memory to the same as the CPU/memory setting of the single container
* Explicitly setting the launch type to 'EC2' is supported, as well as null launch type value


clareliguori:
Restart build

PR #2913, opened 2018-08-27
fix(entitytags): Don't set _type field on Elasticsearch documents

ezimanyi:
Storing multiple mapping types in the same index is deprecated in Elasticsearch 6 and will be removed in Elasticsearch 7. Add a config parameter indicating that only one mapping type per index is supported (defaulting to false, for backwards compatibility). If that parameter is set, all documents will their type set to the config parameter mappingTypeName, which defaults to Elasticsearch's recommended value of _doc.

### DIFF
--- a/clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/config/ElasticSearchConfigProperties.java
+++ b/clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/config/ElasticSearchConfigProperties.java
@@ -66,11 +66,19 @@ public class ElasticSearchConfigProperties {
     this.connectionTimeout = connectionTimeout;
   }
 
-  public void setSingleMappingType(boolean singleMappingType) { this.singleMappingType = singleMappingType; }
+  public void setSingleMappingType(boolean singleMappingType) {
+    this.singleMappingType = singleMappingType;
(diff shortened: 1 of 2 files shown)


clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/config/ElasticSearchConfigProperties.java
```diff
@@ -57,4 +65,12 @@ public int getConnectionTimeout() {
   public void setConnectionTimeout(int connectionTimeout) {
     this.connectionTimeout = connectionTimeout;
   }
+
+  public void setSingleMappingType(boolean singleMappingType) { this.singleMappingType = singleMappingType; }
```
lwander: Usually we break this into multiple lines

ezimanyi: Ha, I thought IntelliJ had just collapsed the functions; fixed

ajordens:
We're still running an old ancient version of ES ... but otherwise this looks good.

Suspect we'd run into it when we get around to upgrading to thanks for getting it in!

PR #2916, opened 2018-08-27
refactor(core): Move models used in clouddriver-web to clouddriver-core

dibyom:
This allows clouddriver-web to not have a compile time dependency on clouddriver-aws
Part of https://github.com/spinnaker/spinnaker/issues/3114#


dibyom:
After discussing this with @ttomsu , I have an alternate PR (#2918 ) that does the same thing -- remove cloud provider models from clouddriver-web. 
But instead of moving a bunch of models to the `clouddriver-core` module, it does the following:
* Moves `RoleController` to the `aws` module since it seems like an AWS/ECS only controller.
* Removes the other `aws` models being used in `ServerGroupController.groovy`. The only usages were two static final strings from 2 models.

@lwander / @jrsquared , any preferences ?

robzienert:
I prefer the #2918 route.

dibyom:
Thanks @robzienert . Closing in favor of #2918

PR #2921, opened 2018-08-28
fix(provider/ecs): Display VPC & security groups based on network config

clareliguori:
Since Fargate tasks don't have an EC2 instance, we need to display their VPC and security groups based on the service's network configuration.  Also applicable to ECS tasks on EC2 capacity, as their ENI security groups can be different from the underlying instance's security groups.


clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsServerClusterProviderSpec.groovy
```diff
@@ -110,9 +114,127 @@ class EcsServerClusterProviderSpec extends Specification {
...
+    when:
+    def retrievedCluster = provider.getCluster("myapp", creds.getName(), familyName)
+
+    then:
+    retrievedCluster == expectedCluster
+  }
+
+  def 'should produce an ecs cluster with VPC network configuration'() {
+    given:
+    def applicationName = 'myapp'
```
tomaslin: consider moving some of these duplicate fields outside the spec, using a `setupSpec()` block or combining the two tests with `where`. It tends to improve the readability of the test when they are more concise.

tomaslin:
Seems like the previous PR broke some of the tests.

tomaslin:
LGTM

PR #2922, opened 2018-08-28
refactor(build): Depend on aws-sdk-s3 instead of the entire sdk

dibyom:
For a modular cloud driver build, with say only the kubernetes provider, this reduces the number of JARs in `lib` from 326 to 185

Part of spinnaker/spinnaker#3114

### DIFF
--- a/clouddriver-artifacts/clouddriver-artifacts.gradle
+++ b/clouddriver-artifacts/clouddriver-artifacts.gradle
@@ -1,8 +1,6 @@
 dependencies {
   compile project(":clouddriver-core")
 
-  spinnaker.group('amazon')
-
   compile spinnaker.dependency("frigga")
   compile spinnaker.dependency("bootActuator")
   compile spinnaker.dependency("bootWeb")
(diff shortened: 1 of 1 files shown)


ezimanyi:
This is great---I think in general there are a lot of places we can split up the amazon sdk into just what we need.  (As an example, half of the size of front50 is because it pulls in clouddriver-aws, which pulls in the entire amazon sdk.)

dibyom:
spinnaker-dependencies does actually split up a few of them : https://github.com/spinnaker/spinnaker-dependencies/blob/master/src/spinnaker-dependencies.yml#L207

The whole sdk gets pulled if one specifies `spinnaker.dependency('aws')` or `spinnaker.group('aws')`

duftler:
I like the idea of defining `awsS3` as future maintainers may see it as a good option when looking for the right dependency.

dibyom:
We'll probably do this in Front50 as well. Added `awsS3` in https://github.com/spinnaker/spinnaker-dependencies/pull/226

PR #2923, opened 2018-08-28
feat(cats/search): Enable substring matching for instance search

alanmquach:
We currently do exact matching for instance search, but substring match for other facets. Enabling substring match for instance searches makes the experience consistent across all facets.

While substring match doesn't make a whole lot of sense for machine generated instance IDs (i.e. `i-{17charhex}`), we're also matching against other attributes like account, region, and server group where substring match would be useful. We've also received requests for this, so it also seems to be what a user would naturally expect (for search to be match substrings, across all facets).

While I initially thought that this would be a performance hit from switching from `Collection.contains()` to `String.contains`, profiling shows that this approach is actually consistently 50% faster, presumably due to not having to collect all the values of all entries.

We instead can simply do a substring match on the cache identifier, which seems to be guaranteed to contain at least all the strings we would want to match against anyways. And since the parsing and storing the data from the identifier seems to only be for search purposes, we also cut the memory cost in half by not having to store the same data twice.


christopherthielen:
lgtm - need to update the tests

ajordens:
Definitely a change in behavior with doing substring matches across each of the facets (thinking about a search for "prod" that would match any account containing "prod" vs. just the "prod" account)  ... maybe that's ok. 

Did the change to still call `provider.parseKey(key)` impact the performance gains?

alanmquach:
> Did the change to still call `provider.parseKey(key)` impact the performance gains?
Nope, it doesn't impact runtime search performance since it's only done on cache refreshes.

Cache refreshes will obviously be slower than when not parsing, but it doesn't seem correct to omit that call as it's the `SearchableProvider`'s way of telling us it's invalid. It should be no slower than baseline which does parsing and a bit of shuffling.

PR #2905, opened 2018-08-23
refactor(build): Move top level provider config to com.spinnaker.config

dibyom:
Any Spring Configuration under com.spinnaker.config will be scanned by
`clouddriver-web` removing the necessity to specify them directly using
an `Import`.

PR #2930, opened 2018-08-28
fix(provider/ecs): EcsAccountMapper depends on EcsCredentialsInitializer

clareliguori:
ECS subnet and security group selection is currently broken on master, because EcsAccountMapper is initialized with zero accounts.  EcsAccountMapper is Autowired with AccountCredentialsProvider already, but it also depends on EcsCredentialsInitializer to have populated the accounts in that bean.

I think that previously (maybe before change #2905?), EcsAccountMapper just happened to be always initialized after EcsCredentialsInitializer, so the ECS and AWS accounts were always already populated.  I'm seeing that EcsCredentialsInitializer is now always initialized after EcsAccountMapper, so the accounts list is empty at the time that EcsAccountMapper is initialized.  This PR makes the dependency explicit.

I'm open to suggestions if there's a better way in Spring to wire up this dependency relationship.

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsAccountMapper.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsAccountMapper.java
@@ -6,6 +6,7 @@ import com.netflix.spinnaker.clouddriver.ecs.security.NetflixECSCredentials;
 import com.netflix.spinnaker.clouddriver.security.AccountCredentials;
 import com.netflix.spinnaker.clouddriver.security.AccountCredentialsProvider;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.DependsOn;
 import org.springframework.stereotype.Component;
 
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


tomaslin:
@jkschneider do you have any suggestions on how to improve this kind of Spring dependency relations?

tomaslin:
LGTM. Merging this to unblock broken master and will accept PR if a better Spring solution is found

jkschneider:
@tomaslin I think it's just the design of the `CredentialsInitializerSynchronizable` in general. Notice that many implementations have to inject `ApplicationContext` to satisfy bean relationships. That's generally not a good sign ;).

In general, if `EcsAccountMapper` relies on `EcsCredentialsInitializer` then it would generally be a dependency somehow. The fact that it isn't a dependency I think is the smell.

Consider that in the AWS case, there is a cyclic dependency between `AmazonCredentialsInitializer` and `AwsProviderConfig` that barely doesn't manifest itself because `catsModule` is `null` on the first invocation of `synchronizeAmazonAccounts`.

- `AmazonCredentialsInitializer` depends on `AmazonProviderSynchronizerTypeWrapper` which comes from `AwsProviderConfig`.
- `AwsProviderConfig` depends on `NetflixAmazonCredentials` which come from `AmazonCredentialsInitializer`.

PR #2934, opened 2018-09-01
fix(provider/google): Fix caching of >500 forwarding rules

ezimanyi:


ezimanyi:
Currently the caching agents that account for pagination all manually implement the logic themselves; I've created a class `PaginatedRequest` that wraps a request and handles checking the page token on the result, and scheduling requests for subsequent pages if necessary.

I have a separate PR to refactor the existing pagination to use this new class, but am keeping it separate as we may want to cherry-pick this change so I'm keeping non-essential refactoring out of it.

ezimanyi:
Addresses https://github.com/spinnaker/spinnaker/issues/3281

PR #2935, opened 2018-09-04
 refactor(provider/google): Move non-batch requests to PaginatedRequest

ezimanyi:


ezimanyi:
The first commit just refactors a few caching agents that had implemented custom pagination logic to use the re-usable logic added in https://github.com/spinnaker/clouddriver/pull/2934.

The second commit updates the `PaginatedRequest` class to handle paginating non-batch requests, and updates a number of caching agents to support pagination in their non-batch list requests.  I haven't gone through and updated all instances of `list` requests in caching agents, as I wanted to get feedback on the change first---if this looks good I'll make another PR to wrap the remaining requests.

ezimanyi:
@ewiseblatt : Please take a look to make sure the change to `timeExecute` looks good from a metrics perspective.

ezimanyi:
Discussed offline with @ewiseblatt---the slightly cumbersome API is due to the fact that the sdk doesn't expose the pagination functions in a structured way.  (That is, the result of all paginated calls support `getNextPageToken` but there is no common interface they implement describing this functionality, so we need to explicitly describe this for each class.)

PR #2938, opened 2018-09-04
fix(projects): remove parallel cluster lookups

cfieber:
A project with a large number of applications can significantly load
the redis cache by requesting all application clusters in parallel.

This removes the Rx code that was parallelizing the lookups in favor
of a linear lookup through all the project's applications, and
additionally uses the RequestQueue when performing the cluster lookups
in case we are under load and the requests are timing out.


cfieber:
Don't merge this, still testing

cfieber:
in perf testing seeing one of our more problematic getClusters requests about 60% slower (~10s vs ~6s). This feels like an acceptable tradeoff vs hammering the redis pool with parallel requests so I'm good for this to go in.

PR #2947, opened 2018-09-07
fix(google): Fix bugs in pagination of caching

ezimanyi:


ezimanyi:
A few silly bugs from https://github.com/spinnaker/clouddriver/pull/2935:
* A couple of places we need to store `this` in a variable, as we're crating the `PaginatedRequest` in a closure that overrides `this`
* `getItems` for GCE returns `null` rather than an empty array when there are no items, so we need to guard against null before calling `addAll`

None of these bugs are in 1.9.x, only master

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleZonalServerGroupCachingAgent.groovy
```diff
@@ -134,14 +134,15 @@ class GoogleZonalServerGroupCachingAgent extends AbstractGoogleCachingAgent impl
...
+    GoogleZonalServerGroupCachingAgent cachingAgent = this
     List<String> zones = credentials.getZonesFromRegion(region)
     List<GoogleServerGroup> serverGroups = []
 
     BatchRequest igmRequest = buildBatchRequest()
     BatchRequest instanceGroupsRequest = buildBatchRequest()
     BatchRequest autoscalerRequest = buildBatchRequest()
 
-    List<InstanceTemplate> instanceTemplates = fetchInstanceTemplates(compute, project)
+    List<InstanceTemplate> instanceTemplates = fetchInstanceTemplates(this, compute, project)
```
ewiseblatt: I thin kyou meant cachingAgent

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleRegionalServerGroupCachingAgent.groovy
```diff
@@ -131,7 +131,7 @@ class GoogleRegionalServerGroupCachingAgent extends AbstractGoogleCachingAgent i
     BatchRequest instanceGroupsRequest = buildBatchRequest()
     BatchRequest autoscalerRequest = buildBatchRequest()
 
-    List<InstanceTemplate> instanceTemplates = GoogleZonalServerGroupCachingAgent.fetchInstanceTemplates(compute, project)
+    List<InstanceTemplate> instanceTemplates = GoogleZonalServerGroupCachingAgent.fetchInstanceTemplates(this, compute, project)
```
ewiseblatt: ok, but everywhere else you assigned cachingAgent = this then passed cachingAgent

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleZonalServerGroupCachingAgent.groovy
ezimanyi: Good catch; technically it doesn't matter here as we're not in a closure, but since I've assigned the variable anyway (for the later closure) it's clearer to use it.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleRegionalServerGroupCachingAgent.groovy
ezimanyi: I'll assign it to `cachingAgent` here too as it's clearer

PR #2948, opened 2018-09-07
fix(provider/ecs): fetch all images for ECR repos with >100 images

fcoelho:
ECR repositories may contain many images, but the `ListImages`
operation will retrieve at most 100 at a time [1]. For repositories with
more than 100 images this used to cause pipelines depending on that
repository to randomly fail when you start getting over the limit (as
the desired image would not be on the first page of results).

This commit builds upon the `ListImages` operation and the `nextToken`
value to fetch all images available in the repository. And only images
matching the desired tag/digest will be sent to `DescribeImages`.

Some tests were slightly modified to include an empty list instead of
`null` in the `imageIds` field as:

1. that's what the `ListImages` returns in case there are no images in the repository
2. the code now uses the results from the `ListImages` operation to filter the images instead of the results of `DescribeImages` as previously done

`[1]`: https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/ecr/model/ListImagesRequest.html#getMaxResults--


tomaslin:
@clareliguori @BrunoCarrier can either of you LGTM this PR?

PR #2956, opened 2018-09-12
fix(provider/kubernetes): Improved log output from KubernetesOperationException

e96wic:
The changes allow for better debugging of failed kubernetes requests. Besides the error message, more parameters like response code, reason, etc.


lwander:
Thanks @e96wic!

PR #2958, opened 2018-09-12
fix(core): Make project clusters caching optional

robzienert:
Turns out trying to cache all project clusters takes forever in larger deployments.
Rather than try to cache everything, this change makes caching opt-in project-by-project.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/ProjectClustersService.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/ProjectClustersService.java
@@ -1,0 +1,414 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.clouddriver.core;
+
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.frigga.Names;
+import com.netflix.spinnaker.clouddriver.core.services.Front50Service;
+import com.netflix.spinnaker.clouddriver.model.Cluster;
+import com.netflix.spinnaker.clouddriver.model.ClusterProvider;
+import com.netflix.spinnaker.clouddriver.model.ServerGroup;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+import javax.inject.Provider;
+import java.util.*;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
+public class ProjectClustersService {
+
+  private final static Logger log = LoggerFactory.getLogger(ProjectClustersService.class);
+
+  private final Front50Service front50Service;
+  private final ObjectMapper objectMapper;
+  private final Provider<List<ClusterProvider>> clusterProviders;
+
+  public ProjectClustersService(Front50Service front50Service,
+                                ObjectMapper objectMapper,
+                                Provider<List<ClusterProvider>> clusterProviders) {
+    this.front50Service = front50Service;
+    this.objectMapper = objectMapper.copy().disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+    this.clusterProviders = clusterProviders;
+  }
+
+  public Map<String, List<ClusterModel>> getProjectClusters(List<String> projectNames) {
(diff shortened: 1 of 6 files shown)


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ProjectController.groovy
```diff
@@ -28,29 +29,45 @@ import org.springframework.web.bind.annotation.RequestMapping
...
   }
 
   @RequestMapping(method= RequestMethod.GET, value = "/clusters")
-  List<ProjectClustersCachingAgent.ClusterModel> getClusters(@PathVariable String project) {
-    CacheData cacheData = cacheView.get(Namespace.PROJECT_CLUSTERS.ns, "v1")
-    if (cacheData == null) {
-      throw new NotFoundException("Projects not cached")
-    }
+  List<ClusterModel> getClusters(@PathVariable String project) {
+    if (projectClustersCachingAgentProperties.allowList.contains(project)) {
```
ajordens: not sure if there's a problem with possible case-sensitivity here but otherwise lgtm.

robzienert: Will change. Just spot-checked the front50 API and that's case insensitive. Was erring on the side of caution here, but looks like that's not necessary here.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/ProjectClustersService.java
```diff
@@ -0,0 +1,415 @@
...
+                                Provider<List<ClusterProvider>> clusterProviders) {
+    this.front50Service = front50Service;
+    this.objectMapper = objectMapper.copy().disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+    this.clusterProviders = clusterProviders;
+  }
+
+  public Map<String, List<ClusterModel>> getProjectClusters(List<String> projectNames) {
+    List<Map> projects = front50Service.searchForProjects(Collections.emptyMap(), 1000);
+
+    Map<String, List<ProjectClustersService.ClusterModel>> projectClusters = new ConcurrentHashMap<>();
```
dreynaud: `ConcurrentHashMap`? Why?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/ProjectClustersService.java
```diff
@@ -0,0 +1,415 @@
...
+    }
+
+    Project project;
+    try {
+      project = objectMapper.convertValue(projectData, Project.class);
+    } catch (IllegalArgumentException e) {
+      throw new MalformedProjectDataException("Could not marshal project to internal model: " + projectName, e);
+    }
+
+    return getProjectClusters(project);
```
dreynaud: nitpick:

```
    try {
      return objectMapper.convertValue(projectData, Project.class);
    } catch (IllegalArgumentException e) {
      throw new MalformedProjectDataException("Could not marshal project to internal model: " + projectName, e);
    }
```

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/ProjectClustersService.java
```diff
@@ -0,0 +1,415 @@
...
+      .filter(appCluster -> {
+        Names clusterNameParts = Names.parseName(appCluster.getName());
+        return appCluster.getAccountName().equals(projectCluster.account) &&
+          nameMatches(clusterNameParts.getStack(), projectCluster.stack) &&
+          nameMatches(clusterNameParts.getDetail(), projectCluster.detail);
+      })
+      .collect(Collectors.toSet());
+  }
+
+  private List<RetrievedClusters> retrieveApplication(String application) {
```
dreynaud: might just be me but I find the `retrieveApplication` name a bit confusing. If it's retrieving cluster for an application, `retrieveClusters(String application)` makes more sense to me

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/ProjectClustersService.java
robzienert: Mmm, that's a leftover - will make it a normal HashMap, thanks.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/ProjectClustersService.java
robzienert: This is all just copied from the old stuff / will be rewritten when Adam gets to it, but will change.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/ProjectClustersService.java
robzienert: Yet, I don't want a single project's malformed configuration to blow up the rest of the caching agent.

dreynaud: Wat? Isn't my piece of code equivalent to your piece of code?

robzienert: Uhh... yes. 🤦‍♂️

PR #2967, opened 2018-09-17
fix(provider/oracle): remove oci-java-sdk-full-shaded from classpath

shihchang:
Fix the clouddriver startup problem due to the oci-java-sdk-full-shaded in the classpath.


shihchang:
@markxnelson @guoyongzhang

PR #2972, opened 2018-09-19
fix(provider/kubernetes): sane namespace lookup

lwander:
The prior approach of always consulting the service account namespace
was causing confusion, especially when no service account was in use

@psalaberria002 fyi

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
@@ -124,24 +124,35 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
     return cachedDefaultNamespace;
   }
 
-  public String lookupDefaultNamespace() {
-    String namespace = defaultNamespace;
+  private Optional<String> serviceAccountNamespace() {
     try {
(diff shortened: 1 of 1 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -124,24 +124,35 @@ public String getDefaultNamespace() {
...
+    } catch (KubectlException e) {
+      log.debug("Failure looking up desired namespace", e);
+      return Optional.empty();
+    }
+  }
+
+  public String lookupDefaultNamespace() {
+    try {
+      if (serviceAccount) {
+        return serviceAccountNamespace().orElse(defaultNamespace);
```
wjoel: Does this mean that when deploying to Kubernetes with Halyard, the default will remain `spinnaker` if that's the namespace Halyard deployed to?

That's what we'd like to avoid.

lwander: Only if you've explicitly set the `--service-account` flag with halyard. This is not the case if you're using a service account token for auth, we check what's stored in the kubeconfig for that instead.

wjoel: Excellent, looks good to me then!

PR #2974, opened 2018-09-19
Update PULL_REQUEST_TEMPLATE.md

christopherthielen:
Put @spinnaker/reviewers in ticks so we don't get mentioned on every PR which has the PR template in it.


christopherthielen:
^ that's what I get for using the Github editor 👎

erikmunson:
game changer 🚢

robzienert:
Man, I don't know. Don't we like... want to have reviewers get pinged when a PR comes open? I'd rather see this rolled back and we make the templates project-specific with reviewers per domain if we're interested in reducing noise.

christopherthielen:
Sure, OK. 

I assumed `spinnaker/reviewers` was intended to be an explicit request for review (as opposed to when people forget to delete the PR template)

PR #2976, opened 2018-09-21
fix(google): Stop retrying if allowable error encountered.

ewiseblatt:
@ezimanyi

### DIFF
--- a/clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/deploy/GoogleCommonSafeRetry.groovy
+++ b/clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/deploy/GoogleCommonSafeRetry.groovy
@@ -28,7 +28,7 @@ abstract class GoogleCommonSafeRetry {
   public static class SafeRetryState {
     Object finalResult
     Exception lastSeenException
-    int tries = 1
+    int tries = 0
     boolean success = false
   }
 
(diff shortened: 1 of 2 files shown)


ewiseblatt:
yea, i looked at that originally but seemed like a complexity tradeoff on conditionalizing the messages (or changing them, but they seemed helpful) so I left it as it was. I can look at rewriting the messages if you prefer.

ezimanyi:
I'm fine with this as is, particularly given how much it would increase the scope of the change to refactor.

clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/deploy/GoogleCommonSafeRetry.groovy
```diff
@@ -83,46 +83,45 @@ abstract class GoogleCommonSafeRetry {
...
-          def tries = ++state.tries
+    while (state.tries < maxAttempts) {
+      def tries = ++state.tries
+      try {
+        if (tries == 1) {
+          task?.updateStatus phase, "Attempting $action of $resource..."
+        } else {
           // Sleep with exponential backoff based on the number of retries. Add retry jitter with Math.random() to
           // prevent clients syncing up and bursting at regular intervals. Don't wait longer than a minute.
           Long thisIntervalWait = TimeUnit.SECONDS.toMillis(Math.pow(retryIntervalBase, tries) as Integer)
```
ezimanyi: This looks pre-exisiting and not overly consequential, but it seems like there's an off-by-one error in the exponential backoff.  Because `tries` includes the initial try, the first wait interval will be `retryIntervalBase ^ 2` instead of `retryIntervalBase` as I'd expect for truncated exponential backoff.

PR #2978, opened 2018-09-24
feat(provider/kubernetes): allow custom resource scope config

lwander:
closes https://github.com/spinnaker/spinnaker/issues/3374


lwander:
https://github.com/spinnaker/spinnaker/issues/3374

PR #2982, opened 2018-09-24
fix(aws): Retries around security group tagging

ajordens:
Creation of a security group is eventually consistent yet we were
immediately tagging a newly created security group.

This PR introduces retries _and_ treats a failure to tag as non-fatal.


ajordens:
FYI @anshrma

PR #2983, opened 2018-09-25
feat(artifacts): Add support for helm/charts artifacts

eromanova:
issue: https://github.com/spinnaker/spinnaker/issues/3284

### DIFF
--- a/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/helm/HelmArtifactCredentials.java
+++ b/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/helm/HelmArtifactCredentials.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2018 Mirantis, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.clouddriver.artifacts.helm;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spinnaker.clouddriver.artifacts.config.ArtifactCredentials;
+import com.netflix.spinnaker.kork.artifacts.model.Artifact;
+import com.squareup.okhttp.OkHttpClient;
+import com.squareup.okhttp.Request;
+import com.squareup.okhttp.Request.Builder;
+import com.squareup.okhttp.Response;
+import lombok.Data;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.StringUtils;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Arrays;
+import java.util.List;
+
+@Slf4j
+@Data
+public class HelmArtifactCredentials implements ArtifactCredentials {
+  private final String name;
+  private final List<String> types = Arrays.asList("helm/chart");
+
+  @JsonIgnore
+  private final IndexParser indexParser;
+
+  @JsonIgnore
+  OkHttpClient okHttpClient;
+
+  @JsonIgnore
+  private final Builder requestBuilder;
+
+  @JsonIgnore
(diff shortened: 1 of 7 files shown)


eromanova:
Thanks for your reply, @lwander. Officially helm chart's distribution organised with repositories (not tarballs) where charts can be collected. I suppose this changes will improve UX of using helm in spinnaker.

lwander:
> I suppose this changes will improve UX of using helm in spinnaker.

I agree this is something we need to improve

Does helm's official distribution mechanism have a set credential & download scheme? As long as its split across either a blob store (GCS/S3), HTTP, and others, I'd prefer to rely on the existing artifact kinds and perhaps find another way to improve the UX. If on the other hand there is a way to download helm charts that's unique to helm, I agree we need a new artifact kind.

eromanova:
@lwander @ethanfrogers Yes, current changes looks like dublication of http provider. But this changes are just the small step to provide the main idea (having an ability to get a list of helm/charts from concrete repository and choose one of them). 
It could be realized this way: we can create some periodic task that downloads index.yaml file at fixed intervals, collects the list of urls somewhere. In pipeline configuration we could choose expected artifact with Helm type, which contains only 2 drop-down fields: name and version. Is would be necessary to parse index.yaml file. This logic is different from another artifact kinds and it requires creating new helm/chart type, I think. So, later this current changes could be developed into this new feature. Thats why I prefer not to improve http artifact kind.

lwander:
> It could be realized this way: we can create some periodic task that downloads index.yaml file at fixed intervals, collects the list of urls somewhere. In pipeline configuration we could choose expected artifact with Helm type, which contains only 2 drop-down fields: name and version. Is would be necessary to parse index.yaml file. This logic is different from another artifact kinds and it requires creating new helm/chart type, I think. So, later this current changes could be developed into this new feature. Thats why I prefer not to improve http artifact kind.

This is a good idea to create dropdowns for artifacts and prepopulate them. Before we merge this, let's agree on a design going forward. Could you write a design doc describing how this will work, what the API should be, etc... for artifacts in general? Then we can apply it to the helm artifact type and others.

eromanova:
@lwander @ethanfrogers I've written the design doc. Take a look, please. Any suggestions are welcome.
https://docs.google.com/document/d/14n4IGisxVo7oqP0quznjBoRpFxrIhqskN4qBc-iGaqs/edit?usp=sharing

lwander:
Can you grant comment access? Also we were hoping for an overview of how the smart dropdowns & other improvements apply to artifacts as a whole, not just this new artifact type.

eromanova:
@lwander Sorry, my fault. This logic is going to work only for helm artifact type because only helm repositories are characterized primarily by index.yaml file from which we get the list of all charts and put it into dropdowns. Afaik, in cases of another artifact types there are no similar way. Correct me if I'm wrong. So, it must be another logic of how to get a list of artifacts for each of another artifact types. I first suggest to provide helm artifacts dropdowns and then start working with another ones.

ethanfrogers:
We might want to take this over to the design doc but I think @lwander is saying we should explore how it might look for other providers so that we can have a common design pattern. While this logic may only work for Helm charts, other providers, like Github or S3, would benefit from a caching/smart dropdown feature since there are APIs we can call to fetch and cache their contents. The design doc should reflect a larger design considering how it might affect the whole, while using a Helm chart artifact as an example.

eromanova:
@ethanfrogers @lwander 
Guys, I've updated the design doc. Could you please take a look again?
https://docs.google.com/document/d/14n4IGisxVo7oqP0quznjBoRpFxrIhqskN4qBc-iGaqs/

lwander:
Looking very solid @eromanova, thank you for taking the time to generalize this. My last hope is that we can come up with some shared controller definitions (see my comment in the doc).

eromanova:
@lwander @ethanfrogers 
Are there any questions we haven't discussed yet?

eromanova:
@lwander @ethanfrogers 
Are there any questions we haven't discussed yet?

eromanova:
@lwander @ethanfrogers 
Are there any questions we haven't discussed yet?

ethanfrogers:
@eromanova LGTM! thanks for taking the time to write it up. i think this will be a great feature!

eromanova:
Please review @lwander . Thanks in advance.

eromanova:
Hi, @lwander, review please.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/helm/HelmArtifactCredentials.java
```diff
@@ -0,0 +1,127 @@
...
+    } else {
+      log.info("No credentials included with helm artifact account {}", account.getName());
+    }
+    requestBuilder = builder;
+  }
+
+  private String credentialsFromFile(String filename) {
+    try {
+      String credentials = FileUtils.readFileToString(new File(filename));
+      return credentials.replace("\n", "");
```
lwander: I would only strip leading/trailing whitespace or newlines just in case.

eromanova: @lwander credentialsFromFIle function has changed. Take a look again please. Thanks.

PR #2985, opened 2018-09-25
fix(provider/kubernetes): fix FCR for cluster-scoped resources w/ namespace

lwander:


lwander:
https://github.com/spinnaker/spinnaker/issues/3379

lwander:
https://github.com/spinnaker/spinnaker/issues/3379

PR #3000, opened 2018-09-28
feat(provider/kubernetes): make checking permissions optional on startup

lwander:
@benjaminws fyi


benjaminws:
Killer, thanks!!!

PR #3003, opened 2018-10-01
fix(provider/kubernetes): Scope custom resource kinds by API group

willgorman:
Partial fix for spinnaker/spinnaker#3387

I've confirmed this works with custom resources like `networkpolicy.crd.projectcalico.org` where the unscoped name would collide with a native resource kind as well as one where it wouldn't, like `prometheusrule.monitoring.coreos.com`  I'm  not sure if this is the best way to handle it.  I'd also considered applying scoping by group for all resource types, custom or not.  This just seemed like a less risky approach.

I've noticed there does seem to be an existing race condition depending on whether the kind is dynamically registered first by the caching agent or by a deploy manifest operation.  Currently if the caching agent gets it first then it is correctly created  with isRegistered set to false: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesUnregisteredCustomResourceCachingAgent.java#L62.  If the deploy manifest operation goes first, then isRegistered is set to true.  With this change, the deploy manifest operation going first would also add it with the unscoped name.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifest.java
```diff
@@ -51,7 +51,9 @@ public KubernetesManifest clone() {
 
   @JsonIgnore
   public KubernetesKind getKind() {
-    return KubernetesKind.fromString(getRequiredField(this, "kind"));
+    //using ApiVersion here allows a translation from a kind of NetworkPolicy in the manifest to something
+    //like  NetworkPolicy.crd.projectcalico.org for custom resources
+    return KubernetesKind.fromString(getRequiredField(this, "kind"), getApiVersion());
```
lwander: This will result in registering a lot of different kinds when multiple cluster versions are used. It also means kind filtering logic won't work unless the user knows which version to expect

willgorman: I'm not sure that's the case since it only gets the api group out of the version rather than making the full version part of the kind.  I could clarify that by getting the group from the version and passing that instead of the version.

Is there an example of a place where kind filtering logic happens?  I can check that to make sure that isn't affected by introducing this change from an unscoped kind in the manifest to a scoped kind in the cache.

willgorman: The api group is added to the `name` of the kind so equality works the same as it had before.  By passing the api group here we are able to construct the scoped name to use when [searching for the kind in the static list of known kinds](https://github.com/cerner/clouddriver/blob/4d8dacb1731f246825a140677943e59c14e266a3/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKind.java#L155) so that we don't end up creating a new kind when a manifest comes in with just `PrometheusRule` as the kind.  The assumption here is that `PrometheusRule.monitoring.coreos.com` has already been dynamically registered by the caching agent.

lwander:
The orphaned keys only exist for kinds that have a group set though, right? Which should be limited to CRDs discovered by Spinnaker? I'm not sure if we can get around orphaning some keys, especially in your case where you have two resources both with the same kind but different api groups mapped to the same key.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKind.java
```diff
@@ -141,10 +146,24 @@ public static KubernetesKind fromString(String name, boolean registered, boolean
...
+    final String apiGroup = apiVersion != null ? apiVersion.getApiGroup() : "";
+    final String scopedName = !apiGroup.isEmpty() ? name + "." + apiGroup : "";
+
+
     synchronized (values) {
       Optional<KubernetesKind> kindOptional = values.stream()
-          .filter(v -> v.name.equalsIgnoreCase(name) || (v.alias != null && v.alias.equalsIgnoreCase(name)))
-          .findAny();
+        .filter(v -> v.name.equalsIgnoreCase(scopedName) || v.name.equalsIgnoreCase(name) || (v.alias != null && v.alias.equalsIgnoreCase(name)))
+        // if there are matches on both scopedName and name, prefer scoped name
```
lwander: One more question: what happens in your case where you have two different kinds with name "networkPolicy"? Does the non-scoped one disappear?

willgorman:
It looks like this still needs some more work to allow custom resource kinds that are dynamically registered from deploy manifest operations (instead of the caching agent) to register with the scoped name. It wasn't the best approach in the first place but it doesn't work at all in our production spinnaker where we run some clouddriver instances just for serving read/write requests with the caching agents disable.

willgorman:
This still isn't ready to merge.  What I've got now seems to cover all the paths I've come across where KubernetesKinds can get dynamically registered by caching or deploying manifests.  However, the ManifestController lookups are broken and return 404s for CRD manifests.  It doesn't seem to affect deployments but it does prevent deck from showing the "stable", "available", etc badges in the stage.

I think I need to modify the kind in whatever data is getting sent back to deck so that it can pass the scoped name back to the manifest controller, but I'm not sure if that would cause any other issues.

willgorman:
@lwander I think this is ready for another look if you don't mind.  The outstanding issue from before was that calls from the UI to clouddriver that needed to find a manifest were only providing the unqualified name (e.g. `NetworkPolicy`).  The additional change I made was to return the fully qualified kind from the deploy manifest stage. That way when the UI queries clouddriver it provides that value and I've confirmed that fixes the stable/available badges and also works with the live manifest calls.   Initially it looks like
screen shot 2019-02-27 at 4 23 10 pm

And then updates to 

screen shot 2019-02-27 at 4 23 02 pm



 What do you think about this approach?  I think I'm a little uncomfortable with it because I don't know if there are other places where the UI having the fully qualified kind could break something else.  Or if there are other places in clouddriver that return manifests where a similar substitution would need to be made.

The other approach I can think of would be to leave it up to the UI/API consumer to construct the fully qualified kind by using the group name from the apiVersion: `http://$GATE/manifests/$CLUSTER/$ACCOUNT/NetworkPolicy.crd.projectcalico.org%20log-then-deny`  That might be preferable, I haven't really looked at deck yet to see how feasible that is.

lwander:
Hey @willgorman I'm no longer on this project -- but I think @ezimanyi can give this a proper look :)

willgorman:
No problem, thanks @ezimanyi.  Good question about the version, and no I hadn't thought about that yet.  I think that for the purposes of the caching agent it would be ok to have multiple versions of a type of resource being cached as the same kind.  In the issue this PR references, the collision on the unqualified kind between the core caching agent and the unregistered CRD caching agent prevented the CRD kinds from being retrieved from the cache.  I think that if there were multiple versions of a kind cached as the same thing they would still be found in the cache and the consumer of the cached data could use the version to decide how to treat the data.

As for the deck behavior, I think that makes sense.  The direct approach here seems likely to cause confusion at best and perhaps break something I haven't thought of.  I'll take that out and look at changing deck instead.

willgorman:
@ezimanyi Ok, I've made the changes discussed above.  One thing to consider about coordinating this change along with the change in deck is that this change alone will break the available/stable badges for all CRDs, not just ones that have an unqualified name that matches a built-in kind.  I haven't tested those other types of CRDs with live manifest calls yet but I'm assuming that it may also prevent those from working.  It might be good to hold off merging this until the deck change is in place, what do you think?

willgorman:
The corresponding deck PR is here: https://github.com/spinnaker/deck/pull/6691

willgorman:
I also tested with live manifest calls for several types of CRDs and those seem to be working correctly too.

ezimanyi:
Awesome---this week has been pretty busy so I haven't had a chance to take a final look over this, but should have some time tomorrow to review this and the deck change.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesApiGroup.java
```diff
@@ -0,0 +1,75 @@
+package com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest;
```
ezimanyi: Missing copyright header

ezimanyi:
Looking into an issue now where the "Patch Manifest" stage no longer works after this change (at least when patching a "Deployment").  We've cut the 1.13 branch, but will wait on either a fix or rollback of this---hopefully there's a reasonable fix.

willgorman:
Well that's no good. What kind of error does it cause? I'll try to take a look at this today but may not be able to get to it until tonight.

ezimanyi:
The error is:
```
c.n.s.c.o.DefaultOrchestrationProcessor  : com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.MalformedManifestException: Missing required field 'apiVersion' on manifest:
{metadata={namespace=docs-site, name=spinnaker-io-nginx}, kind=deployment}
        at com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.MalformedManifestException.missingField(MalformedManifestException.java:27)
        at com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.KubernetesManifest.getRequiredField(KubernetesManifest.java:45)
        at com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.KubernetesManifest.getApiVersion(KubernetesManifest.java:65)
        at com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.KubernetesManifest.getKind(KubernetesManifest.java:55)
        at com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.KubernetesManifest.getFullResourceName(KubernetesManifest.java:284)
        at com.netflix.spinnaker.clouddriver.kubernetes.v2.op.OperationResult.addManifest(OperationResult.java:52)
```

I believe the issue is just that we're getting the `apiVersion` from the manifest using `getRequiredField`, but in some cases (ex: patch manifest) we don't have a fully formed manifest at that point, only the kind field.  We probably just need to guard that case and just pass in `null` as the kind there so it's inferred.

I'll see if that small change fixes it, and if so will make that change.

willgorman:
Yep, I'd just noticed the same thing and I think that sounds like it should work.  

I'm not sure yet what it means for patching CRD manifests since those do require an apiVersion now, but I'm not sure how well supported that was previously since it looks like the list of kinds in the UI doesn't include CRDs.

PR #3007, opened 2018-10-02
feat(provider/kubernetes) k8s resource whitelisting using annotation …

mnalsup:
…moniker.spinnaker.io/application

Added a property called "onlySpinnakerManaged" to the kubernetesV2 provider that whitelists applications that have the spinnaker application moniker annotation. The property is passed from the KubwernetesV2Credentials to the KubernetesManifestAnnotator to affect how it determines whether a resource is ignored.

Use:
Add the onlySpinnakerManaged property to the clouddriver profile under a kubernetes provider set to true. Spinnaker will ignore resources without the application moniker annotation.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestAnnotater.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestAnnotater.java
@@ -208,12 +208,13 @@ public class KubernetesManifestAnnotater {
         .build();
   }
 
-  public static KubernetesCachingProperties getCachingProperties(KubernetesManifest manifest) {
+  public static KubernetesCachingProperties getCachingProperties(KubernetesManifest manifest, boolean onlySpinnakerManaged) {
     Map<String, String> annotations = manifest.getAnnotations();
-
+    boolean shouldIgnore = getAnnotation(annotations, IGNORE_CACHING, new TypeReference<Boolean>() {}, false) ||
+      (onlySpinnakerManaged  && getAnnotation(annotations, APPLICATION, new TypeReference<String>() {}, "").equals(""));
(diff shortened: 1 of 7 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestAnnotater.java
```diff
@@ -208,12 +208,13 @@ public static KubernetesManifestTraffic getTraffic(KubernetesManifest manifest)
         .build();
   }
 
-  public static KubernetesCachingProperties getCachingProperties(KubernetesManifest manifest) {
+  public static KubernetesCachingProperties getCachingProperties(KubernetesManifest manifest, boolean onlySpinnakerManaged) {
     Map<String, String> annotations = manifest.getAnnotations();
-
+    boolean shouldIgnore = getAnnotation(annotations, IGNORE_CACHING, new TypeReference<Boolean>() {}, false) ||
+      (onlySpinnakerManaged  && getAnnotation(annotations, APPLICATION, new TypeReference<String>() {}, "").equals(""));
```
lwander: Extra space before `&&`.

I don't think this belongs in the manifest annotator, I think we should read the properties set from the annotations here, and then make a decision about whether or not to cache this outside the annotator.

mnalsup: That makes sense. It did feel like a bit of a hack. I will look around and make sense of how that is decided.

I now realize that submitting a pull request from my master branch is not ideal. I would like to submit a new one once I fix the implementation.

lwander: Sounds good

PR #3015, opened 2018-10-03
feat(search): Adds parallell executor to search providers.

jtk54:
Previously the search providers ran sequentially. If one provider
took far too long, the upstream service would disconnect before
any results were returned from the other providers. This adds a
circuitbreaker to cap the seach processing time and parallelize
the providers so that one provider's failure won't affect other
providers' results. The new circuitbreaker is configurable; by
default the previous behavior is preserved.


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/search/circuitbreaker/SearchCircuitbreaker.java
```diff
@@ -0,0 +1,90 @@
+package com.netflix.spinnaker.clouddriver.search.circuitbreaker;
```
lwander: Missing license header

jtk54:
The naming here might be too close to Hystrix nomenclature so I will likely rename.

jtk54:
Added metrics to the search executor.

PR #3017, opened 2018-10-03
feat(provider/kubernetes): split kubernetes pod logs by container #3360

maggieneterval:
Closes [issue #3360](https://github.com/spinnaker/spinnaker/issues/3360)

Updates `{account}/{region}/{id}/console` endpoint to return list of container logs for Kubernetes v1 and v2 providers rather than a string concatenation of each container log.

cc @danielpeach @lwander

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/InstanceProvider.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/InstanceProvider.groovy
@@ -16,7 +16,7 @@
 
 package com.netflix.spinnaker.clouddriver.model
 
-interface InstanceProvider<T extends Instance> {
+interface InstanceProvider<T extends Instance, S> {
 
   /**
    * Returns the platform the instance provider
@@ -26,5 +26,5 @@ interface InstanceProvider<T extends Instance> {
 
   T getInstance(String account, String region, String id)
 
-  String getConsoleOutput(String account, String region, String id)
+  S getConsoleOutput(String account, String region, String id)
 }
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2InstanceProvider.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2InstanceProvider.java
@@ -28,6 +28,7 @@ import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.Kube
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.KubernetesManifest;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.op.job.KubectlJobExecutor;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.security.KubernetesV2Credentials;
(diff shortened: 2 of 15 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/InstanceProvider.groovy
```diff
@@ -26,5 +26,5 @@ interface InstanceProvider<T extends Instance> {
 
   T getInstance(String account, String region, String id)
 
-  String getConsoleOutput(String account, String region, String id)
+  S getConsoleOutput(String account, String region, String id)
```
danielpeach: I'd prefer if this interface had two different methods: `getConsoleOutput` and `getConsoleOutputs`. It would implement a `default` for `getConsoleOutputs` that would return an empty list.

The return type of `InstanceController.getConsoleOutput` would be a map with both `output` and `outputs`, which would be populated by `getConsoleOutput` and `getConsoleOutputs`, respectively.

That way, the return type of `InstanceController.getConsoleOutput` will be consistent, and there won't be any special-casing in the UI for Kubernetes. If there are `outputs`, render those, otherwise, render `output`.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2InstanceProvider.java
```diff
@@ -28,11 +28,14 @@
...
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.op.job.KubectlJobExecutor;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.security.KubernetesV2Credentials;
+import com.netflix.spinnaker.clouddriver.model.ContainerLog;
 import com.netflix.spinnaker.clouddriver.model.InstanceProvider;
 import com.netflix.spinnaker.clouddriver.security.AccountCredentialsRepository;
 import com.netflix.spinnaker.clouddriver.security.ProviderVersion;
 import io.kubernetes.client.models.V1Container;
 import io.kubernetes.client.models.V1Pod;
+import java.util.ArrayList;
+import java.util.List;
```
danielpeach: Sort `java` imports together, below the rest of the imports.

lwander: `ctrl-alt-O` does this for you in intellij FYI

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/InstanceProvider.groovy
lwander: From my perspective I like this current approach -- the caller should know what to expect, and we don't litter the interfaces with more fields.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2InstanceProvider.java
maggieneterval: Thanks, updated! Looks like my Intellij's `ctrl-alt-O` defaults to alphabetizing the imports, but I changed it in settings -> code style to sort by all other imports + newline + java imports.

PR #3018, opened 2018-10-03
feat(provider/cf): caching agent backwards compatibility to CAPI 2.94.0/3.29.0

jkschneider:
We originally intended to support PCF 2.0+, but this minor change allows us to support PCF 1.12 as well.


clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/model/v3/LinkTest.java
```diff
@@ -0,0 +1,14 @@
+package com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3;
```
tomaslin: missing license

jkschneider: done

tomaslin:
LGTM

PR #3019, opened 2018-10-04
feat(web): Allow for request queue size to be overridden dynamically

ajordens:
```
Most typically, core and maximum pool sizes are set only upon
construction, but they may also be changed dynamically using
setCorePoolSize(int) and setMaximumPoolSize(int).

If the new value is smaller than the current value, excess existing
threads will be terminated when they next become idle. If larger, new
threads will, if needed, be started to execute any queued tasks.
```

^^ from JDK documentation


ajordens:
This may/may not be valuable.

PR #3022, opened 2018-10-05
feat(provider/kubernetes): Add Builder to KubernetesImageDescription

ezimanyi:
Some existing code is using an auto-generated constructor for KubernetesImageDescription, which is fragile as added fields can change the meaning of the constructor. Add a Builder annotation so this code can use the more robust builder pattern.


ezimanyi:
@lwander : Part 1 of fixing https://github.com/spinnaker/spinnaker/issues/3427.  Once this is merged, I'll bump clouddriver in spinnaker-dependencies, then pull that in to Halyard and change the broken code to use the builder pattern.  (This feels much less fragile than just changing the code in Halyard to match the current order of the fields.)

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/description/servergroup/DeployKubernetesAtomicOperationDescription.groovy
```diff
@@ -82,6 +83,7 @@ class KubernetesContainerPort {
 
 @AutoClone
 @Canonical
+@Builder
```
lwander: This sometimes requires both `@AllArgsConstructor` and `@NoArgsConstructor` for serialization & groovy magic to keep working

ezimanyi: It looks like the `@Canonical` is enough here; I'm able to create objects with a no args constructor an all args constructor as it was before the `@Builder` was added.

PR #3023, opened 2018-10-05
Fix duplication instance health (NLB) [GCP]

gandikun:
Quote from one of commit message : 

```
fix(provider/google): Do not add duplicate instance healths (NLB)  …
Changes:
- Skip health status for a different address
- Only add one health resolution per load balancer, by using a set
- Revert the de-duplication workaround in bc4ed8f

A NLB setup involves a target pool, along with:
- X addresses
- Y forwarding rules ("load balancers") per address
- Z instances

There are X * Y load balancers. Before this change, within each load
balancer, there would be X * Z * Z instance healths, when it should
be just Z.

With 8 addresses, 3 forwarding rules per address, and 60 instances,
there would be 8 * 3 * 8 * 60 * 60 = 691200 instance healths. G1GC, on
a powerful machine with a 32GB heap, will not be able to keep up with
such load, and will repeatedly fallback to serial full GC, causing
long STW pause that lasts for many seconds.

The de-duplication masked the problem, but doesn't help the GC load.

Anyway, the API design for NLB is a little unintuitive, such that:
1. The health check is tied to the target pool.
2. We will make Z API calls to "targetPools.getHealth", once for each
   instance.
3. Each response will contain X health statuses for the instance, one
   for each address.
4. We will then populate the `tpNameToInstanceHealthsMap` from the
   responses.
5. To populate the instance healths for each load balancer, we will
   dip into the map X * Y times.

The duplication could happen, because:
- we didn't deal with the X health statuses correctly in step 3.
- we dipped into the map X * Y * Z times in step 5.
```

More details PR please take a look commit messages.

Thanks

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
@@ -1956,6 +1956,11 @@ class GCEUtil {
       def instanceName = Utils.getLocalName(status.instance)
       def googleLBHealthStatus = GoogleLoadBalancerHealth.PlatformStatus.valueOf(status.healthState)
 
+      if (googleLoadBalancer.type == GoogleLoadBalancerType.NETWORK && googleLoadBalancer.ipAddress != status.ipAddress) {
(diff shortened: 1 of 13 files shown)


jtk54:
There seem to be quite a few duplicated commits in this PR. Can you rebase from master or create a new branch and cherry-pick your actual change onto master?

gandikun:
Hi @jtk54 ,

Sorry for delay, here updated PR.

Thanks

cc : @sayap

PR #3024, opened 2018-10-05
feat(provider/cf): caching agent backwards compatible to CAPI 2.94.0/…

jmelchio:


jkschneider:
@danielpeach This commit could be conceptually seen as a "fix". Leaving commit message as is for consistency with the related commit in master?

PR #3027, opened 2018-10-09
fix(provider/kubernetes): Assign default sequences to manifests

oahayder:
https://github.com/spinnaker/spinnaker/issues/3123

@lwander 

Attempt to use the manifest name to populate the sequence of an RS server group.

Have been trying to add a test but not very familiar with the Name class, trying to mock a situation where we have a replicaset with no sequence field but a sequence in the parsed name 🤔


oahayder:
@lwander you're welcome! I think you need to be the one to push the button here.

PR #3028, opened 2018-10-11
fix(provider/amazon): support setting a http health check on a tcp target group

tomaslin:
HTTP health checks for TCP does not support custom matchers and timeouts. Also, unhealthy and healthy values need to be the same


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -122,9 +95,15 @@ class LoadBalancerV2UpsertHandler {
           .withTargetType(targetGroup.targetType)
 
         if (targetGroup.healthCheckProtocol in [ProtocolEnum.HTTP, ProtocolEnum.HTTPS]) {
-          createTargetGroupRequest.withMatcher(new Matcher().withHttpCode(targetGroup.healthCheckMatcher))
+          createTargetGroupRequest
             .withHealthCheckPath(targetGroup.healthCheckPath)
-            .withHealthCheckTimeoutSeconds(targetGroup.healthCheckTimeout)
+
+          if (targetGroup.protocol == ProtocolEnum.TCP) {
+            createTargetGroupRequest.withUnhealthyThresholdCount(createTargetGroupRequest.getHealthyThresholdCount())
```
jrsquared: this is a little confusing, maybe a comment to explain why this is necessary?

PR #3041, opened 2018-10-17
feat(provider/aws): Add accountId to subnet responses

gardleopard:
accountId added to the subnet object for aws and ecs.

```
{
  "account" : "default",
  "accountId" : "402251857394",
  "availabilityZone" : "eu-west-1c",
  "availableIpAddressCount" : 4091,
  "cidrBlock" : "172.31.16.0/20",
  "deprecated" : false,
  "id" : "subnet-0432b02b",
  "purpose" : null,
  "region" : "eu-west-1",
  "state" : "available",
  "target" : null,
  "type" : "aws",
  "vpcId" : "vpc-d023c5b4"
}
```


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonSubnetProvider.groovy
```diff
@@ -96,6 +106,7 @@ class AmazonSubnetProvider implements SubnetProvider<AmazonSubnet> {
         purpose = "${splits[1]} (${splits[0]})"
       }
     }
+    def accountId = accounts.find{ account -> account.name.equals(parts.account)}.accountId
```
gardleopard: Can we get a miss here? If one account is removed and the clouddriver cache has stale data here?

cfieber: yeah actually that could happen

one thought would be to rework this and stick the accountId into the cached data when it is stored? (would eliminate the need to resolve it on the way back out)

gardleopard: Reworked in https://github.com/spinnaker/clouddriver/pull/3101

gardleopard:
Closing this because https://github.com/spinnaker/clouddriver/pull/3101 is probably a better way to solve this.

PR #3045, opened 2018-10-17
fix(provider/kubernetes): fix enable patch when key doesn't exist

lwander:


lwander:

PR #3047, opened 2018-10-17
fix(kubernetes): add support for json patching #3418

maggieneterval:
Closes #3418

Previously, we assumed that every posted `patchBody` could be deserialized as a `KubernetesManifest` in order to re-use the existing artifact replacement logic. This change adds support for the JSON merge strategy by casting the `patchBody` based on selected merge strategy.

@lwander Hopefully I understood what you meant about skipping the artifact replacement stage for JSON patches -- let me know if there is anywhere outside the code I should document that.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesPatchManifestDescription.java
```diff
@@ -36,7 +37,7 @@
   // This will only be a portion of a full manifest so calls to some required fields can fail.
   // Using the KubernetesManifest type makes it simpler to reuse the ArtifactReplacement logic.
   // TODO: change Orca to only send a single manifest.
-  private KubernetesManifest patchBody;
+  private KubernetesPatchBody patchBody;
```
lwander: If someone already has a pipeline with a patch stage defined, editing this stage's description will break it. I would either: keep this field the same, and add a new one for the jsonpatch object, or, go with what you have and write a pipeline migrator. I think @jtk54 can help with that (it's not too much work).

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/KubernetesPatchManifestOperation.java
```diff
@@ -64,15 +66,26 @@ public OperationResult operate(List _unused) {
...
+    if (manifestPatch != null) {
+      updateStatus("Swapping out artifacts in " + objToPatch + " from context...");
+      ReplaceResult replaceResult = replaceArtifacts(objToPatch, patchHandler);
+
+      updateStatus("Submitting manifest " + description.getManifestName() + " to Kubernetes master...");
+      result.merge(patchHandler.patchWithManifest(credentials, objToPatch.getNamespace(),
+        objToPatch.getName(), description.getOptions(), replaceResult.getManifest()));
+      result.getBoundArtifacts().addAll(replaceResult.getBoundArtifacts());
+    } else if (jsonPatches != null) {
+      // Skip artifact replacement for json patches
```
lwander: :+1:

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/KubernetesPatchManifestOperation.java
```diff
@@ -64,15 +66,26 @@ public OperationResult operate(List _unused) {
...
-    updateStatus("Submitting manifest " + description.getManifestName() + " to Kubernetes master...");
     OperationResult result = new OperationResult();
-    result.merge(patchHandler.patch(credentials, objToPatch.getNamespace(), objToPatch.getName(),
-      description.getOptions(), replaceResult.getManifest()));
 
-    result.getBoundArtifacts().addAll(replaceResult.getBoundArtifacts());
+    KubernetesManifest manifestPatch = description.getPatchBody().getManifestPatch();
+    List<JsonPatch> jsonPatches = description.getPatchBody().getJsonPatches();
+
+    if (manifestPatch != null) {
```
lwander: I would recommend checking the patch strategy in the incoming payload instead of checking which of these fields was defined.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesPatchManifestDescription.java
lwander: Or, even easier, make this an `Object` or `Map`, and decide how to cast it (using an `ObjectMapper`) based on what the strategy is in the below `KubernetesPatchOptions` object.

maggieneterval: Good call, updated!

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/KubernetesPatchManifestOperation.java
maggieneterval: Makes sense, updated!

lwander:
Could you update the docs here? https://www.spinnaker.io/reference/pipeline/stages/#patch-manifest

maggieneterval:
> Could you update the docs here? https://www.spinnaker.io/reference/pipeline/stages/#patch-manifest

Yes! https://github.com/spinnaker/spinnaker.github.io/pull/1055/files

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/KubernetesPatchManifestOperation.java
```diff
@@ -56,6 +61,8 @@ private static Task getTask() {
     return TaskRepository.threadLocalTask.get();
   }
 
+  private static ObjectMapper objectMapper = new ObjectMapper();
```
danielpeach: Any reason to use a new `ObjectMapper` rather than autowire the pre-configured one?

For example, here:
https://github.com/spinnaker/clouddriver/blob/29c5c1435ab0c750c4b491208f600647a3e98901/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2ArtifactProvider.java#L40

maggieneterval: Updated!

PR #3058, opened 2018-10-19
feat(provider/kubernetes): dynamic target selection

lwander:


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2ManifestProvider.java
```diff
@@ -84,6 +86,35 @@ public KubernetesV2Manifest getManifest(String account, String location, String
     }
 
     CacheData data = dataOptional.get();
+
+    return fromCacheData(data, account);
+  }
+
+  @Override
+  public List<KubernetesV2Manifest> getClusterAndSortAscending(String account, String location, String kind, String app, String cluster, Sort sort) {
```
jtk54: Consider a wrapper class if we add any more arguments.

lwander: Agreed

PR #3066, opened 2018-10-21
fix(titus): don't fail when job id is null

emjburns:
Can't get clouddriver to run locally right now, will test Monday.

In an error case, a titus job with have the same server group name as another titus job. This shouldn't happen, but sometimes it does. In that case, we might want to log the error case and then ignore that instance.


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
```diff
@@ -292,17 +292,24 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
       Task task = objectMapper.convertValue(instanceEntry.attributes.task, Task)
 
       Job job
-      if (instanceEntry.attributes.job == null && instanceEntry.relationships[SERVER_GROUPS.ns] && !instanceEntry.relationships[SERVER_GROUPS.ns].empty) {
-        // job needs to be loaded because it was cached separately
-        job = jobData.get(instanceEntry.attributes.jobId)
+      if (instanceEntry.attributes.job != null) {
+        if (instanceEntry.attributes.job == null
+            && instanceEntry.relationships[SERVER_GROUPS.ns]
+            && !instanceEntry.relationships[SERVER_GROUPS.ns].empty) {
```
robzienert: This conditional will always eval `false`. Should remove the first condition, since you're checking for the inverse just a line above.

PR #3068, opened 2018-10-22
fix(provider/cf): environment variables not returned for server groups

jmelchio:


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
```diff
@@ -204,6 +204,8 @@ private CloudFoundryServerGroup map(Application application) {
             .build()))
         .collect(toList());
 
+    Map<String, String> environmentVars = applicationEnv == null || applicationEnv.getEnvironmentJson() == null ? emptyMap() :applicationEnv.getEnvironmentJson();
```
jkschneider: Formatting nit: missing space after `:`.

PR #3071, opened 2018-10-23
fix(provider/kubernetes): associate ingress w/ app (#3069)

lwander:


lwander:
cherry pick into 1.10

PR #3077, opened 2018-10-23
fix(provider/ecs): Rename ResizeServiceAtomicOperationConverter component

clareliguori:
Fixes https://github.com/spinnaker/spinnaker/issues/2746
This is a good candidate for cherry-picking into 1.10.x

Rename ResizeServiceAtomicOperationConverter component name so it doesn't conflict with resizeServerGroup atomic operation

As best I can tell, this is a fun Spring config problem.  Because the atomic operation name is "resizeServerGroup" and the ECS component name is also "resizeServerGroup", the application context starts always returning the ECS component, instead of the right resize converter for the given cloud provider.
https://github.com/spinnaker/clouddriver/blob/29c5c1435ab0c750c4b491208f600647a3e98901/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/ApplicationContextAtomicOperationsRegistry.groovy#L36

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/converters/ResizeServiceAtomicOperationConverter.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/converters/ResizeServiceAtomicOperationConverter.java
@@ -26,7 +26,7 @@ import org.springframework.stereotype.Component;
 
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


tomaslin:
LGTM

tomaslin:
@spinnaker/google-reviewers - can you elaborate on how this can be cherry picked into 1.10?

lwander:
Yup: https://www.spinnaker.io/community/contributing/releasing/#merge-into-the-release-branch

PR #3079, opened 2018-10-24
chore(builds): Remove no longer needed compatibility block

ttomsu:


ttomsu:
Blarg. Seems I can't remove it. Dunno why it worked on my machine.

PR #3081, opened 2018-10-24
fix(provider/openstack): Vendor upstream opstack4j fixes.

msilpala:
A [bug fix in the openstack4j library](https://github.com/ContainX/openstack4j/commit/bb78146aa914b855c2d77e2b4d42e455190fb8eb) has not been added to a release
yet. This bug prevents Spinnaker from caching openstack images when their location is populated.

Until openstack4j makes a new release, we need to vendor the fix within clouddriver.

We've been running with this patch in Production for several weeks now at Target.

### DIFF
--- a/clouddriver-openstack/src/main/java/org/openstack4j/model/image/v2/Image.java
+++ b/clouddriver-openstack/src/main/java/org/openstack4j/model/image/v2/Image.java
@@ -1,0 +1,289 @@
+package org.openstack4j.model.image.v2;
+/**
+ * This file is licensed under the Apache 2 license, quoted below.
+ *
+ * Copyright 2015-16 Jeremy Unruh, ContainX, and OpenStack4j
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+import java.util.Date;
+import java.util.List;
+
+import org.openstack4j.common.Buildable;
+import org.openstack4j.model.common.BasicResource;
+import org.openstack4j.model.image.v2.builder.ImageBuilder;
+import org.openstack4j.openstack.image.v2.domain.GlanceImage.Location;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/*
+ * Works around https://github.com/spinnaker/spinnaker/issues/3018
+ *
+ * TODO: Remove this file after openstack4j fixes this issue upstream. Any release that includes
+ * [this commit](https://github.com/ContainX/openstack4j/commit/bb78146aa914b855c2d77e2b4d42e455190fb8eb) should do it.
+ *
+ * This source file is included in the spinnaker/clouddriver project only because an important fix from openstack4j has
+ * not yet been released. After [this commit](https://github.com/ContainX/openstack4j/commit/bb78146aa914b855c2d77e2b4d42e455190fb8eb)
+ * makes it into a proper release, this file and org.openstack4j.openstack.image.v2.domain.GlanceImage should be removed, and the
+ * spinnaker dependencies updated to include that new openstack4j release.
+ */
+
+/**
+ * A Glance v2.0-2.3 Image
+ * @author emjburns
+ * @see http://developer.openstack.org/api-ref-image-v2.html#showImage-v2
+ */
+public interface Image extends BasicResource, Buildable<ImageBuilder> {
+
+  public enum ImageStatus {
+    /**
+     * Image status is not one of the documented options.
+     * http://docs.openstack.org/developer/glance/statuses.html
+     */
+    UNRECOGNIZED,
+    /**
+     * The image identifier has been reserved for an image in the Glance registry.
+     * No image data has been uploaded to Glance and the image size
+     * was not explicitly set to zero on creation.
+     */
+    QUEUED,
+    /**
+     * Denotes that an image’s raw data is currently being uploaded to Glance.
+     * When an image is registered with a call to POST /images and there is an
+     * x-image-meta-location header present, that image will never be in the saving status
+     * (as the image data is already available in some other location).
+     */
+    SAVING,
+    /**
+     * Denotes an image that is fully available in Glance. This occurs when the
+     * image data is uploaded, or the image size is explicitly set to zero on creation.
+     */
+    ACTIVE,
+    /**
+     * Denotes that access to image data is not allowed to any non-admin user.
+     * Prohibiting downloads of an image also prohibits operations like image
+     * export and image cloning that may require image data.
+     */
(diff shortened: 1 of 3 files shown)


msilpala:
Hmm.

clouddriver-openstack/src/main/java/org/openstack4j/model/image/v2/Image.java
```diff
@@ -0,0 +1,267 @@
...
+import org.openstack4j.common.Buildable;
+import org.openstack4j.model.common.BasicResource;
+import org.openstack4j.model.image.v2.builder.ImageBuilder;
+import org.openstack4j.openstack.image.v2.domain.GlanceImage.Location;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/*
+ * This source file in included in the spinnaker/clouddriver project only because an important fix from openstack4j has
```
jtk54: s/in/is.

clouddriver-openstack/src/main/java/org/openstack4j/model/image/v2/Image.java
```diff
@@ -0,0 +1,267 @@
...
+    PENDING_DELETE;
+
+    @JsonCreator
+    public static ImageStatus value(String v)
+    {
+      if (v == null) return UNRECOGNIZED;
+      try {
+        return valueOf(v.toUpperCase());
+      } catch (IllegalArgumentException e) {
+        return UNRECOGNIZED;
```
jtk54: Log a warning here? Otherwise this is silently failing/defaulting.

clouddriver-openstack/src/main/java/org/openstack4j/model/image/v2/Image.java
```diff
@@ -0,0 +1,267 @@
...
+    PUBLIC,
+    PRIVATE,
+    UNKNOWN;
+
+    @JsonCreator
+    public static ImageVisibility forValue(String value) {
+      if (value != null)
+      {
+        for (ImageVisibility s : ImageVisibility.values()) {
+          if (s.name().equalsIgnoreCase(value)) {
```
jtk54: Does `name()` ever return null?

clouddriver-openstack/src/main/java/org/openstack4j/model/image/v2/Image.java
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  public enum ImageVisibility {
+    PUBLIC,
+    PRIVATE,
+    UNKNOWN;
+
+    @JsonCreator
+    public static ImageVisibility forValue(String value) {
+      if (value != null)
```
jtk54: Nit: prefer the opening bracket on the same line as the `if`. Not sure if this is the convention for the openstack provider, but it is not in the other providers.

clouddriver-openstack/src/main/java/org/openstack4j/openstack/image/v2/domain/GlanceImage.java
```diff
@@ -0,0 +1,633 @@
...
+  @JsonProperty("virtual_size")
+  private Long virtualSize;
+
+  private Map<String, String> additionalProperties = Maps.newHashMap();
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void setName(String name) {
```
jtk54: All the getters and setters can be replaced with a lombok `@Data` on the class.

clouddriver-openstack/src/main/java/org/openstack4j/openstack/image/v2/domain/GlanceImage.java
```diff
@@ -0,0 +1,633 @@
...
+
+  public static ImageBuilder builder() {
+    return new ImageConcreteBuilder();
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public String toString() {
```
jtk54: Can be replaced with lombok `@ToString` on class.

clouddriver-openstack/src/main/java/org/openstack4j/openstack/image/v2/domain/GlanceImage.java
```diff
@@ -0,0 +1,633 @@
...
+      .add("osVersion", osVersion)
+      .add("osDistro", osDistro)
+      .add("ramdiskId", ramdiskId)
+      .add("virtualSize", virtualSize)
+      .toString();
+  }
+
+
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  public static class Images extends ListResult<GlanceImage> {
```
jtk54: Same deal here with this class and its statics - a lot of boilerplate can be removed in favor of lombok annotations if you are so inclined.

clouddriver-openstack/src/main/java/org/openstack4j/model/image/v2/Image.java
msilpala: Doh.

msilpala:
Other than the big comment block at the top, these two files are exact copy-paste jobs of [v2 Image.java](https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/model/image/v2/Image.java) and [v2 GlanceImage.java](https://github.com/ContainX/openstack4j/blob/master/core/src/main/java/org/openstack4j/openstack/image/v2/domain/GlanceImage.java) from the fabric8/kubenetes-client project.

I'm hesitant to make any changes here, since the idea is that at some point we remove these copies and just get the code through our dependency inclusion.

So, I'd suggest opening issues and/or PR's [over there](https://github.com/ContainX/openstack4j/issues) as a better path to getting these valid suggestions into the system.

msilpala:
I will fix the typo and rebase the PR.

duftler:
Please add `TODO:`'s to each of those class comments (ideally referencing either a github id or a relevant bug on spinnaker/spinnaker); otherwise, nobody will find them when searching.

jtk54:
Makes sense for the style/boilerplate comments, feel free to ignore.

msilpala:
Typo fixed; TODOs and links to 
https://github.com/spinnaker/spinnaker/issues/3018 added.

msilpala:
Rebased.

duftler:
Looks like the files need copyright notices added.

msilpala:
Whose copyright would we add to them, given that these are proxies for files in the openstack4j project? It would feel really weird to put any of our copyrights on their source code.

duftler:
Do the original files have copyrights? I would use those (not a lawyer, but that seems most right to me). If they don't, maybe their sibling files do?

msilpala:
If you remove the comment blocks with the TODOs, which I added, the originals are identical to these files.

Just rebased again. In case. :-)

duftler:
I see a top-level LICENSE in that repo. Can copy the contents into the tops of each of these classes (with a note indicating where they came from).

msilpala:
Can do. There are a couple different top-level ones, in the [readme](
https://github.com/ContainX/openstack4j/blob/a3558400c9fb1c1ff231864e2548321e8c91ee1c/README.md#L401) and the [license](
https://github.com/ContainX/openstack4j/blob/a3558400c9fb1c1ff231864e2548321e8c91ee1c/LICENSE#L3). Maybe I'll take the superset.

The only copyrights I'm seeing in source files are ones that we (Target) put there while working on the OST provider a couple years ago.

msilpala:
How's that?

msilpala:
Woo-hoo! Thanks @duftler, @edwinavalos, and @jtk54.

If there's a chance of getting this back-ported to to 1.8.x branch, that would be super-helpful to us as we prep for a minor US retail event happening next month.

PR #3084, opened 2018-10-25
feat(docker): Allow custom config of Docker OkHttpClient

robzienert:
Had a question in OSS about how to provide use custom certs to talk to a Docker Registry and it turns out the current implementation doesn't allow for customization of the OkHttpClient. This PR is gives [similar treatment to what I did in Igor](https://github.com/spinnaker/igor/pull/232), allowing users to add a custom OkHttpClientProvider to Spring. Default functionality is unchanged.


tomaslin:
lgtm

PR #3085, opened 2018-10-25
fix(ha): Add clouddriver-ro-deck.yml halconfig. (#3033)

joonlim:


lwander:
Please tag as a fix

lwander:
(the commit message)

PR #3099, opened 2018-10-29
fix(google): Survive missing, irrelevant target proxies during http l…

duftler:
…p deletion.


duftler:
Closes https://github.com/spinnaker/spinnaker/issues/3338.

PR #3107, opened 2018-10-31
fix(image): don't add beginning glob to image search

emjburns:
When we're searching for an image, there doesn't seem to be a reason why we should add a glob to the front of the query here. Specifying a package of `ate` (query like `find?q=ate`) in a find image stage would lead match `gate`, which doesn't seem like desired behavior. 

Is there some scenario I'm missing here where the first `*` is needed?


ajordens:
breaking change ... not sure it's fair to call something that's been in-place for 4 years a bug?

ajordens:
Since Chris added the globbing, might have something to do with ami searching on the deploy modal. 

Might be a subtle change there but not hugely breaking.

emjburns:
The globbing is needed for the ui dropdown fields. We should add the globs where needed in the UI before merging this PR

robzienert:
Closing - seems this isn't going anywhere for now.

PR #3108, opened 2018-10-31
feat(artifacts): Add support for helm/chart artifacts

Tizaniy:
issue: spinnaker/spinnaker#3284
* add artifact cloud provider for caching
* add helm provider for helm chart caching 
* add API requests for getting artifact definitions


clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/provider/artifacts/helm/HelmArtifactProvider.java
```diff
@@ -0,0 +1,71 @@
...
+
+import com.netflix.spinnaker.cats.agent.Agent;
+import com.netflix.spinnaker.cats.agent.AgentSchedulerAware;
+import com.netflix.spinnaker.clouddriver.cache.SearchableProvider;
+import com.netflix.spinnaker.clouddriver.provider.ArtifactCloudProvider;
+import com.netflix.spinnaker.clouddriver.provider.cache.Keys;
+
+import java.util.*;
+
+public class HelmArtifactProvider extends AgentSchedulerAware implements SearchableProvider {
```
lwander: Caching these is a little heavy-handed, since this will generate a lot of requests for a largely unchanging datastore.

What we found when caching docker images was that we quickly hit rate limits caching large numbers of images, when in reality the big value-add came from populating image drop-downs. 

If the work done by this caching agent can be switched to a live call against the helm repo for the sake of reducing request volume I think this would be in a much better state.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/provider/controller/CacheArtifactController.java
```diff
@@ -0,0 +1,77 @@
...
+
+  Cache cacheView;
+
+  @Autowired
+  CacheArtifactController(Cache cacheView) {
+    this.cacheView = cacheView;
+  }
+
+
+  @RequestMapping(method = RequestMethod.GET, value = "{type}/account/{accountName}/names")
```
ezimanyi: What's a typical (and/or max) number of artifact names we'd expect?  As this endpoint doesn't support filtering, we'd always need to return everything and filter on the front-end, which could become slow if people have a lot of these configured.  (This is somewhat related to @lwander 's comment, as if you change this to fetch on-demand then this would similarly work.)

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/provider/controller/CacheArtifactController.java
```diff
@@ -0,0 +1,77 @@
...
+                        @PathVariable("accountName") String accountName) {
+    switch (type) {
+      case "helm":
+        return findAttributeValue(HELM_NAMES.ns, Keys.getHelmNameKey(accountName), "names");
+      default:
+        throw new NotFoundException("Could not found artifact of type " + type);
+    }
+  }
+
+  @RequestMapping(method = RequestMethod.GET, value = "{type}/account/{accountName}/names/{artifactName}/versions")
```
ezimanyi: Usually these `@RequestMapping`s are prefixed with a `/`; I'm actually not 100% sure what would happen if a nested one (as this) didn't start with a slash (maybe it's implied)...but it's probably clearer to include it.

Tizaniy:
Thank you for your reply @lwander and @ezimanyi. I deleted caching agent and turned requests into live call.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ArtifactController.java
```diff
@@ -66,4 +67,63 @@ StreamingResponseBody fetch(@RequestBody Artifact artifact) {
 
     return outputStream -> IOUtils.copy(artifactDownloader.download(artifact), outputStream);
   }
+
+  @RequestMapping(method = RequestMethod.GET, value = "/{type}/account/{accountName}/names")
+  List<String> getNames(@PathVariable("type") String type,
+                        @PathVariable("accountName") String accountName) {
+    switch (type) {
+      case "helm":
+        return getHelmArtifactNames(accountName);
```
lwander: I would prefer to move away from a growing switch statement, and instead delegate to something autowired by spring to loop over all artifact credentials that can handle this, and take the first/only one that works: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/ArtifactDownloader.java#L53

lwander:
Nice, the live call looks much better already. One comment on how the types are handled

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ArtifactController.java
```diff
@@ -66,4 +67,63 @@ StreamingResponseBody fetch(@RequestBody Artifact artifact) {
...
+    switch (type) {
+      case "helm":
+        return getHelmArtifactVersions(accountName, artifactName);
+      default:
+        throw new NotFoundException("Could not found artifact of type " + type);
+    }
+  }
+
+  private List<String> getHelmArtifactNames(String accountName) {
+    HelmArtifactCredentials helmArtifactCredentials = (HelmArtifactCredentials) findCredentials("helm/chart", accountName);
```
ezimanyi: You'll probably end up addressing this along with @lwander 's comment above (as it's related) but I'm not sure I like the necessity to cast from the base `ArtifactCredentials` to the specific credential type in the base controller, as well as the amount of helm-specific logic.

Ideally the specific downloading code here would be pushed into the `HelmArtifactCredentialsClass` and the controller would call a more general wrapper method rather than making specific calls like `downloadIndex` itself.

I actually wonder if we should even just expose `getArtifactNames` and `getArtifactVersions` in the `ArtifactCredentials` interface with default implementations that throw `Not implemented`, which will set us up for less code duplication if/when we do this for other types.

Tizaniy:
@ezimanyi, @lwander, thank you for review. There is one problem. `HelmArtifactCredentials` class handles `helm/chart` type, but `helm` is expected as type variable. Should it be changed?

eromanova:
@ezimanyi @lwander Full artifact types contains forward slashes (`helm/chart`, `github/file` etc.) and if we include full artifact type in the call, we must encode slashes to `%2F`. But such URLs are rejected. We can disable rejecting by configuring spring to enable `%2F` string, but it can lead to security vulnerabilities. We can replace slashes in request with another symbol (for example `_` or `-`) or change the request format to handle two parts of artifact type. What do you think about it? Is there more appropriate workaround?

ezimanyi:
That's interesting; I hadn't realized that encoded slashes would be rejected but read a bit about the security vulnerabilities in `PATH_INFO`.  I think that replacing the forward slash with something else, such as `-` is probably the best solution here, though don't have a very strong opinion.

Tizaniy:
@ezimanyi, @lwander, I changed type handling, please take a look.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ArtifactController.java
```diff
@@ -66,4 +63,33 @@ StreamingResponseBody fetch(@RequestBody Artifact artifact) {
...
+    }
+    return credentials.getArtifactNames();
+  }
+
+  @RequestMapping(method = RequestMethod.GET, value = "/{type}/account/{accountName}/names/{artifactName}/versions")
+  List<String> getVersions(@PathVariable("type") String type,
+                           @PathVariable("accountName") String accountName,
+                           @PathVariable("artifactName") String artifactName) {
+    ArtifactCredentials credentials = findArtifactCredentials(accountName);
+    if (!credentials.handlesType(type.replace("-","/"))) {
```
lwander: This makes me a little nervous -- it's effectively doing the same thing that a url-encoding does, but instead of relying on a shared standard, it only applies to this endpoint.

What if we made `type` a query param? Also what's wrong with relying on url-encoding? Our http libraries are smart enough to handle the `/` on both ends (request & controller).

Tizaniy: @lwander, the server rejects requests and returns 400 HTTP error when it sees `%2F` in the request. This is default Tomcat behavior. `ALLOW_ENCODED_SLASH` boolean flag may be specified for Tomcat, but it starts to see `%2F` as delimiter. Anyway default implementation of HttpFirewall interface that Spring Security uses is the reason requests are rejected as well.

lwander: Ah right -- is that only for path params or query params as well? Only asking in the hope that we can keep the API a little more readable/understandable

lwander: Does `/artifacts/account/{accountName}/versions?<type>&<name>&<location>...` seem OK? it puts all identifying details aside from account into a query param, allowing a user to specify only the keys they feel are needed.

eromanova: @lwander Appreciate your suggestion, but I would like to clarify. Do you mean we should leave 2 API: 
`/artifacts/account/{accountName}/names?<type>&<location>&...` and
`/artifacts/account/{accountName}/versions?<type>&<name>&<location>...`

Or we should leave only one:
`/artifacts/account/{accountName}/versions?<type>&<name>&<location>...` and get all artifact names if `name` field is missed. If it's not - get artifact versions according to chosen name.

Nevertheless, I'm a little bit confused because `type` is still going after `artifactAccount`. But if we can't use `type` as path variable and there are no solutions, I think the best way is to leave API:
`/artifacts/account/{accountName}/names?<type>`
`/artifacts/account/{accountName}/versions?<type>`

lwander: > Do you mean we should leave 2 API:
`/artifacts/account/{accountName}/names?<type>&<location>&...` and
`/artifacts/account/{accountName}/versions?<type>&<name>&<location>...`

That sounds good to me. 

> Nevertheless, I'm a little bit confused because type is still going after artifactAccount. 

I agree it seems a bit odd -- but in reality there is no reason why an artifactAccount can't download more than 1 type. Even in the helm case, if you have a repository in GCS or S3, you should (in the future) be able to use a GCS or S3 account to read the contents of the repository, knowing that it is a proper helm repo.

Tizaniy: @lwander, please take a look at changes I made according to last replies.

PR #3110, opened 2018-10-31
chore(builds): Conditionalize settings.gradle includes with includeCloudProviders property

jkschneider:
This finishes the work started with https://github.com/spinnaker/spinnaker/issues/2844 to allow us to build a leaner version of clouddriver with only the implementations we are interested in.

* Decouples `clouddriver-web` tests from Amazon and Google specific code
* Remove unused CloudFoundry resources in `clouddriver-web` test resources folder.

Once this is merged, adding `includeCloudProviders=MYPROVIDER` in your `~/.gradle/gradle.properties` (or any other way of setting that property) reduces the build scope of clouddriver to just the provider(s) you are interested in.


clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/controllers/OperationsControllerSpec.groovy
```diff
@@ -69,11 +68,17 @@ class OperationsControllerSpec extends Specification {
...
   }
 
+  static class MockGoogleDeployDescription implements DeployDescription {
+  }
+
+  static class MockAmazonDeployDescription implements DeployDescription {
+  }
+
   @Shared
   def googlePreProcessor = new AtomicOperationDescriptionPreProcessor() {
```
ttomsu: Nothing about this is related to google, right? If not, I'd replace it with `testPreProcessor` or something similar. Same with any other mention of a specific cloud provider.

ttomsu:
@dibyom - can we get away with the settings.gradle file like this? Does it break intellij (i.e. `./gradlew clean idea assemble`)?

jkschneider:
> can we get away with the settings.gradle file like this? Does it break intellij (i.e. `./gradlew clean idea assemble`)?

IntelliJ is fine, both with the new model (`idea .`) and the old model (`./gradlew idea`).

clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/controllers/OperationsControllerSpec.groovy
jkschneider: Renamed types and variables to remove any appearance of a reference to Amazon or Google implementations.

PR #3112, opened 2018-10-31
feat(security): Support multiple applications in `ApplicationNameable`

ajordens:
Allow an implementation of `ApplicationNameable` to support authorization
checks against _multiple_ applications.

This handles scenarios where a particular operation may target multiple
server groups that may or may not exist in the same application.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
@@ -20,12 +20,13 @@ import com.netflix.spinnaker.clouddriver.aws.model.AmazonAsgLifecycleHook
 import com.netflix.spinnaker.clouddriver.aws.model.AmazonBlockDevice
 import com.netflix.spinnaker.clouddriver.deploy.DeployDescription
 import com.netflix.spinnaker.clouddriver.orchestration.events.OperationEvent
+import com.netflix.spinnaker.clouddriver.security.resources.ApplicationNameable
 import groovy.transform.AutoClone
 import groovy.transform.Canonical
 
 @AutoClone
 @Canonical
-class BasicAmazonDeployDescription extends AbstractAmazonCredentialsDescription implements DeployDescription {
+class BasicAmazonDeployDescription extends AbstractAmazonCredentialsDescription implements DeployDescription, ApplicationNameable {
   String application
   String amiName
   String stack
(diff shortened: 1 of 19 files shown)


ajordens:
I err'd on the side of using `ApplicationNameable` and `ServerGroupNameable` since they were more descriptive of the purpose and commonly used. It didn't strike me as abnormal for these to be plural as we do have operations that work on more than one resource.

The alternative would have been to do some conversion from `ServerGroupNameable` usage to `ResourcesNameable` since it already offered a plural `getApplications()`.

I will be going through all aws and titus descriptions and implementing the appropriate interfaces as we're fairly wide open with respect to application restriction.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
```diff
@@ -20,12 +20,13 @@ import com.netflix.spinnaker.clouddriver.aws.model.AmazonAsgLifecycleHook
...
 import com.netflix.spinnaker.clouddriver.deploy.DeployDescription
 import com.netflix.spinnaker.clouddriver.orchestration.events.OperationEvent
+import com.netflix.spinnaker.clouddriver.security.resources.ApplicationNameable
 import groovy.transform.AutoClone
 import groovy.transform.Canonical
 
 @AutoClone
 @Canonical
-class BasicAmazonDeployDescription extends AbstractAmazonCredentialsDescription implements DeployDescription {
+class BasicAmazonDeployDescription extends AbstractAmazonCredentialsDescription implements DeployDescription, ApplicationNameable {
```
ajordens: Subsequently I might pull the `implements ApplicationNameable` up to `DeployDescription` and all cloud providers will pick it up.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/ResizeAsgDescription.groovy
```diff
@@ -17,18 +17,16 @@
...
 
-class ResizeAsgDescription extends AbstractAmazonCredentialsDescription {
-  String serverGroupName
-  String region
+class ResizeAsgDescription extends AbstractAmazonCredentialsDescription implements ServerGroupNameable {
   ServerGroup.Capacity capacity
   List<AsgTargetDescription> asgs = []
 
-  @Deprecated
-  String asgName
```
ajordens: Pretty sure these aren't necessary anymore ... we've got a `DescriptionPreProcessor` that takes all these `@Deprecated` values and places them consistently under `asgs`.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
cfieber: seems reasonable to me

PR #3113, opened 2018-11-01
fix(provider/aws): populate loadbalancer attributes

gardleopard:
This populates loadBalancerNameToAttributes when edda is not enabled


tomaslin:
LGTM

PR #3114, opened 2018-11-01
perf(provider/kubernetes): parallelize calls to kubectl

lwander:


lwander:
Nope!

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2CachingAgent.java
```diff
@@ -75,7 +75,7 @@ protected KubernetesKind primaryKind() {
   }
 
   protected Map<KubernetesKind, List<KubernetesManifest>> loadPrimaryResourceList() {
-    Map<KubernetesKind, List<KubernetesManifest>> result = namespaces.stream()
+    Map<KubernetesKind, List<KubernetesManifest>> result = namespaces.parallelStream()
```
ianferguson: Does it make sense to use an application specific thread pool for parallelizing these? iirc, parallelStream is using a single JVM wide ForkJoinPool under the hood with the thread count set to the number of cpu cores, which may be non-ideal if other classes are using or end up using `parallelStream` calls, and is potentially a lower thread count than optimal for calls that are largely waiting on network i/o to complete.

lwander: You're exactly right -- we've seen issues with too many kubectl calls running at once within docker leaving dozens of zombie processes. Artificially limiting this for now is intentional, and having tested this under load it seems to behave well. If we run into problems in the future we'll swap this out for a local thread pool.

PR #3117, opened 2018-11-02
feat(core): General purpose `DescriptionAuthorizer`

ajordens:
Ported the `authorize()` method from `DescriptionValidator` and added
a metric in the event that a description presents an account _but_ no
application(s).

This metric will emit some false positives but still prove useful in the
short-term to highlight descriptions that might not implement
`ApplicationNameable` (or not implement it properly).

The previous behavior expected a `Description` to implement
`DescriptionValidator` in order to have authorization applied, however
this is not a requirement and missed some legitimate scenarios.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/deploy/DescriptionAuthorizer.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/deploy/DescriptionAuthorizer.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.deploy;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spectator.api.Id;
+import com.netflix.spectator.api.Registry;
+import com.netflix.spinnaker.clouddriver.security.resources.AccountNameable;
+import com.netflix.spinnaker.clouddriver.security.resources.ApplicationNameable;
+import com.netflix.spinnaker.clouddriver.security.resources.ResourcesNameable;
+import com.netflix.spinnaker.fiat.shared.FiatPermissionEvaluator;
+import org.slf4j.Logger;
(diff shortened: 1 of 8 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/deploy/DescriptionAuthorizer.java
```diff
@@ -0,0 +1,96 @@
...
+      applications.forEach(application -> {
+        if (!fiatPermissionEvaluator.hasPermission(auth, application, "APPLICATION", "WRITE")) {
+          errors.reject("authorization", format("Access denied to application %s", application));
+        }
+      });
+    }
+
+
+    if (account != null && applications.isEmpty()) {
+      registry
```
ajordens: Will also log a suitable message ... as mentioned in the PR description I'm looking to get an idea of how many descriptions are account restricted but not application-restricted.

Likely also worth a `success` tag on this counter.

The plan would be to mark the descriptions that legitimately have no applications, and fix the others.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/deploy/DescriptionValidator.groovy
```diff
@@ -33,40 +33,4 @@ public abstract class DescriptionValidator<T> implements VersionedCloudProviderO
   }
 
   abstract void validate(List priorDescriptions, T description, Errors errors)
-
-  @Autowired(required = false)
-  FiatPermissionEvaluator permissionEvaluator
-
-  void authorize(T description, Errors errors) {
```
ajordens: Discovered there are a handful (or more!) descriptions that do not implement `DescriptionValidator` and thus are ignored as far as authorization checks go.

I don't think `implements DescriptionValidator` should be a pre-req and this logic has been extracted / converted to Java.

ajordens:
I have some actual testing to do but wanted to make sure nothing looked off with this general strategy @ttomsu?

I've got another pending PR that touches all the aws and titus descriptions and makes them
 implement the *Nameable interfaces as appropriate. 

Would like to get this change merged prior tho.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/OperationsController.groovy
```diff
@@ -161,24 +164,18 @@ class OperationsController {
         )
         if (validator) {
           validator.validate(descriptions, description, errors)
-          validator.authorize(description, errors)
         } else {
-          def msg = "No validator found for operation `${description?.class?.simpleName}` and cloud provider $cloudProvider"
-
-          switch(opsSecurityConfigProps.onMissingValidator) {
-            case SecurityConfig.SecurityAction.WARN:
```
ajordens: This seems unnecessary now that validation != authorization.

ttomsu:
I like the change.

I made a half-hearted attempt to build in some automatic checking of this stuff: https://github.com/spinnaker/clouddriver/blob/29c5c1435ab0c750c4b491208f600647a3e98901/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/listeners/OperationsTypeChecker.groovy#L64

Not sure if you found that or would find it useful to ferret out additional non-compliant Descriptions.

PR #3118, opened 2018-11-02
fix(aws): Ensure that AWS operations enforce WRITE access to applications

ajordens:
The majority of AWS operations were not implementing one of:
- `ApplicationNameable`
- `ServerGroupNameable`
- `ResourcesNameable`

In these cases, there was no check being performed that a particular
operation had access to the application owning the particular resource
being modified.


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/AbstractRegionAsgInstanceIdsDescription.groovy
```diff
@@ -16,9 +16,16 @@
...
+
+abstract class AbstractRegionAsgInstanceIdsDescription extends AbstractAmazonCredentialsDescription implements ServerGroupNameable {
   String region
   String asgName
   List<String> instanceIds
   Integer targetHealthyDeployPercentage
+
+  @Override
+  Collection<String> getServerGroupNames() {
+    return [asgName]
```
ajordens: there is a check in the `DescriptionAuthorizer` that'll filter out null ... I cannot say for certain whether these values are ever non-null.

PR #3120, opened 2018-11-03
feat(kubernetes/v2): Support statefulset partition rollouts.

benjaminws:
# Statefulset Partition roll out support

This PR adds support for statefulset partition rollouts. In order to do this with some confidence, we've added specs around the `KubernetesStatefulSetHandler` as well.

## Add support for statefulset Partition rollout

The logic was taken directly from the kubectl code here
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/rollout_status.go#L137-L146
c922c2f 

This feature is useful for facilitating staged rollouts of statefulsets. 

Simple Scenario: You want to update a subset of images in a statefulset without triggering a full deployment of everything.

1. Patch statefulset with n partitions:

```
spec:
  updateStrategy:
    rollingUpdate:
      partition: n
    type: RollingUpdate
```

2. Patch statefulset to update the image:

```
- op: replace
  path: /spec/template/spec/containers/0/image
  value: 'image_name:0.0.2'
```

3. Inspect pods to confirm update:

```
for p in $(seq 0 n); do kubectl get po test-$p --template '{{range $i, $c := .spec.containers}}{{$c.image}}{{end}}'; echo; done
```

Any pod with an ordinal > `partition` should be updated, while the remaining pods should be using the original image.

This patch adds support for this type of workflow, commonly used in statefulset deployments.

Looking for input on making sure I've covered everything.

## Add specs for all scenarios

During this process, a small bug was uncovered with `KubernetesManifest.isNewerThanObservedGeneration()`.

The `generation` and `observedGeneration` properties of `KubernetesManifest` are hydrated from JSON as Doubles at some point, so this would end up throwing a `ClassCastException`, and that was being ignored, so the method ultimately returned true. My hypothesis is that these properties are coerced as doubles by `gson` when the JSON response is hydrated from kubectl calls in `src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubectlJobExecutor.java`

Note that these properties are integers according to the kubernetes API docs.

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#objectmeta-v1-meta

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#rollingupdatestatefulsetstrategy-v1-apps


lwander:
Maybe @wjoel should have a look too

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesDeploymentHandler.java
```diff
@@ -79,7 +79,7 @@ public Status status(KubernetesManifest manifest) {
     if (manifest.getApiVersion().equals(EXTENSIONS_V1BETA1)
         || manifest.getApiVersion().equals(APPS_V1BETA1)
         || manifest.getApiVersion().equals(APPS_V1BETA2)) {
-      if (!manifest.isNewerThanObservedGeneration()) {
+      if (manifest.isNewerThanObservedGeneration()) {
```
wjoel: Hmm. What's the logic behind this change? Was the previous logic broken?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesStatefulSetHandler.java
```diff
@@ -130,18 +131,33 @@ private Status status(V1beta2StatefulSet statefulSet) {
...
+      }
+      result.setStable(new Status.Condition(true,"Partitioned roll out complete"));
+      return result;
     }
 
     existing = status.getCurrentReplicas();
     if (existing == null || (desiredReplicas != null && desiredReplicas > existing)) {
       return result.unstable("Waiting for all updated replicas to be scheduled");
-    }
+     }
```
wjoel: Formatting error.

clouddriver-kubernetes/src/test/resources/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/statefulsetpartitionbase.json
```diff
@@ -0,0 +1,151 @@
...
+                ],
+                "dnsPolicy": "ClusterFirst",
+                "restartPolicy": "Always",
+                "schedulerName": "default-scheduler",
+                "securityContext": {},
+                "terminationGracePeriodSeconds": 10,
+                "tolerations": []
+            }
+        },
+        "updateStrategy": {
```
wjoel: Do we also have a test for the case where `updateStrategy` isn't set? I'm pretty sure it gets a default value but it would be good to know that `statefulSet.getSpec().getUpdateStrategy().getRollingUpdate();` never throws an NPE.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesDeploymentHandler.java
benjaminws: It wasn't broken, but inverted. The method name states `isNewerThanObservedGeneration` but the original code in this method would do just the opposite, returning false if `generation > observedGeneration`, and all the usages were negated. I felt like this was a reasonable refactor.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesStatefulSetHandler.java
benjaminws: doh!

clouddriver-kubernetes/src/test/resources/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/statefulsetpartitionbase.json
benjaminws: Great question. I will add a case for it because as you state, I was also pretty sure it gets a default value, but this would code would definitely NPE if `updateStrategy` is missing.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesDeploymentHandler.java
wjoel: Ah right. Agreed!

clouddriver-kubernetes/src/test/resources/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/statefulsetpartitionbase.json
benjaminws: FWIW I did some manual testing around this, and the statefulset always has `updateStrategy` regardless of if it's created/patched with that section or not. I think this is safe. Also pretty sure we would have seen problems in the past, since https://github.com/spinnaker/clouddriver/blob/d455405a66389b2b27f0d0b0ce82c020fa348810/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesStatefulSetHandler.java#L117-L119 should have NPE'd before now. :)

benjaminws:
Do we still feel good about this? Ok to merge?

PR #3132, opened 2018-11-07
feat(provider/cf): add option for user to define service creation/del…

Jammy-Louie:
add configurability to the timeout setting when creating a service instance or deleting a service instance


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/description/DeployCloudFoundryServiceDescription.java
```diff
@@ -24,7 +24,7 @@
 
 @Data
 @EqualsAndHashCode(callSuper = true)
-public class DeployCloudFoundryServiceDescription extends DestroyCloudFoundryServiceDescription {
+public class DeployCloudFoundryServiceDescription extends AbstractCloudFoundryServiceDescription {
   String service;
```
jkschneider: What is the difference between `service` and `serviceName` in the base class?

Jammy-Louie: "service" appears to be the service name that appears in the market place.
"serviceName" appears to be the service instance name

we should be more explicit with the naming and "serviceName" should be "serviceInstanceName"

I can do this in a separate PR which require a minor corresponding change in deck to rename the fields from "serviceName"  to "serviceInstanceName"

jkschneider: The point is, we're actually adding a field here to the original command. Is that intended?

jkschneider: Ah, I see. It extended `DestroyCloudFoundryServiceDescription` previously. Yikes.

PR #3137, opened 2018-11-08
refactor(provider/cf): convert tests to junit5

jmelchio:


jkschneider:
@jmelchio Could you update the branch with the resolved conflict?

jmelchio:
@jkschneider updated the branch and resolved the conflict.

clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/ServiceInstancesTest.java
```diff
@@ -32,23 +32,20 @@
...
+class ServiceInstancesTest {
   private ServiceInstances serviceInstances;
   private CloudFoundrySpace cloudFoundrySpace;
   private ServiceInstanceService serviceInstanceService;
 
-  @Before
-  public void before() {
+  @BeforeEach
+  void before() {
     serviceInstanceService = mock(ServiceInstanceService.class);
```
claymccoy: You could just initialize with the mocks above and avoid a  BeforeEach entirely if this is all it is needed for. I tend to use BeforeEach for doing nontrivial preparation of the class under test.

clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/HttpCloudFoundryClientTest.java
```diff
@@ -21,19 +21,19 @@
...
 import java.io.IOException;
 
 import static com.squareup.okhttp.Protocol.HTTP_1_1;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.Assert.fail;
+import static org.junit.jupiter.api.Assertions.fail;
 import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.*;
 
 public class HttpCloudFoundryClientTest {
```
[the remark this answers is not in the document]
jmelchio: Probably not. I'll take car of this when I fix the timing issue.

clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/ServiceInstancesTest.java
jmelchio: True, I'll fix it.

clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/HttpCloudFoundryClientTest.java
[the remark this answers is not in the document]
jmelchio: Changed

clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/ServiceInstancesTest.java
jmelchio: Changed.

clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/ServiceInstancesTest.java
```diff
@@ -132,46 +124,50 @@ public void shouldSuccessfullyCreateService() {
...
-      cloudFoundrySpace,
-      Duration.ofSeconds(4));
+    assertThrows(CloudFoundryApiException.class, () ->
+      serviceInstances.createServiceInstance("newServiceInstanceName",
+        "serviceName",
+        "ServicePlan1",
+        Collections.emptySet(),
+        null,
+        cloudFoundrySpace,
+        Duration.ofMillis(40))
```
claymccoy: Why 40 ms? What happens if this is 0?

jmelchio: The retry config controller fails if we set 0 ms.  I am hesitant to change the implementation of the main code to cover because we won't be testing production-ready code if we have a code path to cover only the tests in which we set this timeout to 0.

stuart-pollock: The retry config controller fails if we set 0 ms. I am hesitant to change the implementation of the main code to cover because we won't be testing production-ready code if we have a code path to cover only the tests in which we set this timeout to 0.

jmelchio: When you try to set it to anything below 10 ms the you get an error message if I recall correctly.

claymccoy: I understand where you are coming from and I am fine with this going in for now, but I have a few reservations about this long term.

In a situation like this if the number is too high then the tests take too long, and if the number is too low then the tests fail. So you try to find this Goldilocks number in the middle that is not too brittle or too slow. But in reality, it is not ideal because the ultimate value always suffers a little of both.

I'm not too concerned about sidestepping code from another trusted project like resilence4j or our own code that is adequately tested. It doesn't always have to run exactly as it would in production to be a good test especially when some kind of I/O is involved. I like fast unit tests.

I looked through to see where this is used. That raised more questions. It doesn't look like this is actually used as a timeout, or that the retry logic has any real problem with a 0 timeout. 
rather max attempts cannot be 0, which makes sense. And the timeout and polling interval are used to derive the max attempts. So a 0 timeout results in a 0 max attempts. This does not seem intuitive to me. I would expect the max attempts to be defined specifically and separate from a timeout.
`.maxAttempts((int) (timeout.toMillis() / pollingInterval.toMillis()))`

When I trace these calls all the way up to the timeout specified in the operation, I am still left thinking that this is better expressed as max attempts, and maybe not even required to be configurable from the operation itself.

Above that I see `.waitDuration(pollingInterval)` which is the actual part that I believe is extending test execution time. In there `pollingInterval.toMillis() < 10` will result in an exception. But that isn't the only way to specify polling. That just creates an IntervalFunction, and there are other ways to do that. It would seem possible to create an IntervalFunction that did not extend test time.

So my longer-term suggestion would be to specify max attempts in a simpler way that was not bound to timeout and polling, and use an IntervalFunction that avoided sleeping.

jkschneider: On the last point, the alternative is to still specify wait duration but mock the clock. That was also possible in this case, with reflection.

PR #3141, opened 2018-11-09
refactor(artifacts): Clean up spring in artifacts

ezimanyi:


ezimanyi:
@lwander : mind double checking my changes, particularly those in the above comment, make sense?

PR #3144, opened 2018-11-09
fix(provider/kubernetes): consult kind list for crd caching

lwander:


lwander:
That will definitely help, but so far we don't do any kind checks inside of very nested loops, or heavily run code. This example runs once every 30s, and typically that list is quite short.

PR #3149, opened 2018-11-10
fix(provider/gce): Wait for batch requests to finish before returning

ezimanyi:
We currently send GCE batch requests to a thread pool for execution but don't wait for execution to finish. This leads to race conditions where cache refresh requests can return success before the cache is actually updated.

### DIFF
--- a/clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/batch/GoogleBatchRequest.java
+++ b/clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/batch/GoogleBatchRequest.java
@@ -41,9 +41,11 @@ import java.util.concurrent.TimeUnit;
 public class GoogleBatchRequest {
 
   private static final int MAX_BATCH_SIZE = 100; // Platform specified max to not overwhelm batch backends.
-  private static final int DEFAULT_EXECUTE_TIMEOUT_MINUTES = 10;
   private static final int DEFAULT_CONNECT_TIMEOUT_MILLIS = (int) TimeUnit.MINUTES.toMillis(2);
   private static final int DEFAULT_READ_TIMEOUT_MILLIS = (int) TimeUnit.MINUTES.toMillis(2);
+  // Many requests (ex: caching load balancers) have callbacks that execute further requests, up to a depth of 3-4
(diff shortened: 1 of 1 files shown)


clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/batch/GoogleBatchRequest.java
```diff
@@ -41,6 +41,7 @@
 public class GoogleBatchRequest {
 
   private static final int MAX_BATCH_SIZE = 100; // Platform specified max to not overwhelm batch backends.
+  private static final int DEFAULT_EXECUTE_TIMEOUT_MINUTES = 10;
```
lwander: What's the relationship between `DEFAULT_READ_TIMEOUT_MILLIS`, and the total execution time here?

jtk54: Looks like `DEFAULT_READ_TIMEOUT_MILLIS` deals with the open connection in the batch and `DEFAULT_EXECUTE_TIMEOUT_MINUTES` is for threadpool shutdown as it is now. Should those two be the same?

lwander: My first guess was this should be: `DEFAULT_EXECUTE_TIMEOUT_MINUTES = DEFAULT_CONNECT_TIMEOUT_MILLIS + DEFAULT_READ_TIMEOUT_MILLIS + small_constant`, but I wasn't sure if we do multiple sequential reads, and if the fact that it now looks like we allow for ~4 of those assuming the chosen constants were intentional.

ezimanyi:
The reason I left a reasonably large buffer for `DEFAULT_EXECUTE_TIMEOUT_MINUTES` is to account for cases like caching load balancers, where even after the initial connect we have callbacks that process the data and do further API requests.  I believe the depth of requests reaches ~3-4 in the case of load balancers, which is why I picked a factor of 5 above the individual stage timeouts.

lwander:
SGTM, could you drop that into a comment? Or codify it as an expression?

ezimanyi:
Sounds good, added a comment

PR #3150, opened 2018-11-12
feat(provider/appengine): allow limiting the set of indexed services and versions

sbwsg:
Fixes https://github.com/spinnaker/spinnaker/issues/3539

### DIFF
--- a/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/config/AppengineConfigurationProperties.groovy
+++ b/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/config/AppengineConfigurationProperties.groovy
@@ -43,6 +43,10 @@ class AppengineConfigurationProperties {
     String sshKnownHostsFilePath
     boolean sshTrustUnknownHosts
     GcloudReleaseTrack gcloudReleaseTrack
+    List<String> services
+    List<String> versions
+    List<String> omitServices
+    List<String> omitVersions
 
     void initialize(AppengineJobExecutor jobExecutor) {
       if (this.jsonPath) {
--- a/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AbstractAppengineCachingAgent.groovy
+++ b/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AbstractAppengineCachingAgent.groovy
@@ -40,6 +40,30 @@ abstract class AbstractAppengineCachingAgent implements CachingAgent, AccountAwa
     this.credentials = credentials
   }
 
+  boolean shouldIgnoreLoadBalancer(String loadBalancerName) {
+    if (credentials.services != null && !credentials.services.isEmpty() &&
+      credentials.services.every { !loadBalancerName.matches(it) }) {
+      return true
+    }
+    if (credentials.omitServices != null && !credentials.omitServices.isEmpty() &&
+      credentials.omitServices.any { loadBalancerName.matches(it) }) {
+      return true
+    }
+    return false
+  }
+
+  boolean shouldIgnoreServerGroup(String serverGroupName) {
+    if (credentials.versions != null && !credentials.versions.isEmpty() &&
+      credentials.versions.every { !serverGroupName.matches(it) }) {
+      return true
+    }
+    if (credentials.omitVersions != null && !credentials.omitVersions.isEmpty()
+      && credentials.omitVersions.any { serverGroupName.matches(it) }) {
+      return true
+    }
+    return false
+  }
+
   static void cache(Map<String, List<CacheData>> cacheResults,
                     String cacheNamespace,
                     Map<String, CacheData> cacheDataById) {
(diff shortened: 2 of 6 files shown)


clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AppengineServerGroupCachingAgent.groovy
```diff
@@ -100,6 +100,9 @@ class AppengineServerGroupCachingAgent extends AbstractAppengineCachingAgent imp
     }
 
     def serverGroupName = data.serverGroupName.toString()
+    if (credentials.omitVersions != null && credentials.omitVersions.size() > 0 && credentials.omitVersions.contains(serverGroupName)) {
```
sbwsg: replace with `shouldIgnoreVersion(serverGroupName)`

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/config/AppengineConfigurationProperties.groovy
```diff
@@ -43,6 +43,10 @@ class AppengineConfigurationProperties {
     String sshKnownHostsFilePath
     boolean sshTrustUnknownHosts
     GcloudReleaseTrack gcloudReleaseTrack
+    List<String> services
+    List<String> versions
```
danielpeach: If I've blacklisted `myapp-v001`, does that mean that `myapp-v002` will still show up? 

I think it might be more intuitive to whitelist / blacklist using regex so that I could say that I want to exclude any version that starts with `myapp`. You could also whitelist / blacklist using app name (and use Frigga to parse the server group's name).

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AppengineServerGroupCachingAgent.groovy
lwander: Also, we generally prefer `.isEmpty` to `.size() > 0`

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/security/AppengineNamedAccountCredentials.groovy
```diff
@@ -54,6 +54,11 @@ class AppengineNamedAccountCredentials implements AccountCredentials<AppengineCr
   final GcloudReleaseTrack gcloudReleaseTrack
   final List<AppengineGitCredentialType> supportedGitCredentialTypes
 
+  final List<String> services
+  final List<String> versions
+  final List<String> omitServices
+  final List<String> omitVersions
```
lwander: I would put a check somewhere that at most one of `X`, and `omitX` is specified.

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AbstractAppengineCachingAgent.groovy
```diff
@@ -40,6 +40,26 @@ abstract class AbstractAppengineCachingAgent implements CachingAgent, AccountAwa
     this.credentials = credentials
   }
 
+  boolean shouldIgnoreLoadBalancer(String loadBalancerName) {
+    if (credentials.services != null && !credentials.services.contains(loadBalancerName)) {
```
lwander: In terms of default behavior, I would recommend reading empty and null as "not specified", especially since it's easy to either have halyard or another config gen tool generate empty as the default value, and then it's also possible that a change in the default config properties leaves those lists with an empty array as the default.

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/config/AppengineConfigurationProperties.groovy
jtk54: :+1:

sbwsg: ✅ done & doc strings updated here: https://github.com/spinnaker/halyard/pull/1102/files

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AbstractAppengineCachingAgent.groovy
```diff
@@ -40,6 +40,30 @@ abstract class AbstractAppengineCachingAgent implements CachingAgent, AccountAwa
     this.credentials = credentials
   }
 
+  boolean shouldIgnoreLoadBalancer(String loadBalancerName) {
+    if (credentials.services != null && !credentials.services.isEmpty() &&
+      credentials.services.find { loadBalancerName.matches(it) } == null) {
```
danielpeach: Could be an `.any` (or an `.every`)

sbwsg: ✅ Updated `== null` to be `.every` and `!= null` to be `.any`.

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/security/AppengineNamedAccountCredentials.groovy
lwander: Maybe it's useful now that regex is supported -- without it I think it's unnecessary.

sbwsg: Oh wait, do you mean here that I should check if any string appears both in `X` and `omitX`?  That would indeed be wrong.

PR #3153, opened 2018-11-12
feat(artifacts): Add jar artifact account support

ezimanyi:
Adds support for fetching artifacts of type jar/file from a remote repository, using the aether library.

### DIFF
--- a/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/jar/JarArtifactCredentials.java
+++ b/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/jar/JarArtifactCredentials.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2018 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.artifacts.jar;
+
+import com.netflix.spinnaker.clouddriver.artifacts.config.ArtifactCredentials;
+import com.netflix.spinnaker.kork.artifacts.model.Artifact;
+import lombok.Data;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.maven.repository.internal.MavenRepositorySystemUtils;
+import org.eclipse.aether.DefaultRepositorySystemSession;
+import org.eclipse.aether.RepositorySystem;
+import org.eclipse.aether.RepositorySystemSession;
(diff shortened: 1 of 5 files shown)


clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/jar/JarArtifactCredentials.java
```diff
@@ -0,0 +1,118 @@
...
+    locator.addService(RepositoryConnectorFactory.class, BasicRepositoryConnectorFactory.class);
+    locator.addService(TransporterFactory.class, FileTransporterFactory.class);
+    locator.addService(TransporterFactory.class, HttpTransporterFactory.class);
+    locator.addService(RepositoryLayoutProvider.class, DefaultRepositoryLayoutProvider.class);
+    return locator.getService(RepositorySystem.class);
+  }
+
+  private Path getRepositoryPath() throws IOException {
+    String uuid = UUID.randomUUID().toString();
+    Path tempDir = Paths.get(System.getProperty("java.io.tmpdir"), uuid);
```
lwander: Maybe we should be sure to delete the downloaded artifact after we stream it? I can imagine this filling up pretty quickly, and generally this is only cleared after system restarts.

ezimanyi:
@jkschneider : Would be interested in your comments here too; this only handles maven repository layouts for now, but can be improved in the future to add other types.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/jar/JarArtifactCredentials.java
ezimanyi: It also mirrors the repo with metadata, etc...so I don't think that I can delete the artifact itself without having things end up in an inconsistent state.  I wonder if maybe an option is to have each fetch request use its own local repo, and delete it when it's done.

I wish there were a simpler API to just grab a single artifact remotely without copying it locally, but as far as I can tell that's not an option.  Let me think about this some more.

lwander: We can leave this as-is and point it out in the docs for now.

jkschneider: This will be solved if we move away from aether, which we will have to do anyway.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/jar/JarArtifactCredentials.java
```diff
@@ -0,0 +1,118 @@
...
+    session.setLocalRepositoryManager(system.newLocalRepositoryManager(session, localRepo));
+    return session;
+  }
+
+  private File getJarFile(String reference) {
+    org.eclipse.aether.artifact.Artifact jarArtifact = new DefaultArtifact(reference);
+    ArtifactRequest request = new ArtifactRequest(jarArtifact, Collections.singletonList(remoteRepository), null);
+    ArtifactResult result;
+    try {
+      result = repositorySystem.resolveArtifact(repositorySystemSession, request);
```
jkschneider: Does this result in transitives downloading? I'll make sure we don't do that inadvertently in the Ivy impl.

ezimanyi:
Superseded by https://github.com/spinnaker/clouddriver/pull/3177

PR #3156, opened 2018-11-13
fix(provider/gce): Avoid blackholing Exceptions from CachingAgents.

jtk54:
Second commit is the new stuff, first is the re-revert.

### DIFF
--- a/clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/batch/GoogleBatchRequest.java
+++ b/clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/batch/GoogleBatchRequest.java
@@ -27,10 +27,11 @@ import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.extern.slf4j.Slf4j;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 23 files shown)


clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/batch/GoogleBatchRequest.java
```diff
@@ -0,0 +1,124 @@
...
+          qr.getRequest().queue(newBatch, qr.getCallback());
+        } catch (IOException ioe) {
+          log.error("Queueing request {} in batch failed.", qr);
+          throw new RuntimeException(ioe);
+        }
+      });
+      queuedBatches.add(newBatch);
+    });
+
+    ExecutorService threadPool = new ForkJoinPool(10);
```
ezimanyi: This is from the prior PR, but is there a reason to create a new pool each time? Would it be more efficient to have a common pool (or use the default pool) in this case?

jtk54: We could but it's unclear what the performance implications would be for piping all the batches through one pool since we can't tell the size of the batches beforehand when deciding the pool parallelism. We _do_ have an understanding of rough sizes per batch, and which batch sizes are considered large. I'm inclined to keep this as is and tune it later if there are problems.

PR #3159, opened 2018-11-14
fix(provider/gce): Set device name on attached disks.

jtk54:
This solves an issue where deleting MIGs with multiple disks would not delete the additional disks, leaving them orphaned.


lwander:

PR #3160, opened 2018-11-14
fix(provider/oracle): fix missing oci-sdk jars in web startScript

shihchang:
Becasue the oci-sdk jars are not published in maven (https://github.com/oracle/oci-java-sdk/issues/25), the dependency are 'dynamically' declared during fetching and unpacking the oci-sdk. If ':clouddriver-web:startScripts' runs before ':clouddriver-oracle:unpackSdk' finished, the oci-sdk jars then could be missing in the generated script. The fix adds dependsOn and mustRunAfter ':clouddriver-oracle:unpackSdk' to clouddriver-web:startScripts


shihchang:
@spinnaker/reviewers

related PRs:
https://github.com/spinnaker/front50/pull/397

dibyom:
Could you add a condition so that the `dependsOn` and `mustRunAfter` are only added if the `clouddriver-orcale` project is included in the build? Otherwise, this will fail if we do a modular build that does not include the Oracle provider e.g. `./gradlew -PincludeCloudProviders=aws installDist`

shihchang:
> Could you add a condition so that the `dependsOn` and `mustRunAfter` are only added if the `clouddriver-orcale` project is included in the build? Otherwise, this will fail if we do a modular build that does not include the Oracle provider e.g. `./gradlew -PincludeCloudProviders=aws installDist`
Made the 'dependsOn' and 'mustRunAfter' conditional as suggested.

PR #3169, opened 2018-11-18
fix(gce): TerminateGoogleInstancesDescription is not ServerGroupNameable

ewiseblatt:
@ttomsu 

Not sure if this is right or not, however this fixes google_kato_test terminate_instances.
Since https://github.com/spinnaker/clouddriver/commit/85a4901008322c9fba47065437b056d8cee2b0d2 the test was failing from a null pointer error. Note that test is deleting instances that are not bound to a server group.

It could be the bug is somewhere in the interface assuming that in fact there is a server group (i.e. server group is optional for this operation). I am not familiar with the semantics or implications here so am speculating with this and counting on your review.

See out of band email for more information about the failure details.


ttomsu:
Good call. I think I remember @duftler telling me long ago that we need to support killing individual instances that are not part of a server group, and I didn't think about that when putting the initial PR together.

PR #3174, opened 2018-11-20
fix(provider/cf): correct pagination logic when iterating over multiple pages

Jammy-Louie:
Pagination logic was not iterating over the last page when retrieving results.

spinnaker/spinnaker#3625

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryClientUtils.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryClientUtils.java
@@ -55,7 +55,7 @@ final class CloudFoundryClientUtils {
       .orElseThrow(() -> new CloudFoundryApiException("Unable to retrieve " + resourceNamePluralized));
 
     List<R> allResources = new ArrayList<>(firstPage.getResources());
-    for (int page = 2; page < Math.max(2, firstPage.getPagination().getTotalPages()); page++) {
+    for (int page = 2; page <= Math.max(2, firstPage.getPagination().getTotalPages()); page++) {
       final int p = page;
       allResources.addAll(safelyCall(() -> fetchPage.apply(p))
         .orElseThrow(() -> new CloudFoundryApiException("Unable to retrieve " + resourceNamePluralized))
@@ -70,7 +70,7 @@ final class CloudFoundryClientUtils {
       .orElseThrow(() -> new CloudFoundryApiException("Unable to retrieve " + resourceNamePluralized));
 
     List<Resource<R>> allResources = new ArrayList<>(firstPage.getResources());
-    for (int page = 2; page < Math.max(2, firstPage.getTotalPages()); page++) {
+    for (int page = 2; page <= Math.max(2, firstPage.getTotalPages()); page++) {
       final int p = page;
       allResources.addAll(safelyCall(() -> fetchPage.apply(p))
         .orElseThrow(() -> new CloudFoundryApiException("Unable to retrieve " + resourceNamePluralized))
--- a/clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryClientUtilsTest.java
+++ b/clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryClientUtilsTest.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2018 Pivotal, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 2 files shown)


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryClientUtils.java
```diff
@@ -55,7 +55,7 @@ static void safelyCall(RetrofitConsumer r) throws CloudFoundryApiException {
       .orElseThrow(() -> new CloudFoundryApiException("Unable to retrieve " + resourceNamePluralized));
 
     List<R> allResources = new ArrayList<>(firstPage.getResources());
-    for (int page = 2; page < Math.max(2, firstPage.getPagination().getTotalPages()); page++) {
+    for (int page = 2; page <= Math.max(2, firstPage.getPagination().getTotalPages()); page++) {
       final int p = page;
       allResources.addAll(safelyCall(() -> fetchPage.apply(p))
         .orElseThrow(() -> new CloudFoundryApiException("Unable to retrieve " + resourceNamePluralized))
```
claymccoy: I see very similar logic on line 73 below.
Does it suffer from the same error and require a "<=" too?
If so, then isn't there a way to avoid repeating the logic so that there aren't multiple places to update in the first place?

clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryClientUtilsTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.*;
+
+class CloudFoundryClientUtilsTest {
+
+  @Test
+  void collectPagesIteratesOverMultiplePages(){
+    ApplicationService applicationService = mock(ApplicationService.class);
+    Application applicationOne =  new Application().setName("app-name-one").setGuid("app-guid-one").setState("STARTED").setCreatedAt(ZonedDateTime.now()).setLinks(Collections.emptyMap());
```
claymccoy: Line length here is a bit long in a few places.
Are all those attributes on Application even necessary?

clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryClientUtilsTest.java
```diff
@@ -0,0 +1,55 @@
...
+    Pagination<Application> pageOne = new Pagination<>();
+    pageOne.setPagination(new Pagination.Details().setTotalPages(2));
+    pageOne.setResources(pageOneResources);
+    Application applicationTwo =  new Application().setName("app-name-two").setGuid("app-guid-two").setState("STARTED").setCreatedAt(ZonedDateTime.now()).setLinks(Collections.emptyMap());
+    List pageTwoResources = Collections.singletonList( applicationTwo );
+    Pagination<Application> pageTwo = new Pagination<>();
+    pageTwo.setPagination(new Pagination.Details().setTotalPages(2));
+    pageTwo.setResources(pageTwoResources);
+
+    when(applicationService.all(null, null,null)).thenReturn(pageOne);
```
claymccoy: space before the last null on both method calls to `all`

clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryClientUtilsTest.java
```diff
@@ -0,0 +1,55 @@
...
+    when(applicationService.all(2, null,null)).thenReturn(pageTwo);
+
+    List results = CloudFoundryClientUtils.collectPages("applications", page -> applicationService.all(page, null, null));
+
+    assertThat(results.size()).isEqualTo(2);
+    assertThat(results).contains(applicationOne, applicationTwo);
```
claymccoy: you should be able to do just one assert for equality on the list

jkschneider: specifically, call `containsExactly` instead of `contains` and remove the size assertion.

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryClientUtils.java
Jammy-Louie: I've added the logic and an additional test for line 73.

you are correct that we could try to refactor that logic to avoid repeating the code. 
I'm going to look at that and do a separate PR to remove the duplicate logic next. I think it should be a separate PR because a change would require changes in our model and how `page` and `pagination` object are handled.

PR #3182, opened 2018-11-27
feat(scattergather): Coroutine-backed scatter gather strategy

robzienert:
Adds a coroutine-backed scatter/gather implementation.

Two main benefits of this one over the current naive implementation:

1. Timeouts will cancel any unfinished, in-flight work.
2. Requests are performed in parallel.

The parallelism of requests is limited by the size of the OkHttp connection pool.

This also has some extra noise because I made the `ScatteredOkHttpCallFactory` an interface. I was initially going to mock everything out, but realized that'd be more brittle than an embedded server.

### DIFF
--- a/clouddriver-scattergather/src/test/kotlin/com/netflix/spinnaker/clouddriver/scattergather/coroutine/CoroutineScatterGatherSpec.kt
+++ b/clouddriver-scattergather/src/test/kotlin/com/netflix/spinnaker/clouddriver/scattergather/coroutine/CoroutineScatterGatherSpec.kt
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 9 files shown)


clouddriver-scattergather/src/test/kotlin/com/netflix/spinnaker/clouddriver/scattergather/coroutine/CoroutineScatterGatherSpec.kt
```diff
@@ -0,0 +1,75 @@
...
+        expectThat(result) {
+          get { status }.isEqualTo(200)
+          get { body!! }.contains("two")
+          get { body!! }.contains("one")
+        }
+      }
+    }
+
+    afterGroup {
+      server.shutdown()
```
robzienert: One unfortunate thing, the mock server hangs for like 10-15 seconds while shutting down. I couldn't find a way to make this faster. Wondering if it'd be worthwhile marking it ignored if we're running in travis or something?

robzienert: I was informed out-of-band that I should give Wiremock a go, so I'll tackle that tomorrow between meetings 1 and 56.

robzienert:
Closing for the time being.

PR #3189, opened 2018-11-28
fix(provider/cf): correct pagination logic when iterating

Jammy-Louie:
Fix issue when collecting results from a single page and add a test for the scenario of a single page of results.

spinnaker/spinnaker#3662

### DIFF
--- a/clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryClientUtilsTest.java
+++ b/clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryClientUtilsTest.java
@@ -34,6 +34,22 @@ import static org.mockito.Mockito.when;
 class CloudFoundryClientUtilsTest {
 
   @Test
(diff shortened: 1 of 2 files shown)


clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryClientUtilsTest.java
```diff
@@ -33,6 +33,22 @@
 
 class CloudFoundryClientUtilsTest {
 
+  @Test
+  void collectPagesIteratesOveeOnePage() {
```
claymccoy: Over is misspelled.

Jammy-Louie: 👍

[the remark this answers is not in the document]
Jammy-Louie: yup already changed as mentioned above

PR #3190, opened 2018-11-28
fix(provider/kubernetes): hold entries in on-demand longer

lwander:
This should help prevent the force cache refresh task from having to query the on demand cache as often, as processed entries will be kept for an additional cycle (30s)

@benjaminws @davidxia, I think this should speed up executions that seem hung on ForceCacheRefresh tasks


davidxia:
Yay, thanks!

benjaminws:
Nice, going to test this in my staging env now.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2OnDemandCachingAgent.java
```diff
@@ -97,7 +97,7 @@ public CacheResult loadData(ProviderCache providerCache) {
     try {
       primaryResource = loadPrimaryResourceList();
     } catch (KubectlJobExecutor.NoResourceTypeException e) {
-      log.warn(getAgentType() + ": resource for this caching agent is not supported for this cluster");
+      log.error(getAgentType() + ": resource for this caching agent is not supported for this cluster. This will cause problems, please remove it from caching using the `omitKinds` config parameter.");
```
jtk54: Logging `error` and then not failing after seems strange, but I guess is unavoidable given one caching agent. Is there an additional way we can alert users without relying on tribal knowledge of 'force cache refresh hanging -> look in logs'?

PR #3196, opened 2018-11-30
fix(provider/aws): Only describe instance health for ELBs (#2699)

dmart:
fixes https://github.com/spinnaker/spinnaker/issues/2699


robzienert:
Reason we didn't hit this at Netflix is because we proxy all of our calls through Edda, which no-ops the call. Whoopsies.

staciegravy:

staciegravy:

ezimanyi:

PR #3198, opened 2018-12-03
fix(titus): fix when job data is cached apart from instances

asher:


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
```diff
@@ -295,9 +295,10 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
...
 
       Job job
-      if (instanceEntry.attributes.job != null) {
+      if (instanceEntry.attributes.job != null || (instanceEntry.attributes.jobId != null &&
+        jobData.containsKey(instanceEntry.attributes.jobId))) {
         if (instanceEntry.relationships[SERVER_GROUPS.ns]
-            && !instanceEntry.relationships[SERVER_GROUPS.ns].empty) {
+          && !instanceEntry.relationships[SERVER_GROUPS.ns].empty) {
           // job needs to be loaded because it was cached separately
           job = jobData.get(instanceEntry.attributes.jobId)
```
asher: The titusV2 provider caches job data at the serverGroup but not instance level. This code path either maps an instance jobId back to the job data cached with the serverGroup, or reads it directly from the instance if present. Loading the job object from the serverGroup data shouldn't be dependent on an instance level job object.

PR #3199, opened 2018-12-03
fix(cache/introspection): sort by descending execution

lwander:
This makes the results a lot easier to read, as agents that are executed infrequently/on different nodes get pushed to the bottom


lwander:

PR #3201, opened 2018-12-03
perf(provider/kubernetes): share CRD list between threads

lwander:
@gorebala on slack did a lot of research into the k8s force cache
refresh times and found that each thread was updating the list of CRDs
using a live call. This change moves the refresh logic into a shared
parent thread (where the credentials are created), meaning the list is
shared by call caching agents, which was the original intent behind
using a memoizing supplier.


lwander:
Playing with this locally, on 1 clouddriver node with 20 caching agents, the POST /cache/kubernetes/manifest response time went down from an average of 8 seconds down to 1.5

lwander:

PR #3206, opened 2018-12-05
fix(security): Fix CVE-2017-5929 security issue threat level 9

jingsfdc:
https://nvd.nist.gov/vuln/detail/CVE-2017-5929

QOS.ch Logback before 1.2.0 has a serialization vulnerability affecting the SocketServer and ServerSocketReceiver components. 

The RemoteStreamAppenderClient class in logback-classic and the SocketNode classes in logback-classic and logback-access allow data to be deserialized over a Java Socket, via an ObjectInputStream, without validating the data beforehand. When data is received from the Socket, to be logged, it is deserialized into Java objects. An attacker can exploit this vulnerability by sending malicious, serialized Java objects over the connection to the Socket, which may result in execution of arbitrary code when those objects are deserialized.Note that although logback-core is implicated by the Logback project here, the Sonatype Security Research team discovered that the vulnerability is actually present in the logback-classic and logback-access components.Advisory Deviation Notice: The Sonatype Security Research team discovered that this vulnerability was first introduced in version 0.3.0 of logback-classic and version 0.6.0 of logback-access, as opposed to being present in all versions prior to 1.2.0, as stated in the advisory.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,6 @@
 /*
  * Copyright 2014 Netflix, Inc.
+ * Copyright (c) 2018, salesforce.com, inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -111,6 +112,9 @@ subprojects { project ->
           }
           if (it.requested.group == 'org.apache.xbean') {
             it.useVersion '4.3'
+          }
+          if (it.requested.group == 'ch.qos.logback') {
+            it.useVersion '1.2.3'
           }
         }
       }


build.gradle
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright 2014 Netflix, Inc.
+ * Copyright (c) 2018, salesforce.com, inc. All rights reserved.
```
lwander: Is this a requirement from your legal department? If it's not, the convention is to leave the headers pointing at the original authors.

lwander:
Thanks for this! For future patches like this, can you link to where the CVE was reported (which CVE repo)?

build.gradle
jingsfdc: @lwander yes, it's a requirement from the legal department.

jingsfdc:
@lwander I have updated the description with the link to the CVE repo. Thanks for the feedback.

lwander:

build.gradle
jamesward: Typically a CLA is used to transfer the copyright but it doesn't appear that Spinnaker has a CLA.  @lwander can you confirm?

lwander: Yes that's right, there is no CLA at this time

PR #3207, opened 2018-12-05
fix(provider/kubernetes): fix v2 cronjob status

lwander:
Currently Kubernetes only reports if a job scheduled by a cronjob has
completed without a success status. By checking if a job was running we
were only preventing potentially functioning cronjob specs from becoming
"stable", without correctly filtering out "unstable" cronjobs.


lwander:

PR #3215, opened 2018-12-10
feat(provider/cf): add SG clone action

jmelchio:
Co-Authored-By: Joris Melchior <joris.melchior@gmail.com>
Co-Authored-By: Stu Pollock <spollock@pivotal.io>

spinnaker/spinnaker#3656

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/converters/CloneCloudFoundryServerGroupAtomicOperationConverter.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/converters/CloneCloudFoundryServerGroupAtomicOperationConverter.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2018 Pivotal, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.cloudfoundry.deploy.converters;
+
+import com.netflix.spinnaker.clouddriver.artifacts.ArtifactCredentialsRepository;
+import com.netflix.spinnaker.clouddriver.cloudfoundry.CloudFoundryOperation;
+import com.netflix.spinnaker.clouddriver.cloudfoundry.provider.view.CloudFoundryClusterProvider;
+import com.netflix.spinnaker.clouddriver.helpers.OperationPoller;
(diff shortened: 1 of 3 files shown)


jmelchio:
Needs to go in sync with: https://github.com/spinnaker/orca/pull/2551

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/converters/DeployCloudFoundryServerGroupAtomicOperationConverter.java
```diff
@@ -85,7 +85,7 @@ public DeployCloudFoundryServerGroupDescription convertDescription(Map input) {
     converted.setSpace(findSpace(converted.getRegion(), converted.getClient())
       .orElseThrow(() -> new IllegalArgumentException("Unable to find space '" + converted.getRegion() + "'.")));
 
-    Map artifactSource = (Map) input.get("artifactSource");
+    Map artifactSource = (Map) input.get("artifact");
```
jkschneider: Can you elaborate on why the property name changed?

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/converters/CloneCloudFoundryServerGroupAtomicOperationConverter.java
```diff
@@ -0,0 +1,35 @@
+/*
```
jkschneider: Why is a separate operation necessary if it extends from and doesn't modify the behavior of `DeployCloudFoundryServerGroupAtomicOperationConverter`?

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/converters/DeployCloudFoundryServerGroupAtomicOperationConverter.java
jmelchio: The mapping from `artifact` to `artifactSource` happens in Orca. For the create server group operation we have control over that so it works fine, when sending the clone operation we don't have an easy extension point in Orca so we considered it easier to adjust this than to try and implement a custom operation for clone in Orca.

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/converters/CloneCloudFoundryServerGroupAtomicOperationConverter.java
jmelchio: Core Deck determines that you execute a clone action and puts that operation string on it. As a result you have to have a handler for it on the receiving end in clouddriver.

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/converters/DeployCloudFoundryServerGroupAtomicOperationConverter.java
jkschneider: Thanks @jmelchio!

PR #3219, opened 2018-12-11
fix(core): Make uses of snakeyaml threadsafe

ezimanyi:
Also change constructor in CloudFoundry provider to use SafeConstructor to prevent code injection.


jkschneider:
Oh, very interesting. TIL.

PR #3222, opened 2018-12-13
fix(ecs): support multiple network binds for non-network-interfaced containers

gmccue:
Solves https://github.com/spinnaker/spinnaker/issues/3724

@clareliguori @BrunoCarrier 

Happy to add a test case for this, but will need a bit of direction there.


clareliguori:
Thanks for jumping on this!

re: unit tests, I have started breaking up the tests for that file in https://github.com/spinnaker/clouddriver/pull/3220.  You'll want to add a test similar to "should get a list of task health", but with multiple network bindings

cc @iniinikoski who mentioned he can test locally too

clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgentSpec.groovy
```diff
@@ -114,6 +115,75 @@ class TaskHealthCachingAgentSpec extends Specification {
     taskHealth.getTaskId() == CommonCachingAgent.TASK_ID_1
   }
 
+  def 'should get a list of task health for tasks with multiple network bindings'() {
```
clareliguori: If you merge from master (#3220 is now merged), you can reduce a lot of this copy-and-paste

gmccue: Oh excellent, didn't realized that had been merged. Will do.

PR #3223, opened 2018-12-13
fix(aws/region support): Use RegionUtils instead of Regions

gardleopard:
This is to have less coupling to constants in the aws sdk.
Clouddriver will support new regions without updating the aws sdk.

https://github.com/aws/aws-sdk-java/pull/1839


gardleopard:
This will make clouddriver support the new eu-north-1 region

PR #3224, opened 2018-12-13
fix(provider/titus): reworks streaming provider so it uses a buffer

tomaslin:
This also simplifies the update logic so that now snapshots and deltas go through the same write to cache cycle.

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
@@ -45,12 +45,15 @@ import com.netflix.spinnaker.clouddriver.titus.client.TitusRegion;
 import com.netflix.spinnaker.clouddriver.titus.client.model.TaskState;
 import com.netflix.spinnaker.clouddriver.titus.credentials.NetflixTitusCredentials;
 import com.netflix.titus.grpc.protogen.*;
+import lombok.extern.slf4j.Slf4j;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.inject.Provider;
 import java.util.*;
 import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
```diff
@@ -60,12 +61,10 @@
 import static com.netflix.spinnaker.clouddriver.titus.caching.Keys.Namespace.*;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
+@Slf4j
```
emjburns: Why did you flip to using lombok? I prefer defining the log as it was before

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
```diff
@@ -236,255 +208,84 @@ public void executeAgent(Agent agent) {
...
+        changes++;
       }
     }
 
-    private void writeToCache(CacheResult result, boolean evict, Collection<AgentDataType> providedTypes) {
-      Collection<String> authoritative = new HashSet<>(providedTypes.size());
-      for (AgentDataType type : providedTypes) {
-        if (type.getAuthority() == AgentDataType.Authority.AUTHORITATIVE) {
-          authoritative.add(type.getTypeName());
+    private void maybeWriteToCache() {
```
emjburns: Could you add a javadoc comment about when this would write and when it wouldn't?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
```diff
@@ -76,6 +75,8 @@
   private final Registry registry;
   private final Id metricId;
   private final Provider<AwsLookupUtil> awsLookupUtil;
+  private final long TIME_UPDATE_THRESHHOLD_MS = 5000;
+  private final long ITEMS_CHANGED_THRESHHOLD = 1000;
```
robzienert: Since these are going to be modified by a different thread, should we be using `AtomicLong` instead?

tomaslin: they don't get modified

robzienert: Whoops, I meant to comment this on this part: https://github.com/spinnaker/clouddriver/pull/3224/files#diff-ba906249bad8b13af34d50731bc17b94R162

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
tomaslin: to be consistent with the other codes

tomaslin:
Updated to address comments from reviews

PR #3227, opened 2018-12-14
feat(provider/kubernetes): 'liveManifestCall' mode

lwander:
Been meaning to submit this for us to try out:

This is in service of faster pipeline executions, which ignore force
cache refreshes, and only serve manifest status from live calls.

Need to find a cleaner top-level way to distinguish between object
providers that query the cache vs. live data.

Important to note: this leaves caching (& the clusters tab) intact for
now, those can be disabled separately in a future PR.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesNamedAccountCredentialsInitializer.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesNamedAccountCredentialsInitializer.groovy
@@ -114,6 +114,7 @@ class KubernetesNamedAccountCredentialsInitializer implements CredentialsInitial
           .checkPermissionsOnStartup(managedAccount.checkPermissionsOnStartup == null ? true : managedAccount.checkPermissionsOnStartup)
           .kubernetesSpinnakerKindMap(kubernetesSpinnakerKindMap)
           .onlySpinnakerManaged(managedAccount.onlySpinnakerManaged == null ? false : managedAccount.onlySpinnakerManaged)
+          .liveManifestCalls(managedAccount.liveManifestCalls ?: false)
           .build()
 
         accountCredentialsRepository.save(managedAccount.name, kubernetesAccount)
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2LiveManifestProvider.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2LiveManifestProvider.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2018 Google, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.view.provider;
+
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.view.model.KubernetesV2Manifest;
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.KubernetesResourcePropertyRegistry;
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.KubernetesKind;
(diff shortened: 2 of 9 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesNamedAccountCredentialsInitializer.groovy
```diff
@@ -114,6 +114,7 @@ class KubernetesNamedAccountCredentialsInitializer implements CredentialsInitial
           .checkPermissionsOnStartup(managedAccount.checkPermissionsOnStartup == null ? true : managedAccount.checkPermissionsOnStartup)
           .kubernetesSpinnakerKindMap(kubernetesSpinnakerKindMap)
           .onlySpinnakerManaged(managedAccount.onlySpinnakerManaged == null ? false : managedAccount.onlySpinnakerManaged)
+          .liveManifestCalls(managedAccount.liveManifestCalls == null ? false : managedAccount.liveManifestCalls)
```
jtk54: Could probably use an elvis here.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2LiveManifestProvider.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    Pair<KubernetesKind, String> parsedName;
+    try {
+      parsedName = KubernetesManifest.fromFullResourceName(name);
+    } catch (Exception e) {
+      return null;
+    }
+
+    log.info("Live call to lookup manifest '{}:{}' in namespace '{}' under account '{}'", parsedName.getRight(), parsedName.getLeft(), location, account);
```
jtk54: Might want to move this to debug?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesNamedAccountCredentialsInitializer.groovy
lwander: True, groovy

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2LiveManifestProvider.java
lwander: Sorry that's intentional -- it's helpful from a setup & early user perspective to see where the extra load on Kubernetes will come from, esp. with regular caching still running. I'm going to drop a TODO to move this to DEBUG in a future release.

johnmshields:
@lwander I can't find a normal issue that corresponds to this PR so I'm just commenting here...

We're seeing very long force cache refresh times so we're looking forward to trying this PR out. However, I want to understand how this interacts with caching. I think the snippet from one of our clouddriver instances below probably clarifies why I'm curious:

```bash
bash-4.4$ time kubectl --kubeconfig=/home/spinnaker/.hal/default/staging/dependencies/2054179555-config --context=<our context> --namespace=default -o json get apiService,clusterRole,clusterRoleBinding,configMap,controllerRevision,customResourceDefinition,cronJob,daemonSet,deployment,event,horizontalpodautoscaler,ingress,job,namespace,networkPolicy,persistentVolume,persistentVolumeClaim,pod,podSecurityPolicy,podDisruptionBudget,replicaSet,role,roleBinding,secret,service,serviceAccount,statefulSet,storageClass > /tmp/core-cache.json 2> /tmp/core-cache.err

real	3m12.315s
user	3m6.377s
sys	0m2.214s
bash-4.4$ du -h /tmp/core-cache.json 
20.1M	/tmp/core-cache.json
bash-4.4$ cat /tmp/core-cache.err 
Error from server (Expired): The provided from parameter is too old to display a consistent list result. You must start a new list without the from.
```

This is probably one of the reasons we are seeing such issues with caching. We have multiple large clusters with hundreds of pods. As you can see the JSON output is 20MB and takes 3m. Also, the k8s server doesn't seem to like this particular query. :)

Is there a way to tell the caching agents to split up these `kubectl` calls between types or otherwise configure it to perform better? In our case, I think there must be something in this list that is causing it to be very slow.

(I tried using `--chunk-size` on the command line to test and that didn't seem to help in our situation. Only limiting the kinds list seems to help.)

lwander:
Hey @johnmshields 

TL;DR the live manifest call mode is written exactly to fix what you're seeing here for pipeline executions only.

---

@benjaminws might be able to chime in & say what they did for those huge payloads, since they were running into very similar issues.

Caching does two things:

1. It requests all data, sharded across namespaces, at a configurable interval. It used to be sharded by type as well, but the total number of `kubectl` invocations were a huge issue and made Spinnaker nearly unrunnable. 
2. It updates a single resource during any manifest-oriented stage (e.g. deploy, scale, ...) to ensure that stability checks are running against up-to-date resources. This requires that 'force cache refresh' to ensure these results are merged correctly with the results in 1. If either 1 or 2 keep failing, this 'force cache refresh' will keep retrying for a long while (up to 12 minutes).

The live manifest call mode only replaces 2. Even if 1 is failing for the reasons seen above, pipelines will still be able to query resource status, without having to refresh or coordinate with the state of the cache.

PR #3229, opened 2018-12-14
feat(entitytags): a simple `deleteByNamespace` admin api

ajordens:
Dry run (default):
```
curl -X POST 'https://localhost:8081/admin/tags/deleteByNamespace/{namespace}'
```

Remove from elasticsearch only:
```
curl -X POST 'https://localhost:8081/admin/tags/deleteByNamespace/{namespace}?dryRun=false'
```

Remove from elasticsearch and front50:
```
curl -X POST 'https://localhost:8081/admin/tags/deleteByNamespace/{namespace}?dryRun=false&deleteFromSource=true'
```

### DIFF
--- a/clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
+++ b/clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
@@ -417,6 +417,58 @@ public class ElasticSearchEntityTagsProvider implements EntityTagsProvider {
     return getElasticSearchEntityTagsReconciler().reconcile(this, cloudProvider, account, region, dryRun);
   }
 
+  @Override
+  public Map<String, Object> deleteByNamespace(String namespace, boolean dryRun, boolean deleteFromSource) {
+    List<EntityTags> entityTagsForNamespace = front50Service
+      .getAllEntityTags(false)
+      .stream()
+      .filter(e -> e.getTags().stream().anyMatch(t -> namespace.equalsIgnoreCase(t.getNamespace())))
+      .collect(Collectors.toList());
+
+    for (EntityTags entityTags : entityTagsForNamespace) {
+      // ensure that all tags (and their metadata) in the offending namespace are removed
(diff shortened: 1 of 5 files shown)


ajordens:
Specifically opt'd not to make this an exposed API as I'm not quite sure how easy we'd like it to be to remove a whole lot of tags at one time. 

The logic is pretty straight forward, might warrant a unit test.

clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
```diff
@@ -417,6 +417,57 @@ public Map reconcile(String cloudProvider, String account, String region, boolea
     return getElasticSearchEntityTagsReconciler().reconcile(this, cloudProvider, account, region, dryRun);
   }
 
+  @Override
+  public Map<String, Object> deleteByNamespace(String namespace, boolean dryRun, boolean deleteFromSource) {
+    List<EntityTags> entityTagsForNamespace = front50Service
+      .getAllEntityTags(false)
```
ajordens: This just fetches the set of entity tags that front50 has in it's in-memory cache ... should be reasonable quick but the response payload is quite large (could be optimized in the future).

clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
```diff
@@ -417,6 +417,58 @@ public Map reconcile(String cloudProvider, String account, String region, boolea
...
+          .getTags()
+          .stream()
+          .filter(e -> !namespace.equalsIgnoreCase(e.getNamespace()))
+          .collect(Collectors.toList())
+      );
+
+      Set<String> tagNames = entityTags
+        .getTags()
+        .stream()
+        .map(e -> e.getName().toLowerCase())
```
christopherthielen: what is the reason for lowercase compare of the tag names?

christopherthielen:
thanks for doing this!

PR #3231, opened 2018-12-17
fix(provider/ecs): check for null service before getTaskDefinition

gmccue:
Fixes https://github.com/spinnaker/spinnaker/issues/3749

@clareliguori


tomaslin:
please change the commit message per spinnaker bot

PR #3233, opened 2018-12-17
feat(ecs): Migrate to standard server group name resolver

clareliguori:
This makes it slightly easier to implement https://github.com/spinnaker/spinnaker/issues/3733, by standardizing on the same logic that determines the 'next' and 'current' server groups.

This does change the format of the server group names, from basename-v0001 to basename-v001.  It's not a breaking change because the sequence number parses the same in both formats.


tomaslin:
LGTM

PR #3237, opened 2018-12-20
fix(provider/docker): No longer depend on service in authentication

abergmeier:
Should fix #3772

Depending on the structure of the service URIs, there may be no Service in
the Authentication (see https://tools.ietf.org/html/rfc6750#section-3).
This is actually true for OpenShift OKR structure.
As such, ignore missing `service` and ensure that it gets ignored later on
as query parameter.


abergmeier:
Now with tests to ensure this will continue to work.

PR #3241, opened 2018-12-21
fix(provider/kubernetes): bail out of caching on error

lwander:
By returning `null` we were deleting valid (but stale) cache entries
instead of forcing the caching agent to retry later.


lwander:

PR #3244, opened 2019-01-02
fix(core): include instanceCounts when fetching project clusters, fix…

maggieneterval:
… ClassCastException for gce and appengine fetch cluster methods

While investigating https://github.com/spinnaker/spinnaker/issues/3490, I noticed that project clusters were also failing to populate for GCE and Appengine projects because ProjectClusterService.retrieveClusters was throwing the following error: `java.lang.ClassCastException: java.util.ArrayList cannot be cast to java.util.Set`. Additionally, `instanceCounts` was missing from the model for all providers.

(Will open a separate PR once I resolve the k8s issue in https://github.com/spinnaker/spinnaker/issues/3490)

Before:

clusters_before

After:

clusters_after


maggieneterval:
:+1: will do!

maggieneterval:

PR #3247, opened 2019-01-02
fix(kubernetes): fix fetching clusters for project

maggieneterval:
Previously, `KubernetesV2ClusterProvider.getClusterSummaries` was not returning any cluster data for a given application. When calling `KubernetesCacheUtils.getTransitiveRelationship` to identify applications-to-clusters relationships, no relationships were picked up -- I am still trying to figure out the root cause, but adjusting `getClusterSummaries` to instead use `cacheUtils.getAllDataMatchingPattern` resolves the issue. The other changes are to address NPEs in ProjectClustersService because it appears that for k8s clusters, cluster -> moniker -> app and cluster -> serverGroup -> imagesSummary may be null.

Before:

k8s-before

After:

k8s-after


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2ClusterProvider.java
```diff
@@ -75,11 +74,10 @@
 
   @Override
   public Map<String, Set<KubernetesV2Cluster>> getClusterSummaries(String application) {
-    String applicationKey = Keys.application(application);
+    String clusterGlobKey = Keys.cluster("*", application, "*");
```
lwander: This will likely be much slower, especially for large applications.

lwander:
It sounds like the root cause is that cluster keys aren't being written into the cache with a relationship pointing back at the applications they were found under.

maggieneterval:
Ah thanks @lwander, that would make perfect sense -- might ask you to point me in the right direction tomorrow :)

maggieneterval:
This should be ready for a second look -- thanks for the help @lwander!

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesCacheDataConverter.java
```diff
@@ -371,6 +371,28 @@ static void addSingleRelationship(Map<String, Collection<String>> relationships,
     return result;
   }
 
+  static List<CacheData> getClusterRelationships(String account, CacheData cacheData) {
```
lwander: Can this list ever have more than 1 entry? If not, I would use a `Optional` here instead.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2Cluster.java
```diff
@@ -38,12 +38,14 @@
...
   Set<LoadBalancer> loadBalancers = new HashSet<>();
+  String application;
 
   public KubernetesV2Cluster(String rawKey) {
     Keys.ClusterCacheKey key = (Keys.ClusterCacheKey) Keys.parseKey(rawKey).get();
     this.name = key.getName();
     this.accountName = key.getAccount();
-    this.moniker = Moniker.builder().cluster(name).build(); // TODO(lwander) if it turns out that cluster -> app is important, enforce constraints here.
+    this.application = key.getApplication();
+    this.moniker = Moniker.builder().cluster(name).app(application).build(); // TODO(lwander) if it turns out that cluster -> app is important, enforce constraints here.
```
lwander: Looks like` cluster -> app` might become important :)

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesCacheDataConverter.java
lwander: I know it makes the stream below a feel little nicer with the flatmap -- but I think for the sake of consistency having this behave more like the above `convertAs*` methods (which are nullable rather than optional, that works too) will make this more readable/understandable in the future I think.

maggieneterval: Makes sense, updated!

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2Cluster.java
maggieneterval: Haha so prescient of you! Lmk if there are any constraints in particular you'd like me to add with this change.

maggieneterval:
This appears to be behaving as expected even when clusters are shared between applications!

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2Cluster.java
lwander: No I think you solved them by allowing the clusters to belong to multiple projects. You can delete the TODO

PR #3248, opened 2019-01-03
fix(provider/kubernetes): v2 handle 0 replica scale-down

lwander:
https://github.com/spinnaker/spinnaker/issues/3804


lwander:

PR #3255, opened 2019-01-04
fix(artifacts): fixed property prefix for maven artifacts

jkschneider:
* Simplified ArtifactCredentialsRepository to keep a list rather than a map
* Set account name when generating IvyArtifactAccount from MavenArtifactAccount


claymccoy:
LGTM
Nice simplification for the impl.

PR #3257, opened 2019-01-07
fix(cf): repair scale-to-zero issue

Jammy-Louie:
If the request is to scale down to zero, then we should treat the DOWN status as a success.

spinnaker/spinnaker#3670

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/ops/ScaleCloudFoundryServerGroupAtomicOperation.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/ops/ScaleCloudFoundryServerGroupAtomicOperation.java
@@ -27,6 +27,7 @@ import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation;
 import lombok.RequiredArgsConstructor;
 
 import java.util.List;
+import java.util.Optional;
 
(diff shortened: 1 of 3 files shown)


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/ops/ScaleCloudFoundryServerGroupAtomicOperation.java
```diff
@@ -56,17 +57,20 @@ public Void operate(List priorOutputs) {
...
-      inProgressState -> (inProgressState == ProcessStats.State.RUNNING || inProgressState == ProcessStats.State.CRASHED),
+      inProgressState -> (
+        inProgressState == ProcessStats.State.RUNNING
+          || inProgressState == ProcessStats.State.CRASHED
+          || inProgressState == ProcessStats.State.DOWN),
       null, getTask(), description.getServerGroupName(), PHASE);
 
-    if (state != ProcessStats.State.RUNNING) {
+    if (state == ProcessStats.State.RUNNING || (state == ProcessStats.State.DOWN &&
+      Optional.ofNullable(description.getCapacity()).map(cap -> cap.getDesired() == 0).orElse(false))) {
```
claymccoy: What is the case where capacity can be null?

Jammy-Louie: We took a closer look and Capacity should not be null because it will either be provided from `Orca` or specified from the user in `Deck`

PR #3260, opened 2019-01-08
fix(provider/gce): Adds flexibility to autoscaler upsert for scaleDown

jtk54:
tasks.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/converters/ResizeGoogleServerGroupAtomicOperationConverter.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/converters/ResizeGoogleServerGroupAtomicOperationConverter.groovy
@@ -65,6 +65,15 @@ class ResizeGoogleServerGroupAtomicOperationConverter extends AbstractAtomicOper
       upsertGoogleAutoscalingPolicyDescription.autoscalingPolicy = googleAutoscalingPolicy
       upsertGoogleAutoscalingPolicyDescription.autoscalingPolicy.minNumReplicas = input.capacity?.min
       upsertGoogleAutoscalingPolicyDescription.autoscalingPolicy.maxNumReplicas = input.capacity?.max
+      if (input?.writeMetadata != null) {
+        upsertGoogleAutoscalingPolicyDescription.writeMetadata = input?.writeMetadata
+      }
+
+      // Override autoscaling mode. This is useful in situations where we need the resize to happen
+      // regardless of previous autoscaling mode (e.g. scale down in red/black deployment strategies).
+      if (input?.autoscalingMode) {
+        upsertGoogleAutoscalingPolicyDescription.autoscalingPolicy.mode = input.autoscalingMode
+      }
 
       return upsertGoogleAutoscalingPolicyDescription
     } else {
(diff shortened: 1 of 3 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/UpsertGoogleAutoscalingPolicyAtomicOperation.groovy
```diff
@@ -180,18 +180,20 @@ class UpsertGoogleAutoscalingPolicyAtomicOperation extends GoogleAtomicOperation
...
-                           project,
-                           GCEUtil.buildRegionalServerGroupUrl(project, region, serverGroupName),
-                           autoscaler)
-    } else {
-      updatePolicyMetadata(compute,
-                           credentials,
-                           project,
-                           GCEUtil.buildZonalServerGroupUrl(project, zone, serverGroupName),
-                           autoscaler)
+    if (description.writeMetadata) {
```
lwander: No chance this is `null`? The boolean is nullable

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/converters/ResizeGoogleServerGroupAtomicOperationConverter.groovy
```diff
@@ -65,6 +65,15 @@ class ResizeGoogleServerGroupAtomicOperationConverter extends AbstractAtomicOper
...
       upsertGoogleAutoscalingPolicyDescription.autoscalingPolicy.minNumReplicas = input.capacity?.min
       upsertGoogleAutoscalingPolicyDescription.autoscalingPolicy.maxNumReplicas = input.capacity?.max
+      if (input?.writeMetadata != null) {
+        upsertGoogleAutoscalingPolicyDescription.writeMetadata = input?.writeMetadata
+      }
+
+      // Override autoscaling mode. This is useful in situations where we need the resize to happen
+      // regardless of previous autoscaling mode (e.g. scale down in red/black deployment strategies).
+      if (input?.autoscalingMode) {
+       upsertGoogleAutoscalingPolicyDescription.autoscalingPolicy.mode = input.autoscalingMode
```
duftler: Indentation looks off here.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/UpsertGoogleAutoscalingPolicyAtomicOperation.groovy
```diff
@@ -180,18 +180,20 @@ class UpsertGoogleAutoscalingPolicyAtomicOperation extends GoogleAtomicOperation
...
-                           project,
-                           GCEUtil.buildRegionalServerGroupUrl(project, region, serverGroupName),
-                           autoscaler)
-    } else {
-      updatePolicyMetadata(compute,
-                           credentials,
-                           project,
-                           GCEUtil.buildZonalServerGroupUrl(project, zone, serverGroupName),
-                           autoscaler)
+    if (description.writeMetadata == null || description.writeMetadata) {
```
duftler: What's the actual path where this can end up `null`? Looks like the converter only overwrites the default `True` if a non-null value is passed in the request input.

Also, if it can in fact end up `null`, why does that mean we should update the metadata?

jtk54: This will be null if a user explicitly issues an upsert description with `writeMetadata = null`. The converter change in this PR was for resize operations. Not writing metadata is the exception and should be explicitly requested with `writeMetadata = false`, so in all other cases we write metadata.

jtk54:
@spinnakerbot cherry-pick 1.10 1.11

jtk54:

PR #3270, opened 2019-01-11
fix(provider/kubernetes): sync updates to account properties

lwander:
This likely doesn't fix the issue where an otherwise healthy clouddriver
node stops being able to deploy registered CRDs, but does potentially
address a startup bug. However, unless spring boot is doing something
unexpected, access to the now syncronized method should never happen across
more than a single thread.

https://github.com/spinnaker/spinnaker/issues/3439


lwander:

PR #3274, opened 2019-01-11
fix(*): Consolidate default ObjectMapper mutations

robzienert:
This thing makes things more better when you disable Redis... such as not breaking all REST endpoints.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/CloudDriverConfig.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/CloudDriverConfig.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.clouddriver.config
 
+import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.cats.agent.ExecutionInstrumentation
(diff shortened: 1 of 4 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/CloudDriverConfig.groovy
```diff
@@ -81,6 +119,17 @@ class CloudDriverConfig {
     Clock.systemDefaultZone()
   }
 
+  /**
+   * rz - This is a consolidation of mutations made to the default spring objectmapper.
+   * TODO(rz): Remove Autowiring of ObjectMappers from the project entirely.
+   */
+  @Autowired
+  ObjectMapper objectMapper(ObjectMapper objectMapper) {
```
cfieber: was just looking at this:

https://docs.spring.io/spring-boot/docs/current/reference/html/howto-spring-mvc.html#howto-customize-the-jackson-objectmapper

In theory you could replace this bit with some properties in the default environment

fail_on_unknown_properties false should actually be a default

setting
```
spring.jackson.default-property-inclusion=non_null
spring.jackson.serialization.fail_on_empty_beans=false
```

should accomplish the same thing

not sure if that feels like too much magic or not, and the ability to get at this programmatically for tests is also nice so theres that too..

cfieber: if nothing else then maybe just defining a `Jackson2ObjectMapperBuilderCustomizer` Bean here instead of autowiring the objectMapper would be the more booty way to do this

robzienert: Thanks this is neat-o. TIL.

PR #3276, opened 2019-01-14
fix(aws): Fix concurrent list modification exception

robzienert:
`ConcurrentModificationException` while running the Titus streaming agent.


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/utils/AwsLookupUtil.groovy
```diff
@@ -118,7 +118,11 @@ class AwsLookupUtil {
   }
 
   public Map lookupAccount(account, region) {
-    Map awsDetails = awsAccountLookup.find {
+    // rz - avoid a concurrent access exception while interacting with awsAccountLookup (via Titus streaming agent)
+    List accounts = new ArrayList<>(awsAccountLookup.size())
+    accounts.addAll(awsAccountLookup)
```
robzienert: Did `addAll` cuz `Collections.copy` still sometimes threw an exception about `src` being larger than the `dest`.

PR #3282, opened 2019-01-16
fix(google): differentiate among autohealing health check kinds

maggieneterval:
Previously when querying for a health check, we returned the first health check found by name when querying the following kinds in order: healthCheck --> httpHealthCheck --> httpsHealthCheck. However, it is possible to have up to three health checks with the same name (one of each kind), so we should specify `healthCheckKind` and only query for the specified kind, leaving the default querying logic in place if `healthCheckKind` is unspecified.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
@@ -272,28 +272,28 @@ class GCEUtil {
   static def queryHealthCheck(String projectName,
                               String account,
                               String healthCheckName,
+                              GoogleHealthCheck.HealthCheckKind healthCheckKind,
                               Compute compute,
                               Cache cacheView,
                               Task task,
                               String phase,
                               GoogleExecutorTraits executor) {
-    task.updateStatus phase, "Looking up http(s) health check $healthCheckName..."
+    task.updateStatus phase, "Looking up health check $healthCheckName..."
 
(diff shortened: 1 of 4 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -272,28 +272,28 @@ class GCEUtil {
...
+      case GoogleHealthCheck.HealthCheckKind.healthCheck:
+        return queryNestedHealthCheck(projectName, account, healthCheckName, compute, cacheView, task, phase, executor)
+      case GoogleHealthCheck.HealthCheckKind.httpHealthCheck:
+        return queryLegacyHttpHealthCheck(account, healthCheckName, cacheView, task, phase)
+      case GoogleHealthCheck.HealthCheckKind.httpsHealthCheck:
+        return queryLegacyHttpsHealthCheck(projectName, healthCheckName, compute, task, phase, executor)
+      default:
+        // Note: Cache queries for these health checks must occur in this order since queryLegacyHttpsHealthCheck() will make a live
+        // call that fails on a missing health check.
+        // todo(mneterval): return null instead of querying for each type once all health check payloads include `healthCheckKind`
```
jtk54: Are the health checks missing `healthCheckKind` in incoming operations e.g. from the UI?

jtk54: Yep, SGTM.

maggieneterval:

PR #3283, opened 2019-01-16
feat(core): Add DualTaskRepository

robzienert:
Simple little class to migrate task backends without interruption. 

I opted for not doing any special logic treatment around `clientRequestId`. I think this behavior will be fine?

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/task/DualTaskRepository.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/task/DualTaskRepository.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/task/DualTaskRepository.java
```diff
@@ -0,0 +1,101 @@
...
+
+    List<Task> tasks = new ArrayList<>();
+    try {
+      tasks.addAll(primaryList.get(asyncTimeoutSeconds, TimeUnit.SECONDS));
+      tasks.addAll(previousList.get(asyncTimeoutSeconds, TimeUnit.SECONDS));
+    } catch (TimeoutException | InterruptedException | ExecutionException e) {
+      log.error("Could not retrieve list of tasks by timeout", e);
+
+      // Return tasks so we can still get data in partial failures
+      return tasks;
```
dreynaud: you could just take this line out and let the final return do the returning.

PR #3288, opened 2019-01-18
Update GCE base URL to the new domain (spinnaker#3863)

CodingFanSteve:
GCE is rolling out a new domain "compute.googleapis.com/compute".
Update GCE's base URL in Spinnaker code to point to this new domain.
The old domain "www.googleapis.com/compute" will still be valid, so the old release will still be working fine.


jtk54:
Do the paths in [ComputeVersion](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/ComputeVersion.groovy) need changed?

PR #3292, opened 2019-01-22
fix(kubernetes): cherry-pick #3247 into 1.10.x

mrampton:
This is the same as #3247 but for 1.10.x


mrampton:
Thanks for following up @ezimanyi -- was just looking at that as well.  I'll go ahead and make a separate PR for #3054 and update this branch afterward.

PR #3295, opened 2019-01-22
feat(aws/cloudformation): Allow on demand cache request/invalidation

xavileon:
This patch enabled the CloudFormationCachingAgent to also handle on
demand request, so items of the cache are force refreshed. We need
to invalidate this cache/forcefully refresh it when we update an
existing cloudformation stack, so we track the progress of the new
operation, not the result of the previous one.

Related issue https://github.com/spinnaker/spinnaker/issues/736

Signed-off-by: Xavi León <xavi.leon@schibsted.com>


jervi:
@spinnaker/reviewers PTAL

xavileon:
@ethanfrogers you may be interested in reviewing this as well.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonCloudFormationCachingAgent.java
```diff
@@ -39,28 +43,57 @@
...
+  }
+
+  @Override
+  public OnDemandMetricsSupport getMetricsSupport() {
+    return this.metricsSupport;
+  }
+
+  @Override
+  public boolean handles(OnDemandType type, String cloudProvider) {
+    return type == OnDemandType.CloudFormation && cloudProvider.equals(AmazonCloudProvider.ID);
```
ethanfrogers: I think we'd prefer to use `OnDemandType.CloudFormation.equals(type)` instead of `==`.

xavileon: Done in https://github.com/spinnaker/clouddriver/pull/3295/commits/36580ea576634912796418ed75b2578806d782e9

PR #3296, opened 2019-01-22
feat(provider/gce): Support accelerator types for zonal migs.

jtk54:


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/security/GoogleNamedAccountCredentials.groovy
```diff
@@ -386,6 +419,20 @@ class GoogleNamedAccountCredentials implements AccountCredentials<GoogleCredenti
     return zoneToInstanceTypesMap
   }
 
+  static Map<String, Map> convertAcceleratorTypeListToMap(AcceleratorTypeAggregatedList acceleratorTypeList) {
+    def zoneToAcceleratorTypesMap = acceleratorTypeList.items.collectEntries { zone, acceleratorTypesScopedList ->
+      println ",, accel types scoped list: ${acceleratorTypesScopedList}"
```
sbwsg: is this debug log intended to be kept?

jtk54: Nope good catch.

PR #3304, opened 2019-01-23
fix(provider/titus): factor result build time in stream update threshold

asher:


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
```diff
@@ -334,7 +336,7 @@ private void writeToCache(Boolean firstSnapshotWrite) {
           .collect(Collectors.toSet());
 
         cache.putCacheResult(getAgentType(), authoritative, result);
-        lastUpdate.set(startTime);
+        lastUpdate.set(System.currentTimeMillis());
```
asher: `writeToCache` and `buildCacheResults` are currently overly expensive for the goal of incrementally processing the Titus event stream. We load all load balancers and scaling policies from the Titus stack, then build and attempt to persist cacheResults for every job running in the stack, even if processing only the last 10 job events from the stream. This can take longer than `TIME_UPDATE_THRESHOLD_MS`, so setting `lastUpdate` to the start time instead of completion time can result in a large backlog in processing the stream, with most runtime spent handling unchanged resources.

`TitusStreamingUpdateAgent` will not perform well until further refactored to support incremental cache updates. This change is intended to slightly improve the situation while testing for correctness.

PR #3315, opened 2019-01-25
feat(cats): improved support for incremental CacheResults

asher:


cats/cats-core/src/main/java/com/netflix/spinnaker/cats/provider/ProviderCache.java
```diff
@@ -23,10 +23,42 @@
...
+  void putCacheResult(String source, Collection<String> authoritativeTypes, CacheResult cacheResult);
 
-    Collection<CacheData> getAll(String type, Collection<String> identifiers);
+  /***
+   * Equivalent to putCacheResult but no evictions are processed, even if explicitly within cacheResult
+   * @param source The calling caching agent
+   * @param authoritativeTypes Authoritative cache results define graph vertices, informative results create edges.
+   * @param cacheResult Results to store. Since evictions are bypassed, partial authoritative results are supported.
+   */
+  void addCacheResult(String source, Collection<String> authoritativeTypes, CacheResult cacheResult);
```
asher: This PR is primarily intended to add this `addCacheResult` method to ProviderCache. It differs from `putCacheResult` in that it bypasses all eviction logic, and from `putCacheData` in that it correctly stored authoritative vs. informative types. It is intended for incremental result storage, such as from an evented caching agent that manages its own evictions.

asher: @cfieber This is really here for the sql backed cache provider I'm working on. In the process of changing how relationships are stored and mapped, the authoritative vs. informative data types takes on a greater distinction. Only authoritative types store a full cacheResult object. If you consider that a graph vertex, informative results only add edges between them.

asher: Updated the javadoc comment to note that authoritativeTypes here may only be of use with some ProviderCache implementations.

cfieber: cool, makes sense!

PR #3318, opened 2019-01-28
fix(aws/cloudformation): Allow more generic failures to be displayed

xavileon:
We were filtering only the `CREATE_FAILED` failures to be shown in deck by filtering
that specific failure from the list of events of a given stack. However,
stacks could be failed also while updating (`UPDATE_FAILED`). This patch
makes the filtering of the failure reason more generic to handle such
cases.

Related issue: https://github.com/spinnaker/spinnaker/issues/736


jervi:
@spinnaker/reviewers PTAL

xavileon:
@ethanfrogers PTAL :)

PR #3319, opened 2019-01-28
fix(aws): Support disabling cloud formation via `aws.features.cloudFormation`

ajordens:
Defaults to `false` to avoid an AWS permission error when your
clouddriver instance profile has minimal grants.


ajordens:
FYI @xavileon -- we ran into permission errors and it would perhaps make sense if the `cloudformation` support had a feature flag gating it.

Happy to converse about whether it should default to enabled or disabled once more of the implementation is in-place.

xavileon:
@ajordens yes, it completely makes sense, thanks! I'll make the agent resilient to these errors in any case in a later PR, thanks for the report. This way it won't error if we ever decide it's worth having it enabled by default.

I'll keep this in mind for other parts of the feature needs gating :+1:

PR #3320, opened 2019-01-28
refactor(aws): Make cloudformation controllers consistent with rest of project

robzienert:
* Move CloudFormation code from `clouddriver-web` to `clouddriver-aws`
* Change REST endpoints to match style of the rest of the project, nest controller endpoint under `aws` root path


ezimanyi:
:+1:, clouddriver is failing to start for anyone without aws enabled right now because `CloudFormationController` at the top level depends on having aws-specific components to autowire.  I think this PR should fix it.

ajordens:
This looks better to me, thanks for the cleanup.

Not completely sold on needing `cloudFormation` in the URL since `stacks` may be self-explanatory when scoped to aws? I'll leave this to someone with stronger opinions.

It doesn't affect this PR, but can we get the model objects more appropriately named @xavileon? `CloudFormation` is perhaps more aptly named `CloudFormationStack`. This is an AWS-specific controller so the `List<CloudFormationProvider` may be overkill ... are we ever going to have more than one for AWS?

ajordens:
As this progresses, we'll need to get security around these new endpoints and operations. Using `accountId` vs `accountName` is a little non-standard and may complicate the security filters.

xavileon:
@ajordens sure, it makes sense, will do shortly. Actually, having it in the core module was a left over of a first attempt to make it generic enough to support other kind of infrastructure templates (e.g. heat templates for openstack, gcp manifests, etc.). But it got half-done and not appropriately named so making it aws specific makes more sense 👍

PR #3322, opened 2019-01-29
fix(provider/kubernetes): Null boolean on KubernetesV2ServerGroup

ncknt:
* KubernetesV2ServerGroup `disabled` flag is null and causes a NPE.
*`disabled` is set on https://github.com/spinnaker/clouddriver/pull/3060. This fix is just to get past the issue on 1.10.x. It doesn't show on later releases.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/ProjectClustersService.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/ProjectClustersService.java
@@ -255,7 +255,7 @@ public class ProjectClustersService {
         .flatMap(ac -> ac.getServerGroups().stream())
         .filter(serverGroup ->
           serverGroup != null &&
-            !serverGroup.isDisabled() &&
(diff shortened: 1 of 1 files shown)


ncknt:
We could also just protect again null values in `isDisabled()`. It wasn't clear if a `null` value is expected/ok. LMK and I'll update the PR.

ezimanyi:
@ncknt : It looks like this same bug is in `master` as well.  (I know you noted above it doesn't show on other releases---but it may be because it's intermittent somehow?)  I'd probably prefer fixing on master then cherry-picking the fix onto 1.10 and 1.11 than only fixing in 1.10.

As for the actual fix, this looks fine to me---ideally the field would probably be a `boolean` instead of a `Boolean` to avoid this, but given that the interface exposes a `Boolean` it's probably best to just make callers null-safe rather than just fix the kubernetes implementation.  (It looks like almost all other callers are in groovy, and if I'm not mistaken groovy will just handle this fine, treating `null` as falsey, so what you've done here is consistent.)

PR #3323, opened 2019-01-29
refactor(cloudformation): rename cf model and use accountName instead of accountId

xavileon:
This patch renames the cloud formation model to CloudFormationStack to better reflect its meaning. This change also uses the aws account name instead of the account id on the controller and cloud formation caching agent as suggested in https://github.com/spinnaker/clouddriver/pull/3320#issuecomment-458314739. It also makes the cloud formation caching agent resilient to permission errors, so it doesn't fail if the role assumed by spinnaker cannot list cloud formation stacks.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/cache/Keys.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/cache/Keys.groovy
@@ -35,7 +35,7 @@ class Keys implements KeyParser {
     KEY_PAIRS,
     INSTANCE_TYPES,
     ELASTIC_IPS,
-    CLOUDFORMATION,
+    STACKS,
     ON_DEMAND
 
     final String ns
@@ -109,7 +109,7 @@ class Keys implements KeyParser {
       case Namespace.ELASTIC_IPS.ns:
         result << [address: parts[2], account: parts[3], region: parts[4]]
         break
-      case Namespace.CLOUDFORMATION.ns:
+      case Namespace.STACKS.ns:
         result << [id: parts[2], account: parts[3], region: parts[4]]
         break;
       default:
@@ -168,6 +168,6 @@ class Keys implements KeyParser {
   static String getCloudFormationKey(String stackId,
                                      String region,
                                      String accountName) {
-    "$ID:${Namespace.CLOUDFORMATION}:${accountName}:${region}:${stackId}"
+    "$ID:${Namespace.STACKS}:${accountName}:${region}:${stackId}"
   }
 }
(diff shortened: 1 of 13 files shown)


xavileon:
@ajordens could you please take a look? it's addressing the suggestions on https://github.com/spinnaker/clouddriver/pull/3320#issuecomment-458314739

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/cache/Keys.groovy
```diff
@@ -167,7 +167,7 @@ class Keys implements KeyParser {
 
   static String getCloudFormationKey(String stackId,
                                      String region,
-                                     String account) {
-    "$ID:${Namespace.CLOUDFORMATION}:${account}:${region}:${stackId}"
+                                     String accountName) {
+    "$ID:${Namespace.CLOUDFORMATION}:${accountName}:${region}:${stackId}"
```
ajordens: `CLOUDFORMATION` still seems not quite right here ... aren't we specifically dealing with cloudformation stacks?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonCloudFormationCachingAgent.java
```diff
@@ -81,37 +81,41 @@ public CacheResult loadData(ProviderCache providerCache) {
...
+        Map<String, Collection<String>> relationships = new HashMap<>();
+        relationships.put(CLOUDFORMATION.getNs(), Collections.singletonList(stackCacheKey));
+        stackCacheData.add(new DefaultCacheData(stackCacheKey, stackAttributes, relationships));
       }
-      String stackCacheKey = Keys.getCloudFormationKey(stack.getStackId(), region, account.getAccountId());
-      Map<String, Collection<String>> relationships = new HashMap<>();
-      relationships.put(CLOUDFORMATION.getNs(), Collections.singletonList(stackCacheKey));
-      stackCacheData.add(new DefaultCacheData(stackCacheKey, stackAttributes, relationships));
+    } catch (Exception e) {
+      log.info("Error retrieving stacks, ignoring error: {}", e.getMessage());
```
ajordens: Normally this would probably be something like `log.error("Error retrieving stacks", e)` (it's probably worth a full stacktrace)

If there's a more appropriate error to catch for auth failures than we could catch that too.

xavileon: The amazon client throws a generic `AmazonCloudFormationException` but still worth being more specific than `Exception`

xavileon:
Thanks @ajordens for the comments, addressed them in https://github.com/spinnaker/clouddriver/pull/3323/commits/6cd86281cb4329026558306a4852c6e146524afd

PR #3325, opened 2019-01-29
fix(provider/kubernetes): fix NPE on KubernetesV2ServerGroup disabled…

ncknt:
KubernetesV2ServerGroup disabled flag is null and causes a NPE.
Related to spinnaker/spinnaker#3490 and cherry picked PRs: #3292, #3293,  #3294

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/ProjectClustersService.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/ProjectClustersService.java
@@ -255,8 +255,7 @@ public class ProjectClustersService {
         .flatMap(ac -> ac.getServerGroups().stream())
         .filter(serverGroup ->
           serverGroup != null &&
-            serverGroup.isDisabled() != null &&
-            !serverGroup.isDisabled().booleanValue() &&
+            (serverGroup.isDisabled() == null || !serverGroup.isDisabled()) &&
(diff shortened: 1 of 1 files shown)


ncknt:
@ezimanyi Here's #3322 on master this time.

ezimanyi:
Actually as I look at this one more time, we might actually want to treat `null` as equivalent to `false` to be consistent with how this works in the rest of the (groovy) code.  Everywhere else, if this is `null` it will be treated as false as `!serverGroup.isDisabled()` will be true whether `isDisabled` returns `false` or `null`.  It *probably* doesn't matter that much, but is probably better to be consistent.

Would like @lwander to take a look before merging

ncknt:
I'd expect `null` to be equivalent of `false`. Isn't it what we're doing here?

ezimanyi:
No, it took me a second read to realize this:
```
a != null && !a
false when a == null
true when a == false
```

I think we'd want
```
(a == null || !a)
```

ncknt:
haha. Not my day, thanks for double checking. I've updated the PR.

lwander:

ezimanyi:
@ncknt : Was it 1.10 where you wanted this fix?  If so, we can cherry-pick into 1.11 and 1.10 as well

ncknt:
@ezimanyi Yes 1.10 and 1.11 would be great too.

ezimanyi:

ezimanyi:

PR #3332, opened 2019-01-30
feat(provider/kubernetes): recreate annotation support

lwander:


lwander:
https://github.com/spinnaker/spinnaker.github.io/pull/1198

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/KubernetesDeployManifestOperation.java
```diff
@@ -240,6 +241,10 @@ private boolean isVersioned(KubernetesResourceProperties properties, KubernetesM
     return properties.isVersioned();
   }
 
+  private boolean isRecreate(KubernetesManifestStrategy strategy) {
+    return strategy.getRecreate() != null ? strategy.getRecreate() : false;
```
ezimanyi: There's a lot of places where we have `boolean` properties as `Boolean` that leads to logic like this.  I suppose at this point it might be better to be consistent---but is there a reason not to have the property be a `boolean` to avoid this?

lwander: Chatted off-line in more detail, but the intent is to leave this as is since
1. `Boolean` accurately reflects that this can be `null`, `false`, or `true`, and
2. `null` != `false` in all cases, other flags default to `true`, where only new behavior is introduced in the negative case.

PR #3347, opened 2019-02-02
chore(aws): update status with size on ASG creation

dreynaud:


dreynaud:
Should be a little quality of life improvement

PR #3351, opened 2019-02-05
fix(entitytags): Handle non-deserializable entity tags

ajordens:
Specifically situations where a tag identifies itself as an object but
is actually a `String`.


ajordens:
Our delete logic is actually contingent on being able to deserialize an entity tag in the first place ... this PR should make that not fail.

PR #3352, opened 2019-02-06
fix(provider/aws): change STS endpoints for GovCloud and China regions

kevinheins:
Issue link: https://github.com/spinnaker/spinnaker/issues/3941

As detailed in https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/STSAssumeRoleSessionCredentialsProvider.html, STS has special endpoints for GovCloud and China regions.

At credential creation time, we don't have a way to easily check what region we are in, so I have it relying on the `arn` string being passed in. Clouddriver allows for a full arn, so we can check it for `aws-us-gov` or `aws-cn`.

I will also update the documentation as to point it passing in a full ARN is allowed.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/NetflixSTSAssumeRoleSessionCredentialsProvider.java
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/NetflixSTSAssumeRoleSessionCredentialsProvider.java
@@ -28,6 +28,17 @@ public class NetflixSTSAssumeRoleSessionCredentialsProvider extends STSAssumeRol
                                                         String accountId) {
     super(longLivedCredentialsProvider, roleArn, roleSessionName);
     this.accountId = accountId;
+
+    /**
+     Need to explicitly set sts region if GovCloud or China as per
+     https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/STSAssumeRoleSessionCredentialsProvider.html
+     */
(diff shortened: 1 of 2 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/NetflixSTSAssumeRoleSessionCredentialsProvider.java
```diff
@@ -28,6 +28,17 @@ public NetflixSTSAssumeRoleSessionCredentialsProvider(AWSCredentialsProvider lon
...
+    /**
+     Need to explicitly set sts region if GovCloud or China as per
+     https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/STSAssumeRoleSessionCredentialsProvider.html
+     */
+    if(roleArn.contains("aws-us-gov")){
+      setSTSClientEndpoint("sts.us-gov-west-1.amazonaws.com");
+    }
+    if(roleArn.contains("aws-cn")){
+      setSTSClientEndpoint("sts.cn-north-1.amazonaws.com.cn");
+    }
```
robzienert: Looks all good, but do `if (...) {`

kevinheins: This should do

ajordens:
LGTM ... I'll let @robzienert merge at his convenience.

kevinheins:

kevinheins:

PR #3356, opened 2019-02-07
feat(core): `DeploymentResult` support for returning target capacity

ajordens:
AWS deployments are somewhat special in that they create an initial
asg of size 0/0/0 and then subsequently resize it to the target
capacity.

This causes problems if the `ForceCacheRefresh` task as part of a deploy
were to ever be removed from `orca` as it's possible for the 0/0/0 asg
to be cached and the `WaitForUpInstances` task to prematurely complete.

I'd like to remove the `ForceCacheRefresh` task as part of a deploy,
hence this PR and a follow-up in `orca` that makes it aware of the
capacity being passed back from `clouddriver`.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
@@ -100,9 +100,10 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
 
   @Override
   DeploymentResult handle(BasicAmazonDeployDescription description, List priorOutputs) {
-    task.updateStatus BASE_PHASE, "Initializing handler..."
     def deploymentResult = new DeploymentResult()
+
     task.updateStatus BASE_PHASE, "Preparing deployment to ${description.availabilityZones}..."
+
     for (Map.Entry<String, List<String>> entry : description.availabilityZones) {
       String region = entry.key
 
@@ -251,6 +252,12 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
         description.spotPrice = null
       }
 
+      def capacity = new DeploymentResult.Deployment.Capacity(
+          min: description.capacity.min ?: 0,
+          max: description.capacity.max ?: 0,
+          desired: description.capacity.desired ?: 0
+      )
+
       def autoScalingWorker = new AutoScalingWorker(
         application: description.application,
         region: region,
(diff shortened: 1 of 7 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/deploy/DeploymentResult.groovy
```diff
@@ -28,4 +27,83 @@ class DeploymentResult {
   Map <String, List<String>> deployedNamesByLocation = [:]
 
   List<Artifact> createdArtifacts = []
+  Set<Deployment> deployments = []
```
ajordens: I'd love to standardize on this method of returning data back to orca ... the previous `serverGroupNamesByRegion` and `deployedNamesByLocation` are too specific and only capture a single piece of data.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
```diff
@@ -100,9 +100,11 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
 
   @Override
   DeploymentResult handle(BasicAmazonDeployDescription description, List priorOutputs) {
-    task.updateStatus BASE_PHASE, "Initializing handler..."
     def deploymentResult = new DeploymentResult()
+
+    task.updateStatus BASE_PHASE, "Initializing handler..."
```
dreynaud: I don't think we need 2 `task.updateStatus` in a row

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
```diff
@@ -251,6 +253,12 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
         description.spotPrice = null
       }
 
+      def capacity = new DeploymentResult.Deployment.Capacity(
+          min: description.capacity.min != null ? description.capacity.min : 0,
```
dreynaud: isn't that what the elvis operator is for?

`min: description.capacity.min ?: 0,`

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
```diff
@@ -313,6 +321,10 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
       description.events << new CreateServerGroupEvent(
         AmazonCloudProvider.ID, account.accountId, region, asgName
       )
+
+      deploymentResult.deployments.add(
+          new DeploymentResult.Deployment(location: region, serverGroupName: asgName, capacity: capacity)
```
dreynaud: neato! But shouldn't we add the cloud provider and account here for completeness?

ajordens: That won't hurt ... consider it done!

ajordens:
Updated with your feedback @dreynaud.

PR #3361, opened 2019-02-12
refactor(core): Make JobExecutor synchronous

ezimanyi:
## Overview
The `JobExecutor` class is used to run external commands; it currently has a single implementation, `LocalJobExecutor` which forks a subprocess on the local machine to execute the command.

The current implementation is asynchronous; the parent process calls `JobExecutor.startJob`, and receives back an Job ID.  It can then call `JobExecutor.getJobStatus` with the id to get the status of the job.  The status includes:
* whether the job is still running or completed
* if the job is completed, the results of the job: exit status, stdout, stderr

Every single use of the `JobExecutor` interface in `clouddriver` uses the following pattern:
* Initiate a job by calling `startJob`
* Call `getJobStatus` in a loop (with exponential backoff) to wait for the job to finish
* Process the results of the command by reading the `JobStatus` that is returned

For short calls, the overhead of starting the job asynchronously and polling for the result accounts for the majority of the wall-clock time.  It would be more efficient (and make the code less complex) to replace this with a simple synchronous `JobExecutor` that exposes a single method `start` that blocks until the child process exits, and returns the results.

### Feature parity
The same `org.apache.commons.exec.*` library provides a synchronous execute method that supports the same features we're already using:
* Sending `stdout` and `stderr` from the child process to a `ByteArrayOutputStream` in the parent
* Timing out the process after a specified timeout

### Output
Moving to synchronous execution does not have any effect on the piping of output from the child process to the parent process; in both cases `PumpStreamHandler` creates a daemon thread to read any output from the child process and copy it to a buffer in the parent process.  (And another thread to do the same with the error from the child process.)  We then read that buffer when the child process finishes.

The only difference is that in the synchronous case, we'll likely read the buffer sooner, as we'll know it's ready more quickly than if we had to wait for a polling cycle.

### CPU usage

I've done testing by having clouddriver cache a cluster with ~1000 deployments, each with a service (and each with a replicaset scaling out to 2 pods).  Clouddriver was configured with 8 kubernetes accounts, each separately aching the cluster.  There is no noticeable difference in clouddriver CPU usage before and after the change.

### Cache time

There was also no significant difference in total cache time (as measured by logs of `kubernetes:[account]/KubernetesCoreCachingAgent[1/1] completed in [time]s`) before and after the change (both ~59s).

## Overall assessment

This didn't help as I had hoped with cache cycle times (particularly as the depth of consecutive call is not as deep as I thought from experience with GCE), but does significantly speed up startup where we make a bunch of `kubectl` calls in a single-threaded loop.  It also significantly simplifies the code, which may make further optimizations more evident/feasible.


ezimanyi:
@lwander : Did some load testing, and did not see any increased CPU usage with this change.  It didn't help cache cycles as I hoped (given they are not composed of as many short sequential `kubectl` calls as I thought) but it does significantly speed up startup (even without parallelization, which I have a separate PR for).  Based on the considerations above I think it still makes sense to move forward with this to remove the unecessary overhead from each call.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/jobs/JobExecutor.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Google, Inc.
+ * Copyright 2019 Google, Inc.
```
jtk54: Generally this is file creation, not update.

ezimanyi: Yeah, I wasn't sure what to do here...the only thing that hasn't changed here is the name of the class so it felt like more of a new file.

jtk54: I didn't realize it was moved from the git diff, SGTM.

ezimanyi:
This should help a lot with spinnaker/spinnaker#3323

PR #3366, opened 2019-02-12
fix(kubernetes/v2): Parallelize checking for omitKinds

ezimanyi:
For users with a lot of Kubernetes V2 accounts configured, the check of which kinds are readable takes a lot of time on startup. Parallelize this check so that clouddriver starts up faster.


ezimanyi:
Closes spinnaker/spinnaker#3323

ezimanyi:

davidxia:
@ezimanyi just tried this out today in 1.12.5. Clouddriver k8s pod's startup time for our 7 clusters used to be ~10min. Now it's 2! Thanks.

PR #3367, opened 2019-02-12
fix(appengine): Fix appengine deployments

ezimanyi:
The new JobExecutorLocal is in Java, and fails when passed an array containing a mixture of Strings and GStrings.  As a GString is not a subclass of String (which is final) it is incorrect to create a collection containing a mixture (and generates a warning). Fix this at the source by not creating arrays that mix String and Gstring.


clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DeployAppengineAtomicOperation.groovy
```diff
@@ -262,13 +262,13 @@ class DeployAppengineAtomicOperation implements AtomicOperation<DeploymentResult
       deployCommand << gcloudReleaseTrack.toString().toLowerCase()
     }
     deployCommand += ["app", "deploy", *(repositoryFullConfigFilePaths + writtenFullConfigFilePaths + configArtifactPaths)]
-    deployCommand << "--version=$versionName"
+    deployCommand << "--version=" + versionName
```
jtk54: I'd drop a comment here that `deployCommand` should be `String` only and not mixed so nobody re-learns this lesson.

ezimanyi: Good point, will do

PR #3368, opened 2019-02-12
fix(provider/gce): Fix metadata for autoscaler deletes.

jtk54:


jtk54:

PR #3377, opened 2019-02-14
fix(provider/appengine): Fix NPE thrown when deploying GCS object usi…

ttomsu:
…ng default creds

Fixes https://github.com/spinnaker/spinnaker/issues/3975


sbwsg:

PR #3379, opened 2019-02-14
fix(provider/kubernetes): Fix NPE on ingress with no http

ezimanyi:
The http field is optional on v1beta1.IngressRule, but the caching logic currently throws an NPE when this is the case.


ezimanyi:
Here's the spec listing it as [optional](https://kubernetes.io/docs/reference/federation/extensions/v1beta1/definitions/#_v1beta1_ingressrule)

ezimanyi:

PR #3381, opened 2019-02-14
fix(artifacts/front50): Fix serialization of artifacts

ezimanyi:
We need to mark front50Service and objectMapper JsonIgnore so we don't try to serialize them when returning the accounts via /artifacts/credentials.


ezimanyi:
Tested via `curl http://localhost:7002/artifacts/credentials`; fails before, works now

PR #3383, opened 2019-02-15
fix(aws): Fallback if asg instances cannot be described

ajordens:
There has been some back and forth around the method used to determine
instances to enable/disable.

This PR introduces a fallback lookup based on instances tagged with the
asg name in the event that the original lookup failed.

Historically we looked images up this way but moved away due to slowness
in larger account/regions. It seems reasonable to at least try it as a
fallback rather than letting the operation fail.


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AbstractEnableDisableAtomicOperation.groovy
```diff
@@ -126,23 +127,35 @@ abstract class AbstractEnableDisableAtomicOperation implements AtomicOperation<V
...
             if (failedAttempts >= 10) {
               task.updateStatus phaseName, "Failed to describe instances 10 times, aborting. This may happen if the server group has been disabled for a long period of time."
-              return false
+
+              try {
+                // fallback to a tag-based instance lookup (will be slower in large region/accounts)
+                reservations.addAll(fetchInstancesTaggedWithServerGroup(regionScopedProvider, serverGroupName))
+                break
+              } catch (Exception e2) {
+                log.error(
```
cfieber: maybe worth a task.updateStatus in here

PR #3386, opened 2019-02-15
 fix(artifacts/github): Fix threading bug in github artifact resolver

ezimanyi:
GithubArtifactCredentials is storing a Request.Builder in an instance variable. This means that when multiple threads are concurrently downloading Github artifacts, it's possible for one thread to download an artifact from another thread's URL.


ezimanyi:
Attempt to resolve spinnaker/spinnaker#3932; the evidence on that issue points to this being the root cause, though I can't be fully sure.

ezimanyi:

PR #3388, opened 2019-02-16
fix(dependencies): manage common-langs with spinnaker-dependencies

mrampton:
Having spinnaker-dependencies manage commons-lang ensures a consistent version will be provided for the dependency

https://github.com/spinnaker/spinnaker-dependencies/blob/97aaaf2e579daaaaaf851532f23a41f4bf622420/src/spinnaker-dependencies.yml#L17


mrampton:

mrampton:

ezimanyi:
Is there a reason for cherry-picking this into 1.11 and 1.12?  While it's good to clean this up, I'm not sure I see a strong motivation for pulling it into stable Spinnaker versions.

mrampton:
Hi @ezimanyi. We have customers running 1.11 and 1.12 who are experiencing issues which this fix eliminates.  Happy to tackle this another way if you prefer.  The missing version for the common-langs dependency seems like an oversight or mistake. Thoughts?

ezimanyi:
Oh, I see---it's not even specified at all in the old code.  If this is fixing an actual issue, makes sense to pull into older releases---I thought this was just general cleanup.

mrampton:
Yup -- sorry, i should have made that more clear! Will be more explicit with future issues.  Thanks for chiming in.

PR #3390, opened 2019-02-19
fix(artifacts): Fix threading bug in HTTP artifact credentials

ezimanyi:
Many artifact credentials duplicate the same logic of:
- Getting an authorization header from the artifact account config
- Generating a URL from an artifact, and sending a request to download it

This logic is not thread-safe, which was recently fixed in the Github artifact handler.  This commit pulls the common logic into a base class and removes the duplicate (and non-threadsafe) implementations.


ezimanyi:
The last commit broke some tests; I'm going to pull that commit into a separate PR and just submit the first 5 commits here.

PR #3392, opened 2019-02-19
refactor(artifacts): Clean up Spring dependency injection

ezimanyi:
Each artifact type creates its own okHttpClient; just create a single one that is autowired into each.

Instead of having the individual types depend on the repository and modify it, have the repository depend on the individual artifact types. It can then just create a list from these injected types.

Improve the interface of artifactCredentialsRepository. Most calls to getAllCredentials filter based on name and/or type; add this filtering as a supported function in the interface and clean up the calling code.


ezimanyi:
@jtk54 : The first commit I just pulled off of #3390 (so you've already seen it) because I broke some tests and wanted to fix them in a separate PR.  The second commit fixes the tests.

PR #3393, opened 2019-02-19
fix(search/cats): Avoid short-circuiting when filters are empty.

jtk54:
Avoids a bad path where cats tries to optimize the search q by looking for things in the filters. If filters are empty this always short-circuits and returns an empty result list.


jtk54:

PR #3394, opened 2019-02-19
fix(kubernetes): return only valid kinds from UnregisteredCrdCachingA…

maggieneterval:
…gent.primaryKinds

(Hopefully) closes https://github.com/spinnaker/spinnaker/issues/3972

I attempted to reproduce #3972 by enabling a service account w/ RBAC that excluded permissions for a CRD I had previously deployed to a cluster I manage with Spinnaker, and not added to the whitelisted `kinds` in my hal config. Since I could not reproduce the exact errors specified in #3972, my best guess is that at least the first error originated from the following sequence of events in Clouddriver:
`KubernetesV2CachingAgent.loadPrimaryResourceList` calls `KubernetesV2Credentials.list` with the kinds returned from `KubernetesUnregisteredCustomResourceCachingAgent.primaryKinds`, which does not filter by valid kinds. Since we _do_ filter by valid kinds in `KubernetesCoreCachingAgent.primaryKinds`, my guess is that we should do the same here and not be doing anything with non-whitelisted CRDs.

Apologies for bothering you @lwander but since the issue mentions that you had believed this issue to already be resolved, would love your input if you have a second. Thanks!

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesUnregisteredCustomResourceCachingAgent.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesUnregisteredCustomResourceCachingAgent.java
@@ -63,6 +63,8 @@ public class KubernetesUnregisteredCustomResourceCachingAgent extends Kubernetes
 
   @Override
   protected List<KubernetesKind> primaryKinds() {
-    return credentials.getCrds();
+    return credentials.getCrds().stream()
+      .filter(credentials::isValidKind)
+      .collect(Collectors.toList());
   }
 }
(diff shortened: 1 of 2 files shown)


ezimanyi:
Was just thinking about this some more, and this may need another small change to fully fix the bug.  If I'm reading/understanding correctly, the order is:
* When reading the account config, we load all custom kinds in the account's `customResources` config (plus all the hard-coded standard kinds)
* We then check if all kinds currently known are readable in `determineOmitKinds` which adds anything unreadable to `omitKinds`
* Then we set up a `liveCrdSupplier` that polls for new kinds and adds them to the list of known kinds

So I think that Spinnaker only knows about these CRDs after it's already made it's list of unreadable things to omit, so they will all pass the `isValid` check.  In addition to this change, we might want to also move `determineOmitKinds` after the first poll of CRDs [here](https://github.com/spinnaker/clouddriver/blob/05bef64cd3c6c91db81831a21d3def56f5a08b14/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java#L388).

It's also possible that I'm mis-understanding how this works and the change as-is will work.

maggieneterval:
Ahhh thank you @ezimanyi, I think that makes sense -- I moved `determineOmitKinds()` _after_ the first call to `liveCrdSupplier.get()`.

You're off the hook Lars, I'm removing you as a reviewer :cry:

maggieneterval:

davidxia:
@maggieneterval Just tried this out today. Logs are so much more signal now :) Thanks!

maggieneterval:
Hooray, so happy to hear that @davidxia!

PR #3400, opened 2019-02-20
fix(provider/azure): add support for Azure US Government environment

kevinheins:
Issue Link: https://github.com/spinnaker/spinnaker/issues/4006

### DIFF
--- a/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureBaseClient.groovy
+++ b/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureBaseClient.groovy
@@ -54,10 +54,20 @@ public abstract class AzureBaseClient {
    * @param clientId
    * @param tenantId
    * @param secret
+   * @param configuredAzureEnvironment
    * @return
    */
-  static ApplicationTokenCredentials getTokenCredentials(String clientId, String tenantId, String secret) {
-    new ApplicationTokenCredentials(clientId, tenantId, secret, AzureEnvironment.AZURE)
+
+  static ApplicationTokenCredentials getTokenCredentials(String clientId, String tenantId, String secret, String configuredAzureEnvironment) {
+    if ( configuredAzureEnvironment == "AZURE_US_GOVERNMENT") {
+      return new ApplicationTokenCredentials(clientId, tenantId, secret, new AzureEnvironment(
+        "https://login.microsoftonline.us/",
+        "https://management.core.usgovcloudapi.net/",
+        true));
+    }
+    else {
+      return new ApplicationTokenCredentials(clientId, tenantId, secret, AzureEnvironment.AZURE)
+    }
   }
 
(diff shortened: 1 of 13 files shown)


mybayern1974:
1. Please define an enum called say "_AzureEnvironment.US_Gov_" like that of "_AzureEnvironment.AZURE_" to make the codes consistent;
2. How can the code know the current env is _Azure.Gov_? Do you plan to also make PR to the Halyard to let hal set up the env at the very beginning?
3. Suggest you could ensure the end-2-end flow for this role could work, in terms of either theory or proof-of-concept, and after that submit PRs. I'm suggesting this because maybe you need to further add special support for the _Gov_ subs on other Azure resources say Key Vault, Service Principle, etc (I'm just guessing. I'm not the Azure Gov expert). Any future possible breakings may require reverting previous submitted codes. So I'm suggesting you have a full picture offline first (not necessarily to implement all of them).

kevinheins:
> 1. Please define an enum called say "_AzureEnvironment.US_Gov_" like that of "_AzureEnvironment.AZURE_" to make the codes consistent;
> 2. How can the code know the current env is _Azure.Gov_? Do you plan to also make PR to the Halyard to let hal set up the env at the very beginning?
> 3. Suggest you could ensure the end-2-end flow for this role could work, in terms of either theory or proof-of-concept, and after that submit PRs. I'm suggesting this because maybe you need to further add special support for the _Gov_ subs on other Azure resources say Key Vault, Service Principle, etc (I'm just guessing. I'm not the Azure Gov expert). Any future possible breakings may require reverting previous submitted codes. So I'm suggesting you have a full picture offline first (not necessary to implement all of them).

1. The version of the Azure SDK that is used, `com.microsoft.azure:azure-client-authentication:1.0.0-beta1`, only supports _AzureEnvironment.AZURE_ and _AzureEnvironment.AZURE_CHINA_. This is why I have `getTokenCrendetials` create a new environment. I tried upgrading to `com.microsoft.azure:azure-client-authentication:1.0.0-beta2` but there were many breaking changes.

2. I am hooking into the `environment` configuration variable in clouddriver.yml that seems to be unused. You can provide this with a custom cloudriver-local.yml file. It looks something like this
````
azure:
  enabled: true
  accounts:
  - name: my-azure-account
    requiredGroupMembership: []
    providerVersion: V1
    permissions: {}
    clientId: --------
    appKey: --------
    tenantId: --------
    subscriptionId: --------
    defaultResourceGroup: spinnaker-rg
    defaultKeyVault: spinnaker-kv
    environment: AZURE_US_GOVERNMENT
    regions:
    - usgovvirginia
  primaryAccount: my-azure-account
  bakeryDefaults:
    templateFile: azure-linux.json
    baseImages: []
````
3. I have tested this locally with an Azure Government account and it is working. I can create security bakes, start bakes, etc. Clouddriver properly authenticates and caches objects.

kevinheins:
@mybayern1974 anything stopping a merge here?

kevinheins:

PR #3403, opened 2019-02-21
fix(elasticsearch): Fix connection string for elasticsearch tests

ezimanyi:
The elasticsearch tests spin up a new local node agains which to run the tests; this node defaults to port 9200, but will pick another port if that port is not available. The tests currently always send requests to port 9200 (which could be an unrelated node); update them to send requests to the node created by the tests.


ezimanyi:
Finally dug into why the elasticsearch tests were failing for me some days but not others---turns out they failed whenever I had my own elasticsearch node running on port 9200.  (The failure itself was due to the fact I was running an incompatible version of elastic search; this just fixes the test so it always uses the locally-spun-up elasticsearch.)

ajordens:
Oops ... past self apologizes!

Thanks for the fix!

PR #3406, opened 2019-02-22
fix(cf): Provide error message for region case mismatch

jmelchio:
spinnaker/spinnaker#4015

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/converters/AbstractCloudFoundryAtomicOperationConverter.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/converters/AbstractCloudFoundryAtomicOperationConverter.java
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.clouddriver.cloudfoundry.deploy.converters;
 
 import com.netflix.spinnaker.clouddriver.artifacts.ArtifactDownloader;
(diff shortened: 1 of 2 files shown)


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/converters/AbstractCloudFoundryAtomicOperationConverter.java
```diff
@@ -40,9 +41,18 @@
     CloudFoundrySpace space = CloudFoundrySpace.fromRegion(region);
 
     // fully populates the space guid which is what Cloud Foundry's API expects as an input, not the name.
-    return client.getOrganizations()
+    Optional<CloudFoundrySpace> spaceOptional = client.getOrganizations()
       .findByName(space.getOrganization().getName())
       .map(org -> client.getSpaces().findByName(org.getId(), space.getName()));
+
+    spaceOptional.ifPresent(spaceCase -> {
+      String[] parts = region.split(">");
```
claymccoy: The region split and trim here are duplicate what is already done in fromRegion above. This is checking more than just case and seems overly complicated.

jmelchio: Yes, you're right. Let me take a stab at it.

PR #3409, opened 2019-02-22
feat(aws): Support for disabling an account

ajordens:
Accounts can now mark themselves as disabled and opt-out of being
indexed without explicitly having to be removed from clouddriver
configuration.

Initial implementation is `aws`-specific but other providers can update
their respective `AccountCredentials` implementation to take advantage.


emjburns:
LGTM

PR #3414, opened 2019-02-26
test(core): Simple startup integration test

ezimanyi:
It's very common for one of our microservices to fail to start because of a missing autowire dependency. Often this is because someone changes autowiring such that it only works with a particular feature enabled, but fails with it disabled.

To prevent these issues, add an extremely simple integration test that just checks that clouddriver starts up with the default spring profile. This won't catch everything, but should catch most of the common cases I've seen. (We can always add test for other common profiles later if desired.)


ezimanyi:
This also makes two minor changes to make autowiring work when there are no cloud providers defined; I'm not sure how useful clouddriver would be in that case, but it seems reasonable for this controllers to handle the base case gracefully, and it means we can then test clouddriver-core easily without needing a cloud provider.

ezimanyi:
This test will (correctly) fail until rebased on top of #3412.

duftler:
nice

PR #3419, opened 2019-02-27
fix(google): Fix search for Google provider

ezimanyi:
Search is currently broken for the Google provider because it does not implement a key parser, and the search logic returns and empty list for any provider that does not.

Closes spinnaker/spinnaker#4050


ezimanyi:
Closes spinnaker/spinnaker#4050

ezimanyi:

PR #3425, opened 2019-03-04
feat(cf): default env if not set for account

Jammy-Louie:


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/security/CloudFoundryCredentials.java
```diff
@@ -49,7 +50,7 @@
 
   public CloudFoundryCredentials(String name, String appsManagerUri, String metricsUri, String apiHost, String userName, String password, String environment) {
```
jkschneider: Can you mark environment as `@Nullable`?

Jammy-Louie: 👍

PR #3434, opened 2019-03-06
fix(provider/gce): Fix seqence generation for server groups.

jtk54:
Also fixes an NPE when applying monikers. @jervi PTAL.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/names/GoogleLabeledResourceNamer.java
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/names/GoogleLabeledResourceNamer.java
@@ -57,12 +57,12 @@ public class GoogleLabeledResourceNamer implements NamingStrategy<GoogleLabeledR
     Names parsed = Names.parseName(name);
 
     Moniker moniker = Moniker.builder()
-        .app(parsed.getApp())
-        .cluster(parsed.getCluster())
-        .detail(parsed.getDetail())
-        .stack(parsed.getStack())
-        .sequence(parsed.getSequence())
-        .build();
+      .app(parsed.getApp())
+      .cluster(parsed.getCluster())
+      .detail(parsed.getDetail())
+      .stack(parsed.getStack())
+      .sequence(parsed.getSequence())
+      .build();
 
     Map<String, String> labels = labeledResource.getLabels();
     if (moniker.getApp() != null && labels != null) {
@@ -99,9 +99,6 @@ public class GoogleLabeledResourceNamer implements NamingStrategy<GoogleLabeledR
 
   private static <T> void setIfPresent(Consumer<T> setter, T value) {
     if (value != null) {
-      if (value.equals("")) {
-        value = null;
-      }
       setter.accept(value);
     }
   }
--- a/clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/names/GoogleLabeledResourceNamerSpec.groovy
+++ b/clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/names/GoogleLabeledResourceNamerSpec.groovy
@@ -50,24 +50,24 @@ class GoogleLabeledResourceNamerSpec extends Specification {
     }
 
     where:
-    name                                                      | labels                                    || expectedApp  | expectedCluster                                      | expectedStack     | expectedDetail | expectedSequence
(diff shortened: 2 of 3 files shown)


jervi:
Nice 👍

clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/names/GoogleLabeledResourceNamerSpec.groovy
```diff
@@ -50,24 +50,24 @@ class GoogleLabeledResourceNamerSpec extends Specification {
...
+    "cass-nccpintegration-random-junk-d0prod-z0useast1a-v003" | [:]                                   || "cass"       | "cass-nccpintegration-random-junk-d0prod-z0useast1a" | "nccpintegration" | "random-junk" | 3
+    "cass-nccpintegration-random-junk-d0prod-z0useast1a-v003" | [(APP): "myApp"]                      || "myApp"      | "cass-nccpintegration-random-junk-d0prod-z0useast1a" | "nccpintegration" | "random-junk" | 3
+    "cass-nccpintegration-random-junk-v003"                   | [(CLUSTER): "myCluster"]              || "cass"       | "myCluster"                                          | "nccpintegration" | "random-junk" | 3
+    "cass-nccpintegration-random-junk-v003"                   | [(STACK): "myStack"]                  || "cass"       | "cass-myStack"                                       | "myStack"         | "random-junk" | 3
+    "cass-nccpintegration-random-junk-v003"                   | [(STACK): "myStack", (DETAIL): ""]    || "cass"       | "cass-myStack"                                       | "myStack"         | ""            | 3
```
jtk54: What is the particular use tested in this case? Is it OK to have empty string detail -> empty string expected detail @jervi ?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/names/GoogleLabeledResourceNamer.java
```diff
@@ -99,9 +99,6 @@ private static String getClusterName(String app, String stack, String detail) {
 
   private static <T> void setIfPresent(Consumer<T> setter, T value) {
     if (value != null) {
-      if (value.equals("")) {
-        value = null;
-      }
```
jervi: It is actually this way on purpose. The use case is if you have some legacy infrastructure that is named in a way that Frigga parses as something, you can overwrite that part by setting an empty label. Without this feature, you can only overwrite it with another value, not remove/ignore it. E.g. if you have a legacy instance group named `myApp-foo-bar` and you want it to appear in Spinnaker as `myApp-foo`, you can do that by setting an empty `spinnaker-moniker-detail` label.

clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/names/GoogleLabeledResourceNamerSpec.groovy
jervi: See my comment above

jervi: But I guess it's OK. The result will be the same, I think.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/names/GoogleLabeledResourceNamer.java
jtk54: I think you could omit/delete the `details` moniker label (while including the others) to remove the detail from the legacy server group.

clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/names/GoogleLabeledResourceNamerSpec.groovy
jtk54: :+1:

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/names/GoogleLabeledResourceNamer.java
jervi: Yeah. I think this block was something I should have removed earlier anyway and never should have ended up in the PR. Thanks for fixing :)

PR #3438, opened 2019-03-08
refactor(cf): Adopt artifact model for deploy stages

jkschneider:


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/ops/DeployCloudFoundryServerGroupAtomicOperation.java
```diff
@@ -120,45 +129,48 @@ private static CloudFoundryServerGroup createApplication(DeployCloudFoundryServe
...
+      () -> client.getApplications().packageUploadComplete(packageId),
+      Function.identity(), null, getTask(), description.getServerGroupName(), PHASE);
+
+    getTask().updateStatus(PHASE, "Completed creating package for application '" + description.getServerGroupName() + "'");
 
     return packageId;
   }
 
   private void buildDroplet(String packageId, String serverGroupId, DeployCloudFoundryServerGroupDescription description) {
-    final CloudFoundryClient client = description.getClient();
```
claymccoy: why remove final?

jkschneider: why should final be on a local variable?

PR #3442, opened 2019-03-08
chore(kubernetes): Improved error messages on failing deployments

german-muzquiz:
When a `Deploy (Manifest)` stage fails because of a misconfigured kubernetes account (like lack of permissions in `kubeconfig` file), UI shows a misleading error message about a "notValidKind", where it hides the underlying issue:

default screen shot 2019-03-08 at 10 49 34 am
 
The code keeps track of kubernetes kinds that it can and cannot access using lists. When clouddriver starts, by default it tries to read all known kinds, and in case of errors, it adds them to the `omitKinds` list.

This PR changes the list of `omitKinds` to a map, where the value is a string explaining the reason for being omitted. This string is used to render the final error message to users:

screen shot 2019-03-08 at 11 38 06 am

screen shot 2019-03-08 at 11 20 51 am

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
@@ -131,22 +131,14 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
   }
 
   public boolean isValidKind(KubernetesKind kind) {
-    if (kind == KubernetesKind.NONE) {
-      return false;
-    } else if (!this.kinds.isEmpty()) {
-      return kinds.contains(kind);
-    } else if (!this.omitKinds.isEmpty()) {
-      return !omitKinds.keySet().contains(kind);
-    } else {
-      return true;
-    }
+    return getInvalidKindReason(kind) != null;
   }
 
   public InvalidKindReason getInvalidKindReason(KubernetesKind kind) {
     if (kind == KubernetesKind.NONE) {
       return InvalidKindReason.KIND_NONE;
-    } else if (!this.kinds.isEmpty() && !kinds.contains(kind)) {
-      return InvalidKindReason.MISSING_FROM_ALLOWED_KINDS;
+    } else if (!this.kinds.isEmpty()) {
+      return !kinds.contains(kind) ? InvalidKindReason.MISSING_FROM_ALLOWED_KINDS : null;
     } else {
       return this.omitKinds.getOrDefault(kind, null);
     }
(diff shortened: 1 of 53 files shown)


german-muzquiz:
cc @ezimanyi

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -115,15 +115,16 @@ public boolean getOnlySpinnakerManaged() {
 
   private final Path serviceAccountNamespacePath = Paths.get("/var/run/secrets/kubernetes.io/serviceaccount/namespace");
 
-  public boolean isValidKind(KubernetesKind kind) {
+  public Optional<String> checkIfInvalidKind(KubernetesKind kind) {
```
ezimanyi: The contract of this function has gotten a lot more complicated, changing from a simple `boolean` `isValid` to an `Optional` that all callers have to unwrap and negate.  We should think about how to support the need for additional info in `validateKind` without complicating the interface for all other callers.  As one possibility, `isValidKind` could keep the same contract, but another function `getInvalidReason` could be used in the one place where it's needed.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -455,12 +457,13 @@ private void determineOmitKinds() {
...
     log.info("Checking permissions on configured kinds for account {}... {}", accountName, allKinds);
-    Set<KubernetesKind> unreadableKinds = allKinds.parallelStream()
+    Map<KubernetesKind, String> unreadableKinds = allKinds.parallelStream()
       .filter(k -> k != KubernetesKind.NONE)
-      .filter(k -> !omitKinds.contains(k))
-      .filter(k -> !canReadKind(k, checkNamespace))
-      .collect(Collectors.toCollection(ConcurrentHashMap::newKeySet));
-    omitKinds.addAll(unreadableKinds);
+      .filter(k -> !omitKinds.keySet().contains(k))
+      .collect(Collectors.toConcurrentMap(k -> k, k -> checkIfKindError(k, checkNamespace))).entrySet().parallelStream()
```
ezimanyi: Should keep it as a `Stream` instead of calling `Collect` then immediately converting back to a `Stream`

german-muzquiz:
I think going with the enum of reasons is a good option if we don't want to show the underlying error in UI. Spotting the actual error from the log files has been proven not so easy to track, because by default all kinds are validated when clouddriver starts, so when a "Deploy (Manifest)" stage executes it doesn't actually log the real reason of why it failed.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -115,18 +114,44 @@ public boolean getOnlySpinnakerManaged() {
...
+    InvalidKindReason(String errorMessage) {
+      this.errorMessage = errorMessage;
+    }
+
+    public String getErrorMessage(KubernetesKind kind) {
+      return String.format(this.errorMessage, kind);
+    }
+  }
+
   public boolean isValidKind(KubernetesKind kind) {
```
ezimanyi: Could this be refactored now to just `return getInvalidKindReason(kind) != null`?  Just to make sure the two functions don't diverge?

german-muzquiz: That sounds good!

PR #3447, opened 2019-03-11
fix(aws): ensure any deployments objects are persisted

asher:


ajordens:
That is awesome, I totally didn't know this was happening :(

Wonder if we should also copy `createdArtifacts`.

PR #3453, opened 2019-03-11
fix(appengine): Allow suppression of sequence in deployed servergroup names

sbwsg:
There are some specific use cases that require an appengine version's URL remain completely stable between deployments. Spinnaker normally adds sequence strings to names, which changes the URL that appengine exposes with each deployment. This fixes that.


sbwsg:

PR #3456, opened 2019-03-12
fix(cloudformation): template body as string

ethanfrogers:
accept the cloudformation template body as a string which could either
bt JSON or YAML


ethanfrogers:
@xavileon PTAL!

PR #3458, opened 2019-03-12
feat(cloudformation): support capabilities

ethanfrogers:
adds support for passing cloudformation capabilities to stack
create/update. needed if you want to create IAM roles with
as part of your stack.


ethanfrogers:
@xavileon PTAL :)

PR #3462, opened 2019-03-13
chore(cf): remove usage of task fail and throw an exception

Jammy-Louie:


Jammy-Louie:
How an error would look like previously in the UI:
Screen Shot 2019-03-13 at 4 19 10 PM

How an error would look like with the proposed changes in the UI:
Screen Shot 2019-03-13 at 4 26 52 PM

PR #3464, opened 2019-03-13
fix(provider/kubernetes): Fix isValidKind

ezimanyi:
spinnaker/clouddriver#3442 inverted the check in isValidKind; fix this


ezimanyi:
@german-muzquiz : FYI.  I think the suggestion in my comment had this bug too, and I didn't catch it when reviewing the PR, but our integration tests did.

german-muzquiz:
> @german-muzquiz : FYI. I think the suggestion in my comment had this bug too, and I didn't catch it when reviewing the PR, but our integration tests did.

Glad it was caught, I would have thought that travis build would notice the bug.

ezimanyi:
@german-muzquiz : Looks like there wasn't any test coverage for this, so there was no way for Travis to notice.  I've added some in #3464 to prevent issues like this in the future.

PR #3468, opened 2019-03-14
feat(provider/azure): Enable Ssh public key for VM provisioning

mybayern1974:
With this change there will be two options to provision Azure VM Scale Set: Password and SSH public key. The switch of choosing one of them will be at halyard config when setting up an Azure account, so there will be a near future change at Halyard. Only the switch flag will be propagated from Halyard to CloudDriver, with no secure content being exposed


mybayern1974:
@duftler , if possible please help merge this PR at your earliest convenience given this issue is a blocker to one of our customers who only use ssh public key rather than pwd. Thank you in advance.

PR #3470, opened 2019-03-15
feat(amazon): delete snapshot atomic operation

emjburns:
Support for deleting snapshots that swabbie will use.


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/DeleteAmazonSnapshotAtomicOperation.java
```diff
@@ -0,0 +1,57 @@
+package com.netflix.spinnaker.clouddriver.aws.deploy.ops;/*
```
cfieber: weird spot for a header line ya got there... is this a long con to steal @robzienert 's thunder if he ever tries to rename packages?

emjburns: Intellij and I are no longer friends. It's trying to sabotage me

PR #3477, opened 2019-03-18
fix(provider/kubernetes): Properly set unstable flag for deployment

ezimanyi:
If a deployment reports its "Available" deployment condition as false, we currently mark the manifest as unavailable but don't mark it as unstable, which means that downstream stages will continue before it stabilizes.

In most cases the logic below for comparing replica sets will end up marking the deployment as unstable as well as unavailable, but there are edge cases were Kubernetes (correctly) marks the deployment as unavailable but the Spinnaker logic does not.

One example is if the new replicas have not been created yet (such as if we poll quickly). In that case the replica counts will all match in the below logic and we'll mark the deployment stable.

It might make sense to just remove the logic below as it just duplicates what Kubernetes is doing to determine the status (though as noted here, does not completely account for all edge cases).


ezimanyi:
For future reference, the logic that Kubernetes uses to determine of the deployment is complete: https://github.com/kubernetes/kubernetes/blob/b61baea13594f8cacd8e7db86aa68cbd967b9a5e/pkg/controller/deployment/util/deployment_util.go#L727

ezimanyi:

PR #3479, opened 2019-03-18
fix(cf): Destroy service should be idempotent

jasonjchu:
spinnaker/spinnaker#4137


jasonjchu:
idempotent-destroy-service

PR #3481, opened 2019-03-19
feat(cats): initial support for indexing resource by application

asher:
The upcoming sql backed cache implementation adds support for indexing
and retrieving resources by application. This can provide significant
performance improvements for Spinnaker's application clusters view when
application contain >thousands of instances and associated health
objects.

The new CacheData.attributes.application value is used by SqlCache for
this purpose. This PR selectively implements this in several of the AWS
caching agents. AmazonClusterProvider will take advantage when serving
/applications/{application}/serverGroups requests when the SqlCache
implementation is in use.

### DIFF
--- a/cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/Cache.java
+++ b/cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/Cache.java
@@ -19,11 +19,16 @@ package com.netflix.spinnaker.cats.cache;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Map;
 
 /**
  * Cache provides view access to data keyed by type and identifier.
  */
 public interface Cache {
+  enum StoreType {
+    REDIS, IN_MEMORY, SQL
+  }
+
   /**
    * Gets a single item from the cache by type and id
    *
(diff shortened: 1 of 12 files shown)


cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/Cache.java
```diff
@@ -115,4 +120,24 @@ default boolean exists(String type, String identifier) {
...
+
+  default Map<String, Collection<CacheData>> getAllByApplication(Collection<String> types,
+                                                                 String application,
+                                                                 Map<String, CacheFilter> cacheFilters) {
+    throw new InvalidCacheStateException("Method only implemented for StoreType.SQL");
+  }
+
+  default StoreType storeType() {
+    return StoreType.REDIS;
+  }
```
robzienert: I'd like to see method & type docs for all interface members going forward.

cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/CompositeCache.java
```diff
@@ -125,6 +127,25 @@ public CacheData get(String type, String id, CacheFilter cacheFilter) {
...
+                                                                  CacheFilter filter) {
+        return caches.iterator().next().getAllByApplication(type, application, filter);
+    }
+
+    @Override
+    public Map<String, Collection<CacheData>> getAllByApplication(Collection<String> types,
+                                                                  String application,
+                                                                  Map<String, CacheFilter> cacheFilters) {
+        return caches.iterator().next().getAllByApplication(types, application, cacheFilters);
+    }
```
robzienert: I think we should iterate all caches here, rather than blindly selecting the first, so that behavior of the class stays consistent across all of the methods. In the case of these specific methods being SQL-only, we'd probably do well to filter out any caches that don't successfully implement the new methods (via try/catch, likely).

cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/Cache.java
```diff
@@ -115,4 +120,24 @@ default boolean exists(String type, String identifier) {
    * @return the items matching the type and identifiers
    */
   Collection<CacheData> getAll(String type, String... identifiers);
+
+  default Map<String, Collection<CacheData>> getAllByApplication(String type, String application) {
+    throw new InvalidCacheStateException("Method only implemented for StoreType.SQL");
```
robzienert: I don't think this exception type is representative of the actual error. I'd rather see something like `UnsupportedCacheMethodException` or something?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/EddaLoadBalancerCachingAgent.groovy
```diff
@@ -88,6 +90,14 @@ class EddaLoadBalancerCachingAgent implements CachingAgent, HealthProvidingCachi
       String healthId = Keys.getInstanceHealthKey(ilb.instanceId, account.name, region, healthId)
       Map<String, Object> attributes = objectMapper.convertValue(ilb, ATTRIBUTES)
       Map<String, Collection<String>> relationships = [(INSTANCES.ns): [instanceId]]
+
+      CacheData instance = providerCache.get(INSTANCES.ns, instanceId, RelationshipCacheFilter.none())
+      if (instance != null) {
+        String application = instance.attributes.get("application")
+        if (application != null)
+          attributes.put("application", application)
```
robzienert: Stylistic preference here, but I prefer all conditionals to have squigglies, even if they're one-liners.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonClusterProvider.groovy
```diff
@@ -198,10 +256,78 @@ class AmazonClusterProvider implements ClusterProvider<AmazonCluster>, ServerGro
     if (application == null) {
       return null
     }
-    Collection<AmazonCluster> clusters = translateClusters(resolveRelationshipData(application, CLUSTERS.ns), includeDetails)
+
+    Collection<AmazonCluster> clusters
+
+    if (sqlApplicationIndexEnabled &&
+      ((cacheView.class == CompositeCache.class &&
+      (cacheView as CompositeCache).getStoreTypes().every { (it == SQL) }) ||
```
robzienert: This seems like a risky cast? Are all Caches implementations of CompositeCache?

cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/CompositeCache.java
asher: Made it consistent but left out the try/catch. The sql config only allows `CompositeCache` to be write up with a single provider of type `SqlProviderCache`. We don't actually want the functionality of `CompositeClass`, it just happens to get wired by default in all cases currently. Will address that as a separate issue.

asher:
@robzienert Good comment re: testing. Tests for the byApplication methods are going into the netflix internal repo since the functionality is dependent on `cats-sql`, but they'll come over to OSS soon.

PR #3482, opened 2019-03-19
fix(kubernetes): Infer API version when missing on patch manifest

ezimanyi:
* fix(kubernetes): Infer API version when missing on patch manifest 

  The patch manifest stage is currently failing, because we're fetching the apiGroup as a required field. If there is no apiGroup specified, we should try to match the supplied kind against native groups.

* refactor(kubernetes): Add optional apiGroup to KubernetesKind 

  As I was working on the last commit, I found it a bit confusing knowing whether a given 'kind' string I was dealing with was a fully-qualified kind (with API group) or just the kind string.

  To make this less ambiguous, include 'apiGroup' as a field on the KubernetesKind object, and handle scoping as part of toString and fromString so this happens in one place.

  This adds the default api group for built-in kinds, but explicitly considers two built-in kinds as matching even if the API group doesn't to handle cases such as deployments which can exist in either apps or extensions.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesApiGroup.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesApiGroup.java
@@ -58,8 +58,8 @@ public class KubernetesApiGroup {
     return name;
   }
 
-  public boolean isCustomResourceGroup() {
-    return !NATIVE_GROUPS.contains(this);
+  public boolean isNativeGroup() {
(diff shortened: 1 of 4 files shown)


ezimanyi:
@willgorman : The first commit fixes the bug; the second one is a small refactor that keeps the API group object on the `KubernetesKind` object itself and moves the string parsing to only occur in the `toString` and `fromString` methods.  The second commit is not necessary to fix the bug, but I think makes it a bit cleaner---let me know your thoughts.

ezimanyi:
Merged this to fix the integration tests, but happy to do a follow-up if there are further comments.

willgorman:
👍 I think the refactor greatly improves the clarity.

ezimanyi:

PR #3485, opened 2019-03-20
feat(provider/azure): Add instance tags in AzureServerGroupDescription

michaeljqzq:
* Add property `instanceTags` for user defined tags
* Remove unnecessary pre-defined tags since the upper limit of tag number is 15


michaeljqzq:
This PR is linked with https://github.com/spinnaker/deck/pull/6722

PR #3486, opened 2019-03-20
fix(cf): Remove "//" from routes

stuart-pollock:
spinnaker/spinnaker#4145


jkschneider:

PR #3489, opened 2019-03-20
fix(cf): Unmap LBs too aggressive

jasonjchu:
Attempting to unmap a load balancer from a single server group would remove it universally.  It should be limited to a single version of an application.

spinnaker/spinnaker#4164


jkschneider:

PR #3496, opened 2019-03-21
fix(provider/cf): Form-based manifest input for clone server group

jkschneider:
We have to handle form-based manifest input differently for clone operations because they pass through clone logic in Orca that is common to all providers.


jkschneider:

PR #3499, opened 2019-03-22
fix(titus): remove TitusHealthIndicator

asher:
Titus master availability shouldn't impact clouddriver startup or availability. Nothing used TitusHealthIndicator, apart from it being wired into spring.


ajordens:
👍 

I'm in favor of not having Titus knock the entirety of `clouddriver` OOS.

PR #3501, opened 2019-03-25
feat(kuberenetes): v2 run job stage

ethanfrogers:
adds support for run job in kubernetes v2 provider. very similar to the
Deploy (Manifest) stage but only supports the Job kind. satisifies the
methods that are needed by the Run Job stage to start, monitor and
collect jobs.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/KubernetesModelUtil.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/KubernetesModelUtil.java
@@ -30,8 +30,12 @@ import java.util.stream.Collectors;
 @Slf4j
 public class KubernetesModelUtil {
   public static long translateTime(String time) {
+    return KubernetesModelUtil.translateTime(time, "yyyy-MM-dd'T'HH:mm:ssX");
+  }
+
+  public static long translateTime(String time, String format) {
     try {
-      return StringUtils.isNotEmpty(time) ? (new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssX").parse(time)).getTime() : 0;
+      return StringUtils.isNotEmpty(time) ? (new SimpleDateFormat(format).parse(time)).getTime() : 0;
     } catch (ParseException e) {
       log.error("Failed to parse kubernetes timestamp", e);
       return 0;
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/view/KubernetesJobProvider.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/view/KubernetesJobProvider.groovy
@@ -22,17 +22,19 @@ import com.netflix.spinnaker.clouddriver.kubernetes.v1.security.KubernetesV1Cred
 import com.netflix.spinnaker.clouddriver.model.JobProvider
 import com.netflix.spinnaker.clouddriver.model.JobState
 import com.netflix.spinnaker.clouddriver.security.AccountCredentialsProvider
-import org.springframework.beans.factory.annotation.Autowired
+import io.fabric8.kubernetes.api.model.Pod
 import org.springframework.stereotype.Component
+
 
 @Component
 class KubernetesJobProvider implements JobProvider<KubernetesJobStatus> {
   String platform = "kubernetes"
 
-  @Autowired
   AccountCredentialsProvider accountCredentialsProvider
 
-  KubernetesJobProvider() { }
+  KubernetesJobProvider(AccountCredentialsProvider accountCredentialsProvider) {
+    this.accountCredentialsProvider = accountCredentialsProvider
+  }
 
   @Override
   KubernetesJobStatus collectJob(String account, String location, String id) {
@@ -41,7 +43,7 @@ class KubernetesJobProvider implements JobProvider<KubernetesJobStatus> {
       return null
     }
     def trueCredentials = (credentials as KubernetesNamedAccountCredentials).credentials
-    def pod = trueCredentials.apiAdaptor.getPod(location, id)
+    Pod pod = trueCredentials.apiAdaptor.getPod(location, id)
     def status = new KubernetesJobStatus(pod, account)
 
     String podName = pod.getMetadata().getName()
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/converter/job/KubernetesRunJobOperationConverter.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/converter/job/KubernetesRunJobOperationConverter.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2019 Armory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.kubernetes.v2.converter.job;
+
+import com.netflix.spinnaker.clouddriver.kubernetes.KubernetesOperation;
+import com.netflix.spinnaker.clouddriver.kubernetes.v1.deploy.converters.KubernetesAtomicOperationConverterHelper;
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.view.provider.KubernetesV2ArtifactProvider;
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.KubernetesResourcePropertyRegistry;
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.job.KubernetesRunJobOperationDescription;
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.op.job.KubernetesRunJobOperation;
+import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation;
+import com.netflix.spinnaker.clouddriver.security.AbstractAtomicOperationsCredentialsSupport;
+import com.netflix.spinnaker.clouddriver.security.ProviderVersion;
+import org.springframework.stereotype.Component;
+
+import java.util.Map;
+
+import static com.netflix.spinnaker.clouddriver.orchestration.AtomicOperations.RUN_JOB;
+
+@KubernetesOperation(RUN_JOB)
+@Component
+public class KubernetesRunJobOperationConverter extends AbstractAtomicOperationsCredentialsSupport {
+
+  private KubernetesResourcePropertyRegistry registry;
+
+  private KubernetesV2ArtifactProvider artifactProvider;
+
+  public KubernetesRunJobOperationConverter(KubernetesResourcePropertyRegistry registry, KubernetesV2ArtifactProvider artifactProvider) {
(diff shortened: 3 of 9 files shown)


ethanfrogers:
@ezimanyi PTAL!

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/model/KubernetesJobStatus.groovy
```diff
@@ -97,4 +116,37 @@ class KubernetesJobStatus implements JobStatus, Serializable {
...
+
+    int completions = 1
+    V1JobSpec spec = job.getSpec()
+    if (spec != null && spec.getCompletions() != null) {
+      completions = spec.getCompletions()
+    }
+    int succeeded = 0
+    if (status.getSucceeded() != null) {
+      succeeded = status.getSucceeded()
+    }
```
robzienert: In groovy, you could trim this up:

```
int succeeded = status.succeeded ?: 0
```

... Or via a more Java approach:

```
int succeeded = Optional.of(status.getSucceeded()).orElse(0)
```

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/view/KubernetesJobProvider.groovy
```diff
@@ -65,6 +80,32 @@ class KubernetesJobProvider implements JobProvider<KubernetesJobStatus> {
     return status
   }
 
+  KubernetesJobStatus collectV2Job(KubernetesV2Credentials credentials, String account, String location, String id) {
+    List<Manifest> manifests = manifestProviders.stream()
+      .collect{ provider ->
+        provider.getManifest(account, location, id)
+    }.findResults { manifest -> manifest }
+
+    if (!(manifests.size() > 0)) {
```
robzienert: `manifests.isEmpty()`

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/view/KubernetesJobProvider.groovy
```diff
@@ -19,27 +19,42 @@ package com.netflix.spinnaker.clouddriver.kubernetes.v1.provider.view
...
   KubernetesJobStatus collectJob(String account, String location, String id) {
     def credentials = accountCredentialsProvider.getCredentials(account)
-    if (!(credentials?.credentials instanceof KubernetesV1Credentials)) {
-      return null
+    if (credentials?.credentials instanceof KubernetesV2Credentials) {
+      return collectV2Job(credentials.credentials, account, location, id)
     }
+
     def trueCredentials = (credentials as KubernetesNamedAccountCredentials).credentials
     def pod = trueCredentials.apiAdaptor.getPod(location, id)
```
ezimanyi: We may want to be explicit about the type as `Pod` here as I think Groovy can't infer the type and may call the wrong constructor (now that there are two).

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/view/KubernetesJobProvider.groovy
```diff
@@ -19,27 +19,42 @@ package com.netflix.spinnaker.clouddriver.kubernetes.v1.provider.view
...
+  List<ManifestProvider> manifestProviders
+
   KubernetesJobProvider() { }
 
   @Override
   KubernetesJobStatus collectJob(String account, String location, String id) {
     def credentials = accountCredentialsProvider.getCredentials(account)
-    if (!(credentials?.credentials instanceof KubernetesV1Credentials)) {
-      return null
+    if (credentials?.credentials instanceof KubernetesV2Credentials) {
```
ezimanyi: Rather than doing an `if/else` on the type of the `credentials` it probably makes sense to just make a new `KubernetesV2JobProvider` that implements the `JobProvider<KubernetesJobStatus>` interface.  It looks like there's actually no shared logic between the V1/V2 implementations, so having them in different classes will make it more maintainable.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/converter/job/KubernetesRunJobOperationConverter.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.Map;
+
+import static com.netflix.spinnaker.clouddriver.orchestration.AtomicOperations.RUN_JOB;
+
+@KubernetesOperation(RUN_JOB)
+@Component
+public class KubernetesRunJobOperationConverter extends AbstractAtomicOperationsCredentialsSupport {
+
+  @Autowired
```
ezimanyi: Constructor injection is generally preferred over field injection for new classes (though we have a lot of existing field injection around).

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubernetesRunJobOperation.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    if (jobSpec.get("metadata") == null) {
+      jobSpec.put("metadata", new HashMap<>());
+    }
+
+    if (!this.description.getNamespace().isEmpty()){
+      jobSpec.setNamespace(this.description.getNamespace());
+    }
+    // append a random string to the job name to avoid collision
```
ezimanyi: Is this to allow concurrent jobs?  It looks like we're deleting the job after it's done.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubernetesRunJobOperation.java
```diff
@@ -0,0 +1,109 @@
...
+    moniker.setApp(description.getApplication());
+
+    deployManifestDescription.setManifests(manifests);
+    deployManifestDescription.setSource(KubernetesDeployManifestDescription.Source.text);
+    deployManifestDescription.setMoniker(moniker);
+    deployManifestDescription.setCredentials(description.getCredentials());
+
+    KubernetesDeployManifestOperation deployManifestOperation = new KubernetesDeployManifestOperation(deployManifestDescription, registry, provider);
+    OperationResult operationResult = deployManifestOperation.operate(new ArrayList());
+    DeploymentResult deploymentResult = new DeploymentResult();
```
ezimanyi: I wonder if we could pull this logic out so that the run job stage first has a "generate manifest" task then, a "deploy manifest" task that deploys that.  Not sure exactly what that would look like, but maybe that would work?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/model/KubernetesJobStatus.groovy
```diff
@@ -40,6 +44,8 @@ class KubernetesJobStatus implements JobStatus, Serializable {
   Set<String> securityGroups
   @JsonIgnore
   Pod pod
+  @JsonIgnore
```
ezimanyi: I wonder if this should be split into two classes that implement a `KubernetesJobStatus` interface...rather than have a pair of fields where exactly one has to be set (and the have the logic check which is null).

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/converter/job/KubernetesRunJobOperationConverter.java
ethanfrogers: I know this but always forget it for some reason! Doh!

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubernetesRunJobOperation.java
ethanfrogers: Correct. Jobs are immutable once created which would prevent you from running them concurrently. I'm open to suggestions on how better to accomplish this, though. This is what the V1 provider does.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubernetesRunJobOperation.java
ethanfrogers: Perhaps. I feel like that would take a fair bit of refactoring in the `RunJobStage` to understand the difference between Kubernetes V1/V2 and Titus.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/KubernetesModelUtil.java
```diff
@@ -30,8 +30,12 @@
 @Slf4j
 public class KubernetesModelUtil {
   public static long translateTime(String time) {
+    return KubernetesModelUtil.translateTime(time, "yyyy-MM-dd'T'HH:mm:ssX");
```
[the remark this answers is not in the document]
ethanfrogers: I thought the same thing but the format that was used with the Fabric8 models doesn't work with the `client-java` models, for some reason.

ethanfrogers:
@ezimanyi i've addressed your comments! lmk if this needs anything else!

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/KubernetesModelUtil.java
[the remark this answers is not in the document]
ezimanyi: Interesting, maybe Fabric8 was doing something to translate

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubernetesRunJobOperation.java
ezimanyi: I think appending the string is fine, no need to change anything.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubernetesRunJobOperation.java
ezimanyi: Yeah, it's probably not worth the effort---the other things above about making separate classes were more obvious refactors, this one has less of an obvious benefit.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/provider/view/KubernetesV2JobProvider.java
```diff
@@ -0,0 +1,102 @@
...
+import lombok.Data;
+import org.apache.commons.lang.NotImplementedException;
+import org.springframework.stereotype.Component;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+@Data
```
ezimanyi: Are we using the getters/setters generated by this?  Looks like `@Getter` on the `platform` attribute is all we need, which would leak less of the implementation.

ethanfrogers: I don't think so, actually. Will remove! Thanks.

PR #3512, opened 2019-03-27
feat(provider/kubernetes): Add Tekton CR types to KubernetesKind

abergmeier:
Add Tekton CR types for working with them.


ezimanyi:
@abergmeier : The current pattern for Kubernetes caching is to only hard-code Kubernetes-provided kinds, and to leave any custom resource definitions to be configured by the end user via the `customResources` parameter on their Kubernetes account.

It looks like configuring custom resources is not well documented, but the [PR that implemented it](https://github.com/spinnaker/clouddriver/pull/2354) has some useful info on how to configure your account to cache custom resources.

There are a lot of interesting opportunities for support for Tekton CRDs in Spinnaker---if/when Spinnaker has first-class support for these concepts, I could see the argument for adding them as built-in kinds.  But for now, I don't see a strong argument for including these in the list vs. any other popular CRDs that many users have installed.

abergmeier:
IMO this should be a precursor to implementation of Handler for Tekton (according to https://www.spinnaker.io/guides/developer/crd-extensions/).
@ethanfrogers WDYT?

skim1420:
Has this general area of interaction with Tekton been discussed in larger circle, and with the TOC? Please hold off on any implementation until there has been agreement.

ezimanyi:
Closing this for now based on the above discussion, we can re-open/reconsider once this has been discussed more broadly.

PR #3515, opened 2019-03-27
feat(artifacts): Support relative chart urls

dongo0223:
spinnaker/spinnaker#4078

Helm supports relative chart urls in index.yaml. 
https://github.com/helm/helm/blob/75b4afaa33a02ff01f041cf58a01535bca770cd6/pkg/repo/chartrepo.go#L252


clouddriver-artifacts/src/test/java/com/netflix/spinnaker/clouddriver/artifacts/helm/IndexParserTest.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+public class IndexParserTest {
+	
+	private String buildIndexYaml(String chartName, String version, List<String> urls) {
+		StringBuilder indexYamlBuilder = new StringBuilder();
+		indexYamlBuilder
```
ezimanyi: Might have been a bit easier to read to just construct as an object, then convert to yaml at the end

PR #3516, opened 2019-03-28
feat(moveTitusTasks): Add support for tasks moving between jobs

aravindmd:
- Adding support for tasks moving between jobs
- Bump titus api definition


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
```diff
@@ -322,6 +328,18 @@ private void updateTask(StreamingCacheState state, Task task) {
       state.changes.incrementAndGet();
     }
 
+    private void updateMovedTask(StreamingCacheState state, Task task, String destinationJobId,
+                                 String sourceJobId) {
+      if (sourceJobId != null) {
+        if (state.tasks.containsKey(sourceJobId)) {
+          state.tasks.get(destinationJobId).add(task);
+          state.tasks.get(sourceJobId).remove(task);
```
asher: Check that `source.tasks.containsKey(sourceJobId)` before removing. This doesn't look like it's necessary, but there isn't a guarantee of event ordering or only-once delivery. Note the "..but task not present in current snapshot set" debug logging in `updateTask()` - we see this fire in prod occasionally as a result.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
```diff
@@ -322,6 +328,18 @@ private void updateTask(StreamingCacheState state, Task task) {
       state.changes.incrementAndGet();
     }
 
+    private void updateMovedTask(StreamingCacheState state, Task task, String destinationJobId,
+                                 String sourceJobId) {
+      if (sourceJobId != null) {
+        if (state.tasks.containsKey(sourceJobId)) {
+          state.tasks.get(destinationJobId).add(task);
```
asher: If the destination job doesn't already have running tasks,`state.tasks.get(destinationJobId)` will return null and the add would throw an NPE. `state.tasks.computeIfAbsent(jobId, t -> new HashSet<>()).add(task)` or something similar would be safer.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
```diff
@@ -217,6 +217,13 @@ public void executeAgent(Agent agent) {
                   updateJob(state, notification.getJobUpdate().getJob());
                   break;
                 case TASKUPDATE:
+                  if (notification.getTaskUpdate().getMovedFromAnotherJob()) {
+                      Task task = notification.getTaskUpdate().getTask();
+                      String destinationJobId = task.getJobId();
+                      String sourceJobId = task.getTaskContextOrDefault("task.movedFromJob", null);
+                      log.info("{} task moved from another job ", sourceJobId);
```
asher: May be helpful to include the taskId and source/dest job ids, something like "task {} moved from job {} to {}".

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
asher: Since `updateTask()` is still called after this method, you don't need to do anything wrt the destination job here. Can remove this line, as well as the `state.updatedJobs.add(destinationJobId)`.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
aravindmd: updated per your inputs

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
aravindmd: done

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
aravindmd: fixed

PR #3517, opened 2019-03-28
fix(core): add `cloudProvider` to `LoadBalancerServerGroup` as expect…

maggieneterval:
…ed by deck

Closes https://github.com/spinnaker/spinnaker/issues/4074

Before:
1748bd74-20a7-4400-be97-58550867a49e

After:
c7c2c2c1-85ac-491b-a307-44a7dc31adb2

Deck expects `LoadBalancerServerGroup`s to have a `cloudProvider` field in order to correctly route to the Server Group side-view from the Load Balancers tab (per code [here](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/loadBalancer/LoadBalancerServerGroup.tsx#L56)).


maggieneterval:
Just updated PR description with screen captures of the behavior this enables.

maggieneterval:
@ezimanyi I did a global search for `new LoadBalancerServerGroup` and `LoadBalancerServerGroup.builder` which is the only way this java dummy knows how to create an new instance of a class...but this did fix the UI issue for GCE and Appengine.

ezimanyi:
@maggieneterval : You're entirely correct, ignore my last comment.  For some reason I thought there were other constructors, but I'm not sure what I was looking at.  Sorry about that!

maggieneterval:
@ezimanyi well I'm glad you asked because I double checked and found one I'd missed the first time!

maggieneterval:

PR #3520, opened 2019-03-28
fix(cf): Clone server group sets environment variables in new server group

jasonjchu:
spinnaker/spinnaker#4211


jkschneider:
@Jammy-Louie Please test this end-to-end one last time with the latest changes.

Jammy-Louie:
@jkschneider I just verified it. The environment variables are set

jkschneider:

PR #3531, opened 2019-04-01
 feat(kubernetes): Use streaming job executor for k8s manifests

ezimanyi:
This PR builds on an idea that @benjaminws implemented in DataDog/clouddriver#15 to use a streaming JSON parser to parse the output of `kubectl` in order to reduce the amount of data we're copying around for large clusters.  This change pushes the support for streaming output down to the `JobExecutorLocal` class, and then updates the `KubectlJobExecutor` to use this new support.

* feat(core): Add streaming Job functionality 

  Currently the LocalJobExecutor accumulates all output from the job into a stream and returns the entire output (as a string) to the calling function.  In some cases, it would be useful to allow the calling function to process the output as it is produced to reduce the amount of output that needs to be stored (and to reduce the number of times that it is copied around).

  Add an optional parameter to runJob specifying a function that will consume the output as it is produced. When this parameter is passed, all output will be passed to the supplied callback, and the output on the JobResult will be null.

* feat(kubernetes): Use streaming job executor for k8s manifests 

  When fetching lists of kubernetes manifests, the returned data can be quite large for large clusters. Reduce the amount of times this data needs to be copied by using a streaming JSON parser to parse the output as it is produced.


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/jobs/local/StreamConsumer.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.jobs.local;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+public interface StreamConsumer {
```
ianferguson: Who is responsible for closing an InputStream passed to an implementation of this interface? The implementation or the caller? Can documentation on that contract be added?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubectlJobExecutor.java
```diff
@@ -589,6 +586,26 @@ private Void patch(KubernetesV2Credentials credentials, KubernetesKind kind, Str
     return null;
   }
 
+  private StreamConsumer parseManifestList(List<KubernetesManifest> manifestList) {
+    return (InputStream is) -> {
+      JsonReader reader = new JsonReader(new InputStreamReader(is, StandardCharsets.UTF_8));
```
ianferguson: if we're always immediately assuming UTF_8 encoding and wrapping the inputstream in `StreamConsumer`s, would it make more sense to have them be `ReadableConsumer=` or `ReaderConsumer` and only have to deal in characters and not bytes?

If they do need to stay as InputStream consumers, does the encoding need to come from `Charset.defaultCharset()` to avoid any issues on systems that are not using UTF-8 character encoding for standard out. Admittedly, I think its unlikely that in practice this is being run on any systems that that would be the case on, so feel free to ignore

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubectlJobExecutor.java
```diff
@@ -332,7 +333,8 @@ public KubernetesManifest get(KubernetesV2Credentials credentials, KubernetesKin
       command.add("-l=" + selectors.toString());
     }
 
-    JobStatus status = jobExecutor.runJob(new JobRequest(command));
+    List<KubernetesManifest> result = new ArrayList<>();
```
ianferguson: if this List is being written to potentially on a different thread by the JobExecutor, does this List need to be a ConcurrentList to ensure there's no visibility edge cases with items added on the JobExecutor thread not showing up on this thread?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/jobs/JobExecutor.java
```diff
@@ -15,6 +15,33 @@
...
+
+  /**
+   * Runs the specified JobRequest, streaming output to the supplied StreamConsumer, and returns the status and standard
+   * error in a JobStatus object.  The returned JobStatus will have standard output set to null, as the output will have
+   * already been consumed by the supplied consumer.
+   * @param jobRequest The job request
+   * @param streamConsumer A callback that consumes the job's standard output as it is produced
+   * @return The result of the job
+   */
+  JobStatus runJob(JobRequest jobRequest, StreamConsumer streamConsumer);
```
ianferguson: If StreamConsumer were changed to something along the lines of a `Function<InputStream, T>` and JobStatus were changed to have a type parameter for its output (which for `JobStatus runJob(JobRequest jobRequest);` would end up being `JobStatus<String>`), the results of the job in this case could be assembled within the function and returned as part of that all the way to the caller of `runJob` as a `JobStatus<List<KubernetesManifest>>` rather than having the `StreamConsumer` lambda close over a mutable collection in the callers scope in order to populate that collection via side effects from whatever call site the lambda gets called at.

That may also be overkill for this, but would likely make certain aspects of testing and future changes a bit more easy by keeping all the various state pretty tightly scoped and with less mutation across method calls

ianferguson: I also think an approach like that might make the implementation/code paths of the two methods on this interface more similar, and not necessarily require two separate executeJobStreaming and executeJob methods within `JobExecutorLocal.java`, but I haven't tried that out to be sure

ezimanyi: I like this idea a lot, and I think it will fix a few of the other comments you've noted below.  I'll give this a try, then will revisit the other comments to see which are still relevant.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/jobs/local/StreamConsumer.java
ezimanyi: Pushed this responsibility to the implementing class and added this to the class documentation.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/jobs/local/JobExecutorLocal.java
```diff
@@ -73,6 +82,31 @@ private ExecuteResult executeJob(JobRequest jobRequest) throws IOException {
...
+    DefaultExecuteResultHandler resultHandler = new DefaultExecuteResultHandler();
+    executor.execute(jobRequest.getCommandLine(), jobRequest.getEnvironment(), resultHandler);
+
+    try (PipedInputStream is = new PipedInputStream(stdOut)) {
+      consumer.consume(is);
+    }
+
+    try {
+      resultHandler.waitFor();
+    } catch (InterruptedException e) {
```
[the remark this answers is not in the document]
ezimanyi: Great point; I've updated the `catch` block to:
* kill the forked job
* reset the interrupt flag on the thread

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/jobs/local/JobExecutorLocal.java
```diff
@@ -90,5 +124,13 @@ private Executor buildExecutor(ExecuteStreamHandler streamHandler) {
     final OutputStream stdErr;
     final int exitValue;
     final boolean wasKilled;
+
+    String getStdOut() {
+      return Optional.ofNullable(stdOut).map(Objects::toString).orElse(null);
```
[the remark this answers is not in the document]
ezimanyi: Yes, this was originally always a `ByteArrayOutputStream` so `toString()` worked correctly, then as part of the refactor I made the type an `OutputStream` (but only ever called `.toString()` in the case where it was a `ByteArrayOutputStream`).  That was a poor design, as we were effectively relying on a specific implementation of the class.   The refactor to use generics in the interface of `JobResult` sidesteps this entire issue.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubectlJobExecutor.java
ezimanyi: Great point...I hadn't made it concurrent as it's only ever being written to by a single thread, but did not enforce any "happens-after" semantics between writing to the list in one thread and reading from it in the other.  Now that `runJob` returns the results rather than modifying a variable in the calling thread, this is no longer an issue.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubectlJobExecutor.java
ezimanyi: I think it makes sense to remove the responsibility for converting characters to bytes from the implementation and just give them a `Reader`.  If at some point we need a consumer that handles a stream of bytes, we can refactor to account for that, but I don't forsee that so it doesn't make sense to make this overly general.

I'd imagine non-UTF-8 is pretty unlikely, though it's probably better to just use the default encoding, which I've done as part of moving this translation to `JobExecutorLocal`.

ezimanyi:
Thanks a lot @ianferguson for the detailed feedback!  Much appreciated.  I've pushed a few commits that I believe should address all of the points---please let me know if you have any further comments!

* refactor(core): Improve interface of JobExecutor 

  * JobStatus is now parametrized and can return any type in its
   stdOut field.  (This field will be renamed output in a follow-up
   commit.)
  * StreamConsumer is now parametrized and returns the results of
   its processing on the output stream as a type T instead of
   mutating an array in the calling method.
  * Pushed a flag indicating if the job was killed up to JobStatus,
   which eliminated the need for the (almost identical) ExecuteResult
  * Changed some of the control flow in JobExecutorLocal.
  * Explicitly stated that implementations of StreamConsumer must close
   the stream, and updated the existing implemenation to do so.

* refactor(core): Improve interrupt handling 

  If the thread is interrupted, we should kill the forked job that is running, and set the interrupt flag on the thread before returning.

* refactor(core): Rename JobStatus and member variables 

  JobStatus is really holding the result of a job; it is never generated until the job is finished.  Rename it to JobResult.

  Now that stdOut can be either stdOut or some processing of the output, rename it to the more general output.  Also rename stdErr as error, though either would probably be appropriate for it.

* refactor(core): Change StreamConsumer to ReaderConsumer 

  We'll likely only ever want to read characters from a job, so push the conversion from InputStream -> InputStreamReader to the JobExecutor.

  Also:
  * Don't explicitly specify a character class; just use the default
  * buffer the input stream before passing it to the ReaderConsumer

* refactor(core): A few minor changes 

  * Actually close the JsonReader (and thus the underlying reader/stream),
   which I thought I had done in an earlier commit.
  * Rename ResultSupplier to RequestExecutor which is a bit more
   descriptive.

ianferguson:
@ezimanyi last round of changes looks really good! I noticed on re-reading it that `  <T> JobResult<T> runJob(JobRequest jobRequest, ReaderConsumer<T> readerConsumer);` doesn't appear to be called anywhere now? Is there some slight of polymorphic hand I'm missing that results in the `List<KubernetesManifest>` jobs getting routed through to the streaming JobExecutor call in the current code?

ezimanyi:
@ianferguson : Thanks again for reviewing!

The two calls to `<T> JobResult<T> runJob(JobRequest jobRequest, ReaderConsumer<T> readerConsumer)` are [here](https://github.com/spinnaker/clouddriver/blob/16f2ba364366a58869b49467fa59f41968869bec/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubectlJobExecutor.java#L332) and [here](https://github.com/spinnaker/clouddriver/blob/16f2ba364366a58869b49467fa59f41968869bec/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubectlJobExecutor.java#L309).  I think that Github isn't expanding the diff on `KubectlJobExecutor` as it's reasonably large.  (Probably a sign that some of the mechanical changes I made in that file can be factored out, but that's for another day.)

ianferguson:
@ezimanyi :+1: thanks!

PR #3537, opened 2019-04-03
feat(cf): Implement on-demand caching

stuart-pollock:
- for performance reasons, we introduced the caching of individual
CloudFoundryServerGroup objects

spinnaker/spinnaker#4083

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
@@ -16,14 +16,17 @@
 
 package com.netflix.spinnaker.clouddriver.cloudfoundry.client;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.netflix.frigga.Names;
 import com.netflix.spinnaker.clouddriver.cloudfoundry.client.api.ApplicationService;
 import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v2.ApplicationEnv;
 import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v2.MapRoute;
 import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v2.Resource;
+import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.*;
 import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.Package;
 import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.Process;
-import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.*;
 import com.netflix.spinnaker.clouddriver.cloudfoundry.model.*;
 import com.netflix.spinnaker.clouddriver.model.HealthState;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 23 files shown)


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
```diff
@@ -88,6 +117,14 @@ public CloudFoundryServerGroup findById(String guid) {
       .collect(toList());
   }
 
+  @Nullable
+  public CloudFoundryServerGroup findServerGroupByNameAndSpaceId(String name, String spaceId) {
```
jkschneider: `findServerGroupId` hits API, but `findById` uses the cache. Is there potential drift between them? If we are going to limit what `findServerGroupByNameAndSpaceId` can find to those things in the cache, can we instead just iterate over the cache looking for this name/space combination? And a view on this cache if you think this is called frequently enough to optimize?

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Organizations.java
```diff
@@ -70,4 +71,35 @@ public CloudFoundryOrganization findById(String orgId) throws CloudFoundryApiExc
...
+    // fully populates the space guid which is what Cloud Foundry's API expects as an input, not the name.
+    Optional<CloudFoundryOrganization> organization = findByName(space.getOrganization().getName());
+
+    Optional<CloudFoundrySpace> spaceOptional = organization
+      .map(org -> findSpaceByName(org, space.getName()));
+
+    spaceOptional.ifPresent(spaceCase -> {
+      if (!(space.getName().equals(spaceCase.getName()) &&
+        space.getOrganization().getName().equals(spaceCase.getOrganization().getName()))) {
+        throw new CloudFoundryApiException("Org or Space name not in correct case");
```
jkschneider: Is this possible? Is CAPI's name filter case insensitive? (https://apidocs.cloudfoundry.org/7.9.0/organizations/list_all_organizations.html)

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/provider/CloudFoundryProvider.java
```diff
@@ -50,7 +50,7 @@
     singletonMap(new SearchableResource(APPLICATIONS.getNs(), "cloudfoundry"), new ApplicationSearchResultHydrator());
 
   private final Map<String, String> urlMappingTemplates = emptyMap();
-  private final String id = "cloudfoundry";
+  public static final String PROVIDER_ID = "cloudfoundry";
```
jkschneider: I believe this is a breaking change somehow, that there are places that rely on `getId()` on the provider. Notice the `@Getter` on the type.

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Organizations.java
stuart-pollock: Yes, the CAPI name filter is case-insensitive.

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/provider/CloudFoundryProvider.java
stuart-pollock: If `CloudFoundryProvider` implemented `CloudProvider`, then there might be a problem.  It implements `SearchableProvider`, which does not have that method.

PR #3539, opened 2019-04-04
refactor(cf): Use Spring Cloud's refresh mechanism for refreshing providers & credentials

andreasevers:
Design doc: https://docs.google.com/document/d/1qrSxHMwZHrpiyAM25986a5V_rzfSOlVLhlC-BpOruyE/edit#

Config refreshing feels like reinventing the wheel because:
Spring offers environment change events, through the Actuator refresh endpoint,
ConfigurationProperties are refreshed upon calling the endpoint anyway (https://gist.github.com/dsyer/a43fe5f74427b371519af68c5c4904c7#configurationpropertiesrebinder).

The current way of discovering credentials and providers is using prototype beans. I think this can be simplified, while setting us up nicely for integration with e.g. the Spring Cloud Config server (or perhaps Halyard).

See https://github.com/spinnaker/spinnaker/issues/3969#issuecomment-479945060


andreasevers:
Closing this PR for now. We're working on an alternative approach.

PR #3540, opened 2019-04-04
fix(provider/cf): Fix clone from manifest artifact

jkschneider:
Form and artifact inputs for manifest are now sent to clouddriver as an "artifact". In the case of form input, it is orca's responsibility to convert the form to the manifest form and send it along to clouddriver as a base64 artifact.

image


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/artifacts/CloudFoundryArtifactCredentials.java
```diff
@@ -29,18 +29,20 @@
 
 @AllArgsConstructor
 @Getter
-public class PackageArtifactCredentials implements ArtifactCredentials {
-  private final String name = "package";
+public class CloudFoundryArtifactCredentials implements ArtifactCredentials {
+  public static final String TYPE = "cloudfoundry/app";
+
+  private final String name = "cloudfoundry";
```
claymccoy: CloudFoundryCloudProvider.ID is commonly used for this string

clouddriver-cloudfoundry/src/test/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/converters/DeployCloudFoundryServerGroupAtomicOperationConverterTest.java
```diff
@@ -102,131 +102,6 @@ void initializeClassUnderTest() {
     converter.setObjectMapper(new ObjectMapper());
   }
 
-  @Test
```
claymccoy: Why aren't these tests useful anymore? The class and methods being tested still exist even though they have been trimmed down.

jkschneider: It tests manifest parsing that has moved to Orca now.

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/artifacts/CloudFoundryArtifactCredentials.java
jkschneider: Thanks. Eventually this string will be per-account when we expand `CloudFoundryArtifactCredentials` to be an artifact account you can select on stage "Produces Artifacts". It could be `cheeseburgers` right now and the effect would be nil.

jkschneider:

PR #3541, opened 2019-04-04
fix(provider/cf): Get latest server group from API rather than cache

jkschneider:
The cache doesn't pick up on names fast enough on failures. So subsequent runs of deploy/clone can fail with a non-unique app name. By going to CAPI directly with a cheap API call, we can avoid the eventual consistency problem here.

image


jkschneider:

PR #3549, opened 2019-04-09
fix(cf): Fix server group name resolver

jkschneider:
The previous criteria selected server groups lexicographically less than `clusterName-v9999`, but this  selected `clusterName-prod-v046` because `p` < `v`.

We're now doing `<clusterName-v9999` and `>=clusterName`, selecting more than one result, filtering on matching cluster names, and returning all the taken server groups in the cluster.

Additionally, now filtering on `space_guid`.

cc / @ciberkleid

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
@@ -41,10 +41,12 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.Charset;
 import java.util.*;
+import java.util.stream.Collectors;
 
-import static com.netflix.spinnaker.clouddriver.cloudfoundry.client.CloudFoundryClientUtils.collectPages;
-import static com.netflix.spinnaker.clouddriver.cloudfoundry.client.CloudFoundryClientUtils.safelyCall;
+import static com.netflix.spinnaker.clouddriver.cloudfoundry.client.CloudFoundryClientUtils.*;
+import static java.util.Arrays.asList;
 import static java.util.Collections.*;
+import static java.util.Optional.ofNullable;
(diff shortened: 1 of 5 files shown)


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
```diff
@@ -380,15 +382,14 @@ public void setCurrentDroplet(String appGuid, String dropletGuid) throws CloudFo
...
-    return safelyCall(() -> api.appsLessThanName("name<" + clusterName + "-v9999", "desc", 1))
-      .flatMap(pr -> pr.getResources().stream()
-        .filter(app -> {
-          Names names = Names.parseName(app.getEntity().getName());
-          return clusterName.equals(names.getCluster()) && spaceId.equals(app.getEntity().getSpaceGuid());
-        })
-        .findAny())
-      .orElse(null);
+  public List<Resource<com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v2.Application>> getTakenSlots(String clusterName, String spaceId) {
+    List<String> filter = asList("name<" + clusterName + "-v9999", "name>=" + clusterName, "space_guid:" + spaceId);
```
claymccoy: Wouldn't '>=clusterName-v000' remove the ambiguity?

jkschneider: @claymccoy An application called `clusterName` deployed outside of Spinnaker is interpreted by Spinnaker as a v000 sequenced server group, so unfortunately we can't do that.

jkschneider: It still doesn't totally remove the ambiguity anyway. An app named `clusterName-zulu-v001` > `clusterName-v000` but needs to be filtered away by cluster name.

jkschneider:

PR #3551, opened 2019-04-09
 feat(provider/kubernetes): Add traffic options to deploy description

ezimanyi:
* refactor(provider/kubernetes): Improve some encapsulation 

  KubernetesDeployManifestDescription is a Data class, so the member variables can be private.

  KubernetesManifestTraffic is simple enough that using a builder is less clear; give it a single-arg constructor and remove setters.

  Also lightly refactor KubernetesDeployManifestOperationSpec to make it easier to add tests in the next commit.

* feat(provider/kubernetes): Add traffic options to deploy description 

  In order to support better first-class deployment strategies in the Kubernetes V2 provider, add functionality to the deploy manifest stage.  In particular:
  * Allow the services to be specified on the deploy description, rather than only as an annotation on the manifest.
  * Add a boolean to control whether to send traffic to the new workload, defaulting to true.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestAnnotater.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestAnnotater.java
@@ -205,9 +205,8 @@ public class KubernetesManifestAnnotater {
   public static KubernetesManifestTraffic getTraffic(KubernetesManifest manifest) {
     Map<String, String> annotations = manifest.getAnnotations();
 
-    return KubernetesManifestTraffic.builder()
-        .loadBalancers(getAnnotation(annotations, LOAD_BALANCERS, new TypeReference<List<String>>() {}, new ArrayList<>()))
-        .build();
+    List<String> loadBalancers = getAnnotation(annotations, LOAD_BALANCERS, new TypeReference<List<String>>() {}, new ArrayList<>());
+    return new KubernetesManifestTraffic(loadBalancers);
   }
 
   public static KubernetesCachingProperties getCachingProperties(KubernetesManifest manifest) {
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestAnnotater.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestAnnotater.java
@@ -209,6 +209,26 @@ public class KubernetesManifestAnnotater {
     return new KubernetesManifestTraffic(loadBalancers);
   }
 
+  public static void setTraffic(KubernetesManifest manifest, KubernetesManifestTraffic traffic) {
+    Map<String, String> annotations = manifest.getAnnotations();
+    List<String> loadBalancers = traffic.getLoadBalancers();
+
+    if (annotations.containsKey(LOAD_BALANCERS)) {
+      KubernetesManifestTraffic currentTraffic = getTraffic(manifest);
+      if (currentTraffic.getLoadBalancers().equals(loadBalancers)) {
+        return;
+      } else {
+        throw new RuntimeException(String.format(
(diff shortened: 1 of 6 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestTraffic.java
```diff
@@ -17,16 +17,14 @@
...
+import lombok.Getter;
 
 import java.util.List;
 
-@Data
-@Builder
 @AllArgsConstructor
-@NoArgsConstructor
+@EqualsAndHashCode
+@Getter
```
jtk54: `@Data` gives you both getters and setters for member fields here. Not sure if there's a difference when it comes to serializing this object.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestAnnotater.java
```diff
@@ -209,6 +209,26 @@ public static KubernetesManifestTraffic getTraffic(KubernetesManifest manifest)
...
+  public static void setTraffic(KubernetesManifest manifest, KubernetesManifestTraffic traffic) {
+    Map<String, String> annotations = manifest.getAnnotations();
+    List<String> loadBalancers = traffic.getLoadBalancers();
+
+    if (annotations.containsKey(LOAD_BALANCERS)) {
+      KubernetesManifestTraffic currentTraffic = getTraffic(manifest);
+      if (currentTraffic.getLoadBalancers().equals(loadBalancers)) {
+        return;
+      } else {
+        throw new RuntimeException(String.format(
```
jtk54: Would this prevent adding a new load balancer to an existing (already load balanced) replica set? Not sure about the k8s best practices/constraints, just wondering what the implications are here.

maggieneterval: I think the idea was that folks should either 1) have all relevant load balancers stored as metadata on the manifest in source control or 2) have all relevant load balancers selected in the UI and added by Spinnaker to the manifest, since any combination of the two where Spinnaker adds to or overrides existing load balancers might be unexpected

maggieneterval: When we write up docs on this we should make this clear

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestTraffic.java
ezimanyi: Yeah, removing the setters would make serialization no longer work without annotating the constructor, but we never actually serialize/deserialize this object.

jtk54: :+1:

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestAnnotater.java
jtk54: :+1:

ezimanyi: Good question.  We might also want to think about how this would interact with things like "find manifest" where a manifest already has another annotation...maybe at some point what you selected in the services section could just override the manifest if we need it for that feature.  But for the simple use cases I think this is simpler and we can revisit later.

jtk54: Sounds good.

PR #3552, opened 2019-04-10
 fix(provider/azure): Enable Azure Load Balancer from Azure VM Scale Set

neil-yechenwei:
Background:
Enable Azure Load Balancer which supports inbound NAT port-forwarding rules to connect to azure vm instances.
There is one another related PR (https://github.com/spinnaker/deck/pull/6829) in deck. Please also help to review.


neil-yechenwei:
I've updated all commit title to specified format. Please go ahead to review the code.

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/templates/AzureServerGroupResourceTemplate.groovy
```diff
@@ -57,6 +58,14 @@ class AzureServerGroupResourceTemplate {
     mapper.writeValueAsString(template)
   }
 
+  /**
+   * Initialize variables that will be used in mulitple places
```
duftler: s/mulitple/multiple

PR #3553, opened 2019-04-10
fix(titus): Don't copy migrationPolicy from source during job creation

aravindmd:
- We use to copy the `migrationPolicy` from source even when it was not specified in the payload and `jobDisruptionBudget` was set as part of the payload, this fix it to ignore the source `migrationPolicy` when `jobDisruptioBuget` is set and copy only when its not specified.

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.clouddriver.titus.deploy.handlers
 
+import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.frigga.Names
 import com.netflix.spinnaker.clouddriver.aws.deploy.ops.loadbalancer.TargetGroupLookupHelper
 import com.netflix.spinnaker.clouddriver.aws.deploy.ops.loadbalancer.TargetGroupLookupHelper.TargetGroupLookupResult
@@ -172,7 +173,9 @@ class TitusDeployHandler implements DeployHandler<TitusDeployDescription> {
         if (description.inService == null) {
           description.inService = sourceJob.inService
         }
-        description.migrationPolicy = description.migrationPolicy ?: sourceJob.migrationPolicy
+        if(description.disruptionBudget == null) {
+          description.migrationPolicy = description.migrationPolicy ?: sourceJob.migrationPolicy
+        }
         description.jobType = description.jobType ?: "service"
         if (!description.hardConstraints) description.hardConstraints = []
         if (!description.softConstraints) description.softConstraints = []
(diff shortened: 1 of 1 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
```diff
@@ -172,7 +173,9 @@ class TitusDeployHandler implements DeployHandler<TitusDeployDescription> {
         if (description.inService == null) {
           description.inService = sourceJob.inService
         }
-        description.migrationPolicy = description.migrationPolicy ?: sourceJob.migrationPolicy
+        if(description.disruptionBudget == null) {
```
ajordens: space before the `(` and perhaps an inline comment saying that the `migrationPolicy` should only be used when the `disruptionBudget` has not been specified.

ajordens: what if both are specified?

ajordens: Should the `disruptionBudget` be propagated from the `sourceJob`? (is that a thing?)

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
```diff
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.clouddriver.titus.deploy.handlers
 
+import com.fasterxml.jackson.databind.ObjectMapper
```
ajordens: where did this import come from?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
aravindmd: good question , let me circle back on this with titus

aravindmd: i think that will need some more work and most likely changes to the caching agents to, i will take a look at this next in a separate PR , this should fix the immediate issue i think.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
aravindmd: removed , i was using this for some local logging test

ajordens:
LGTM

PR #3555, opened 2019-04-11
fix(cf): don’t attempt to start app if instance count is 0

claymccoy:


jkschneider:

PR #3556, opened 2019-04-11
fix(cf): cf exception should display message from throwable

claymccoy:
I was tired of seeing error messages like the following for a failed deploy stage.
“Exception ( Monitor Deploy ) Cloud Foundry API returned with error(s):”

It is missing anything useful, like the actual error info. That is because in some cases CloudFoundryApiException is only sent a throwable which is ignored even though it has useful information about the failure.

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryApiException.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryApiException.java
@@ -58,7 +58,7 @@ public class CloudFoundryApiException extends RuntimeException {
   }
 
   private static String getMessage(Throwable t, String... errors) {
-    String[] allErrors = Arrays.copyOf(errors,errors.length + 1);
+    String[] allErrors = Arrays.copyOf(errors, errors.length + 1);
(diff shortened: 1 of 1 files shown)


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryApiException.java
```diff
@@ -54,4 +56,10 @@ private static String getMessage(String... errors) {
     return "Cloud Foundry API returned with error(s): " +
       stream(errors).filter(Objects::nonNull).collect(Collectors.joining(" and "));
   }
+
+  private static String getMessage(Throwable t, String... errors) {
+    String[] allErrors = Arrays.copyOf(errors,errors.length + 1);
```
jkschneider: nit: missing space?

claymccoy: thanks!

claymccoy: The way that IJ puts the parameter names in line for you makes this difficult to see.

claymccoy:

PR #3557, opened 2019-04-11
fix(cf): handle timeouts in CF api calls better

claymccoy:
Previous logic returned the last response if an exception was thrown. We should check that last response is not null before returning it, null does more harm than good. If it is a problem that we are not expecting, just throw the error up the chain. Let’s also expect timeouts and treat that as a retryable situation.


jkschneider:

PR #3562, opened 2019-04-12
fix(google): correctly filter instanceCacheData in GoogleClusterProvi…

maggieneterval:
…der.clusterFromCacheData

Closes https://github.com/spinnaker/spinnaker/issues/4266

- When filtering `instanceCacheData` by cluster relationship, we need to check if any of the keys contain the cluster name, rather than checking if the list of keys contains an exact match for the cluster name.
- This fixes an issue where instance health is reported incorrectly in Deck's Clusters view.


maggieneterval:

PR #3563, opened 2019-04-12
fix(kubernetes): Initialize credentials after CRDs are registered

ezimanyi:
There are currently two mechanisms for Spinnaker to cache CRDs:
* Registered CRDs that are included in clouddriver's config and can set CRD-specific properties (such as a handler class)
* Unregistered CRDs that are discovered at startup by asking the cluster for all CRDs. These use default properties that are not overriden.

We are currently polling for CRDs in the cluster and configuring all found CRDs as unregistered before we register configured CRDs. Registration of a CRD is a no-op if the CRD has already been registered, so this means the user-configured registered CRD properties are being ignored.

We should instead wait until we've registered all configured CRDs before having the caching agent try to auto-discover CRDs.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderConfig.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderConfig.java
@@ -103,6 +103,7 @@ class KubernetesV2ProviderConfig {
             log.warn("Error encountered registering {}: ", cr, e);
           }
         });
+        v2Credentials.initialize();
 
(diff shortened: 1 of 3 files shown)


ezimanyi:
This fixes something that @jfillo and I came across while looking into spinnaker/spinnaker#4237.  While I don't believe it will fully fix that issue, it should at least fix the workaround of registering the CRD you want to delete.

It looks like this was initially broken in #3272, which ensured that we polled for CRDs at least once before starting up.  This change preserves that behavior, but moves that initial poll to after we have registered configured CRDs.

ezimanyi:
Whoops, broke the tests I wrote a couple of weeks go...fixed now.

ezimanyi:

jfillo:
Thanks so much for all the work you put in for tracking down and resolving this issue!

PR #3566, opened 2019-04-14
feat(provider/google): Support Shielded VM policies

haykbaluyan:
Shielded VMs are virtual machines (VMs) on Google Cloud Platform hardened by a set of security controls that help defend against rootkits and bootkits. Using Shielded VMs helps protect enterprise workloads from threats like remote attacks, privilege escalation, and malicious insiders. Shielded VMs leverage advanced platform security capabilities such as secure and measured boot, a virtual trusted platform module (vTPM), UEFI firmware, and integrity monitoring. Please see more details and screenshots using the link below.

https://github.com/spinnaker/spinnaker/issues/4281

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
@@ -196,17 +196,19 @@ class GCEUtil {
 
     boolean isUefiCompatible = false;
     boolean isSecureBootCompatible = false;
-    for (def feature : guestOsFeatureList) {
+
+    guestOsFeatureList.each { feature ->
       if (feature.getType() == "UEFI_COMPATIBLE") {
         isUefiCompatible= true
-        continue
+        return
       }
 
       if (feature.getType() == "SECURE_BOOT") {
         isSecureBootCompatible = true
-        continue
+        return
       }
     }
+
     return isUefiCompatible && isSecureBootCompatible
   }
 
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
@@ -188,6 +188,28 @@ class GCEUtil {
     }
   }
 
+  static boolean isShieldedVmCompatible(Image image) {
+    java.util.List<GuestOsFeature> guestOsFeatureList = image.getGuestOsFeatures()
+    if (guestOsFeatureList == null || guestOsFeatureList.size() == 0) {
+      return false
+    }
+
+    boolean isUefiCompatible = false;
+    boolean isSecureBootCompatible = false;
+    for (def feature : guestOsFeatureList) {
+      if (feature.getType() == "UEFI_COMPATIBLE") {
+        isUefiCompatible= true
+        continue
+      }
+
+      if (feature.getType() == "SECURE_BOOT") {
+        isSecureBootCompatible = true
+        continue
+      }
+    }
+    return isUefiCompatible && isSecureBootCompatible
+  }
+
   static GoogleNetwork queryNetwork(String accountName, String networkName, Task task, String phase, GoogleNetworkProvider googleNetworkProvider) {
     task.updateStatus phase, "Looking up network $networkName..."
 
@@ -612,6 +634,7 @@ class GCEUtil {
 
     def networkInterface = instanceTemplateProperties.networkInterfaces[0]
     def serviceAccountEmail = instanceTemplateProperties.serviceAccounts?.getAt(0)?.email
+    def shieldedVmConfig = instanceTemplateProperties.shieldedVmConfig
 
     return new BaseGoogleInstanceDescription(
       image: image,
@@ -625,7 +648,10 @@ class GCEUtil {
       network: Utils.decorateXpnResourceIdIfNeeded(project, networkInterface.network),
       subnet: Utils.decorateXpnResourceIdIfNeeded(project, networkInterface.subnet),
       serviceAccountEmail: serviceAccountEmail,
-      authScopes: retrieveScopesFromServiceAccount(serviceAccountEmail, instanceTemplateProperties.serviceAccounts)
+      authScopes: retrieveScopesFromServiceAccount(serviceAccountEmail, instanceTemplateProperties.serviceAccounts),
+      enableSecureBoot: shieldedVmConfig?.enableSecureBoot,
+      enableVtpm: shieldedVmConfig?.enableVtpm,
+      enableIntegrityMonitoring: shieldedVmConfig?.enableIntegrityMonitoring
     )
   }
 
@@ -741,6 +767,7 @@ class GCEUtil {
                                                String phase,
                                                String clouddriverUserAgentApplicationName,
                                                List<String> baseImageProjects,
+                                               Image bootImage,
                                                SafeRetry safeRetry,
                                                GoogleExecutorTraits executor) {
     def credentials = description.credentials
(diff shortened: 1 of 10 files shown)


haykbaluyan:
@ezimanyi @jtk54 kindly please give this a look. I sure might have missed lots of things so all the feedback is very welcome. For more screenshots please refer to https://github.com/spinnaker/spinnaker/issues/4281

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -188,6 +188,28 @@ class GCEUtil {
...
+    if (guestOsFeatureList == null || guestOsFeatureList.size() == 0) {
+      return false
+    }
+
+    boolean isUefiCompatible = false;
+    boolean isSecureBootCompatible = false;
+    for (def feature : guestOsFeatureList) {
+      if (feature.getType() == "UEFI_COMPATIBLE") {
+        isUefiCompatible= true
+        continue
```
jtk54: Are there any guest features that are compatible with both UEFI and secure boot?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -188,6 +188,28 @@ class GCEUtil {
...
 
+  static boolean isShieldedVmCompatible(Image image) {
+    java.util.List<GuestOsFeature> guestOsFeatureList = image.getGuestOsFeatures()
+    if (guestOsFeatureList == null || guestOsFeatureList.size() == 0) {
+      return false
+    }
+
+    boolean isUefiCompatible = false;
+    boolean isSecureBootCompatible = false;
+    for (def feature : guestOsFeatureList) {
```
jtk54: Can do a groovy `.each` here: `guestOsFeatureList.each { feature -> <logic> }`.

jtk54: Not that big of a deal though, pretty readable as-is.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/description/BaseGoogleInstanceDescription.groovy
```diff
@@ -43,6 +43,9 @@ class BaseGoogleInstanceDescription extends AbstractGoogleCredentialsDescription
   Boolean preemptible
   Boolean automaticRestart
   OnHostMaintenance onHostMaintenance
+  Boolean enableSecureBoot;
```
jtk54: Can you drop a comment block to describe what each of these are designating?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/ModifyGoogleServerGroupInstanceTemplateAtomicOperation.groovy
```diff
@@ -168,6 +168,14 @@ class ModifyGoogleServerGroupInstanceTemplateAtomicOperation extends GoogleAtomi
         clonedDescription.disks = overriddenProperties.disks
 
         clonedDescription.baseDeviceName = description.serverGroupName
+
+        def bootImage = GCEUtil.getBootImage(description,
```
jtk54: Is there a particular reason for factoring out the boot image calculation from `buildAttachedDisks()`? I didn't notice any logic changes and we've duplicated this call in a few spots. If not I think this should remain as-is.

haykbaluyan: Thanks @jtk54  I have been thinking on this for a while and could not find a better solution. Here is the thing. The only way to check if shielded VM policies should be passed is to check guestOSFeatures on Image object (**passing shielded vm policies to non-shielded image will result in error**).

To access guestOsFeatures i need Image object and i noticed that getBootImage call in buildAttachedDisks gets me this object. Of course I can leave buildAttachedDisks as it is and separately call getBootImage but that will mean calling it twice (once as part of shielded VM decision and one as part of buildAttachedDisks call). I wanted to avoid calling it twice so did this refactoring. Maybe not the best decision though.

On the other hand calling getBootImage twice will force to make changes in the below test since it expects "images" call to be made only once. Sorry for the long comment and please let me know what is the best course of action in your opinion. 

com.netflix.spinnaker.clouddriver.google.deploy.ops.CreateGoogleInstanceAtomicOperationUnitSpec > should create instance FAILED
    junit.framework.AssertionFailedError: No call to 'images' expected at this point. Still 1 call(s) to 'instances' expected.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
haykbaluyan: According the Google's documentation both features must be available to recognize the image as a shielded image.

**Once the disk is ready, the next step is to create an official GCP image using the gcloud
command. In particular, you must specify UEFI_COMPATIBLE and SECURE_BOOT in the
--guest-os-features flag. These values allow GCE to recognize the image as a shielded
image. For**

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
haykbaluyan: Will change it. I am all for using groovy's built-in controls :)

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/description/BaseGoogleInstanceDescription.groovy
haykbaluyan: Will do.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/ModifyGoogleServerGroupInstanceTemplateAtomicOperation.groovy
jtk54: I see the issue -- we actually only include a [subset](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy#L797) of the image's attributes and a reference in the disks built up by `buildAttachedDisks()`. Furthermore, we're using the underlying raw compute model classes so we can't simply change the representation to support what we need for shielded VMs. I'm OK with the change as you proposed, but can you drop a comment in the callsites? Thanks.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
jtk54: :+1:

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/ops/ModifyGoogleServerGroupInstanceTemplateAtomicOperation.groovy
haykbaluyan: Exactly, it just seem to make sense to inject image as parameter instead of a call within buildAttachedDisks since image can be later needed for other purposes too (just like this case with shielded VMs).

Also sorry for a naive question, but what are your referring to by "callsites"? You mean places where buildAttachedDisks is called?

jtk54: Yes.

PR #3567, opened 2019-04-15
fix(aws/caching): only add application attribute to authoritative types

asher:


asher:
The AWS `ClusterCachingAgent` was recently changed to add an application attribute to all cached types (#3481). The upcoming SQL cache backend explicitly only stores data for `AUTHORITATIVE` cache types, using `INFORMATIVE` types to build a DAG. The default redis backed cache uses the absence of attribute keys to make a similar distinction. Adding an attribute key to `INFORMATIVE` types had the unintended side effect of them being treated as `AUTHORITATIVE`, essentially dueling with the truly `AUTHORITATIVE` agents.

ajordens:
Good dig and quick turnaround ... welcome back @asher.

LGTM

ajordens:
`Adding an attribute key to INFORMATIVE types had the unintended side effect of them being treated as AUTHORITATIVE, essentially dueling with the truly AUTHORITATIVE agents.`

Given this is unintended ... is there a better way this could have been logged or otherwise detected?

asher:

asher:
@ajordens Not thinking of a good way to detect this issue at runtime. Redis constraints rule out our ability to cheaply detect if caching agents are overwriting each other, while `DefaultProviderCache` only uses the `AUTHORITATIVE` vs. `INFORMATIVE` distinction to limit cache evictions to `AUTHORITATIVE` types in `putCacheResult`. We could potentially search for `INFORMATIVE` types with data that includes attributes there and log, but it's currently a valid condition. Prior to 2019Q1, no caching agent declared itself as `AUTHORITATIVE` for AWS Clusters, and I wouldn't be surprised if agents for other cloud providers depend on this behavior, and would thus break/not store some data types when run with `SqlProviderCache`. Logging a warning here (where it could legitimately fire today for every run of some non-aws caching agents) strikes me as more likely to produce noise than give us a preemptive signal.

Detecting pre-merge, I can imagine us building tests around mocked cloud provider responses, run a set of caching agents in a defined order, make assertions on controller outputs, re-run the agents in reverse order, then repeating the same assertions. There's a decent chance we'll be refactoring many of the providers and perhaps controllers over the next year to take advantage of and mature the sql backend, so I think there'd be a lot of value in this sort of testing, especially if it can be implemented and maintained for all major cloud providers.

PR #3575, opened 2019-04-17
fix(aws/deploy): support image lookup when deploying by imageId rather than name

cfieber:
supports finding an already resolved image when the deploy was supplying AMI id rather than AMI name


cfieber:
thanks me too!!! :heart_eyes:

cfieber:
all the tests passed so probably for sure

PR #3576, opened 2019-04-17
fix(cloudformation): force update only when account and region match

xavileon:
Updating the cloud formation cache on demand is not a cheap operation in
terms of API calls to aws (linear on the number of stacks).
Updating the cache for all the accounts/region pairs can take a
non-negligible amount of time and make the force cache task in orca timeout,
not considering the unnecessary updates.

This patch filters those updates that do not belong to the
account/region of the caching agent in question.

Related PR: https://github.com/spinnaker/orca/pull/2843


xavileon:
@spinnaker/reviewers PTAL

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonCloudFormationCachingAgent.java
```diff
@@ -86,7 +86,20 @@ public boolean handles(OnDemandType type, String cloudProvider) {
 
   @Override
   public OnDemandResult handle(ProviderCache providerCache, Map<String, ?> data) {
-    return new OnDemandResult(getOnDemandAgentType(), loadData(providerCache), Collections.emptyMap());
+    if (shouldHandle(data)) {
+      log.info("Updating CloudFormation cache for account: {} and region: {}", account.getName(), this.region);
+      return new OnDemandResult(getOnDemandAgentType(), loadData(providerCache), Collections.emptyMap());
```
ajordens: Am I wrong in assuming that you actually just want to refresh a particular stack vs. everything in this account/region?

xavileon: Yes, that would be ideal. Looking at how the on demand caching works though, apparently the whole set of cached elements are replaced by whatever this method returns, evicting everything not contained on the CacheResult [[1]][[2]] (not what we want I guess?). I may be misunderstanding how we handle the result of this.

[1]: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy#L100
[2]: https://github.com/spinnaker/clouddriver/blob/master/cats/cats-core/src/main/java/com/netflix/spinnaker/cats/provider/ProviderCache.java#L27

xavileon: @ajordens what do you think?

robzienert: @xavileon Your assertion (as far as I’m aware) is correct. I’d love to see a “patch cache” operation additionally available for caching agents, in addition to our current replacement strategy.

PR #3581, opened 2019-04-18
fix(aws): DetachInstances op: only run min capacity checks when decrementDesiredCapacity is true

christopherthielen:


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/DetachInstancesAtomicOperation.groovy
```diff
@@ -78,9 +78,9 @@ class DetachInstancesAtomicOperation implements AtomicOperation<Void> {
         return false
       }
 
-      if ((autoScalingGroup.desiredCapacity - validInstanceIds.size()) < autoScalingGroup.minSize) {
+      int newMin = autoScalingGroup.desiredCapacity - validInstanceIds.size()
+      if (description.decrementDesiredCapacity && newMin < autoScalingGroup.minSize) {
```
ajordens: Might need some default valuing for `decrementDesiredCapacity`?

Default to `true` if `null`?

christopherthielen: doesn't it default to `false` by virtue of being a `boolean`?

ajordens: Totally correct, I glossed over that.

👍

PR #3597, opened 2019-04-24
 refactor(provider/kubernetes): Remove reflection from cache constructors

ezimanyi:
* refactor(provider/kubernetes): Remove reflection from cache constructors 

  Constructing caching agents currently uses reflection, which breaks static analysis and makes it harder to reason about and refactor this code. Instead of having each handler return a reference to the class that should be constructed, instead have the handler return a functional interface to the actual constructor so that the base class does not need to use reflection.

* refactor(kubernetes): Runtime -> compile time error 

  All subclasses of KubernetesV2Caching agent need to implement either primaryKinds or primaryKind, and a failure to do so is a runtime error.  Make this a compile-time error by standardizing on primaryKinds and making it abstract in the base class.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2CachingAgentFactory.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2CachingAgentFactory.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2019 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.agent;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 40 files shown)


ezimanyi:
@plumpy : Realized I can't formally request review from you yet, but this is the PR I just mentioned.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesHandler.java
```diff
@@ -96,43 +88,15 @@ public KubernetesV2CachingAgent buildCachingAgent(
       int agentCount,
       Long agentInterval
   ) {
-    Constructor constructor;
-    Class<? extends KubernetesV2CachingAgent> clazz = cachingAgentClass();
```
jtk54: Always feels good to get rid of variables named `clazz`.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2CachingAgentFactory.java
```diff
@@ -0,0 +1,35 @@
...
+
+package com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.agent;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spectator.api.Registry;
+import com.netflix.spinnaker.clouddriver.kubernetes.security.KubernetesNamedAccountCredentials;
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.KubernetesResourcePropertyRegistry;
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.security.KubernetesV2Credentials;
+
+public interface KubernetesV2CachingAgentFactory {
```
plumpy: Maybe mark this @FunctionalInterface?

ezimanyi: Good idea, will do that

PR #3600, opened 2019-04-24
fix(provider/kubernetes): Fix cache refresh timeout for HA clouddriver

ezimanyi:
When clouddriver is deployed in HA mode, on-demand cache refresh request are sent to the RW clouddriver, but requests for the list of currently-pending requests are sent to the RO clouddriver.

In principle, this should not be a problem, as the pending refresh requests are stored in Redis (which both query). The problem is that each caching agent keeps a list of namespaces it should keep track of, and only returns pending results for those namespaces.

If you don't have an explicit list of namespaces to cache, clouddriver periodically calls reloadNamespaces() on each agent to update its list of namespaces. This call happens when either processing an on-demand request or as part of a regular caching cycle, so that the list of namespaces in each agent stays up-to-date and all namespaces are cached by one agent.

The problem is that read-only clouddriver does not have scheduled caching cycles and does not process on-demand requests (by design) so it never refreshes its list of namespaces. New namespaces that are added after clouddriver-ro restarts will never be assigned to a caching agent, so requests for pending cache refreshes will always omit those namespaces.

There's actually no advantage to sharding out requests for pending requests to multiple caching agents. Each is making exactly the same request to redis, then filtering the results to its namespaces, only to have all those results combined in the end. Fix the bug (and reduce some redundant work) by just designating a single caching agent as the one to receive all read requests and have it return all of them, regardless of namespace.


ezimanyi:

PR #3601, opened 2019-04-25
fix(cf): Create / Delete Service Key timeouts

jmelchio:
- Moved from /v2/spaces/{guid}/space_summary to /v2/.../service_instances
because it is faster and it gives us everything we need

spinnaker/spinnaker#4338


jkschneider:
Very interesting... Did we tell CAPI about this by chance?

PR #3603, opened 2019-04-25
feat(provider/kubernetes): replace annotation support

alperkokmen:
this addresses https://github.com/spinnaker/spinnaker/issues/2348.

essentially, it adds support for deploying k8s manifests using `replace` command. changes are pretty similar to the ones for recreate support that was recently added [here](https://github.com/spinnaker/clouddriver/pull/3332). we ran into the issue below while deploying fairly large `configMap` instances using `kubectl apply -f`. workaround was to use `kubectl replace -f`.

> metadata.annotations: Too long: must have at most 262144 characters

here are some things i considered but didn't implement until i get more feedback from the team –
- [`KubectlJobExecutor#deploy`](https://github.com/spinnaker/clouddriver/blob/9282c27ca9e8c0e015aa823e76ca4e711e68a89c/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubectlJobExecutor.java#L349-L369) is actually running `apply` which is one way of deploying/updating resources in K8s. other update methods such as `patch` have [matching method names in `KubectlJobExecutor`](https://github.com/spinnaker/clouddriver/blob/9282c27ca9e8c0e015aa823e76ca4e711e68a89c/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubectlJobExecutor.java#L546-L579). we could rename `deploy` as `apply` to wrap `kubectl` more consistently; but, not really a required functional change.
- instead of handling the split between `apply` and `replace` at calling sites, it could be handled in `KubectlJobExecutor#deploy` by passing `replace` as an argument and changing [the first argument of `command`](https://github.com/spinnaker/clouddriver/blob/9282c27ca9e8c0e015aa823e76ca4e711e68a89c/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubectlJobExecutor.java#L355). `CanDeploy#deploy` would just pass `replace` to `KubernetesV2Credentials#deploy` which would pass them to `jobExecutor`.

this would also require an update to docs [here](https://github.com/spinnaker/spinnaker.github.io/blob/master/reference/providers/kubernetes-v2/index.md) to mention the new annotation.


alperkokmen:
@lwander could you please take a look when you get a chance? thanks!

emjburns:
@ezimanyi can you take a look? (@alperkokmen Lars no longer works on Spinnaker)

PR #3604, opened 2019-04-25
feat(ecs): Add service discovery support

clareliguori:
Cache available service registries from AWS Cloud Map, and specify service discovery configuration for ECS server groups


clareliguori:
Ping @spinnaker/reviewers

PR #3612, opened 2019-04-30
feat(cf): add build info to server group

claymccoy:
Every cloud provider will have a different way to put metadata on their server group. Here we use app environment variables because we don’t have resource tagging yet.
The builds are not cloud provider specific so it makes sense to have more consistency here. The CFBuildInfo could be more general purpose than just CF. It could be moved to the core model as part of a server group and a consistent way to represent ci build reference info. It is basic build metadata that is produced by Artifactory build info and relevant attributes to just about any CI system.

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/model/BuildEnvVar.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/model/BuildEnvVar.java
@@ -1,11 +1,11 @@
 /*
- * Copyright 2019 Netflix, Inc.
+ * Copyright 2019 Pivotal, Inc.
  *
- * Licensed under the Apache License, Version 2.0 (the "License")
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/model/CloudFoundryBuildInfo.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2019 Netflix, Inc.
```
Jammy-Louie: Can you fix the headers?

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/model/BuildEnvVar.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright 2019 Netflix, Inc.
```
Jammy-Louie: Can you fix the headers?

Jammy-Louie:
👍

PR #3613, opened 2019-04-30
fix(provider/google): ServiceAccount should be a Set<ServiceAccount>

seanpeters86:
Right now, each firewall rule is properly being matched by tag or service account. However, if the accounts are using shared VPC the Spinnaker caching agent will show each firewall rule existing in every project, even though it is hosted in the shared vpc only. The idea for this change is to only show firewall rules in the project they originate from, instead of all of them.

**This is still a WIP and is currently not functioning properly and should not be merged.**

The expression below is correctly evaluating the firewall Project, and is correctly jumping into the if statement if there is not a match, but the rules are still being returned in the Spinnaker UI firewalls section and the `/securityGroups` endpoint


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/view/GoogleSecurityGroupProvider.groovy
```diff
@@ -216,6 +216,15 @@ class GoogleSecurityGroupProvider implements SecurityGroupProvider<GoogleSecurit
         targetServiceAccountsInCommon = (securityGroup.targetServiceAccounts).intersect(serviceAccounts)
       }
 
+      /** Due to how shared VPC projects share firewall rules, they look "duplicated" in Spinnaker and show
+      * once per account. This derives the project the firewall is hosted in, and confirms that it matches where the
+      * Security group is located.
+      */
+      def firewallProjectAndAccountMatch = GCEUtil.deriveProjectId(securityGroup.selfLink) == account
```
jtk54: I don't think `projectId` and `account` are equivalent here.

My understanding here is that the point of VPC is to manage network resources in one project called the 'host project' and then utilize those network resources in a separate project called the 'service project', so the two are segregated as far as permissions are concerned. I think the duplication comes from the [caching agent](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleSecurityGroupCachingAgent.groovy) writing the duplicate shared VPC security groups to redis for each configured GCE account, so this is likely the problem.

seanpeters86:
@jtk54 Just tried out a few things in thew new commit:
- Swapped to using Set<ServiceAccount> instead
- Working on fixing the security groups caching agent as you suggested (took a similar approach as I did before, and it also failed)

Testing this out now.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleSecurityGroupCachingAgent.groovy
```diff
@@ -210,6 +210,19 @@ class GoogleSecurityGroupCachingAgent extends AbstractGoogleCachingAgent impleme
...
+            if (GCEUtil.deriveProjectId(it.selfLink) != project) {
+              log.debug("${GCEUtil.deriveProjectId(it.selfLink)} does not equal ${project}")
+              newFirewalls.add(it)
+            }
+          } catch(Exception ex) {
+            log.debug(ex.getMessage())
+          }
+        }
+        log.debug("Old Size: ${firewalls.size()}")
+        log.debug("New Size: ${(firewalls.size() - newFirewalls.size())}")
```
seanpeters86: This is definitely not the write way to cache things properly, as right now the "New Size" is always 0, meaning none of them pass through the logic I wrote above. Which makes sense considering this is a similar approach I took in the function call itself. Any ideas where this should go instead? I assumed it would be around here since just about we get `firewalls = (firewalls ?: []) + (hostFirewalls ?: [])` which is the crux if my problems I believe.

PR #3614, opened 2019-05-01
fix(runJob/kubernetes): defer cleanup to k8s

ethanfrogers:
This is the beginning of some work to redo how logs are handled for Jobs
in the V2 provider. In order to implement the new approach we need to
prevent Jobs from being cleaned up by Spinnaker and defer to Kubernetes
to do that work. Versions >= 1.12 implement a `ttlSecondsAfterFinished`
field for Jobs that will tell Kubernetes to cleanup the Job after that
TTL has expired. We'll communicate this to customers via documentation.


ethanfrogers:
It should be noted that `ttlSecondsAfterFinished` is an alpha feature and requires enabling a feature flag on the API server for it to work. If we find that a lot of customers don't have this enabled we can implement a `strategy.spinnaker.io/ttl` annotation to overcome that.

PR #3616, opened 2019-05-02
fix(docker): Added missing insecureRegistry property

jasonjchu:
spinnaker/spinnaker#3722


ezimanyi:
I just cleaned up the Kubernetes account code to remove a couple of levels of builders (#3630); looks like a similar change would be helpful for the Docker account code to prevent issues like this.

PR #3628, opened 2019-05-06
fix(ecs): Retry on ECS service eventual consistency

clareliguori:
Fixes spinnaker/spinnaker#4371


jwach:
@clareliguori @jkschneider Any chance for releasing this fix within 1.13.x ? This is quite a blocker for us.

PR #3630, opened 2019-05-06
refactor(kubernetes): Simplify account creation code

ezimanyi:
The code to translate Kubernetes account properties into accounts uses multiple levels of `Builder` classes, each of which do a bit of defaulting/initialization.  This makes it difficult to understand where configured account properties are defaulted, and where they are actually used.

As a first pass to simplifying this, clean up the interface between `KubernetesNamedAccountInitializer` and `KubernetesNamedAccountCredentials`.  Remove the builder class and have defaulting either occur directly on the `KubernetesConfigurationProperties.ManagedAccount` class (for simple defaults) or in the constructor (for more complex cases that aren't just setting a single default value.)

* test(kubernetes): Add tests to credentials initializing 

  There are currently no tests on initializing credentials; before refactoring this code, add some tests.

  Also, a number of the Kubernetes tests depend on being able to explicitly pass credentials to a builder for KubernetesNamedAccountCredentials, which will be removed. Instead just use a mock object and specify the desired properties.

* refactor(kubernetes): Remove unused fields from account credentials 

  Most of the fields on KubernetesNamedAccountCredentials are just copies of the fields on the contained KubernetesCredentials object. There's no need to also have these on the parent class (and in fact we never look these up on the parent class).  Remove all of the redundant fields.

  For the remaining fields:
  - Apply 'private final' consistently
  - Replace the accessors with a @Getter annotation

* refactor(kubernetes): Consolidate defaulting of account parameters 

  Defaulting of account paramters is scattered across a few different classes; consolidate this to the definition of the properties class so it's easier to see the values of these properties.

* refactor(kubernetes): Pass ManagedAccount to builder 

  Rather than pass each property of the ManagedAccount individually to the Builder for KubernetesNamedAccountCredentials, just pass the entire object.

* refactor(kubernetes): Use a factory class instead of a builder 

  Now that we're passing the entire ManagedAccount class to the builder rather than the properties individually, there are no optional parameters, so a builder class is less optimal. Other than the ManagedAccount, all properties of the builder are beans that can be autowired. Create a factory bean that autowires in these other beans, and exposes methods to create a v1 or a v2 account from a supplied ManagedAccount.

  Ideally most of the logic in CredentialFactory would actually live in as factories in the credential classes themselves; I'll likely make this change in a future PR, but for now this is a reasonable shim to avoid making this PR too large.

### DIFF
--- a/clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesNamedAccountCredentialsInitializerSpec.groovy
+++ b/clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesNamedAccountCredentialsInitializerSpec.groovy
@@ -117,7 +117,6 @@ class KubernetesNamedAccountCredentialsInitializerSpec extends Specification {
     1 * accountCredentialsRepository.save("test-account", _ as KubernetesNamedAccountCredentials) >> { _, creds ->
       credentials = creds
     }
-    1 * namerRegistry.getNamingStrategy("kubernetesAnnotations") >> Mock(KubernetesManifestNamer)
 
     credentials.getName() == "test-account"
     credentials.getProviderVersion() == ProviderVersion.v1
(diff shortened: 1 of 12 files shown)


clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesNamedAccountCredentialsInitializerSpec.groovy
```diff
@@ -117,7 +117,6 @@ class KubernetesNamedAccountCredentialsInitializerSpec extends Specification {
     1 * accountCredentialsRepository.save("test-account", _ as KubernetesNamedAccountCredentials) >> { _, creds ->
       credentials = creds
     }
-    1 * namerRegistry.getNamingStrategy("kubernetesAnnotations") >> Mock(KubernetesManifestNamer)
```
ezimanyi: This used to run for both v1 and v2, even though it was only needed for v2.  After the refactor it only runs for v2.

ezimanyi:
Realized I overzealously removed two properties that were used on the front-end, so restored those in the last commit.

PR #3631, opened 2019-05-06
fix(cf): previously unhandled exception from bad application state

Jammy-Louie:
spinnaker/spinnaker#4357


jkschneider:

PR #3638, opened 2019-05-07
feat(cf): Use Spring Cloud Config for external configuration of providers and credentials

scottfrederick:
See https://github.com/spinnaker/spinnaker/issues/3969

Design document: https://docs.google.com/document/d/1qrSxHMwZHrpiyAM25986a5V_rzfSOlVLhlC-BpOruyE/

Adds support for loading configuration of accounts and other properties from an external source such as a git repository, a Vault server, or a JDBC-accessible database. 

* Embeds a Spring Cloud Config Server in clouddriver
  * Disabled by default by activating a `native` Spring Cloud Config Server profile with no further configuration
  * Enabled by providing external Spring Cloud Config Server configuration (see https://docs.google.com/document/d/16EWk-ChZkae87hCSCFdl3ElepTLC-ac-4uv1tCyfZcc/edit#)
* Provides a Spring Cloud-native way for configuration to be refreshed 
  * Listens for Spring `RefreshScopeRefreshedEvent`s and calls a new `synchronize()` method on all `CredentialsInitializerSynchronizable` beans when the app context is refreshed
  * Refreshes the app context on a schedule, on approximately the same interval as cache agents are updated
* Implements config refresh synchronization in the Cloud Foundry provider


clouddriver-cloudfoundry/clouddriver-cloudfoundry.gradle
```diff
@@ -28,5 +28,7 @@ dependencies {
   testCompile "org.junit.jupiter:junit-jupiter-engine:${spinnaker.version("jupiter")}"
   testCompile "org.junit.jupiter:junit-jupiter-params:${spinnaker.version("jupiter")}"
 
+  compile 'org.springframework.cloud:spring-cloud-context:1.3.5.RELEASE'
```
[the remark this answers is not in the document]
scottfrederick: The versions of `spring-cloud-context` and `spring-cloud-config-server` will have to change, as Cloud versions are closely tied to Boot versions. If we want to hold off on merging this PR until after `boot2` is merged to master, I can update the versions and re-test everything. Otherwise we'll have to update these versions on the `boot2` branch. 

Also see https://github.com/spinnaker/spinnaker-dependencies/commit/4ad3fa68eae1797219bb752043c3ace99f0cd2ac.

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/security/CloudFoundryCredentialsSynchronizer.java
```diff
@@ -0,0 +1,113 @@
...
+
+  @Override
+  @PostConstruct
+  @SuppressWarnings("unchecked")
+  public void synchronize() {
+    List<?> deltas = ProviderUtils.calculateAccountDeltas(accountCredentialsRepository,
+      CloudFoundryCredentials.class,
+      cloudFoundryConfigurationProperties.getAccounts());
+
+    List<CloudFoundryConfigurationProperties.ManagedAccount> addedAccounts =
```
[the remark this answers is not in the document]
scottfrederick: I couldn't agree more, that untyped list of size 2 is not very clear. I'm happy to make that change if it's not too much for this PR.

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/security/CloudFoundryCredentialsSynchronizer.java
```diff
@@ -0,0 +1,113 @@
...
+    this.cloudFoundryConfigurationProperties = cloudFoundryConfigurationProperties;
+    this.accountCredentialsRepository = accountCredentialsRepository;
+    this.catsModule = catsModule;
+    this.registry = registry;
+  }
+
+  @Override
+  @PostConstruct
+  @SuppressWarnings("unchecked")
+  public void synchronize() {
```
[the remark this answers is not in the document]
scottfrederick: Good question, I'll re-test and make sure that case is covered. I know that the `@ConfigurationProperties` beans (e.g. `CloudFoundryConfigurationProperties` in this provider) will get updated with any change from the backing config repository, but I'm not positive that this code will update the `accountCredentialsRepository` with the changes.

scottfrederick:
@ezimanyi 

> Is there something here that might have changed the priority order?

These changes shouldn't change that priority. Per the [Boot docs](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html), the relative priority should be: 

> 12. Profile-specific application properties outside of your packaged jar (application-{profile}.properties and YAML variants).
> 13. Profile-specific application properties packaged inside your jar (application-{profile}.properties and YAML variants).
> 14. Application properties outside of your packaged jar (application.properties and YAML variants).
> 15. Application properties packaged inside your jar (application.properties and YAML variants).

which would translate to: 

1. `/home/.spinnaker/clouddriver-local.yml`
2. `/home/.spinnaker/clouddriver.yml`
3. `clouddriver.yml`

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/security/CloudFoundryCredentialsSynchronizer.java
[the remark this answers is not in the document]
scottfrederick: Existing account modification was addressed in a new commit.

clouddriver-cloudfoundry/clouddriver-cloudfoundry.gradle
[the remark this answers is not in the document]
ezimanyi: Sounds good; the `boot2` branch was merged this afternoon

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/security/CloudFoundryCredentialsSynchronizer.java
```diff
@@ -90,12 +80,10 @@ private void synchronizeRepository(List<CloudFoundryConfigurationProperties.Mana
     });
   }
 
-  private void synchronizeProvider(List<Agent> addedAgents) {
-    cloudFoundryProvider.getAgents().addAll(addedAgents);
-  }
-
   private void synchronizeAgentCache(List<Agent> addedAgents, List<String> deletedAccountNames) {
+    cloudFoundryProvider.getAgents().addAll(addedAgents);
```
ezimanyi: I'm not too familiar with the Cloud Foundry provider, or with the scheduling of caching agents in general, so this could use a look from someone else.  But if I'm reading this properly (which I may not be), it looks like if an account changes we'll store the new account details in the repository but we won't create a new caching agent and will leave the old caching agent running with a reference to the old `CloudFoundryCredentials`.  We probably don't want to reschedule all agents on every polling cycle, so it might be necessary to do some diffing of the existing and new account if we want to support live modifications.

ezimanyi:
@scottfrederick : Regarding the priority order, my reading of the docs is the same as yours, and I don't see anything here that *should* change the priority order.  But somehow I'm still observing that things I have set in my `/home/.spinnaker/clouddriver.yml` are not taking precedence over the `clouddriver.yml` after this change.

As one example, overriding `server.port` in `/home/.spinnaker/clouddriver.yml` worked before this change, but after this change it's always using the default in `clouddriver.yml`.  I tried turning on `DEBUG` level logging to see if there were any obvious clues but it wasn't obvious from that output what the change was.

cfieber:
@ezimanyi try starting with -Ddebug as well, that does log out configuration sources

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CloudConfigRefreshConfig.java
```diff
@@ -0,0 +1,55 @@
...
+ * from an environment repository backend on a schedule that matches the cache refresh
+ * schedule.
+ *
+ * The specified {@code @Profile}s are used by Spring Cloud Config Server to indicate which
+ * environment repository backend(s) are used. Only configure a {@link CloudConfigRefreshScheduler}
+ * if at least one Spring Cloud Config Server profile is active.
+ */
+@Configuration
+@AutoConfigureAfter({RedisCacheConfig.class, DynomiteCacheConfig.class})
+@Profile({"git", "vault", "jdbc", "credhub", "composite"})
```
cfieber: Is there an alternative here to profile naming? (@ConditionalOnBean or something?)

profile naming seems pretty personal to the application configurer and this would be pretty much the first time we've driven behaviour off of it

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy
```diff
@@ -68,8 +74,12 @@ class Main extends SpringBootServletInitializer {
   }
 
   static void main(String... args) {
+    BOOTSTRAP_SYSTEM_PROPS.each { key, value -> System.setProperty(key, value) }
```
cfieber: if these are defaults, they should only setProperty if there isn't already a value for that property (otherwise I suspect its impossible to change for example bootstrap.location)

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy
```diff
@@ -48,16 +48,22 @@ import java.security.Security
...
-    'spring.application.name': 'clouddriver',
-    'spring.config.name'     : 'spinnaker,${spring.application.name}',
-    'spring.profiles.active' : '${netflix.environment},local'
+    'spring.config.name'     : 'spinnaker,${spring.application.name}'
   ]
 
+  static final Map<String, String> BOOTSTRAP_SYSTEM_PROPS = [
+    'spring.application.name'               : 'clouddriver',
+    'spring.cloud.bootstrap.location'       : '${user.home}/.spinnaker/',
+    'spring.cloud.bootstrap.name'           : 'spinnaker-config,${spring.application.name}-config',
```
cfieber: I'd suggest avoid a dash in these names since there are semantics around `-profile` and this could be interpreted as configuration for spinnaker or application in the `config` profile (something like  `spinnakerconfig,${spring.application.name}config` avoids that collision case)

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy
```diff
@@ -48,16 +48,22 @@ import java.security.Security
 @EnableScheduling
 class Main extends SpringBootServletInitializer {
 
-  static final Map<String, String> DEFAULT_PROPS = [
+  private static final Map<String, String> DEFAULT_PROPS = [
```
cfieber: this wasn't private because it is referenced in ConfigRefreshController (probably not breaking because thanks groovy)

you should be able to leave this private, delete launchArgs in this class, and delete ConfigRefreshController since spring-cloud-context should already be providing that functionality

cfieber:
more general of a question for team pivotal / spring experts:

is there a better way than the existing mechanism (of manually constructing an Agent per account) of trying to create a bean per list of items in configuration?

I've never been super happy with the bootstrapping code where we are spinning up all the caching agents, and even less happy with the code that went in to make that dynamic (we literally never use the /config-refresh behaviour instead we just redeploy a new clouddriver).

Would it be worth re-considering the caching agents to be singletons that used the ConfigurationProperties bean (which becomes dynamic with config server) to figure out what accounts to cache or operate on?

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy
scottfrederick: I can make that change

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy
scottfrederick: Agreed on the confusion with `-profile`. I've struggled to find something I like, but `spinnakerconfig,${spring.application.name}config` works for me.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy
scottfrederick: I was hoping the `ConfigRefreshController` and supporting code could be deleted after this new refresh mechanism was in place, but wasn't sure if I should do that in this PR. I'm happy to delete some more code.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CloudConfigRefreshConfig.java
scottfrederick: These profiles are what [Spring Cloud uses to decide](https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java#L280-L379) which environment repository (or repositories) to configure. It's a bit unfortunate that the profiles don't have names that are more specific to Spring Cloud, but we can't change that. I can see if there's a bean we could depend on instead, but you get most of the Config Server infrastructure with these changes - just with a default `NativeEnvironmentRepository` via the `native` profile without a location configured.

jkschneider:
> my local configuration ... is taking lower precedence than the default config in ...

@ezimanyi I think this was unrelated to this PR and fixed by https://github.com/spinnaker/clouddriver/pull/3656.

ezimanyi:
@jkschneider : I'd done my earlier testing before this was rebased onto `boot2` so I don't think it would be related to that PR.  I just tried to re-test and now ran into :
```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'defaultEnvironmentRepository' defined in class path resource [org/springframework/cloud/config/serv
er/config/DefaultRepositoryConfiguration.class]: Invocation of init method failed; nested exception is java.lang.IllegalStateException: You need to configure a uri for the git repository.
```

I just want to make sure some smoke testing is done to make sure existing configs still work before we merge this; I'm happy to try my config again once the above startup failure is fixed.

ethanfrogers:
@cfieber @scottfrederick @jkschneider @ezimanyi has Eric's last comment been resolved? I'd be happy to do what I can to help push this forward. Let me know if there's anything I can do!

scottfrederick:
@ethanfrogers I'm working through that now, figuring how to "no-op" config server by default while allowing backing repos to be configured. Things changed with Boot 2 / Cloud 2 that requires some re-work there.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CloudConfigRefreshConfig.java
scottfrederick: Changed conditional to detect bean names instead of profiles.

scottfrederick:
> my local configuration ... is taking lower precedence than the default config in ...

@ezimanyi I've tested this property precedence after working out the `boot2` updates, and it looks right to me. Properties in `/home/.spinnaker/spinnaker[-local].yml` or `/home/.spinnaker/clouddriver[-local].yml` override properties in the bundled `clouddriver.yml`. I enabled Boot actuators in looked at the `/env` endpoint to verify that properties sources were loaded in the expected order. 

Can you update with the latest commits in this PR and see if things look right now with newer Boot and Cloud?

ezimanyi:
@scottfrederick : I'm now getting a circular bean dependency error with either the Google or Kubernetes provider enabled.

```
The dependencies of some of the beans in the application context form a cycle:

   kubernetesV1ApplicationProvider defined in URL [jar:file:/home/ezimanyi/dev/spinnaker/clouddriver/clouddriver-kubernetes/build/libs/clouddriver-kubernetes.jar!/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/view/KubernetesV1ApplicationProvider.class]
      ↓
   cacheView defined in class path resource [com/netflix/spinnaker/clouddriver/cache/CacheConfig.class]
┌─────┐
|  catsModule defined in class path resource [com/netflix/spinnaker/clouddriver/cache/CacheConfig.class]
↑     ↓
|  kubernetesNamedAccountCredentials defined in class path resource [com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesNamedAccountCredentialsInitializer.class]
↑     ↓
|  synchronizeKubernetesAccounts defined in class path resource [com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesNamedAccountCredentialsInitializer.class]
└─────┘
```
I think it was introduced by changing the scope of the `synchronizeKubernetesAccounts` from a prototype to a singleton.  As this removes the `CredentialsRefreshListener` class that was fetching that prototype bean, I suspect that we can remove the `@Bean` annotation from all of the `synchronize*` methods and just leave them as private methods called by the `*namedAccountCredentials` beans.

I made that change locally just to get past the startup error, and it does look like the properties are now being defaulted in the expected order for me.

scottfrederick:
@ezimanyi I pushed (yet) another commit with the change you suggested. I started with all the providers enabled that I could enable without other configuration. Hopefully it will start up for you now!

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/security/CloudFoundryCredentialsSynchronizer.java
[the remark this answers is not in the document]
scottfrederick: On further review, this `calculateAccountDeltas` is used in most providers, usually in Groovy code where the "multiple return values" syntax is much cleaner on the calling side. I'd prefer not to pile more sweeping refactorings into this PR, but delay that for another day.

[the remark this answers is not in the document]
ezimanyi: Sounds good to me

ethanfrogers:
This is more of an FYI than anything.  Halyard sets `SPRING_PROFILES_ACTIVE=local` on the Clouddriver deployment but for this to work we'll need to update it to be `SPRING_PROFILES_ACTIVE=local,composite`. We should probably do that _before_ this is merged to make sure we don't break anything.

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/security/CloudFoundryCredentialsSynchronizer.java
scottfrederick: Unschedling and rescheduling accounts was added in another commit.

scottfrederick:
@ezimanyi @cfieber @ethanfrogers I think all comments/feedback has been addressed, and this is ready for review again. 

I can squash all these commits down if/when you think it's ready to merge.

PR #3642, opened 2019-05-08
fix(cf): missing null check for env vars

Jammy-Louie:


claymccoy:
@Jammy-Louie This was a great catch. It never got merged though, I probably should have done it when I approved it. I am redoing this code a bit right now and looked back at this to make sure I was handling this case. I'm just going to add it into that new PR rather than resolve the conflict here. Thanks!

PR #3645, opened 2019-05-08
fix(titus): If a run job spits out a yaml file, we were parsing it as properties.  Now we parse it as yaml.

christopherthielen:


clouddriver-titus/src/test/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusJobProviderSpec.groovy
```diff
@@ -0,0 +1,153 @@
+/*
+ * Copyright 2015 Netflix, Inc.
```
marchello2000: living in the past?

marchello2000:
LGTM

PR #3649, opened 2019-05-09
 feat(kubernetes): Add metrics support to live manifest mode

ezimanyi:
Fixes spinnaker/spinnaker#4243

* test(kubernetes): Add tests to pod metrics caching 

* refactor(kubernetes): Improve typing of pod metrics 

  The pod metrics code converts the metrics to Maps before using them, which makes it more difficult to use these metrics. Simplify this code by using the KubernetesPodMetric class that already exists.

  In order to preserve the serialized version stored in the cache, rename some fields to match what we're currently using to serialize. Then serialize/deserialize to/from that object when we write/read from the cache.

* feat(kubernetes): Add metrics support to live manifest mode 

  The live manifest supplier does not currently support returning metrics, so users with it enabled can't see metrics on the clusters tab. Add support for this.


ezimanyi:
Whoops, accidentally deleted the branch; re-created it and reopening this now.

PR #3653, opened 2019-05-09
feat(titus) : Return ServiceJobProcesses details for titus jobs

aravindmd:
- Return Titus `ServiceJobProcesses` details as part of the Titus job details , This is equivalent to `Scaling Processes` in AWS


marchello2000:
LGTM with the caveat that I don't know anything about Titus. But the code itself seems reasonable.

ajordens:
Same boat as @marchello2000.

👍

PR #3656, opened 2019-05-10
fix(core): Spring does not override default configuration files

claymccoy:
In the Spring Boot 2 upgrade, spring.config.location overrides all other config locations. The equivalent property is now called spring.config.additional-location.
This fix allows the app the to run without additional config files.


marchello2000:
Can you also fix this in other boot2 services: e.g. `orca` & `fiat` please?

jkschneider:
@marchello2000 Fixed in orca with https://github.com/spinnaker/orca/pull/2908

marchello2000:
> @marchello2000 Fixed in orca with [spinnaker/orca#2908](https://github.com/spinnaker/orca/pull/2908)

nice, thanks - didn't see that :) I think Fiat still needs it, right?

jkschneider:
@marchello2000 https://github.com/spinnaker/fiat/pull/406

PR #3659, opened 2019-05-10
fix(gate/boot2): Disable LdapHealthIndicatorAutoConfiguration

plumpy:
We can reenable it once we upgrade to a version of spring-boot with the
fix for spring-projects/spring-ldap#473


plumpy:
cc @ezimanyi

plumpy:
Wait a minute, this isn't gate. 😳

PR #3661, opened 2019-05-13
chore(*): Bump spinnaker-dev plugin

ezimanyi:
* chore(*): Apply codestyles to non-project files 

  The next commit pulls in the codestyles plugin, and applies it to the base project (but not to subprojects). In preparation, run the formatter and commit the changes as a single commit.

* chore(*): Bump spinnaker-dev plugin 

  We need the new plugin to set spring.config.additional-location instead of spring.config.location when running services.


ezimanyi:
First commit is just auto-formatting.  I had trouble disabling the codestyle plugin for the base project as it completely removed the `spotlessApply` and `spotlessCheck` tasks; it was easier just to format the base project than to try to debug that.

(Enabling this for subprojects will be more work as it is having trouble parsing some of the files; I haven't looked into that.)

PR #3662, opened 2019-05-13
chore(*): Apply codestyles to clouddriver

ezimanyi:
* chore(*): Fix a few things that break the code formatter 

  The code formatter breaks on:
  * Variables named _
  * Wildcard imports in kotlin files
  * Static imports of annotated classes defined in the same file (as importing the annotation must come before the static import, but the formatter puts all static imports at the top).

  Manually fix those so the formatter can run.

* chore(*): Apply codestyles to clouddriver 

  Run ./gradlew spotlessApply on the entire project

* chore(*): Turn on codestyle enforcement


ezimanyi:
Cherry-picks will be much easier if we do this before rather than after cutting the 1.14 branch.

(Also, I think the static import order issue noted above is fixed in jdk9; I found [this bug ](https://bugs.openjdk.java.net/browse/JDK-8056066)though haven't tested yet as just removing the import fixed it.)

PR #3665, opened 2019-05-14
fix(RRB): Minor fixes in rolling red black

marchello2000:
* Correctly considers instances during disable in RRB if they are not part of discovery
* Better math (e.g. 20% of 5 = 1 not 2!)


clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/deploy/ops/AbstractEurekaSupport.groovy
```diff
@@ -264,36 +264,50 @@ abstract class AbstractEurekaSupport {
     return serverGroup
   }
 
+  /**
+   * Returns a list of instanceIds to disable. Only really used for RollingRedBlack strategy.
+   * The list represents the given percentage of "enabled" instances.
+   * Enabled instance is one that that has at least 1 health provider indicating it's UP
```
dreynaud: seems technically correct, but maybe this no longer belongs in `AbstractEurekaSupport`?

marchello2000: I didn't want to refactor/change too much I will move it around when i do more RRB stuff, how do you feel about that?

PR #3666, opened 2019-05-14
fix(logging): add logstash-logback-encoder to runtimeClasspath

marchello2000:
boot2 regression, adding the library back to facilitate logging


ezimanyi:

PR #3671, opened 2019-05-15
fix(provider/azure): fail to delete application for azure

neil-yechenwei:
Background:
Currently it failed to delete application after deleted all server groups.
Root cause:
While deleting application, it would check whether there is cluster in the cache which is associated with application. If it exists, it throws exception "Application is associated with one or more clusters". After checked, currently there is no logic to delete cluster data from the cache so that it failed. 
The fix:
So this pr would add logic to delete the cluster when there is no server group which is associated with cluster.


neil-yechenwei:
@duftler , a kindly ping, please help to review the change. Thanks.

PR #3674, opened 2019-05-15
fix(appengine): Allow gcloud path to be configured

ezimanyi:
Fixes spinnaker/spinnaker#3349

Clouddriver searches the the PATH for the gcloud executible; in some cases it is easier to provide a direct path to the executable than to fully control the environment of clouddriver.

(One example is running it on Ubuntu 18.04 on GCE; gcloud is installed in /snap/bin, which is added to the path for login shells but is not in the path for systemd.)


ezimanyi:

PR #3676, opened 2019-05-16
fix(runJob): fix v1 job collection

ethanfrogers:
because of the way `JobController` implements collecting jobs, returning
null is the way to signal that a job provider does not apply to a given
request. prior to this fix, jobs run with a v1 account would error out
because the job controller would attempt to gather the status of the job
against the incorrect job provider.


ethanfrogers:

PR #3682, opened 2019-05-16
refactor(*): Remove ProviderSynchronizerTypeWrapper and usage

scottfrederick:
Following aaf1572d9e299a5f4cf3eea91d51a334a77cd251, remove the `ProviderSynchronizerTypeWrapper` class and usage of it from all providers. This class was used to facilitate refreshing configuration via the `/config-refresh` controller, which was removed as part of the integration of Spring Cloud Config refresh and config server. 

Also remove `SCOPE_PROTOTYPE` from `@ConfigurationProperties` beans, as these beans will get refreshed by Spring Cloud Config refresh when they are in the `RefreshScope` context.


jkschneider:
@cfieber I think this gets us closer to being able to just use singletons and get rid of this prototype scoped weird anti-cyclic dependency thing.

scottfrederick:
After these change, I only see one remaining use of `SCOPE_PROTOTYPE` in [`RetrofitConfig`](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/RetrofitConfig.groovy#L41). I suspect that one can be deleted too, but I don't understand that config well enough to know for sure.

scottfrederick:
@ezimanyi @cfieber Any feedback on this mass deletion of code?

PR #3687, opened 2019-05-17
fix(cf): only create artifact metadata for Maven artifacts

claymccoy:
All artifacts don’t use the Artifact fields consistently. For example, the Jenkins artifact uses the build number in the version field. It then looks incorrect when it is later displayed as a version.


jkschneider:

PR #3688, opened 2019-05-17
fix(kubernetes): Improve error message on unbound artifact

ezimanyi:
Fixes spinnaker/spinnaker#3571

The error message when a required artifact is not bound in a Kubernetes deploy or patch stage is confusing. Make this clearer so that users get more actionable feedback when they misconfigure a stage.


teabolt:
Maybe this message could be further improved by adding a "Possible causes:" text:
`
The following required artifacts could not be bound: '$ARTIFACTS'. Possible causes: * Check that the Docker image name above matches the name used in the image field of your manifest. Failing the stage as this is likely a configuration error.
`

The reason for this is that the docker image not being present is not always the actual cause for the failure. In my case I had accidentally set `Required Artifacts to Bind` in my deploy stage which resulted in missing artifacts.

PR #3690, opened 2019-05-20
fix(titus): Support Titus value base constraints

aravindmd:
**Context** : The Titus API supports `soft` and `hard` constraints which allow a user to change the behavior of the placement decision , The constraints are modeled as key/value pairs such that you can submit a value. The current Spinnaker implementation assumes that constraints are just keys or names and values cannot be passed in . 

This change allows users to set `softConstraints` and `hardConstraints` as key/value pair as part of `constraints` . It continues to support the previous format as well , however `constraints` takes precedence when specified.


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/SubmitJobRequest.java
```diff
@@ -56,6 +58,13 @@ public String getConstraint() {
     }
   }
 
+  @Data
+  public static class Constraints {
+
+    public Map hardConstraints;
```
ajordens: thoughts on mapping to just `hard` and `soft` 

```
constraints: {
  hard: {},
  soft: {}
}  
```

vs

```
constraints: {
  hardConstraints: {},
  softConstraints: {}
}  

```

ajordens:
So is the bigger problem here that we don't have the ability to supply a value for a constraint?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/SubmitJobRequest.java
aravindmd: that looks better , i will update it

aravindmd:
Yes the problem is not being able to set values and also not break the old contract

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/JobDescription.java
```diff
@@ -500,12 +512,21 @@ public JobDescriptor getGrpcJobDescriptor() {
       containerBuilder.putAllEnv(env);
     }
 
-    if (!softConstraints.isEmpty()) {
-      containerBuilder.setSoftConstraints(constraintTransformer(softConstraints));
-    }
+    if (constraints != null) {
```
marchello2000: just curious, when is `constraints` there vs. `soft/hardConstraints`?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/SubmitJobRequest.java
```diff
@@ -91,6 +100,9 @@ public String getConstraint() {
   private MigrationPolicy migrationPolicy = null;
   private DisruptionBudget disruptionBudget = null;
 
+  @JsonProperty("constraints")
```
marchello2000: do you need this annotation given that you have one on line 428?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
```diff
@@ -313,22 +313,25 @@ class TitusDeployHandler implements DeployHandler<TitusDeployDescription> {
...
-
-    if (description.softConstraints) {
-      description.softConstraints.each { constraint ->
-        submitJobRequest.withConstraint(SubmitJobRequest.Constraint.soft(constraint))
-      }
-    }
-
-    if (description.jobType == "service" && !description.hardConstraints?.contains(SubmitJobRequest.Constraint.ZONE_BALANCE) && !description.softConstraints?.contains(SubmitJobRequest.Constraint.ZONE_BALANCE)) {
-      submitJobRequest.withConstraint(SubmitJobRequest.Constraint.soft(SubmitJobRequest.Constraint.ZONE_BALANCE))
+    // constraints map takes precedence when specified in the request
```
marchello2000: this is a bit convoluted (not your fault - obv titus requirements) but maybe you can add a more descriptive comment that explains what's up

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/SubmitJobRequest.java
aravindmd: yes that's not required , removed

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusDeployHandler.groovy
aravindmd: Agree , added a comment for now , I think i will follow up this with what @ajordens suggested with a preprocessor , that will make this better i think

PR #3692, opened 2019-05-20
chore(dependencies): align on guava version from caffeine

dreynaud:


emjburns:
Can you explain to me why we need to use this other package? I have no knowledge about guava so I just picked one of them when I upgraded deps

robfletcher:
@emjburns the `spring-boot-dependencies` BOM's opinion on Guava is derived from `caffeine:guava`. Relying on that ensures we won't pull in a version of Guava that breaks Spring.

PR #3696, opened 2019-05-21
chore(dependencies): korkVersion 5.3.5 for guava alignment

cfieber:


cfieber:
it actually looks like this doesn't work as expected in practice - something is version-managing our constraint up to 26.0-android..

cfieber:
google cloud http-client api from the looks of it..

cfieber:
so this seems to trim it down to two guava versions remain, 24.0-jre in any module that doesn't bring in the google http-client transitively and 26.0-android otherwise (with 26.0-android landing in clouddriver-web)

progress, but still not guava-highlander

cfieber:
closing in favour of #3701

PR #3697, opened 2019-05-21
fix(runJob/kubernetes): collect pods and status

ethanfrogers:
In order for the Deck to display logs in a reliable manner we need
access to any Pod names created by the Job. When collecting the status
of the Job, collect the names and statuses of any Pods owned by the Job.
When displaying logs, Deck will determine the Pod name use when fetching
console output.


clanesf:
@ethanfrogers @ezimanyi - Does this really need to wait until 1.15 or can we get it cherry picked into 1.14.1. We really need this for our EKS implementation at SFDC.

ethanfrogers:
@clanesf my goal is to pick this into 1.14.1. I haven't done it yet because the UI isn't complete. Once that has been merged I'll cherry pick both.

ethanfrogers:

PR #3698, opened 2019-05-21
fix(kubernetes): Fix compiler warnings

ezimanyi:
There are a few compiler warnings in clouddriver-kubernetes; fix these.

* Defaulting a field on a class that uses a `@Builder`, as `@Builder` ignores the default.  For these I either added `@Builder.Default` to tell the builder to use the default, or removed `@Builder` if it was not necessary (unused or for a simple class where all builders passed the same parameters).
* EqualsAndHashCode without calling the super class
* Passing an array instead of a list to a log call

Also, for some of the modified classes, took the opportunity to remove un-needed lombok annotations and clean up the interface (as well as make some fields final where possible).


ezimanyi:
I've been meaning to do this for a while, every time I see a bunch of warnings compiling `clouddriver-kubernetes`

PR #3704, opened 2019-05-22
fix(cf): cacheTime should be stored as a long

claymccoy:
This will avoid exceptions this:
"Cannot compare java.lang.String with value '2019-05-22T03:51:57.222+0000' and java.lang.Long with value '1,558,496,930,016'"

That happen here:

https://github.com/spinnaker/orca/blob/fdc5f6a53affddc56051171accd97493fdc7b660/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/ServerGroupCacheForceRefreshTask.groovy#L181

because cacheTime is expected to be a long.

Screen Shot 2019-05-21 at 11 41 29 PM


jkschneider:

PR #3711, opened 2019-05-22
fix(kubernetes): add cleaned up artifact manifests to output

maggieneterval:
Addresses: https://github.com/spinnaker/spinnaker/issues/4422
Related PR: https://github.com/spinnaker/orca/pull/2933

When using Kubernetes V2 Rollout Strategies with the `max-version-history` annotation, Clouddriver sometimes tries to disable a previously-deleted ReplicaSet because we are not force refreshing the cache after the Deploy Manifest stage's `CleanupArtifactsTask`. In order to add a functioning `ManifestForceCacheRefreshTask`, we need to add the cleaned-up manifest(s) to the `OperationResult`'s `manifestNamesByNamespace`.

This fix only addresses this issue when `liveManifestCalls` is not enabled.


maggieneterval:

PR #3712, opened 2019-05-23
fix(provider/azure): Fix several issues in Azure Load Balancer

michaeljqzq:
This pull request fix the following issues:
* When sshpublickey is not presented, deploy server group breaks
* When VNet and SG is not in same resource group, deployment fails
* When deployment SG with ALB failed, the rollback operations can't finish correctly
* When update ALB, the backend information in the ALB lost
* SG NIC bind with ALB

We hope this PR could be applied to v1.14.0 as a hotfix, thanks!


michaeljqzq:
This pull request is linked with https://github.com/spinnaker/deck/pull/7054

PR #3715, opened 2019-05-23
refactor(gce): introduce a WaitableComputeOperation

plumpy:
cc @jtk54


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/compute/RegionalOperation.java
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  @Override
+  public Operation waitForDone(Task task, String phase) {
+    return poller.waitForRegionalOperation(
+        credentials.getCompute(),
+        credentials.getProject(),
+        operation.getRegion(),
+        operation.getName(),
+        /* timeoutSeconds= */ null,
```
jtk54: What's the rationale for keeping this null vs making it a parameter?

plumpy: Just YAGNI... if I ever get to a point where we need it, I'll make it a parameter then. I didn't check all the call sites, but I checked more than a handful and I never saw any that actually used a value other than null.

PR #3716, opened 2019-05-23
feat(provider/kubernetes): Deploy kubernetes kind List

german-muzquiz:
This PR adds support to Deploy Manifest stage (kubernetes v2 provider) to deploy manifests of kind `List`.

`List` is a special kind that is used only to have several kubernetes resource definitions in a single yaml file, the same as separating each resource with dashes (`---`). `List` is not a kind that is available in Kubernetes API like `Service`, `Pod`, etc.

Today trying to deploy this List manifest fails:

```
apiVersion: v1
kind: List
metadata:
  name: list-test
  namespace: default
items:
  - apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: nginx-deployment
      namespace: default
    spec:
      replicas: 1
      selector:
        matchLabels:
          app: nginx
      template:
        metadata:
          labels:
            app: nginx
        spec:
          containers:
            - image: 'nginx:1.14.0'
              name: nginx
              ports:
                - containerPort: 80
  - apiVersion: v1
    kind: Service
    metadata:
      labels:
        app: nginx
      name: nginx-service
      namespace: default
    spec:
      ports:
        - port: 80
          protocol: TCP
          targetPort: 80
      selector:
        app: nginx
      sessionAffinity: None
      type: ClusterIP
```

But doing a `kubectl apply -f file.yml` is successful.


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/security/AbstractAtomicOperationsCredentialsSupport.groovy
```diff
@@ -37,6 +37,10 @@ abstract class AbstractAtomicOperationsCredentialsSupport implements AtomicOpera
 
   }
 
+  ObjectMapper getObjectMapper() {
```
ncknt: I don't think it's necessary. The getter should be implicit.

ezimanyi:
I'm on the fence about whether adding this is worth the added complexity to the deploy manifest stage.  Have users expressed a desire to do multi-manifest deploys using this syntax vs. separating the manifests with `---`?

dbyron0:
I'm pretty sure I ran into this when I tried using an istio helm chart which has e.g. https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/charts/tracing/templates/service.yaml

ncknt:
Yep. @german-muzquiz mentioned he ran into the issue when deploying Spinnaker with Istio.

ezimanyi:
Ah, ok, I didn't realize this was a common pattern for applying things to your cluster (vs. just a container for getting things back).  In that case, it probably makes sense for us to support it the same way we support splitting manifests with `---`.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifest.java
```diff
@@ -52,7 +52,7 @@ public KubernetesKind getKind() {
     // using ApiVersion here allows a translation from a kind of NetworkPolicy in the manifest to
     // something
     // like  NetworkPolicy.crd.projectcalico.org for custom resources
-    String kindName = getRequiredField(this, "kind");
+    String kindName = getKindName();
```
ezimanyi: Why did you pull this out into a function?

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/converter/manifest/KubernetesDeployManifestConverter.java
```diff
@@ -50,13 +62,61 @@ public AtomicOperation convertOperation(Map input) {
...
+  /**
+   * If present, converts a KubernetesManifest of kind List into a list of KubernetesManifest
+   * objects.
+   *
+   * @param mainDescription deploy manifest description as received.
+   * @param objectMapper used to generate the input deploy manifest description.
+   * @return updated description.
+   */
+  @SuppressWarnings("unchecked")
+  private KubernetesDeployManifestDescription convertListDescription(
```
ezimanyi: I'm wondering whether this would live better in the actual `KubernetesDeployManifestOperation` class.  In general the converter is intended to just deserialize the operation with more complex logic being handled in the actual operation itself.  Anything else that requires actually looking at the manifests and manipulating them (such as looking up namespaces, etc) is in the operation.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifest.java
german-muzquiz: I pulled this out to be public a method because I needed the kind name as a string instead of a `KubernetesKind` object, since there's no `KubernetesKind` for `List`. Also, using `getKind` on a non-registered kind has the effect of dynamically creating a new kind for it (used for custom resource definition objects), which then has other side effects like trying to cache those kinds.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/converter/manifest/KubernetesDeployManifestConverter.java
german-muzquiz: That's also another possibility, `operate` function already performs some initial logic over the list of input `KubernetesManifest` objects. I don't know however if there are other side effects of delaying manifest splitting until the operation actually executes, but I can test if there are no problems related with caching for instance. 

A drawback is that there's no `ObjectMapper` in the operation, then we would end up having two different paths for converting an input `Map` from the request into `KubernetesManifest`: one in the converter used for everything and other in the operation used only when resources are grouped in `List` manifests. If the way that deserializing manifests changes (i.e. using a custom ObjectMapper), there would be two places to change and the `List` one would not be obvious. That said, I have no problem moving the split to the operation itself.

german-muzquiz: @ezimanyi given the above, do you think it's better to move this logic to `KubernetesDeployManifestOperation`?

ezimanyi: Yeah, as I think of it more, it probably makes sense to have all of the unwrapping of manifests happen in the same place---both so that they can use the same `objectMapper` but also so that downstream logic can just assume it has a list of manifests to deploy.  I think it's fine to leave in this file.

clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/converter/manifest/KubernetesDeployManifestConverterTest.groovy
```diff
@@ -0,0 +1,127 @@
...
+
+class KubernetesDeployManifestConverterTest extends Specification {
+
+  def converter = new KubernetesDeployManifestConverter()
+  def inputMap = new HashMap()
+  def manifestList = new ArrayList()
+  def gsonObj = new Gson()
+
+  def setup() {
+    converter.accountCredentialsProvider = Mock(AccountCredentialsProvider)
```
ezimanyi: If you convert the `KubernetesDeployManifestConverter` to take the other dependencies in the constructor (and set the fields to `private final`) then it will be easier to just create the mock via the constructor rather than having to set all properties.

In general, we're moving to use constructor injection instead of field injection as a best practice, so this would also be moving us towards that.

clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/converter/manifest/KubernetesDeployManifestConverterTest.groovy
```diff
@@ -0,0 +1,127 @@
...
+import groovy.text.SimpleTemplateEngine
+import spock.lang.Specification
+
+
+class KubernetesDeployManifestConverterTest extends Specification {
+
+  def converter = new KubernetesDeployManifestConverter()
+  def inputMap = new HashMap()
+  def manifestList = new ArrayList()
+  def gsonObj = new Gson()
```
ezimanyi: I'd suggest just using the existing `ObjectMapper` and doing `objectMapper.readValue()` rather than introduce a new deserialization library.

clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/converter/manifest/KubernetesDeployManifestConverterTest.groovy
german-muzquiz: Agreed with constructor injection. I moved dependencies of this converter to constructor, but there are a couple defined in the parent class. I didn't changed the parent because that would impact almost 200 subclasses though.

PR #3717, opened 2019-05-24
fix(titus): Return job constraints with serverGroup details

aravindmd:
- follow up to #3690, return `constraints` map as part for `TitusServerGroup` to keep it consistent.

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/Job.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/Job.java
@@ -71,6 +71,8 @@ public class Job {
   private String jobState;
   private ServiceJobProcesses serviceJobProcesses;
 
+  private SubmitJobRequest.Constraints constraints;
+
   public Job() {}
 
   public Job(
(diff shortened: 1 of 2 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/Job.java
```diff
@@ -146,6 +148,17 @@ public Job(
     iamProfile = grpcJob.getJobDescriptor().getContainer().getSecurityProfile().getIamRole();
     allocateIpAddress = true;
     submittedAt = new Date(grpcJob.getStatus().getTimestamp());
+    constraints = new SubmitJobRequest.Constraints();
+    if (grpcJob.getJobDescriptor().getContainer().getHardConstraints().getConstraintsMap()
```
marchello2000: nit: prob be nicer to have a "temp" var:
```
constraintsMap = grpcJob.getJobDescriptor().getContainer().getHardConstraints().getConstraintsMap()
if (constraintsMap != null) {
    ....setHard(constraintsMap);
}
```

aravindmd: 👍

PR #3718, opened 2019-05-24
fix(gce): fix waiting on regional operations from the new MIG compute wrapper

plumpy:
I tested this locally with a ZonalOperation and fixed it there. I'm not
sure why it didn't occur to me to also make the same fix in
RegionalOperation.


plumpy:
cc @jtk54

PR #3724, opened 2019-05-28
feat(cats): Allow partial cache update

xavileon:
Currently, cache providers (either Default or SQL) evicts all those
keys not present on the CacheResult. This prevents doing partial
updates in on-demand caching agents, which may come in handy when
the query operation is a costly one and only a subset of the
keys may actually need updating.

This patch introduces the notion of a partial CacheResult, which allows
keeping the keys not present on the CacheResult if the partialResult
flag is set to true. The default behaviour is kept as it is today to keep backwards
compatibility.

Related comment: https://github.com/spinnaker/clouddriver/pull/3576#discussion_r283163664


xavileon:
@robzienert This is related to a comment on a separate PR, to allow partial updates of the cache. I'd love to hear your thoughts on this! :)

xavileon:
@asher ah cool, I wonder how I missed that 😓 I can move this at the CachingAgent level then, so it can be reused by other agents. Now, the CachingAgent unconditionally calls the `putCacheResult` no matter what, the idea is to check if the result is a partial update and call the `addCacheResult` instead. How does that sound?

asher:
@xavileon that sounds reasonable. Note, if you're trying to target OnDemand updates specifically, I think you'll also want to update `CatsOnDemandCacheUpdater.handle` (https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.groovy#L100)

xavileon:
addressed in https://github.com/spinnaker/clouddriver/pull/3724/commits/64169e5f6896d1d8cee208b10ec3b1d0c47452cd

xavileon:
hi @asher ! could you double-check when you have some time please?

xavileon:
@asher bump :)

xavileon:
hello @asher does the PR looks good after the changes?

PR #3731, opened 2019-05-31
feat(cfn/changeset): Introduce CFN change sets

xavileon:
This patch is a step towards fully supporting CloudFormation change
sets in Spinnaker. Specifically, this patch brings:
- Extend the deployCloudFormation operation with a flag indicating
wether it's a regular stack or a change set.
- If it's a change set, the create change set API call in AWS is
called.
- The on-demand cahing agent now only updates a single stack
if the stack is provided in the data map. This makes the
force cache refresh operation much more efficient by only
updating the relevant stack (instead of iterating over all
stacks in a given account/region).

Depends on https://github.com/spinnaker/clouddriver/pull/3724


thecodesmith:
@xavileon I have a situation where I would like to deploy an RDS database with Cloudformation, and then use the database hostname output from the stack as a parameter to a Deploy (Manifest) stage, so that the deployed app can connect to the database. This is a common pattern for apps that my team deploys. Currently the Spinnaker Deploy (Cloudformation Stack) stage does not capture the stack outputs. We are trying to decide whether to stick with a workaround, or invest the time to contribute such a feature back. Does your PR here handle this? I see that it is adding some metadata with a `DescribeStacksRequest`, but can't tell if it solves our use case. Thanks!

xavileon:
hi @thecodesmith! I think this PR solves already what you describe :) https://github.com/spinnaker/orca/pull/2911 We just need getting it merged and will support your use case

xavileon:
@spinnaker/reviewers PTAL

xavileon:
@spinnaker/reviewers bump :)

maggieneterval:
Hey @xavileon, I'm really sorry these PRs have sat here for so long. Is there anyone in particular you think would be a good reviewer for this PR? Looks like you were the author of most of the affected files. I'm also happy to merge these along with the Deck PR if there isn't anyone else you think would be useful to have as an approver.

xavileon:
Hi @maggieneterval no one in particular. I understand the business logic can be time-consuming to review if not familiar with the functionality. I was pinging just in case someone wanted to check. I'm ok if you merge the related PRs as is, if nobody else can review it 👍 
also:
https://github.com/spinnaker/orca/pull/2950
https://github.com/spinnaker/deck/pull/7071
Thanks!!

PR #3734, opened 2019-05-31
fix(kubernetes): Properly handle empty kubectl input stream

ezimanyi:
If kubectl throws an error, it will generaly return a nonzero status, an empty stdout, and the error in stderr. The streaming manifest parsing logic fails when the output stream is empty, so we're propagating the error parsing the empty input stream instead of the actual error message.

If the output of kubectl is empty, just return an empty list of manifests. The calling code will the properly look at the return code and stderr to determine whether the call succeeded and what message to show the user.


ezimanyi:

PR #3735, opened 2019-05-31
fix(eureka): Handles duplicate eureka records

cfieber:
This adds deterministic handling when we get multiple records for the
same instance id. Previously we would potentially generate two cache
data objects, however since they both had the same id only one would be
written (last wins style most likely).

This now prefers a record in a 'worse' state (based on ordering the
HealthState enum from worst to best) to ensure that an instance showing
as Down is not accidentally considered ready for traffic during a
deploy.

If all records agree on health state then takes the most recently
updated.

### DIFF
--- a/clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/model/EurekaInstance.groovy
+++ b/clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/model/EurekaInstance.groovy
@@ -86,6 +86,16 @@ class EurekaInstance extends DiscoveryHealth {
       healthState = HealthState.Down
     }
 
+    // if this has an asgName and is not part of a titus task registration,
+    // prefer the app name derived from the asg name rather than the supplied
+    // app name. We index these records on application to associate them to
+    // a particular cluster, and with the name incorrect then the record will
+    // not be properly linked
+    if (metadata?.titusTaskId == null && asgName != null) {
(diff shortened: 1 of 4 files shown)


clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/model/EurekaInstance.groovy
```diff
@@ -86,6 +86,16 @@ class EurekaInstance extends DiscoveryHealth {
...
 
+    // if this has an asgName and is not part of a titus task registration,
+    // prefer the app name derived from the asg name rather than the supplied
+    // app name. We index these records on application to associate them to
+    // a particular cluster, and with the name incorrect then the record will
+    // not be properly linked
+    if (metadata?.titusTaskId == null && asgName != null) {
+      def idx = asgName.indexOf('-')
+      def appFromAsg = idx == -1 ? asgName : asgName.substring(0, idx)
+      app = appFromAsg
```
dreynaud: I don't see how this relates to the duplicate eureka records 🤔

cfieber: it has to do with how we index and load the health records by server group, if the app name is incorrect our lookup by app query doesn't find the records ( I stuck a big comment in the code to that effect )

PR #3736, opened 2019-06-01
perf(kubernetes): Reduce memory allocation during caching cycles

ezimanyi:
* test(kubernetes): Add tests to getClusterRelationships 

  Also simplify the logic in the invertRelationships test

* perf(kubernetes): Reduce memory allocation during caching cycles 

  The caching logic for the kubernetes v2 provider allocates a lot of short-term memory during each caching cycle, putting pressure on the garbage collector, in some cases exceeding the garbage collection overhead.

  A significant contributor is the logic to compute relationships between kubernetes objects. We currently create a CacheData object to hold each relationship and rely on downstream (and inefficient) logic to merge these into one CacheData per object, containing all of its relationships.

  Improve this by having invertRelationships return one CacheData per object (containing all of its relationships) rather than on CacheData per relationship. Likewise, have getClusterRelationships return a single CacheData for each application, rather than a separate one for each
 application-object relationship.


ezimanyi:
This should help significantly with spinnaker/spinnaker#3287.

ezimanyi:

PR #3740, opened 2019-06-03
fix(ecs): Don't specify load balancer role for non-load-balanced service

clareliguori:
Also improve unit tests around CreateService calls

This will need to be cherry-picked into 1.14

Fixes spinnaker/spinnaker#4475


ezimanyi:

PR #3741, opened 2019-06-03
perf(kubernetes): Improve performance of cache kind lookup

ezimanyi:
* test(kubernetes): Add tests to Keys 

  An upcoming commit will make some performance improvements to the Keys.Kind and Keys.LogicalKind enums, so add some tests to these.

  Also add some tests to validate the parsing of cache keys containing semicolons. This is currently broken, but add the test now validating the incorrect behavior and update the test in the commit that fixes it.

* fix(kubernetes): Properly substitute characters in key parts 

  As kubernetes objects can contain the character ':' which conflicts with our default cache key separator, the caching logic replaces ':' with ';' before generating a cache key and then inverts this by replacing ';' with ':' when splitting a cache key into its component parts.

  The current logic that runs when reading cache keys is an expensive no-op; it substitutes ';' with ':' in the key part but ignores the results (as the result of replaceAll is ignored).

  In addition to fixing the bug, add a performance optimization. replaceAll always allocates a new string, regardless of whether it actually needs to perform a substitution; this means that we're allocating a new string for every cache key part we read from the cache (~6x the number of keys we read) even though most keys will not have a ':'. Perform a relatively cheap contains check before doing the replacement.

* perf(kubernetes): Improve performance of cache kind lookup 

  Looking up a kind and a logicalKind by name (which happens frequently on reads from the cache) are fairly inefficient.  We use a stream to iterate over all current enum values comparing the input string to calling toString on the enum using a case-insensitive comparison.

  toString is unnecessarily expensive as it allocates a new string on each call via calling .toLowerCase.  Instead, store the lower-case name as a property of the enum and return a reference to it in toString.

  Improve fromString by just converting the input string to upper case and using the enum's valueOf function, rather than looping over the values and doing a case-insensitive search.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/Keys.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/Keys.java
@@ -131,8 +131,10 @@ public class Keys {
       return Optional.empty();
     }
 
-    for (String part : parts) {
-      part.replaceAll(";", ":");
+    for (int i = 0; i < parts.length; i++) {
+      if (parts[i].contains(";")) {
+        parts[i] = parts[i].replaceAll(";", ":");
+      }
     }
 
     try {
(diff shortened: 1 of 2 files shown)


ezimanyi:

PR #3748, opened 2019-06-05
fix(kubernetes): Allow namespaces that don't exist in validation

ezimanyi:
Fixes spinnaker/spinnaker#4482

In 1.14, we tried to improve the failure mode for users with misconfigured accounts by using the actual live list of namespaces while validating accounts.

This fails when users are creating a namespace and deploying something to the namespace as part of the same operation; in that case the live list does not know about the namespace (but will by the time the deployment completes).

This reverts spinnaker/clouddriver#3639 (and adds a comment and test to make sure this subtle bug is not re-introduced).


ezimanyi:
:bug: :no_entry_sign: :hammer:

maggieneterval:

PR #3752, opened 2019-06-05
fix(cf): remove sensitive information from credentials payload

Jammy-Louie:


jkschneider:

PR #3753, opened 2019-06-05
feat(ecs): DRAFT: Add support for 'command' field and volume/mounts on ECS

justinrlee:
Support specifying 'command', 'volume', and 'mountPoints' in an ECS deployment definition.

Does not yet have tests (these will be forthcoming).  PR to track work.

A bit of weirdness: Due to the fact that ECS is currently limited to a single container, 'volumes' goes to the task definition and 'mountPoints' goes to the container definition.


robzienert:
Ping - what's the status of this PR?

PR #3758, opened 2019-06-06
perf(kubernetes): Remove cluster relationship from pod kind

ezimanyi:
Pods are currently set as having a cluster relationship; this means that we create a cluster (and associated relationship to the pod) for every pod in a kubernetes deployment.

This is not necessary, and doesn't match the definition of cluster used by other providers.  These clusters and relationships are also not used when we construct the server group model for
a Kubernetes application; we end up ignoring any cluster that does not map to a server group (which pods do not).

This change will significantly improve the efficiency of both the caching logic, and the logic to return server groups for kubernetes (which now will not need to process each pod individually).


maggieneterval:
oh wait you haven't seen star wars! tl;dr is baby darth vader competes in a "pod" racing competition in order to buy his freedom and escape from his mum / the planet of tatooine (aka his "cluster relationships")

PR #3763, opened 2019-06-07
fix(docker): Tweak messages for exception and log

dragon3:
## WHAT

This pull request tweaks messages for exception and log.
It's tiny change 😄 

## WHY

I found that it's better to add white space and ":" to the messages.
For example, I saw this log:
```
2019-06-07 07:40:24.315  WARN 1 --- [cutionAction-83] c.n.s.c.d.r.a.v.c.DockerRegistryClient   : Error encountered during catalog of v2/_catalog?n=100&last=foo/bar429 Too Many Requests
```
There is no space before `429 Too Many Requests`, it's hard to read or parse.


clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -322,7 +322,7 @@ class DockerRegistryClient {
         String userDefinedCatalog = new File(catalogFile).getText()
         return (DockerRegistryCatalog) new Gson().fromJson(userDefinedCatalog, DockerRegistryCatalog.class)
       } catch (Exception e) {
-        throw new DockerRegistryOperationException("Unable to read catalog file $catalogFile:" + e.getMessage())
+        throw new DockerRegistryOperationException("Unable to read catalog file $catalogFile: " + e.getMessage())
```
marchello2000: since you're changing this
```suggestion
        throw new DockerRegistryOperationException("Unable to read catalog file $catalogFile: " + e.getMessage(), e)
```

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -336,7 +336,7 @@ class DockerRegistryClient {
           registryService.getCatalog(paginateSize, token, userAgent)
       }, "_catalog")
     } catch (Exception e) {
-      log.warn("Error encountered during catalog of $path" + e.getMessage())
+      log.warn("Error encountered during catalog of $path: " + e.getMessage())
```
marchello2000:
```suggestion
      log.warn("Error encountered during catalog of $path", e)
```

that way you get the whole stack trace, etc

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
dragon3: Fixed in a2a414c

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
dragon3: Fixed in a2a414c as well.

dragon3:
Hi @marchello2000 
Thank you for your feedback.

I'd just like to fix the messages, but as your feedback, it's better to set cause to the exception and log.
So I pushed another commit for it.
Please take a look at it.

Thanks!

PR #3765, opened 2019-06-07
fix(ecs): Handle container instances no longer being cached

clareliguori:
Handle container instances no longer being cached, but showing up in stopped ECS tasks in the account.  This seems to happen when there is a high rate of change among container instances, for example when using a Spot fleet.

This can be cherry-picked into 1.14.

Fixes spinnaker/spinnaker#4466


ezimanyi:

PR #3771, opened 2019-06-10
fix(appengine): Use getOutput() instead of getStdOut().

duftler:


duftler:

PR #3776, opened 2019-06-12
fix(ecs): If in "host" network mode, host and container port should b…

justinrlee:
…e the same

When creating a task definition with the host network mode and specifying a containerPort, the containerPort should be used for both the host and container port.

Can be cherry-picked into 1.14.


justinrlee:

ezimanyi:

PR #3779, opened 2019-06-12
fix(cf): fail destoryService stage if service not found

Jammy-Louie:


Jammy-Louie:
Screen Shot 2019-06-12 at 12 10 39 PM

jkschneider:

PR #3782, opened 2019-06-13
fix(ecs): Fix incorrect de-duping of clusters

clareliguori:
Fix incorrect de-duping of clusters with same name, but different accounts

Fixes spinnaker/spinnaker#4448

Can be cherry-picked into 1.14


ezimanyi:

PR #3785, opened 2019-06-13
perf(google): improve performance of gce image fetching

maggieneterval:
Addresses spinnaker/spinnaker#4510

Per the above issue and comments on [this Slack conversation](https://spinnakerteam.slack.com/archives/C091CCWRJ/p1560262702208300), some Spinnaker users have tens of thousands of images per account. This currently results in a nearly unusable Deploy modal UI, and failing Find Image stages.

Before:
- Fetch all images from all accounts from the cache, even if the `account` param is specified.
- Call `matchesTagFilter` and `getTagFilter` for every image fetched from the cache, even if no tag filters are specified.

After:
- If `account` param is specified, fetch only images of that account from the cache.
- Call `getTagFilter` once per request. Only call `matchesTagFilter` once per image if there is a tag filter on the request.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleNamedImageLookupController.java
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleNamedImageLookupController.java
@@ -31,22 +31,18 @@ import com.netflix.spinnaker.clouddriver.google.GoogleCloudProvider;
 import com.netflix.spinnaker.clouddriver.google.cache.Keys;
 import groovy.util.logging.Slf4j;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import javax.servlet.http.HttpServletRequest;
-import org.apache.commons.lang.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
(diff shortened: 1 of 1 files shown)


maggieneterval:
@duftler Thank you for adding wonderful tests to this controller, it made me so much less scared to refactor!

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleNamedImageLookupController.java
```diff
@@ -73,31 +69,41 @@ private GoogleNamedImageLookupController(Cache cacheView) {
...
-        NamedImage newImage =
-            new NamedImage(imageAccount, i.getName(), attributes, buildTagsMap(i));
-        results.add(newImage);
-      }
+  private Collection<CacheData> getImageCacheData(String account) {
+    // If no account supplied, return images from all accounts
+    String pattern =
+        account == null
+            ? String.format("%s:images:*", GoogleCloudProvider.getID())
+            : String.format("%s:images:%s*", GoogleCloudProvider.getID(), account);
```
ezimanyi: I think you want a colon after `%s` to avoid pulling images for all accounts starting with `account` (rather than accounts exactly matching `account`).

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleNamedImageLookupController.java
```diff
@@ -73,31 +69,41 @@ private GoogleNamedImageLookupController(Cache cacheView) {
...
-        Map<String, Object> attributes = new HashMap<>();
-        attributes.put("creationDate", i.get("creationTimestamp"));
-        NamedImage newImage =
-            new NamedImage(imageAccount, i.getName(), attributes, buildTagsMap(i));
-        results.add(newImage);
-      }
+  private Collection<CacheData> getImageCacheData(String account) {
+    // If no account supplied, return images from all accounts
+    String pattern =
+        account == null
```
ezimanyi: I'm not sure if this is a significant change, but prior to this change we'd return images for all accounts when the account is the empty string, whereas now we only return all accounts when it's null.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleNamedImageLookupController.java
maggieneterval: Wow really good catch, thanks Eric!

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleNamedImageLookupController.java
maggieneterval: Another great catch for @ezimanyi :baseball: Adding `.orEmpty` check!

maggieneterval:
@duftler Not yet, Eric's going to show me how to use his profiling tool so I plan to!

PR #3790, opened 2019-06-14
refactor(gce): Add a wrapper around compute.instanceTemplates()

plumpy:
cc @ezimanyi


plumpy:
Oops, added copyright headers to everything in the 'compute' directory (new and old).

PR #3795, opened 2019-06-19
feat(auth): Optionally skip authentication for image tagging

jervi:
Adds a new feature flag `operations.security.allowUnauthenticatedImageTaggingInAccounts` that takes a list of account names, and will skip the authentication check for tagging cloud images in AWS or GCE if the account name matches. The motivation is that if the bakery puts images in a protected account, any tagging stage will fail if the user or service account that triggered the pipeline does not have permissions to write to the account that contains the image, even if the same user just created the image. The feature defaults to disabled to not change the current behaviour.


ajordens:
Not a huge fan of doing it in this spot ... I'd probably lean towards something like the `requiresApplicationRestriction()` that exists.

Maybe `requiresAccountRestriction()` and the Image tagging converter could have responsibility for setting it on the description.

It strikes me as heavy handed to pull this into the `OperationsController`.

jervi:
@ajordens Thank you for the review :) I was vacationing last week, so sorry for the slow feedback. I've updated the PR now, would you mind taking another look? I skipped the converter-suggestion, but I can move the functionality there if you like that better.

clouddriver-core/src/test/groovy/com/netflix/spinnaker/clouddriver/deploy/DescriptionAuthorizerSpec.groovy
```diff
@@ -68,4 +100,22 @@ class DescriptionAuthorizerSpec extends Specification {
       this.names = names
     }
   }
+
+  class TestImageTaggingDescription implements AccountNameable {
```
jervi: I couldn't access the actual implementations because `clouddriver-core` doesn't depend on the aws or google modules. Do you prefer this test class implementation, or should I depend on the aws and google modules using gradle `testImplementation`?

enrichman:
Any update on this? :)

jervi:
@spinnaker/reviewers PTAL :)

clouddriver-core/src/test/groovy/com/netflix/spinnaker/clouddriver/deploy/DescriptionAuthorizerSpec.groovy
cfieber: I prefer this rather than adding the dependency, thanks!

PR #3796, opened 2019-06-19
feat(kubernetes): Add support for excluding event loading on manifest…

mtweten:
…s retrieval

This change allows optionally skipping kubernetes event retrieval when retrieving manifests. 

When live manifest mode is enabled, scenarios that result in a lot of manifests being loaded will put a lot of extra pressure on the kubernetes API server due to event loading. 

Specifically, we ran into this scenario when we had a lot of deployment pipelines running concurrently, and each deployment involved a decent number of manifests (~30). In Orca, the WaitForManifestStableTask (when live manifest mode is enabled) iterates through the manifests involved in the deployment and calls the retrieve manifest API for each one, which causes a lot of full event loads to occur, which tanks (in our case) the kubernetes API server. However, WaitForManifestStableTask does not actually need any event information to do its job, so in that particular case we can skip loading events entirely.

See https://github.com/spinnaker/orca/pull/2997 for the corresponding Orca change.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ManifestProvider.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ManifestProvider.java
@@ -25,19 +25,8 @@ public interface ManifestProvider<T extends Manifest> {
     SIZE
   }
 
-  T getManifest(String account, String location, String name);
-
   T getManifest(String account, String location, String name, boolean includeEvents);
 
   List<T> getClusterAndSortAscending(
       String account, String location, String kind, String app, String cluster, Sort sort);
-
-  List<T> getClusterAndSortAscending(
-      String account,
-      String location,
-      String kind,
-      String app,
-      String cluster,
-      Sort sort,
-      boolean includeEvents);
 }
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ManifestProvider.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ManifestProvider.java
@@ -27,6 +27,17 @@ public interface ManifestProvider<T extends Manifest> {
 
   T getManifest(String account, String location, String name);
 
+  T getManifest(String account, String location, String name, boolean includeEvents);
+
   List<T> getClusterAndSortAscending(
       String account, String location, String kind, String app, String cluster, Sort sort);
+
+  List<T> getClusterAndSortAscending(
+      String account,
+      String location,
+      String kind,
+      String app,
+      String cluster,
+      Sort sort,
+      boolean includeEvents);
 }
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2LiveManifestProvider.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2LiveManifestProvider.java
@@ -49,6 +49,12 @@ public class KubernetesV2LiveManifestProvider extends KubernetesV2AbstractManife
 
   @Override
   public KubernetesV2Manifest getManifest(String account, String location, String name) {
+    return getManifest(account, location, name, true);
+  }
+
+  @Override
+  public KubernetesV2Manifest getManifest(
+      String account, String location, String name, boolean includeEvents) {
     if (!isAccountRelevant(account)) {
       return null;
     }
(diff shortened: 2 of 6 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ManifestProvider.java
```diff
@@ -27,6 +27,17 @@
 
   T getManifest(String account, String location, String name);
 
+  T getManifest(String account, String location, String name, boolean includeEvents);
+
   List<T> getClusterAndSortAscending(
       String account, String location, String kind, String app, String cluster, Sort sort);
+
+  List<T> getClusterAndSortAscending(
```
ezimanyi: This is only ever called in one place (`getDynamicManifestFromCluster`); the only existing call to that REST endpoint is from `DynamicResolveManifestTask` in `orca` which also doesn't use the events, so I think we can safely just always omit events from this call.  If someone adds a use case that needs the events, they can add in the flag and pass it through.  (Notably this is never exposed through gate, only internal REST endpoints which we don't guarantee backwards compatibility for.)

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ManifestProvider.java
```diff
@@ -27,6 +27,17 @@
 
   T getManifest(String account, String location, String name);
 
+  T getManifest(String account, String location, String name, boolean includeEvents);
```
ezimanyi: Given that this is only called in 2 places, it might be simpler to just always require the boolean flag.  The only remaining location that would call the old version is `KubernetesV2JobProvider` which also doesn't need events so should just pass in `false`.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2LiveManifestProvider.java
```diff
@@ -84,7 +90,10 @@ public KubernetesV2Manifest getManifest(String account, String location, String
     String namespace = manifest.getNamespace();
     KubernetesKind kind = manifest.getKind();
 
-    List<KubernetesManifest> events = credentials.eventsFor(kind, namespace, parsedName.getRight());
+    List<KubernetesManifest> events =
+        !includeEvents
```
ezimanyi: This would be easier to read if we checked the positive condition and inverted to:
```
includeEvents ? credentials.eventsFor(kind, namespace, parsedName.getRight()) : Collections.emptyList()
```

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2ManifestProvider.java
```diff
@@ -127,19 +147,24 @@ public KubernetesV2Manifest getManifest(String account, String location, String
...
     String key = data.getId();
 
     List<KubernetesManifest> events =
-        cacheUtils
-            .getTransitiveRelationship(
-                kind.toString(), Collections.singletonList(key), KubernetesKind.EVENT.toString())
-            .stream()
-            .map(KubernetesCacheDataConverter::getManifest)
-            .collect(Collectors.toList());
+        !includeEvents
```
ezimanyi: Same comment as above about checking the positive condition `includeEvents` rather than `!includeEvents`

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ManifestController.java
```diff
@@ -52,15 +53,24 @@ public ManifestController(List<ManifestProvider> manifestProviders, RequestQueue
...
   @PreAuthorize("hasPermission(#account, 'ACCOUNT', 'READ')")
   @PostAuthorize("hasPermission(returnObject?.moniker?.app, 'APPLICATION', 'READ')")
   @RequestMapping(value = "/{account:.+}/{location:.+}/{name:.+}", method = RequestMethod.GET)
   Manifest getForAccountLocationAndName(
-      @PathVariable String account, @PathVariable String location, @PathVariable String name) {
+      @PathVariable String account,
+      @PathVariable String location,
+      @PathVariable String name,
+      @RequestParam(value = "includeEvents", required = false, defaultValue = "true")
+          boolean includeEvents) {
```
ezimanyi: Looks like you forget to pass this parameter to the call to `getManifest` below?

mtweten: Good catch, looks like I messed this up when rebasing some of my changes

mtweten:
Thanks for the review Eric!

Addressed your comments, I agree with simplifying the number of APIs - I was just initially a little worried and wanted to do it in a super passive way so I didn't inadvertently break anything.

Only thing I wanted to be sure of was the change in `getDynamicManifestFromCluster` - essentially we're just saying the contract for that method is that events aren't returned, and if that functionality is eventually needed it can be added in the future?

PR #3797, opened 2019-06-19
fix(titus): calling non-existant method

marchello2000:
when Titus isn't available in a region instead of giving the appropriate error messge we get:

```
No signature of method: com.netflix.spinnaker.clouddriver.deploy.DescriptionValidationErrors.rejectValue() is applicable for argument types: (String, String, String, String) values: [region, titusDeployDescription.region.not.configured, us-east-1 , ...] Possible solutions: rejectValue(java.lang.String, java.lang.String, java.lang.String), rejectValue(java.lang.String, java.lang.String, [Ljava.lang.Object;, java.lang.String), rejectValue(java.lang.String, java.lang.String)
```


emjburns:
LGTM

PR #3801, opened 2019-06-19
fix(kubernetes): update fabric HPA model in v1 provider

maggieneterval:
- [This fabric8 commit](https://github.com/fabric8io/kubernetes-client/pull/1288/files) updated the Kubernetes HPA version from v1 to v2beta1, which includes a breaking change in the CPU Utilization model detailed in the docs [here](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/): `Notice that the targetCPUUtilizationPercentage field has been replaced with an array called metrics`.


maggieneterval:
Oh man, good catch, I only searched for `withTargetCPUUtilizationPercentage`...will update!

PR #3804, opened 2019-06-20
fix(cf): allow environmentVars values to be deserializable to objects

Jammy-Louie:


Jammy-Louie:

Jammy-Louie:

PR #3806, opened 2019-06-20
fix(appengine): Properly account for GCS object version in deploy

ezimanyi:
Fixes spinnaker/spinnaker#4530

The appengine deploy stage does not account for the fact that GCS objects can have a version (such as gs://bucket/file.tar#123). Currently sending a file such as the preceding breaks as the provider doesn't recognize it as a tar file (and also includes 123 as part of the file name rather than as an object generation).

Fix this so that the version is parsed out of the file path and is properly sent to the GCS API.


ezimanyi:

PR #3808, opened 2019-06-21
fix(provider/openstack): Make the openstack provider compatible wit hthe latest openstack releases.

randomswdev:
The pull request includes the following changes:

- Upgrade to Openstack4j 3.2.0; this library allows to parse the description of some newer OpenStack objects, like the Octavia load balancers. The old library failed to deserialize their content.
- Replace the deprecated LoadBalance V2 interface of Neutron with the newer Octavia interface. The latest version of Openstack use Octavia directly, without having Neutron to intermediate the load balancer operations.
- Replace references to the deprecated ceilometer interface. The heat templates refer to Ceilometer that has been deprecated and removed; replaced with Gnocchi.

I know that the Openstack Clouddriver has been deprecated and removed from Spinnaker starting with version 1.14, but I would like to have fixed the 1.13 code, that still supports Openstack, and then evaluate if it is worth reintroduce support in 1.14.

### DIFF
--- a/clouddriver-openstack/clouddriver-openstack.gradle
+++ b/clouddriver-openstack/clouddriver-openstack.gradle
@@ -15,7 +15,7 @@
  */
 
 ext {
-  openstack4jVersion = '3.1.0'
+  openstack4jVersion = '3.2.0'
   commonsNetVersion = '3.5'
   commonsIOVersion = '1.3.2'
 }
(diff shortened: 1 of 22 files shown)


randomswdev:
Updated the commits.

emjburns:
Unfortunately, this is a feature that we won’t merge into the minor patch versions. If you’d like to do feature work on OpenStack I suggest you submit a proposal to add the provider back into the codebase. In order for that proposal to be accepted you’ll need a plan for who will maintain the code.

cc @spinnaker/toc

PR #3809, opened 2019-06-23
feat(docker): Use cache for "/dockerRegistry/images/tags" instead of requesting docker registry every time

dragon3:
## WHAT

This pull request changes `/dockerRegistry/images/tags` endpoint to use cache in Redis made by caching agent.

It provides better performance especially the image has a lot of tags. 

In order for sorting tags (`sortTagsByDate` option), this commit adds new `creationDate` attribute in the taggedImage cache.

## WHY

I've been having a bad performance issue in loading docker image tags when I want to start manual pipeline execution, like this modal:

Screen Shot 2019-06-23 at 14 45 06

I have some huge private docker registry and there are so many images and tags.
In that case, the image tag loading is like forever... 

---

I found the docker image tags listing uses the endpoint `/dockerRegistry/images/tags` in clouddriver, and I it doesn't use cache, instead it accesses docker registry every time.

That's why I created this pull request.


dragon3:
The `/dockerRegistry/images/tags` endpoint was added in https://github.com/spinnaker/clouddriver/pull/1562 without using cache.

But we can store the creation date into cache by caching agent and use it in the endpoint.
That's this pull request's change.

If there is strong reason not use cache for the endpoint, let me know, I'll close this pull request.

Thanks!

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/provider/agent/DockerRegistryImageCachingAgent.groovy
```diff
@@ -157,11 +158,17 @@ class DockerRegistryImageCachingAgent implements CachingAgent, AccountAware, Age
...
+          creationDate = credentials.client.getCreationDate(repository, tag)
+        } catch (Exception e) {
+          log.warn("Unable to fetch tag creation date, reason: {} (tag: {}, repository: {})", e.message, tag, repository)
+        }
 
         cachedTags[tagKey].with {
           attributes.name = "${repository}:${tag}".toString()
           attributes.account = accountName
           attributes.digest = digest
+          attributes.creationDate = creationDate
```
asher: Would you mind changing this to `attributes.date`? We have an alternate docker tag caching agent internally @ Netflix that store a `date` attribute used for sorting via an alternate controller, just like in this PR. It reads from ElasticSearch instead of an actual registry which is why it's currently internal-only, but we do plan on merging its functionality into `DockerRegistryImageCachingAgent` at some point, at which point naming consistency will help.

dragon3: Hi @asher 
Thank you for the comment.

Sure! I've changed it to `attributes.date`. Please have a look at it.
And good to know you have a plan to merge Netflix internal functionality in the future 👍

dragon3:
@asher 
Thank you for reviewing and merging!

1 question,
Is this change going to be available in Spinnaker 1.13 and 1.14 as well? or just 1.15?

Thanks!

asher:
@dragon3 since this isn't strictly a bug fix, I'm leaning towards not back porting it. 1.15.0 should be coming out next week though.

dragon3:
@asher 
Thank you for the reply. I got it.
I'm looking forward to 1.15.0 👍

PR #3812, opened 2019-06-24
feat(kubernetes,google): Support retrieving config files from config server.

scottfrederick:
This PR adds support for reading files from an external Spring Cloud Config Server backend (e.g. a git repository) for the purposes of keeping account secrets out of Spinnaker configuration files. See the design document for more information: https://docs.google.com/document/d/1Y7SIgobc8_CCFmFZLRUGl-76sfHWlprFYGy992S5U4A/edit#heading=h.sai17pl3aws8. 

The following fields have external configuration enabled with these changes: 
- `kubernetes` account field `kubeconfig` 
- `google` account fields `json-path` and `user-data-file`

If these fields have a value that starts with `configserver:` then the value following the prefix will be loaded as a resource from a Config Server. Without the prefix the value will be used as an absolute path to a file on the filesystem as before. 

The `google` account fields are read from the Config Server backend into memory on the clouddriver pod, and are never written to the filesystem. The `kubernetes` `kubeconfig` must be read from Config Server and stored as a temp file on the pod filesystem because the Fabric8 API used by the v1 provider and the `kubectl` commands used by the v2 provider both require a local file. This is an incremental improvement over storing kubernetes context info as a persistent file on the pod filesystem, as the temp file is more obfuscated and ephemeral, but we should investigate a way to remove this need for a temp file.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CloudConfigRefreshConfig.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CloudConfigRefreshConfig.java
@@ -17,12 +17,16 @@
 package com.netflix.spinnaker.clouddriver.cache;
 
 import com.netflix.spinnaker.cats.cluster.DefaultAgentIntervalProvider;
+import com.netflix.spinnaker.clouddriver.data.ConfigFileService;
 import com.netflix.spinnaker.clouddriver.refresh.CloudConfigRefreshScheduler;
 import org.jetbrains.annotations.NotNull;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.cloud.config.server.EnableConfigServer;
+import org.springframework.cloud.config.server.environment.EnvironmentRepository;
+import org.springframework.cloud.config.server.resource.ResourceRepository;
 import org.springframework.cloud.context.refresh.ContextRefresher;
 import org.springframework.context.annotation.*;
 import org.springframework.core.type.AnnotatedTypeMetadata;
@@ -34,7 +38,14 @@ import org.springframework.core.type.AnnotatedTypeMetadata;
 @Configuration
 @AutoConfigureAfter({RedisCacheConfig.class, DynomiteCacheConfig.class})
 @Conditional(RemoteConfigSourceConfigured.class)
+@EnableConfigServer
 public class CloudConfigRefreshConfig {
+
+  @Bean
+  ConfigFileService configFileService(
+      ResourceRepository resourceRepository, EnvironmentRepository environmentRepository) {
+    return new ConfigFileService(resourceRepository, environmentRepository);
+  }
 
   @Bean
   @ConditionalOnBean(DefaultAgentIntervalProvider.class)
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CloudConfigRefreshConfig.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CloudConfigRefreshConfig.java
@@ -37,28 +37,40 @@ import org.springframework.core.type.AnnotatedTypeMetadata;
  */
 @Configuration
 @AutoConfigureAfter({RedisCacheConfig.class, DynomiteCacheConfig.class})
-@Conditional(RemoteConfigSourceConfigured.class)
-@EnableConfigServer
 public class CloudConfigRefreshConfig {
 
-  @Bean
-  ConfigFileService configFileService(
-      ResourceRepository resourceRepository, EnvironmentRepository environmentRepository) {
-    return new ConfigFileService(resourceRepository, environmentRepository);
+  @Configuration
+  @Conditional(RemoteConfigSourceConfigured.class)
+  @EnableConfigServer
+  static class RemoteConfigSourceConfiguration {
+    @Bean
+    ConfigFileService configFileService(
+        ResourceRepository resourceRepository, EnvironmentRepository environmentRepository) {
+      return new ConfigFileService(resourceRepository, environmentRepository);
+    }
+
+    @Bean
+    @ConditionalOnBean(DefaultAgentIntervalProvider.class)
+    public CloudConfigRefreshScheduler intervalProviderConfigRefreshScheduler(
+        ContextRefresher contextRefresher, DefaultAgentIntervalProvider agentIntervalProvider) {
+      return new CloudConfigRefreshScheduler(contextRefresher, agentIntervalProvider.getInterval());
+    }
+
+    @Bean
+    @ConditionalOnMissingBean(DefaultAgentIntervalProvider.class)
+    public CloudConfigRefreshScheduler defaultIntervalConfigRefreshScheduler(
+        ContextRefresher contextRefresher) {
+      return new CloudConfigRefreshScheduler(contextRefresher, 60);
+    }
   }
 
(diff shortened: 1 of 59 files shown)


clouddriver-core/src/test/groovy/com/netflix/spinnaker/clouddriver/data/ConfigFileServiceTest.java
```diff
@@ -0,0 +1,154 @@
...
+            RuntimeException.class, () -> configFileService.getContents(CLOUD_TEST_FILE_NAME));
+    assertThat(exception.getMessage()).contains(CLOUD_TEST_FILE_NAME);
+  }
+
+  private String baseName(String fileName) {
+    return new File(fileName).getName();
+  }
+
+  private void createExpectedFile() throws IOException {
+    File file = new File(TEST_FILE_NAME);
```
ezimanyi: Would it be better to write this file into the configured temp directory instead of the current working directory?

clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/MainSpec.java
```diff
@@ -24,7 +24,12 @@
 
 @RunWith(SpringJUnit4ClassRunner.class)
 @SpringBootTest(classes = {Main.class})
-@TestPropertySource(properties = {"redis.enabled = false", "sql.enabled = false"})
+@TestPropertySource(
+    properties = {
+      "redis.enabled = false",
+      "sql.enabled = false",
+      "spring.cloud.config.server.git.uri = test"
```
ezimanyi: Is this needed to have the test pass?  In general, `MainSpec` tries to specify the minimum config needed for a successful startup, the main goal being to catch when people accidentally make a config parameter required (ex: by gating a bean behind a parameter, but autowiring it into a core class).  For that reason, I'd rather omit that config parameter here.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -116,7 +118,8 @@ public KubernetesV2Credentials(
 
     this.kubectlExecutable = managedAccount.getKubectlExecutable();
     this.kubectlRequestTimeoutSeconds = managedAccount.getKubectlRequestTimeoutSeconds();
-    this.kubeconfigFile = managedAccount.getKubeconfigFile();
+    this.kubeconfigFile =
+        configFileService.getLocalPath(managedAccount.getKubeconfigFile(), "kube", "config");
```
ezimanyi: Should this be just calling `getKubeconfigFile()` directly (which will itself either fetch the kube config or write the contents to a file)?

scottfrederick: Do you mean `KubernetesNamedAccountCredentials.CredentialFactory#getKubeconfigFile`? It would make sense for `KubernetesNamedAccountCredentials.CredentialFactory#buildV2Credentials` to pass the results of `getKubeconfigFile` into the `KubernetesV2Credentials` ctor, since the defaulting of the file location would also happen then.

ezimanyi: Yes, that is what I meant, though looking at it more closely (assuming I'm reading it correctly) it looks like the V2 provider always just directly read `kubeconfigFile` instead of going through the logic to either read `kubeconfigFile` or write `kubeconfigContents` to a file so this change correctly maintains that behavior.  (I suspect that difference between V1 and V2 is not deliberate, but that's not something we need to investigate here.)

If we do want to make V1 and V2 consistent, I agree that consolidating the logic to live in `KubernetesNamedAccountCredentials` and passing the result of `getKubeconfigFile` to both constructors makes sense.  But I think upon a closer read now that what you have is fine given that it replicates the existing behavior.

scottfrederick: It's definitely hard to figure out what's intentional and not intentional when looking at those differences without any historical context :-). I've made the recommended change and it's pretty clean, so I'm fine committing that or leaving as-is.

ezimanyi: I've you've made that change, I'd say go for it...at some point I plan to further refactor the V1/V2 to share more code so that's a step towards that (where they at least treat getting the kube config the same way).

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/ConfigFileService.java
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  private String retrieveFromLocalPath(String path) {
+    try {
+      Path filePath = Paths.get(path);
+      return new String(Files.readAllBytes(filePath));
+    } catch (FileNotFoundException e) {
+      throw new RuntimeException("File \"" + path + "\" not found or is not readable");
+    } catch (IOException e) {
+      throw new RuntimeException("Exception reading file " + path);
```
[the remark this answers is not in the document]
scottfrederick: Resolved in 2nd commit.

clouddriver-core/src/test/groovy/com/netflix/spinnaker/clouddriver/data/ConfigFileServiceTest.java
scottfrederick: Resolved in 2nd commit.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
scottfrederick: Resolved in 2nd commit.

clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/MainSpec.java
scottfrederick: I moved where the new `ConfigFileService` configuration was getting done so its conditional on a config source being defined. That cleaned this up and gets all the Config Server related config in one place.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CloudConfigRefreshConfig.java
```diff
@@ -34,8 +38,15 @@
 @Configuration
 @AutoConfigureAfter({RedisCacheConfig.class, DynomiteCacheConfig.class})
 @Conditional(RemoteConfigSourceConfigured.class)
+@EnableConfigServer
 public class CloudConfigRefreshConfig {
 
+  @Bean
+  ConfigFileService configFileService(
```
ezimanyi: Moving the `ConfigFileService` bean to this class means it won't exist unless there's a remote source, but it's a required dependency of the credentials for a few cloud providers, so these providers won't work in the absence of a remote source.

scottfrederick: That's a testing failure on my part. You have to have `kubernetes` or `google` enabled in the main spinnaker config without a config server backend to get the autowiring failure, and I missed that combo. Sorry about that. Smarter autoconfig takes care of it.

ezimanyi: Fix for this looks good, thanks!

PR #3821, opened 2019-06-26
WIP feat(kubernetes): dynamic account config

ethanfrogers:
☠️  DO NOT MERGE ☠️ 

Posting this to get initial feedback as it deals with areas of the codebase that I'm not super familiar with.

This PR aims to implement the work in the Cloud Foundry provider around dynamic configuration, particularly for accounts. Instead of using the `KubernetesNamedAccountInitializer`, `KubernetesCredentialsSynchronizer` is responsible for synchronizing the `AccountCredentialsRepository` with any added/changed/deleted accounts. This `@Bean` gets called whenever the `/refresh` endpoint is hit. Additionally, it will be called on a fixed interval. 

Some concerns I have about this refactor:
1. Credential instantiation for a large number of accounts takes a long time, even with `checkPermissionsOnStartup = false`. We've heard this from a number of users at scale. For example, I've heard reports of Clouddriver taking up to an hour to boot. With the way dynamic config works, we run the risk of that happening during account refresh as the same initialization process occurs when hitting the `/refresh` endpoint or when the periodic refresh triggers. Furthermore, calls to `/refresh` are not returned until this process finishes which will make integrating with it difficult (i.e. "how long should i wait before timeout"). Something to consider.

2. Previously, the `KubernetesV*Provider` classes were instantiated after the caching agents had been populated. In order to break the dependency loop between `CatsModule` and `KubernetesCredentialsSynchronizer` (described below) I had to instantiate those beans _before_ the synchronizer. I'm unsure if that will have any side effects or not. Additionally, there were instantiations of  `ScheduledExecutorService` in each of the `KubernetesV*ProviderConfig` classes which are unaccounted for in this patch. It'd be great to get some suggestions on how we might account for this i n the new setup.

3. In order for the `KubernetesCredentialsSynchronizer` to de-schedule any defunct caching agents it needs to depend on `CatsModule`. However, due to the way the `KubernetesV*Provider` instances were created this created a circular dependency between `CatsModule` and the synchronizer because `CatsModule` depends on `List<Provider>` and `KubernetesV*Provider` classes couldn't be created until _after_ the synchronizer had been created. In order to break this loop, I created the 2 provider beans _before_ the synchronizer. Again, I'm unsure of any consequences here. This change also turns the `KubernetesV*ProviderConfig` classes into `@Components` that we can call from the synchronizer to handle agent refreshing. 

It should be notes that Clouddriver does start and all of the logs look normal. Refresh behavior works as expected, too. As noted in point 2, i'm concerned that the agent polling isn't being setup properly. I _think_ the config refresh scheduler is taking care of that but that isn't really what we want.

### DIFF
--- a/clouddriver-kubernetes/clouddriver-kubernetes.gradle
+++ b/clouddriver-kubernetes/clouddriver-kubernetes.gradle
@@ -23,6 +23,7 @@ dependencies {
   implementation "org.springframework.boot:spring-boot-starter-web"
   implementation 'com.jayway.jsonpath:json-path:2.3.0'
   implementation "com.github.ben-manes.caffeine:guava"
+  implementation "org.springframework.cloud:spring-cloud-context"
 
   testImplementation "cglib:cglib-nodep"
   testImplementation "org.objenesis:objenesis"
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.groovy
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.groovy
@@ -20,9 +20,16 @@ import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.KubernetesSpi
 import com.netflix.spinnaker.clouddriver.security.ProviderVersion
 import com.netflix.spinnaker.fiat.model.resources.Permissions
 import groovy.transform.ToString
+import lombok.Data
+import org.springframework.beans.factory.DisposableBean
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.cloud.context.config.annotation.RefreshScope
+import org.springframework.stereotype.Component
 
+@Component
 @ToString(includeNames = true)
-class KubernetesConfigurationProperties {
+@ConfigurationProperties("kubernetes")
+class KubernetesConfigurationProperties implements DisposableBean {
   private static final Integer DEFAULT_CACHE_THREADS = 1
 
   @ToString(includeNames = true)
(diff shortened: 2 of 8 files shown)


ethanfrogers:
cc @scottfrederick

scottfrederick:
> 1. [...] With the way dynamic config works, we run the risk of that happening during account refresh as the same initialization process occurs when hitting the /refresh endpoint or when the periodic refresh triggers. [...] Something to consider.

Automatic config refresh is scheduled at the [same interval as cache refresh, or a default of 60 seconds](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CloudConfigRefreshConfig.java#L39-L51). We could consider making the config refresh a multiple of the cache refresh, and/or having a separate property to configure the config refresh interval to give users control. 

For 2 and 3, I'd suggest making these changes to clean up the bean dependencies and see how that improves the situation: 

* Make `KubernetesV1ProviderConfig` and `KubernetesV2ProviderConfig` implement `CredentialsInitializerSynchronizable` (renaming the classes accordingly)
* Register those 2 new synchronizers as `@Bean`s in  `KubernetesConfiguration` as is done with `KubernetesCredentialsSynchronizer`, removing the `@Component` from the classes
* Remove the delegation by `KubernetesCredentialsSynchronizer` to those provider synchronizers, since the refresh listener will call the three synchronizers directly

It looks to me like that would remove the dependency cycle by de-coupling Provider and Credential sync. Hopefully it will also remove the need for the `@DependsOn`, which is a smell to me that dependencies aren't right (since declaring dependencies via `@Bean` method params or bean constructor params is usually sufficient).

ethanfrogers:
@scottfrederick thanks for the suggestion on the refactoring of those classes. I'll try to make those changes and see if that changes anything.

PR #3824, opened 2019-06-27
feat(k8s): dynamic account config

ethanfrogers:
implement credential synchronization for kubernetes accounts. defines
`CredentialsInitializerSynchronizable` on the
`KubernetesV*ProviderConfig` classes. Each of these classes is
responsible for handling accounts for each provider version.


ethanfrogers:
@scottfrederick @ezimanyi submitting a new PR for feedback. It was easier to implement @scottfrederick suggestions on a wholly new branch. 

Per his feedback, I've implemented credential synchronization on the 2 provider config classes to prevent the circular dependency between the synchronizer and `CatsModule`. This seems much cleaner to me.

ethanfrogers:
Oh, one more thing I forgot to mention - each time credentials are refreshed we go through the whole check permissions process unless `checkPermissionsOnStartup` is set to `false`. As noted in the previous PR, this is a pretty lengthy process which may prove troublesome with a large number of accounts. One one hand, we _should_ do this in the event that permissions on the service account have changed.

scottfrederick:
> This seems much cleaner to me.

Me too! Aside from some cleanup (which I assume is why it's still WIP), this looks like it should work. 

It looks like you'll need `equals()` and `hashcode()` methods in `KubernetesNamedAccountCredentials` to get the detection of changed accounts to work.

ethanfrogers:
@scottfrederick I started working on that in the next commit but got distracted by something else. I'll either have it up today or next week.

ethanfrogers:
I ran into a bit of a blocker while implementing the `equals` and `hashCode` methods. In order to properly determine if the account has changed you need to compare the underlying credentials (`KubernetesV2Credentials`) which means that you need the full list of `omitKinds` and `namespaces` which, unless defined explicitly,  are collected from the Kubernetes API, which is a lengthy process today. In fact, this is one of the exact problems this feature is meant to overcome - the long startup time of Clouddriver. I think we need to address that before this really makes sense. If we don't then it will take the same amount of time it takes Clouddriver to boot.

cc @ezimanyi

ethanfrogers:
With respect to the above comment - while the initialization of credentials is a performance problem it isn't necessarily a blocker to implementing this functionality (dynamic accounts) but more of a tradeoff that can be addressed in a subsequent PR. I'll ensure that this PR is cleaned up and ready to merge before addressing that issue.

ethanfrogers:
@ezimanyi @scottfrederick I think this is ready for an official review!

scottfrederick:
See #3833 to fix the issue of temp files on refresh.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/security/KubernetesV1Credentials.java
```diff
@@ -294,4 +295,32 @@ public Boolean isRegisteredImagePullSecret(String secret, String namespace) {
     }
     return secrets.contains(secret);
   }
+
+  @Override
+  public boolean equals(Object o) {
```
ezimanyi: Similar to the comment above, `@EqualsAndHashCode` would remove some boilerplate code here.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/security/KubernetesV1Credentials.java
```diff
@@ -294,4 +295,32 @@ public Boolean isRegisteredImagePullSecret(String secret, String namespace) {
...
+    if (this == o) return true;
+    if (!(o instanceof KubernetesV1Credentials)) {
+      return false;
+    }
+    KubernetesV1Credentials that = (KubernetesV1Credentials) o;
+    return Objects.equals(namespaces, that.namespaces)
+        && Objects.equals(omitNamespaces, that.omitNamespaces)
+        && Objects.equals(dockerRegistries, that.dockerRegistries)
```
ezimanyi: These are `List<LinkedDockerRegistryConfiguration>` but `LinkedDockerRegistryConfiguration` does not override `equals` and `hashCode` so the lists will only be equal if they contain the same object references.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/security/KubernetesV1Credentials.java
```diff
@@ -294,4 +295,32 @@ public Boolean isRegisteredImagePullSecret(String secret, String namespace) {
...
+        && Objects.equals(kubeconfigFile, that.kubeconfigFile)
+        && Objects.equals(kubeconfigFileHash, that.kubeconfigFileHash);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(
+        namespaces,
+        omitNamespaces,
+        dockerRegistries,
```
ezimanyi: Same comment above about `LinkedDockerRegistryConfiguration`.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/security/KubernetesV1Credentials.java
```diff
@@ -294,4 +295,32 @@ public Boolean isRegisteredImagePullSecret(String secret, String namespace) {
...
+        imagePullSecrets,
+        configureImagePullSecrets,
+        kubeconfigFile,
+        this.kubeconfigFileHash);
```
ezimanyi: Nit: remove `this` to be consistent with other member variables.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderSynchronizable.java
```diff
@@ -0,0 +1,195 @@
...
+    this.kubernetesV2Provider = kubernetesV2Provider;
+    this.accountCredentialsRepository = accountCredentialsRepository;
+    this.kubernetesV2CachingAgentDispatcher = kubernetesV2CachingAgentDispatcher;
+    this.kubernetesResourcePropertyRegistry = kubernetesResourcePropertyRegistry;
+    this.kubernetesConfigurationProperties = kubernetesConfigurationProperties;
+    this.credentialFactory = credentialFactory;
+    this.kubernetesSpinnakerKindMap = kubernetesSpinnakerKindMap;
+    this.catsModule = catsModule;
+
+    ScheduledExecutorService poller =
```
ezimanyi: The `poller` object looks like it's not used anywhere.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderSynchronizable.java
```diff
@@ -0,0 +1,195 @@
...
+                  new KubernetesNamedAccountCredentials(
+                      managedAccount, kubernetesSpinnakerKindMap, credentialFactory);
+
+              AccountCredentials existingCredentials =
+                  accountCredentialsRepository.getOne(managedAccount.getName());
+
+              if (existingCredentials == null) {
+                // account didn't previously exist
+                newAndChangedAccounts.add(managedAccount.getName());
+              } else if (existingCredentials != null && !existingCredentials.equals(credentials)) {
```
ezimanyi: The `!= null` will always be `true`

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderSynchronizable.java
```diff
@@ -0,0 +1,195 @@
...
+            .map(AccountCredentials::getName)
+            .collect(Collectors.toList());
+
+    List<String> newNames =
+        kubernetesConfigurationProperties.getAccounts().stream()
+            .map(KubernetesConfigurationProperties.ManagedAccount::getName)
+            .collect(Collectors.toList());
+
+    return existingNames.stream()
+        .filter(name -> !newNames.contains(name))
```
ezimanyi: Similar to the comment for the V1 provider, this is `O(N^2)` in number of accounts.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderSynchronizable.java
```diff
@@ -0,0 +1,195 @@
...
+    List<String> newAndChangedAccounts = synchronizeAccountCredentials();
+
+    // we only want to initialize caching agents for new or updated accounts
+    Set<KubernetesNamedAccountCredentials> allAccounts =
+        ProviderUtils.buildThreadSafeSetOfAccounts(
+                accountCredentialsRepository,
+                KubernetesNamedAccountCredentials.class,
+                ProviderVersion.v2)
+            .stream()
+            .filter(account -> newAndChangedAccounts.contains(account.getName()))
```
ezimanyi: Would be more efficient if `newAndChangedAccounts` were a `Set`; currently is `O(M*N)` where `M` is total accounts and `N` is changed accounts.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -71,11 +66,13 @@
   @Getter private final List<String> omitNamespaces;
   private final List<KubernetesKind> kinds;
   private final Map<KubernetesKind, InvalidKindReason> omitKinds;
+  private final Map<KubernetesKind, InvalidKindReason> omitKindsComputed = new HashMap<>();
   @Getter private final List<CustomKubernetesResource> customResources;
 
   @Getter private final String kubectlExecutable;
   @Getter private final Integer kubectlRequestTimeoutSeconds;
   @Getter private final String kubeconfigFile;
+  @Getter private final String kubeconfigFileHash;
```
ezimanyi: Do we need the `@Getter` here?  I'd think this should be an implementation detail of the class.

clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesNamedAccountCredentialsInitializerSpec.groovy
```diff
@@ -1,149 +0,0 @@
...
-import com.netflix.spinnaker.clouddriver.kubernetes.security.KubernetesNamedAccountCredentialsInitializer
-import com.netflix.spinnaker.clouddriver.kubernetes.v1.security.KubernetesV1Credentials
-import com.netflix.spinnaker.clouddriver.kubernetes.v2.names.KubernetesManifestNamer
-import com.netflix.spinnaker.clouddriver.kubernetes.v2.op.job.KubectlJobExecutor
-import com.netflix.spinnaker.clouddriver.names.NamerRegistry
-import com.netflix.spinnaker.clouddriver.security.AccountCredentialsRepository
-import com.netflix.spinnaker.clouddriver.security.ProviderVersion
-import spock.lang.Specification
-
-class KubernetesNamedAccountCredentialsInitializerSpec extends Specification {
```
ezimanyi: These tests were all recently added as part of a refactor that I did on the account creation code.  While the specific class they're testing might be gone, the business logic of the tests is still valid so these tests should be moved/modified rather than just deleted.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -71,11 +66,13 @@
   @Getter private final List<String> omitNamespaces;
   private final List<KubernetesKind> kinds;
   private final Map<KubernetesKind, InvalidKindReason> omitKinds;
+  private final Map<KubernetesKind, InvalidKindReason> omitKindsComputed = new HashMap<>();
```
ezimanyi: Since we're not modifying `omitKinds` anymore and just leaving it to represent the configured omit kinds, it can probably just be a `List<String>` and the constructor can just grab the value from `managedAccount.getOmitKinds()` without any post-processing.

It's a bit confusing now when we need to check one vs the other, so I'd suggest just populating `omitKindsComputed` with `omitKinds` in the constructor and just always using `omitKindsComputed` going forward.  The original `omitKinds` would exist only for determining if there was any change.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -586,4 +588,59 @@ public void patch(
...
+      return false;
+    }
+
+    KubernetesV2Credentials that = (KubernetesV2Credentials) o;
+    return accountName.equals(that.accountName)
+        && Objects.equals(namespaces, that.namespaces)
+        && Objects.equals(omitNamespaces, that.omitNamespaces)
+        && Objects.equals(kinds, that.kinds)
+        && Objects.equals(omitKinds, that.omitKinds)
+        && Objects.equals(customResources, that.customResources)
```
ezimanyi: `KubernetesCustromResource` does not override `equals` and `hashCode`.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -586,4 +588,59 @@ public void patch(
...
+        && Objects.equals(kubectlExecutable, that.kubectlExecutable)
+        && Objects.equals(kubectlRequestTimeoutSeconds, that.kubectlRequestTimeoutSeconds)
+        && Objects.equals(kubeconfigFile, that.kubeconfigFile)
+        && Objects.equals(kubeconfigFileHash, that.kubeconfigFileHash)
+        && Objects.equals(serviceAccount, that.serviceAccount)
+        && Objects.equals(context, that.context)
+        && Objects.equals(onlySpinnakerManaged, that.onlySpinnakerManaged)
+        && Objects.equals(liveManifestCalls, that.liveManifestCalls)
+        && Objects.equals(checkPermissionsOnStartup, that.checkPermissionsOnStartup)
+        && Objects.equals(cachingPolicies, that.cachingPolicies)
```
ezimanyi: `KubernetesCachingPolicy` does not override `equals` and `hashCode`

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -586,4 +588,59 @@ public void patch(
...
+        && Objects.equals(oAuthServiceAccount, that.oAuthServiceAccount)
+        && Objects.equals(oAuthScopes, that.oAuthScopes)
+        && Objects.equals(metrics, that.metrics)
```
ezimanyi: `metrics` is not `final` and can be changed (from `true` to `false` if the metrics aren't readable).  This means that the current value may not be the configured one, similar to with `omitKinds`.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -586,4 +588,59 @@ public void patch(
       }
     }
   }
+
+  @Override
+  public boolean equals(Object o) {
```
ezimanyi: Can replace with `@EqualsAndHashCode`

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/KubernetesV1ProviderSynchronizable.groovy
```diff
@@ -0,0 +1,152 @@
...
+import com.netflix.spinnaker.clouddriver.security.ProviderVersion
+import groovy.util.logging.Slf4j
+
+import javax.annotation.PostConstruct
+import java.util.concurrent.Executors
+import java.util.concurrent.ScheduledExecutorService
+import java.util.stream.Collectors
+
+@Slf4j
+class KubernetesV1ProviderSynchronizable implements CredentialsInitializerSynchronizable {
```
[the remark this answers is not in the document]
ethanfrogers: Heh, I guess the body of this file changed enough that Git just assumed it was an entirely new file. This was previously `KubernetesV1ProviderConfig` and written in Groovy.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/KubernetesV1ProviderSynchronizable.groovy
```diff
@@ -0,0 +1,152 @@
...
+    this.kubernetesV1Provider= kubernetesV1Provider
+    this.kubernetesCloudProvider = kubernetesCloudProvider
+    this.accountCredentialsRepository = accountCredentialsRepository
+    this.kubernetesV1CachingAgentDispatcher = kubernetesV1CachingAgentDispatcher
+    this.kubernetesConfigurationProperties = kubernetesConfigurationProperties
+    this.credentialFactory = credentialFactory
+    this.kubernetesSpinnakerKindMap = kubernetesSpinnakerKindMap
+    this.catsModule = catsModule
+
+    ScheduledExecutorService poller = Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory(KubernetesV1ProviderSynchronizable.class.getSimpleName()))
```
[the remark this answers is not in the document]
ethanfrogers: That's something I was confused about. We were doing this in the previous version of this file but I could not figure out what it was actually used for.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/security/KubernetesV1Credentials.java
ethanfrogers: Ah, yes. I remember attempting to implement `equals` on `LinkedDockerRegistryConfiguration` but encountered some issues. I don't remember what they were exactly, though. Will revisit.

ethanfrogers:
@ezimanyi I've addressed most of the feedback so far. I only need to address the bit about the `metrics` flag and add tests. Thanks for the thorough review and the perf recommendations!

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -62,31 +66,49 @@
...
-  @Getter private final List<KubernetesCachingPolicy> cachingPolicies;
+  @Include @JsonIgnore @Getter private final String oAuthServiceAccount;
 
-  @JsonIgnore @Getter private final String oAuthServiceAccount;
-  @JsonIgnore @Getter private final List<String> oAuthScopes;
+  @Include @JsonIgnore @Getter private final List<String> oAuthScopes;
 
-  @Getter private boolean metrics;
-  @Getter private final boolean debug;
+  @Include @Getter private boolean metrics;
```
ezimanyi: I think we could remove the `@Getter` on the `metrics` now to make it more clear that consumers shouldn't care.  (I think it's used in one test, but that test should probably just check the computed one.)

clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/KubernetesV1ProviderSynchronizableSpec.groovy
```diff
@@ -0,0 +1,126 @@
+/*
+ * Copyright 2019 Netflix, Inc.
```
ezimanyi: s/Netflix/Armory/

PR #3825, opened 2019-06-28
feat(ecs): Add support for task definition artifact

allisaurus:
Add support for deploying an ECS Task Definition as an artifact.

* Implements https://github.com/spinnaker/spinnaker/issues/4070
* Fixes https://github.com/spinnaker/spinnaker/issues/3341
* Fixes https://github.com/spinnaker/spinnaker/issues/3340

This feature allows Task Definition attributes, including multiple containers, to be defined in an artifact (JSON file) and used in conjunction with pipeline images and other server group settings to deploy to ECS.


allisaurus:
Associated PRs:
https://github.com/spinnaker/orca/pull/3016
https://github.com/spinnaker/deck/pull/7162

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -263,6 +265,90 @@ protected RegisterTaskDefinitionRequest makeTaskDefinitionRequest(
...
+              setSpinnakerEnvVars(c.getEnvironment(), newServerGroupName);
+          c.setEnvironment(updatedEnv);
+
+          Map<String, String> updatedLabels =
+              setSpinnakerDockerLabels(c.getDockerLabels(), newServerGroupName);
+          c.setDockerLabels(updatedLabels);
+        });
+
+    if (FARGATE_LAUNCH_TYPE.equals(description.getLaunchType())) {
+      requestTemplate.setExecutionRoleArn(ecsServiceRole);
```
clareliguori: I think ideally if the role is already specified in the template, we should leave it there, and only use the Spinnaker role if it's missing.
Related: https://github.com/spinnaker/spinnaker/issues/3340

clareliguori:
Yay!  I only had a single comment on the logic.  We talked offline already about adding unit tests.

FYI for other reviewers, our goal is to get this merged for the 1.15 release (release window opens July 8).

allisaurus:
NOTE: For folks upgrading to **1.15** or otherwise using this - there is a **risk associated with downgrading** to a previous Spinnaker version after having used this change to create ECS server groups. 
(CC @maggieneterval  re: 1.15 release notes)

### Explanation
This PR includes a change in the behavior of `ContainerInformationService.java` which [allows](https://github.com/spinnaker/clouddriver/pull/3825/files#diff-23c06b7bfaad28e2198498b3811ddffbR152) for tasks with multiple containers to be described. Previous to this change, `getTaskPrivateAddress` [would fail](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/services/ContainerInformationService.java#L149) if more than one container was found in the task.

Consequently, this means that after adopting this change and creating an ECS server group with multiple containers, downgrading to an older version of Spinnaker that *does not* include this change will result in `clouddriver` returning errors when it attempts to describe clusters or server groups, because it will attempt to describe the tasks with multiple containers and throw an exception. 

### Example of error scenario
1. upgrade to Spinnaker 1.15
2. use the new ECS artifact support to create a server group that includes >1 container.
3. deploy to the server group (creates a service in your ECS account).
4. downgrade to a previous version of Spinnaker (e.g., 1.14.7)
5. open the application containing the previously-created ECS server group w/ multiple containers
6. RESULT: Spinnaker fails to describe clusters for your application, or deploy to server groups that include the SG w/ >1 container.

### Recovery
The only way to recover from this scenario is to manually delete the service with >1 container from your ECS account. OR deploy a version of `clouddriver` that includes the change to `ContainerInformationService` that permits multiple containers in a task.

maggieneterval:
Hey there @allisaurus and @clareliguori, thanks so much for the release note! I've added it to the 1.15 curated changelog doc, which I'll share with you both so you can adjust any wording/formatting as needed or add any other features you'd like to highlight for 1.15!

PR #3826, opened 2019-06-28
refactor(kubernetes): Make cache keys immutable

ezimanyi:
* refactor(kubernetes): Make cache keys immutable 

  We don't ever change cache keys after construction, but they still support (unused) setters.  Make these keys immutable so we can safely use them as keys in sets/maps.

* fix(kubernetes): Fix serialization of cache keys 

  Some cache key types don't have symmetry between serialization and deserialization logic.  Fix them so that they do so we can create a key and call toString() and get the same result as if we directly created the string key.

* refactor(kubernetes): Move key constructors into inner classes 

  The top-level Keys class supports a number of methods for creating a String representation of various key types. Move these methods into the relevant inner classes and rename them createKey.

* perf(kubernetes): Short-circuit string replacement 

  As a performance improvement, don't allocate a new string for every key part we're processing via the replaceAll(). Instead check whether they key has a ':' first which is relatively cheap compared to alwyas allocating a new string.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/Keys.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/Keys.java
@@ -225,7 +225,7 @@ public class Keys {
 
     @Override
     public String toString() {
-      return createKey(kind, type, name, version);
+      return createKey(kind, type, name, location, version);
     }
 
     @Override
@@ -289,7 +289,7 @@ public class Keys {
 
     @Override
     public String toString() {
-      return createKey(getKind(), logicalKind, account, name);
+      return createKey(getKind(), logicalKind, account, application, name);
     }
 
     @Override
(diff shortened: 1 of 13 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/Keys.java
```diff
@@ -252,6 +248,11 @@ public ApplicationCacheKey(String[] parts) {
       name = parts[3];
     }
 
+    @Override
+    public LogicalKind getLogicalKind() {
```
maggieneterval: Just curious, do we have to add a getter here because logicalKinds is static/final, or is there another reason the lombok getter isn't doing the trick?

ezimanyi: Good question---I couldn't get Lombok to do it because the getter it was creating was static, and you can't override an instance method with a static method.  So I had to write my own instance method even though it only references a static variable.

maggieneterval: :+1: thank you for explaining!

PR #3832, opened 2019-07-01
chore(sql): convert tests to use tc-mysql, unpin jooq

asher:
- Also removes H2 specific query special handling. Depends on https://github.com/spinnaker/kork/pull/330.


cats/cats-sql/src/test/groovy/com/netflix/spinnaker/cats/sql/SqlCacheSpec.groovy
```diff
@@ -38,7 +40,7 @@ class SqlCacheSpec extends WriteableCacheSpec {
     ((SqlCache) cache).merge('foo', data)
 
     then:
-    1 * ((SqlCache) cache).cacheMetrics.merge('test', 'foo', 1, 1, 0, 0, 2, 1, 0)
+    1 * ((SqlCache) cache).cacheMetrics.merge('test', 'foo', 1, 1, 0, 0, 1, 1, 0)
```
asher: H2 special handling resulted in an additional select statement, which is no longer the case.

cats/cats-test/src/main/groovy/com/netflix/spinnaker/cats/cache/CacheSpec.groovy
```diff
@@ -95,7 +95,7 @@ abstract class CacheSpec extends Specification {
         '*TEST*'                | ['blaTEST', 'blaTESTbla', 'TESTbla']
         'bla*'                  | ['blaTEST', 'blaTESTbla', 'blaPest', 'blaFEST']
         'bla[TF]EST'            | ['blaTEST', 'blaFEST']
-        'bla?EST'               | ['blaTEST', 'blaFEST']
```
asher: This test previously passed because H2 treats regex statements as case sensitive by default. MySQL defaults to case insensitive, but that's preferable. Case insensitive search was a requested feature which cat-sql delivers.

lukaseder:
Great to see that Testcontainers provided a better solution here. I'm curious though, what was the reason for having to "pin" jOOQ to an old release? Something that could have been avoided on the jOOQ side?

asher:
> Great to see that Testcontainers provided a better solution here. I'm curious though, what was the reason for having to "pin" jOOQ to an old release? Something that could have been avoided on the jOOQ side?

@lukaseder Thanks for asking! The problem was primarily around interactions with H2 after jooq added support for some non standard sql extensions with H2, like `ON DUPLICATE KEY UPDATE` which seemed to assume running H2 in MySQL mode. H2's MySQL mode seems much buggier than native however, as well as incomplete. I don't remember all the details beyond that it didn't seem to support compound primary keys correctly. Upgrading jooq and leaving H2 in its native mode caused one set of tests to fail, enabling H2's MySQL mode caused a different set to fail. We knew we wanted to go towards Testcontainers to have greater confidence in the test coverage, so pinned jooq back to punt on digging deep into the H2 weirdness.

PR #3833, opened 2019-07-01
fix(kubernetes): Use deterministic temp file names

scottfrederick:
When writing a temp kube config file from `kubeconfigContents` or from a `kubeconfigFile` that contains a config server reference like `configserver:kubeconfig`, a temp file with a randomly-generated and unique name was being generated. This causes problems when dynamic account refresh is introduced to the kubernetes provider in #3824. Every refresh will result in a new file being created with a non-deterministic name, making it difficult to determine of the `kubeconfigContents` or `kubeonfigFile` value actually changed and potentially filling the file system with temp files. 

This commit changes the behavior to create the temp file with a known name in the system temp directory, and overwrite the file contents on each refresh instead of creating a new file.


scottfrederick:
Superceded by #3850

PR #3834, opened 2019-07-01
refactor(kubernetes): A few improvements to KubernetesKind

ezimanyi:
* test(kubernetes): Add tests to KubernetesKind 

  Add some tests to KubernetesKind before doing some refactoring of that class for performance reasons.

* refactor(kubernetes): A few improvements to KubernetesKind 

  In preparation for some more subtle changes, make a few improvements to the KubernetesKind class:
  * Make the public static fields also final
  * Make it more clear what is nullable and what isn't
  * Add equals and hash code methods that allow comparing to kinds via
  .equals()

* refactor(kubernetes): Change from == to equals for KubernetesKind 

  Most comparisons of KubernetesKind use == instead of equals(), which works because we currently keep a static pool of all KubernetesKinds and never create duplicates.

  The cost of keeping this static pool is much higher than the benefit, so move to using .equals() instead of == to prepare for a future where there can be two different KubernetesKind objects that are equal.

* fix(kubernetes): Use case-insensitive comparison 

  Kinds should be considered equal if their name is equal ignoring case. This matches the current logic where getOrRegisterKind will return a kind that matches ignoring case.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKind.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKind.java
@@ -26,6 +26,9 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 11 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKind.java
```diff
@@ -131,24 +139,30 @@
...
-      String alias,
+      @Nonnull String name,
+      @Nonnull KubernetesApiGroup apiGroup,
+      @Nullable String alias,
       boolean isNamespaced,
       boolean hasClusterRelationship,
       boolean isDynamic,
       boolean isRegistered) {
     this.name = name;
+    this.lcName = name.toLowerCase();
```
maggieneterval: I probably just missed it but where is `lcName` being used?

ezimanyi: It's used above in the `@EqualsAndHashCode.Include @Nonnull private final String lcName;`  I wanted case-insensitive equality so decided to store the lower-case version of the name as its own variable and use that in the auto-generated equality check.  It's not directly exposed to consumers of the class anywhere though.

maggieneterval: Oh cool I didn't fully grok @EqualsAndHashCode.Include until I read about it just now, nice!!

PR #3836, opened 2019-07-02
perf(redis): Replace String.format with String.join

ezimanyi:
When caching large Kubernetes clusters, ~25% of allocated objects per caching cycle are from calling String.format on cache keys.

String.format is significantly more expensive than String.join as it parses the input as a regular expression. In this case, we can replace the calls to .format() with calls to .join() while keeping exactly the same functionality.


ezimanyi:
This seems like a micro-optimization but it actually makes a noticeable difference in allocations for the Kubernetes V2 provider.

maggieneterval:

PR #3838, opened 2019-07-02
perf(redis): Don't re-encode string when computing hash

ezimanyi:
When caching values in redis, we compare the hash of the value we're about to write against the value already in the cache to avoid writing unchanged values again.

The hash function we're using is expensive as it encodes the string to bytes before hashing it. Per the documentation of the putString function, it is only useful for cross-language compatibility and we should instead use putUnencodedChars for much better performance when this is not needed.

Link to header block:
https://github.com/google/guava/blob/af57f19e99110695d2b1329ae313aa5158580417/guava/src/com/google/common/hash/Hasher.java#L110

### DIFF
--- a/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cache/RedisCache.java
+++ b/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cache/RedisCache.java
@@ -15,8 +15,6 @@
  */
 package com.netflix.spinnaker.cats.redis.cache;
 
-import static java.nio.charset.StandardCharsets.UTF_8;
-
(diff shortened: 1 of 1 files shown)


ezimanyi:
@robzienert : Wanted to run this by you---this will change the hash implementation for storing keys in redis to be more efficient.  The only downside I could see is that it could (in some cases if there are non-ASCII characters) cause a one-time unenecessary write to redis as the hash value could change when this merges.  That seems to me to be a small price to pay for the performance improvement here, but wanted a second opinion here.

(The string encoding accounts for ~25% of allocated memory on Kubernetes caching cycles.)

PR #3839, opened 2019-07-02
fix(SqlCache): modify typeSanitization regex to match unpermitted chars

perek:
Fixes: https://github.com/spinnaker/spinnaker/issues/4591

Previously the type sanitization only had a limited set of matching characters, this creates bugs when encountering a cache type which included non-permitted mysql identifier characters. Change the regex to include the ASCII set of permitted chars outlined here: https://dev.mysql.com/doc/refman/5.7/en/identifiers.html

Added a test which previously failed to prove this fixes the problem.


robzienert:

PR #3845, opened 2019-07-03
perf(kubernetes): Use a more efficient kind registry

ezimanyi:
* refactor(kubernetes): Pull kind identifying information to ScopedKind 

  I created ScopedKind as a helper class in KubernetesKind but it really represents the "key" for a given Kind (ie, the unique information that determines what kind we're talking about as opposed to other attributes).

  Move the name and apiGroup information to live in ScopedKind to encapsulate this.

* perf(kubernetes): Use a more efficient kind registry 

  We're always looking up kinds by either name or alias; instead of storing the kinds as an array and looping on every request, create a HashMap mapping the name -> KubernetesKind and another mapping alias -> KubernetesKind so it's easy to look up a kind by name or alias.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKindRegistry.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKindRegistry.java
@@ -74,13 +74,13 @@ public class KubernetesKindRegistry {
     Predicate<KubernetesKind> groupMatches =
         kind -> {
           // Exact match
-          if (Objects.equals(kind.getApiGroup(), apiGroup)) {
+          if (Objects.equals(kind.getScopedKind().getApiGroup(), apiGroup)) {
             return true;
           }
 
           // If we have not specified an API group, default to finding a native kind that matches
           if (apiGroup == null || apiGroup.isNativeGroup()) {
-            return kind.getApiGroup().isNativeGroup();
+            return kind.getScopedKind().getApiGroup().isNativeGroup();
           }
 
           return false;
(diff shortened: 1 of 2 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKindRegistry.java
```diff
@@ -17,23 +17,27 @@
...
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Supplier;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import org.apache.commons.lang3.StringUtils;
 
 public class KubernetesKindRegistry {
-  private final List<KubernetesKind> values = Collections.synchronizedList(new ArrayList<>());
+  private final Map<KubernetesKind.ScopedKind, KubernetesKind> nameMap = new ConcurrentHashMap<>();
+  private final Map<KubernetesKind.ScopedKind, KubernetesKind> aliasMap = new ConcurrentHashMap<>();
```
maggieneterval: Is the reason we want a `ConcurrentHashMap` instead of a boring regular `HashMap` because (quoting from the internet) "If you have very highly concurrent application in which you may want to modify or read key value in different threads then it’s ideal to use Concurrent Hashmap."?

ezimanyi: I used a `ConcurrentHashMap` because we might be reading from multiple threads at a time (though we can only be updating one thread at a time because the update methods are `synchronized`) and might be reading from a thread while another thread is performing an update.  Using a `ConcurrentHashMap` guarantees we'll see consistent results in that case, whereas a regular `HashMap` would require manually adding thread-safety on top.

maggieneterval: :+1: :+1: :+1: thanks!

PR #3849, opened 2019-07-04
feat(cloudformation): Treat template as string instead of yaml

gregjones:
To allow for templates containing instinsic functions (e.g. !Ref)
this changes the cloudformation operation so the parameter is processed
as a string, instead of a map.

For backwards compatibility, if the received value is a map, it will
be dumped as a string before continuing.


gregjones:
cc @xavileon

xavileon:
@spinnaker/reviewers PTAL  related to https://github.com/spinnaker/orca/pull/3024

xavileon:
@ethanfrogers if you could also check out this one, it's related to https://github.com/spinnaker/orca/pull/3024 thanks!

ethanfrogers:
sorry for the delay here @xavileon! it fell off my radar.

PR #3854, opened 2019-07-08
fix(titus): adjust page size for titus queries

marchello2000:
based on feedback from titus team, adjust page size for titus job/tasks
requests to 1k


marchello2000:
@asher @robzienert : After another convo with Titus team, they say to just remove "largepage" support

PR #3859, opened 2019-07-10
fix(bootstrap): add default config path

ethanfrogers:
adds `/opt/spinnaker/config` to the default bootstrap config path.
without this, clouddriver doesn't pick up spring cloud configuration
when deployed by halyard.


ethanfrogers:
cc @scottfrederick  - i'm not sure if this is the proper way to fix this but it works locally. if there's a more springy approach LMK!

ethanfrogers:
Done in spinnaker/kork#338.

PR #3862, opened 2019-07-11
fix(appengine): Fix threading bugs in AppengineMutexRepository

ezimanyi:
I found these bugs while looking into spinnaker/spinnaker#4574.  I'm not sure if this is the root cause of those issues, but it certainly could be and is worth fixing.

* refactor(appengine): Remove unnecessary catch block 

  There's no need to catch an exception only to re-throw it in the finally block. Replace with an idiomatic try/finally.

  Also add some public/private modifiers that will make conversion to java in the next commit easier.

* refactor(appengine): Convert mutex repository to java 

* fix(appengine): Fix threading bugs in AppengineMutexRepository 

  There are a couple of threading bugs here:
  (1) We do an unsyncronized containsKey + put to generate the mutex for a given String, which can lead to two threads holding different Mutexes for the same String.
  (2) The Mutex itself has a bug whereby two threads calling lock can actually acquire different semaphores.

  Remove the Mutex class entirely and replace it with a ReentrantLock which supports the same operations. Use a ConcurrentHashMap and the atomic computeIfAbsent to store the mappings from keys to locks.


ezimanyi:

ezimanyi:

PR #3863, opened 2019-07-11
fix(kubernetes): Fix daemonset stability condition

ezimanyi:
Fixes spinnaker/spinnaker#4571.

Currently we'll consider a daemonset stable if it has the correct number of replicas, even if they are of the wrong generation. Fix this by adding a check for the generation, just as we do for replica sets.

Also, fix a bug in the replica set implementation where we are comparing Integers using == instead of using .equals.


ezimanyi:

ezimanyi:

PR #3870, opened 2019-07-12
fix(k8s): improve determineOmitKinds

ethanfrogers:
improved perf on determineOmitKinds while we consider longer term
    approaches. previously, we made a kubectl call for every kind spinnaker
    knew about and every custom resource, regardless of if the API could
    handle it or not. this approach uses 2 methods to determine readability:
    1. fetch all resources the API says it knows about via `kubectl
    api-resources`. this allows us to build a list of all known kinds.
    `determineOmitKinds` first checks against this list, preventing us from
    making list calls to kinds we already know we can't deploy.
    2. for all available kinds, use the `SelfSubjectAccessReview` API to
    determine if the account has access to the kind. We're checking against
    the `list` verb, as we were previously, but don't have to actually
    deserialize the entire list of manifests that a regular `kubectl list`
    call returns.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKind.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKind.java
@@ -36,92 +36,152 @@ public final class KubernetesKind {
 
   public static final KubernetesKind API_SERVICE =
       createAndRegisterKind(
-          "apiService", KubernetesApiGroup.APIREGISTRATION_K8S_IO, null, false, false);
+          "apiService",
+          KubernetesApiGroup.APIREGISTRATION_K8S_IO,
+          null,
+          "apiservices",
+          false,
+          false);
   public static final KubernetesKind CLUSTER_ROLE =
       createAndRegisterKind(
-          "clusterRole", KubernetesApiGroup.RBAC_AUTHORIZATION_K8S_IO, null, false, false);
+          "clusterRole",
+          KubernetesApiGroup.RBAC_AUTHORIZATION_K8S_IO,
+          null,
+          "clusterroles",
+          false,
+          false);
   public static final KubernetesKind CLUSTER_ROLE_BINDING =
       createAndRegisterKind(
-          "clusterRoleBinding", KubernetesApiGroup.RBAC_AUTHORIZATION_K8S_IO, null, false, false);
+          "clusterRoleBinding",
+          KubernetesApiGroup.RBAC_AUTHORIZATION_K8S_IO,
+          null,
+          "clusterrolebindings",
+          false,
+          false);
   public static final KubernetesKind CONFIG_MAP =
-      createAndRegisterKind("configMap", KubernetesApiGroup.CORE, "cm", true, false);
+      createAndRegisterKind("configMap", KubernetesApiGroup.CORE, "cm", "configmaps", true, false);
   public static final KubernetesKind CONTROLLER_REVISION =
-      createAndRegisterKind("controllerRevision", KubernetesApiGroup.APPS, null, true, false);
+      createAndRegisterKind(
+          "controllerRevision", KubernetesApiGroup.APPS, null, "controllerrevisions", true, false);
   public static final KubernetesKind CUSTOM_RESOURCE_DEFINITION =
       createAndRegisterKind(
-          "customResourceDefinition", KubernetesApiGroup.EXTENSIONS, "crd", false, false);
+          "customResourceDefinition",
+          KubernetesApiGroup.EXTENSIONS,
+          "crd",
+          "customresourcedefinitions",
+          false,
+          false);
   public static final KubernetesKind CRON_JOB =
-      createAndRegisterKind("cronJob", KubernetesApiGroup.BATCH, null, true, false);
+      createAndRegisterKind("cronJob", KubernetesApiGroup.BATCH, null, "cronjobs", true, false);
   public static final KubernetesKind DAEMON_SET =
-      createAndRegisterKind("daemonSet", KubernetesApiGroup.APPS, "ds", true, true);
+      createAndRegisterKind("daemonSet", KubernetesApiGroup.APPS, "ds", "daemonsets", true, true);
   public static final KubernetesKind DEPLOYMENT =
-      createAndRegisterKind("deployment", KubernetesApiGroup.APPS, "deploy", true, true);
+      createAndRegisterKind(
+          "deployment", KubernetesApiGroup.APPS, "deploy", "deployments", true, true);
   public static final KubernetesKind EVENT =
-      createAndRegisterKind("event", KubernetesApiGroup.CORE, null, true, false);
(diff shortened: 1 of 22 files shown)


ethanfrogers:
Some extra data. I took timing samples of `determineOmitKinds` before and after this change.

Before - 14s average/account
After - 5s average/account

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKind.java
```diff
@@ -134,23 +194,26 @@ private static KubernetesKind createAndRegisterKind(
...
         new KubernetesKind(
-            name, apiGroup, alias, isNamespaced, hasClusterRelationship, false, true));
+            name, apiGroup, alias, plural, isNamespaced, hasClusterRelationship, false, true));
   }
 
   private KubernetesKind(
       @Nonnull String name,
       @Nonnull KubernetesApiGroup apiGroup,
       @Nullable String alias,
+      @Nullable String plural,
```
ezimanyi: I worry about adding another field here to hold the plural version of a name; this is something else that we'll need to make sure stays in sync.  I think this will also break people who have registered CRDs in their clouddriver config, as they (presumably) have not added the plural version there so the kind will be registered with a null plural.

I think it would be more robust to use the separate `apigroup` and `kind` columns from the `api-resources` call and to compare base on `apiGroup` and `name`.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -409,6 +443,28 @@ private boolean canReadKind(KubernetesKind kind, String checkNamespace) {
     }
   }
 
+  private boolean canReadKindRbac(KubernetesKind kind, String checkNamespace) throws Exception {
+    ApiClient client = Config.fromConfig(this.kubeconfigFile);
```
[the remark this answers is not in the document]
ethanfrogers: Nice catch, I def didn't think about that. I'll see if there's any way I can incorporate that.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKind.java
ethanfrogers: Totally agree. The only reason I opted for adding the plurals is because it made parsing the result of `api-resources` easier due to it's semi-tabular nature and the fact that I can't just `-o json` the result. I'll see if there's a better way to tease the data out.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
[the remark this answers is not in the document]
ethanfrogers: @ezimanyi I dug your concerns about `kubectl` vs the API client. Based on what I've seen in the `kubectl` code [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/auth/cani.go) it doesn't look like it's doing anything different than what we were doing with `SelfSubjectAccessReview`. However, in my testing, the overhead of shelling out to `kubectl auth-can i` seems negligible vs using the API client directly.

In this chart, the yellow line is using the `kubectl` version instead of the API client:


We can still get a perf boost by using this new approach but continuing to shell out to `kubectl` for each call.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -365,13 +371,29 @@ private void determineOmitKinds() {
 
     log.info(
         "Checking permissions on configured kinds for account {}... {}", accountName, allKinds);
+    long startTime = System.nanoTime();
+
+    // compute list of kinds we explicitly know the server doesn't support
+    List<String> availableResources = jobExecutor.apiResources(this);
+    Map<KubernetesKind, InvalidKindReason> unavailableKinds =
+        allKinds
+            .parallelStream()
```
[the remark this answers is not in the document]
ethanfrogers: Converted to a `Set` and removed the `parallelStream`.

ethanfrogers:
@ezimanyi Updated per your feedback:
1. Moved to using a `Set<ScopedKind>` instead of the `parallelStream` and `List<String>`.
2. Defer to `kubectl auth can-i` vs the API client.
3. Removed `plural` and grabbed the api version and kind by parsing `kubectl api-resources`.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubectlJobExecutor.java
```diff
@@ -577,6 +578,59 @@ private String getOAuthToken(KubernetesV2Credentials credentials) {
...
+  public boolean authCanI(KubernetesV2Credentials credentials, String kind, String verb) {
+    List<String> command = kubectlAuthPrefix(credentials);
+    command.add("auth");
+    command.add("can-i");
+    command.add(verb);
+    command.add(kind);
+
+    JobResult<String> status = jobExecutor.runJob(new JobRequest(command));
+
+    if (!status.getResult().equals(JobResult.Result.SUCCESS)) {
```
ezimanyi: Minor: This could just be return `status.getResult().equals(JobResult.Result.SUCCESS)` to avoid the `if` (and same for the next function).  Also, for enums where we can't ever create new ones on the fly, it's generally safe to compare with `==` (unlike some places where we've overused this for classes that can be created on the fly).

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesApiResourceParser.java
```diff
@@ -0,0 +1,44 @@
...
+public class KubernetesApiResourceParser {
+
+  public static Set<KubernetesKind.ScopedKind> parse(String input) {
+    String[] lines = input.trim().split("\n");
+    String headerRow = lines[0];
+    int apiGroupIndex = headerRow.indexOf("APIGROUP");
+    int namespaceIndex = headerRow.indexOf("NAMESPACED");
+    int kindIndex = headerRow.indexOf("KIND");
+
+    Set<KubernetesKind.ScopedKind> kinds = new HashSet<>();
```
ezimanyi: It might be good to throw an explicit exception here if the data is not of the format we expect (ie, if we don't find those headers).  Otherwise it will probably be a reasonably hard-to-debug NPE or out-of-bounds exception that happens downstream.

PR #3873, opened 2019-07-12
chore(intellij): Add IDEA project files and upgrade the Spinnaker Gradle project

plumpy:


.idea/copyright/ALS2.xml
```diff
@@ -0,0 +1,6 @@
+<component name="CopyrightManager">
+  <copyright>
+    <option name="notice" value="Copyright &amp;#36;today.year Netflix, Inc.&#10;&#10;Licensed under the Apache License, Version 2.0 (the &quot;License&quot;);&#10;you may not use this file except in compliance with the License.&#10;You may obtain a copy of the License at&#10;&#10;  http://www.apache.org/licenses/LICENSE-2.0&#10;&#10;Unless required by applicable law or agreed to in writing, software&#10;distributed under the License is distributed on an &quot;AS IS&quot; BASIS,&#10;WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.&#10;See the License for the specific language governing permissions and&#10;limitations under the License.&#10;" />
+    <option name="myName" value="ALS2" />
```
cfieber: ASL2?

cfieber:
feels like some duplication of what this is trying (apparently unsuccessfully) to do:

https://github.com/spinnaker/spinnaker-gradle-project/blob/master/spinnaker-dev-plugin/src/main/groovy/com/netflix/spinnaker/gradle/idea/SpinnakerIdeaConfigPlugin.groovy

(that used to require generating an idea project via gradlew idea which kind of really sucks anyway. I had thought the idea plugin was supposed to interact with idea during project import but apparently not)

I'm happy to go the route of XML files in 2019 instead, if so we can delete some stuff out of spinnaker-gradle-plugin

.idea/copyright/ALS2.xml
plumpy: I have no idea why it's called that, but that's what the SpinnakerIdeaConfigPlugin called it, so I just copied that name :)

plumpy:
Yeah, the gradle idea plugin kinda sucks too because it generates old-style IDEA project files (*.ipr files instead of the new-style .idea directory). This way has its own drawbacks (to start, it means we need to check these files into every single repository), but I feel like it's better overall. Now you can just open a directory in IntelliJ without having to run './gradlew idea' first.

There's some support for this in our gradle plugin (spinnaker/spinnaker-gradle-project#76), which is why I bumped the version. I checked these files into echo a few weeks ago (spinnaker/echo#586) just to see if anyone yelled, but no one did, so I think it's time for a test on a busier project. I'll let this soak for a few more weeks, then check them into all the repositories.

PR #3874, opened 2019-07-12
feat(kubernetes): allow multi-manifest with rollout strategies

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4446

- Removes validation preventing multi-manifests to be deployed as part of a Spinnaker-coordinated rollout strategy. Still expects exactly one ReplicaSet, but now permits other resources to be deployed alongside the ReplicaSet. (Also, the designated Service must still be deployed prior to the rollout.)

Related PRs:
Deck: spinnaker/deck#7219
Orca: https://github.com/spinnaker/orca/pull/3034

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/KubernetesDeployManifestOperation.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/KubernetesDeployManifestOperation.java
@@ -262,11 +262,11 @@ public class KubernetesDeployManifestOperation implements AtomicOperation<Operat
   }
 
   private void validateManifestsForRolloutStrategies(List<KubernetesManifest> manifests) {
-    if (description.getStrategy() != null
-        && (manifests.size() != 1
-            || !manifests.get(0).getKind().equals(KubernetesKind.REPLICA_SET))) {
+    long numReplicaSets =
+        manifests.stream().filter(m -> m.getKind().equals(KubernetesKind.REPLICA_SET)).count();
+    if (description.getStrategy() != null && numReplicaSets != 1) {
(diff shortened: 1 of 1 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/KubernetesDeployManifestOperation.java
```diff
@@ -262,11 +262,13 @@ private void applyTraffic(KubernetesManifestTraffic traffic, KubernetesManifest
   }
 
   private void validateManifestsForRolloutStrategies(List<KubernetesManifest> manifests) {
-    if (description.getStrategy() != null
-        && (manifests.size() != 1
-            || !manifests.get(0).getKind().equals(KubernetesKind.REPLICA_SET))) {
+    List<KubernetesManifest> replicaSetManifests =
+        manifests.stream()
+            .filter(m -> m.getKind().equals(KubernetesKind.REPLICA_SET))
+            .collect(Collectors.toList());
```
ezimanyi: Not worth changing here (as this isn't performance-sensitive code) but this could have been `.count()` as the terminal operation to directly populate a `long numReplicaSets` if we don't actually need a list of the replicatsets.

maggieneterval: Oh that is much nicer, I changed it!

maggieneterval:
> To make sure I understand, clouddriver never cares that there are other manifests being deployed, right? That is, the orca change to pull out the single replica set is the only change we need beyond removing validation?

I think so! Orca kicks off the disable/deploy operations for the "old" manifests depending on what we determine the "new" RS to be, so we should just need to update that one place!

PR #3875, opened 2019-07-12
fix(appengine): check for null versions in servergroup caching agent

perek:
Appengine always requires that a service have at least 1 version, except for the `default` version which is not deletable. This causes an NPE when the caching agent attempts to cache a `default` service with 0 versions. Put the removeIf inside of the version check to fix this NPE


ezimanyi:

ezimanyi:

PR #3880, opened 2019-07-15
refactor(google): convert GoogleApplicationProvider to Java

plumpy:
I did this while debugging something. It serves no particular purpose
aside from the helping along the slow gradual death of Groovy.


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/view/GoogleApplicationProvider.java
```diff
@@ -0,0 +1,115 @@
...
+    GoogleApplication.View applicationView = application.getView();
+
+    Collection<String> clusters = getRelationships(cacheData, CLUSTERS);
+    clusters.forEach(
+        key -> {
+          Map<String, String> parsedKey = Keys.parse(key);
+          applicationView
+              .getClusterNames()
+              .get(parsedKey.get("account"))
+              .add(parsedKey.get("name"));
```
ezimanyi: Does the `withDefault` on the `GoogleApplication.clusterNames` work even when called from Java or do we need to handle the case where `.get(parsedKey.get("account"))` is null?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/view/GoogleApplicationProvider.java
```diff
@@ -0,0 +1,115 @@
+/*
+ * Copyright 2019 Netflix, Inc.
```
ezimanyi: Unintentional header update

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/view/GoogleApplicationProvider.java
plumpy: Yeah, it works. I wanted to convert GoogleApplication to Java too, and turn clusterNames into a SetMultimap, which makes this simpler... but I was worried I'd break the serialization.

PR #3885, opened 2019-07-17
fix(kubernetes): fix property parser exception

ethanfrogers:
when the property parser exountered a malformed it would throw an out of
bounds exception and fail. we should just skip properties with empty
values.


ethanfrogers:
cc @milesmatthias - is this the behavior you would expect for malformed properties? I can see either skipping it or defaulting it to an empty string as valid behaviors.

milesmatthias:
Yea, I would expect it to be skipped as well. Thanks for the quick fix @ethanfrogers! 👏

milesmatthias:
1 follow up here... either way you handle this scenario, there's a potential for extra work to be created. For example, I'm using the return value here as an override in a bake stage, and an empty string in a bake override is a valid value, so manifests are created.

For me, my bake step is generating job manifests, so the command being run in the job needs to exit cleanly in the use case of `cmd ""`. That's fine since we control the command the job being run, but may be a problem elsewhere.

cc @maggieneterval

ethanfrogers:

PR #3895, opened 2019-07-23
wip(events): Making `Saga` a domain model

jeyrschabu:
- adding internal events


clouddriver-saga/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/models/Saga.kt
```diff
@@ -68,6 +78,20 @@ class Saga(
   var compensating: Boolean = false
```
jeyrschabu: Wondering if we should add a default compensate in SagaEventHandler and have that one just turn the flag to true. Then have impl have to call `super.compensate`. Has to be a better way

clouddriver-saga/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/events.kt
```diff
@@ -60,6 +60,42 @@ class SagaLogAppended(
...
+  sagaId: String,
+  val oldValue: Long,
+  val newValue: Long
+) : SagaEvent(sagaName, sagaId)
+
+class SagaRequiredEventsAdded(
+  sagaName: String,
+  sagaId: String,
+  val eventNames: List<String>,
+  val currentNumOfEvents: Int
```
robzienert: I don't see any value in tracking `currentNumEvents`.

clouddriver-saga/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/events.kt
```diff
@@ -60,6 +60,42 @@ class SagaLogAppended(
...
+) : SagaEvent(sagaName, sagaId)
+
+class SagaCompleted(
+  sagaName: String,
+  sagaId: String
+) : SagaEvent(sagaName, sagaId)
```
robzienert: We'll want a `val success: Boolean`. `SagaCompleted` will be fired at the completion of the compensation lifecycle as well, in which case the success value would be `false`.

clouddriver-saga/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/models/Saga.kt
```diff
@@ -60,6 +66,10 @@ class Saga(
   var completed: Boolean = false
     set(value) {
       if (value != field) {
+        addEvent(SagaCompleted(
+          name,
+          id
+        ))
         dirty = true
```
robzienert: Get rid of `dirty` everywhere.

clouddriver-saga/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/models/Saga.kt
robzienert: What do you mean by "default compensate"? How could there be a generic default?

jeyrschabu:
Closing this guy. Destroyed my local branch, going to start fresh in another PR coming soon

PR #3898, opened 2019-07-24
feat(alibabacloud): add alibaba cloud provider's credential config

xiaozhu36:
The PR aims to add credential config for the clouddriver-alicloud. Currently, the credential only supports accessKey and secretKey. In the next future, the driver will support more credential, like assume role.


xiaozhu36:
HI . @robzienert Is there any problem on this PR?

PR #3901, opened 2019-07-25
fix(ecs): Additional null checks around cached EC2 instance data

clareliguori:
Fixes spinnaker/spinnaker#4657


clareliguori:

PR #3904, opened 2019-07-26
refactor(cats): Make the AgentDataType constructor private

plumpy:
just a tiny semi-pointless cleanup


cfieber:
the best kind of cleanup

PR #3906, opened 2019-07-26
perf(google): improve performance of /serverGroups endpoint

maggieneterval:
As the number of GCE resources increases, the `/serverGroups` request becomes unacceptably slow. Because I would like to cherry-pick this into 1.15, this is intended to be the least invasive refactor possible. Changes by class:

`GoogleApplicationProvider`
- Abstracts retrieving application cache data, and the identifiers of clusters and instances associated with an application, into reusable methods (to be used by `GoogleClusterProvider`).

`GoogleClusterProvider`
- Previously, `GoogleClusterProvider.getClusters` called `GoogleApplicationProvider.getApplication`, which retrieves the application cache data, maps it to a `GoogleApplication` object, transforms the related clusters into an account-to-cluster map, and transforms the related instances into a list of maps created by calling `Keys.parse`, which is expensive (more on that later). These transformations were unnecessary, because `getClusters` proceeded to transform the account-to-cluster map and the expensively parsed instance maps back into a list of keys of each. This change instead retrieves the raw application cache data and reads the list of related clusters and instances directly from its relationships.
- `GoogleClusterProvider.getClusters` calls `GoogleClusterProvider.clusterFromCacheData` for each cluster in the application. Previously, in each `clusterFromCacheData` call, we were calling `GoogleSecurityGroupProvider.getAllByAccount`, when in fact we should only need to fetch a given account's security groups once. This change lifts the security groups retrieval to the parent method, creating an account-to-security-groups map. We were also calling `GoogleInstanceProvider.getInstanceCacheData` on a list of every application instance for every cluster in the application, so this change also lifts that logic to the parent method and creates a second signature for `clusterFromCacheData` to preserve backwards compatibility with other consumers.

Next steps:
- Based on profiling a Clouddriver pod of a GCE installation with many resources, a non-trivial source of allocations is `Pattern.compile` calls in frigga utils, which is called by `Keys.parse`. I will investigate if any of the patterns can be made static.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/view/GoogleApplicationProvider.java
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/view/GoogleApplicationProvider.java
@@ -24,7 +24,6 @@ import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.collect.ImmutableSet;
 import com.netflix.spinnaker.cats.cache.Cache;
 import com.netflix.spinnaker.cats.cache.CacheData;
 import com.netflix.spinnaker.cats.cache.RelationshipCacheFilter;
@@ -35,6 +34,8 @@ import com.netflix.spinnaker.clouddriver.google.model.GoogleApplication;
 import com.netflix.spinnaker.clouddriver.model.Application;
 import com.netflix.spinnaker.clouddriver.model.ApplicationProvider;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -66,14 +67,25 @@ final class GoogleApplicationProvider implements ApplicationProvider {
     return data.stream().map(this::applicationFromCacheData).collect(toSet());
   }
 
+  public CacheData getApplicationCacheData(String name) {
+    return cacheView.get(
+        APPLICATIONS.getNs(),
+        Keys.getApplicationKey(name),
+        RelationshipCacheFilter.include(CLUSTERS.getNs(), INSTANCES.getNs()));
+  }
+
+  public Set<String> getApplicationClusterIdentifiers(CacheData applicationCacheData) {
+    return getRelationships(applicationCacheData, CLUSTERS);
+  }
+
+  public Set<String> getApplicationInstanceIdentifiers(CacheData applicationCacheData) {
+    return getRelationships(applicationCacheData, INSTANCES);
+  }
+
   @Override
   public Application getApplication(String name) {
+    CacheData cacheData = getApplicationCacheData(name);
 
-    CacheData cacheData =
-        cacheView.get(
-            APPLICATIONS.getNs(),
-            Keys.getApplicationKey(name),
-            RelationshipCacheFilter.include(CLUSTERS.getNs(), INSTANCES.getNs()));
     if (cacheData == null) {
       return null;
     }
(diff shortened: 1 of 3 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/view/GoogleClusterProvider.groovy
```diff
@@ -77,26 +78,37 @@ class GoogleClusterProvider implements ClusterProvider<GoogleCluster.View> {
...
+    Collection<CacheData> clusterCacheData = cacheView.getAll(
+      CLUSTERS.ns,
+      applicationClusterIdentifiers,
+      RelationshipCacheFilter.include(SERVER_GROUPS.ns)
+    )
+
+    Set<String> applicationInstanceIdentifiers = includeInstanceDetails ?
+      applicationProvider.getApplicationInstanceIdentifiers(applicationCacheData) :
+      Collections.<String>emptySet()
+    Collection<CacheData> applicationInstanceCacheData = instanceProvider.getInstanceCacheData(applicationInstanceIdentifiers)
```
plumpy: supernit: I kept getting `applicationCacheData` and `applicationInstanceCacheData` mixed up. IMO `instanceCacheData` is just as clear but more distinct. Feel free to ignore if you disagree :grinning:

maggieneterval:
Update on next steps: Opened this Frigga PR: https://github.com/Netflix/frigga/pull/26

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/view/GoogleApplicationProvider.java
```diff
@@ -66,14 +67,25 @@
     return data.stream().map(this::applicationFromCacheData).collect(toSet());
   }
 
+  public CacheData getApplicationCacheData(String name) {
```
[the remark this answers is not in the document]
maggieneterval: GREAT idea, tis done, thank you!

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/view/GoogleClusterProvider.groovy
maggieneterval: :+1: :+1: updated!

maggieneterval:
@plumpy thanks for the code review, if this is ready to merge I hope I can be rewarded with understanding what :dog2: :sweet_potato: means? :pray: :pray: :pray:

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/view/GoogleApplicationProvider.java
```diff
@@ -66,32 +68,50 @@
...
             Keys.getApplicationKey(name),
             RelationshipCacheFilter.include(CLUSTERS.getNs(), INSTANCES.getNs()));
-    if (cacheData == null) {
-      return null;
-    }
+    return getApplicationCacheData(cacheData);
+  }
 
-    return applicationFromCacheData(cacheData);
+  ApplicationCacheData getApplicationCacheData(CacheData cacheData) {
```
plumpy: I think this can be private? go/java-practices/visibility

plumpy:
:dog2: is because you like dogs and I also like dogs so why not make this review better by adding a dog
:sweet_potato: is because I wanted to say "this review is sweet!" and when I typed `:sweet`, a sweet potato autocompleted.

I'm not really as advanced with you with emojis. (yet?)

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/view/GoogleApplicationProvider.java
maggieneterval: Thanks Plumpy good call :pray: :pray:

PR #3910, opened 2019-07-31
Add CircleCI config to (hopefully) reduce build times. 

eddiewebb:
This PR is based on a slack convo with @robzienert that expressed interest in exploring other CI systems to reduce the build times of Spinnaker components.  CloudDriver was suggested as a good test of longer/complex builds.

The change introduces a single .circleci/config.yml file that uses install/buildWithTravis.sh scripts as guidance.

To see the actual builds, please see https://circleci.com/gh/eddiewebb/clouddriver/11

I am happy to help an admin setup the CircleCI project for the Spinnaker org to explore a full build & publish to bintray build (I don't have acess to a bintray repo)


.circleci/config.yml
```diff
@@ -0,0 +1,86 @@
...
+                $GRADLE -Prelease.travisci=true -Prelease.useLastTag=true -PbintrayUser="${bintrayUser}" -PbintrayKey="${bintrayKey}" -x test candidate --stacktrace
+                ;;
+              *)
+                $GRADLE -Prelease.travisci=true -Prelease.useLastTag=true -PbintrayUser="${bintrayUser}" -PbintrayKey="${bintrayKey}" -x test final --stacktrace
+                ;;
+              esac
+            else
+              echo -e 'WARN: Should not be here => Branch ['$CIRCLE_BRANCH']  Tag ['$CIRCLE_TAG']  Pull Request ['$CIRCLE_PULL_REQUEST']'
+              $GRADLE build
+            fi
```
robzienert: Let's just no-op the build process if it's not a PR. That way we can merge this and run side-by-side for PRs, but keep our current publishing flow on Travis, rather than having dualing CI's trying to publish to Bintray.

eddiewebb: i just removed all but assemble and build. it will still run on master with current config, but only building.

eddiewebb: Circle has a +build only pr'setting that will take care of the rest

robzienert:
Looks like this build needs to get docker into the mix - can this be updated for that? 

https://circleci.com/gh/eddiewebb/clouddriver/20

robzienert:
Hmm, one last change, I hope... we use testcontainers and according to the docs it looks like `machine` is required?

eddiewebb:
I don't know testcontainers, but if the requirement is just docker daemon, then current use of remote docker should work - https://github.com/spinnaker/clouddriver/pull/3910/files#diff-1d37e48f9ceff6d8030570cd36286a61R19

I committed a machine build, but this move away from a docker container might create the need to ensure a JDK version is consistent and deterministic across builds.

PR #3917, opened 2019-08-05
fix(kubernetes/v1): re-add currentCpuUtilization

mrampton:
@maggieneterval -- just realized I stripped out currentCpuUtilization from the class and forgot to add it back with my final commit on Friday.

Embarrassing -- but here's the fix!


maggieneterval:

PR #3919, opened 2019-08-05
fix(provider/k8s): fix v2 run job cancel

ethanfrogers:
implements the destroy operation for v2 jobs so that canceling the run
job stage actually works. adds an operation and converter. also adding
`region` to the job status so that the destroy op can be properly
created in orca.


ethanfrogers:
Related to spinnaker/orca#3082

ethanfrogers:
I'm closing this PR as the user who requested it doesn't want this functionality. The expected result of canceling a job is really "pausing" a job (resources stays around in exited state, logs are still viewable) but Kubernetes doesn't support this.

emjburns:
@ethanfrogers why is that the expected result of canceling a job? My expectation would be that if you cancel a job everything is removed. For Titus, the logs become archived and you can still view them but in a different way than if something is running.

ethanfrogers:
@emjburns I think that is the expectation, as well. The experience is similar for Kubernetes as well but the user who requested this feature doesn't have their logging infra in place so this would be a breaking change for them. Since they're the only person who has reported this, I figured we'd keep the old behavior in place for the time being.

emjburns:
Ah okay. Thanks!

PR #3921, opened 2019-08-06
fix(google): return null guards to GoogleApplicationProvider

maggieneterval:
My bad! :see_no_evil: :sob: 
`GoogleApplicationProvider.getApplication` can be called with non-GCE application names, so we cannot assume 1) That we can build an instance of `ApplicationCacheData` (should default to null) or 2) That we will have a non-null instance of `ApplicationCacheData` downstream.


maggieneterval:
Thank you, it has been nullabled!

PR #3922, opened 2019-08-06
feat(event): Adding eventing lib

robzienert:
This is PR 1 of N for some new idempotency functionality in River. The implementation will be confined to the Titus provider only while we iterate ergonomics of the various components.

The idempotency functionality implements a Saga pattern that is backed by an immutable event log. While not strictly required for the design pattern, backing operations by an eventing system will make it more palatable for debugging and auditing functionality later on, as well as new interaction possibilities between clouddriver & orca in the future.

Current implementation targets an in-memory store only. Future contributions will add a SQL backend so that operations can be retried by Orca in the face of downstream cloud provider failures.

It is possible that this functionality could be lifted into kork at some point.

### DIFF
--- a/clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/Aggregate.kt
+++ b/clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/Aggregate.kt
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.clouddriver.event
+
+/**
+ * The identifiable collection of an event log.
+ *
+ * Aggregates are grouped by a [type] which should be unique for each domain entity, with unique
+ * [id] values therein. A [version] field is used to ensure business logic is operating on the
+ * latest event state; any modification to an [Aggregate] event log will increment this value.
+ * When an operation is attempted on an [version] which is not head, the event framework will
+ * reject the change.
+ */
+class Aggregate(
+  val type: String,
+  val id: String,
+  var version: Long
+) {
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (javaClass != other?.javaClass) return false
+
+    other as Aggregate
+
+    if (type != other.type) return false
+    if (id != other.id) return false
+
+    return true
+  }
+
+  override fun hashCode(): Int {
+    var result = type.hashCode()
+    result = 31 * result + id.hashCode()
+    return result
+  }
+}
--- a/clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/EventMetadata.kt
+++ b/clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/EventMetadata.kt
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.clouddriver.event
+
+import java.time.Instant
+
+/**
+ * Metadata for a [SpinnakerEvent].
+ *
+ * @param sequence Auto-incrementing number for event ordering
+ * @param originatingVersion The aggregate version that originated this event
+ * @param timestamp The time at which the event was created
+ * @param serviceVersion The version of the service (clouddriver) that created the event
+ * @param provenance Where the event was generated and by what
+ */
+data class EventMetadata(
+  val sequence: Long,
+  val originatingVersion: Long,
+  val timestamp: Instant = Instant.now(),
+  val serviceVersion: String = "unknown",
+  val provenance: String = "unknown"
+)
--- a/clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/SpinnakerEvent.kt
+++ b/clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/SpinnakerEvent.kt
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.clouddriver.event
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo
+import java.util.UUID
+
+/**
+ * The base event class for the event sourcing library.
+ *
+ * @param aggregateType The type of aggregate the event is for
+ * @param aggregateId The id of the aggregate the event is for
+ * @property id A unique ID for the event. This value is for tracing, rather than loading events
+ * @property metadata Associated metadata about the event; not actually part of the "event proper"
+ */
+@JsonTypeInfo(
+  use = JsonTypeInfo.Id.NAME,
+  include = JsonTypeInfo.As.PROPERTY,
+  property = "spinEventType"
+)
+abstract class SpinnakerEvent(
+  val aggregateType: String,
+  val aggregateId: String
+) {
+  val id = UUID.randomUUID().toString()
+
+  lateinit var metadata: EventMetadata
+}
--- a/clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/config/MemoryEventRepositoryConfig.kt
+++ b/clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/config/MemoryEventRepositoryConfig.kt
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.clouddriver.event.config
+
+import com.netflix.spectator.api.Registry
+import com.netflix.spinnaker.clouddriver.event.persistence.EventRepository
+import com.netflix.spinnaker.clouddriver.event.persistence.InMemoryEventRepository
+import org.slf4j.LoggerFactory
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.boot.context.properties.EnableConfigurationProperties
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.validation.annotation.Validated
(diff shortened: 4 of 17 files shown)


clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/config/MemoryEventRepositoryConfig.kt
```diff
@@ -0,0 +1,100 @@
...
+      }
+      if (value.maxAggregateAgeMs == null && value.maxAggregatesCount == null) {
+        context.buildConstraintViolationWithTemplate("One of 'maxAggregateAgeMs' and 'maxAggregatesCount' must be set")
+          .addConstraintViolation()
+        return false
+      }
+      return true
+    }
+  }
+}
```
robzienert: Here's me, making an attempt at some better patterns for handling configuration within Spinnaker.

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/persistence/MemoryEventRepository.kt
```diff
@@ -0,0 +1,168 @@
...
+
+  override fun save(aggregateType: String, aggregateId: String, originatingVersion: Long, events: List<SpinEvent>) {
+    registry.counter(aggregateWriteCountId).increment()
+
+    val aggregate = getAggregate(aggregateType, aggregateId)
+
+    if (aggregate.version != originatingVersion) {
+      // If this is being thrown, ensure that the originating process is retried on the latest aggregate version
+      // by re-reading the events list.
+      throw AggregateChangeRejectedException(
```
ajordens: The expected retry behavior for this is something upstream correct and up to the caller right?

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/persistence/MemoryEventRepository.kt
```diff
@@ -0,0 +1,168 @@
...
+        sequence = currentSequence + (index + 1),
+        originatingVersion = originatingVersion
+      )
+    }
+
+    registry.counter(eventWriteCountId).increment(events.size.toLong())
+    this.events[aggregate]!!.addAll(events)
+    aggregate.version = aggregate.version + 1
+
+    events.joinToString(",") { it.javaClass.simpleName }
```
ajordens: Looks unnecessary, it's already in the log below explicitly.

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/SpinEvent.kt
```diff
@@ -0,0 +1,41 @@
...
+ * @param aggregateId The id of the aggregate the event is for
+ * @property id A unique ID for the event. This value is for tracing, rather than loading events
+ * @property metadata Associated metadata about the event; not actually part of the "event proper"
+ */
+@JsonTypeInfo(
+  use = JsonTypeInfo.Id.NAME,
+  include = JsonTypeInfo.As.PROPERTY,
+  property = "spinEventType"
+)
+abstract class SpinEvent(
```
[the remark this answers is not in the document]
marchello2000: I prefer `SpinnakerEvent`

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/config/MemoryEventRepositoryConfig.kt
```diff
@@ -0,0 +1,100 @@
...
+import org.springframework.validation.annotation.Validated
+import java.time.Duration
+import javax.validation.Constraint
+import javax.validation.ConstraintValidator
+import javax.validation.ConstraintValidatorContext
+import javax.validation.constraints.Min
+import kotlin.reflect.KClass
+
+@Configuration
+@ConditionalOnProperty("spinnaker.clouddriver.eventing.memory-repository.enabled", matchIfMissing = true)
```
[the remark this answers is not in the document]
robzienert: Ah, yes! Thanks for calling that out. I've been dabbling with trying to come up with a standard config pattern. Extracted this gist from some private gist I cooked up a few months ago while I was out of commission for some color: https://gist.github.com/robzienert/68dde8c895e93c0a391953eae43fcf0f

I was planning on submitting an RFC for this but never got around to it. Happy to debate this, I'm dropping the gist here mostly to get wheels turning.

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/persistence/MemoryEventRepository.kt
robzienert: That's correct. Upstream could be code interacting with the repository, or bubbled all the way out of the service boundary.

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/persistence/MemoryEventRepository.kt
```diff
@@ -0,0 +1,168 @@
...
+
+    if (aggregate.version != originatingVersion) {
+      // If this is being thrown, ensure that the originating process is retried on the latest aggregate version
+      // by re-reading the events list.
+      throw AggregateChangeRejectedException(
+        "Attempting to save events against an old aggregate version " +
+        "(version: ${aggregate.version}, originatingVersion: $originatingVersion)")
+    }
+
+    val currentSequence = this.events[aggregate]!!.map { it.metadata.sequence }.max() ?: 0
```
marchello2000: unless i am missing something, i don't think the sequence calculation is thread-safe

marchello2000: actually also the sequence assignment on L69 is not safe either unless this whole method is in a critical section (or equivalent)

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/persistence/MemoryEventRepository.kt
```diff
@@ -0,0 +1,168 @@
...
+  private val aggregateCountId = registry.createId("eventing.aggregates")
+  private val aggregateWriteCountId = registry.createId("eventing.aggregates.writes")
+  private val aggregateReadCountId = registry.createId("eventing.aggregates.reads")
+  private val eventCountId = registry.createId("eventing.events")
+  private val eventWriteCountId = registry.createId("eventing.events.writes")
+  private val eventReadCountId = registry.createId("eventing.events.reads")
+
+  private val events: MutableMap<Aggregate, MutableList<SpinEvent>> = ConcurrentHashMap()
+
+  override fun save(aggregateType: String, aggregateId: String, originatingVersion: Long, events: List<SpinEvent>) {
```
marchello2000: nit: having `events` and `MemoryEventRepository.events` makes it harder to grok which is being operated on while reading this function. Maybe `newEvents` for the local var?

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/persistence/MemoryEventRepository.kt
```diff
@@ -0,0 +1,168 @@
...
+
+    registry.counter(eventWriteCountId).increment(events.size.toLong())
+    this.events[aggregate]!!.addAll(events)
+    aggregate.version = aggregate.version + 1
+
+    events.joinToString(",") { it.javaClass.simpleName }
+    log.debug("Saved $aggregateType/$aggregateId@${aggregate.version}: " +
+      "[${events.joinToString(",") { it.javaClass.simpleName }}]")
+
+    events.forEach { eventPublisher.publish(it) }
```
marchello2000: given that the `eventPublisher` can be (as is now) synchronous, I think it's a slightly weird/unexpected for the caller of this method to have to wait on all publishers having to process the events being saved. I think the caller would expect the events to be saved durably but not necessarily broadcast on the same blocking operation

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/SynchronousEventPublisher.kt
```diff
@@ -0,0 +1,47 @@
...
+
+  private lateinit var context: ApplicationContext
+
+  override fun publish(event: SpinEvent) {
+    context.getBeansOfType(EventListener::class.java).values.toList()
+      .also {
+        log.trace("Publishing event: $event to ${it.joinToString(",")}")
+      }.forEach {
+        try {
+          it.onEvent(event)
```
marchello2000: Always scares me when things like this are synchronous - one bad actor/EventListener and everyone gets hurt

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/EventMetadata.kt
```diff
@@ -0,0 +1,35 @@
...
+ * @param timestamp The time at which the event was created
+ * @param serviceVersion The version of the service (clouddriver) that created the event
+ * @param provenance Where the event was generated and by what
+ */
+data class EventMetadata(
+  val sequence: Long,
+  val originatingVersion: Long,
+  val timestamp: Instant = Instant.now(),
+  val serviceVersion: String = "unknown",
+  val provenance: String = "unknown"
```
marchello2000: this might be just me, but i hate the word `provenance`. it always takes me a few seconds to think about what it means. I am going to go on a limb and say that, esp for developers whose primary language is not English, will find this work similarly difficult to parse.
I think `source` or `origin` convey the same information and are much more widely used

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/SynchronousEventPublisher.kt
robzienert: That's what the try/catch is for - it negates the exact scenario you're talking about.

marchello2000: what about one actor taking 10 minutes to process the event?

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/persistence/MemoryEventRepository.kt
robzienert: My assumption was that an aggregate would be written to by a single thread. Probably a poor assumption, so I'll update.

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/persistence/MemoryEventRepository.kt
robzienert: This is a result of the current Saga framework design. You raise a good point, so I'm working on refactoring the Saga lib to separate the flow propagation from being event-driven to just informed by event data. This will let me remove this specific call and allow event publishing to be handled entirely asynchronously. This will also have the side effect of making `SynchronousEventPublisher` basically useless (and therefore, I'll re-evaluate the default `EventPublisher` implementation).

Thanks!

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/persistence/MemoryEventRepository.kt
marchello2000: ah, yeah, that makes sense - didn't realize it was meant that only 1 thread per aggregate

robzienert:
Cleaned a bunch of stuff up, including getting rid of the `EventPublisher` in favor of the Spring `ApplicationEventPublisher`, which is async. The Saga lib I've refactored to not be driven off synchronous retrieval of events - how it should've been the whole time. 😬

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/Aggregate.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    if (type != other.type) return false
+    if (id != other.id) return false
+
+    return true
+  }
+
+  override fun hashCode(): Int {
+    var result = type.hashCode()
+    result = 31 * result + id.hashCode()
```
marchello2000: do you need mod `Int.MAX_VALUE`?

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/SpinnakerEvent.kt
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @param aggregateType The type of aggregate the event is for
+ * @param aggregateId The id of the aggregate the event is for
+ * @property id A unique ID for the event. This value is for tracing, rather than loading events
+ * @property metadata Associated metadata about the event; not actually part of the "event proper"
+ */
+@JsonTypeInfo(
+  use = JsonTypeInfo.Id.NAME,
+  include = JsonTypeInfo.As.PROPERTY,
+  property = "spinEventType"
```
marchello2000: where is this prop defined?

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/persistence/InMemoryEventRepository.kt
```diff
@@ -0,0 +1,165 @@
...
+          .sortedBy { it.second.metadata.timestamp }
+          .subList(0, max(events.size - maxCount, 0))
+          .forEach {
+            log.trace("Cleaning up ${it.first}")
+            events.remove(it.first)
+          }
+      }
+  }
+
+  @Scheduled(fixedRate = 1_000)
```
marchello2000: every second seems a bit aggressive, no? not sure what the actual usage is (how many events are coming /going).
do you think it's worth making this recurrence interval configurable?

marchello2000:
LGTM

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/Aggregate.kt
robzienert: This is autogenerated.

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/SpinnakerEvent.kt
robzienert: In the JSON payload.

PR #3924, opened 2019-08-08
feat(provider/k8s): support artifacts in v2 runjob

ethanfrogers:
adds support for artifacts in the v2 run job implementation


ethanfrogers:
Related to - https://github.com/spinnaker/spinnaker/issues/4605

PR #3927, opened 2019-08-09
feat(provider/eureka): retry/http configuration for eureka caching agent

cfieber:


clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/provider/agent/EurekaCachingAgent.groovy
```diff
@@ -92,7 +93,7 @@ class EurekaCachingAgent implements CachingAgent, HealthProvidingCachingAgent, C
   @Override
   CacheResult loadData(ProviderCache providerCache) {
     log.info("Describing items in ${agentType}")
-    EurekaApplications disco = eurekaApi.loadEurekaApplications()
+    EurekaApplications disco = retry.retry({ eurekaApi.loadEurekaApplications() }, 3, 100, false)
```
cfieber: This is more to handle hitting a bad node than eureka being completely offline (hence the short retry interval)

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/config/EurekaProviderConfiguration.groovy
```diff
@@ -38,12 +53,56 @@ import java.util.regex.Pattern
...
+  @Autowired
+  ConfigurableEnvironment environment
+
   @Bean
   @ConfigurationProperties("eureka.provider")
   EurekaAccountConfigurationProperties eurekaConfigurationProperties() {
     new EurekaAccountConfigurationProperties()
   }
 
+  private OkHttpClientConfigurationProperties eurekaClientConfig() {
```
cfieber: This feels wrong-but-right - I don't want to publish a different OkHttpClientConfigurationProperties bean into the context so this is duplicating the functionality of `@ConfigurationProperties`

PR #3931, opened 2019-08-12
fix(provider/k8s): return operation result data

ethanfrogers:
return the operation result data alongside the data required by the run
job task. the operation result data will be used when refreshing the
cached manifests in downstream refresh tasks


ethanfrogers:

ethanfrogers:

PR #3932, opened 2019-08-12
chore(dependencies): Get the Oracle API jar from Maven.

plumpy:
Thanks to Clare's [careful analysis](https://github.com/spinnaker/spinnaker/issues/4740), I think it should be safe to merge this now. Sounds like the problem we were having before was some sort of negative interaction between the AWS client and a specific version of httpclient, both of which have now been upgraded to versions that aren't affected.

Yay! :clinking_glasses: 

Fixes spinnaker/spinnaker#4731
Fixes spinnaker/spinnaker#4740


cfieber:
This is exciting.

We should add [this bom](https://search.maven.org/artifact/com.oracle.oci.sdk/oci-java-sdk-bom/1.5.17/pom) to the spinnaker-dependencies module in kork and get rid of the explicit version numbers in this PR.

clanesf:
I just ran into a bunch of these one of our 1.14.X systems to 1.15.2 using Clouddriver build
`gcr.io/spinnaker-marketplace/clouddriver:6.1.0-20190812153835`. Do we need to cherry-pick this back to 1.15 so we don't run into these during upgrades?

plumpy:
Unforunately, this isn't really a cherry-pickable fix. It changes a few dozen dependency jars and would reintroduce spinnaker/spinnaker#4484 unless we also cherry-pick spinnaker/kork#336, which likely introduces even more dependency changes, etc. The surface area is just way too large to be safe.

Aside from the warnings in the log, are you experiencing any actual issues because of it?

danopia:
The linked SDK issue mentions S3 signature mismatches as the main symptom, if you use non-normalized paths as s3 keys. So someone who uses s3 artifacts with interesting paths would be impacted by this issue.

I do not use s3 from clouddriver and haven't seen any issues in almost a week of seeing this warning. I get over 10 million WARNs a day, though 😄 

Screen Shot 2019-08-27 at 9 21 28 AM

PR #3936, opened 2019-08-13
feat(titus): Refactor deploy operation to use sagas

robzienert:
Refactors a bunch of the Titus cloud provider (specifically the deploy operation):

1. Switched from Groovy to Java for affected areas.
1. Deploy operation is now backed by the saga library.

A couple of additional things:

1. Widespread access into thread local `Task` is no longer a thing. The Saga domain object can handle `log` statements, which was its primary use anyway.
1. A few of the `SagaAction`s emit out additional events, which could eventually play into an event stream (`TitusJobSubmitted`, `TitusLoadBalancersApplied`, `TitusScalingPoliciesCopied`, ...)
1. The `LoadFront50App` `SagaAction` is designed to be composable into other, future Saga workflows. I'd like to move Clouddriver in more of a direction where we have common building blocks within a cloud provider, or in this case, across cloud providers.

Followup after this: I want to add a bunch of tests for `clouddriver-titus` now that the deploy operation has been broken down into smaller pieces.


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/AbstractTitusDeployAction.java
```diff
@@ -0,0 +1,94 @@
...
+
+      TitusUtils.assertTitusAccountCredentialsType(sourceCredentials);
+
+      return titusClientProvider.getTitusClient(
+          (NetflixTitusCredentials) sourceCredentials, source.getRegion());
+    }
+    return null;
+  }
+
+  static class InsufficientDeploySourceStateException extends TitusException {
```
emjburns: Should the exceptions be in another class? This strikes me as not how we usually structure things.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/AttachTitusServiceLoadBalancers.java
```diff
@@ -0,0 +1,107 @@
...
+import javax.annotation.Nullable;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import org.jetbrains.annotations.NotNull;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class AttachTitusServiceLoadBalancers
+    implements SagaAction<AttachTitusServiceLoadBalancers.AttachTitusServiceLoadBalancersCommand> {
```
emjburns: Add an import for `AttachTitusServiceLoadBalancersCommand` so that this is nicer to read?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/AttachTitusServiceLoadBalancers.java
```diff
@@ -0,0 +1,107 @@
...
+
+      saga.log("Load balancers applied");
+    }
+
+    return new Result();
+  }
+
+  @EqualsAndHashCode(callSuper = true)
+  @Getter
+  public static class AttachTitusServiceLoadBalancersCommand extends SagaCommand {
```
emjburns: Why is this a static class inside this other class? Trying to wrap my brain around this structure.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/CopyTitusServiceScalingPolicies.java
```diff
@@ -0,0 +1,182 @@
...
+import lombok.Getter;
+import org.jetbrains.annotations.NotNull;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class CopyTitusServiceScalingPolicies extends AbstractTitusDeployAction
+    implements SagaAction<CopyTitusServiceScalingPolicies.CopyTitusServiceScalingPoliciesCommand> {
+
+  private static final List<ScalingPolicyStatus.ScalingPolicyState> IGNORED_STATES =
```
emjburns: Imports for all these guys?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/PrepareTitusDeploy.java
```diff
@@ -0,0 +1,484 @@
...
+      this.description = description;
+    }
+
+    @Override
+    public void setFront50App(@Nonnull LoadFront50App.Front50App app) {
+      this.front50App = app;
+    }
+  }
+
+  private static class SecurityGroupNotFoundException extends TitusException {
```
emjburns: Feels like there could be a better way to share common exceptions for things that most cloud providers would need to use.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/SubmitTitusJob.java
```diff
@@ -0,0 +1,191 @@
...
+
+    if (jobUri == null) {
+      throw new TitusException("could not create job");
+    }
+
+    saga.log("Successfully submitted job request to Titus (Job URI: %s)", jobUri);
+
+    return new Result(
+        new ManyCommands(
+            new AttachTitusServiceLoadBalancersCommand(
```
asher: Even though its a noop, can we avoid running this command entirely for batch job operations?

robzienert: Yes, we do via the condition in the `SagaFlow`:

```java
SagaFlow flow = new SagaFlow()
        .then(LoadFront50App.class)
        .then(PrepareTitusDeploy.class)
        .then(SubmitTitusJob.class)
        .on(
          TitusServiceJobPredicate.class,                       // <-- HERE!
          sagaFlow -> {
            sagaFlow
              .then(AttachTitusServiceLoadBalancers.class)
              .then(CopyTitusServiceScalingPolicies.class);
          })
        .completionHandler(TitusDeployCompletionHandler.class);
```

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/PrepareTitusDeploy.java
robzienert: So, yes I agree, but I think that Titus should still have an explicit exception for this. We don't have this ability yet, but being able to catch exceptions & then correlate them to a specific cloud provider would be immensely valuable for being able to track the reliability and correctness of a specific cloud provider.

Perhaps `clouddriver-core` could have a marker interface of `FirewallNotFoundException`, but I'd still want it as: `class SecurityGroupNotFoundException : TitusException, FirewallNotFoundException`

Does this make sense?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/AttachTitusServiceLoadBalancers.java
robzienert: This is just an organizational thing.

If you need to modify anything for attaching load balancers to a Titus service, it's all in one place. Furthermore, the likelihood that two `SagaAction` implementations using the same `SagaCommand` is pretty unlikely, but if that were the case, it would then make sense to break that `SagaCommand` to its own file. Since these commands are only being used within a single `SagaAction`, it makes sense to me to embed it as an inner class.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/SubmitTitusJob.java
asher: Lovely, thanks for sharing the code snippet here!

PR #3937, opened 2019-08-13
feat(titus): Adding optional sharedMemory resource attribute

alanmquach:
Adding support for sharedMemory.

Shared memory was added to the protobuf in [rc62](https://github.com/Netflix/titus-api-definitions/releases/tag/0.0.1-rc62) so bumping the protobuf dependency to `rc62`


robzienert:
Looks tasty, but will need to rebase on the saga work I just merged.

alanmquach:
Yum.

PR #3943, opened 2019-08-14
refactor(eventing): Simplify interaction with SpinnakerEvent

robzienert:
Removing `aggregateType` and `aggregateId` from the `SpinnnakerEvent` constructor. This reduces the boilerplate amount of code that people will have to write, as well as avoid any potential bugs from incorrectly passing an incorrect type or ID, as these values are strictly written to by the event lib internals.


clouddriver-saga-test/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/types.kt
```diff
@@ -20,28 +20,26 @@ import com.netflix.spinnaker.clouddriver.saga.flow.SagaAction
 import com.netflix.spinnaker.clouddriver.saga.models.Saga
 import java.util.function.Predicate
 
-class ShouldBranch(sagaName: String, sagaId: String) : SagaEvent(sagaName, sagaId)
+class ShouldBranch : SagaEvent()
```
emjburns: ooo this is more better, I like

PR #3946, opened 2019-08-15
fix(artifacts): Make HttpCredentials fetchUrl throws exception on non-successful response

evansb:
Fixes https://github.com/spinnaker/spinnaker/issues/4756

The OkHTTP method used to download artifact silently ignores application-layer error such as 4xx/5xx responses. This PR catches such errors and re-throw an `IOException` (let me know if you'd like me to throw a more specific exception class, I am unable to find a standard exception class for HTTP error used in the codebase).


evansb:

evansb:

PR #3947, opened 2019-08-15
refactor(kubernetes): Convert all V2 code in clouddriver-kubernetes to Java

ezimanyi:
This PR fixes a few places where code in `clouddriver-kubernetes` was in the wrong place (shared code living in the `/v1` subdirectory and v1-only code in the top-level) then converts everything outside of the `/v1` directory to Java (of which there were just a few remaining classes).

* fix(appengine): Fix package name of MutableCacheData 

  At some point the MutableCacheData class was moved to the appengine provider, but the package name still references kubernetes. Update the package name in the class, and all of the references.

* refactor(kubernetes): Move shared logic out of V1 provider 

  KubernetesAtomicOperationConverterHelper is used by both the V1 and V2 providers, so should not live in the V1 directory tree; move it up to the base level.

  KubernetesApiClientConfig is only used by the V1 provider, so it should be moved into the v1 directory tree.

* refactor(kubernetes): Convert ConfigurationProperties to Java 

  Also, move the top-level classes in that file to their own files as part of the Java conversion.

  In order to make this work better with Lombok-generated accessors, change calls to the accessors: booleans should use isXXX instead of getXXX, and change the capialization of getoAuthScopes.

* refactor(kubernetes): Convert a few more classes to Java 

  Also, as part of this conversion update KubernetesAtomicOperationConverterHelper to use generics, which avoids a lot of necessary casting and provides extra type safety. This extra type safety surfaced a bug in the resize server group converter where we were incorrectly converting the description to a deploy manifest description which compiled before but no longer does.

* refactor(kubernetes): Convert KubernetesCloudProvider to Java 

  Also standardize static access to the ID to just use the public static final field instead of mixing use of the field and the auto-generated static accessor.

* refactor(kubernetes): Convert configuration to Java 

  This is the last non-test class in clouddriver-kubernetes that is still in groovy (excluding classes used exclusively by the V1 provider).


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/KubernetesCloudProvider.java
```diff
@@ -0,0 +1,43 @@
...
+
+  @Override
+  public String getId() {
+    return ID;
+  }
+
+  @Override
+  public String getDisplayName() {
+    return "Kubernetes";
+  }
```
ethanfrogers: Do you still need `@Getter` if you define these overrides yourself?

ezimanyi: Good catch, I don't---will remove!

robzienert:
I don't know what I'm looking at, but I love it 😍 😍 😍 😍 😍

PR #3955, opened 2019-08-16
fix(titus): Loading front50 app should be optional

robzienert:


emjburns:
Why is it ok not to load the app?

robzienert:
Some apps (such as nimble shadow apps) do not have front50 applications.

PR #3942, opened 2019-08-14
perf(docker): Fix for ConcurrentHashMap cast error. Added unit tests

german-muzquiz:
Fix for issue running `DockerRegistryImageCachingAgent` introduced here: https://github.com/spinnaker/clouddriver/pull/3926

Unit tests were added.

PR #3958, opened 2019-08-16
Revert "perf(docker): Fix for ConcurrentHashMap cast error. Added unit tests"

mrampton:
Reverts spinnaker/clouddriver#3942


mrampton:
Hi @ezimanyi -- our integration tests have not been happy with this commit. Would you mind reverting it? We will revise and test internally -- and then push out a new PR when we are happy with the results.

I'll send you a second message for the commit which was cherry-picked into 1.15.x.

Thanks!

PR #3960, opened 2019-08-19
fix(sql): enable ProjectClustersCachingAgent in non-redis modes

dreynaud:


dreynaud:
> Need to do something with: https://github.com/spinnaker/clouddriver/blob/master/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlCacheConfiguration.kt#L136

ah, do you think it would be better to add `ProjectClustersCachingAgent` there instead?

dreynaud:
I removed the empty `CoreProvider` from `SqlCacheConfiguration` in dd71e578cec5a00f92d67726ac65f041a5852d18 instead, my thinking is that it is simpler to reason about if there's only one `CoreProvider` definition

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/CloudDriverConfig.groovy
```diff
@@ -296,19 +298,23 @@ class CloudDriverConfig {
...
+    List<Agent> agents = [
       new ProjectClustersCachingAgent(
         projectClustersService,
         projectClustersCachingAgentProperties
       )
-    ])
+    ]
+
+    if (redisCacheOptions.isPresent() && redisClientDelegate.isPresent()) {
+      agents.add(new CleanupPendingOnDemandCachesAgent(redisCacheOptions, redisClientDelegate, applicationContext))
```
robzienert: You'll need to `redisCacheOptions.get()` and `redisClientDelegate.get()` here.

dreynaud: Groovy, why you let me be an idiot?

PR #3961, opened 2019-08-19
perf(docker): Set cache data id, missing from un-grooving code

german-muzquiz:
Changes of https://github.com/spinnaker/clouddriver/pull/3926 and https://github.com/spinnaker/clouddriver/pull/3942 (both reverted) adding the following:

* Explicitly set id field of `DefaultCacheDataBuilder` and include it in tests. 

This was previously set by groovy code implicitly rather than explicitly, by means of this code:
```
cachedTags[tagKey].with {
    attributes.name = "${repository}:${tag}".toString()
    attributes.account = accountName
    attributes.digest = digest
    attributes.date = creationDate
}
```
cachedTags was this `MapWithDefault` implementation:
```
[:].withDefault { String id -> new DefaultCacheDataBuilder(id: id) }
```
Above code first tried to fetch an entry from the map with key `tagKey`, and that caused the map to create and return a new instance of `DefaultCacheDataBuilder` automatically, with its `id` field set to the value of `tagKey`. When the code was ungroovied, id was missing to be set and cache entries were not properly saved, causing Igor to never find a new docker image.


german-muzquiz:
@ezimanyi

mrampton:
Hi @ezimanyi -- this PR addresses the issue that was causing the previous PR to fail in our integration testing. It has been successfully tested in our environment and additional test coverage was added to ensure that the cache entry is stored with the correct key.

clouddriver-docker/src/test/groovy/com/netflix/spinnaker/clouddriver/docker/registry/provider/agent/DockerRegistryImageCachingAgentTest.groovy
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    when:
+    def cacheResult = agent.loadData(null)
+
+    then:
+    cacheResult.cacheResults.get("taggedImage").size() == total
+    cacheResult.cacheResults.get("imageId").size() == total
+    for (def i = 0; i < total; i++) {
+      cacheResult.cacheResults.get("taggedImage")[i].getId() != ""
```
ezimanyi: It would probably make sense to ensure this is set to the expected value, not just that it's not the empty string.  (This might involve stubbing the key generation logic if you want to keep it outside of the "unit" we're testing.)

clouddriver-docker/src/test/groovy/com/netflix/spinnaker/clouddriver/docker/registry/provider/agent/DockerRegistryImageCachingAgentTest.groovy
```diff
@@ -0,0 +1,113 @@
...
+
+  def "cached tags should include digest"() {
+    given:
+    credentials.trackDigests >> true
+    credentials.repositories >> ["repo-1"]
+    client.getTags("repo-1") >> new DockerRegistryTags().tap {
+      name="repo-1"
+      tags=["tag-1", "tag-2"]
+    }
+    client.getDigest(*_) >> "123"
```
ezimanyi: It might be a bit more robust to return different digests for the different tags, and ensure the digests appear correctly in the output.

clouddriver-docker/src/test/groovy/com/netflix/spinnaker/clouddriver/docker/registry/provider/agent/DockerRegistryImageCachingAgentTest.groovy
```diff
@@ -0,0 +1,113 @@
...
+      total += repoTags[i].size()
+    }
+
+    when:
+    def cacheResult = agent.loadData(null)
+
+    then:
+    cacheResult.cacheResults.get("taggedImage").size() == total
+    cacheResult.cacheResults.get("imageId").size() == total
+    for (def i = 0; i < total; i++) {
```
ezimanyi: The logic in this for loop is somewhat hard to follow, and also looks like it's only testing that each returned tag is something we have in `repoTags` but not the reverse---that everything in `repoTags` is returned in the result.

I'd suggest being more explicit about the expected results, and then asserting that the set of expected results is the same as the set of actual results.

german-muzquiz:
@ezimanyi I improved tests validation by asserting on all expected exact values, they should be easier to read, and to troubleshoot failed tests on expected vs actual values.
I also added a few more tests to cover for exceptions, increasing class coverage to 75%

mrampton:

PR #3965, opened 2019-08-20
refactor(kubernetes): Add global and account-specific registry

ezimanyi:
This is a first step towards addressing spinnaker/spinnaker#4407.  In particular, here we give each account credentials object its own registry of account-specific properties (that falls back to the global properties) and also remove the account-specific `registered` property from `KubernetesKind`.

The next step will be to move the global `KubernetesKind` registry to this same structure, so there is a global registry of built-in kinds and account-specific kinds do not leak to other accounts.  Sending this PR first as this was becoming a large enough change.  I'm sure that the follow-up will require some small changes to the `*AccountResourcePropertyRegistry` classes but the general structure here should work.

* refactor(kubernetes): Remove registered field from KubernetesKind 

  The registered field is only ever used in in the KubernetesResourcePropertyRegistry, where we short-circuit if a kind isn't registered. But the definition of a registered kind is one that has been registered with KubernetesResourcePropertyRegistry, so there's no need to also store this on the kind itself.

  Remove this field, and update the property registry to figure out for itself if a kind is registered.

* refactor(kubernetes): Add global and account-specific registry 

  We currently have a single KubernetesResourcePropertyRegistry that tracks both global properties and has a map of account-specific properties.

  Rather than store the account-specific properties in a global object, let's create a new class to store account-specific properties, AccountResourcePropertyRegistry. The credentials for an account will contain an AccountResourcePropertyRegistry, making it easy to look up relevant information from code that just has the credentials.

  The global properties will live in a GlobalResourcePropertyRegistry; each AccountResourcePropertyRegistry will contain a reference to a singleton GlobalResourcePropertyRegistry and will fall back to that registry if needed.

  Finally, to simplify the few cases where we are looking up account-specific properties without credentials, create a class KubernetesResourcePropertyResolver that handles the boilerplate of resolving accounts and casting them to Kubernetes accounts.

  Finally, replace all uses of the global singleton property registry with the one on the account relevant to the operation.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesAccountResolver.java
```diff
@@ -0,0 +1,54 @@
...
+  private final ResourcePropertyRegistry globalResourcePropertyRegistry;
+
+  public KubernetesAccountResolver(
+      AccountCredentialsRepository credentialsRepository,
+      ResourcePropertyRegistry globalResourcePropertyRegistry) {
+    this.credentialsRepository = credentialsRepository;
+    this.globalResourcePropertyRegistry = globalResourcePropertyRegistry;
+  }
+
+  public Optional<KubernetesV2Credentials> getCredentials(String account) {
```
ethanfrogers: +1, I like this.

PR #3972, opened 2019-08-22
fix(core): Create better task exception summaries

robzienert:
Can move this to kork immediately, but wanted to show a non-generic implementation of (potentially) a more common error summary that can pass between service boundaries.

This should also help make our exceptions a little more obvious to end users when a pipeline fails due to something in clouddriver. Argument could be had for not making the `message` the generic top-most exception message and instead use `cause`; which I'm leaning towards, although this would be a breaking change for existing exception handling. Thoughts?


clouddriver-core/src/test/groovy/com/netflix/spinnaker/clouddriver/orchestration/DefaultOrchestrationProcessorSpec.groovy
```diff
@@ -127,9 +128,53 @@ class DefaultOrchestrationProcessorSpec extends Specification {
...
+  }
+
   private void submitAndWait(AtomicOperation atomicOp) {
     processor.process([atomicOp], taskKey)
     processor.executorService.shutdown()
     processor.executorService.awaitTermination(5, TimeUnit.SECONDS)
   }
+
+  private static class ExternalLibraryException extends RuntimeException {
+    ExternalLibraryException(String var1) {
```
marchello2000: `var1`?

robzienert:
Moved to https://github.com/spinnaker/kork/pull/366

PR #3979, opened 2019-08-23
fix(core): Create better task exception summaries

robzienert:
Re-opening with use of the kork lib.


marchello2000:
thanks! i was just about to make this PR :)

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/DefaultOrchestrationProcessor.groovy
```diff
@@ -172,4 +177,28 @@ class DefaultOrchestrationProcessor implements OrchestrationProcessor {
       log.error("Unable to clear thread locals, reason: ${e.message}")
     }
   }
+
+  /**
+   * For backwards compatibility.
+   *
+   * TODO(rz): Not 100% sure we should keep these two methods.
```
marchello2000: this or hand construction of the object in the 4 places it's used above. seems fine to keep it for now

PR #3980, opened 2019-08-23
fix(ecs): Handle task def not cached yet

clareliguori:


louisjimenez:
@spinnaker/netflix-reviewers @ezimanyi  anyone able to review? thanks 🙂

PR #3981, opened 2019-08-23
fix(ecs): set correct key for clusterNames in applications list

clareliguori:
Cluster list key is assumed by gate to be account name, not app name


louisjimenez:
@spinnaker/netflix-reviewers @ezimanyi  this one as well

PR #3983, opened 2019-08-26
fix(ecs): new task ID format is valid for instance details and termination

clareliguori:
Fixes spinnaker/spinnaker#4799


louisjimenez:
@spinnaker/netflix-reviewers @ezimanyi  last one  🙂

PR #3989, opened 2019-08-28
fix(kubernetes): Use consistent casing for built-in kinds

ezimanyi:
* fix(kubernetes): Revert consistent casing for kind 

  The prior commit to always serialize kinds as lowercase is still something we should do longer-term to reduce the fragility of code that is depending on particular casings
  (ex: replicaSet).

  That being said, it turns out there are more places depending on the current casing than I initially thought. In particular, the logic to substitute parts of a kubernetes manifest (ex: to replace a config map with the fully-qualified config map with a version) depends on the kind being camelCased.

  As we're approaching 1.16, I think that this is a more invasive change that I'd like to do right now, and would also like to decouple that change from the more important work to refactor Kubernetes kinds.  So this commit rolls back the change that always converts kinds to lower-case; the following commit fixes the issue with the refactor a different (and less invasive way).

* fix(kubernetes): Use consistent casing for built-in kinds 

  A lot of code depends on case-sensitive comparisons between kinds; the recent refactor removed the global kind registry so when Spinnaker encounters ReplicaSet it is leaving it cased as ReplicaSet instead of (as before) converting it to replicaSet.

  This change builds a small collection of the built-in kinds and handles that conversion to Spinnaker-canonical casing when we're reading kinds. This is only done for built-in kinds as these are the only ones where we have built-in logic that depends on the case.

  A later PR (after 1.16) will re-implement always converting to lower case and will fix places where we're incorrectly doing case-sensitive comparisons.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKind.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKind.java
@@ -147,8 +147,8 @@ public class KubernetesKind {
   @Override
   public String toString() {
     if (apiGroup.isNativeGroup()) {
-      return lcName;
+      return name;
     }
-    return lcName + "." + apiGroup.toString();
+    return name + "." + apiGroup.toString();
   }
 }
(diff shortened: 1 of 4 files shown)


ezimanyi:
This PR reverts #3987, and applies a less invasive approach.  I still think that the change in #3987 is a better longer-term fix, but I discovered more places that would need to be changed and want to decouple that change from the existing `KubernetesKind` refactor, and push it to after 1.16.

PR #3990, opened 2019-08-28
fix(kubernetes): Dynamic accounts reload when kubeconfig content changed

ncknt:
This PR fixes dynamic account reloading when configured via `git`. I'm not entirely sure but it seems that when `configserver:` is used on `kubeconfigfile`, a new file is created at each polling. Since the file is different, the account is considered new and reinitialized.

The fix is to remove the test on `kubeconfigfile`. We already test on `kubeconfigFileHash`.

This may potentially be a problem if the operator deletes the old file and decide to add a new account with the exact same parameter since `kubeconfigFileHash` is computed once.

Would appreciate some review from @ethanfrogers and @scottfrederick


scottfrederick:
> when `configserver:` is used on `kubeconfigfile`, a new file is created at each polling. Since the file is different, the account is considered new and reinitialized.

I'm curious about this. I believe it's true that a new temp file for the `configserver:kubeconfigfile` will be created on each config refresh, but each file should have the same name and absolute path (after [a change to make this so](https://github.com/spinnaker/clouddriver/pull/3833), which was subsequently moved to `kork-config`). 

Nonetheless, these changes look good to me too and more tests is always a good thing.

PR #3991, opened 2019-08-29
fix(kubernetes): Add Json annotations

ezimanyi:
These annotations were dropped in converting KubernetesKind; this means that we're serializing a KubernetesKind as a full object rather than just as a string. This slightly changes the result of calls to /serverGroups and /loadBalancers as now the 'kind' field is an object not a string. This could cause issues with consumers that expect a specific format; our integration tests expect to see just a string there which is how I found this.


ezimanyi:
I'll follow-up with a PR adding a test; want to merge this to validate that it fixes the integration tests.

PR #3998, opened 2019-09-04
fix(appengine): add RANDOM as a supported ShardBy

perek:
RANDOM is a missing shardBy method causing errors to be logged during CloudDriver boot. This adds RANDOM as a supported enum as seen in the appengine docs: https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1/apps.services#ShardBy


ezimanyi:

PR #4002, opened 2019-09-04
refactor(gce): convert GoogleZonalServerGroupCachingAgent to Java

plumpy:


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleZonalServerGroupCachingAgent.java
```diff
@@ -0,0 +1,906 @@
...
+      serverGroup.setRegion(credentials.regionFromZone(zone));
+    }
+  }
+
+  private static List<String> getZones(DistributionPolicy distributionPolicy) {
+    if (distributionPolicy == null || distributionPolicy.getZones() == null) {
+      return ImmutableList.of();
+    }
+    return distributionPolicy.getZones().stream()
+        .map(z -> Utils.getLocalName(z.getZone()))
```
ezimanyi: Maybe more stream-y as:
```
      .map(DistributionPolicyZoneConfiguration::getZone)
      .map(Utils::getLocalName)
```
🤷‍♂

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleZonalServerGroupCachingAgent.java
```diff
@@ -0,0 +1,906 @@
...
+    if (instanceTemplate != null) {
+      List<AttachedDisk> disks = getDisks(instanceTemplate);
+      instanceTemplate.getProperties().setDisks(disks);
+      launchConfig.put("instanceTemplate", instanceTemplate);
+      launchConfig.put("launchConfigurationName", instanceTemplate.getName());
+      launchConfig.put("instanceType", instanceTemplate.getProperties().getMachineType());
+      launchConfig.put("minCpuPlatform", instanceTemplate.getProperties().getMinCpuPlatform());
+      setSourceImage(serverGroup, launchConfig, disks, credentials, providerCache);
+    }
+    serverGroup.setLaunchConfig(launchConfig);
```
ezimanyi: This looks like one of the few places we set a mutable collection on the `serverGroup`.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleZonalServerGroupCachingAgent.java
```diff
@@ -0,0 +1,906 @@
...
+    }
+
+    List<AttachedDisk> sortedDisks = new ArrayList<>();
+    Optional<AttachedDisk> firstBootDisk =
+        persistentDisks.stream().filter(AttachedDisk::getBoot).findFirst();
+    firstBootDisk.ifPresent(sortedDisks::add);
+    template.getProperties().getDisks().stream()
+        .filter(disk -> !disk.getBoot())
+        .forEach(sortedDisks::add);
+    return sortedDisks;
```
ezimanyi: Could be immutable (particularly since the first `return` does return an immutable list).

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleZonalServerGroupCachingAgent.java
```diff
@@ -0,0 +1,906 @@
...
+    AutoscalingPolicy autoscalingPolicy = autoscaler.getAutoscalingPolicy();
+    if (autoscalingPolicy != null) {
+      serverGroup.setAutoscalingPolicy(autoscalingPolicy);
+      serverGroup.getAsg().put("minSize", autoscalingPolicy.getMinNumReplicas());
+      serverGroup.getAsg().put("maxSize", autoscalingPolicy.getMaxNumReplicas());
+    }
+    if (autoscaler.getStatusDetails() != null) {
+      serverGroup.setAutoscalingMessages(
+          autoscaler.getStatusDetails().stream()
+              .filter(details -> details.getMessage() != null)
```
ezimanyi: nit: I might invert the `filter` and `map` since we're just filtering on the result of the `map`

plumpy: Good call, looks much nicer too:
```
          autoscaler.getStatusDetails().stream()
              .map(AutoscalerStatusDetails::getMessage)
              .filter(Objects::nonNull)
              .collect(toImmutableList())
```

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleZonalServerGroupCachingAgent.java
plumpy: Absolutely, thanks.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleZonalServerGroupCachingAgent.java
plumpy: Okay, fixed this and the other place (serverGroup.setAsg).

Sadly a GoogleServerGroup is going to be very difficult to make fully immutable because it has several fields that are mutable objects we don't control (from the Google Compute APIs) :/

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleZonalServerGroupCachingAgent.java
plumpy: I dunno, I'm generally of the mind that merging consecutive map calls aids readability instead of hinders it? (Obviously up to some limit, where the lambda is still relatively readable.)

IOW: I don't think that's necessary.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleZonalServerGroupCachingAgent.java
```diff
@@ -0,0 +1,906 @@
...
+  }
+
+  @Override
+  public boolean handles(OnDemandType type, String cloudProvider) {
+    return OnDemandType.ServerGroup.equals(type)
+        && GoogleCloudProvider.getID().equals(cloudProvider);
+  }
+
+  @Override
+  public OnDemandResult handle(ProviderCache providerCache, Map<String, ?> data) {
```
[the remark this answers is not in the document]
plumpy: Yeah, good point. I think that because `@Nullable` is not marked `@Inherited`, you do have to add it here too.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleZonalServerGroupCachingAgent.java
```diff
@@ -0,0 +1,906 @@
...
+  @Override
+  public CacheResult loadData(ProviderCache providerCache) {
+
+    try {
+      CacheResultBuilder cacheResultBuilder = new CacheResultBuilder();
+      cacheResultBuilder.setStartTime(System.currentTimeMillis());
+
+      List<GoogleServerGroup> serverGroups = getServerGroups(providerCache);
+
+      // If an entry in ON_DEMAND was generated _after_ we started our caching run, add it to the
```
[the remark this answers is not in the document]
plumpy: Huh, yeah, that definitely seems like a crazy way of doing things. I can think of other adjectives too! But thanks for the backstory. I updated the comment; let me know if it appears inaccurate or incomplete:

```
// We don't evict things unless they've been processed because Orca expects to see these keys
// when calling pendingOnDemandRequests, and then knows the on-demand cache request has
// finished when they disappear. If they go away before Orca has a chance to see them, Orca
// decides the on-demand cache request was ignored completely and sends it again. So we have
// to leave them around a while so Orca can observe them. Is this weird? Yes. Yes, it is.
```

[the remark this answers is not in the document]
ezimanyi: Only slight comment is that orca doesn't know they've succeded "when they disappear" but rather when it sees they have a `processedCount` of >0, but other than that looks great!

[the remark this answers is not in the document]
plumpy: Ah, great, thanks! Updated the comment (again).

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleZonalServerGroupCachingAgent.java
```diff
@@ -0,0 +1,906 @@
...
+          INFORMATIVE.forType(LOAD_BALANCERS.getNs()));
+
+  private static final String ON_DEMAND_TYPE =
+      String.join(":", GoogleCloudProvider.getID(), OnDemandType.ServerGroup.name());
+
+  private static final Splitter COMMA = Splitter.on(',').omitEmptyStrings().trimResults();
+  private static final MapSplitter IMAGE_DESCRIPTION_SPLITTER =
+      Splitter.on(',').withKeyValueSeparator(": ");
+
+  private final GoogleNamedAccountCredentials credentials;
```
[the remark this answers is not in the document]
plumpy: Added `@ParametersAreNonnullByDefault` to the class. (Which then caused IntelliJ to highlight a missing `@Nullable` annotation.)

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleZonalServerGroupCachingAgent.java
```diff
@@ -0,0 +1,906 @@
...
+            (String)
+                cacheResultBuilder
+                    .getOnDemand()
+                    .getToKeep()
+                    .get(serverGroupKey)
+                    .getAttributes()
+                    .get("cacheResults"),
+            new TypeReference<Map<String, List<DefaultCacheData>>>() {});
+    onDemandData.forEach(
+        (namespace, cacheDatas) -> {
```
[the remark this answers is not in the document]
plumpy: I think I briefly had this as `cacheDataObjects` but decided to go with `cacheDatas`. :confused:

PR #4004, opened 2019-09-04
feat(sql): Support event storage in SQL backend

robzienert:


clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/event/SqlEventCleanupAgent.kt
```diff
@@ -0,0 +1,83 @@
...
+    log.info("Deleting aggregates last updated earlier than $cutoff ($duration)")
+
+    withPool(ConnectionPools.EVENTS.value) {
+      jooq.transactional(retryRegistry.retry(RETRY_CONFIG_NAME)) { ctx ->
+
+        val deleted = registry.timer(timingId).record<Int> {
+          ctx.deleteFrom(table("event_aggregates"))
+            .where(
+              timestampDiff(field("last_change_timestamp", Timestamp::class.java), currentTimestamp())
+                .greaterThan(DayToSecond.valueOf(duration))
```
asher: This is going to block `SqlEventRepository.save()` and cause a bad time.

clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/event/SqlEventRepository.kt
```diff
@@ -0,0 +1,298 @@
...
+              // The aggregate doesn't exist and we're already expecting a non-zero version.
+              throw AggregateChangeRejectedException(-1, originatingVersion)
+            }
+
+            // The aggregate doesn't exist yet, so we'll go ahead and seed it immediately.
+            // Aggregate tokens are globally unique across all aggregates and calculated monotonically.
+            val initialAggregate = mapOf(
+              field("aggregate_type") to aggregateType,
+              field("aggregate_id") to aggregateId,
+              field("token") to getNextMonotonicToken(ctx),
```
asher: What value does reading from the DB via `getNextMonotonicToken` provides? What prevents collisions if 2 instances enter that in the same millisecond? Is that worth having to rerun the transactional closure?

clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/event/SqlEventRepository.kt
```diff
@@ -0,0 +1,298 @@
...
+        .limit(perPage)
+        .fetchAggregates()
+
+      val remaining = jooq.selectCount().from(AGGREGATES_TABLE)
+        .withConditions(conditions)
+        .fetchOne(0, Int::class.java) - perPage
+
+      EventRepository.ListAggregatesResult(
+        aggregates = aggregates.map { it.model },
+        nextPageToken = if (remaining > 0) aggregates.lastOrNull()?.token else null
```
asher: Token is effectively used as an aggregate row creation timestamp, but since aggregate rows are updated in place over time without the token incrementing, is this the right thing to paginate by? (Maybe, seems similar to sorting by `buildTime` in orca, but I could also see sorting by last updated being desirable.)

clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/event/SqlEventRepository.kt
```diff
@@ -0,0 +1,298 @@
...
+              newEvents.forEach {
+                nextSequence = it.initialize(aggregateType, aggregateId, originatingVersion, nextSequence)
+                step = step.values(it.toSqlValues(objectMapper))
+              }
+              step
+            }
+            .execute()
+
+          // Update the event table with a new version
+          ctx.update(AGGREGATES_TABLE)
```
asher: Comment doesn't match the table being updated.

clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/event/SqlEventRepository.kt
```diff
@@ -0,0 +1,298 @@
...
+    // TODO(rz): Get this from Spring?
+    val retry = RetryConfig.custom<Retry>()
+      .maxAttempts(retryProperties.transactions.maxRetries)
+      .waitDuration(Duration.ofMillis(retryProperties.transactions.backoffMs))
+      .ignoreExceptions(AggregateChangeRejectedException::class.java)
+      .build()
+
+    try {
+      withPool(POOL_NAME) {
+        jooq.transactional(retryRegistry.retry("eventSave", retry)) { ctx ->
```
asher: Potentially long running transaction here (esp. with potential for retries in `getNextMonotonicToken`), be sure to profile under heavy concurrent load.

clouddriver-sql/src/main/resources/db/changelog/20190822-initial-event-schema.yml
```diff
@@ -0,0 +1,159 @@
...
+            tableName: event_aggregates
+            columns:
+              - column:
+                  name: aggregate_type
+                  type: varchar(255)
+                  constraints:
+                    nullable: false
+                    primaryKey: true
+              - column:
+                  name: aggregate_id
```
asher: What do `aggregate_id`'s look like?

clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/event/SqlEventRepository.kt
robzienert: Removed after IRL convo; no longer getting monotomic.

clouddriver-sql/src/main/resources/db/changelog/20190822-initial-event-schema.yml
robzienert: They're hashes (sent from Orca).

clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/event/SqlEventRepository.kt
robzienert: `listAggregates` isn't going to be consumed by anyone but operators / developers as far as I can tell so far. I wanted pagination because it was The Right Thing to do, and also something easy to reason about. If the pagination token were based off of something like a `lastUpdated` field, the pagination results would be inconsistent and would probably lead to some odd behavior for consumers.

clouddriver-sql/src/main/resources/db/changelog/20190822-initial-event-schema.yml
robzienert: https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/KatoService.groovy#L56

PR #4005, opened 2019-09-05
perf(kubernetes): Defer checking valid kinds until after startup

ezimanyi:
* refactor(kubernetes): Rename InvalidKindReason and avoid nulls 

  Instead of having an enum InvalidKindReason, and returning null when a kind is valid, jsut call it KindStatus and use an enum value VALID to represent a valid kind. This avoids the need to use null as a sentinel value.

* refactor(kubernetes): Change lists to immutable sets 

  We store a list of kinds and kinds to omit on each account; these lists are never modified after they are created, and we only ever do set-like operations on them (contains, size).

  In order to make lookups more efficient, and prevent bugs (and make the following refactor easier), change the lists to immutable sets.

* perf(kubernetes): Defer checking valid kinds until after startup 

  Instead of blocking the startup of clouddriver while we check whether we can read every kind in every account, defer these checks until first use.

  Encapsulate this logic in an inner class called KindValidator; this class handles keeping track of what kinds are valid for an account as well as whether metrics are readable for the account. It will only ever attempt to read a kind once (the first time it is asked), and short-circuits just like the earlier logic based on the contents of kinds and omitKinds.

* refactor(kubernetes): Rename isMetricsComputed 

  The fact that whether metrics is enabled or not is computed by reading metrics is an implementation details; make it easier for consumers by renaming this function isMetricsEnabled.

  Also, remove the getter for metrics, which should never be read by consumers of the class, which should always use isMetricsEnabled.

* perf(kubernetes): Don't check valid kinds in agent scheduling 

  We currently check whether kinds are valid while scheduling caching agents which blocks agents starting. Caching agents themselves should
  (and currently do) check whether kinds are valid before actually querying them; remove the check that occurs before scheduling agents.

  This means that at worst we'll schedule a no-op agent that immediately returns for an invalid kind (though given that kinds are generally grouped in agents that handle many kinds, this likely won't happen much in practice).

  While the earlier commit in this PR deferred checking readable kinds, the fact that we're checking all readable kinds during agent scheduling means we're still effectively blocking startup on checking readable kinds. Removing this allow startup to proceed before we've checked all readable kinds.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
@@ -17,6 +17,7 @@
 
 package com.netflix.spinnaker.clouddriver.kubernetes.v2.security;
 
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static lombok.EqualsAndHashCode.Include;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
@@ -78,9 +79,9 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
 
   @Include @Getter private final List<String> omitNamespaces;
 
-  @Include private final Set<KubernetesKind> kinds;
+  @Include private final ImmutableSet<KubernetesKind> kinds;
 
-  @Include private final Set<KubernetesKind> omitKinds;
+  @Include private final ImmutableSet<KubernetesKind> omitKinds;
 
   @Include @Getter private final List<CustomKubernetesResource> customResources;
 
@@ -117,7 +118,7 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
   private final Supplier<List<KubernetesKind>> liveCrdSupplier;
   @Getter private final ResourcePropertyRegistry resourcePropertyRegistry;
   @Getter private final KubernetesKindRegistry kindRegistry;
-  private final KindValidator kindValidator;
+  private final PermissionValidator permissionValidator;
 
   public KubernetesV2Credentials(
       Registry registry,
@@ -148,14 +149,14 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
                           return new KubernetesKindProperties(k, true, false, true);
                         }))
             .map(KubernetesKindProperties::getKubernetesKind)
-            .collect(ImmutableSet.toImmutableSet());
+            .collect(toImmutableSet());
     // omitKinds is a simple placeholder that we can use to compare one instance to another
     // when refreshing credentials.
     this.omitKinds =
         managedAccount.getOmitKinds().stream()
             .map(KubernetesKind::fromString)
-            .collect(ImmutableSet.toImmutableSet());
-    this.kindValidator = new KindValidator(this.kinds, this.omitKinds);
+            .collect(toImmutableSet());
+    this.permissionValidator = new PermissionValidator();
 
     this.customResources = managedAccount.getCustomResources();
 
@@ -293,7 +294,7 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
 
   @Nonnull
   public KubernetesKindStatus getKindStatus(@Nonnull KubernetesKind kind) {
-    return kindValidator.getKindStatus(kind);
+    return permissionValidator.getKindStatus(kind);
   }
 
   public String getDefaultNamespace() {
@@ -367,7 +368,7 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
   }
 
   public boolean isMetricsEnabled() {
-    return metrics && kindValidator.isMetricsReadable();
+    return metrics && permissionValidator.isMetricsReadable();
   }
 
   public KubernetesManifest get(KubernetesKind kind, String namespace, String name) {
@@ -578,18 +579,10 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
    * Handles validating whether kubernetes kinds should be cached by the current account, and
    * whether metrics are enabled for the account.
    */
-  private class KindValidator {
+  private class PermissionValidator {
     private final Supplier<String> checkNamespace = Suppliers.memoize(this::computeCheckNamespace);
     private final Map<KubernetesKind, Boolean> readableKinds = new ConcurrentHashMap<>();
     private final Supplier<Boolean> metricsReadable = Suppliers.memoize(this::checkMetricsReadable);
-
-    private final Set<KubernetesKind> kinds;
-    private final Set<KubernetesKind> omitKinds;
-
-    KindValidator(Set<KubernetesKind> kinds, Set<KubernetesKind> omitKinds) {
-      this.kinds = kinds;
-      this.omitKinds = omitKinds;
-    }
 
     private String getCheckNamespace() {
       return checkNamespace.get();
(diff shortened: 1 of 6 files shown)


ezimanyi:
Fixed the tests; I had an extra commit locally that I forgot to push.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -679,4 +573,130 @@ public void patch(
...
+      // otherwise, checking all namespaces for all kinds is too expensive in large clusters
+      // (imagine
+      // a cluster with 100s of namespaces).
+      return namespaces.get(0);
+    }
+
+    private boolean skipPermissionChecks() {
+      // checkPermissionsOnStartup exists from when permission checks were done at startup (and took
+      // a long time); this flag was added to skip the checks and assume all kinds were readable.
+      // Now that permissions are checked on-the-fly, this flag is probably not necessary, but for
```
plumpy: > but for now...

If possible, I like to put clear removal timelines in the comment like "this can be removed for the 1.18 release unless we see major problems with the new startup behavior in 1.17". So that when we forget about this and stumble across it again in 2021 we can feel confident in removing it. But maybe this requires more discussion with customers so a hard timeline isn't appropriate.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -679,4 +573,130 @@ public void patch(
...
+
+    private String getCheckNamespace() {
+      return checkNamespace.get();
+    }
+
+    private String computeCheckNamespace() {
+      List<String> namespaces = getDeclaredNamespaces();
+
+      if (namespaces.isEmpty()) {
+        log.warn(
```
plumpy: use `{}` instead of String concatenation for log messages? It seems unlikely `warn` messages would be disabled, but still, good habit...

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -679,4 +573,130 @@ public void patch(
       }
     }
   }
+
+  /**
+   * Handles validating whether kubernetes kinds should be cached by the current account, and
+   * whether metrics are enabled for the account.
+   */
+  private class KindValidator {
+    private final Supplier<String> checkNamespace = Suppliers.memoize(this::computeCheckNamespace);
```
plumpy: nit: a lot of people (including me, but certainly not everyone) leave out `private` visibility declarations inside private classes, since they're arguably redundant noise. But up to you.

Java Practices seems to take no real position: go/java-practices/redundancy#visibility-specifiers-in-private-classes

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -679,4 +573,130 @@ public void patch(
...
+                + accountName
+                + "' doesn't prevent access from all namespaces in this cluster, or that the cluster is reachable.");
+        return null;
+      }
+
+      // we are making the assumption that the roles granted to spinnaker for this account in all
+      // namespaces are identical.
+      // otherwise, checking all namespaces for all kinds is too expensive in large clusters
+      // (imagine
+      // a cluster with 100s of namespaces).
```
plumpy: formatting: pull this line up

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -679,4 +573,130 @@ public void patch(
...
+      return readableKinds.computeIfAbsent(kind, this::canReadKind)
+          ? KubernetesKindStatus.VALID
+          : KubernetesKindStatus.READ_ERROR;
+    }
+
+    /**
+     * Returns whether metrics are readable for the current kubernetes account. This check is cached
+     * for the lifetime of the process, and subsequent calls return the cached value.
+     */
+    boolean isMetricsReadable() {
```
plumpy: Putting this metrics stuff in a class called `KindValidator` is sort of confusing (it seems to be entirely unrelated to validating kinds?) but not a huge deal, I guess...

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -76,9 +77,9 @@
 
   @Include @Getter private final List<String> omitNamespaces;
 
-  @Include private final List<KubernetesKind> kinds;
+  @Include private final Set<KubernetesKind> kinds;
```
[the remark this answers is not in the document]
ezimanyi: Yeah, that makes sense; I'll change that.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -149,10 +150,13 @@ public KubernetesV2Credentials(
                           return new KubernetesKindProperties(k, true, false, true);
                         }))
             .map(KubernetesKindProperties::getKubernetesKind)
-            .collect(Collectors.toList());
+            .collect(ImmutableSet.toImmutableSet());
```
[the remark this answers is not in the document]
ezimanyi: Yeah, I tend to avoid static imports as I find myself always scrolling up to figure out where something came from, though I suppose this is one of the more common ones.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -679,4 +573,130 @@ public void patch(
...
+  /**
+   * Handles validating whether kubernetes kinds should be cached by the current account, and
+   * whether metrics are enabled for the account.
+   */
+  private class KindValidator {
+    private final Supplier<String> checkNamespace = Suppliers.memoize(this::computeCheckNamespace);
+    private final Map<KubernetesKind, Boolean> readableKinds = new ConcurrentHashMap<>();
+    private final Supplier<Boolean> metricsReadable = Suppliers.memoize(this::checkMetricsReadable);
+
+    private final Set<KubernetesKind> kinds;
```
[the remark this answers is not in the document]
ezimanyi: Oh, good point.  This is a relic from back when I wanted them to be `Set`s in the inner class, but they were still `List`s in the outer class.  Now that I updated the outer class to store `ImmutableSet<KubernetesKind>`s there's no need for them here anymore.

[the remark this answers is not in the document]
ezimanyi: At some point if I disentangle this in a later change and actually make this static (or pull it out of being an inner class) these can get added back but for now it seems silly to have them when the inner class is reaching out for everything else.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
ezimanyi: Just chatted about this; in this case I added `private` to anything not intended to be called from outside the inner class, and left the two methods intended to be called from the outer class as package private.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
ezimanyi: Would it be more clear if this were called something like `PermissionValidator`?  While `metrics` is not really a kind, the thing we're doing is roughly the same---checking to see if the current service account has permission to read something in the cluster.

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
ezimanyi: Added some more detail

clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
plumpy: Yeah, I think it would.

PR #4007, opened 2019-09-06
refactor(gce): Convert GoogleRegionalServerGroupCachingAgent to java

plumpy:


clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/AbstractGoogleServerGroupCachingAgentTest.java
```diff
@@ -0,0 +1,748 @@
...
+        new GoogleComputeApiFactory(
+            new GoogleOperationPoller(),
+            new DefaultRegistry(),
+            "user-agent",
+            MoreExecutors.listeningDecorator(Executors.newCachedThreadPool())),
+        new DefaultRegistry(),
+        REGION,
+        new ObjectMapper()) {
+      @Override
+      Collection<String> getOnDemandKeysToEvictForMissingServerGroup(
```
ezimanyi: Is there some way to annotate all these methods with `@ParametersAreNonnullByDefault`, as they are overriding methods that have that annotation.  Maybe there's not a clean way, in which case it probably doesn't make sense to clutter up this test code with it.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/AbstractGoogleServerGroupCachingAgent.java
```diff
@@ -0,0 +1,941 @@
...
+  public OnDemandMetricsSupport getMetricsSupport() {
+    return onDemandMetricsSupport;
+  }
+
+  @Override
+  public String getAccountName() {
+    return credentials.getName();
+  }
+
+  GoogleNamedAccountCredentials getCredentials() {
```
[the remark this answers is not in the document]
plumpy: It's kinda like putting a `public` method in a `private` class... maybe it's useful documentation, but it doesn't actually do anything. So, I dunno. I've done both, but I think for now I'm leaning towards limiting visibility and matching the class-level visibility.

clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/AbstractGoogleServerGroupCachingAgentTest.java
plumpy: I did `Refactor→Convert Anonymous to Inner`, which is probably better anyway... an anonymous class with 8 method implementations is pretty ugly, thanks.

PR #4008, opened 2019-09-06
perf(kubernetes): Defer loading namespaces until first use

ezimanyi:
The goal here is to avoid blocking startup while waiting to read namespaces from an unreachable/down cluster.  Just as we did with checking readable kinds, this defers reading namespaces until first use, where waiting for the call to timeout will not block caching other clusters.

* refactor(kubernetes): Push namespace logic down to V1 and V2 

  A following commit will simplify the logic for handling namespaces in the kubernetes caching agents. There is a lot of legacy V1 code in groovy that depends on the particular implementation that I don't want to refactor and test as part of this change. In order to restrict the change to V2, push this logic down from the base caching agent to the V1 and V2 caching agents, where it is for now duplicated but will soon be changed in V2.

  This also makes the KubernetesMetricCachingAgent extend the KubernetesV2CachingAgent rather than the base KubernetesCachingAgent so it can share the new namespace functionality.

* refactor(kubernetes): Make namespaces private 

  Change namespaces to be private, and have all the consumers use the accessor (which can be protected instead of public). Also, return an ImmutableList of namespaces rather than a list.

* perf(kubernetes): Defer loading namespaces until first use 

  The kubernetes v2 caching agents store a list of namespaces of interest, and have a complex system of calling loadNamespaces() to refresh this list then getNamespaces() to fetch them.

  It turns out that we always just call loadNamespaces() immediately before a call to getNamespaces() so the cache is doing nothing. Remove the cached namespaces variable and all calls to loadNamespaces(). Instead just have getNamespaces() directly look up the namespaces (which was happening anyway before).

  The main benefit is that now we can remove the call to loadNamespaces() to prime the cache from the caching agent constructor. This means that caching agents don't need to communicate with the cluster in their constructor, so startup will not be blocked waiting for calls to an unreachable cluster to time out.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2CachingAgent.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2CachingAgent.java
@@ -18,6 +18,7 @@
 package com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.agent;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.ImmutableList;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.cats.agent.AgentIntervalAware;
 import com.netflix.spinnaker.cats.agent.CacheResult;
@@ -52,7 +53,7 @@ public abstract class KubernetesV2CachingAgent
   @Getter protected String providerName = KubernetesCloudProvider.ID;
 
   @Getter protected final Long agentInterval;
-  protected List<String> namespaces;
+  private ImmutableList<String> namespaces;
 
   protected KubernetesV2CachingAgent(
       KubernetesNamedAccountCredentials<KubernetesV2Credentials> namedAccountCredentials,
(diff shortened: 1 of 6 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2CachingAgent.java
```diff
@@ -193,4 +193,10 @@ protected CacheResult buildCacheResult(Map<KubernetesKind, List<KubernetesManife
             });
     return result;
   }
+
+  protected ImmutableList<String> getNamespaces() {
+    return credentials.getDeclaredNamespaces().stream()
+        .filter(n -> agentCount == 1 || Math.abs(n.hashCode() % agentCount) == agentIndex)
```
ethanfrogers: What exactly is this doing?

ezimanyi: It's handling splitting of namespaces among the caching agents.

So if you have 5 caching agents and this is caching agent 2, this agent only handles namespaces where `Math.abs(n.hashCode() % 5) == 2`.  (If there's only one agent, we need to handle all namespaces.)

The hashing makes sure the assignment of namespaces to caching agents is stable (and not dependent on the order we get the list back from the cluster).

PR #4009, opened 2019-09-06
refactor(web): Remove logic from operations controller into service class

robzienert:
Just some cleanup before I do some other work. Also some annotations floating around on some interfaces as I figure out what I need to do and I'm figuring out the implicit contracts of these interfaces.


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AtomicOperationConverterNotFoundException.java
```diff
@@ -0,0 +1,18 @@
+package com.netflix.spinnaker.clouddriver.orchestration;
```
scottfrederick: Missing copyright

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AtomicOperationException.java
```diff
@@ -0,0 +1,36 @@
+package com.netflix.spinnaker.clouddriver.orchestration;
```
scottfrederick: Missing copyright

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AtomicOperationNotFoundException.java
```diff
@@ -0,0 +1,20 @@
+package com.netflix.spinnaker.clouddriver.orchestration;
```
scottfrederick: Missing copyright

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/OperationsController.groovy
```diff
@@ -49,165 +31,53 @@ import org.springframework.web.bind.annotation.RestController
...
 
-  @Autowired DescriptionAuthorizer descriptionAuthorizer
-
-  /*
-   * APIs
-   * ----------------------------------------------------------------------------------------------------------------------------
-   */
-
-  /**
+  @Autowired
```
scottfrederick: `@Autowired` isn't required on ctors on Java classes, is it required for Groovy?

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/OperationsController.groovy
```diff
@@ -49,165 +31,53 @@ import org.springframework.web.bind.annotation.RestController
...
   @Deprecated
   @RequestMapping(value = "/ops", method = RequestMethod.POST)
-  Map<String, String> operations(@RequestParam(value = "clientRequestId", required = false) String clientRequestId,
-                                 @RequestBody List<Map<String, Map>> requestBody) {
-    List<AtomicOperation> atomicOperations = collectAtomicOperations(requestBody)
-    start(atomicOperations, clientRequestId)
+  OperationsService.StartOperationResult operations(
+    @RequestParam(value = "clientRequestId", required = false) String clientRequestId,
+    @RequestBody List<Map<String, Map>> requestBody) {
+    List<AtomicOperation> atomicOperations = operationsService.collectAtomicOperations(requestBody)
```
scottfrederick: Calling one method of `operationsService` to get a value (`atomicOperations`), just to pass that value to another method of the same service, is a bad smell. The pattern is repeated several times in this class. 

Can both calls be encapsulated into one method? Or is that an indication that `collectAtomicOperations` and `start` should belong to separate services?

robzienert: Good point. Thank you - I'll split these into two services. I wasn't too pleased with this myself, thanks for keeping me honest. :)

PR #4013, opened 2019-09-09
refactor(kubernetes): Remove extraneous null checks

ezimanyi:
* refactor(kubernetes): Initialize property registry in credentials class 

  The property registry is owned by the credentials class; rather than have the agent scheduler call into the class to initialize it, just have the class handle this in its initialize method.

* refactor(kubernetes): Make AccountResourcePropertyRegistry immutable 

  AccountResourcePropertyRegistry support a register method, but it is only used a single time to initialize the registry with the account's KubernetesResourceProperties.

  Instead, require the KubernetesResourceProperties in the constructor, and remove the register method.

* refactor(kubernetes): Make GlobalResourcePropertyRegistry immutable 

  GlobalResourcePropertyRegistry has a register method that is only ever called from the constructor. Remove this method and instead build an immutable map in the constructor.

* refactor(kubernetes): Update return values from registries to be immutable 

  Now that both AccountResourcePropertyRegistry and GlobalResourcePropertyRegistry are immutable, update the public methods on these classes, and the interface to return immutable collections where approriate. Also add NonNull annotations where appropriate.

* refactor(kubernetes): Add Nonnull annotation to handler kinds 

  In an upcoming commit, we're going to be storing the KubernetesKind and SpinnakerKind associated with handlers in an ImmutableMap, which won't allow null values.

  As we already have null objects for each of these (KubernetesKind.NONE and SpinnakerKind.UNCLASSIFIED) there's no need to ever have these methods return null.  Add annotations to the interface and implementations, and update the one implementation returning null to instead return SpinnakerKind.UNCLASSIFIED.

* refactor(kubernetes): Let KubernetesSpinnakerKindMap initialize itself 

  It's a bit strange that GlobalResourcePropertyRegistry is responsible for initializing KubernetesSpinnakerKindMap. (It takes the singleton kind map in the constructor, mutates it, then forgets about it.) This is a vestige of the way the code worked before the kind registries were created.

  Instead, let KubernetesSpinnakerKindMap handle its own initialization; its constructor should expect the same list of KubernetesHandlers and should register the kinds found on each of them.

  This also means that we can make the register method private as it's only called from the constructor. If you were to guess that the next commit would remove the register method entirely and make the class immutable, you would probably be correct...

* refactor(kubernetes): Make KubernetesSpinnakerKindMap immutable 

  Now that KubernetesSpinnakerKindMap handles its own initialization, the constructor has all the information needed to completely construct the maps. Create immutable maps in the constructor and remove the register method.

* refactor(kubernetes): Enforce that defaultProperties are not null 

  We've added the NonNull annotation to GlobalResourcePropertyRegistry.get because we fall back to the properties for KubernetesKind.NONE for a kind that's not in the registry. While it is statically true that there will be properties for the NONE kind (because KubernetesUnregisteredCustomResourceHandler handles the NONE kind) it's a bit confusing to see how this defaulting happens and is liable to break if one day we remove the NONE kind from that handler.

  Instead, let's be more explicit and expect to receive the unregistered custom resource handler bean in the constructor of the global registry, and set that as the default to return. Then if we somehow don't have this default value, autowiring will fail at startup (instead of causing null-pointer exceptions later).

  This makes it much more clear that get() always returns a non-null value, even though it always did.

* refactor(kubernetes): Remove extraneous null checks 

  Now that we have statically guaranteed that the return value of KubernetesKindRegistry.get is non-null, we can remove multiple extraneous non-null checks from the code.

### DIFF
--- a/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesSpinnakerKindMap.java
+++ b/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesSpinnakerKindMap.java
@@ -17,15 +17,13 @@
 
 package com.netflix.spinnaker.clouddriver.kubernetes.v2.description;
 
-import static com.google.common.collect.ImmutableMap.toImmutableMap;
-
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableSetMultimap;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.KubernetesKind;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.op.handler.KubernetesHandler;
 import java.util.Arrays;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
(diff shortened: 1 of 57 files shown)


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesSpinnakerKindMap.java
```diff
@@ -60,35 +62,37 @@ public static SpinnakerKind fromString(String name) {
     }
   }
 
-  private final Map<SpinnakerKind, Set<KubernetesKind>> spinnakerToKubernetes = new HashMap<>();
-  private final Map<KubernetesKind, SpinnakerKind> kubernetesToSpinnaker = new HashMap<>();
+  private final ImmutableMap<SpinnakerKind, ImmutableSet<KubernetesKind>> spinnakerToKubernetes;
```
plumpy: If you make this an `ImmutableSetMultimap`, you can make that constructor a whole lot simpler. I don't think anything else needs to change. (A `Multimap` is essentially just a fancy Guava wrapper around `Map<Key, Collection<Value>>` with some nice convenience methods.)

ezimanyi: That's awesome, I briefly searched for a cleaner way of doing this and didn't find one...I'll make that change now.

PR #4014, opened 2019-09-09
feat(titus): Mimicker property testing

robzienert:
I wanted to have a concrete example to show for the https://github.com/spinnaker/kork/pull/373 work. Surprisingly, there weren't any changes I needed to make as a result. The tests do likely need to get more coverage regardless.

### DIFF
--- a/clouddriver-titus/src/test/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/actions/PrepareTitusDeployActionSpec.groovy
+++ b/clouddriver-titus/src/test/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/actions/PrepareTitusDeployActionSpec.groovy
@@ -32,18 +32,55 @@ import com.netflix.spinnaker.clouddriver.titus.deploy.actions.PrepareTitusDeploy
 import com.netflix.spinnaker.clouddriver.titus.deploy.actions.SubmitTitusJob
 import com.netflix.spinnaker.clouddriver.titus.deploy.description.TitusDeployDescription
 import com.netflix.spinnaker.config.AwsConfiguration
+import com.netflix.spinnaker.kork.test.mimicker.DataContainer
+import com.netflix.spinnaker.kork.test.mimicker.Mimicker
+import com.netflix.spinnaker.moniker.Moniker
+import com.netflix.spinnaker.moniker.frigga.FriggaReflectiveNamer
+import spock.lang.Shared
 import spock.lang.Specification
 import spock.lang.Subject
 import spock.lang.Unroll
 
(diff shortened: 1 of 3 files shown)


clouddriver-titus/src/test/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/actions/PrepareTitusDeployActionSpec.groovy
```diff
@@ -126,24 +168,34 @@ class PrepareTitusDeployActionSpec extends Specification {
...
 
-    description.securityGroups == ["sg-2", "sg-1"]
+    description.securityGroups == [sg2Id, sg1Id]
   }
 
   @Unroll
   def "security groups include app security group (label=#labelValue, desc=#descriptionValue, includesAppGroup=#includesAppGroup)"() {
     given:
-    TitusDeployDescription description = createTitusDeployDescription()
+    String sg1Id = mimicker.aws().securityGroupId
```
cfieber: as a mild bit of test coding style I'd generally prefer to stick these kinds of test data in the `where:` section. I suspect that @robfletcher would have the most valid opinion as senior VP of spock testing though

robzienert: Oh, sure yeah that makes sense... I'll change.

gradle.properties
```diff
@@ -2,6 +2,6 @@
 includeCloudProviders=all
 fiatVersion=1.4.0
 enablePublishing=false
-korkVersion=6.6.2
+korkVersion=6.5.0-SNAPSHOT
```
emjburns: should this be the snapshot version, or is this leftover from testing locally?

clouddriver-titus/src/test/resources/mimicker-titus.yml
```diff
@@ -0,0 +1,2 @@
+mimicker:
+  titus:
```
emjburns: does this give you a place to define some inputs or configure the behavior for the instance of mimicker that you create?

gradle.properties
emjburns: oh I see the PR wasn't merged into kork yet.

clouddriver-titus/src/test/resources/mimicker-titus.yml
robzienert: It gives you a place to define inputs. Example:

```yaml
mimicker:
  titus:
    cells:
    - cell1
    - cell2
    - cell3
```

```kotlin
dataContainer.list<String>("titus/cells") // returns full list of titus cells.
```

PR #4016, opened 2019-09-09
perf(kubernetes): Fix startup when a cluster is unreachable

ezimanyi:
This PR removes that last two live calls to a kubernetes cluster during clouddriver startup.  This means that clouddriver will successfully start up even if there is an account pointing to an unreachable cluster; the errors will then be restricted to the affected accounts and caching will continue for unaffected accounts.

* perf(kubernetes): Defer checking if metrics are enabled 

  Rather than checking if metrics are enabled before scheduling caching agents, push the responsibility for determining this down to the metrics caching agent. This means we'll always schedule the metrics caching agent, but it will be a no-op if metrics are disabled.

  The advantage here is that we don't need to make a live lookup to determine if metrics are readable during startup.

* perf(kubernetes): Remove reading of CRDs on startup 

  We currently try to read the list of CRDs for a cluster when the account is initialized on clouddriver startup. This means that if a cluster is unreachable, we'll block the entire startup waiting for the connection to time out. Remove this check on startup, and defer it until the first actual request to read CRDs.

  There is a subtle history behind why that check was initially added on startup, but the refactoring that has been done since obviates the need for it. In particular, we needed to check for CRDs on startup to:
  (1) Make sure we included them in the check for unavailable kinds that happened on startup
  (2) Ensure that we did not end up with null resource properties for the resource, preventing deployment (as in spinnaker/spinnaker#3439).

  (1) is no longer an issue because we no longer check for readable kinds on startup; the first time we encouter a kind, we'll check if it's readable.
  (2) is no longer an issue because the refactor of the property registries has enabled us to statically guarantee that no kinds will return null properties. (The error messages in the linked issue have been completely removed.) Dynamically-discovered kinds always deployed using the unregistered custom resource handler; this will now be returned for them regardless of whether the account has polled for crds yet or not.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2CachingAgentDispatcher.java
```diff
@@ -80,20 +80,18 @@ public KubernetesV2CachingAgentDispatcher(ObjectMapper objectMapper, Registry re
...
-                      new KubernetesMetricCachingAgent(
-                          credentials,
-                          objectMapper,
-                          registry,
-                          i,
-                          credentials.getCacheThreads(),
-                          agentInterval)));
-    }
+    IntStream.range(0, credentials.getCacheThreads())
+        .boxed()
```
plumpy: while you're in here, remove the call to boxed?

ezimanyi: :+1: 
:no_entry_sign: :package:

PR #4017, opened 2019-09-10
refactor(*): Remove credentialAccount from all AWS-derived cloud providers

robzienert:
What could possibly go wrong with this? Really. What could possibly go wrong with this? I know we have duplication for `credentials` vs `account`, this is standardizing on `credentials`, since like... I don't even know why `credentialAccount` ever existed to begin with.

It looks to me like most of the code for descriptions expects `credentials`, whereas `account` isn't really referenced directly anyhow? (Looking at `AtomicOperationConverter`s for the AWS-derived cloud providers).


robzienert:
yolo

PR #4019, opened 2019-09-11
feat(saga): Add support for re-entrance of sagas

robzienert:
Now that we're actually persisting sagas into a durable store, we can start the work on making saga-backed operations retryable. From a saga point of view, when we re-apply a saga, it will inspect the committed event log and determine what the next `Step` is in the `SagaFlow` that needs to be executed, and will internally seek the `SagaFlowIterator` to that point.


clouddriver-saga/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/flow/SagaFlowIterator.kt
```diff
@@ -81,6 +96,29 @@ class SagaFlowIterator(
...
   }
 
+  /**
+   * Seeks the iterator to the next step that needs to be (re)started, if the saga has already begun.
+   *
+   * Multiple strategies are used to locate the correct index to seek to. The highest index returned from the [Seeker]
+   * strategies will be used for seeking.
+   *
+   * TODO(rz): What if there is more than 1 of a particular command in a flow? :thinking_face: May need more metadata
+   * in the [SagaCommandCompleted] event passed along...
```
robzienert: This currently isn't an issue, so I'm kicking the problem down the road.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/LoadFront50App.java
```diff
@@ -124,11 +124,11 @@ public Result apply(@NotNull LoadFront50AppCommand command, @NotNull Saga saga)
             "Failed to convert front50 application to internal model", e);
       }
     } catch (Exception e) {
-      log.error("Failed to load front50 application attributes for {}", command.getAppName(), e);
       if (command.isAllowMissing()) {
         // It's ok to not load the front50 application
         return new Result(command.nextCommand, Collections.emptyList());
       }
+      log.error("Failed to load front50 application attributes for {}", command.getAppName(), e);
```
robzienert: This little guy snuck in here on accident, but I'm not gonna make a PR just for this change anyway. This reduces false error log noise.

PR #4020, opened 2019-09-11
refactor(kubernetes): Clean up ArtifactReplacer

ezimanyi:
This is not in service of any other change, I've just been meaning to clean up this class for a while as I find it confusing whenever I need to read it.  There's definitely more that can be done here, but this is what I was able to do in the few hours I allotted for this.

Since I moved a lot of code, I'd definitely recommend reading the commits separately.

* test(kubernetes): Fix ArtifactReplacer tests 

  We should be using when/then instead of expect

* refactor(kubernetes): Make ArtifactReplacer immutable 

  We always create an ArtifactReplacer then immediately call registerReplacer a number of times to register all replacers. Replace this with just passing the replacers into the constructor.

* refactor(kubernetes): Delete unused static variables 

  At some point in the past, this class was refactored and all references to these static variables were removed, but the variables themselves were never removed.  Remove them.

* refactor(kubernetes): Replace static methods with constants 

  As the Replacer class is immutable, it is safe to just return the same Replacer ever time we need one. Let's replace the static methods to generate a replacer with static constants.

* refactor(kubernetes): Improve encapsulation of Replacer class 

  The ArtifactReplacer and Replacer classes don't clearly define a boundary of responsibility; this commit is a few changes to disentangle the classes a bit, as well as make a few minor other improvements to them.

  In particular:
  * Replace stream .filter() and .map() operations that have side- effects with forEach
  * Add nullity annotations where appropriate
  * Remove the unused (always null) namePattern field from replacer
  * Consolidate the API of Replacer down to getArtifacts and ReplaceArtifacts

* refactor(kubernetes): Move Replacer class to top level 

  This is a reasonably complex class to be an inner class, which reduces its ability to encapsulate its complexity. This commit is just moving the existing class, with no other changes.

* refactor(kubernetes): Make Replacer builder private 

  Defining a replacer is reasonably complex, so ideally this would only be done by the class itself, exposing useful replacers for consumers. Move the one replacer defined outside the class into the class, and make the builder private.


clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/artifact/ArtifactReplacerFactory.java
```diff
@@ -21,112 +21,96 @@
...
-        .findPath("$..spec.template.spec['containers', 'initContainers'].*.image")
-        .nameFromReference(
-            ref -> {
-              int atIndex = ref.indexOf('@');
-              // @ can only show up in image references denoting a digest
-              // https://github.com/docker/distribution/blob/95daa793b83a21656fe6c13e6d5cf1c3999108c7/reference/regexp.go#L70
-              if (atIndex >= 0) {
-                return ref.substring(0, atIndex);
-              }
+  public static final Replacer DOCKER_IMAGE =
```
plumpy: FYI I would have been inclined to make the constants `private` and kept the static methods since
1. the fact that they're singletons is not particularly relevant to clients
2. if you later need to add one that _isn't_ a singleton (perhaps it takes a parameter), the resulting static method would be more consistent if the rest of them were too

I generally try to limit my `public` constants to primitives. But whatever, definitely not a huge deal. I only really point it out  because everything else is too ✨ 💯  to comment on.

PR #4021, opened 2019-09-11
fix(aws): Alter AllowLaunchDescription to not overload account

robzienert:
We asked, "What could possibly go wrong?" with #4017. It turns out `AllowLaunchDescription` overloaded `account`, which broke any deploy from a newly baked image. That's what can go wrong.

This PR fixes that, making the description use a new field and fix-forward whatever Orca sends us.


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/preprocessors/AllowLaunchPreProcessor.java
```diff
@@ -0,0 +1,40 @@
...
+  @Override
+  public boolean supports(Class descriptionClass) {
+    return AllowLaunchDescription.class.isAssignableFrom(descriptionClass);
+  }
+
+  @Override
+  public Map process(Map description) {
+    // Backwards-compatibility from when AllowLaunch used to overload "account" from the abstract
+    // AWS description.
+    description.put("targetAccount", description.get("account"));
```
cfieber: I'd propose for this to do a putIfAbsent("targetAccount") instead otherwise callers will have to continue to supply account once they have been updated

robzienert: Good one. OK

PR #4023, opened 2019-09-12
refactor(kubernetes): Fix some unsafe casts in credential creation

ezimanyi:
The motivation here is to fix a number of unsafe casts in the creation of credentials, and to clean up that code a bit.  I ran into this while working on my next performance change (in an upcoming PR) to fix a few endpoints that hang when a cluster is unreachable (now that startup is fixed in that case). 

* refactor(kubernetes): Split CredentialFactory to V1 and V2 

  A lot of the unecessary casts that we do when creating credentials stem from the fact that we have a single CredentialFactory with different methods for creating V1 and V2 credentials.

  Instead, make a generic interface KubernetesCredentialFactory that pulls out the common logic to default methods.  (These default methods should probably both eventually live somewhere else, but I think having them on the interface is a resonable solution for now.)

  Then give both the V1 and V2 credentials an inner factory class that implements that interface and allows us to create credentials in a type-safe way. It also logically makes more sense, as the factory for credentials lives with the credentials it's creating.

* refactor(kubernetes): Add some generic bounds to the account code 

  Add some parametrized types to eliminate a bunch of unsafe casts.

* refactor(kubernetes): Push getSpinnakerKindMap to V1/V2 

  The getSpinnakerKindMap function on the base KubernetesNamedAccountCredentials class is needed to support sending the kind map to the UI. It forks using an if block depending on which implementation we're using, which means we should push the logic down to the actual implementations.

  One solution would have been to keep the kind map at the top level and just implement an "add custom resources" in the V2 provider class (that is a no-op in V1). I didn't do that because:
  (1) It's strange that there's logic in this class anyway, which is just supposed to hold some metadata and delegate down to the actual credentials.
  (2) Ultimately, it probably makes sense for the V2 provider to implement its own account-specific KubernetesSpinnakerKindMap instead of building this map on the fly. Pushing this down to the V1/V2 implementations allows us to do this later.

### DIFF
--- a/clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/ops/securitygroup/UpsertKubernetesV1SecurityGroupAtomicOperationSpec.groovy
+++ b/clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/ops/securitygroup/UpsertKubernetesV1SecurityGroupAtomicOperationSpec.groovy
@@ -60,7 +60,7 @@ class UpsertKubernetesV1SecurityGroupAtomicOperationSpec extends Specification {
     dockerRegistry = Mock(LinkedDockerRegistryConfiguration)
     dockerRegistries = [dockerRegistry]
     accountCredentialsRepositoryMock = Mock(AccountCredentialsRepository)
-    credentials = new KubernetesV1Credentials(apiMock, NAMESPACES, [], [], accountCredentialsRepositoryMock)
+    credentials = new KubernetesV1Credentials(apiMock, NAMESPACES, [], [], accountCredentialsRepositoryMock, new com.netflix.spinnaker.clouddriver.kubernetes.v2.description.KubernetesSpinnakerKindMap(java.util.Collections.emptyList()))
     namedAccountCredentials = Mock(KubernetesNamedAccountCredentials) {
       getCredentials() >> credentials
     }
--- a/clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/ops/securitygroup/UpsertKubernetesV1SecurityGroupAtomicOperationSpec.groovy
+++ b/clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/ops/securitygroup/UpsertKubernetesV1SecurityGroupAtomicOperationSpec.groovy
@@ -25,6 +25,7 @@ import com.netflix.spinnaker.clouddriver.kubernetes.v1.api.KubernetesApiAdaptor
 import com.netflix.spinnaker.clouddriver.kubernetes.v1.deploy.description.securitygroup.KubernetesIngressTlS
 import com.netflix.spinnaker.clouddriver.kubernetes.v1.deploy.description.securitygroup.KubernetesSecurityGroupDescription
 import com.netflix.spinnaker.clouddriver.kubernetes.v1.security.KubernetesV1Credentials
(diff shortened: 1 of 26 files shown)


clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderSynchronizableSpec.groovy
```diff
@@ -32,28 +32,29 @@ import com.netflix.spinnaker.clouddriver.security.AccountCredentials
 import com.netflix.spinnaker.clouddriver.security.AccountCredentialsRepository
 import com.netflix.spinnaker.clouddriver.security.ProviderVersion
 import com.netflix.spinnaker.kork.configserver.ConfigFileService
+import groovy.transform.CompileStatic
```
maggieneterval: Should we start adding this to any groovy tests we touch?

clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/ops/securitygroup/UpsertKubernetesV1SecurityGroupAtomicOperationSpec.groovy
```diff
@@ -60,7 +60,7 @@ class UpsertKubernetesV1SecurityGroupAtomicOperationSpec extends Specification {
     dockerRegistry = Mock(LinkedDockerRegistryConfiguration)
     dockerRegistries = [dockerRegistry]
     accountCredentialsRepositoryMock = Mock(AccountCredentialsRepository)
-    credentials = new KubernetesV1Credentials(apiMock, NAMESPACES, [], [], accountCredentialsRepositoryMock)
+    credentials = new KubernetesV1Credentials(apiMock, NAMESPACES, [], [], accountCredentialsRepositoryMock, new com.netflix.spinnaker.clouddriver.kubernetes.v2.description.KubernetesSpinnakerKindMap(java.util.Collections.emptyList()))
```
plumpy: import this (weird IntelliJ refactoring artifact, I'm sure)

It happens 9 other places in this ~~CL~~PR too...

clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderSynchronizableSpec.groovy
ezimanyi: It's not always trivial to add, as there are some Groovy constructs that aren't allowed in `@CompileStatic`.  (`@TypeChecked` is an alternative in that case that will still throw type errors if you have them.)

I've actually had some trouble in the past getting `@CompileStatic` to work with tests because I think some of the `Spock` constructs we use don't work well with it.  But in cases where it is easy to add it to the class I'd definitely say go ahead.

I added this because I was getting hard-to-debug errors in the tests and wanted some type-safety to help me out.

clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/ops/securitygroup/UpsertKubernetesV1SecurityGroupAtomicOperationSpec.groovy
ezimanyi: Good catch!  It as indeed an artifact of IntelliJ refactoring.  Fixed.

PR #4024, opened 2019-09-13
refactor(kubernetes): Move V2-only code to clouddriver-kubernetes-v2

ezimanyi:
* refactor(kubernetes): A few minor fixes 

  Remove a few unused variables/imports, and use parametrized types in a few places.

* refactor(kubernetes): Remove SpinnakerKindMap from V1 

  While we've exposed the general SpinnakerKindMap in the shared KubernetesNamedAccountCredentials class, this map is only ever used for the V2 provider. Constructing the map depends on a number of classes that are V2-only, so to better factor the code, we'll just return an empty map for V1 accounts.

* refactor(kubernetes): Split some shared and V2-only code 

  KubernetesSpinnakerKindMap is only used by the V2 provider, but SpinnakerKind is needed by both. In order to support having the V1 provider use SpinnakerKind but not the KubernetesSpinnakerKindMap, move SpinnakerKind to the top level.

  Instead of building all the Kubernetes beans in KubernetesConfiguration, move the construction of V2-only beans to KubernetesV2Configuration.

* refactor(kubernetes): Move V2-only code to clouddriver-kubernetes-v2 

  In order to better control the code sharing between the V1 and V2 kubernetes providers, pull all the (Java-only) code from the V2 provider into its own module. We'll also use the Java compiler instead of the Groovy compiler to compile it.

  In a follow-up PR, I'll move all V1-only code into clouddriver-kubernetes-v1, leaving only shared code in clouddriver-kubernetes. This will make it easier to edit code in the V2 provider without accidentally affecting the V1 provider (and vice versa).


clouddriver-kubernetes-v2/clouddriver-kubernetes-v2.gradle
```diff
@@ -0,0 +1,33 @@
+tasks.compileGroovy.enabled = false
```
maggieneterval: yaaaaaaaay!

clouddriver-kubernetes/clouddriver-kubernetes.gradle
```diff
@@ -1,5 +1,4 @@
 dependencies {
-  implementation project(":clouddriver-artifacts")
```
maggieneterval: whoah I forgot there were no artifacts in v1!

PR #4025, opened 2019-09-14
feat(saga): Add support for retryable tasks within OrchestrationProcessor

robzienert:
Sorry this is big. Wiring up Sagas into the existing Task functionality was more intrusive than I expected it to be. I've kept bodies of work in different commits, which I'll squash when the PR goes in.

1. I combined `OperationsController` and `TaskController`. It was confusing to have two different controllers for this (similar to Orca). The endpoints themselves don't change in location or shape.
2. I moved all of the business logic out of `OperationsController` into an `OperationsService`. I just couldn't deal.
3. Added a `Saga *-1 Task` relationship. The `/task/{id}` endpoint is the entrypoint to the getting Saga state. This required making a lot of boilerplate changes to the `Task` descendants and is a lot of noise on this PR.
4. Added a new `TaskState.FAILED_RETRYABLE` enum. This is what gets used to signal clients if a a task is retryable (shock)!. A new `fun Task.fail(retryable: boolean)` method was added which deprecates the previous `fun Task.fail()` method.
5. Refactored `DefaultOrchestrationProcessor` to understand how to use the new `Task` method; which determines whether an error is retryable either by delegating to an `ExceptionClassifier`. This classifier inspects an exception to determine if it's retryable or not: Exceptions implementing `SpinnakerException` allows developers to fine-tune on a case-by-case basis.


ajordens:
Need to review in more detail — but is there a sensible rollback story of something doesn’t go quite right on the initial rollin?

robzienert:
Should be no problems with a rollback if needed.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/documentation/Nullable.groovy
```diff
@@ -24,8 +24,9 @@ import java.lang.annotation.Target
 /**
  * Marker annotation for documentation purposes. Methods annotated with @Nullable indicate that the method may return a null value.
  *
- *
+ * @deprecated Because why use this and not {@code javax.annotation.Nullable}?
```
dreynaud: 🙃

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AtomicOperationException.java
```diff
@@ -0,0 +1,51 @@
...
+public class AtomicOperationException extends SpinnakerException {
+  public AtomicOperationException(String message, List<String> errors) {
+    super(message);
+    this.errors = errors;
+  }
+
+  @Override
+  public Map<String, Object> getAdditionalAttributes() {
+    if (errors == null || errors.isEmpty()) {
+      return new HashMap<>();
```
dreynaud: is the output supposed to be writable? Otherwise a `Collections.emptyMap()` would do

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/task/Task.groovy
```diff
@@ -82,4 +94,17 @@ public interface Task {
...
+   *
+   * @param sagaId The Saga name/id pair
+   */
+  void connectSaga(@Nonnull SagaId sagaId);
+
+  /**
+   * A list of Sagas associated with this Task, if any.
+   */
+  @Nonnull
+  List<SagaId> getSagaIds();
```
marchello2000: nit: maybe
```suggestion
  List<SagaId> getConnectedSagas();
```
to be symmetrical with `connectSaga`?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/task/jedis/JedisTask.groovy
```diff
@@ -57,12 +69,22 @@ class JedisTask implements Task {
     repository.addToHistory(repository.currentState(this).update(TaskState.COMPLETED), this)
   }
 
+  @Deprecated
```
marchello2000: would be nice to, in a subsequent PR, to just get rid of these - i don't love how we use `@Deprecated` and never actually deprecate

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/task/jedis/RedisTaskRepository.java
```diff
@@ -202,6 +225,11 @@ public void set(String id, JedisTask task) {
     data.put("id", task.getId());
     data.put("startTimeMs", Long.toString(task.getStartTimeMs()));
     data.put("ownerId", task.getOwnerId());
+    try {
+      data.put("sagaIds", mapper.writeValueAsString(task.getSagaIds()));
+    } catch (JsonProcessingException e) {
+      throw new SystemException("Failed to serialise saga ids into Task", e);
```
marchello2000:
```suggestion
      throw new SystemException("Failed to serialize saga ids into Task", e);
```

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/DefaultOrchestrationProcessor.groovy
```diff
@@ -201,4 +212,42 @@ class DefaultOrchestrationProcessor implements OrchestrationProcessor {
...
   }
+
+  @Nonnull
+  private GetTaskResult getTask(String clientRequestId) {
+    def existingTask = taskRepository.getByClientRequestId(clientRequestId)
+    if (existingTask) {
+      if (!isTaskRetryable(existingTask)) {
+        return new GetTaskResult(existingTask, false)
+      }
+      existingTask.updateStatus(TASK_PHASE, "Re-initializing Orchestration Task...")
```
dreynaud: Minorest of nits, but I think this would be better without the `...`

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/DefaultOrchestrationProcessor.groovy
```diff
@@ -201,4 +212,42 @@ class DefaultOrchestrationProcessor implements OrchestrationProcessor {
...
+    }
+    return new GetTaskResult(
+      taskRepository.create(TASK_PHASE, "Initializing Orchestration Task...", clientRequestId),
+      true
+    )
+  }
+
+  private void failTask(@Nonnull Task task, @Nonnull Exception e) {
+    if (task.getSagaIds().isEmpty()) {
+      // Tasks that are not Saga-backed are automatically assumed to be not retryable.
```
dreynaud: ❤️ this comment

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/DefaultOrchestrationProcessor.groovy
```diff
@@ -201,4 +212,42 @@ class DefaultOrchestrationProcessor implements OrchestrationProcessor {
...
+      task.fail(false)
+    } else {
+      task.fail(exceptionClassifier.isRetryable(e))
+    }
+  }
+
+  private boolean isTaskRetryable(Task task) {
```
marchello2000: it feels awkward that this method is on the `.OrchestrationProcessor`... I would expect it to be on the `Task` but it's not really how `Task` is setup... Not very concrete, i know.. thoughts out loud

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/DefaultOrchestrationProcessor.groovy
dreynaud: also minor but I think this would read a bit nicer without the low-level-ish list check. Maybe some sweet sugar like `task.hasSagas()` or something

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/ExceptionClassifierConfigurationProperties.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties("exception-classifier")
+@Data
+public class ExceptionClassifierConfigurationProperties {
+
+  /**
+   * A list of Exception classes that are not retryable within the scope of Saga-backed
+   * orchestrations.
```
marchello2000: nit: maybe specify that this should be the fully qualified name?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
```diff
@@ -0,0 +1,252 @@
...
+    this.atomicOperationsRegistry = atomicOperationsRegistry;
+    this.descriptionAuthorizer = descriptionAuthorizer;
+    this.allowedAccountValidators = allowedAccountValidators.orElse(Collections.emptyList());
+    this.atomicOperationDescriptionPreProcessors =
+        atomicOperationDescriptionPreProcessors.orElse(Collections.emptyList());
+    this.accountCredentialsRepository = accountCredentialsRepository;
+    this.registry = registry;
+    this.objectMapper = objectMapper;
+
+    validationErrorsCounterId = registry.createId("validationErrors");
```
dreynaud: the error counter name could be a little more descriptive? `operationValidationErrors`?

clouddriver-saga/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/flow/SagaAction.kt
```diff
@@ -39,6 +39,17 @@ interface SagaAction<in T : SagaCommand> {
...
+  /**
+   * In the event of an exception being raised from [apply], a [SagaAction] can implement custom error handling logic.
+   *
+   * By default, nothing happens.
+   *
+   * @param command The input [SagaCommand] that was acted on
+   * @param saga The [Saga] state used to apply the [command]
+   * @param exception The resulting exception
+   */
+//  fun recover(command: T, saga: Saga, exception: Exception): Result = Result()
```
marchello2000: intended to be commented out?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
```diff
@@ -0,0 +1,252 @@
...
+  @Nonnull
+  public List<AtomicOperation> collectAtomicOperations(
+      @Nullable String cloudProvider, @Nonnull List<Map<String, Map>> inputs) {
+    List<AtomicOperationBindingResult> results = convert(cloudProvider, inputs);
+
+    List<AtomicOperation> atomicOperations = new ArrayList<>();
+    results.forEach(
+        bindingResult -> {
+          if (bindingResult.errors.hasErrors()) {
+            throw new DescriptionValidationException(bindingResult.errors);
```
dreynaud: without being a language lawyer I'm not quite sure what the expected behavior is here, throwing exceptions in a lambda in a `forEach`? Are we expecting an exception for each result that has an error? Are we getting out on the first error?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
```diff
@@ -0,0 +1,252 @@
...
+        AccountCredentials credentials = accountCredentialsRepository.getOne(accountName);
+        return credentials.getProviderVersion();
+      } catch (Exception e) {
+        log.warn("Unable to determine provider version for account {}", accountName, e);
+      }
+    }
+    return ProviderVersion.v1;
+  }
+
+  private static Map processDescriptionInput(
```
dreynaud: that's some unusually fancy magic here, can you throw in a comment explaining what is going on for the rest of us?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
```diff
@@ -0,0 +1,252 @@
...
+    Class<?> convertDescriptionReturnType =
+        ResolvableType.forMethodReturnType(convertDescriptionMethod).getRawClass();
+
+    for (AtomicOperationDescriptionPreProcessor preProcessor : descriptionPreProcessors) {
+      if (preProcessor.supports(convertDescriptionReturnType)) {
+        descriptionInput = preProcessor.process(descriptionInput);
+      }
+    }
+
+    return descriptionInput;
```
dreynaud: updated at every step of the loop so we just return the last one?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/task/jedis/JedisTask.groovy
robzienert: Oh, that's definitely the plan! I actually delete `@Deprecated` 💯💯💯

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/DefaultOrchestrationProcessor.groovy
robzienert: Agreed. I've been removing those from other places but missed these. Thanks!

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
robzienert: This isn't new code, I just moved it out from the controller, so I don't want to change the metric names, even if they're pretty rubbish. :(

We probably should have a plan around how to change metric names, though.

clouddriver-saga/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/flow/SagaAction.kt
robzienert: Yes, this behavior isn't written yet / haven't decided the specific contract yet.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
robzienert: We're getting out on the first error. Changing this logic would alter backwards compatibility. Again, this was originally in a controller, so it was just a lift/shift.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
robzienert: No, the preprocessors mutate the map in a chain. It's the same map the whole way through, for better or worse.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/task/Task.groovy
robzienert: I went the other way and just got rid of the `connect` taxonomy. It was a silly thing.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
dreynaud: oh :/

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
dreynaud: ooh I see it's on both sides of the expression

PR #4029, opened 2019-09-16
refactor(kubernetes): Reduce API surface of KubernetesKindProperties

ezimanyi:
* refactor(kubernetes): Reduce API surface of KubernetesKindProperties 

  Only the statically-defined kinds are allowed to set isDynamic to false, or hasClusterRelationship to true. In order to make this statically true, make the full constructor private and expose a factory method that doesn't allow those properties to be set for use outside the class.

  (Some of the usages of the constructor in tests *did* set isDynamic and hasClusterRelationship, but these were just testing the API that was available; now that these are not exposed outside the class the tests can just use the factory method.)

* refactor(kubernetes): Simplify getRegisteredKinds 

  The only time we ever use the getRegisteredKinds method in KubernetesKindRegistry, what we actually want is a list of the global kinds (so we can set up the core caching agent).

  We do this by getting all the kinds and filtering on the isDynamic property; it would be simpler to just return the global kinds in the first place. A kind is global if and only if it is not dynamic.

* refactor(kubernetes): Remove isDynamic property from kind properties 

  The prior commit removed the last usage of isDynamic in the KubernetesKindProperties class. Remove the member variable from the class.

* refactor(kubernetes): Move defaulting logic to account registry 

  In preparation for a fix to reduce live CRD lookups, move the defaulting of unknown kinds from the GlobalKubernetesKindRegistry to the account KubernetesKindRegistry.

  This doesn't change the result of calling getRegisteredKind on the account registry, but moves where the defaulting happens.

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesCoreCachingAgent.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesCoreCachingAgent.java
@@ -62,8 +62,7 @@ public class KubernetesCoreCachingAgent extends KubernetesV2OnDemandCachingAgent
 
   @Override
   protected List<KubernetesKind> primaryKinds() {
-    return credentials.getKindRegistry().getRegisteredKinds().stream()
-        .filter(k -> !k.isDynamic())
+    return credentials.getKindRegistry().getGlobalKinds().stream()
         .map(KubernetesKindProperties::getKubernetesKind)
         .filter(credentials::isValidKind)
         .collect(Collectors.toList());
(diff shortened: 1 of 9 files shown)


ezimanyi:
@maggieneterval : Good question, here are the definitions of the relevant terms:
**global kinds**: The set of kinds built-in to Spinnaker (ie, the static entries in `KubernetesKind.java`) that exist across all accounts.
**registered kinds**: Any kind that is in a registry (so the definition depends on the actual registry we're using).  For the global registry this would just be global kinds; for an account-specific registry it would also include kinds configured in the config for the account.
**dynamic kinds**: Kinds that are not global kinds (as you've said)

I've found these terms pretty confusing and so am slowly trying to get rid of them; this PR finally gets rid of the concept of a *dynamic* kind.

maggieneterval:
@ezimanyi thank you! :pray:

PR #4030, opened 2019-09-16
Escape label names when attaching/detaching a replicaset from a service

jzinkweg:
Fixes spinnaker/spinnaker#4893.

Implements escaping of '~' and '/' characters when used in labels as defined in RFC 6901.
https://tools.ietf.org/html/rfc6901#section-3

Allows for the use of labels such as "app.kubernetes.io/name" and "app.kubernetes.io/instance" in the service selector.

### DIFF
--- a/clouddriver-kubernetes-v2/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesServiceHandlerSpec.groovy
+++ b/clouddriver-kubernetes-v2/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesServiceHandlerSpec.groovy
@@ -1,0 +1,133 @@
+/*
+ * Copyright 2019 Air France-KLM Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


jzinkweg:
I wasn't sure where to leave the utility code to escape the label.
It looks like this scenario only applies to the KubernetesServiceHandler for now but adding it to JsonPatch allows for reuse.

clouddriver-kubernetes-v2/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesServiceHandlerSpec.groovy
```diff
@@ -0,0 +1,133 @@
+/*
+ * Copyright 2017 Google, Inc.
```
ezimanyi: This should be 2019, and you/your company.

jzinkweg: Thanks, fixed!

PR #4031, opened 2019-09-16
Add more details in call reason header for Titus

aravindmd:
- Adding human readable details to `X-Titus-CallReason` header that is passed along to Titus . This was an ask from Titus which can help then with internal debugging.


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/TitusClientAuthenticationUtil.java
```diff
@@ -20,6 +20,8 @@
 import io.grpc.Metadata;
 import io.grpc.stub.AbstractStub;
 import io.grpc.stub.MetadataUtils;
+import org.apache.commons.lang.StringUtils;
```
emjburns: I think you need to use `commons.lang3`

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/TitusClientAuthenticationUtil.java
```diff
@@ -33,7 +35,21 @@
...
+        String.format(
+            "Invoked by Spinnaker execution %s, Application name %s",
+            getSpinnakerExecutionId(
+                AuthenticatedRequest.getSpinnakerExecutionId().orElse("unknown")),
+            MDC.get(AuthenticatedRequest.Header.APPLICATION.getHeader())));
     return serviceStub.withInterceptors(MetadataUtils.newAttachHeadersInterceptor(metadata));
   }
+
+  public static String getSpinnakerExecutionId(String executionIdHeader) {
+    if (!executionIdHeader.equalsIgnoreCase("unknown")) {
```
emjburns: Probably worth also making sure this is not empty in this condition, so that the split + `ids[0]` won't fail with an NPE.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/TitusClientAuthenticationUtil.java
```diff
@@ -33,7 +35,21 @@
...
+        String.format(
+            "Invoked by Spinnaker execution %s, Application name %s",
+            getSpinnakerExecutionId(
+                AuthenticatedRequest.getSpinnakerExecutionId().orElse("unknown")),
+            MDC.get(AuthenticatedRequest.Header.APPLICATION.getHeader())));
     return serviceStub.withInterceptors(MetadataUtils.newAttachHeadersInterceptor(metadata));
   }
+
+  public static String getSpinnakerExecutionId(String executionIdHeader) {
+    if (!executionIdHeader.equalsIgnoreCase("unknown") && !executionIdHeader.isEmpty()) {
```
robzienert:
```suggestion
    if (!"unknown".equalsIgnoreCase(executionIdHeader) && !executionIdHeader.isEmpty()) {
```

Protect against NPEs! Also, would recommend adding `@Nonnull` or `@Nullable` on public method return types & arguments.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/TitusClientAuthenticationUtil.java
```diff
@@ -33,7 +35,21 @@
...
+      String[] ids = StringUtils.split(executionIdHeader, ":");
```
robzienert: I'd recommend using `Splitter` instead.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/TitusClientAuthenticationUtil.java
```diff
@@ -33,7 +37,21 @@
   public static <STUB extends AbstractStub<STUB>> STUB attachCaller(STUB serviceStub) {
     Metadata metadata = new Metadata();
     metadata.put(CALLER_ID_KEY, AuthenticatedRequest.getSpinnakerUser().orElse("spinnaker"));
-    metadata.put(CALL_REASON_KEY, AuthenticatedRequest.getSpinnakerExecutionId().orElse("unknown"));
+    metadata.put(
+        CALL_REASON_KEY,
+        String.format(
+            "Invoked by Spinnaker execution %s, Application name %s",
+            getSpinnakerExecutionId(
+                AuthenticatedRequest.getSpinnakerExecutionId().orElse("unknown")),
```
marchello2000: nit: sort of defeats the purpose of `nullables` i suppose, but i think this would be more readable as `orElse(null)`

then `getSpinnakerExecutionId` below would read simpler as 
```
if (Strings.isNullOrEmpty(executionIdHeader) {
  return "unknown";
}

return Splitter.on(':').splitToList(executionIdHeader).get(0);
```

marchello2000: never mind my comment as i see you followed, Robs :)

PR #4032, opened 2019-09-16
feat(sagas): Add resume task endpoint

robzienert:
Adds a new controller endpoint that will be called by Orca when a Task needs to be retried. 

We opted for making a specific endpoint for this behavior, although re-submitting to the operations endpoint _would_ work as well, provided the input description & client request id is exactly the same. A new endpoint was chosen so that Orca doesn't need to remember what it sent across the wire.

 Due to how `AtomicOperation`s are created _(see below)_, there was no context available on resume to actually re-construct the `AtomicOperation`. So I had to add a new interface `SagaContextAware` that `AtomicOperation` classes can implement to get the necessary request context down to Sagas. Since this wiring is pretty boilerplate-y, I modified the Saga lib to allow some more framework-y use cases, like adding a new `SnapshotAtomicOperationInputs` step to the beginning of all `AtomicOperation`s backed by sagas. Having the full request state let's us look up the correct information to synthesize the original request.


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/OperationsController.groovy
```diff
@@ -122,16 +123,21 @@ class OperationsController {
...
+  StartOperationResult resumeTask(@PathVariable("id") String id) {
     Task t = taskRepository.get(id);
     if (t == null) {
       throw new NotFoundException("Task not found (id: $id)")
     }
 
-    // TODO(rz): Check if task is failed: Only allow resuming failed tasks
-    // TODO(rz): Lookup saga
+    if (!t.status.retryable) {
+      throw new ConstraintViolationException("Task is not retryable").with {
```
cfieber: I think you could `.also` here instead of `.with` ( and not need the `it` line )

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/task/DefaultTask.groovy
```diff
@@ -145,6 +150,8 @@ class TaskDisplayStatus implements Status {
   Boolean isRetryable() {
     return taskStatus.isRetryable()
   }
+
+
```
dreynaud: very newline-y

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/OperationsController.groovy
robzienert: `.also` isn't available in Groovy, is it?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/task/DefaultTask.groovy
robzienert: Stop pointing out my tricks. Line count == more compensation.

PR #4033, opened 2019-09-17
refactor(kubernetes): Look up CRD details on-demand

ezimanyi:
* refactor(kubernetes): Remove useless call to register kind 

  On account creation, we iterate over all kinds in the account and call registerKind. Since we only have the name of the kind and no other information, we just assume the kind is namespaced and pass true.

  Since the kind registry will return a set of default properties
  (including isNamespaced set to true) for something not in the registry, skip the needless call to register these kinds with default properties.

* refactor(kubernetes): Pass custom resources to registry constructor 

  Each account has a set of custom resources that are statically defined in the account config. We register each of these into the KubernetesKindRegistry, but currently do so in a somewhat opaque way, by depending on a side-effect of the KubernetesResourceProperties.fromCustomResource function we use when creating a different property on the account.

  Remove the side effect from the fromCustomResource function and instead pass the list of custom resources to the constructor for KubernetesKindRegistry, allowing it to register resources there.

* refactor(kubernetes): Use ObjectMapper to parse CRD 

  The current logic for reading CRD details is full of unsafe casts and is hard to read. Define a class that has the fields we expect on a CRD and use ObjectMapper to convert the manifest back from the cluster into a typed object.

* refactor(kubernetes): Allow KubernetesKindRegistry to look up CRDs 

  To avoid needing to pre-register all CRDs into the kind registry, the registry should be able to look up kind information in the cluster if it comes across a kind that it does not have any information about.

  When creating a KubernetesKindRegistry, the caller can provide a Function<KubernetesKind, Optional<KubernetesKindProperties>>; when looking up a kind that is not in the registry, the registry will call that function to try to resolve the kind properties. If the function returns a result, the properties are recorded in the registry
  (so subsequent lookups of the same kind don't call the function again). If the function returns an empty Optional, we return a set of default properties as before.

  This commit does not actually use the new functionality; this will come in a later commit.

* refactor(kubernetes): Look up CRD details on-demand 

  Create a new function getCrdProperties that on-demand looks up the properties of a CRD and pass this to the constructor for the kind registry. This removes the need to pre-register kinds that are fround when listing CRDs as the registry will look up any kinds it needs, and also removes the last remaining side-effect of calling liveCrdSupplier.

  The only call to getCrds is from the unregistered CRD caching agent. Given that this happens once per caching cycle (along with a lot of other lookups) there's no real benefit to memoizing the result with a timed expiration. Remove the memoization and just have getCrds do a live lookup.

* refactor(kubernetes): Rename getRegisteredKind to getKindProperties 

  The term registered is confusing, and also somewhat redundant given that the method is on a kind regsitry. (It's also not entirely correct as the function can also live lookup properties and register them.)

  Let's rename the function getKindProperties to be more in line with what it does.

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKindRegistry.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKindRegistry.java
@@ -17,8 +17,8 @@
 package com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest;
 
 import com.google.common.collect.ImmutableCollection;
+import com.google.common.collect.ImmutableList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
@@ -110,7 +110,7 @@ public class KubernetesKindRegistry {
     @Nonnull
     public KubernetesKindRegistry create() {
       return new KubernetesKindRegistry(
-          globalKindRegistry, k -> Optional.empty(), Collections.emptyList());
+          globalKindRegistry, k -> Optional.empty(), ImmutableList.of());
     }
   }
 }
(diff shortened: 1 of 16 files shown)


clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesKindRegistry.java
```diff
@@ -17,55 +17,73 @@
...
+              "Dynamically registering {}, (namespaced: {})",
+              kindProperties.getKubernetesKind().toString(),
+              kindProperties.isNamespaced());
+          return kindProperties;
+        });
   }
 
   /**
-   * Searches the registry for a {@link KubernetesKindProperties} with the supplied {@link
+   * Searches the the registry for a {@link KubernetesKindProperties} with the supplied {@link
```
maggieneterval:
```suggestion
   * Searches the registry for a {@link KubernetesKindProperties} with the supplied {@link
```

PR #4041, opened 2019-09-18
fix(kubernetes): Fix credentials endpoint with unreachable cluster

ezimanyi:
Fixes spinnaker/spinnaker#4407.  See #3770 for prior work on this.

* refactor(kubernetes): Make namespaces immutable 

  The namespaces and omitNamespaces fields on KubernetesV2Credentials never change; let's enforce that by making them immutable.

* fix(kubernetes): Fix credentials endpoint with unreachable cluster 

  Calling the endpoint to list credentials currently calls getNamespaces on KubernetesNamedAccountCredentials, which in turn calls getDeclaredNamespaces on the account.

  Because of the caching logic, this sometimes results in a live call (if the current entry is expired) which can cause the entire endpoint to hang. We should always serve this call out of the cache (and handle any refreshes asynchronously) so that we never hang on this call. I think a slight increase in stale reads here is prefereable to blocking the endpoint synchronously getting live data.

  The other place this call is used is when caching agents decide which namespaces they are responsible for caching. I don't think this change will lead to more stale reads because the caching agent runs at regular intervals, so we'll be regularly reading the namespaces and thus triggering async refreshes of the cached data. (Performance-wise it also would probably be fine to just always read live in this case, but for simplicity I'm having this call also go through the cache.)

  I think ideally we would not have the credentials endpoint get the namespaces in every account. My understanding is that this is primarily used by deck to populate namespace dropdowns; we could probably instead fetch the namespaces for a given account on-demand once the account has been selected and remove the need to return them as part of the credentials endpoint. With that change we could potentially completely remove the caching here, but for now this should improve the stability of clouddriver when an account is unreachable.

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
@@ -90,9 +90,9 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
 
   @Include private final String accountName;
 
-  @Include @Getter private final List<String> namespaces;
+  @Include @Getter private final ImmutableList<String> namespaces;
 
-  @Include @Getter private final List<String> omitNamespaces;
+  @Include @Getter private final ImmutableList<String> omitNamespaces;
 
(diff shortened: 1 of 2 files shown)


ezimanyi:
@maggieneterval : Let's definitely chat about adding a new endpoint for getting namespaces per cluster when you're working on that component---I think that's probably the idea fix for this issue.

PR #4044, opened 2019-09-23
fix(titus): Do not automatically assume all Titus exceptions are retryable

robzienert:


clouddriver-titus/src/test/groovy/com/netflix/spinnaker/clouddriver/titus/TitusExceptionSpec.groovy
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.clouddriver.titus
+
+import spock.lang.Specification
+
+class TitusExceptionSpec extends Specification {
+
+  def "should set retryable to false on invalid arguments"() {
```
jonsie: Maybe add a small test here of a message that does not start with `INVALID_ARGUMENT`.

PR #4049, opened 2019-09-24
feat(artifacts/gitRepo): add support for git repo

ethanfrogers:
adds support for a git repo artifact type which will clone a given git
repo, package it up as a tar.gz and return the value. callers of this
artifact type should expect a .tar.gz and act accordingly, decompressing
if necessary. this impl. doesn't include support for cloning the
repository at a particular history depth as jgit doesn't support that
option. the alternative is to perform git operations by shelling out to
git but i opted not to do that because it makes authentication much more
difficult than it is _with_ jgit.

### DIFF
--- a/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
+++ b/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
@@ -1,0 +1,150 @@
+/*
+ * Copyright 2019 Armory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.artifacts.gitRepo;
+
+import com.netflix.spinnaker.clouddriver.artifacts.config.ArtifactCredentials;
+import com.netflix.spinnaker.kork.artifacts.model.Artifact;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.eclipse.jgit.api.ArchiveCommand;
+import org.eclipse.jgit.api.CheckoutCommand;
+import org.eclipse.jgit.api.CloneCommand;
+import org.eclipse.jgit.api.Git;
+import org.eclipse.jgit.archive.ArchiveFormats;
+import org.eclipse.jgit.transport.UsernamePasswordCredentialsProvider;
+
+@Slf4j
+public class GitRepoArtifactCredentials implements ArtifactCredentials {
+  @Getter private final List<String> types = Collections.singletonList("git/repo");
+
+  @Getter private final String name;
+  private final String username;
+  private final String password;
+
+  public GitRepoArtifactCredentials(GitRepoArtifactAccount account) {
+    this.name = account.getName();
+    this.username = account.getUsername();
+    this.password = account.getPassword();
+  }
+
+  @Override
+  public InputStream download(Artifact artifact) throws IOException {
+    String repoReference = artifact.getReference();
+    Path stagingPath =
+        Paths.get(System.getProperty("java.io.tmpdir"), UUID.randomUUID().toString());
+
+    try {
+      log.info("Cloning git/repo {} into {}", repoReference, stagingPath.toString());
+      Git localRepository = clone(artifact, stagingPath);
+      ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+      log.info("Creating archive for git/repo {}", repoReference);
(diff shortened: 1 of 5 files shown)


ethanfrogers:
I plan on following this PR up with support for SSH based authentication but wanted to get something small in review first. 

I should point that this code is very similar to the `AppengineGitRepositoryClient` and we should consider consolidating it at some point in the future. I don't know App Engine at all so I'm probably not the best person to make those changes.

dbyron0:
@ethanfrogers  Do you mind mentioning the other PRs you have queued up to use this?

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
```diff
@@ -0,0 +1,150 @@
...
+              .setFormat("tgz")
+              .setOutputStream(outputStream);
+
+      if (!StringUtils.isEmpty(subPath)) {
+        archiveCommand.setPaths(subPath);
+      }
+
+      archiveCommand.call();
+    } finally {
+      ArchiveFormats.unregisterAll();
```
ezimanyi: I think calling `registerAll()` followed by `unregisterAll()` each time we archive a repo is going to introduce threading bugs, as this is updating static variables that are used by all threads.  I suspect we really just want to call `registerAll()` (or even a more specific register of `tgz`) just once on startup.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
```diff
@@ -0,0 +1,150 @@
...
+      return new ByteArrayInputStream(outputStream.toByteArray());
+    } catch (Exception e) {
+      throw new IOException("Failed to clone git/repo " + repoReference, e);
+    } finally {
+      log.info("Cleaning up git/repo {}", repoReference);
+      try {
+        FileUtils.deleteDirectory(stagingPath.toFile());
+      } catch (IOException e) {
+        log.info("Failed to cleanup git/repo for {}", repoReference);
+        throw new RuntimeException(
```
ezimanyi: In general it's not great to throw exceptions from within a `finally` block, primarily because it will mask any exception that happened in the `try` block...so if we fail to create the archive at all and throw an exception `Failed to clone git/repo...` then also fail to delete it there won't be any trace left of the original exception.

One workaround would be to use a try-with-resources construct:
```
try (AutoCloseable ignored = () -> FileUtils.deleteDirectory(stagingPath.toFile()))
```
and completely eliminate the `finally`.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
```diff
@@ -0,0 +1,150 @@
...
+        Paths.get(System.getProperty("java.io.tmpdir"), UUID.randomUUID().toString());
+
+    try {
+      log.info("Cloning git/repo {} into {}", repoReference, stagingPath.toString());
+      Git localRepository = clone(artifact, stagingPath);
+      ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+      log.info("Creating archive for git/repo {}", repoReference);
+      archiveToOutputStream(artifact, localRepository, outputStream);
+      return new ByteArrayInputStream(outputStream.toByteArray());
+    } catch (Exception e) {
```
ezimanyi: It would be better to catch a more specific exception here (which may be possible in light of the comment below on declaring something more specific in the `throws` clauses).

ethanfrogers:
@dbyron0 the initial use case for this is to clean up the Kustomize implementation which will involve work in Deck, Rosco and (I think) Orca. Is there anything specific you're looking for?

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
ethanfrogers: Ah, glad you called that out. I've removed the calls to `register/unregister` and added a call to register the single `tgz` format when the credentials are constructed. It will get re-registered for every account but that seems like a fine tradeoff to me. The alternative is that the `JGit` implementation leaks into other areas of this. I tried to register only if unregistered but there isn't an API for determining what types are already registered.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
ethanfrogers: So, I think we'll still have the same problem using try-with-resources because `AutoCloseable.close` throws an exception which will need to be handled. In either case, to prevent the the main exception from being swallowed I think we'll just log and move on. I've taken a stab at implementing this. LMK what you think.

ezimanyi: One of the [advantages of try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) is that it will suppress the exception from `.close()` if (and only if) the `try` block threw an exception.  Always suppressing and logging the exception is also reasonable, but that will swallow the exception in cases where the `try` completes normally and we fail to close.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
ezimanyi: It looks like the `registerFormat` does reference counting, so maybe we don't want to create one on every account...perhaps this would be better in a static initializer?  Though honestly I think it's probably fine where it is too.

The javadoc suggests calling `registerFormat` then `unregisterFormat` on each use, but it definitely seems that it's not threadsafe so I think what you have here or a static initializer is definitely better.

dbyron0:
@ethanfrogers Hoping for the ability to use this kind of artifact for baking helm charts, and a way to populate them from a github trigger.

ethanfrogers:
@dbyron0 Good to know. I've heard that request from a couple of folks as well. I'll make a note of that and see if we can fit it in!

dbyron0:
@ethanfrogers Thank you!

PR #4051, opened 2019-09-24
fix(gce): Fix an NPE in AbstractGoogleServerGroupCachingAgent

plumpy:


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/AbstractGoogleServerGroupCachingAgent.java
```diff
@@ -768,25 +767,30 @@ private void setSourceImage(
...
-      jenkinsInfo.put("name", appversion.getBuildJobName());
-      jenkinsInfo.put("number", appversion.getBuildNumber());
+      ImmutableMap.Builder<String, String> jenkinsInfo = ImmutableMap.builder();
+      putIfNotNull(jenkinsInfo, "name", appversion.getBuildJobName());
+      putIfNotNull(jenkinsInfo, "number", appversion.getBuildNumber());
       if (tags.containsKey("build_host")) {
         jenkinsInfo.put("host", tags.get("build_host"));
       }
-      buildInfo.put("jenkins", jenkinsInfo);
+      buildInfo.put("jenkins", jenkinsInfo.build());
```
ezimanyi: Should we do the same for `build_host` and `build_info_url` to guard against the `tags` (mutable) map containing these keys but mapped to `null`?  (And also just to make the logic consistent, as right now it's not clear why these two use different logic.)

plumpy: Yeah, okay, good call. I switched to a slightly different approach, building into a `HashMap` and then copying to an `ImmutableMap` at the end. It's still annoying but better, I think.

Thanks...

ezimanyi: :+1: new approach looks good!

PR #4052, opened 2019-09-25
feat(artifacts/gitRepo): support SSH auth

ethanfrogers:
adds support for SSH authentication and also improves UX around what
happens if the users supplies a reference that isn't supported by the
auth type.

### DIFF
--- a/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
+++ b/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
@@ -63,6 +63,14 @@ public class GitRepoArtifactCredentials implements ArtifactCredentials {
   private final String sshPrivateKeyPassphrase;
   private final String sshKnownHostsFilePath;
   private final boolean sshTrustUnknownHosts;
+  private final AuthType authType;
+
+  private enum AuthType {
+    HTTP,
+    TOKEN,
+    SSH,
+    NONE
+  }
 
   public GitRepoArtifactCredentials(GitRepoArtifactAccount account) {
     this.name = account.getName();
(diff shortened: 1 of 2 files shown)


clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
```diff
@@ -127,12 +180,62 @@ private String artifactVersion(Artifact artifact) {
...
     }
+  }
 
-    return cloneCommand;
+  private CloneCommand configureSshAuth(CloneCommand cloneCommand) {
+    SshSessionFactory sshSessionFactory =
+        new JschConfigSessionFactory() {
+          @Override
+          protected void configure(OpenSshConfig.Host hc, Session session) {
+            if (sshKnownHostsFilePath == null && sshTrustUnknownHosts) {
```
ezimanyi: Is there a use case for trusting unknown hosts even if you provide a known hosts file?

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
```diff
@@ -127,12 +180,62 @@ private String artifactVersion(Artifact artifact) {
...
+            }
+          }
+
+          @Override
+          protected JSch createDefaultJSch(FS fs) throws JSchException {
+            JSch defaultJSch = super.createDefaultJSch(fs);
+            defaultJSch.addIdentity(sshPrivateKeyFilePath, sshPrivateKeyPassphrase);
+
+            if (sshKnownHostsFilePath != null && sshTrustUnknownHosts) {
+              log.warn(
```
ezimanyi: It feels like this warning could be closer to where we actually ignore the value of `sshTrustUnknownHosts` (ie, in `confgure`).  (Or maybe that gets called too often to log...)

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
```diff
@@ -44,11 +58,41 @@
...
+    this.sshPrivateKeyFilePath = account.getSshPrivateKeyFilePath();
+    this.sshPrivateKeyPassphrase = account.getSshPrivateKeyPassphrase();
+    this.sshKnownHostsFilePath = account.getSshKnownHostsFilePath();
+    this.sshTrustUnknownHosts = account.isSshTrustUnknownHosts();
+
+    if (!StringUtils.isEmpty(username) && !StringUtils.isEmpty(password)) {
+      authType = AuthType.HTTP;
+    } else if (!StringUtils.isEmpty(token)) {
+      authType = AuthType.TOKEN;
+    } else if (!StringUtils.isEmpty(sshPrivateKeyFilePath)
```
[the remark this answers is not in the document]
ethanfrogers: Possibly? I copied all of this logic from the App Engine logic so I assume most cases have been accounted for since it's been around longer. If cases like this come up I don't see why we couldn't just relax this condition in a patch.

PR #4053, opened 2019-09-25
fix(titus): Fixing titus instance id search

aravindmd:
- It turns out titus instances where not getting cached in the `instances` in memory cache , this was due to the fact that `TitusStreamingUpdateAgent` was not implementing `getProvidedDataTypes` method 
Refer here for more context : https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/SearchableProvider.java#L91


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
```diff
@@ -159,6 +159,16 @@ public String getProviderName() {
...
   }
 
+  @Override
+  public Collection<AgentDataType> getProvidedDataTypes() {
+    return TYPES;
+  }
+
+  @Override
+  public CacheResult loadData(ProviderCache providerCache) {
+    return null;
```
asher: I'd throw an explicit exception here instead. This method shouldn't get called, but if it did, would result in an odd NPE further up the stack.

PR #4067, opened 2019-10-03
fix(cf): when scaling cf disabled SG we should use max capacity

Jammy-Louie:
When scaling a disabled server group in cloud foundry we should use the "max" value because a disabled stated in CF would be a stopped state which would result in 0 instances running but we should still adjust the number instances to the target number if the server group is enabled/started.

fix for spinnaker/spinnaker#4786


Jammy-Louie:

PR #4070, opened 2019-10-03
fix(titus): Fix SagaContext wiring in DeployHandler cases

robzienert:
It turns out that TitusDeployHandler (and other DeployHandler classes) are singleton components
that get injected into prototype AtomicOperation components, which I didn't realize until now.
The issue with this is that I was setting request-scoped context into a singleton bean, which
could lead to some neat concurrency bugs. This change modifies things for DeployAtomicOperation
specifically to inject the SagaContext into the description (which is actually request-scoped)
rather than passing through to the TitusDeployHandler object.

There's also this neat feature I discovered while working through this where we convert the raw request input into an atomic operation description twice, throwing away the first conversion after we've validated the input. Not sure why we do this, but it seems like something we could make better someday.


aravindmd:
good to know , thanks for looping me in !

PR #4071, opened 2019-10-03
feat(retry): Resilience4j Spring Boot2 implementation in clouddriver-sql

jonsie:
If we go this route, I would like to see the work to back Spectator with a Micrometer registry get resurrected, so we can surface some of these Resilience4j metrics easily (example PR: https://github.com/spinnaker/kork/pull/268)

### DIFF
--- a/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/sql.kt
+++ b/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/sql.kt
@@ -15,15 +15,11 @@
  */
 package com.netflix.spinnaker.clouddriver.sql
 
-import com.netflix.spinnaker.kork.core.RetrySupport
-import com.netflix.spinnaker.kork.sql.config.RetryProperties
-import io.github.resilience4j.retry.Retry
+import io.github.resilience4j.retry.annotation.Retry
 import org.jooq.DSLContext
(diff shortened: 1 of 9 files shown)


cfieber:
I think we should already be able to get the metrics (https://github.com/spinnaker/kork/blob/master/kork-core/src/main/java/com/netflix/spinnaker/kork/metrics/SpectatorMeterRegistry.java#L28) - the work I was considering was inverting that relationship in kork but we should be bridging those frameworks already

clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/sql.kt
```diff
@@ -34,36 +30,29 @@ internal val taskStatesFields = listOf("id", "task_id", "created_at", "state", "
...
-  }
+@Retry(name = "sqlRead")
+internal fun <T> DSLContext.read(fn: (DSLContext) -> T): T {
+  return fn(this)
 }
 
 /**
- * Run the provided [fn] with retry support.
+ * Run the provided [fn], retrying on failures using resilience4j.retry.instances.sqlDelete configuration.
+ * TODO(csmalley):  Investigate if usage of this is necessary, or if delete calls can be transactional
```
robzienert: Yeah, we can use `transactional` for deletes. It's really "read" vs "write" connections (replicas vs master).

PR #4078, opened 2019-10-04
fix(kubernetes): Fix matching of artifacts in deploy/patch stages

ezimanyi:
Fixes spinnaker/spinnaker#4956

When determining whether the deploy and patch manifest stages bound all required artifacts, the artifacts in the list of required artifacts have an artifact account set while those in the list we're trying to bind don't.

As the .equals function of Artifact includes the account in its comparison, this means that we don't recognize the replaced artifacts as the ones we expected to replace and fail the stage.

Trying to fix this more generally is riskier than I'd like to cherry pick, so as a temporary fix, define a class ArtifactKey that contains the fields we'd like to use when comparing artifacts and use that class for the comparison.

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/ArtifactKey.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/ArtifactKey.java
@@ -27,6 +27,19 @@ import javax.annotation.Nullable;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
 
+/**
+ * When determining whether the deploy and patch manifest stages bound all required artifacts, the
+ * artifacts in the list of required artifacts have an artifact account set while those in the list
+ * we're trying to bind don't.
+ *
+ * <p>As the .equals function of Artifact includes the account in its comparison, this means that we
+ * don't recognize the replaced artifacts as the ones we expected to replace and fail the stage.
+ *
+ * <p>As a temporary fix until we can refactor the artifact passing code to consistently include (or
+ * not) account, or decide that account should always be excluded from Artifact.equals(), create a
+ * class to hold the fields of Artifact that these two stages should use when deciding whether
+ * artifacts are equal.
+ */
 @EqualsAndHashCode
 @ToString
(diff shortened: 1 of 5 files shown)


clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/ArtifactKey.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Collection;
+import java.util.Objects;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+@EqualsAndHashCode
+@ToString
+class ArtifactKey {
```
plumpy: maybe paste some of your commit message as Javadoc?

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/KubernetesDeployManifestOperation.java
```diff
@@ -130,8 +131,10 @@ public OperationResult operate(List _unused) {
       boundArtifacts.addAll(replaceResult.getBoundArtifacts());
     }
 
-    Set<Artifact> unboundArtifacts = new HashSet<>(requiredArtifacts);
-    unboundArtifacts.removeAll(boundArtifacts);
+    Set<ArtifactKey> unboundArtifacts =
+        Sets.difference(
```
maggieneterval: This method is fancy! I love it!

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/ArtifactKey.java
ezimanyi: Good :telephone_receiver: ; done.

ezimanyi:
@spinnakerbot :cherries: :pick: 1.16

ezimanyi:
Manual cherry pick in #4079

PR #4081, opened 2019-10-07
fix(titus): Serde fix for SagaContext

robzienert:
I tested this and it works.


clouddriver-titus/src/test/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/description/TitusDeployDescriptionSpec.groovy
```diff
@@ -13,12 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.netflix.spinnaker.clouddriver.titus.deploy.descriptions
+package com.netflix.spinnaker.clouddriver.titus.deploy.description
```
robzienert: This was the wrong package, just moving it to the right place.

robzienert:
Lombok was causing troubles with using `@JsonIgnore`, so I just set the methods explicitly.

PR #4089, opened 2019-10-14
feat(dynamicRetry): Pull comma-delimited list of non-retryable exceptions from dynamic config store

jonsie:
~~No tests exist for this class, maybe I should write a test before we merge this.~~

Wrote a test.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/ExceptionClassifier.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/ExceptionClassifier.java
@@ -15,10 +15,18 @@
  */
 package com.netflix.spinnaker.clouddriver.orchestration;
 
+import com.google.common.base.Splitter;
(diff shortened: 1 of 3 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/ExceptionClassifier.java
```diff
@@ -29,16 +36,35 @@
...
   }
 
   /** Returns whether or not a given Exception is retryable or not. */
   public boolean isRetryable(@Nonnull Exception e) {
     if (e instanceof SpinnakerException) {
       return Optional.ofNullable(((SpinnakerException) e).getRetryable()).orElse(false);
     }
-    return !properties.getNonRetryableClasses().contains(e.getClass().getName());
+
+    String dynamicNonRetryableClasses =
```
marchello2000: i saw wrap in a try/catch for extra defensiveness and return `false` if things fail?

PR #4090, opened 2019-10-15
fix(titus): Adjust retries on upsert scaling policy operation

jeyrschabu:
- wrap a retry on scaling policy create call for good measure
- fix retry count/timeout for checking if the scaling policy was applied
- make retry to be exponential

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/ops/UpsertTitusScalingPolicyAtomicOperation.groovy
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/ops/UpsertTitusScalingPolicyAtomicOperation.groovy
@@ -81,7 +81,9 @@ class UpsertTitusScalingPolicyAtomicOperation implements AtomicOperation<Map> {
 
       task.updateStatus BASE_PHASE, "Create Scaling Policy request constructed, sending..."
 
-      ScalingPolicyID result = client.createScalingPolicy(requestBuilder.build())
+      ScalingPolicyID result = retrySupport.retry({ ->
+        client.createScalingPolicy(requestBuilder.build())
+      }, 10, 3000, false)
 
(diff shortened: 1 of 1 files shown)


jeyrschabu:
When creating a new scaling policy, Spinnaker polls titus for approximately 50s to check if the policy was applied.
Spinnaker will fail the pipeline if the policy isn't reported as applied by Titus in that timeframe.
This happens whenever there is increase in latency between titus and aws and the new status isn't reflected in time. 
Having an exponential retry here sounds like a good idea.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/ops/UpsertTitusScalingPolicyAtomicOperation.groovy
```diff
@@ -99,9 +101,9 @@ class UpsertTitusScalingPolicyAtomicOperation implements AtomicOperation<Map> {
     retrySupport.retry({ ->
       ScalingPolicyResult updatedPolicy = client.getScalingPolicy(result.id)
       if (!updatedPolicy || (updatedPolicy.getPolicyState().state != ScalingPolicyState.Applied)) {
-        throw new IllegalStateException("New policy did not transition to applied state within 45 seconds")
+        throw new IllegalStateException("Timed out while waiting for policy to be applied")
       }
-    }, 5000, 10, false)
```
robzienert: lol omg

jeyrschabu: Lol

dreynaud: omg wat

PR #4093, opened 2019-10-15
refactor(google): use Optional.ofNullable() instead of if/else

plumpy:


plumpy:
Honestly mostly put this up to try out @robzienert's auto-merge thing, but I feel like it doesn't work so maybe I'll just submit it :grin:

plumpy:
It's alllliiiiiiiiiiive 🤖

PR #4096, opened 2019-10-16
fix(kubernetes/v1): Fix NPE in autoscaler caching

ezimanyi:
We are throwing an NPE if the resource field on a metric is null.

This PR is directly against the 1.16 branch as this has already been fixed by bumping the fabric8 library and changing how we look up metrics on master in #4060


ezimanyi:
@spinnakerbot :cherries: :pick: 1.15

PR #4097, opened 2019-10-16
fix(saga-retry): Opt-in pattern for retrying non-SpinnakerExceptions

jonsie:
The pattern of opting-out of non-SpinnakerExceptions is too broad and results in retries that quickly spiral out of control.  Until we have wide spread adoption of SpinnakerException and can make all of our retryable decisions off that, we will specifically opt-in to retrying exceptions.


clouddriver-core/src/test/groovy/com/netflix/spinnaker/clouddriver/orchestration/DefaultOrchestrationProcessorSpec.groovy
```diff
@@ -108,11 +108,12 @@ class DefaultOrchestrationProcessorSpec extends Specification {
     where:
     exception                     | sagaId               || retryable
     new RuntimeException()        | null                 || false
-    new NonRetryableException()   | null                 || false
-    new RuntimeException()        | new SagaId("a", "a") || true
+    new RetryableException()      | null                 || false
```
marchello2000: nit: this is a bit confusing now... maybe add a comment that the retry logic is only allowed for saga based operations?

clouddriver-core/src/test/groovy/com/netflix/spinnaker/clouddriver/orchestration/DefaultOrchestrationProcessorSpec.groovy
```diff
@@ -108,11 +108,12 @@ class DefaultOrchestrationProcessorSpec extends Specification {
...
+    new RuntimeException()        | new SagaId("a", "a") || false
     new NonRetryableException()   | new SagaId("a", "a") || false
-    new SomeDynamicException()    | new SagaId("a", "a") || false
-    new AnotherDynamicException() | new SagaId("a", "a") || false
+    new RetryableException()      | new SagaId("a", "a") || true
+    new SomeDynamicException()    | new SagaId("a", "a") || true
```
marchello2000: im confused... how come `SomeDynamic` and `AnotherDynamic` are retryable now?

jonsie: I reversed the logic in `ExceptionClassifier` so now we opt-in to retrying these exceptions.

clouddriver-core/src/test/groovy/com/netflix/spinnaker/clouddriver/orchestration/DefaultOrchestrationProcessorSpec.groovy
jonsie: Added a comment.

PR #4100, opened 2019-10-17
feat(kubernetes): support rolling restart operation for deployments

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4757

As the many comments/emojis on this [Kubernetes feature request](https://github.com/kubernetes/kubernetes/issues/13488) indicate, there are several valid user stories around issuing a rolling restart of a Deployment. The `kubectl rollout restart` command was released with kubectl version 1.15, and we currently use kubectl version 1.16 in Clouddriver. Like the other `kubectl rollout` commands, it is valid for Deployments, StatefulSets, and DaemonSets. We currently only expose the ad-hoc rollout commands for Deployments in Deck, but could at a later date expose them for StatefulSets and DaemonSets if there is evidence from users that it would be useful.

Related PRs:
Orca: https://github.com/spinnaker/orca/pull/3233
Deck: https://github.com/spinnaker/deck/pull/7538

863ee1d0-ca4a-4d50-b8ef-870bcfa72908

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/KubernetesRollingRestartManifestOperation.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/KubernetesRollingRestartManifestOperation.java
@@ -35,7 +35,7 @@ public class KubernetesRollingRestartManifestOperation implements AtomicOperatio
   public KubernetesRollingRestartManifestOperation(
       KubernetesRollingRestartManifestDescription description) {
     this.description = description;
-    this.credentials = (KubernetesV2Credentials) description.getCredentials().getCredentials();
+    this.credentials = description.getCredentials().getCredentials();
   }
 
   private static Task getTask() {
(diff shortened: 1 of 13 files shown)


clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/manifest/KubernetesRollingRestartManifestOperation.java
```diff
@@ -0,0 +1,68 @@
...
+
+public class KubernetesRollingRestartManifestOperation implements AtomicOperation<Void> {
+  private final KubernetesRollingRestartManifestDescription description;
+  private final KubernetesV2Credentials credentials;
+  private static final String OP_NAME = "ROLLING_RESTART_KUBERNETES_MANIFEST";
+
+  public KubernetesRollingRestartManifestOperation(
+      KubernetesRollingRestartManifestDescription description) {
+    this.description = description;
+    this.credentials = (KubernetesV2Credentials) description.getCredentials().getCredentials();
```
ezimanyi: Looks like all of the V2 operations need to do this casting, but I think we can avoid it by adding a generic `<C extends KubernetesCredentials>` to `KubernetesAtomicOperationDescription` and then making `KubernetesAtomicOperationDescription` be `KubernetesAtomicOperationDescription<KubernetesV2Credentials>`.

maggieneterval: Thank you for being the type of code reviewer that teaches me a new trick every time! Just updated in a separate commit; will open a separate PR to remove the now-unnecessary casts from the other Kubernetes V2 classes, and add the appropriate param to the V1 only classes.

PR #4101, opened 2019-10-17
feat(cfn): delete CFN changeset if empty upon request

xavileon:
This patch introduces the DeleteCloudFormationChangeSet async operation,
so Orca can request to delete a CFN change set under some conditions
(e.g. when it's empty, or after the change set has been executed). This
will allow full control of the change set life cycle by Spinnaker.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/DeleteCloudFormationChangeSetAtomicOperation.java
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/DeleteCloudFormationChangeSetAtomicOperation.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2019 Adevinta
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.clouddriver.aws.deploy.ops;
+
+import com.amazonaws.services.cloudformation.AmazonCloudFormation;
+import com.amazonaws.services.cloudformation.model.*;
+import com.netflix.spinnaker.clouddriver.aws.deploy.description.DeleteCloudFormationChangeSetDescription;
+import com.netflix.spinnaker.clouddriver.aws.security.AmazonClientProvider;
+import com.netflix.spinnaker.clouddriver.data.task.Task;
+import com.netflix.spinnaker.clouddriver.data.task.TaskRepository;
+import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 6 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/DeleteCloudFormationChangeSetAtomicOperation.java
```diff
@@ -0,0 +1,68 @@
...
+    DeleteChangeSetRequest deleteChangeSetRequest =
+        new DeleteChangeSetRequest()
+            .withStackName(description.getStackName())
+            .withChangeSetName(description.getChangeSetName());
+    try {
+      task.updateStatus(BASE_PHASE, "Deleting CloudFormation ChangeSet");
+      amazonCloudFormation.deleteChangeSet(deleteChangeSetRequest);
+    } catch (AmazonCloudFormationException e) {
+      log.error(
+          "Error removing change set {} on stack {}",
```
gregjones: What do you think about including something from the exception in the message to give an idea why it failed?

xavileon: I think it's a good idea 🙂 I'd add the message.

xavileon:
@spinnaker/reviewers PTAL

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/DeleteCloudFormationChangeSetAtomicOperation.java
```diff
@@ -0,0 +1,70 @@
...
+
+    DeleteChangeSetRequest deleteChangeSetRequest =
+        new DeleteChangeSetRequest()
+            .withStackName(description.getStackName())
+            .withChangeSetName(description.getChangeSetName());
+    try {
+      task.updateStatus(BASE_PHASE, "Deleting CloudFormation ChangeSet");
+      amazonCloudFormation.deleteChangeSet(deleteChangeSetRequest);
+    } catch (AmazonCloudFormationException e) {
+      log.error(
```
marchello2000: is it OK that this error is ignored?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/DeleteCloudFormationChangeSetAtomicOperation.java
```diff
@@ -0,0 +1,70 @@
...
+          "Error removing change set "
+              + description.getChangeSetName()
+              + " on stack "
+              + description.getStackName(),
+          e.getMessage());
```
marchello2000: log the full exception
```suggestion
          e;
```

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/DeleteCloudFormationChangeSetAtomicOperation.java
xavileon: Do you mean not returning any error back to orca? Fair comment, I'll return an error so orca can handle it appropiately, probably with a retryable task :+1:

xavileon: Done in https://github.com/spinnaker/clouddriver/pull/4101/commits/c7cd1349d81479884881e3937a0e497146128f0d

PR #4102, opened 2019-10-17
fix(runJob): fix output on multi container

ethanfrogers:
fixes a bug where you couldn't fetch output on jobs with multiple
containers. this was because we weren't specifying the exact container
name to fetch. update the method to take the container name.

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
@@ -452,9 +452,12 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
         () -> jobExecutor.logs(this, namespace, podName, containerName));
   }
 
-  public String jobLogs(String namespace, String jobName) {
+  public String jobLogs(String namespace, String jobName, String containerName) {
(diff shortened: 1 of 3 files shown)


ethanfrogers:
spinnaker/spinnaker#5031

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -457,6 +457,14 @@ public String jobLogs(String namespace, String jobName) {
         "logs", KubernetesKind.JOB, namespace, () -> jobExecutor.jobLogs(this, namespace, jobName));
   }
 
+  public String jobLogs(String namespace, String jobName, String containerName) {
```
ezimanyi: It looks like we can delete the other `jobLogs` which is now only used it a test, which would allow us to also delete the other `KubectlJobExecutor.jobLogs()`.

ethanfrogers: Yep, fooled by IntelliJ not graying the other one out. Removed!

ethanfrogers:

maheshopsmx:
If same container run job  fails with exit 1, it is not able to call the spinnaker property in the next stage

PR #4104, opened 2019-10-17
refactor(core): Move RequestQueue to core

ezimanyi:
Does this seems like a reasonable move, to try to keep the kubernetes code out of `clouddriver-core` and `clouddriver-web`?

* refactor(core): Move RequestQueue to core 

  A number of controllers currently use a RequestQueue to help balance requests from different applications. One such controller is the kubernetes-only ManifestController that I'd like to move out of clouddriver-web and into clouddriver-kubernetes-v2 as it is causing a fair amount of kubernetes-specific code to live in core.

  In order to support controllers (such as this one) that live in their respective modules still using RequestQueue, move the RequestQueue logic to core.

* refactor(kubernetes): Move ManifestController to Kubernetes 

  The ManifestController is currently in clouddriver-web even though it only applies to Kubernetes accounts; this is causing a number of kubernetes-specific classes to live in clouddriver-core.

  Move the ManifestController to live in clouddriver-kubernetes-v2, and also move the classes it depdends on out of clouddriver-core and into clouddriver-kubernetes-v2.


cfieber:
seems reasonable to me, thanks

PR #4110, opened 2019-10-18
perf(kubernetes): Fix all agents handling on-demand requests

ezimanyi:
* test(kubernetes): Add tests to KubernetesCoreCachingAgent 

  The next commit will fix a bug in the Kubernetes caching agents; there are currently no tests validating these agents. Add some tests on the KubernetesCoreCachingAgent.

* perf(kubernetes): Fix all agents handling on-demand requests 

  Currently sending an on-demand cache update for a resource that is not namespaced will cause all caching agents to try to handle the update. This can cause a very high load on redis if there are a lot of agents (as might be expected in a large installation of Spinnaker).

  Designate one agent to handle requests that are not namespaced instead of having them handled by all agents.

### DIFF
--- a/clouddriver-kubernetes-v2/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesCoreCachingAgentTest.java
+++ b/clouddriver-kubernetes-v2/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesCoreCachingAgentTest.java
@@ -16,41 +16,40 @@
 
 package com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.agent;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyZeroInteractions;
 import static org.mockito.Mockito.when;
-import static org.mockito.internal.verification.VerificationModeFactory.times;
 
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableCollection;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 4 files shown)


clouddriver-kubernetes-v2/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesCoreCachingAgentTest.java
```diff
@@ -0,0 +1,317 @@
...
+    ProviderCache providerCache = mock(ProviderCache.class);
+    Map<String, String> data =
+        ImmutableMap.of(
+            "account", ACCOUNT,
+            "location", NAMESPACE1,
+            "name", "deployment " + DEPLOYMENT_NAME);
+    ImmutableList<OnDemandAgent.OnDemandResult> results =
+        processOnDemandRequest(data, providerCache);
+
+    verify(providerCache, times(1)).putCacheData(any(String.class), any(CacheData.class));
```
plumpy: arguably these tests would be better if you used an in-memory cache instead of a mock (`new DefaultProviderCache(new InMemoryCache())`) and tested to see what _actually_ exists in the cache instead of just testing what methods were called...

But I obviously appreciate they likely already took a long time to write and they're certainly better with mocks than not existing at all, so I 100% would understand if you didn't want to rework them all at this point...

In this case, however, I think you'd probably want to specify the arguments to the mocks. `any()` is useful for stubby, but not as great for verification. Can you just move these to the end of the method and do
`verify(providerCache).putCacheData(cacheData.getId(), cacheData)`?

ezimanyi: Great point...added a commit to use an `InMemoryCache()` instead of stubbing the `DefaultProviderCache`.  I also made a few other improvements to the tests while I was looking at them again.

maggieneterval:
:arrow_right: :thought_balloon: :memo: :chart_with_upwards_trend: :tada: 
:man: :hammer_and_pick: :computer:

ezimanyi:
@maggieneterval : I think you said you wanted to merge this...it's all set!

maggieneterval:
@ezimanyi thank you, tis squashed and merged! :pray:

PR #4118, opened 2019-10-24
fix(cloudconfig): Log errors and continue

ncknt:
When refresh fails, the exception is not logged.  This causes hard to debug issues and more importantly Clouddriver no longer runs the scheduler afterwards.


ncknt:
@scottfrederick It's a follow up to https://github.com/spinnaker/kork/pull/404

ncknt:
@ethanfrogers Can you PTAL?

PR #4119, opened 2019-10-24
refactor(google): Update to latest google credentials style.

duftler:


duftler:
@german-muzquiz please review

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gcs/GcsArtifactCredentials.java
```diff
@@ -50,25 +52,27 @@
     JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();
     String credentialsPath = account.getJsonPath();
 
-    GoogleCredential credential;
+    GoogleCredentials credentials;
 
     if (!StringUtils.isEmpty(credentialsPath)) {
```
ttomsu: I know you didn't touch this line, but it seems like this logic could be flipped and easier to read:

```
if (StringUtils.isEmpty(credentialsPath)) {
  credentials = GoogleCredentials.getApplicationDefault();
} else {
  credentials = GoogleCredentials.fromStream(stream)...
}
```

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/security/AppengineCredentials.groovy
```diff
@@ -16,24 +16,29 @@
...
 import com.google.api.client.json.JsonFactory
 import com.google.api.client.json.jackson2.JacksonFactory
 import com.google.api.services.appengine.v1.Appengine
+import com.google.auth.http.HttpCredentialsAdapter
+import com.google.auth.oauth2.GoogleCredentials
 import com.netflix.spinnaker.clouddriver.googlecommon.security.GoogleCommonCredentials
+import groovy.transform.CompileStatic
 import groovy.transform.TupleConstructor
 
+@CompileStatic
```
german-muzquiz: Looking at what this annotation does, seems that it gets rid of some Groovy dynamic features: https://stackoverflow.com/a/29611007/19136 . This class is small, would it be better to just migrate it to full Java instead? I'm just trying to avoid that in the future somebody adds some groovy dynamic code without noticing (or knowing) what are the effects of `@CompileStatic`.

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/security/AppengineJsonCredentials.groovy
```diff
@@ -20,8 +20,11 @@ import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
 import com.google.api.client.http.HttpTransport
 import com.google.api.client.json.JsonFactory
 import com.google.api.services.appengine.v1.AppengineScopes
+import com.google.auth.oauth2.GoogleCredentials
 import com.netflix.spinnaker.clouddriver.googlecommon.security.GoogleCommonCredentialUtils
+import groovy.transform.CompileStatic
 
+@CompileStatic
```
german-muzquiz: Same comment about considering migrating to Java instead of using `@CompileStatic`.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gcs/GcsArtifactCredentials.java
duftler: Ok, done.

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/security/AppengineJsonCredentials.groovy
duftler: Have converted those 2 classes to java.

PR #4123, opened 2019-10-25
fix(google): don't leave orphaned applications in the cache

plumpy:
If an `AbstractGoogleServerGroupCachingAgent` has some applications stored,
but in the next run it finds no applications, it doesn't put anything
under the `applications` key. `DefaultProviderCache` then has some weird
behavior where it removes all relationships associated with that
application, but leaves the existing application in the cache. That
means we'll just repeat that behavior again the next time the caching
agent runs. The application data is just orphaned but triggers
relationship evictions every single caching cycle.

Amazon doesn't have this problem because they always stick a
`List<CacheData>` under the `applications` key, even if it's empty. This is
enough to tell `putCacheResult()` to store that empty list, overwriting
whatever was there before.

The orphaned data isn't the cause of spinnaker/spinnaker#4511, but it
dramatically exacerbates that issue. (The real issue is that every
single regional/zonal caching agent asserts ownership over the
application and will happily delete it, which triggers deletions of
relationships that were put there by other caching agents.)

I was working on a larger refactor of CATS to make this bug less likely
to happen, but since we're about to cut a release branch, I'd rather
just check in this smaller fix.


clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/cache/CacheResultBuilderTest.java
```diff
@@ -0,0 +1,123 @@
...
+        .namespace("auth2")
+        .keep("id2")
+        .setAttributes(ImmutableMap.of("attr2", "value2"));
+    cacheResultBuilder
+        .namespace("auth3")
+        .keep("id3")
+        .setAttributes(ImmutableMap.of("attr3", "value3"));
+
+    Map<String, Collection<CacheData>> cacheResults = cacheResultBuilder.build().getCacheResults();
+    assertThat(cacheResults.get("auth1")).isEmpty();
```
duftler: nice

PR #4129, opened 2019-10-29
feat(gce): support autoscaler scale-down controls

plumpy:


maggieneterval:
Update: I tried submitting a `scaleDownControl` config with each member set to `null`, and that didn't work either (received the error `"Invalid value for field 'resource.autoscalingPolicy.scaleDownControl.maxScaledDownReplicas': ''. None of the fields has a value, exactly one is required."`)

clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/AbstractGoogleServerGroupCachingAgentTest.java
```diff
@@ -710,6 +715,101 @@ void serverGroupAutoscalerProperties() {
...
+        .isEqualTo(input.getLoadBalancingUtilization().getUtilizationTarget());
+    assertThat(converted.getMaxNumReplicas()).isEqualTo(input.getMaxNumReplicas());
+    assertThat(converted.getMinNumReplicas()).isEqualTo(input.getMinNumReplicas());
+    assertThat(converted.getMode().toString()).isEqualTo(input.getMode());
+
+    assertThat(converted.getCustomMetricUtilizations())
+        .hasSize(input.getCustomMetricUtilizations().size());
+    for (int i = 0; i < converted.getCustomMetricUtilizations().size(); ++i) {
+      CustomMetricUtilization convertedCustomMetric =
+          converted.getCustomMetricUtilizations().get(0);
```
ezimanyi: Should `.get(0)` be `.get(i)` here and on the next line?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/AbstractGoogleServerGroupCachingAgent.java
```diff
@@ -887,6 +894,77 @@ private static void populateAutoscaler(
...
+    if (input == null) {
+      return null;
+    }
+    LoadBalancingUtilization output = new LoadBalancingUtilization();
+    output.setUtilizationTarget(input.getUtilizationTarget());
+    return output;
+  }
+
+  @Nullable
+  private static List<CustomMetricUtilization> convertCustomMetricUtilizations(
```
ezimanyi: Maybe return an `ImmutableList` here to indicate to the caller it can't be modified?  (Or perhaps not as important for private methods.)

clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/AbstractGoogleServerGroupCachingAgentTest.java
plumpy: Whoops, fixed, thanks!

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/AbstractGoogleServerGroupCachingAgent.java
plumpy: :+1:

plumpy:
Thanks for catching that Maggie. As discussed, this now works kinda like the other fields: if left empty, the previous value will be retained. If you set `scaleDownControl:{}`, it will erase what was there previously.

maggieneterval:
Clearing a scale-down policy works now! :tada: :tada:

PR #4130, opened 2019-10-29
fix(cf): reduce unnecessary calls to foundation to improve performance.

Jammy-Louie:
when caching was occurring we were making a lot of out going API calls the to the cf foundation. Some of the API calls are not performant and should not be called very often.

reduce the number of outgoing calls to help improve performance


Jammy-Louie:

PR #4135, opened 2019-10-30
fix(kubernetes): Fix mis-spelled API group

ezimanyi:
The networking.k8s.io API group is mis-spelled as network.k8s.io; fix this. This bug is preventing security groups from being cached by clouddriver.

We also should accept extensions/v1beta1 as the API group, which can still be returned from clusters running Kubernetes < 1.16.


ezimanyi:

PR #4138, opened 2019-10-31
fix(titus): allow setting jobGroupDetails and jobGroupStack for Titus…

tomaslin:
At one point in 2018 we disabled setting the job group detail and stack for batch jobs under the incorrect assumption that it was increasing the time for deploying Titus instances. 

This PR restores the ability to set job group details, stack and sequence number which is available in the Titus API.


tomaslin:
There is no need to change the read side since it was not bound to the actual job type https://github.com/spinnaker/clouddriver/blob/master/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/Job.java#L142-L144

PR #4139, opened 2019-11-01
fix(provider/aws): Handling an exception to prevent clouddriver crash

kirangodishala:
Issue: [spinnaker/spinnaker#5071](https://github.com/spinnaker/spinnaker/issues/5071)

If the AWS credentials - secret key id and secret access key - are incorrect, the clouddriver pod crashes.

The fix allows clouddriver to continues with other cloud providers instead of crashing.


cfieber:
closing this for now, lets discuss in spinnaker/spinnaker#5071

PR #4141, opened 2019-11-04
fix(aws): support for cross-application CLBs with the sql backend

asher:
- Fixes https://github.com/spinnaker/spinnaker/issues/5061

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonLoadBalancerInstanceStateCachingAgent.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonLoadBalancerInstanceStateCachingAgent.groovy
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.clouddriver.aws.provider.agent
 
 import com.amazonaws.services.elasticloadbalancing.model.DescribeInstanceHealthRequest
+import com.amazonaws.services.elasticloadbalancing.model.DescribeInstanceHealthResult
 import com.amazonaws.services.elasticloadbalancing.model.LoadBalancerNotFoundException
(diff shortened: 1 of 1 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonLoadBalancerInstanceStateCachingAgent.groovy
```diff
@@ -98,39 +100,60 @@ class AmazonLoadBalancerInstanceStateCachingAgent implements CachingAgent, Healt
...
-
           Map<String, Object> attributes = objectMapper.convertValue(ilb, ATTRIBUTES)
-          if (idObj.containsKey("application")) {
-            attributes.put("application", idObj.get("application"))
+          Map<String, Collection<String>> relationships = [(INSTANCES.ns): [instanceId]]
+
+          if (instances[instanceId] != null) {
+            String application = instances[instanceId].attributes.get("application")
+            if (application != null) {
+              attributes.put("application", application)
```
asher: This ensures that health objects stored by `AmazonLoadBalancerInstanceStateCachingAgent` are indexed by the application of instances the objects relate to, instead of the application the load balancer itself. This was already done by `EddaLoadBalancerCachingAgent` which runs instead of `AmazonLoadBalancerInstanceStateCachingAgent` when edda is enabled.

asher:

costimuraru:
@asher do you think it's possible to cherry pick this in 1.16 as well? Not seeing any breaking changes. Furthermore, I made a custom build where I cherry picked this commit over the `release-1.16.x` branch and it seems to be working, but it would be great to not use custom images for our prod spinnaker deployment. Thanks.

jonsie:

PR #4147, opened 2019-11-06
fix(titus): backoff for caching agent observeJob retries

asher:


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
```diff
@@ -275,15 +270,20 @@ public void executeAgent(Agent agent) {
...
                         e);
-                    notificationIt =
-                        titusClient.observeJobs(
-                            ObserveJobsQuery.newBuilder()
-                                .putFilteringCriteria("jobType", "SERVICE")
-                                .putFilteringCriteria("attributes", "source:spinnaker")
-                                .build());
+
+                    try {
+                      Thread.sleep(2000);
```
jonsie: Thoughts on making this dynamically configurable?

PR #4150, opened 2019-11-06
feat(*): Github Actions Experiment

ajordens:


.github/workflows/pr.yml
```diff
@@ -0,0 +1,25 @@
+name: PR
+
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
jonsie: We could probably just pull down a public open JDK 8 docker image instead of running on this vanilla VM and then skip the `actions/setup-java@v1` step below.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,25 @@
...
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 8
+    # - uses: actions/cache@v1
+    #   with:
+    #     path: ~/.gradle
+    #     key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+    #     restore-keys: |
+    #       ${{ runner.os }}-gradle-
+    - uses: eskatos/gradle-command-action@v1
```
jonsie: I think we want to use the repositiries Gradle wrapper (`./gradlew`) rather than the version of Gradle that this action uses.

ajordens: I believe we are ...

image

.github/workflows/pr.yml
ajordens: That's a fair idea -- I've not seen any github actions posts that mention doing that and I suspect that `ubuntu-latest` is locally cached so the fetch is quick.

Takes ~7s to install java.

PR #4155, opened 2019-11-06
feat(saga): Saga exception handler and TitusExceptionHandler implementation

jonsie:
Rather than scattering the exception catching logic (for determining if something is retryable) all throughout the cloudprovider code, this pattern should help keep things simple.

### DIFF
--- a/clouddriver-saga/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/flow/CompensatingSagaAction.kt
+++ b/clouddriver-saga/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/flow/CompensatingSagaAction.kt
@@ -15,10 +15,10 @@
  */
 package com.netflix.spinnaker.clouddriver.saga.flow
 
-import com.google.common.annotations.Beta
 import com.netflix.spinnaker.clouddriver.saga.SagaCommand
 import com.netflix.spinnaker.clouddriver.saga.SagaRollbackCommand
 import com.netflix.spinnaker.clouddriver.saga.models.Saga
+import com.netflix.spinnaker.kork.annotations.Beta
 
 /**
  * A [SagaAction] that has a companion [rollback] method.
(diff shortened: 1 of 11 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/TitusException.java
```diff
@@ -21,36 +21,16 @@
 public class TitusException extends IntegrationException {
   public TitusException(String message) {
     super(message);
-    setRetryable(true);
```
jonsie: Removed setting TitusException as retryable when only a message is provided - this makes sense to me given where this exception is thrown.

clouddriver-saga/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/flow/CompensatingSagaAction.kt
```diff
@@ -15,10 +15,10 @@
  */
 package com.netflix.spinnaker.clouddriver.saga.flow
 
-import com.google.common.annotations.Beta
 import com.netflix.spinnaker.clouddriver.saga.SagaCommand
 import com.netflix.spinnaker.clouddriver.saga.SagaRollbackCommand
 import com.netflix.spinnaker.clouddriver.saga.models.Saga
+import com.netflix.spinnaker.kork.annotations.Beta
```
robzienert: Thanks for changing these.

clouddriver-saga/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/flow/SagaExceptionHandler.kt
```diff
@@ -13,24 +13,18 @@
...
-    expect:
-    new TitusException("Something bad happened", downstream).retryable == retryable
-
-    where:
-    message                                  || retryable
-    "INVALID_ARGUMENT: Image does not exist" || false
-    "Something else entirely"                || true
-  }
+/**
+ * The exception handler is an optional interface for implementors to use when determining how to
```
robzienert:
```suggestion
 * The [ExceptionHandler] is an optional interface for implementors to use when determining how to
```

PR #4158, opened 2019-11-07
fix(caching): Config to disable fully qualified caching agents

jonsie:
This works in conjunction with the pattern matcher to enable caching agents (which I think is still useful, just not something we should have available to dynamic configuration)

### DIFF
--- a/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
+++ b/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
@@ -57,6 +57,7 @@ class SqlClusteredAgentScheduler(
   private val nodeStatusProvider: NodeStatusProvider,
   private val dynamicConfigService: DynamicConfigService,
   enabledAgentPattern: String,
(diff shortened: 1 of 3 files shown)


cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
```diff
@@ -90,6 +92,11 @@ class SqlClusteredAgentScheduler(
     val lockInterval = agentLockAcquisitionIntervalSeconds ?: 1L
     lockPollingScheduler.scheduleAtFixedRate(this, 0, lockInterval, TimeUnit.SECONDS)
     enabledAgents = Pattern.compile(enabledAgentPattern, CASE_INSENSITIVE)
+    disabledAgents = dynamicConfigService.getConfig(
```
asher: I think this will require a clouddriver restart to pick up changes to the dynamic property, not sure if that's the intent?

jonsie: That's not the intent - good catch.  Let me fix that.

PR #4165, opened 2019-11-12
fix(securityGroup): Add tags field to AmazonSecurityGroup model

aravindmd:
- Add `tags` field to `AmazonSecurityGroup` model


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonSecurityGroup.groovy
```diff
@@ -39,6 +39,8 @@ class AmazonSecurityGroup implements SecurityGroup {
   final String region
   final Set<Rule> inboundRules
   final Set<Rule> outboundRules
+  final List<Map> tags
```
jeyrschabu: Left a comment on the other PR about using the amazon Tag model

aravindmd: updated to use `com.amazonaws.services.ec2.model.Tag`

PR #4166, opened 2019-11-13
fix(cats/sql): fix table names with long type names

ncknt:
This should fix table with names > `sql.constraints.max-table-name-length:64` characters. It works by hashing part of the name. It will hash the `type` first, then if not enough the `type` and the suffix (`_rel`). It will throw an exception if it cannot shorten the name while preserving `cats_v1_<table namespace>`.

Also exposed `MAX_ID_LENGTH` in the configuration class (`sql.constraints.max-id-length:352`).

```
cats_v1_foo => cats_v1_foo
cats_v1_superduper....longname => cats_v1_superduper<hash>
cats_v1_superduper....longname_rel => cats_v1_superduper<hash>_rel
```

Fix for https://github.com/spinnaker/spinnaker/issues/4809

### DIFF
--- a/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
+++ b/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
@@ -912,22 +912,17 @@ class SqlCache(
     }
 
     // Hash the name and keep the suffix
-    val hash = Hashing.murmur3_128().hashBytes(name.toByteArray()).toString().substring(0..15)
+    val hash = Hashing.murmur3_128().hashBytes((name + suffix).toByteArray()).toString().substring(0..15)
     val available = sqlConstraints.maxTableNameLength - base.length - suffix.length - hash.length - 1
     if (available >= 0) {
       return base + name.substring(0..available) + hash + suffix
     }
 
-    // Hash name and suffix
-    val pHash = Hashing.murmur3_128().hashBytes((name + suffix).toByteArray()).toString().substring(0..15)
-    val pAvailable = sqlConstraints.maxTableNameLength - base.length - hash.length - 1
-    if (pAvailable >= 0) {
-      return base + name.substring(0..pAvailable) + pHash
+    // Remove suffix
+    if (available + suffix.length >= 0) {
+      return base + name.substring(0..(available + suffix.length)) + hash
     }
-    if (base.length + pHash.length > sqlConstraints.maxTableNameLength) {
-      throw IllegalArgumentException("property sql.table-namespace $tableNamespace is too long")
-    }
-    return base + pHash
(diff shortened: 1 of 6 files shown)


ncknt:
@ezimanyi  Could you PTAL (you had self-assigned it)?

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -884,15 +885,51 @@ class SqlCache(
...
+
+    // Hash the name and keep the suffix
+    val hash = Hashing.murmur3_128().hashBytes(name.toByteArray()).toString().substring(0..15)
+    val available = sqlConstraints.maxTableNameLength - base.length - suffix.length - hash.length - 1
+    if (available >= 0) {
+      return base + name.substring(0..available) + hash + suffix
+    }
+
+    // Hash name and suffix
+    val pHash = Hashing.murmur3_128().hashBytes((name + suffix).toByteArray()).toString().substring(0..15)
```
asher: Avoid hashing twice in this case. They're all known up front so determine what to hash based on the length of the relevant components.

ncknt: Thanks, Updated

ncknt:
Can we port to 1.17.x? I don't know if there's a bot for that.

ezimanyi:
@ncknt : You can ask spinnakerbot to cherry-pick (see an example [here](https://github.com/spinnaker/deck/pull/7600#issuecomment-551190651)).  It will ultimately be up to the release manager (who is @ttomsu this release) to decide whether the fix is small/safe enough to merge.  This is not a tiny fix, but it does fix a blocking bug for kubernetes/SQL users so I'd lean towards cherry picking but it's roughly on the limit of what I'd feel safe cherry-picking.

clanesf:
We would second cherry picking back to 1.17 if possible.

cats/cats-sql/src/test/groovy/com/netflix/spinnaker/cats/sql/SqlCacheSpec.groovy
```diff
@@ -96,6 +97,23 @@ class SqlCacheSpec extends WriteableCacheSpec {
...
+  @Unroll
+  def 'max length of table name is checked'() {
+    when:
+    def realName = ((SqlCache) cache).checkTableName("cats_v1_", name, suffix)
+
+    then:
+    realName == expected
+
+    where:
+    name              || expected                                                             || suffix
```
ttomsu: Style nit: Spock tests usually split inputs with a singe pipe `|`from the outputs with a double pipe `||`. http://spockframework.org/spock/docs/1.0/data_driven_testing.html#_syntactic_variations

ttomsu:

PR #4167, opened 2019-11-13
feat(cf): add SpEL support for cf manifests

Jammy-Louie:
refactor the code to support SpEL evaluations


Jammy-Louie:
relies on https://github.com/spinnaker/orca/pull/3299

PR #4168, opened 2019-11-15
chore(dependencies): Upgrade Spring Boot to 2.2.1

scottfrederick:
See https://github.com/spinnaker/spinnaker/issues/5134

Dependent on https://github.com/spinnaker/kork/pull/419 for a Spring Boot dependency upgrade

### DIFF
--- a/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cache/AbstractRedisCache.java
+++ b/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cache/AbstractRedisCache.java
@@ -234,7 +234,7 @@ public abstract class AbstractRedisCache implements WriteableCache {
     return String.join(":", prefix, type, "members");
   }
 
-  protected TypeReference<?> getRelationshipsTypeReference() {
+  protected TypeReference<? extends Collection<String>> getRelationshipsTypeReference() {
     return options.isTreatRelationshipsAsSet() ? RELATIONSHIPS_SET : RELATIONSHIPS_LIST;
   }
 }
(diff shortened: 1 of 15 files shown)


cfieber:
this one troubles me a bit with the typeref changes requiring more specialized types on the left hand side - going to poke around a bit and see if there are any other options there (and live with it as is if we have to)

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cache/AbstractRedisCache.java
```diff
@@ -234,7 +234,7 @@ protected String allOfTypeId(String type) {
     return String.join(":", prefix, type, "members");
   }
 
-  protected TypeReference getRelationshipsTypeReference() {
+  protected TypeReference<?> getRelationshipsTypeReference() {
```
cfieber:
```suggestion
  protected TypeReference<? extends Collection<String>> getRelationshipsTypeReference() {
```

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cache/RedisCache.java
```diff
@@ -306,7 +306,7 @@ private CacheData extractItem(String id, List<String> keyResult, List<String> kn
         if (rel != null) {
           String relType = knownRels.get(relIdx - 1);
           Collection<String> deserializedRel =
-              objectMapper.readValue(rel, getRelationshipsTypeReference());
+              (Collection<String>) objectMapper.readValue(rel, getRelationshipsTypeReference());
```
cfieber: cast should be unnecessary with the above comment

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cache/AbstractRedisCache.java
scottfrederick: Good catch. Any improvement in the types is a good thing. Resolved in 43f237c16f37208c069aceec126271f773eb57b9

PR #4170, opened 2019-11-20
fix(kubernetes): Add missing limitrange kind

ezimanyi:
Fixes spinnaker/spinnaker#5139

Deployments containing a limitrange time out waiting for the manifest to appear, as clouddriver is missing the entry for this kind.

Ideally there would be a better way of auto-detecting kinds like this (particularly ones that don't need any custom logic) but for now just adding another Handler to handle this.


ezimanyi:
@spinnakerbot 🍒 ⛏ 1.17

ezimanyi:
@spinnakerbot 🍒 ⛏ 1.17

PR #4177, opened 2019-11-22
fix(titus): Resolve next sequence number when cluster details match sequence pattern

jonsie:


clouddriver-titus/src/test/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/TitusServerGroupNameResolverSpec.groovy
```diff
@@ -0,0 +1,93 @@
...
+
+    then:
+    result == nextServerGroupName
+
+    where:
+    details      | sequence | nextSequence
+    "v82589065"  |  "v000"  | "v001"
+    "v82589065"  |  "v001"  | "v002"
+    "v82589065"  |  "v998"  | "v999"
+    "v8258c06b"  |  "v000"  | "v001" //one a-z0-9 test for good measure
```
cfieber: Try adding v00001 as a detail test case... Suspect that bombs

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/TitusServerGroupNameResolver.groovy
```diff
@@ -41,8 +41,14 @@ class TitusServerGroupNameResolver extends AbstractServerGroupNameResolver {
...
     def clusterNameParts = Names.parseName(clusterName)
+
+    // The sequence does not exist as part of the clusterName.  Therefore, if we have a sequence,
+    // this is because frigga parsed the detail as a sequence.
+    def cluster = clusterNameParts.sequence ?
+      clusterNameParts.cluster + "-v${clusterNameParts.sequence}" : clusterNameParts.cluster
+
     List<Job> jobs = titusClient.findJobsByApplication(clusterNameParts.app)
-      .findAll { Names.parseName(it.name).cluster == Names.parseName(clusterName).cluster }
+      .findAll { Names.parseName(it.name).cluster == cluster }
```
cfieber: Suspect this could just be Names.parseName(it.name) == clusterName and get rid of the logic above to derive cluster

cfieber:
Suspect this would also break with a stack named vNNN and no detail component

jonsie:
Removed unnecessary usage of Frigga to parse out cluster name.

PR #4178, opened 2019-11-22
feat(kubernetes): do not set replicas to 1 when removed

brianwolfe:
WIP

When adding a horizontal pod autoscaler, it is typical to remove the
replicas: N from the Deployment specification. Unfortunately, that will
set the pod count to 1 on the subsequent deployment. This PR is an early
attempt at preventing that by explicitly patching the
last-applied-configuration in kubernetes.

For more details about this issue (and the original proposed workaround in kubernetes that I'm drawing this from, please look at: https://github.com/kubernetes/kubernetes/issues/25238#issuecomment-406415297


brianwolfe:
@ezimanyi

Thanks for the feedback. @gregorymfoster is cleaning up a minimal version of this on an internal PR, will update this one when we  have something that works.

1. We can short-circuit, it's a good idea to keep things clean for those cases.
2. Guarding replicas == 0. I'm open to removing this guard. In our internal case it is there to make sure that we actually scale up from 0 in this case so the autoscaler has something to work with. I think this is unnecessary in the general case.
3. Yes. This should be a onetime edit (and in fact, it will only apply one time with the current implementation if I haven't screwed something up). This sort of modification is quite dangerous (we have seen two internal outages because of this issue because people didn't follow the manual steps it wasn't apparent until the production deploy scaled to 1 pod). I'd honestly rather use blue-green rollouts, but we can't transition globally to that right now...so this is my only real idea.

ezimanyi:
Sounds good, thanks for the details! I think this looks good---just let me know when the final PR is ready and I'll take a more detailed look.

robzienert:
@brianwolfe ping

robzienert:
Closing. @brianwolfe you are free to re-open when you finish up.

abhiroop93:
Hey this is still happening with us, our deployments get scaled to one on redeploying. Any clues on how to fix it?

PR #4179, opened 2019-11-22
refactor(docker): Remove unused getTags() function

ezimanyi:
Fixes spinnaker/spinnaker#5002 though that issue no longer actually manifests since 1.15 due to the fact this code has been unused since then.

The function getTags() in DockerRegistryNamedAccountCredentials has a bug. Depending on whether we return from the try or the catch block, we'll get either an Instant or a Date, which we then put into a single array and try to sort. In some cases this can lead to an error 'Comparison method violates its general contract' when we try to sort this heterogenous array.

While fixing this, I realized that this function is completely unused since #3809, so just removing the function instead of fixing it.


dbyron0:
🎉 Thank you!

emjburns:
@ezimanyi I was looking to use this function for the docker artifact support I'm working on in `keel` - would you mind fixing the bug instead? :) if not, I can fix the bug.

ezimanyi:
@emjburns : I would love to fix the bug.  (Or, in other words, I already fixed the bug and was writing tests on it when I realized it was unused, so I will fell better about those 2 hours of my life if I fix it.)  I'll make another PR to do that.

emjburns:
Thanks so much @ezimanyi !

PR #4180, opened 2019-11-22
perf(kubernetes): improve pendingOnDemandRequests lookups when a scoped manifest is provided

alperkokmen:
changes are split into two commits. it's probably easier to review them separately.

- f2d6fbe
this introduces a new parameter called `data` (`Map<String, String>`) to `handle()` and `pendingOnDemandRequests()` methods in `OnDemandCacheUpdater` and `OnDemandAgent` interfaces. this is done primary to allow `CacheController` endpoint to accept parameters other than `id` and pass them to agents. primary motivation here was to use these new arguments in `KubernetesV2OnDemandCachingAgent`.

- d7b5fad
this updates `KubernetesV2OnDemandCachingAgent#pendingOnDemandRequests` to
make use of the newly introduced parameter mentioned above. in order to preserve the current functionality, method uses presence of scoped manifest fields in data parameter (account and name) to determine whether or not to derive `matchingKeys` without any cache lookups. when a scoped manifest is provided, we use `InfrastructureCacheKey.createKey` to derive the key we need to look up in the cache.

with these changes, @fkyc and i have seen substantial increase in # of concurrent k8s deploys (100 -> 700) we can do without pegging redis CPU utilization at 100% as it eliminates the need to do a scan and then multi-get.

note that this is against `1.15.x` which doesn't have some of the other perf improvements made since then. if these changes look reasonable, we will be submitting another patch against master.

### DIFF
--- a/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AppengineLoadBalancerCachingAgent.groovy
+++ b/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AppengineLoadBalancerCachingAgent.groovy
@@ -36,6 +36,7 @@ import com.netflix.spinnaker.clouddriver.cache.OnDemandMetricsSupport
 import com.netflix.spinnaker.clouddriver.kubernetes.v1.provider.view.MutableCacheData
 import groovy.util.logging.Slf4j
 
+import javax.annotation.Nonnull
 import java.util.concurrent.TimeUnit
 import java.util.stream.Collectors
 
@@ -82,7 +83,7 @@ class AppengineLoadBalancerCachingAgent extends AbstractAppengineCachingAgent im
   }
 
   @Override
-  boolean handles(OnDemandAgent.OnDemandType type, String cloudProvider) {
+  boolean handles(OnDemandAgent.OnDemandType type, String cloudProvider, @Nonnull Map<String, String> data) {
     type == OnDemandAgent.OnDemandType.LoadBalancer && cloudProvider == AppengineCloudProvider.ID
   }
 
@@ -210,7 +211,7 @@ class AppengineLoadBalancerCachingAgent extends AbstractAppengineCachingAgent im
   }
 
   @Override
-  Collection<Map> pendingOnDemandRequests(ProviderCache providerCache) {
+  Collection<Map> pendingOnDemandRequests(ProviderCache providerCache, @Nonnull Map<String, String> data) {
     def keys = providerCache.getIdentifiers(ON_DEMAND.ns)
     keys = keys.findResults {
       def parse = Keys.parse(it)
--- a/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AppengineServerGroupCachingAgent.groovy
+++ b/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/agent/AppengineServerGroupCachingAgent.groovy
@@ -42,6 +42,8 @@ import com.netflix.spinnaker.clouddriver.cache.OnDemandMetricsSupport
 import com.netflix.spinnaker.clouddriver.kubernetes.v1.provider.view.MutableCacheData
 import groovy.util.logging.Slf4j
 
+import javax.annotation.Nonnull
+
 import static com.netflix.spinnaker.cats.agent.AgentDataType.Authority.AUTHORITATIVE
 import static com.netflix.spinnaker.cats.agent.AgentDataType.Authority.INFORMATIVE
 import static com.netflix.spinnaker.clouddriver.appengine.cache.Keys.Namespace.*
@@ -89,7 +91,7 @@ class AppengineServerGroupCachingAgent extends AbstractAppengineCachingAgent imp
   }
 
   @Override
-  boolean handles(OnDemandAgent.OnDemandType type, String cloudProvider) {
+  boolean handles(OnDemandAgent.OnDemandType type, String cloudProvider, @Nonnull Map<String, String> data) {
     type == OnDemandAgent.OnDemandType.ServerGroup && cloudProvider == AppengineCloudProvider.ID
   }
 
@@ -406,7 +408,7 @@ class AppengineServerGroupCachingAgent extends AbstractAppengineCachingAgent imp
   }
 
   @Override
-  Collection<Map> pendingOnDemandRequests(ProviderCache providerCache) {
+  Collection<Map> pendingOnDemandRequests(ProviderCache providerCache, @Nonnull Map<String, String> data) {
     def keys = providerCache.getIdentifiers(ON_DEMAND.ns)
     keys = keys.findResults {
       def parse = Keys.parse(it)
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AbstractAmazonLoadBalancerCachingAgent.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AbstractAmazonLoadBalancerCachingAgent.groovy
@@ -37,6 +37,8 @@ import com.netflix.spinnaker.clouddriver.cache.OnDemandMetricsSupport
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
+import javax.annotation.Nonnull
+
 import static com.netflix.spinnaker.cats.agent.AgentDataType.Authority.AUTHORITATIVE
 import static com.netflix.spinnaker.cats.agent.AgentDataType.Authority.INFORMATIVE
 import static com.netflix.spinnaker.clouddriver.core.provider.agent.Namespace.INSTANCES
@@ -121,14 +123,14 @@ abstract class AbstractAmazonLoadBalancerCachingAgent implements CachingAgent, O
   }
 
   @Override
-  boolean handles(OnDemandAgent.OnDemandType type, String cloudProvider) {
+  boolean handles(OnDemandAgent.OnDemandType type, String cloudProvider, @Nonnull Map<String, String> data) {
     type == OnDemandAgent.OnDemandType.LoadBalancer && cloudProvider == amazonCloudProvider.id
   }
 
   abstract CacheResult loadDataInternal(ProviderCache providerCache)
 
   @Override
-  Collection<Map> pendingOnDemandRequests(ProviderCache providerCache) {
+  Collection<Map> pendingOnDemandRequests(ProviderCache providerCache, @Nonnull Map<String, String> data) {
     return []
   }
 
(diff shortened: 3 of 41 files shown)


alperkokmen:
thanks for the review, @ezimanyi. i really appreciate it.

based on my understanding of the intention behind `OnDemandAgent` interface, i think it would be better to ship both of these changes. but, it's probably easier to review them separately 👍 

> I'm still not fully convinced that this is better than re-using the existing logic from #2523 that passes in an id.

let's table `Map<String, String>` argument for `OnDemandAgent#pendingOnDemandRequests` approach for now. we can revisit *if* we cannot leverage `id` argument.

i understand it's used as `keys` on the receiving end ([here](https://github.com/spinnaker/clouddriver/blob/72857c181f4c6ced9cd580c931ad0cd6c1087b7f/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java#L539-L553) and [here](https://github.com/spinnaker/clouddriver/blob/72857c181f4c6ced9cd580c931ad0cd6c1087b7f/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ClusterCachingAgent.groovy#L470-L484)); however, i didn't find any references to this endpoint in `orca`. [retrofit interface](https://github.com/spinnaker/orca/blob/0e4f4fcb961282c7cdda5f068b5644c655c0e379/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/CloudDriverCacheStatusService.groovy#L22-L26) doesn't even have `id` parameter in there.

in order to make `id` (passed as a single `String` query parameter) work for this, we need to serialize a `ScopedManifest` into a `String` which can then be interpreted on the `clouddriver`-side as an `InfrastructureKey`. so, the only way i see this working would be to make `orca` and `clouddriver` agree on a way to serialize & deserialize a `ScopedManifest` (e.g. join `account`, `location`, `name` by `:`).

1. `orca` would serialize `ScopedManifest` into a `String` which can be passed via `id` parameter.
1. `clouddriver` would receive the `id` parameter which would only get used for types that explicitly override `OnDemandAgent#pendingOnDemand(ProviderCache, String)`.
1. `KubernetesV2OnDemandCachingAgent` would override `OnDemandAgent#pendingOnDemand(ProviderCache, String)` in order to deserialize `id` into `account`, `location`, and `name` parts.

this would be a lot cleaner (in terms of number of lines touched); however, i am not too excited about `orca` having to know how to serialize a `ScopedManifest` into a `String`. is this what you had in mind?

> If anything it will be hard to beat agentIndex == 0 as short-circuit logic for the agents that shouldn't do anything.

i don't disagree; however, reading the `OnDemandAgent` interface, i thought the whole idea with `#handles` was to check _prior_ to calling `#handle` to avoid early returns such as [this](https://github.com/spinnaker/clouddriver/blob/26000a0ff96d192d3760be95a9a158eac322adf9/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2OnDemandCachingAgent.java#L278-L280), [this](https://github.com/spinnaker/clouddriver/blob/26000a0ff96d192d3760be95a9a158eac322adf9/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2OnDemandCachingAgent.java#L290-L292), [this](https://github.com/spinnaker/clouddriver/blob/26000a0ff96d192d3760be95a9a158eac322adf9/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2OnDemandCachingAgent.java#L295-L297), and [this](https://github.com/spinnaker/clouddriver/blob/26000a0ff96d192d3760be95a9a158eac322adf9/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2OnDemandCachingAgent.java#L314-L316).

doesn't it make more sense to let `#handles` determine whether or not an `OnDemandAgent` can handle a specific request (identified by `onDemandType`, `cloudProvider`, and `data`) rather than expecting each `OnDemandAgent` implement short-circuit logic?

finally, [handleReadRequests](https://github.com/spinnaker/clouddriver/blob/bd69f48051e0fbfcc54d1bb02f31cf26a6db6171/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/agent/KubernetesV2OnDemandCachingAgent.java#L393-L401) seems to be specifically implemented for kubernetes. it looked more like a workaround for the problem mentioned in the comment above.

ezimanyi:
I definitely agree that having `orca` and `clouddriver` agree on how to serialize a `ScopedManifest` is fragile, so we should not go with any solution that would require it.  My understanding of #2523 is that the call to `handle()` should now be returning the `id` of the key that was accepted for on-demand cache refresh, so we could update the code in `orca` to store this `id` and use it when checking on the pending status.

I agree it's hard to follow exactly what the intended pattern was as it looks like the corresponding change to `orca` was never implemented (or at least I can't seem to find it either).

One thing I realize looking at this again is that the `Map<String, String> data` you've added to the class to fetch pending refreshes map exactly onto the `Map<String, String>` that we send when requesting the refresh.  This wasn't obvious to me at first, but looking at it again if we do make it clear that the client should send the same data again when looking up the status of a refresh, then this is a more reasonable solution than I'd first thought and my concerns about that parameter are lessened/eliminated.  I do still think that we might want to follow the pattern of #2523 and add a function in the base class that handles the case of no `data` being passed and calls the existing `pendingRefreshes` function so we don't need to add it as a required parameter everywhere.

I think that moving the `account` and `namespace` check into the `handles()` function should significantly reduce the load.  (For `kubernetes` it's really `account` that will help as the `namespace` check will be equivalent to the `agentIndex == 0` though I agree that that is more readable sharding.)  If we do move forward with the `id`-based solution I think that would require parsing the incoming key somewhere in `clouddriver` so we have these fields to check against.  I suppose an advantage of the `Map<String, String>` solution is that we have the data readily available to check.

So overall I think I'm convinced that the broad solution of `Map<String, String>` is good.  I'd definitely like some input from @ajordens on this, particularly any context around #2523 and whether there was a corresponding change to `orca` or whether that perf improvement ended up being abandoned.

alperkokmen:
while we wait for @ajordens to chime in, i did look into how we could use the respnose from handle. unfortunately, it doesn't just return a list of `id`s; but, we may be able to extract what we need from the response. below is the response i got from `clouddriver` when i hit it manually via `curl`.

```
$ curl http://localhost:7002/cache/kubernetes/manifest -d'{"account":"mgmt-a", "location":"kayenta-staging", "name":"deployment kayenta-staging"}' -X POST -H 'Content-type: application/json' | jq .
{
  "cachedIdentifiersByType": {
    "artifact": [
      "kubernetes.v2:artifact:kubernetes/deployment:kayenta-staging:kayenta-staging:"
    ],
    "clusters": [
      "kubernetes.v2:logical:clusters:mgmt-a:kayenta:kayenta"
    ],
    "applications": [
      "kubernetes.v2:logical:applications:kayenta"
    ],
    "deployment": [
      "kubernetes.v2:infrastructure:deployment:mgmt-a:kayenta-staging:kayenta-staging"
    ]
  }
}

$ curl http://localhost:7002/cache/kubernetes/manifest -d'{"account":"mgmt-a", "location":"kayenta-staging", "name":"configMap kayenta-staging-config-5b03bdac:kayenta-staging:v077"}' -X POST -H 'Content-type: application/json' | jq .
{
  "cachedIdentifiersByType": {
    "artifact": [
      "kubernetes.v2:artifact:kubernetes/configMap:kayenta-staging-config-5b03bdac:kayenta-staging:v077"
    ],
    "configMap": [
      "kubernetes.v2:infrastructure:configMap:mgmt-a:kayenta-staging:kayenta-staging-config-5b03bdac-v077"
    ]
  }
}
```

so, it's a map of cache `id`s grouped by type where type will vary based on cloud provider. if we store this `Map<String, List<String>>` within `stageData` similar to `deployedManifests`, we _could_ reference it again in `ManifestForceCacheRefreshTask#checkPendingRefreshes(String, StageData, long)`. i am not sure what implications this has with regards to `stageData` growing in size though. assuming we proceed with the configuration-based approach @fkyc implemented in https://github.com/spinnaker/orca/pull/3314, implementation for the "optimized" method would use `cachedIdentifiersByType` to make calls to `CacheStatusService#pendingForceCacheUpdates`.

ideally, we would cross-reference `cachedIdentifiersByType` against `deployedManifests` to come up with the `id`s that we actually want to check. if `deployedManifests` only include `deployment` and `configMap` manifests, we would just pull those `id`s out.

alternatively, we could pull all `id`s from `cachedIdentifiersByType` and check every single one; but, i am not sure if this would be functionally correct as `artifact`, `clusters`, and `applications` aren't really manifests that would be deployed in our specific case.

alperkokmen:
@ezimanyi @ajordens happy new year!

i wanted to revive the discussion we were having with regards to how to proceed with these changes. could you weigh in on your preferred method for passing locations (i.e. namespaces) between orca and clouddriver? thanks!

ezimanyi:
@ajordens : Would love some input on this PR, particularly any context or background on #2523 which looks to be addressing a similar issue.  Thanks!

ajordens:
@ezimanyi and I talked about this in Slack earlier this week.

The motivations and impact of #2523 have since slipped my mind, certainly avoiding the scan + get is helpful. 

I don’t have strong feelings against `data` or in favor of `id` — my sentiment would be to try and actually remove the need to force cache refresh. We’ve been able to successfully remove them from our AWS / Titus deploy paths which (a) eliminated the overhead and (b) resulted in faster deploy pipelines.

Historically the on demand controller showed up fairly consistently as a high-latency offender. We did experiment with various cleanups and optimizations before just dropping the need to even call it.

ezimanyi:
I've created spinnaker/spinnaker#5607 to account for some work greatly improving (and in many cases removing) force cache refresh for the Kubernetes V2 provider, that I think will eliminate the need for this PR.  In any case, given that this is against a branch, it would need to be ported to `master` to merge...happy discuss more or collaborate on any improvements, but for now going to close this in favor of the larger refactor that I plan on doing in the near future.

alperkokmen:
@ezimanyi thanks for the update. we actually have been running this version since november 2019 and without it we weren't able to scale successfully since concurrent executions were creating a huge load on redis, rendering it unusable for several minutes and deployments timing out.

it is nice that there are plans to mitigate the performance issue; but, assuming this doesn't conflict with your plans (doesn't seem like it as far as i can tell after reading https://github.com/spinnaker/spinnaker/issues/5607), we would still like to upstream these changes to master. that way, we can safely upgrade to newer versions (1.18.x or higher) without having to worry about the concurrent pipeline execution bottleneck.

ezimanyi:
I don't think this should conflict with my plans; my performance improvements might end up making changes to what is done here and/or removing part of it, but certainly it's reasonable to apply these first then I can refactor as needed.  Do you mind opening a PR against `master` with these changes then?  (And the same for spinnaker/orca#3314).  Thanks!

alperkokmen:
@ezimanyi sounds good, will do! thanks!

PR #4182, opened 2019-11-25
fix(kubernetes): do not throw NPE on NetworkPolicies with ingresses w…

gregjones:
…ith null ports

Similar to #4172, the code is assuming that a NetworkPolicy that has an ingress
will return an empty list of ports, but instead it defaults to null for this[1]
and so crashes.

[1] https://github.com/kubernetes-client/java/blob/master/kubernetes/src/main/java/io/kubernetes/client/openapi/models/V1NetworkPolicyIngressRule.java#L45

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2SecurityGroup.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2SecurityGroup.java
@@ -32,9 +32,13 @@ import com.netflix.spinnaker.clouddriver.model.SecurityGroup;
 import com.netflix.spinnaker.clouddriver.model.SecurityGroupSummary;
 import com.netflix.spinnaker.clouddriver.model.securitygroups.Rule;
 import io.kubernetes.client.models.V1NetworkPolicy;
+import io.kubernetes.client.models.V1NetworkPolicyEgressRule;
+import io.kubernetes.client.models.V1NetworkPolicyIngressRule;
 import io.kubernetes.client.models.V1NetworkPolicyPort;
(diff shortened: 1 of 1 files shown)


gregjones:
@maggieneterval & @ezimanyi as author/reviewer of the related PR, would you be kind enough to review this one?

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/model/KubernetesV2SecurityGroup.java
```diff
@@ -119,7 +120,13 @@ public static KubernetesV2SecurityGroup fromCacheData(CacheData cd) {
       return Collections.emptySet();
     }
     return policy.getSpec().getIngress().stream()
-        .map(i -> i.getPorts().stream().map(KubernetesV2SecurityGroup::fromPolicyPort))
+        .map(
```
ezimanyi: The original code was nesting streams, which made things hard to read (possibly contributing to the bug).  It would be great to clean this up into a single stream along with this fix.  Something like the following should be equivalent and also fix the NPE:
```
    return policy.getSpec().getIngress().stream()
      .map(V1NetworkPolicyIngressRule::getPorts)
      .filter(Objects::nonNull)
      .flatMap(Collection::stream)
      .map(KubernetesV2SecurityGroup::fromPolicyPort)
      .collect(Collectors.toSet());
```
Do you mind changing this (and the stream below) to use that single-stream style?

gregjones: Not at all, that version is much nicer. Changed with 2757398ebece6c43f22034c53f727c2ea8cb8469

ezimanyi:

PR #4183, opened 2019-11-25
fix(docker): Fix sorting array of non-comparable types

ezimanyi:
Fixes spinnaker/spinnaker#5002.  The second commit is what actually fixes the bug, but I wanted to clean up some of the unsafe groovy and add tests, so this PR is a bit bigger than just that commit.

* test(docker): Add tests on docker registry tag fetching 

  There are currently no tests that the fetching of docker registry tags and the sorting of these tags by creation date works properly; add some tests before touching that code.

* fix(docker): Fix sorting array of non-comparable types 

  getCreationDate returns an Instant; when the call fails, we're returning a default value as a Date. This means we can end up with an array with a mix of Instant and Date objects, which don't together obey the general Comparable contract.

  Fix this by returning an Instant from the catch block; also pull the error handling logic into a CompileStatic function so we get some additional type safety.

* refactor(docker): Pull key-based sorter into its own class 

  When sorting docker tags by creation date, we are effectively doing a Schwartzian transform to map each item to a key, sort by that key, then re-map back to the original item. Rather than have this ad-hoc in Groovy, make a simple Java class that handles this algorithm and that can be tested. This also allows us to mark the method that contained this logic as CompileStatic, as we've removed the non-safe operations into Java (and made them safe).

  Right now, leaving the class alongside the single place it is used, but if we need this in other places we can move it to a more general location.

### DIFF
--- a/clouddriver-docker/clouddriver-docker.gradle
+++ b/clouddriver-docker/clouddriver-docker.gradle
@@ -3,9 +3,13 @@ dependencies {
   implementation project(":clouddriver-security")
   implementation project(":cats:cats-core")
 
+  compileOnly "org.projectlombok:lombok"
+  testCompile "org.projectlombok:lombok"
+
   implementation "org.springframework.boot:spring-boot-actuator"
   implementation "org.springframework.boot:spring-boot-starter-web"
   implementation "org.codehaus.groovy:groovy-all"
+  implementation "com.google.guava:guava"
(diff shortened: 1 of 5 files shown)


jonsie:
This looks useful.  Is there a reason for this to be in `clouddriver-docker` as opposed to `clouddriver-core`?

PR #4184, opened 2019-11-25
fix(thread): make sure to provide NamedThreadFactory to all executors

marchello2000:
It's a good practice to name all the threads/executors so that it's easy to see what threads are running away (if any).
Add NamedThreadFactory to all places we create a new executor

### DIFF
--- a/clouddriver-kubernetes-v1/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/ops/servergroup/AbstractEnableDisableKubernetesAtomicOperation.groovy
+++ b/clouddriver-kubernetes-v1/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/ops/servergroup/AbstractEnableDisableKubernetesAtomicOperation.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.clouddriver.kubernetes.v1.deploy.ops.servergroup
 
+import com.netflix.spinnaker.cats.thread.NamedThreadFactory
 import com.netflix.spinnaker.clouddriver.data.task.Task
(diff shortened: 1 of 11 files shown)


cfieber:
there is also a constructor-created ExecutorService in DefaultOrchestrationProcessor that could use a name

marchello2000:
> there is also a constructor-created ExecutorService in DefaultOrchestrationProcessor that could use a name

Thanks, update it and `PooledResourceQueue`

clouddriver-kubernetes-v1/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/ops/servergroup/AbstractEnableDisableKubernetesAtomicOperation.groovy
```diff
@@ -141,7 +142,7 @@ abstract class AbstractEnableDisableKubernetesAtomicOperation implements AtomicO
 
     task.updateStatus basePhase, "Resetting service labels for each pod..."
 
-    def pool = Executors.newWorkStealingPool((int) (pods.size() / 2) + 1)
+    def pool = Executors.newWorkStealingPool((int) (pods.size() / 2) + 1, new NamedThreadFactory(AbstractEnableDisableKubernetesAtomicOperation.class.getSimpleName()))
```
ttomsu: This change started throwing errors in the integration test, specifically:

```
No signature of method: static java.util.concurrent.Executors.newWorkStealingPool() is applicable for argument types: (Integer, com.netflix.spinnaker.cats.thread.NamedThreadFactory) values: [1, com.netflix.spinnaker.cats.thread.NamedThreadFactory@3d3bb4d1]
Possible solutions: newWorkStealingPool(), newWorkStealingPool(int).
```

PR #4185, opened 2019-11-26
fix(titus): shutdown the executor for titus streaming agent

marchello2000:
We are leaking an executor (with a single thread in) from the TitusStreamingAgent because it's not shutdown properly
I think it would be better to refactor this code to not create the executor (at least not all the time), but I want to make a minimal change to address the immediate problem first.

There is also some unnamed thread pool that appears to be leaking that I can't track down. Perhaps it's some API that the streaming agent calls, will debug it more after this change goes in


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
```diff
@@ -306,6 +306,7 @@ public void executeAgent(Agent agent) {
           getTimeoutMillis(),
           TimeUnit.MILLISECONDS);
       CompletableFuture.completedFuture(handler).join();
+      executor.shutdown();
```
dreynaud: 😳

marchello2000:
I am going to come back and fix the cancellation bit (since it's not actually working today because the threadpool is configured with 1 max thread so the canceller can't get scheduled till the main work completes - sort of defeats the purpose)

PR #4189, opened 2019-11-27
feat(mergify): Allow OSS approvers to auto-submit without review

ttomsu:
We can click the button to merge anyway, so this just makes maintenance life a bit easier. Thoughts?


ttomsu:
Going to apply this same pattern to all repos. Follow along on https://github.com/spinnaker/spinnaker/issues/5189

PR #4200, opened 2019-12-05
fix(provider/kubernetes): validate spec.template is a map before cast

ivanphdz:
Currently a CRDs from Istio kind expect a string as value of spec.template instead of a map, this cause an error in clouddriver at the moment to cast from object to map to get the annotations inside spec.template.

Issue: https://github.com/spinnaker/spinnaker/issues/4664

Example manifest: https://github.com/signalfx/signalfx-istio-adapter/blob/master/resources/metric-instances.yaml

Reference: https://istio.io/docs/reference/config/policy-and-telemetry/istio.policy.v1beta1/#Instance

### DIFF
--- a/clouddriver-kubernetes-v2/src/test/resources/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/crd-manifest.json
+++ b/clouddriver-kubernetes-v2/src/test/resources/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/crd-manifest.json
@@ -8,5 +8,6 @@
     "name": "default-custom1"
   },
   "spec": {
+    "template": "template1"
   }
 }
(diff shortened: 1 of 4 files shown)


clouddriver-kubernetes-v2/src/test/resources/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/crd-manifest.json
```diff
@@ -8,5 +8,6 @@
     "name": "default-custom1"
   },
   "spec": {
+    "template": "template1"
```
ezimanyi: It might be better to add a new manifest with the properties you want instead of updating this existing one.  The reason I suggest this is that now we're losing test coverage of the case where `spec.template` is `null`.  It might be good to leave this one having `spec.template` `null` and then add another one with `spec.template` a String and test that neither throws an NPE.

ivanphdz: thanks for your help and feedback

PR #4206, opened 2019-12-07
fix(cfn): do not add roleARN if empty or null

xavileon:
If no RoleARN is provided via Deck or in a templated pipeline, the
creation/updating of stacks fails, as en empty value is not allowed by
the AWS API. This patch handles this case by including this field
into the API call only when it's a non-empty string.


xavileon:

xavileon:

PR #4209, opened 2019-12-10
fix(sagas): Use a Set instead of List to avoid duplicate Task SagaIds

jonsie:
So we don't add duplicate SagaIds to tasks (and thereby double the number of atomic operations we retry), use a Set instead of a List.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/task/DefaultTask.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/data/task/DefaultTask.groovy
@@ -104,8 +104,8 @@ public class DefaultTask implements Task {
   }
 
   @Override
-  List<SagaId> getSagaIds() {
-    return sagaIdentifiers.toList()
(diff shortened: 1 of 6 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
```diff
@@ -33,13 +33,7 @@
...
 import com.netflix.spinnaker.security.AuthenticatedRequest;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
```
srekapalli: wildcard import here. ?

jonsie: Argh come on Inteliij, I mean... come on.

PR #4222, opened 2019-12-12
feat(titusJobSearch): Support searching with titus jobIds

aravindmd:
- Adding support to search based on titus jobIds.
- The `jobCache` here is populated only when `titus` is enabled 
- The `cache` has a mapping for the `jobId:serverGroupKey`
- `types=jobIds` to be used for this search

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsSearchProvider.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsSearchProvider.groovy
@@ -137,6 +137,7 @@ class CatsSearchProvider implements SearchProvider, Runnable {
     } catch (Exception e) {
       log.error("Unable to refresh cached identifiers (instances)", e)
     }
+    //Populate jobId cache only when titus is enabled
     if (titusEnabled) {
       try {
         log.info("Refreshing Cached Identifiers (jobIds)")
@@ -145,6 +146,7 @@ class CatsSearchProvider implements SearchProvider, Runnable {
           provider.supportsSearch('serverGroups', Collections.emptyMap())
         }.collect { provider ->
           def cache = providerRegistry.getProviderCache(provider.getProviderName())
+          //Populate job cache based of the ids that are mapped to titus server groups
           def ids = cache.filterIdentifiers("serverGroups", "titus*")
           ids.collect {
             def job = cache.get("serverGroups", it)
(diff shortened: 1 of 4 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsSearchProvider.groovy
```diff
@@ -50,17 +51,16 @@ class CatsSearchProvider implements SearchProvider, Runnable {
...
-    [:]
-  )
+  private final ConcurrentHashMap<String, Collection<String>> cachedIdentifiersByType = new ConcurrentHashMap<>()
 
   private final FiatPermissionEvaluator permissionEvaluator
   private final List<KeyParser> keyParsers
 
   private final ScheduledExecutorService scheduledExecutorService
 
+  @Value('${titus.enabled:false}')
```
robzienert: Do not use `@Value` annotations. Create a `ConfigurationProperties` class and inject that in instead.

robzienert: Also, no. Do not leak cloud providers into core code.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsSearchProvider.groovy
```diff
@@ -128,13 +130,39 @@ class CatsSearchProvider implements SearchProvider, Runnable {
...
       log.info("Refreshed Cached Identifiers (found ${instanceIdentifiers.size()} instances)")
     } catch (Exception e) {
       log.error("Unable to refresh cached identifiers (instances)", e)
     }
+    //Populate jobId cache only when titus is enabled
+    if (titusEnabled) {
+      try {
+        log.info("Refreshing Cached Identifiers (jobIds)")
+        def jobCache = [:]
+        def instanceIdentifiers = providers.find { provider ->
```
asher: I think this should be called `serverGroupIdentifiers`

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsSearchProvider.groovy
```diff
@@ -128,13 +130,39 @@ class CatsSearchProvider implements SearchProvider, Runnable {
...
+          provider.supportsSearch('serverGroups', Collections.emptyMap())
```
asher: @robzienert's feedback that cloud provider specific logic should be avoided in core is valid. This specific line shows the way, in that should should be able to provide a `SearchProvider` in `clouddriver-titus` that implements the following logic.

aravindmd:
Going to address some of these comments soon

robzienert:
Ping.

aravindmd:
Closing this , this is prioritized for now , will revisit and address the comments here.

PR #4225, opened 2019-12-14
fix(tests): Fixed failing SagaService test and also enable Junit5 pla…

srekapalli:
- Fixed  failing test.
- Also enabled JUnit5 platform to execute tests in all sub projects.


srekapalli:
> Looks like something failing in CI -- otherwise seems alright.

Fixed now.

PR #4227, opened 2019-12-16
feat(config/dynamic): Support for local dynamic config files

ncknt:
- When `dynamic-config.enabled` is true, make new PropertySource from `dynamic-config.files` file references
- Observe the files and refresh RefreshScope and application context upon changes/deletion.

This PR adds support to reload config files that have not been added via Spring Cloud Config. It is more lightweight because it only refreshes `@RefreshScope` configuration and doesn't go through the whole SCC environment refresh.

The use case is to reload accounts defined in a file that is mounted on the container and not managed by SCC. e.g. a direct mount from a Kubernetes secret or a file managed by the Spinnaker operator.

*This is not the final PR*: If it makes sense, I'd make a PR to `kork-config/` so it can be used in all services where dynamic config makes sense.

@robzienert @ezimanyi @scottfrederick


ethanfrogers:
Didn't @scottfrederick add support for this in [this Kork PR](https://github.com/spinnaker/kork/pull/413)?

ncknt:
As far as I understand the kork PR still requires SCC to be set up. I've had a few issues with SCC (like deletion not being reflected, all the contexts being quite intensive to refresh) so I was looking for a simpler method. It doesn't have the bells and whistle of SCC and won't let you store your config in git. It just watches files and reload them.

scottfrederick:
> Didn't @scottfrederick add support for this in this Kork PR?

I do think there's overlap between Spring Cloud Config's `native` backend and what this PR aims to provide. 

> It is more lightweight because it only refreshes `@RefreshScope` configuration and doesn't go through the whole SCC environment refresh.

What are the parts of the SCC environment refresh that this feature aims to avoid? By far the most expensive part of the refresh process (as well as initial application load) is the parsing of YAML file contents to an `@ConfigurationProperties` model object, and this change doesn't appear to avoid that part. 

> If it makes sense, I'd make a PR to `kork-config/` so it can be used in all services where dynamic config makes sense.

My biggest concern with this is that it adds yet another approach to externalizing Spinnaker configuration, in addition to the `kork-secrets` and Spring Cloud Config-backed `kork-config` features that are already in place. Will an additional solution to similar problems confuse users? Is there a way we could provide the benefits that this PR aims to provide with one of the existing external config mechanisms?

scottfrederick:
> As far as I understand the kork PR [adding support for Spring Cloud Config `native` backend] still requires SCC to be set up. 

Yes, that's the case. 

> I've had a few issues with SCC (like deletion not being reflected...

That sounds like something we should have an issue to fix, regardless of whether this PR is accepted or not. 

> ... all the contexts being quite intensive to refresh)

Again, parsing large amounts of YAML is definitely intensive - no disagreement about that. What other parts of the Spring Cloud Config refresh have been troublesome?

ncknt:
> What are the parts of the SCC environment refresh that this feature aims to avoid?
That part: https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/ContextRefresher.java#L91
It's basically going through all the property sources, making a copy of all the values (some of which may be computationally intensive), and compare to the new values. I get why it's being done that way as there's no way for the SCC client to know what has changed but this is repeated every 30s and is quite intensive. I think parsing YAML is a small part of that but I don't have numbers to back it up - just debugger sessions :)

More importantly, we've had customers running into issues with SCC such as:
- the issue mentioned above
- a hard time setting up SCC with the right options
- can't refer to secrets as the SCC is exposing its own map of changes at runtime and that doesn't get wrapped in either CloudConfigAware or SecretAware property source.

robzienert:
We need to get on the same page WRT configuration for Spinnaker.

It seems to me like this flavor of discussion has been spinning through multiple PRs and Issues over months and there's two groups diverging on a piece of functionality that is categorically undifferentiated heavy-lifting development. It's my personal opinion that we should be embracing functionality that is offered via Spring (making improvements to it where necessary), rather than building something bespoke that we will have to maintain alone ad infinitum.

That said, we need a single solution that we agree on and are happy with, I'd be keen to have @ncknt, @scottfrederick and whomever else is interested join the next Platform SIG to discuss and we can decide the direction as a group.

robzienert:
Clarifying: I don't think that this PR should be blocked on this conversation. The state of the world today is that there are two solutions that are being used actively and this PR supports one of those two solutions.

I'd say let's not stop momentum, but come prepared to talk the pro's and con's of the solutions we have in front of us and figure out how we can move forward in 2020 towards a single solution.

scottfrederick:
> there are two solutions that are being used actively and this PR supports one of those two solutions

Is that the case? I assume you're talking about `kork-secrets` and `kork-config`, and it looked to me like this PR was a third option that is configured differently from and behaves differently from either of those existing solutions.

robzienert:
My understanding, which should be corrected if wrong, is that this supports the existing `kork-secrets` work.

ncknt:
This PR enables local reload of config in a `kork-secrets` compatible way (you can reference a secret in there). For context, this is a stopgap solution to enable dynamic Kubernetes accounts via the Kubernetes operator. This is not meant for end users and we don't need to advertise that flag and can mark it as experimental if there's such a thing.

Once we resolve all the pending issues with SCC, I'd be happy to remove/deprecate it and converge on a solution.

robzienert:
There is `kork-annotations`, which has both an `@Alpha` and a `@Beta` annotation. I haven't really advertised these much outside of the plugins work - but use them as needed.

ncknt:
I've added the `@Alpha` annotation - hopefully in a good way. I'll be on vacation the next 2 weeks. It'd be great if this could be merged before 1.18 is cut.

PR #4228, opened 2019-12-18
fix(saga): Removing the `list` endpoints as the current implementation is a performance liability

ajordens:


ajordens:
Talked about this offline with @robzienert and @jonsie -- we'll revisit.

PR #4229, opened 2019-12-18
fix(saga): Handle concurrency issue when same op is sent more than once

robzienert:
We experienced an issue with sagas where Orca submitted the same operation into clouddriver at the same time (separated by a few milliseconds). The second request returned a 500 due to the SQL integrity constraint violation, but the work was already being handled successfully by the other clouddriver instance.

What _should_ happen is that the second instance returns a pointer to the original task so that Orca doesn't have to reason about its mistake and just carry on monitoring the operation that's being performed.

Also changed the duration that Kato Tasks are kept around. 1 hour is too short to be able to get notified of an error and to diagnose. 4 days seems a reasonable default so that if an error occurs EOW or on the weekend, there's still sufficient time to get information the following week.


ajordens:
Thanks for increasing the task cleanup interval.

PR #4232, opened 2019-12-24
feat(sql): Add cache cleanup agent for removed accounts

robzienert:
Adds a background agent that regularly scans the database for cache records that are
related to caching agents that do not exist anymore, and deletes them. This addresses
https://github.com/spinnaker/spinnaker/issues/4803

The originally discussed plan was to add an admin endpoint that people would call to cleanup the database, but after re-reading the thread, I feel like this option is lowest impact on end-users, even if it's more costly.


cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCacheCleanupAgent.kt
```diff
@@ -0,0 +1,131 @@
...
+          while (rs.next()) {
+            if (!agentTypes.contains(rs.getString(2))) {
+              idsToClean.add(rs.getString(1))
+            }
+          }
+
+          if (idsToClean.isNotEmpty()) {
+            log.info("Found {} records to cleanup from '{}'", idsToClean.size, relTable)
+            idsToClean.chunked(100) { chunk ->
+              // TODO(rz): Add retry
```
jonsie: Looks like resilience4j is already used here in `sql-cats`, I would recommend doing it like this:

https://github.com/spinnaker/clouddriver/blob/master/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/sql.kt#L36

maggieneterval:
>  The originally discussed plan was to add an admin endpoint that people would call to cleanup the database, but after re-reading the thread, I feel like this option is lowest impact on end-users, even if it's more costly.

+1, especially since from the perspective of a Kubernetes end-user with high account churn who is migrating their Clouddriver to the MySQL backend, it likely would have been a less than excellent experience to need to discover and enable a feature flag just for it to work as well as the Redis backend in terms of cleanup. Thank you so much, @robzienert! :hugs:

robzienert:
🎉 This PR is ready to review.

cats/cats-sql/cats-sql.gradle
```diff
@@ -27,10 +27,6 @@ dependencies {
   implementation project(":clouddriver-security")
   implementation project(":clouddriver-sql")
 
-  compileOnly "org.projectlombok:lombok"
-  annotationProcessor "org.projectlombok:lombok"
-  testAnnotationProcessor "org.projectlombok:lombok"
-
```
robzienert: Java isn't used anywhere in this module, so these deps were unnecessary.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -1,7 +1,6 @@
 package com.netflix.spinnaker.cats.sql.cache
```
robzienert: Everything in this file is a result of moving some internal naming convention code into a new `SqlNames` class.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCacheCleanupAgent.kt
```diff
@@ -0,0 +1,181 @@
...
+  private val providerRegistry: ProviderRegistry,
+  private val jooq: DSLContext,
+  private val registry: Registry,
+  private val sqlNames: SqlNames
+) : RunnableAgent, CustomScheduledAgent {
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  private val deletedId = registry.createId("sql.cacheCleanupAgent.dataTypeRecordsDeleted")
+  private val timingId = registry.createId("sql.cacheCleanupAgent.dataTypeCleanupDuration")
```
robzienert: Longer metric names. I'm often left having no other option but read code to know what some of our metric names actually mean (e.g. `sql.cacheCleanupAgent.deleted` would probably be a likely metric name elsewhere).

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlCacheConfiguration.kt
```diff
@@ -161,6 +163,17 @@ class SqlCacheConfiguration {
   ): SqlCleanupStaleOnDemandCachesAgent =
     SqlCleanupStaleOnDemandCachesAgent(applicationContext, registry, clock)
 
+  @Bean
+  @ConditionalOnProperty(value = ["sql.cache-cleanup-agent.enabled"], matchIfMissing = true)
```
asher: Not sure if this should be on by default, at least initially. Could be flipped to on by default once demonstrated to be non-disruptive to installs that don't need it, or in the second OSS release that contains the cleanup agent, with enough opportunity to advise a config change.

robzienert: Good point - I like this rollout strategy. Thanks!

robzienert:
**Help needed**: Sorry for the delay on this PR merging. I've been having trouble getting this to work locally. Everything wires up correctly when configured to do so, but the agent never gets run. Has anyone seen this before? It's written - as far as I can tell - in the same convention as other core agents, and the other ones are working fine. 🤷‍♂

Would like to verify against our envs before merging.

ezimanyi:
@robzienert : It looks like the agent is never actually being scheduled by calling `AgentScheduler.schedule(...)` with the SQL caching agent.  It looks like this happens automatically for agents that are owned by a `Provider` in the `ProviderRegistry` [here](https://github.com/spinnaker/clouddriver/blob/008e111d64c11a0448921c888635459637a37b00/cats/cats-core/src/main/java/com/netflix/spinnaker/cats/agent/AgentController.java#L34) but given that the SQL agent isn't owned by a provider, it will probably be necessary to manually schedule this agent (or somehow make the logic in `AgentController` more general so it's not tied to a provider).  I could be wrong about this, but that's my guess of what's happening here.

robzienert:
Thanks! I'll take a look tonight and hopefully get this across the finish line finally. :tada:

robzienert:
This functionality defaults to disabled. It can be enabled with `sql.unknown-agent-cleanup-agent.enabled=true`

PR #4233, opened 2019-12-27
Trigger compilation whenever lombok config is updated

theosotr:
Hello

This pull request enhances the build script of the project.
Notably, compilation uses the lombokproject which depends on the configuration `lombok.config`.
When there is any update to this configuration file, compilation of source files should be re-triggered in case of incremental builds.

This commit registers `lombok.config` as input to the Gradle tasks that are responsible for compilation.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -57,6 +57,12 @@ allprojects {
     apply plugin: 'nebula.kotlin'
     apply plugin: "kotlin-allopen"
 
+    def lombokConf = "$rootProject.projectDir/lombok.config"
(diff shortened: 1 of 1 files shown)


plumpy:
This is interesting and definitely a great idea, but I think ideally it would live in `spinnaker-gradle-plugin`. So that we didn't have to add it to all six of our projects. I'll see if I can figure that out and merge this if not...

Thanks again!

robzienert:
@plumpy Did you get anywhere on this for the gradle plugin? Will leave clicking the button to you.

PR #4235, opened 2019-12-31
feat(kubernetes): add initContainer logs to kubernetes v2 provider container logs

sgarlick987:
@maggieneterval @ethanfrogers I've gone ahead and made the changes @maggieneterval  and I discussed yesterday. Let me know what you think.

This is the first step that partially implements https://github.com/spinnaker/spinnaker/issues/4869 by adding initContainer logs to the existing logs being returned. We add all the initContainer logs before the current container logs so they'll show up in deck with initContainers to the left of the existing ones (screenshot in issue). Differentiating the different types of containers in the gui will come in a later PR for clouddriver and deck.

This also handles the refactoring mentioned by @maggieneterval on the issue. I ended up leaving the current ContainerLogs where it was in core and deprecating since the v1 provider was using it, and I opted to not many any code changes to v1. The new ContainerLogs lives in the shared kubernetes project and is immutable and used in the v2 provider now. Let me know what you think, and I can change how I handled it if needed.

Lastly this adds tests KubernetesV2InstanceProviderTest that cover the new and existing code in KubernetesV2InstanceProvider.

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2InstanceProvider.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2InstanceProvider.java
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.view.provider;
 
 import com.netflix.spinnaker.cats.cache.CacheData;
 import com.netflix.spinnaker.clouddriver.kubernetes.KubernetesCloudProvider;
+import com.netflix.spinnaker.clouddriver.kubernetes.model.ContainerLog;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.Keys;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.agent.KubernetesCacheDataConverter;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.view.model.KubernetesV2Instance;
@@ -27,13 +28,16 @@ import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.Kube
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.KubernetesManifest;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.op.job.KubectlJobExecutor;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.security.KubernetesV2Credentials;
-import com.netflix.spinnaker.clouddriver.model.ContainerLog;
 import com.netflix.spinnaker.clouddriver.model.InstanceProvider;
 import io.kubernetes.client.models.V1Container;
+import io.kubernetes.client.models.V1ObjectMeta;
 import io.kubernetes.client.models.V1Pod;
-import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import javax.annotation.Nonnull;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.tuple.Pair;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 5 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ContainerLog.java
```diff
@@ -1,7 +1,7 @@
 /*
- * Copyright 2018 Google, Inc.
+ * Copyright 2019 Netflix, Inc.
```
ethanfrogers: No need to change the header here!

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ContainerLog.java
```diff
@@ -1,7 +1,7 @@
...
  *
- * Licensed under the Apache License, Version 2.0 (the "License")
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
ethanfrogers: Extraneous `;`.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/model/ContainerLog.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2019 Netflix, Inc.
```
ethanfrogers: This should probably be Discovery, Inc.

clouddriver-kubernetes-v2/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2InstanceProviderTest.java
```diff
@@ -0,0 +1,194 @@
+/*
+ * Copyright 2019 Netflix, Inc.
```
ethanfrogers: This should probably be Discovery, Inc.

sgarlick987:
@ethanfrogers ah ok, I'd noticed those were being generated/update automatically so I assumed to just leave them. I'll go ahead and update those now.

sgarlick987:
@ethanfrogers those licenses are cleaned up now

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2InstanceProvider.java
```diff
@@ -27,15 +29,16 @@
...
 import com.netflix.spinnaker.clouddriver.model.InstanceProvider;
 import io.kubernetes.client.models.V1Container;
+import io.kubernetes.client.models.V1ObjectMeta;
 import io.kubernetes.client.models.V1Pod;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.tuple.Pair;
+import org.jetbrains.annotations.NotNull;
```
maggieneterval: Any opposition to the IDE-agnostic alternative? I see we are using the `jetbrains` NotNull a few places elsewhere in Clouddriver, but I think the `javax` one is more frequently used.

```suggestion
import javax.annotation.Nonnull;
```

sgarlick987: No opposition here, I happened to see the other one being used and just went with that. I'll update this.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ContainerLog.java
```diff
@@ -23,6 +23,11 @@
 @AllArgsConstructor
 @Data
 @NoArgsConstructor
+@Deprecated
```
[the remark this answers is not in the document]
sgarlick987: I'll move it over to kubernetes-v1 since that would be the smaller of the two changes.

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2InstanceProvider.java
```diff
@@ -82,49 +85,76 @@ public KubernetesV2Instance getInstance(String account, String location, String
   }
 
   @Override
-  public List<ContainerLog> getConsoleOutput(String account, String location, String fullName) {
-    Optional<KubernetesV2Credentials> optionalCredentials = accountResolver.getCredentials(account);
+  public List<ContainerLog> getConsoleOutput(String account, String namespace, String fullName) {
+    final Optional<KubernetesV2Credentials> optionalCredentials =
```
ezimanyi: We don't tend to mark local variables as `final` in the Spinnaker codebase, though I suppose this is more of a personal preference thing.  We definitely *do* mark fields `final` wherever possible, but in the case of local variables in short methods I think the benefit of the compiler preventing reassignment is not really outweighed by the visual noise.  That being said, I know opinions differ on this, so either way is fine.

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2InstanceProvider.java
```diff
@@ -82,49 +85,76 @@ public KubernetesV2Instance getInstance(String account, String location, String
...
+    return getPodLogs(credentials, pod);
+  }
+
+  @Nonnull
+  private List<ContainerLog> getPodLogs(
+      @Nonnull final KubernetesV2Credentials credentials, @Nonnull final V1Pod pod) {
+    final List<V1Container> initContainers = pod.getSpec().getInitContainers();
+    final List<V1Container> containers = pod.getSpec().getContainers();
 
+    final List<ContainerLog> result = new ArrayList<>(containers.size() + initContainers.size());
```
ezimanyi: This might be a bit clearer using streams (which in a number of places in our code are overused, but which I think could help here).  Something like the following could eliminate a lot of the intermediate arrays, and also remove the need for the `getContainerLogs` which is just looping over containers:
```
return Stream.concat(initContainers.stream(), containers.stream())
    .map(container -> getContainerLog(credentials, pod, container))
```

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2InstanceProvider.java
sgarlick987: I think its worth matching the rest of the codebase, I'll update this tomorrow morning

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2InstanceProvider.java
sgarlick987: I'll update this in the morning with the above.

sgarlick987: I think this made it a lot more clear, thanks for the tip.

PR #4239, opened 2020-01-09
fix(kubernetes): Fix stability for stateful/daemon set

ezimanyi:
Fixes spinnaker/spinnaker#5283.

* test(kubernetes): Add tests to stateful/daemon set stability 

  The next commit will fix bugs in the stateful set and daemon set stability conditions. Add tests to cover the cases that will be fixed; in the case of the daemon set handler, there were no existing tests so this commit adds some basic tests in addition to tests of the bug that will be fixed.

  This commit also does some minor cleanup on the stateful set tests, removing some unused variables and referring to member variables by their name instead of via their groovy-generated accessors.

* fix(kubernetes): Fix stability for stateful/daemon set 

  When scaling a stateful set down to zero replicas, we never consider the stateful set stable. This is because the status field for a stateful set with no replicas omits some fields
  (instead of setting them to 0) and our logic does not properly handle the null case. Fix this and add some tests to validate the fix.

  A similar bug exists with daemon sets; for daemon sets, you don't explicitly set replicas but if you configure the daemon set such that it runs on 0 nodes (for example by including a node selector that doesn't match anything), Spinnaker will never consider the daemon set stable.

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesDaemonSetHandler.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesDaemonSetHandler.java
@@ -34,6 +34,7 @@ import io.kubernetes.client.models.V1beta2DaemonSet;
 import io.kubernetes.client.models.V1beta2DaemonSetStatus;
 import java.util.Map;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import org.springframework.stereotype.Component;
 
 @Component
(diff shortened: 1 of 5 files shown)


ezimanyi:
This exact bug was fixed for deployments in spinnaker/clouddriver#2848 and for replica sets in spinnaker/clouddriver#3248.  I looked through all of the `KubernetesHandler`s and these were the only two others that any similar logic (comparing number of something to desired number of something) so I think this general class of bug should now be fixed.

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesDaemonSetHandler.java
```diff
@@ -123,27 +124,32 @@ private Status status(V1beta2DaemonSet daemonSet) {
...
+    existing = defaultToZero(status.getNumberReady());
+    if (desiredReplicas > existing) {
       return result.unstable("Waiting for all replicas to be ready");
     }
 
     return result;
   }
+
+  // Unboxes an Integer, returning 0 if the input is null
+  private int defaultToZero(@Nullable Integer input) {
```
ezimanyi: This looked a bit cleaner than doing `Optional.ofNullable(status.someField()).orElse(0)` everywhere.  It also has the really minor benefit of not boxing then unboxing the `0` every time we default, but I did this mostly for readability.  The same function is in two classes, but it seemed tiny enough and trivial enough to no be worth trying to pull out to somewhere more general :man_shrugging:

PR #4243, opened 2020-01-09
fix(appengine): Propagate gcloud failures from job executor to task.

duftler:
Without this change, executions that fail while emitting only stderr output, but not stdout output, would be considered successes and the stderr output would be squashed.


duftler:
Fixes https://github.com/spinnaker/spinnaker/issues/5293.

PR #4247, opened 2020-01-09
fix(kubernetes): Core caching agent is authoritative for artifacts

ezimanyi:
Fixes spinnaker/spinnaker#5274.

There is currently no caching agent that is authoritative for kubernetes artifacts; this is causing them to fail to be cached by the SQL provider (which strictly enforces that only caching agents can only write types for which they are authoritative).

The KubernetesCoreCachingAgent should be authoritative for artifacts, just as it is for other logical kinds (clusters and applications).


jasonmcintosh:
Hrmm, can this be cherry picked to 1.17?

alchen99:

PR #4249, opened 2020-01-10
fix(kubernetes): return provider field with kubernetes /search results

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/5068

Deck expects each security group returned from `/search` to have a `provider`, `cloudProvider`, or `type` field corresponding to the correct cloud provider in order to render them in the Firewalls tab. [This commit](https://github.com/spinnaker/clouddriver/pull/3826/commits/0331e1f9e03100ad0b72662b724e72fdecd27287) changed the `provider` property of `CacheKey` from "kubernetes" to "kubernetes.v2" (Deck expects it to be "kubernetes"). It also removed `provider`'s getter (this was being generated by `@Data` on `CacheKey` before it was removed) and made it static, so we were no longer deserializing it when returning `/search` results from the cache.


maggieneterval:
I'm going to cherry-pick this back to 1.17 and 1.18 (and the corresponding Deck fix, which made it into 1.18, back to 1.17) so that the Infrastructure/Firewalls view is functional in each supported Spinnaker release.

maggieneterval:

maggieneterval:

maggieneterval:

PR #4250, opened 2020-01-10
chore(kubernetes): bump kubernetes java client from 5.0.0 to 7.0.0

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/5291

Changes:

- Generated packages moved from `io.kubernetes.client` to `io.kubernetes.client.openapi `
- `V1PodStatus.isReady()` --> `V1PodStatus.getReady()`
- `V1DeleteOption.isOrphanDependents()` --> `V1DeleteOption.getOrphanDependents()`
- `V1beta1Ingress*` --> `NetworkingV1beta1Ingress*`

PR #4251, opened 2020-01-13
fix(kubernetes): do not throw NPE when no initContainers

maggieneterval:
Fixes potential NPE introduced by https://github.com/spinnaker/clouddriver/pull/4235

- `spec.initContainers` defaults to null, not an empty list. Do not throw NPE when no `initContainers` are configured.
- Once https://github.com/spinnaker/clouddriver/pull/4250 is merged, we will get IDE warnings for these nullable properties :smile: 

cc @sgarlick987


maggieneterval:
cc @duftler we should cherry-pick this to 1.18, as any logs from pods without configured initContainers will not surface in the UI. Will tag you on the cherry-pick PR.

maggieneterval:

PR #4261, opened 2020-01-15
feat(artifacts/gitRepo): Allow version value to be branch or commit SHA

aelath:
Until now the artifact.version value, which contains the git reference,
was required to be a branch name. These changes no longer assume version
is a branch and allow the caller to pass in commit SHA as well.

### DIFF
--- a/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
+++ b/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
@@ -38,7 +38,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.eclipse.jgit.api.ArchiveCommand;
-import org.eclipse.jgit.api.CheckoutCommand;
 import org.eclipse.jgit.api.CloneCommand;
 import org.eclipse.jgit.api.Git;
 import org.eclipse.jgit.api.errors.GitAPIException;
@@ -98,6 +97,8 @@ public class GitRepoArtifactCredentials implements ArtifactCredentials {
   @Override
   public InputStream download(Artifact artifact) throws IOException {
     String repoReference = artifact.getReference();
+    String subPath = artifactSubPath(artifact);
+    String remoteRef = artifactVersion(artifact);
     Path stagingPath =
         Paths.get(System.getProperty("java.io.tmpdir"), UUID.randomUUID().toString());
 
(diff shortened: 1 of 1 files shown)


clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
```diff
@@ -29,10 +29,7 @@
 import java.io.OutputStream;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
```
ethanfrogers: Prefer explicit imports instead of `*`. This is something your IDE usually changes without you noticing it. There's a setting to change it.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
```diff
@@ -64,6 +62,7 @@
   private final String sshKnownHostsFilePath;
   private final boolean sshTrustUnknownHosts;
   private final AuthType authType;
+  private String remoteReference;
```
ethanfrogers: We shouldn't need a class-level property for the remote reference since this will be variable for each time the `download` method is invoked. It looks like you've added it so you have access to it in the archive method. In this case, you might want to resolve the reference inside the `download` method and pass it into the archive method.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
```diff
@@ -122,8 +121,32 @@ public InputStream download(Artifact artifact) throws IOException {
...
+    String remoteRef;
+    Boolean branch = false;
+    if (gitRef.contains("/")) {
+      String[] path = gitRef.split("/");
+      branchOrCommit = path[path.length - 1];
+    } else {
+      branchOrCommit = gitRef;
+    }
+    String possibleRemoteRef = "refs/heads/" + branchOrCommit;
+    Collection<Ref> call = repo.lsRemote().setHeads(true).call();
```
ethanfrogers: Would it be possible to get this information without making another call out to the remote? Is this data available when we do the initial clone?

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
aelath: Noted, let me move that out of the class-level.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
aelath: I agree, it'd be nice to keep from making another api call. I might be able to use `branchList` instead. I'm going to give that a try.

aelath:
@ethanfrogers I updated this PR in response to your feedback. I split the `clone` method into `clone` and `checkout` so that between calls to the two I could resolve the git reference we need to checkout and use later. This eliminates the use of the class-level property. I also figured out a way to use `branchList` instead of `lsRemote`, which *should* eliminate that extra api call.

aelath:
After talking this over with @ethanfrogers we decided to change the approach. Instead of trying to determine the intent of the caller ("branch or commit SHA?")we are leveraging JGit's builtin handling of references in the clone command. This reduces the number of calls to the remote by 1 and maintains the interface. The trade off is that using checkout in the clone command does not allows us to restrict to a sub path of the repository. At the time of writing this seems like a suitable trade as the archive subpath functionality stays intact.

PR #4262, opened 2020-01-16
fix(sql): SqlUnknownAgentCleanupAgent should not throw errors on nonexistent tables

maggieneterval:
Related to https://github.com/spinnaker/spinnaker/issues/4803

Kubernetes caching agents can be authoritative for resource types for which there are zero instantiated resources. For example, I can register a CRD to a cluster that I never instantiate, and a caching agent will become authoritative for that type but a SQL table will never be created for it. We should ensure a table exists before scanning it for entries to clean up.


maggieneterval:
@ezimanyi 10/10 on the emojis, wow.

@robzienert let me know if my jooq or kotlin is too embarrassing to merge. Would love to cherry-pick this into the 1.18 release branch before 1.18.0 is released next week (cc @duftler)!

robzienert:
@maggieneterval Perfect kotlin and jooq, exquisite. You are now the spinnaker database person!

maggieneterval:
@spinnakerbot :cherries: :pick: 1.18

PR #4268, opened 2020-01-21
fix(sql): Ensure SqlEventCleanupAgent is scheduled

robzienert:
A recent refactor to get other SQL agents to wire up caused this agent to be missed.

### DIFF
--- a/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/event/SqlEventCleanupAgent.kt
+++ b/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/event/SqlEventCleanupAgent.kt
@@ -17,6 +17,7 @@ package com.netflix.spinnaker.clouddriver.sql.event
 
 import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.cats.agent.RunnableAgent
+import com.netflix.spinnaker.clouddriver.sql.SqlAgent
 import com.netflix.spinnaker.clouddriver.cache.CustomScheduledAgent
 import com.netflix.spinnaker.clouddriver.core.provider.CoreProvider
 import com.netflix.spinnaker.config.ConnectionPools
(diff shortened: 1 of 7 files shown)


clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/event/SqlEventCleanupAgent.kt
```diff
@@ -40,7 +41,7 @@ class SqlEventCleanupAgent(
   private val jooq: DSLContext,
   private val registry: Registry,
   private val properties: SqlEventCleanupAgentConfigProperties
-) : RunnableAgent, CustomScheduledAgent {
+) : RunnableAgent, CustomScheduledAgent, SqlAgent {
```
robzienert: This is the fix.

clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConfiguration.kt
```diff
@@ -100,6 +100,7 @@ class SqlConfiguration {
   }
 
   @Bean
+  @ConditionalOnExpression("\${sql.read-only:false} == false")
```
robzienert: And this fixes a separate problem with this agent wiring up - shouldn't be run if clouddriver is in readonly mode.

marchello2000: Is this needed, isn't this taken into account here:
https://github.com/spinnaker/clouddriver/blob/master/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlCacheConfiguration.kt#L181 ?

robzienert: The agent would still be created, which would be unnecessary. It won't run, but it'd also just be sitting in memory needlessly.

marchello2000: gotcha, makes sense

PR #4272, opened 2020-01-21
chore(kork): use kork's NamedThreadFactory

marchello2000:
Since kork now has this same NamedThreadFactory, move to use it


plumpy:
FYI, Guava has a [`ThreadFactoryBuilder`](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/util/concurrent/ThreadFactoryBuilder.html) that does this, but has more features. You can create an equivalent `ThreadFactory` by doing:

```
new ThreadFactoryBuilder()
    .setNameFormat(DefaultOrchestrationProcessor.class.getSimpleName()).build()
```

I'd be in favor of switching to that instead personally...

plumpy:
Whoops, looks like I'm 4 minutes too late.

PR #4273, opened 2020-01-21
fix(eureka): make non-404 error codes non fatal on disable ops

dreynaud:
There already are a lot of special cases in this class, but marking the task as failed can have pretty bad consequences like failing entire deploy stages and pipelines, triggering rollbacks, etc. If the stars are really poorly aligned and we do get 10 failures (timeouts, 404s, 503s...) I think the kato task should still be considered a success, allowing orca to proceed to the next steps (most likely waiting for down instances) which have a chance of succeeding.

### DIFF
--- a/clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/deploy/ops/AbstractEurekaSupport.groovy
+++ b/clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/deploy/ops/AbstractEurekaSupport.groovy
@@ -49,9 +49,11 @@ abstract class AbstractEurekaSupport {
                                          Task task,
                                          String phaseName,
                                          DiscoveryStatus discoveryStatus,
-                                         List<String> instanceIds) {
+                                         List<String> instanceIds,
+                                         boolean strict = false) {
     updateDiscoveryStatusForInstances(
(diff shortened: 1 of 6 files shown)


clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/deploy/ops/AbstractEurekaSupport.groovy
```diff
@@ -146,8 +146,12 @@ abstract class AbstractEurekaSupport {
           }
         }
       } catch (RetrofitError retrofitError) {
-        if (retrofitError.response?.status == 404 && discoveryStatus == DiscoveryStatus.Disable) {
-          task.updateStatus phaseName, "Could not find ${instanceId} in application $applicationName in discovery, skipping disable operation."
+        if (discoveryStatus == DiscoveryStatus.Disable) {
```
marchello2000: does this get called when I take a single instance out of discovery?
if so... maybe we should also check that `instanceIds.size > 1`?

dreynaud: updated in 4a7237bf87f858a890a1a1096fbbb4b023bacae9, instead of checking `instanceIds.size` I pass a `strict=true` flag when we disable one or more instances explicitly but `strict=false` when disabling a server group. What do you think?

PR #4274, opened 2020-01-21
feat(ecs): Add TargetHealthCachingAgent

allisaurus:
**PR 1 of 2** to partially address https://github.com/spinnaker/spinnaker/issues/4495

Adds a dedicated caching agent for ALB target health descriptions.

PR #4275, opened 2020-01-21
feat(ecs): Refactor TaskHealthCachingAgent to use TargetHealthCache

allisaurus:
**PR 2 of 2** to partially address https://github.com/spinnaker/spinnaker/issues/4495

* this PR includes changes from https://github.com/spinnaker/clouddriver/pull/4274 and should be reviewed/merged **after** that one is approved.

Purpose of the change was to reduce calls to `elasticloadbalancing:DescribeTargetHealth` from per-task to per-targetGroup. The below examples were run against Spinnaker instances running 1 ECS service with ~10 tasks each. 

NOTE: the reduction seen depends on the makeup of the account - many services running fewer tasks will see less improvement than accounts running fewer services with many tasks each.

## Testing
Logs show call count reduced to about 1/3 of what's currently observed.

EX: Number of `DescribeTargetHealth` calls made by Spinnaker instance running `master`:

upstream_calls

EX: Number of `DescribeTargetHealth` calls made by Spinnaker instance running against this PR:

refactored_calls

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
@@ -221,12 +221,14 @@ public class TaskHealthCachingAgent extends AbstractEcsCachingAgent<TaskHealth>
 
   private TaskHealth makeTaskHealth(
       Task task, String serviceName, TargetHealthDescription healthDescription) {
-    log.debug("Task target health is: {}", healthDescription.getTargetHealth());
-    String targetHealth =
-        healthDescription.getTargetHealth().getState().equals("healthy")
-            ? STATUS_UP
-            : STATUS_UNKNOWN;
-
+    String targetHealth = STATUS_UNKNOWN;
+    if (healthDescription != null) {
+      log.debug("Task target health is: {}", healthDescription.getTargetHealth());
+      targetHealth =
+          healthDescription.getTargetHealth().getState().equals("healthy")
+              ? STATUS_UP
+              : STATUS_UNKNOWN;
+    }
     TaskHealth taskHealth = new TaskHealth();
     taskHealth.setType("loadBalancer");
     taskHealth.setState(targetHealth);
@@ -317,7 +319,7 @@ public class TaskHealthCachingAgent extends AbstractEcsCachingAgent<TaskHealth>
     if (targetHealth == null) {
       log.debug("Cached EcsTargetHealth is empty for targetGroup {}", targetGroupArn);
       evictStaleData(task, loadBalancerService);
-      return overallTaskHealth;
+      return makeTaskHealth(task, serviceName, null);
     }
     TargetHealthDescription targetHealthDescription =
         findHealthDescription(targetHealth.getTargetHealthDescriptions(), targetId, targetPort);
@@ -329,7 +331,7 @@ public class TaskHealthCachingAgent extends AbstractEcsCachingAgent<TaskHealth>
           targetId,
           targetPort);
       evictStaleData(task, loadBalancerService);
-      return overallTaskHealth;
+      return makeTaskHealth(task, serviceName, null);
     }
 
     log.debug("Retrieved health of targetId {} for targetGroup {}", targetId, targetGroupArn);
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
@@ -31,10 +31,6 @@ import com.amazonaws.services.ecs.model.NetworkBinding;
 import com.amazonaws.services.ecs.model.NetworkInterface;
 import com.amazonaws.services.ecs.model.PortMapping;
 import com.amazonaws.services.ecs.model.TaskDefinition;
-import com.amazonaws.services.elasticloadbalancingv2.AmazonElasticLoadBalancing;
-import com.amazonaws.services.elasticloadbalancingv2.model.DescribeTargetHealthRequest;
-import com.amazonaws.services.elasticloadbalancingv2.model.DescribeTargetHealthResult;
-import com.amazonaws.services.elasticloadbalancingv2.model.TargetDescription;
 import com.amazonaws.services.elasticloadbalancingv2.model.TargetHealthDescription;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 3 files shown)


allisaurus:
CC @pkandasamy91 for input on the `TaskHealthCacheAgent` changes

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
```diff
@@ -299,34 +289,52 @@ private TaskHealth inferHealthNetworkBindedContainer(
...
+    TargetHealthDescription targetHealthDescription =
+        findHealthDescription(targetHealth.getTargetHealthDescriptions(), targetId, targetPort);
+
+    if (targetHealthDescription == null) {
+      log.debug(
+          "TargetHealthDescription is empty on targetGroup '{}' for {}:{}",
+          targetGroupArn,
+          targetId,
+          targetPort);
       evictStaleData(task, loadBalancerService);
```
clareliguori: I am not entirely clear on why these extra evictions happen tbh, but should we be evicting the target health key too?  Honestly no idea...

clareliguori: tbh I'm not sure why we do these evictions of the service and task definition here, but should we be evicting the target group health too?  Really not sure what impact that would have

allisaurus: yeah, I'm not sure what role evictions have if the cache is being rebuilt (as in, described with the SDK and fully re-populated) every time the agent executes - but maybe that's not truly what's going on? I'll need to poke around some.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
```diff
@@ -299,34 +289,52 @@ private TaskHealth inferHealthNetworkBindedContainer(
...
-                .withTargetGroupArn(targetGroupArn)
-                .withTargets(new TargetDescription().withId(targetId).withPort(targetPort)));
-
-    if (describeTargetHealthResult.getTargetHealthDescriptions().isEmpty()) {
-      log.debug("Target health description is empty");
+
+    String targetHealthKey = Keys.getTargetHealthKey(accountName, region, targetGroupArn);
+    EcsTargetHealth targetHealth = targetHealthCacheClient.get(targetHealthKey);
+
+    if (targetHealth == null) {
```
pkandasamy91: Is there a possibility that the cache doesn't contain an entry but the DescribeHealthResult would show that the task is unhealthy? (i.e. the cache is a bit stale, hasn't pulled/updated recently). Would it make sense to call describeTargetHealth if the cache doesn't have an entry, and then return if that is also null?

Also, not sure if we really need to evict here if the cache is always getting everything (on pull).

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
pkandasamy91: The eviction would only make sense if you're calling AWS, it would remove the stale entries from the cache. Don't think you need it anymore.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
pkandasamy91: In the scenario mentioned above, if a new task was started as part of a deployment but was not added to the cache, this method would return success if any one of the tasks yet to start up was not in the cache.

allisaurus: it's possible it may not find the given target (due to staleness), and right now the approach is to wait for the downstream caches to catch up and declare the task healthy at that point. This applies to the the other ECS caches called here as well. 

While I like the immediacy of calling ELB as a fallback, I'm worried this might be invoked too frequently to retain most of the benefit of the dedicated TargetHealth cache. My preference would be to see how this goes and incorporate additional/fallback logic if increased wait times for "up instances" become a problem (anecdotally, I deployed a few times and didn't see a noticeable increase). However, my hunch is that rate limiting is the bigger bottleneck here ATM.

allisaurus: > In the scenario mentioned above, if a new task was started as part of a deployment but was not added to the cache, this method would return success if any one of the tasks yet to start up was not in the cache.

You're right that if the task has multiple target groups attached then this would return "healthy" if one was found but other's were empty (b/c it's returning the previous overallHealth, not "unknown"). I'll make that change and test it out, then update this.

pkandasamy91: Awesome! Thanks @allisaurus!

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
allisaurus: per offline discussion: punting on reevaluating evictions for a future PR

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
```diff
@@ -299,34 +291,52 @@ private TaskHealth inferHealthNetworkBindedContainer(
...
-      return overallTaskHealth;
+      return makeTaskHealth(task, serviceName, null);
     }
 
-    log.debug(
-        "Target health description is not empty and has a size of {}",
-        describeTargetHealthResult.getTargetHealthDescriptions().size());
-    TargetHealthDescription healthDescription =
-        describeTargetHealthResult.getTargetHealthDescriptions().get(0);
+    log.debug("Retrieved health of targetId {} for targetGroup {}", targetId, targetGroupArn);
```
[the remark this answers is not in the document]
allisaurus: hmm since we're returning at the end of the previous `if` block, I'm inclined to leave it as is. But +1 for keeping an eye on readability in general.

PR #4284, opened 2020-01-27
refactor(kubernetes): Move a few classes to better packages

ezimanyi:
* refactor(kubernetes): Move a few classes to better packages 

  KubernetesV2SearchProvider should be in the same package as all the other Kubernetes providers; move it there.

  KubernetesAccountResolver is really just a helper class for the classes in .*kubernetes.v2.caching.view.provider; move it there so it can be made package-private and only visible to those classes.

  GlobalKubernetesKindRegistry is closely associated with the account credentials and should live in the same package (as the account-specific kind registry does).

* refactor(kubernetes): Remove CacheUtils from handlers 

  The hydrateSearchResult function in all the kubernetes handlers accepts a KubernetesCacheUtils, but never uses it. Remove the unneeded parameter from these functions.

* refactor(kubernetes): Make classes package-private 

  The KubernetesAccountResolver and KubernetesCacheUtils classes are now only used within their package (after the prior commits); make them package-private.

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesCacheUtils.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesCacheUtils.java
@@ -25,7 +25,6 @@ import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.Keys;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.agent.KubernetesCacheDataConverter;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.view.model.ManifestBasedModel;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.view.provider.data.KubernetesV2CacheData;
-import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.KubernetesAccountResolver;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.KubernetesResourceProperties;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.KubernetesSpinnakerKindMap;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.KubernetesManifest;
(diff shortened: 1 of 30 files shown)


clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesCacheUtils.java
```diff
@@ -46,7 +45,7 @@
 
 @Component
 @Slf4j
-public class KubernetesCacheUtils {
+class KubernetesCacheUtils {
```
ethanfrogers: Given all the work I've been doing with CRD extensions lately I feel like we might actually want want to leave this public so it can be used in any custom caching agents. Thoughts?

ezimanyi: In general when people write custom caching agents, what package do they put them in?  I'd also be interested in how people are using this in custom caching agents if none of the built-in ones use it.

ezimanyi: (The reason I ask is that classes like this that become implicit public contracts are effectively impossible to refactor if we don't have any boundaries on how people are using them in custom builds.)

ethanfrogers: I actually haven't seen anyone using a custom caching agent. I'd assume they put it in a package alongside their custom handler. For example, we use `io.armory.spinnaker` for a lot of our internal extensions. If it were me, I'd probably put it under something like `io.armory.spinnaker.kubernetes.{handler, cache}`. We should probably ask some of the users we know have custom handlers to be sure. 

Totally understand the need to make extension points more explicit. Just want to be sure we don't break anyone in the process w/o warning.

ezimanyi: I pushed a commit that un-does making `KubernetesCacheUtils` package-private to not hold this up while seeing if that will break anyone.  I'll reach out to a couple of users that I know have build custom handlers to see if they use this.

PR #4288, opened 2020-01-28
Fix typo in Kubernetes Deployment load message

KeisukeYamashita:
## What

* Fix typo `deployemnts` to `deployments`.


clouddriver-kubernetes-v1/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/agent/KubernetesDeploymentCachingAgent.groovy
```diff
@@ -55,7 +55,7 @@ class KubernetesDeploymentCachingAgent extends KubernetesV1CachingAgent {
 
   @Override
   CacheResult loadData(ProviderCache providerCache) {
-    log.info("Loading deployemnts in $agentType")
+    log.info("Loading deployements in $agentType")
```
louisjimenez: Looks like there is an extra e between the y and m.

PR #4294, opened 2020-01-29
refactor(titus): Move DeleteTitusScalingPolicy to Sagas

robzienert:
I'd like to get some momentum back into making Titus AtomicOperations use Sagas. I chose a simple operation so I can use it as an example of what's needed to do this work.

- Converts operation to Java
- Created `AbstractSagaAtomicOperation` to cut down on boilerplate for non-`DeployHandler` operations
- Added a new `TitusScalingPolicyDeleted` event that is emitted at the end of the the operation


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/sagas/AbstractSagaAtomicOperation.java
```diff
@@ -0,0 +1,94 @@
...
+import javax.annotation.Nonnull;
+import org.springframework.beans.factory.annotation.Autowired;
+
+/**
+ * Removes some of the boilerplate for AtomicOperations to use Sagas.
+ *
+ * @param <T> The AtomicOperation description
+ * @param <R> The operation result type
+ */
+public abstract class AbstractSagaAtomicOperation<T, R>
```
jonsie: My first thought when reading this is - why can't we use this for `DeployHandler` implementations too?  It's a similar process - we create a saga flow and then return a result via `SagaAtomicOperationBridge`.  I think if we make this work with for both cases, that would be better.

PR #4295, opened 2020-01-29
fix(eureka): kato result object for skipped instances

asher:
This is a follow-up to (#4273) where certain failures to deregister an instance with eureka are skipped as potentially safe to handle upstream. This stores skipped instances to a kato result object for use by orca.


clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/deploy/ops/AbstractEurekaSupport.groovy
```diff
@@ -189,6 +192,11 @@ abstract class AbstractEurekaSupport {
         AbstractEurekaSupport.log.info("[$phaseName] - Failed marking discovery $discoveryStatus.value for instances ${errors}")
       }
     }
+    if (skipped) {
+      task.addResultObjects([
+        ["discoverySkippedInstanceIds": instanceIds]
```
asher: Orca's `WaitForRequiredInstancesDownTask` operates over an `instanceIdsToDisable` list provided via kato results. We've observed task timeouts caused by instances that were skipped by `updateDiscoveryStatusForInstances` due to a 404 but that never pass the `someAreDownAndNoneAreUp` test. When this is the case as part of a RRB deploy stage where < 100% of the prior asg is being disabled, we can potentially punt such instances to a latter rolling group, or explicitly retry from orca.

marchello2000: 👍

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/deploy/ops/AbstractEurekaSupport.groovy
```diff
@@ -189,6 +192,11 @@ abstract class AbstractEurekaSupport {
         AbstractEurekaSupport.log.info("[$phaseName] - Failed marking discovery $discoveryStatus.value for instances ${errors}")
       }
     }
+    if (skipped) {
```
marchello2000: nit: i would prefer something like 
```suggestion
    if (!skipped.isEmpty()) {
```
to make it less groovy and slightly more obvious what's going on

PR #4296, opened 2020-01-29
refactor(kubernetes): A few minor improvements to the kind registry

ezimanyi:
* test(kubernetes): Convert KubernetesKindRegistry tests to java 

  Also replace mocking with using real objects where possible.

* refactor(kubernetes): Move tests to same package as classes 

  I recently moved a couple of classes but forgot to also move the tests to the same package; move the tests now.

* refactor(kubernetes): Resolve kind properties higher in call stack 

  Rather than pass the kind registry down into convertAsResource (and the functions it calls), just get the properties for the kind we're interested in and pass them directly to convertAsResource.

  This reduces the number of places we have a KubernetesKindRegistry passed around.

* refactor(kubernetes): Support kind lookup directly on credentials 

  I'd originally allowed callers to get the KindRegistry associated with a KubernetesV2Credentials object as it's a smaller thing to pass around in cases where we don't need the credentials and I didn't want to add more functionality to an already large class.

  But the reality is that every time we look something up on a registry, we do so via credentials.getKindRegistry().getSomething() so this really is something that clients are querying on the credentials object.

  By having clients just call getGlobalKinds and getKindProperties directly on the credentials object, we make the existence of the kind registries an implementation detail which makes them easier to change as we account for more conditions and edge cases in computing the status of a kind. (An upcoming commit will require us to consider both the status in the registry and whether a kind is readable when computing its status which is now much easier to do without affecting callers.)

* refactor(kubernetes): Make registries package-private 

  Now that the kind registries are implementation details and all callers just directly query the credentials for information on kinds of interest, make these package-private so they aren't used in other places.

  This also changes GlobalKubernetesKindRegistry to be a component instead of being generated as a bean (which would have required it to be visible to the configuration class).

  Also add default nullity annotations to these registries.

* refactor(kubernetes): Avoid getting kind status twice 

  When validating a manifest we effectively get the status twice any time it's not valid. Replace this with a single check to get the status, and check if it's valid.

* refactor(kubernetes): Simplify some credentials functions 

  The only callers of getGlobalKinds() and getCrds() take the result and:
  * Discard the properties to get only the actual kind
  * Filter out invalid kinds

  Build this into the credentials, so that these functions just directly return kinds, and only return valid ones, and remove the filtering logic from the callers.

  This also allows us to make isValidKind private as we're doing filtering within the class.

  Also update getRegisteredKinds in the kind registries to more directly return what the only caller is now asking for.

* refactor(kubernetes): Add isRegistered function to kind registry 

  There's currently no way for a consumer of the KubernetesKindRegistry to know if a kind is registered as getKindProperties will return default properties if it isn't.

  Address this in two ways:
  (1) Rename getKindProperties to getKindPropertiesOrDefault to make it more clear what this function does.
  (2) Add an isKindRegistered that will query if a kind is registered (and try to register it by looking up CRD properties) and returns true if the kind is registered at the end of the function (either initially or by CRD lookup).


ezimanyi:
This is, I promise, the last refactor before I actually fix the bug I've been working in (spinnaker/spinnaker#4545) which will be a very small fix after these refactors.

duftler:
I'm really looking forward to seeing how this changes things for that bug.

PR #4297, opened 2020-01-29
fix(kubernetes): Validate dynamic accounts before adding to repo, don't replace existing

ncknt:
- Moved initial account loading to `setup`
- On refresh, don't replace accounts that have not changed
- On refresh, attempt to get the namespaces before adding the account to the credential repository. Without this, a bad account blocks `/credentials?expand=true` calls. This is not a problem at startup where bad accounts may block and prevent Clouddriver from being declared healthy.
- Changed CloudConfigRefreshScheduler to be at fixed delay so 2 refreshes don't follow each other as they can take a long time


clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderSynchronizable.java
```diff
@@ -117,17 +118,38 @@ public void synchronize() {
...
 
     return newAndChangedAccounts;
   }
 
+  /**
+   * Validate and save credentials to repository
+   *
+   * @param account
+   */
+  protected void saveToCredentialsRepository(KubernetesNamedAccountCredentials account) {
```
ezimanyi: I think this can be private?

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderSynchronizable.java
```diff
@@ -117,17 +118,38 @@ public void synchronize() {
...
+    // Attempt to get namespaces to resolve any connectivity error without blocking /credentials
+    List<String> namespaces = account.getCredentials().getDeclaredNamespaces();
```
ezimanyi: One worry here is what happens if the call to get namespaces takes a long time---in that case, I think we'll block the account refresh process while waiting for that network connection.

Is this better than just adding the account, and letting that account's caching thread hang while trying to get namespaces (which wouldn't block other threads)?

ncknt: All we're blocking here is the config refresh not the initial startup of Clouddriver (now in `setup()`). The idea is to do a light validation to make sure adding a bad (or temporarily unavailable) account does not bring Clouddriver down. If the account becomes available later, namespaces will be populated.

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderSynchronizable.java
ncknt: I was hoping to keep it `protected` to have options that performs different levels of validations but we can revisit if needed. Updated to `private`

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderSynchronizable.java
ezimanyi: Ok, I think that makes sense---I *thought* that an unreachable account in clouddriver should only affect the caching thread for that account, which might be better than blocking the whole config refresh thread.  But as this is only testing new accounts that are added in that refresh cycle, the timeout should be fairly quick (and the only effect is that new accounts from that cycle would not be added while we're waiting on an unreachable account).

ncknt: Unfortunately, since the account is added to the account repository, it is used for all operations. New and modified accounts would be affected and potentially blocked but the blocking should only occur once thanks to the memoization you added.

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderSynchronizable.java
```diff
@@ -26,12 +25,7 @@
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.agent.KubernetesV2CachingAgentDispatcher;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.security.KubernetesV2Credentials;
 import com.netflix.spinnaker.clouddriver.security.*;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
+import java.util.*;
```
ethanfrogers: Small nit, prefer explicit imports instead of `*`. Freaking IntelliJ.

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderSynchronizable.java
ezimanyi: Ok, makes sense, thanks for explaining this!

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/KubernetesV2ProviderSynchronizable.java
devorbitus: I know this is an old comment but I happened to see it and wanted to let future people know where to adjust this https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcards-for-class

PR #4299, opened 2020-01-29
fix(ALB): Don't try to fetch LB when ALB is requested

marchello2000:
We see an issue where whenever you create an ALB we will actually fail in edda because we try to get the updated details of the ALB
from both the ALB and the non-ALB list.
Don't try to fetch an ALB from regular LBs (this is symmetric to what ALB caching agent does)

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonLoadBalancerCachingAgent.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonLoadBalancerCachingAgent.groovy
@@ -63,7 +63,7 @@ class AmazonLoadBalancerCachingAgent extends AbstractAmazonLoadBalancerCachingAg
     if (!data.containsKey("loadBalancerName")) {
       return null
     }
-    if (data.containsKey("loadBalancerType") && data.loadBalancerType.equals("application")) {
+    if ("application".equals(data.loadBalancerType)) {
       return null
     }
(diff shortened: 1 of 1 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonLoadBalancerCachingAgent.groovy
```diff
@@ -63,6 +63,9 @@ class AmazonLoadBalancerCachingAgent extends AbstractAmazonLoadBalancerCachingAg
     if (!data.containsKey("loadBalancerName")) {
       return null
     }
+    if (data.containsKey("loadBalancerType") && data.loadBalancerType.equals("application")) {
```
dreynaud: nit:
```suggestion
    if ("application".equals(data.loadBalancerType)) {
```

marchello2000: this is symmetric to:
https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonApplicationLoadBalancerCachingAgent.groovy#L98

marchello2000: nice, thanks

PR #4301, opened 2020-01-31
fix(kubernetes): On refresh, don't replace credentials that have not changed

ncknt:
Similar to https://github.com/spinnaker/clouddriver/pull/4297 but limited to not losing the memoized values. It will only address dynamic accounts becoming unreachable.


ncknt:
@duftler Can you PTAL? Fix for 1.18.

PR #4302, opened 2020-02-03
test(kubernetes): Add tests to status function of handlers

ezimanyi:
* fix(kubernetes): Allow integer values as generation 

  The getObservedGeneration function assumes that the generation on the manifest is a Double. This is an artifact of the particular JSON parser we're using to create a KubernetesManifest, which uses a Double for any integer values.

* test(kubernetes): Add tests to status function of handlers 

  There are very few tests on the implementations of the status function in KubernetesHandler. Before doing some refactoring of the status computation logic, add a robust suite of tests to document the current behavior.

  Some notes:
  * There were already tests on daemon set and stateful set; here I've augmented the tests to assert more thing on the returned status, and in the case of stateful set have converted the groovy test to Java.
  * Some of these might be a bit overspecified (ex: checking the specific string in the unstable reason) but I'd rather know if something changes during the refactor (and may make some slight deliberate changes).
  * I tried as much as possible to put realistic YAML in the resources
  (ex: by doing a rollout and watching the status in my cluster). In some cases this led me to suspect that we are testing conditions that could never be true; I added comments in places where this is the case.


clouddriver-kubernetes-v2/src/test/resources/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/pod/null-phase.yml
```diff
@@ -0,0 +1 @@
+status: {}
```
maggieneterval: Ohhhhh. :laughing:

PR #4303, opened 2020-02-04
refactor(kubernetes): Simplify status handlers

ezimanyi:
* test(kubernetes): Add test for progress deadline exceeded 

  The test for progress deadline exceeded only considers the case where the deployment reports it is available; also add a case where the deadline is exceeded and the deployment reports it is not available.

* refactor(kubernetes): Improve Status and Condition classes 

  There are way too many ways to create and mutate these classes, as both have fully public fields as well as setters; in the case of Status there are also custom setters, and in the case of Condition there is additionally a builder.

  Clean this up as follows:
  * Update the custom setters (ex: failed, unstable, etc.) in Status to create a new condition and assign it to the appropriate value instead of mutating the existing condition. (This fixes a null-safety issue as some of these fields can be null, if unknown() has been called.)
  * This now allows us to make Condition immutable, as we never change an existing condition. Make it immutable by removing the setters and marking the fields final. Replace builders with a constructor (which is reasonable given that we have only two fields with different types). As it's common to create a condition with a null message (because it shows up in defaults for Status) make a convenience factory method withState that returns a Condition with null message.  (Because the class is immutable we can just return a pre-created static Condition in this case.)
  * Add a defaultStatus() factory method to Status; it's the same as the constructor new Status() but I think makes the intent much more clear when reading the Handlers; that is, start with some default status and mutate it if the desired status differs from the default.  I didn't make the no-arg constructor private because I didn't want to needlessly break people who used it in their own custom handlers, but ideally they'd migrate to defaultStatus() for clarity.
  * Add a stable() mutator to replace the one case where we explicitly called setStable()
  * Add nullity annotations.  Many things are nullable; the new annotations just document that existing behavior.

* refactor(kubernetes): Simplify status reporting in handlers 

  The logic to report workload status is fairly complex to follow; one of the reasons it's complex is that in general we create an empty Status object (which defaults some fields) at the beginning of the status function and mutate it throughout the functions, sometimes returning immediately after the mutation but sometimes not.

  In order to understand what we're returning at any given point, we need to read the entire function up to that point to see if any of the prior conditions wrote to the result variable.

  Where possible, simplify this by just creating a new Status inline at the point of return (and change from using new Status() to the factory method Status.defaultStatus()). This is possible in any cases where there are no mutations of result earlier in the function that don't immediately return. By doing this analysis now once, we make it easier for future readers of the function to see what is returned in each case.

  This commit only touches the handlers where it was trivial to make this change (ie, we never mutate the result without immediately returning it).

* refactor(kubernetes): Simplify Deployment and ReplicaSet handlers 

  Now that there is a robust test suite on the Deployment and ReplicaSet handlers, refactor these to make the control flow more clear when determining status. This commit only makes changes that do not affect results (as seen by the lack of changes to the tests in this commit).

* fix(kubernetes): Fix ready condition check for replicaset 

  We currently check the available replicas count before we check the ready replica count; this isn't correct as there will always be more ready replicas than available replicas. (The definition of an available replica is one that has been ready for at least minReadySeconds.) Switch the order of the check so that we generate the correct message in each case.

* fix(kubernetes): Simplify status messages 

  When a replicaset is unstable and unavailable we use slightly different wording in the unstable and unavailable message even though the cause is the same. This isn't complicates both the code and the user's parsing of messages (as they need to parse both to realize they are the same error).

  Simplify this; the tests show exactly where the text of errors has changed and shows that the actual stable/unstable status does not change.

  Also, push the Deployment's UnstableReason up to the top level so we can re-use it for replica sets.

* fix(kubernetes): Don't return early after setting failed flag 

  When a deployment reports that it is failed we exit early instead of checking the remaining condition. This means that we don't always properly set the other flags of the status in this case; update the logic to set the other flags.

  The logic in orca to decide how to handle the status of a manifest checks both the stable and the failed flags (so we won't risk breaking it by sometimes setting a deployment to unstable when it is also failed, which we already do in some cases anyway if the condition reports unavailable).

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/CustomKubernetesHandlerFactory.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/CustomKubernetesHandlerFactory.java
@@ -33,6 +33,7 @@ import com.netflix.spinnaker.clouddriver.kubernetes.v2.caching.view.provider.dat
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.KubernetesKind;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.description.manifest.KubernetesManifest;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.model.Manifest;
+import com.netflix.spinnaker.clouddriver.kubernetes.v2.model.Manifest.Status;
 import com.netflix.spinnaker.clouddriver.kubernetes.v2.security.KubernetesV2Credentials;
 import javax.annotation.Nonnull;
 import lombok.extern.slf4j.Slf4j;
@@ -88,7 +89,7 @@ public class CustomKubernetesHandlerFactory {
 
     @Override
     public Manifest.Status status(KubernetesManifest manifest) {
-      return new Manifest.Status();
+      return Status.defaultStatus();
     }
 
     @Override
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/model/Manifest.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/model/Manifest.java
@@ -17,12 +17,16 @@
 
 package com.netflix.spinnaker.clouddriver.kubernetes.v2.model;
 
+import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
 import com.netflix.spinnaker.moniker.Moniker;
 import java.util.List;
-import lombok.AllArgsConstructor;
+import javax.annotation.Nullable;
 import lombok.Builder;
 import lombok.Data;
-import lombok.NoArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
+import lombok.Value;
 
 public interface Manifest {
   Moniker getMoniker();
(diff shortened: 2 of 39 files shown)


clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/CustomKubernetesHandlerFactory.java
```diff
@@ -88,7 +89,7 @@ public SpinnakerKind spinnakerKind() {
 
     @Override
     public Manifest.Status status(KubernetesManifest manifest) {
-      return new Manifest.Status();
+      return Status.defaultStatus();
```
ethanfrogers: How would this change effect users who are extending Clouddriver w/ custom CRD handlers? Will these changes break those implementations still using `Manifest.Status`? If so, we should figure out how to alert users to these new changes.

ezimanyi:
Good question about how this will affect users who have extended clouddriver.  I tried as much as possible to make any breakage minimal (and to make it obvious, leading to compilation errors, rather than sutble, leading to runtime bugs).

The things that could break custom implementations are:
* `Manifest.Status.Condition` now has final fields and can only be constructed with the all-args constructor `Condition(boolean, String)`.  So anyone who was using the removed builder or the removed no-arg constructor and/or was setting fields using the removed `setState()` or `setMessage()` fields will get a compliation error; they will need to instead just just the all-arg constructor.  I'm not overly worried about this because:
  * In general handlers are using the `Manifest.Status` methods to update the status and aren't creating conditions themselves; of all the built-in handlers there was only a [single case](https://github.com/spinnaker/clouddriver/blob/9f2c141fc47c7bd53b7bdfdc181b8482c9234f8a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesStatefulSetHandler.java#L157) where we created a condition manually; this is easily replaced by just calling the appropriate method in `Manifest.Status`
* The fields of `Manifest.Status` are now private and don't have setters.  Anyone who was accessing these fields directly will get a compilation error (and will have to move to using the getter); anyone who was using the setter will get a compilation error and will need to use one of the custom methods on the class to set what they need (`.stable()`, `.unstable()`, etc.).  I'm not overly worried about this because:
  * Nowhere in the built-in handlers were we accessing the fields directly, and in only [one place](https://github.com/spinnaker/clouddriver/blob/9f2c141fc47c7bd53b7bdfdc181b8482c9234f8a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/KubernetesStatefulSetHandler.java#L157) (the same as above) were we directly calling a settter.

I think that most people who wrote custom handlers probably used the existing ones as examples, which is why I'm using the very minimal changes needed to those handlers as a signal that custom handlers should similarly need minimal changes.  (It's worth looking at [this commit](https://github.com/spinnaker/clouddriver/pull/4303/commits/31a75e89bb8d240cb9a22cbbbed1859bd46df1cc) separately as that's the only one that could break custom implmentations, and shows the single change that was needed across all of the built-in handlers.  The later commits are all handler-specific cleanups that take advantage of new functionality but people don't need to do to their handlers if they don't want.)

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/handler/CustomKubernetesHandlerFactory.java
ezimanyi: The change on this particular line won't affect users at all, as it changes only the implementation of this `status()` function but leaves its behavior identical---`new Manifest.Status()` and `Status.defaultStatus()` are identical.

But the general question is a good one; I've replied in general as non-line-specific PR comment.

ezimanyi:
Added one small commit, as writing the comment above made me realize there's no way someone who creates a `Condition` could actually use it anymore, so making the constructor private will make the API simpler.

PR #4305, opened 2020-02-05
fix(cf): Parallelize apps/lbs cache building

pdelagrave:
Creating a cache entry for one CF app requires 5 distinct HTTP calls to the CF API which adds up quickly when there's thousands of apps to process. The change takes advantage of Java's parallelStream and OKHttp's thread safety/connection pool management.
A better solution will be to use the CF-java client which is fully reactive and will allow to better configure the parameters of what's being done in parallel; but that's a non trivial refactor compared to the current change.


pdelagrave:
Doing some more benchmarks and will possibly add a configurable parallel connection count configurable parameter before merging.

PR #4313, opened 2020-02-07
fix(kubernetes): Move generation checks to be object-specific

ezimanyi:
* test(kubernetes): Fix observed generation in stateful set test 

  One of the stateful set tests incorrectly had the observed generation newer than the metadata one; fix this.

* fix(kubernetes): Replace null status with false 

  When the observed generation in the status does not match the object's generation, we set the stable and failed conditions to null. This should not be treated any differently than if the status generation did match but the object had not fully rolled out---it just means we need to wait a bit longer to see the new status.

  The recent changes to orca have explicitly made it so that from orca's perspective there is no different beteween stable/failed being null and stable/failed being false. Make things more clear (and null-safe) by removing the ability to null these conditions.

* fix(kubernetes): Move generation checks to be object-specific 

  We have a generic function in KubernetesManifest to determine whether the status generation is correct. It has a lot of type-safety and null-safety issues as it's operating on an untyped map.

  As each status handler already converts the manifest to a strongly-typed representation of its handled object, let's move the status checks to happen on that object.

  This in particular fixes some null-pointer issues that were documented in the tests.

* fix(kubernetes): Fix pod phase status reporting 

  This was the last remaining TODO I had noted when adding tests to all of the handlers.

* refactor(kubernetes): Pull unknown status to factory method 

  We frequently create an unknown status with the same text; pull it up to a factory method so we don't keep repeating the same text.

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/model/Manifest.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/model/Manifest.java
@@ -46,19 +46,13 @@ public interface Manifest {
   @NonnullByDefault
   @ToString
   class Status {
-    private @Nullable Condition stable = Condition.withState(true);
+    private Condition stable = Condition.withState(true);
     private Condition paused = Condition.withState(false);
     private Condition available = Condition.withState(true);
-    private @Nullable Condition failed = Condition.withState(false);
+    private Condition failed = Condition.withState(false);
 
     public static Status defaultStatus() {
       return new Status();
-    }
-
-    public Status unknown() {
-      stable = null;
-      failed = null;
-      return this;
     }
 
     public Status failed(@Nullable String message) {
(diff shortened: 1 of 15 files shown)


clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/model/Manifest.java
```diff
@@ -46,19 +46,19 @@
...
     private Condition paused = Condition.withState(false);
     private Condition available = Condition.withState(true);
-    private @Nullable Condition failed = Condition.withState(false);
+    private Condition failed = Condition.withState(false);
 
     public static Status defaultStatus() {
       return new Status();
     }
 
-    public Status unknown() {
```
ezimanyi: The removal of `unknown` is the only change in this PR that could break custom handlers (as the other changes are all implementation-specific changes to built-in handlers).  I think the benefits of null-safety and making it easier to understand outweigh this; anyone who had called `unknown()` to null out fields on the status will need to pick a new appropriate mutation; likely `Status.unstable()` with an appropriate message about why it's unstable.  I opted to remove the method rather than just make it do that automatically because I think it's worth clients figuring out why they were calling `unknown()` and instead doing something more appropriate rather than changing the meaning of `unknown()` but leaving their code compiling.

PR #4324, opened 2020-02-14
feat(provider/aws): Adding delete launch config operation

jeyrschabu:
- added launch config delete operation


jeyrschabu:
I don't see a lot of value for unit test in this case. No real logic to assert other than the to aws.

PR #4325, opened 2020-02-14
feat(core): add SpinnakerRetrofitErrorHandler

maggieneterval:
Users are frequently frustrated by generic `RetrofitErrors` surfaced in pipeline execution details in the UI. A `RetrofitError`'s message is not surfaced when the error was thrown in a microservice more than one network call from the Orca task in which it originated. For example:

- An Orca task that calls out to Rosco, in which Rosco throws an exception, successfully surfaces the exception's message because Orca's `RetrofitExceptionHandler` [parses the RetrofitException.Response's `body.message`](https://github.com/spinnaker/orca/blob/master/orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/RetrofitExceptionHandler.groovy#L50).
- However, if the same Orca task calls out to Rosco, which calls out to Clouddriver, and Clouddriver throws an exception, we can no longer parse the message since we now have a `RetrofitError` body within the body of another `RetrofitError` (at `body.body.message`).

This PR is a proof-of-concept solution to this problem. By registering an error handler with each `Retrofit.RestAdapter` that re-throws `RetrofitErrors` as some subclass of `SpinnakerException` with a useful detail message, Orca is able to surface the useful detail message to end users. As an added benefit, we can centralize logic that is duplicated in many `catch (RetrofitError e)` blocks:

- Checking if a `RetrofitError` is an HTTP error in order to safely read its status.
- Checking if a `RetrofitError` is an HTTP error with a status of 404.

If this seems reasonable, I'm happy to move forward by:

1. Moving `SpinnakerRetrofitErrorHandler`, `SpinnakerNetworkException`, and `SpinnakerHttpException` to Kork.
2. Registering `SpinnakerRetrofitErrorHandler` with each `Retrofit.RestAdapter`.
3. Replacing each `RetrofitError` catch-block with a `SpinnakerNetworkException` catch-block, preceded by more specific `SpinnakerHttpException` and/or `NotFoundException` catch-blocks where appropriate.
(I would complete 2 and 3 for Clouddriver before moving on to the other microservices)

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/LocalFileUserDataProvider.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/LocalFileUserDataProvider.groovy
@@ -20,6 +20,7 @@ import com.netflix.spinnaker.clouddriver.aws.deploy.LaunchConfigurationBuilder
 import com.netflix.spinnaker.clouddriver.core.services.Front50Service
 import com.netflix.spinnaker.clouddriver.exceptions.SpinnakerHttpException
 import com.netflix.spinnaker.clouddriver.exceptions.SpinnakerNetworkException
+import com.netflix.spinnaker.clouddriver.exceptions.SpinnakerServerException
 import com.netflix.spinnaker.kork.web.exceptions.NotFoundException
 import org.springframework.beans.factory.annotation.Autowired
 import retrofit.RetrofitError
@@ -43,7 +44,7 @@ class LocalFileUserDataProvider implements UserDataProvider {
         return Boolean.valueOf(application.legacyUdf)
       } catch (NotFoundException e) {
         return localFileUserDataProperties.defaultLegacyUdf
-      } catch (SpinnakerNetworkException e) {
+      } catch (SpinnakerServerException e) {
         throw e
       }
     }
@@ -60,10 +61,8 @@ class LocalFileUserDataProvider implements UserDataProvider {
           Thread.sleep(retryBackoff)
         }
       } catch (SpinnakerNetworkException e) {
-        if (e.getKind() == RetrofitError.Kind.NETWORK) {
-          Thread.sleep(retryBackoff)
-        }
-      }
+        Thread.sleep(retryBackoff)
+      } catch (SpinnakerServerException e) {}
     }
     throw new IllegalStateException("Failed to read legacyUdf preference from front50 for $account/$applicationName")
   }
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/LocalFileUserDataProvider.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/LocalFileUserDataProvider.groovy
@@ -48,6 +48,7 @@ class LocalFileUserDataProvider implements UserDataProvider {
       }
     }
 
+    // TODO(rz) standardize retry logic
     final int maxRetry = 5
     final int retryBackoff = 500
     final Set<Integer> retryStatus = [429, 500]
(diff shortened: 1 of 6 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/LocalFileUserDataProvider.groovy
```diff
@@ -52,8 +54,12 @@ class LocalFileUserDataProvider implements UserDataProvider {
     for (int i = 0; i < maxRetry; i++) {
       try {
         return result.call()
-      } catch (RetrofitError re) {
-        if (re.kind == RetrofitError.Kind.NETWORK || (re.kind == RetrofitError.Kind.HTTP && retryStatus.contains(re.response.status))) {
+      } catch (SpinnakerHttpException e) {
+        if (retryStatus.contains(e.getResponse().getStatus())) {
+          Thread.sleep(retryBackoff)
```
robzienert: Oh my.

Can you do me a good one and add a `TODO(rz)` comment here to replace this with retry logic with something standard? 🤗

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/exceptions/SpinnakerNetworkException.java
```diff
@@ -0,0 +1,65 @@
...
+import retrofit.RetrofitError;
+
+// todo(mneterval): move to kork-exceptions
+
+/** An exception that exposes the message and kind of a {@link RetrofitError}. */
+@Getter
+@NonnullByDefault
+public class SpinnakerNetworkException extends SpinnakerException {
+  private final String rawMessage;
+  private final RetrofitError.Kind kind;
```
marchello2000: is it worth abstracting the `RetrofitError.Kind` out of here? Not that we would ever move away from retrofit (one hopes..?) but decoupling it here might be useful esp if a bunch of error handlers are going to be checking the `kind`

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/RetrofitConfig.groovy
```diff
@@ -55,6 +56,7 @@ class RetrofitConfig {
       .setConverter(new JacksonConverter())
       .setLogLevel(retrofitLogLevel)
       .setLog(new Slf4jRetrofitLogger(Front50Service))
+      .setErrorHandler(new SpinnakerRetrofitErrorHandler())
```
ezimanyi: I don't know if we'll ever be creating more than one of these in a single service, but this might be a good candidate for a static factory, as this is stateless so can be shared.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/LocalFileUserDataProvider.groovy
maggieneterval: Yes :hugs:

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/exceptions/SpinnakerNetworkException.java
```diff
@@ -0,0 +1,65 @@
...
+import lombok.Getter;
+import retrofit.RetrofitError;
+
+// todo(mneterval): move to kork-exceptions
+
+/** An exception that exposes the message and kind of a {@link RetrofitError}. */
+@Getter
+@NonnullByDefault
+public class SpinnakerNetworkException extends SpinnakerException {
+  private final String rawMessage;
```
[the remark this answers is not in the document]
maggieneterval: Great call, replaced the lombok getter for `rawMessage` with a final package-private one (can expose it further as protected if/when necessary?). Ended up removing `getKind` in favor of a NETWORK-kind specific exception class.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/RetrofitConfig.groovy
maggieneterval: :chart_with_upwards_trend: :coffee: :pouch: :scroll: :heavy_check_mark:

maggieneterval:
Thanks for the quick review, friends.

> I'm all for promoting this into kork, however we'll need to roll it out to each service one-by-one since there's so many places that look for RetrofitError specifically throughout our services (especially in Orca).

I created [this issue](https://github.com/spinnaker/spinnaker/issues/5473) to track rolling it out to each service individually. I'll start by handling each `RestAdapter` and caught `RetrofitError` in Clouddriver first (in a separate PR), then move the new exception classes to `kork-exceptions` and consume it in each microservice one by one.

ezimanyi:
Updates look great! :tada: :100: :1st_place_medal:

PR #4336, opened 2020-02-20
chore(reservations): Disable reservation reports by default

robzienert:
No one outside of Netflix uses reservation reports, and it apparently causes a lot of errors / noise in OSS deployments.


robzienert:
Note for Netflix: I've already made a commit into our config to set this property to true.

ethanfrogers:
@robzienert is this okay to merge yet?

maggieneterval:
@robzienert @jeyrschabu @ethanfrogers The OSS builds are failing on master because Clouddriver won't start up. I'm seeing the following error in the logs, which led me back to this PR:


> BeanCreationException: Error creating bean with name 'awsProvider' defined in class path resource [com/netflix/spinnaker/clouddriver/aws/provider/config/AwsProviderConfig.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.netflix.spinnaker.clouddriver.aws.provider.AwsProvider]: Factory method 'awsProvider' threw exception; nested exception is groovy.lang.GroovyRuntimeException: Could not find matching constructor for: com.netflix.spinnaker.clouddriver.aws.provider.agent.ReservationReportCachingAgent(com.netflix.spectator.api.DefaultRegistry, com.netflix.spinnaker.clouddriver.aws.security.AmazonClientProvider, com.netflix.spinnaker.clouddriver.aws.provider.view.AmazonS3DataProvider, LinkedHashSet, com.fasterxml.jackson.databind.ObjectMapper, Optional, org.springframework.boot.web.servlet.context.AnnotationConfigServletWebServerApplicationContext)


Any idea if we need to change some config in the /buildtool tests to resolve this? Thanks!

PR #4340, opened 2020-02-21
fix(ecs): provide port mappings if legacy target group setting is set

clareliguori:
A bug in the unit tests caused the check for the legacy target group setting to be flipped

When the server group has the legacy settings:
```
      "containerPort": 80,
      "targetGroup": "spinnaker-ecs-ec2-demo-tg",
```
Then the deployment will currently fail with the error message:
```
The container v002 did not have a container port 80 defined. (Service: AmazonECS; Status Code: 400; Error Code: InvalidParameterException; Request ID: 93dc346b-02e8-4ded-a651-b452e44736c6)
```

cc @pkandasamy91


clareliguori:

PR #4341, opened 2020-02-21
feat(ecs): Check for draining services for new server group names

clareliguori:
Fixes https://github.com/spinnaker/spinnaker/issues/5476

### DIFF
--- a/clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/deploy/EcsServerGroupNameResolverSpec.groovy
+++ b/clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/deploy/EcsServerGroupNameResolverSpec.groovy
@@ -53,7 +53,7 @@ class EcsServerGroupNameResolverSpec extends Specification {
       request.cluster == ecsClusterName
       request.services == ["arn:aws:ecs:region:account-id:service/cluster-name/application-stack-details-v000"]
     }) >> new DescribeServicesResult().withServices(
-      new Service(serviceName: "application-stack-details-v000", createdAt: new Date(1)))
+      new Service(serviceName: "application-stack-details-v000", createdAt: new Date(1), status: "ACTIVE"))
 
     takenSlots == [
         new AbstractServerGroupNameResolver.TakenSlot('application-stack-details-v000', 0, new Date(1))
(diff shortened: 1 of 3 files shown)


clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/deploy/EcsServerGroupNameResolverSpec.groovy
```diff
@@ -92,4 +110,85 @@ class EcsServerGroupNameResolverSpec extends Specification {
...
+      request.cluster == ecsClusterName
+      request.services == ["application-stack-details-v001", "application-stack-details-v002"]
+    }) >> new DescribeServicesResult().withServices(
+      new Service(serviceName: "application-stack-details-v001", createdAt: new Date(1), status: "ACTIVE"),
+      new Service(serviceName: "application-stack-details-v002", createdAt: new Date(2), status: "ACTIVE"))
+    ecsClient.describeServices({DescribeServicesRequest request ->
+      request.cluster == ecsClusterName
+      request.services == ["application-stack-details-v003"]
+    }) >> new DescribeServicesResult().withServices(
+      new Service(serviceName: "application-stack-details-v003", createdAt: new Date(3), status: "DRAINING"))
```
allisaurus: what happens if the service returned is `INACTIVE` ? will we create a new server group with `-V003` ? this feels like weird behavior if there's an active service called `-V004` but i'm also not sure how to prevent it w/o describing until we find one that's `MISSING`

clareliguori: Correct, v003 would be created.  The "should not skip names for inactive ECS services" unit test below covers this.

The resolver already has the full list of "taken slots" aka the list of active services, so it would know that there is a v004 and pick the right number appropriately.  This check just makes sure that we don't attempt to use a name that isn't a "taken slot", but can't actually be used.

allisaurus: > The resolver already has the full list of "taken slots" aka the list of active services, so it would know that there is a v004 and pick the right number appropriately.

Ah ha! that's the piece i was missing. in that case :shipit:

PR #4343, opened 2020-02-22
fix(aws): Fixing CLBs listener policies getting dropped on upsert

jeyrschabu:
- updated to now handle listener policies earlier in the operation
- fixed a bug where if a call to drop a listener failed,
the load balancer is left in a dirty state


jeyrschabu:
The previous fix was unsuccessful as the call to reset the policies was happening way after listeners were dropped and recreated.

This patch also contains a bug fix where the load balancer would be left in a dirty state, with some of its listeners deleted in case of an exception while removing listeners.
This fix ensures we rollback the load balancer to its previous state and fail the operation.

FYI @alanmquach we no longer even need to pass the `policyNames` in the request because that info is already available.

PR #4354, opened 2020-02-26
fix(docker): Fix docker registry health check

ezimanyi:
Fixes spinnaker/spinnaker#5173.

* test(kubernetes): Add tests to KubernetesHealthIndicator 

  Before refactoring this class to push some logic to the base class, add some tests on it.  This also adds a couple of useful stub classes to make it easier to create stub accounts.

* refactor(kubernetes): Push logic for kubernetes health to base 

  The logic for checking the health of a Kubernetes account is similar to what we want for docker accounts (and likely for other accounts). Namely, we want to report errors with an individual account, but don't want to bring down all of clouddriver because of failures with individual accounts. Make a base class that can be extended by cloud providers to implement this functionality.

  This also makes one small change to the way unhealthy accounts are reported in detailed health check results; as the accounts will already appear under a key of 'kubernetes', I've eliminated the redundant prefix of
  'kubernetes:' from each of the account names.

* feat(kubernetes): Add metrics to kubernetes health check 

  Add support to the base AccountHealthIndicator to report the number of unhealthy accounts, which then adds support for this to the Kubernetes health indicator.

* fix(docker): Fix docker registry health check 

  The docker registry health check currently fails if any single docker account is unreachable.  (Though it will not fail after startup if it has passed once.)

  This means that a single bad account can cause all of clouddriver to fail to start, which is bad for failure isolation. Rewrite the health check to use the new AccountHealthIndicator class which will report errors (both as metrics and in the detailed health view) but will not cause clouddriver to report itself as DOWN.


clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/health/DockerRegistryHealthIndicator.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright 2016 Google, Inc.
```
ethanfrogers: Is this a new flie? If so `s/2016/2020`.

ezimanyi:
* fix(kubernetes): Fix tests depending on map iteration order 

  When applied to a Map, containsExactly requires that iterating over the map returns the supplied entries in order. As we are not using an ordered map, this will lead to intermittent test failures.

  Replace this with containsOnly, which asserts the map contains exactly the supplied entries, in any order.

  Also fix a copyright header.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/AccountHealthIndicator.java
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * <p>It will always report a status of UP, to prevent issues with a single cloud provider from
+ * bringing down all of clouddriver, but any errors associated with individual accounts will appear
+ * in the detailed health information.
+ *
+ * @param <T> The type of account credentials this health indicator supports
+ */
+public abstract class AccountHealthIndicator<T extends AccountCredentials>
+    implements HealthIndicator {
+  @Nonnull private volatile Health health = new Health.Builder().up().build();
```
maggieneterval: Further inspiration to finish the concurrency chapter of EJ!

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/AccountHealthIndicator.java
```diff
@@ -32,11 +35,30 @@
...
+ *
  * @param <T> The type of account credentials this health indicator supports
  */
 public abstract class AccountHealthIndicator<T extends AccountCredentials>
     implements HealthIndicator {
   @Nonnull private volatile Health health = new Health.Builder().up().build();
+  @Nonnull private final AtomicLong unhealthyAccounts = new AtomicLong(0);
+
+  /**
+   * Create an {@code AccountHealthIndicator} reporting metrics to the supplied registry, using the
```
maggieneterval: This might be a stupid question, my brian is :hankey: today. What's an example of a "supplied registry?" Where are these health metrics going?

ezimanyi: This is a `com.netflix.spectator.api.Registry`, which allows you to register data points that you want to track and report metrics on.  So in this case by calling `PolledMeter.using(registry).withName("name").monitorValue(value)` we're telling the registry to keep track of the number behind the reference that `value` points to.  (So it has to be some reference rather than a primitive since it needs to hold a constant reference to monitor and will periodically look and record the value behind that reference.)

Then if you send a GET request to `http://localhost:7002/spectator/metrics` you'll get a data dump of all the metrics that clouddriver knows about.  I think that most integrations around monitoring effectively poll that endpoint and feed the data into monitoring systems.  (The monitoring daemon for example will poll that endpoint and handle pushing to any systems you have.)

So I guess in summary, here `Registry` is the global registry of all metrics that clouddriver has (which is a bean that is created in the base config) and we need it to be able to register another metric.

maggieneterval: Thank you!

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/AccountHealthIndicator.java
ezimanyi: Fair point, I think I may have been over-influenced by the fact that the class I was refactoring from used `AtomicReference` in a similar way.  Is there some guaranteed bound on when a write to a non-volatile variable would be visible by another thread?  I suppose it would never approach minutes (or even seconds) but just curious.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/health/KubernetesHealthIndicator.java
```diff
@@ -16,62 +16,39 @@
...
-      } catch (Exception e) {
-        String accountName = String.format("kubernetes:%s", accountCredentials.getName());
-        warnings.put(accountName, e.getMessage());
-      }
+  @Override
+  protected Optional<String> accountHealth(KubernetesNamedAccountCredentials accountCredentials) {
+    try {
+      accountCredentials.getCredentials().getDeclaredNamespaces();
+      return Optional.empty();
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
ezimanyi: Good point, fixed!

ezimanyi:
This is so exciting, I now have one of each of the following commits on this PR: `test`, `refactor`, `feat`, `fix`, `style`

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/AccountHealthIndicator.java
plumpy: No guaranteed bound that I know of, but yeah, I can't imagine it even being seconds...

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/health/KubernetesHealthIndicator.java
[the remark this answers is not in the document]
plumpy: Oh, there's one more in `DockerRegistryHealthIndicator`, sorry...

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/AccountHealthIndicator.java
plumpy: I'm definitely now musing about things I know very little about, but it [looks like the kernel context switches every 1.2–1.5µs](https://eli.thegreenplace.net/2018/measuring-context-switching-and-memory-overheads-for-linux-threads/)

Also from that link: "Every thread has some working set of memory, all or some of which is in the cache; when we switch to another thread, all this cache data becomes unneeded and is slowly flushed out, replaced by the new thread's data."

Of course, in a container where there are very few applications running, probably the cache flushes less often than in a VM, but still...

This could make an interesting java-users@ thread. Martin Buchholz, who held the title "JDK Concurrency Czar" at Sun, would almost certainly weigh in :)

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/health/KubernetesHealthIndicator.java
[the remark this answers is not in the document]
ezimanyi: Whoops, fixed now!

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/AccountHealthIndicator.java
ezimanyi: Thanks for the link!  Bookmarking it to read later when I have a block of time and won't have to context switch.

PR #4355, opened 2020-02-26
fix(aws): add .get() to optional known to be present

maggieneterval:
Fixes "could not find matching constructor" error introduced in https://github.com/spinnaker/clouddriver/pull/4336


plumpy:
👊 🕺 

That is meant to portray me angrily shaking my fist at Groovy.

PR #4360, opened 2020-02-28
feat(model): Adds extension point for model customization on read

cfieber:
Cluster/ServerGroupViewModelPostProcessor can augment the view model
returned from the clusters and serverGroups endpoints.

supports extended attributes on Cluster and ServerGroup models.

Empty by default, but allows a customization hook to enhance the
cloudprovider cached model via an integration and serve
extra/computed/enhanced data along with the response payload.

Updates Titus and AWS cloudproviders to implement via JsonAnyGetter/Setter


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ApplicationProvider.groovy
```diff
@@ -17,7 +17,8 @@
 package com.netflix.spinnaker.clouddriver.model
 
 import com.netflix.spinnaker.clouddriver.documentation.Empty
-import com.netflix.spinnaker.clouddriver.documentation.Nullable
+
+import javax.annotation.Nullable
```
robzienert: 🎉

PR #4364, opened 2020-02-28
feat(huaweicloud): implement cache agent of subnet

zengchen1024:
This pr implements the cache agent of subnet. The cloud provider proposal is spinnaker/governance#17


zengchen1024:
@marchello2000 Could you take a look at this pr? Thanks very much!

clouddriver-huaweicloud/src/main/java/com/netflix/spinnaker/clouddriver/huaweicloud/cache/Keys.java
```diff
@@ -77,59 +78,50 @@ private static String getCloudProviderId() {
     return HuaweiCloudProvider.ID;
   }
 
+  private static Map emptyMap() {
```
marchello2000: you had this in your other PR too, just curious what does this buy you (vs. just calling `Collection.emptyMap` which itself is already a `final static` (https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/Collections.java#l4557)

I don't care/think it matters either way, but curious why you went this route

zengchen1024: I just don't like to write Collections.emtyMap so many times. Any better suggestions? thanks!

PR #4379, opened 2020-03-03
fix(cloudFoundry): fix cloudFoundry job provider

ethanfrogers:
if cloud foundry is enabled, this job provider breaks run job for all
other cloud providers because of the exception thrown in
getFileContents. other cloud providers simply return null if this method
isn't implemented. this whole area could probably use a refactored.
currently, every job provider is queried regardless of whether or not
the account is for that cloud provider.


ethanfrogers:

ethanfrogers:

ethanfrogers:

PR #4383, opened 2020-03-03
feat(plugins): add plugin dep and config

claymccoy:
this allows loading of plugins like other spin services


clouddriver-core/clouddriver-core.gradle
```diff
@@ -1,4 +1,6 @@
 dependencies {
+  api("com.netflix.spinnaker.kork:kork-plugins")
```
cfieber: should this be api (vs implementation)

where else in the code is stuff from this lib referenced

claymccoy: I copied what was done in orca

PR #4384, opened 2020-03-03
feat(exceptions): consume SpinnakerRetrofitErrorHandler

maggieneterval:
Part of https://github.com/spinnaker/spinnaker/issues/5473 (Followup to POC https://github.com/spinnaker/clouddriver/pull/4325)

Registers `SpinnakerRetrofitErrorHandler` with each `RestAdapter` and replaces caught `RetrofitError`s with new exceptions that extend `SpinnakerException`. Each module has its own commit. I made no changes to `clouddriver-cloudfoundry` because they already wrote their own [`CloudFoundyApiException`](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryApiException.java).

I will add a followup commit to this PR when when we have merged the [Kork PR](https://github.com/spinnaker/kork/pull/538) that exports `SpinnakerRetrofitErrorHandler`, `SpinnakerServerException`, `SpinnakerNetworkException`, and `SpinnakerHttpException` from `kork-exceptions` rather than duplicating them here in Clouddriver.

### DIFF
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
@@ -23,11 +23,14 @@ import com.netflix.spinnaker.clouddriver.docker.registry.api.v2.auth.DockerBeare
 import com.netflix.spinnaker.clouddriver.docker.registry.api.v2.auth.DockerBearerTokenService
 import com.netflix.spinnaker.clouddriver.docker.registry.api.v2.exception.DockerRegistryAuthenticationException
 import com.netflix.spinnaker.clouddriver.docker.registry.api.v2.exception.DockerRegistryOperationException
+import com.netflix.spinnaker.clouddriver.exceptions.SpinnakerHttpException
+import com.netflix.spinnaker.clouddriver.exceptions.SpinnakerRetrofitErrorHandler
+import com.netflix.spinnaker.clouddriver.exceptions.SpinnakerServerException
 import groovy.util.logging.Slf4j
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
+import org.springframework.http.HttpStatus
 import retrofit.RestAdapter
-import retrofit.RetrofitError
 import retrofit.client.Response
 import retrofit.converter.GsonConverter
 import retrofit.http.GET
@@ -162,6 +165,7 @@ class DockerRegistryClient {
       .setEndpoint(address)
       .setClient(okClientProvider.provide(address, clientTimeoutMillis, insecureRegistry))
       .setLogLevel(RestAdapter.LogLevel.NONE)
+      .setErrorHandler(SpinnakerRetrofitErrorHandler.getInstance())
       .build()
       .create(DockerRegistryService)
     this.converter = new GsonConverter(new GsonBuilder().create())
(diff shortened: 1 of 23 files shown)


clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/deploy/ops/AbstractEurekaSupport.groovy
```diff
@@ -149,26 +154,10 @@ abstract class AbstractEurekaSupport {
...
-            errors[instanceId] = retrofitError
-          } else {
-            skipped.add(instanceId)
-          }
-
-          task.updateStatus phaseName, errorMessage
-        } else {
-          errors[instanceId] = retrofitError
-        }
+      } catch (NotFoundException e) {
```
maggieneterval: @robzienert because of the special error handling in this class, I'm not actually sure catching the new error types improves the readability; let me know if you'd prefer I leave this file as-is.

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -458,6 +463,8 @@ class DockerRegistryClient {
         } else {
           throw error
         }
+      } catch (SpinnakerServerException error) {
+        throw error
```
plumpy: is there some groovy reason why this is needed? Catching and immediately rethrowing serves no purpose that I can see?

(See also `DockerRegistryNamedAccountCredentials` and `LocalFileUserDataProvider`)

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/deploy/ops/AbstractEurekaSupport.groovy
```diff
@@ -216,23 +233,35 @@ abstract class AbstractEurekaSupport {
...
+      } catch (SpinnakerNetworkException e) {
+        if (retryCount >= (maxRetries - 1)) {
+          throw e
+        }
+        retryCount++
+        AbstractEurekaSupport.log.debug("[$phaseName] - Failed calling external service ${e.getMessage()}")
+        sleep(getDiscoveryRetryMs())
+      } catch (SpinnakerServerException e) {
+        AbstractEurekaSupport.log.debug("[$phaseName] - Failed calling external service ${e.getMessage()}")
+        throw e
```
plumpy: Not a huge deal, logging and rethrowing is generally an anti-pattern... either handle the exception or let the caller decide what to do with it. Most likely the caller is going to log this somewhere anyway...

maggieneterval:
> It feels like it would be a lot easier to register some sort of global exception handler to handle RetrofitError exceptions, as described here: https://www.toptal.com/java/spring-boot-rest-api-error-handling

:100: :book: Great read, will investigate creating a global error handler that extends `ResponseEntityExceptionHandler` and report back. Thanks, @plumpy! :hugs:

maggieneterval:
Breaking news: kork has [`GenericExceptionHandlers`](https://github.com/spinnaker/kork/blob/master/kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java#L81) that already should be processing `RetrofitError`s for each microservice: I think we just need to augment the existing logic to actually extract a useful message. Going to close these PRs for now and validate that approach before re-opening.

PR #4386, opened 2020-03-04
fix(elasticsearch): Upgrading to ElasticSearch 6 for entity tags

marchello2000:
This change follows up on the change https://github.com/spinnaker/clouddriver/pull/2913.
    * Remove ES2.x support - now all docs using mapping type `_doc` recommended by ES6 and only one is allowed.
    * Change the queries to use `entityRef.entityType` instead of the document type.
    * Fixup ES template to not use `not_analyzed` but instead use `keyword` (with exception of `entityRef.entityType` which we wildcard query)
    * Adding an `id` field to the document because the built-in `_id` field is no longer wildcard queryable
    * Changing to use a container for the unittests

### DIFF
--- a/clouddriver-elasticsearch/clouddriver-elasticsearch.gradle
+++ b/clouddriver-elasticsearch/clouddriver-elasticsearch.gradle
@@ -7,15 +7,16 @@ dependencies {
   implementation "com.netflix.spinnaker.kork:kork-security"
   implementation "com.squareup.retrofit:retrofit"
   implementation "org.codehaus.groovy:groovy-all"
-  implementation("org.elasticsearch:elasticsearch:2.4.1") {
+  implementation("org.elasticsearch:elasticsearch:6.8.6") {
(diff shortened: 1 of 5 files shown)


marchello2000:
There is currently an issue with this PR in that we use `nodeBuilder` in the unittests. NodeBuilder has been deprecated and removed in ES5. I am not sure/haven't been able to figure out how to work around this, so any advice is appreciated!

ajordens:
This looks reasonable to me ... as we talked about in person the elasticsearch test container might be a legitimate path forward to replace what that `Spec` was doing previously.

marchello2000:
Thanks @ezimanyi and @ajordens . I ended up going the test container route which found a few issues in the change. They are all resolved now, if you can take another quick look.

ezimanyi:
Still looks reasonable to me!

marchello2000:
Thank you!

PR #4394, opened 2020-03-06
fix(repository): Enforce serialization order for JedisTask

ezimanyi:
* test(repository): Add tests to JedisTask 

  Before making a minor change to JedisTask, add some tests of the serialization logic.

* fix(repository): Enforce serialization order for JedisTask 

  There have been a number of end-to-end test failures recently due to race conditions where orca fetches a task status from clouddriver and receives a status of complete, but incomplete task results.

  The fields of the task are computed on-demand by querying the repository. This means that the serialized task may not be internally consistent; each field will reflect the state of the task in the repository at the time that field's accessor was called during serialization.

  This is in general a difficult problem to solve with redis, which does not support atomic reads of multiple keys, but has been solved in the SQL repository by fetching all data in a single query. As a workaround, we'll instruct Jackson to serialize the status first. The reason is that consumers tend to use the status field to check if a task is complete, and expect the other fields to be filled out if it is. If there is an inconsistency between the status and other fields, we'd rather return a stale value in the status field than in other fields.

  In general, returning an older status (ie, still running) and newer other fields will just cause clients to poll again until they see the updated status. Returning a newer status (ie, completed or failed) but stale values in other fields will in general cause clients to use these stale values, leading to bugs.

### DIFF
--- a/clouddriver-core/src/test/java/com/netflix/spinnaker/clouddriver/data/task/jedis/JedisTaskTest.java
+++ b/clouddriver-core/src/test/java/com/netflix/spinnaker/clouddriver/data/task/jedis/JedisTaskTest.java
@@ -28,7 +28,6 @@ import com.netflix.spinnaker.clouddriver.data.task.DefaultTaskStatus;
 import com.netflix.spinnaker.clouddriver.data.task.TaskState;
 import java.nio.charset.StandardCharsets;
 import org.assertj.core.api.AssertionsForClassTypes;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.mockito.InOrder;
 import org.mockito.Mockito;
@@ -62,8 +61,6 @@ final class JedisTaskTest {
         .isEqualTo(objectMapper.readTree(expectedResult));
   }
 
-  // Ignore this test as the order is currently not deterministic
-  @Disabled
   @Test
   void statusComputedFirst() throws Exception {
     RedisTaskRepository taskRepository = mock(RedisTaskRepository.class);
--- a/clouddriver-core/src/test/java/com/netflix/spinnaker/clouddriver/data/task/jedis/JedisTaskTest.java
+++ b/clouddriver-core/src/test/java/com/netflix/spinnaker/clouddriver/data/task/jedis/JedisTaskTest.java
@@ -61,6 +61,7 @@ final class JedisTaskTest {
         .isEqualTo(objectMapper.readTree(expectedResult));
   }
 
+  // See the large comment on the top of JedisTask for this test's rationale
   @Test
   void statusComputedFirst() throws Exception {
     RedisTaskRepository taskRepository = mock(RedisTaskRepository.class);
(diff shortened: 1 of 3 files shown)


ezimanyi:
This fixes the root cause that was worked around in spinnaker/orca#3495.  After opening that PR, I found another failed end-to-end test (this time in the kubernetes provider) with the same root cause.  That one can't be easily fixed with the `kato.result.expected` flag, as that particular task does report some results along the way; so we do have a some results (just not all of them) if the results haven't been fully updated.  Rather than make a more complex workaround (such as adding a key `kato.result.expectThisFieldToExist`) I tried to fix the race condition, which I think this PR accomplishes.

clouddriver-core/src/test/java/com/netflix/spinnaker/clouddriver/data/task/jedis/JedisTaskTest.java
```diff
@@ -0,0 +1,76 @@
...
+    when(taskRepository.getHistory(eq(task))).thenReturn(ImmutableList.of(oldStatus, status));
+    when(taskRepository.getResultObjects(eq(task))).thenReturn(ImmutableList.of(results));
+    when(taskRepository.currentState(eq(task))).thenReturn(status);
+
+    String result = objectMapper.writeValueAsString(task);
+    String expectedResult =
+        Resources.toString(JedisTaskTest.class.getResource("task.json"), StandardCharsets.UTF_8);
+
+    // Compare the parsed trees of the two results, which is agnostic to key order
+    AssertionsForClassTypes.assertThat(objectMapper.readTree(result))
```
ezimanyi: Using the normal `assertThat` doesn't work here as the result of `objectMapper.readTree(String)` is a `JsonNode`, which implements `Iterable`, so `AssertJ` tries to use the equality check for an iterable rather than the built-in `.equals` of `JsonNode`.

plumpy: You could also cast: `assertThat((Object) objectMapper.readTree(result))` but I guess that's no better really.

clouddriver-core/src/test/java/com/netflix/spinnaker/clouddriver/data/task/jedis/JedisTaskTest.java
```diff
@@ -0,0 +1,76 @@
...
+        .isEqualTo(objectMapper.readTree(expectedResult));
+  }
+
+  @Test
```
plumpy:
```suggestion
  // See the large comment on the top of JedisTask for this test's rationale
  @Test
```

clouddriver-core/src/test/java/com/netflix/spinnaker/clouddriver/data/task/jedis/JedisTaskTest.java
ezimanyi: Oh, I figured out where to reply to this...[AssertsionsForClassTypes](https://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/api/AssertionsForClassTypes.html) is the AssertJ recommended way of dealing with ambiguity in asserting on a class or one of its interfaces.

ezimanyi:

PR #4395, opened 2020-03-06
fix(config): Fix OkHttpClientConfiguration instantiation

jonsie:
This is a very groovy PR.

### DIFF
--- a/clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/config/EurekaProviderConfiguration.groovy
+++ b/clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/config/EurekaProviderConfiguration.groovy
@@ -46,6 +46,7 @@ import org.springframework.core.env.ConfigurableEnvironment
 import retrofit.converter.Converter
 import retrofit.converter.JacksonConverter
 
+import javax.inject.Provider
 import java.util.regex.Pattern
 
 @Configuration
@@ -88,7 +89,7 @@ class EurekaProviderConfiguration {
   }
 
   private OkHttpClientConfiguration eurekaOkHttpClientConfig() {
-    new OkHttpClientConfiguration(eurekaClientConfig(), new OkHttpMetricsInterceptor(registry, true))
+    new OkHttpClientConfiguration(eurekaClientConfig(), new OkHttpMetricsInterceptor({ registry }, true))
   }
 
   private static Converter eurekaConverter() {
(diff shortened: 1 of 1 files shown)


clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/config/EurekaProviderConfiguration.groovy
```diff
@@ -46,6 +46,7 @@ import org.springframework.core.env.ConfigurableEnvironment
 import retrofit.converter.Converter
 import retrofit.converter.JacksonConverter
 
+import javax.inject.Provider
```
plumpy: You don't need this, do you?

plumpy:
Looks like [this is the only other one](https://github.com/spinnaker/kork/blob/7dc4e98d14b476189c88eb17852bf9925e96a430/kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttpClientComponents.groovy#L44), I think?

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/config/EurekaProviderConfiguration.groovy
jonsie: Ah indeed, left here by mistake.

jonsie:
> Looks like [this is the only other one](https://github.com/spinnaker/kork/blob/7dc4e98d14b476189c88eb17852bf9925e96a430/kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttpClientComponents.groovy#L44), I think?

This one is actually passing in a `Provider<Registry>` so we're good there.

jonsie:
> Oof, I [fixed this in a Java class in `fiat`](https://github.com/spinnaker/fiat/pull/575) since it caused a compilation error and wondered why it didn't lead to more build breakage... but of course Groovy won't complain.
> 
> I wonder how many more of these are floating out there in Groovy files. We probably need to cherry-pick them all to 1.19.

Yes, this will need to be cherry picked in.  We'll also want kork 7.25.1 in 1.19 as it has a critical fix for OkHttp client configuration properties -- https://github.com/spinnaker/kork/commit/6e885cdd3638358d253fda03cc17ba41504cb333#diff-74a9793617cfd7572ba83c8e96b904d1

plumpy:

PR #4401, opened 2020-03-06
fix(aws): fix more extraAttributes renaming spots that were missed

cfieber:


cfieber:

PR #4405, opened 2020-03-10
fix(titus/entitytags): remove schema version from key returned by TitusClusterProvider

marchello2000:
EntityTag reconciliation process gets all server groups from the provider and matches them against the existing entity tags.
Titus returns keys in the form
```
titus:servergroups:v2:<application>:...
```
while the tags are stored in the form
```
titus:servergroups:<application>:...
```

This change removes the schema key (`v1` or `v2`) from the key IDs that are returned. This API (`getServerGroupIdentifiers`) is only used by the tag reconciler


marchello2000:
going to merge since i want to this to go in soon, please send me feedback

PR #4409, opened 2020-03-11
fix (ecs): Unstable deployments for task definitions with more than one container  (#5544)

atyutyunnik:
The original code will not finish deployment and will eventually time out, in spite of task being successfully deployed by ECS and running (spinnaker/spinnaker#5544)

        "containers": [:

            "containers": [
                {
                    "networkBindings": [],
                    "networkInterfaces": [],
                },
                {
                    "lastStatus": "RUNNING",
                    "networkBindings": [
                        {
                            "bindIP": "0.0.0.0",
                            "containerPort": 8080,
                            "hostPort": 32773,
                            "protocol": "tcp"
                        }
                }]


atyutyunnik:
> can we add unit tests to verify that tasks with network bindings in containers other than the first one are correctly evaluated?

Normally, that's the first thing I do when introducing a change, but I was quite in a hurry with the fix for my own purposes and I also need to figure out how these particular unit tests work.

allisaurus:
quoting @atyutyunnik from the slack convo about this issue:

> I have re-verified it with the original 1.19.0 - the problem is there, UT detects it; applied the fix - UT is happy

🎉

clareliguori:

allisaurus:

atyutyunnik:
For 1.18, you have to use PR# 4411. PR4409 is not compatible with the code base

PR #4411, opened 2020-03-11
fix (ecs): Unstable deployments for task definitions with more than one container (#5544)

atyutyunnik:
The original code will not finish deployment and will eventually time out, in spite of task being successfully deployed by ECS and running (https://github.com/spinnaker/spinnaker/issues/5544)

            "containers": [
                {
                    "networkBindings": [],
                    "networkInterfaces": [],
                },
                {
                    "lastStatus": "RUNNING",
                    "networkBindings": [
                        {
                            "bindIP": "0.0.0.0",
                            "containerPort": 8080,
                            "hostPort": 32773,
                            "protocol": "tcp"
                        }
                }]

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
@@ -144,17 +144,24 @@ public class TaskHealthCachingAgent extends AbstractEcsCachingAgent<TaskHealth>
           continue;
         }
 
-        TaskHealth taskHealth;
-        if (task.getContainers().get(0).getNetworkBindings().size() >= 1) {
-          taskHealth =
-              inferHealthNetworkBindedContainer(
-                  amazonloadBalancing,
-                  task,
-                  containerInstance,
-                  serviceName,
-                  service,
-                  taskDefinition);
-        } else {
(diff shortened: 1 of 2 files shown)


ezimanyi:
@atyutyunnik : Thanks for the PR!  In general our process is to open fixes against the `master` branch then cherry-pick them back to release that need them.  Can you please open this agains the `master` branch?  Thanks!

atyutyunnik:
ezimanyi, yes, I already have (see PR 4409), but it's not source-compatible with release 1.18. Hence an additional PR

ezimanyi:
Oh, thanks, I mis-understood!  If the code has changed enough that it can't be cleanly cherry-picked to 1.18, then it's fine to open this against 1.18 to manually back-port the fix.

It will be important to cherry-pick the master change to 1.19 though, as otherwise users upgrading from 1.18 to 1.19 will get re-broken.

allisaurus:
@atyutyunnik can you please add the unit test from #4409 to this PR as well?

atyutyunnik:
> @atyutyunnik can you please add the unit test from #4409 to this PR as well?

@allisaurus, cherry-picking doesn't seem possible. I had to extend the unit-test with:
...
`
    then:
    amazonloadBalancing.describeTargetHealth({ DescribeTargetHealthRequest request ->
`
...

otherwise, there would be an NPE at 332 of TaskHealthCachingAgent.java:

`
    if (describeTargetHealthResult.getTargetHealthDescriptions().isEmpty()) {
`

Now the unit-test is passing - please review

PR #4412, opened 2020-03-12
fix(sql): use jooq constructs instead of string concatenations

marchello2000:
There were a few places where we use string concatenation instead of using first class jooq constructs for interacting with SQL.
Using jooq gives us proper sql escaping and it's also easier to read


jasonmcintosh:

jasonmcintosh:

PR #4417, opened 2020-03-12
fix(ecs): Remove extra evictions from TaskHealthCachingAgent

allisaurus:
potential fix or remediation for https://github.com/spinnaker/spinnaker/issues/5551

Prior to 1.19.x, `TaskHealthCachingAgent` was calling AWS ELB to describe target (task) health, and evicting services and task definitions when an associated target health was not found. 

However, now this agent is no longer calling AWS ELB directly, so it's not an authority on the staleness of these resources.

### Testing
Deployed this change locally, deployed successfully.


allisaurus:

awsiv:
Awesome stuff @allisaurus  This seems to have fixed it! :)
Both deck and clouddriver behave as expected now 💯

PR #4420, opened 2020-03-13
feat(build): Remove TravisCI and replace with GitHub Actions

jonsie:


marchello2000:
prob need to fix https://github.com/spinnaker/clouddriver/blob/master/.mergify.yml to not look at travis builds anymore and instead look at github

PR #4421, opened 2020-03-13
fix(elasticsearch): Fixing entity type case

jeyrschabu:
- lowercased entity type


marchello2000:
would be good to slap a UT for this

jeyrschabu:
@marchello2000 agreed

PR #4429, opened 2020-03-15
fix(ecs): Correctly scope TargetHealthCachingAgent describe calls

allisaurus:
Fixes https://github.com/spinnaker/spinnaker/issues/5574

### Testing

Before change: reproduced errors in env with mulitple AWS accounts configured in same region
```
com.netflix.spinnaker.clouddriver.ecs.provider.EcsProvider:ecs-my-aws-devel-acct/us-east-1/TargetHealthCachingAgent completed with one or more failures

com.amazonaws.services.elasticloadbalancingv2.model.AmazonElasticLoadBalancingException: 'arn:aws:elasticloadbalancing:us-east-1:ACCOUNT:targetgroup/nlb-targetGroup/7e540bba02fc1ac2' is not a valid target group ARN (Service: AmazonElasticLoadBalancing; Status Code: 400; Error Code: ValidationError; Request ID: ...)
```

After: no such errors found in clouddriver.log 🎉

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TargetHealthCachingAgent.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TargetHealthCachingAgent.java
@@ -72,6 +72,7 @@ public class TargetHealthCachingAgent extends AbstractEcsAwsAwareCachingAgent<Ec
     EcsTargetGroupCacheClient targetGroupCacheClient =
         new EcsTargetGroupCacheClient(awsProviderCache, objectMapper);
     Set<String> targetGroups = fetchTargetGroups(targetGroupCacheClient);
+    log.debug("Fetched {} target groups for which to get target healths", targetGroups.size());
 
     AmazonElasticLoadBalancing amazonLoadBalancing =
         amazonClientProvider.getAmazonElasticLoadBalancingV2(account, region, false);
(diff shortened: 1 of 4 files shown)


clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TargetHealthCachingAgent.java
```diff
@@ -114,7 +115,15 @@ public TargetHealthCachingAgent(
     Collection<String> targetGroupKeys =
         awsProviderCache.filterIdentifiers(TARGET_GROUPS.getNs(), searchKey);
 
-    List<EcsTargetGroup> tgList = cacheClient.find(targetGroupKeys);
+    if (targetGroupKeys.size() > 0) {
+      log.debug("Found " + targetGroupKeys.size() + " target group keys in " + getAgentType());
```
clareliguori: Would it be useful to also log the number of results for this account after filtering below? Or is that logged elsewhere already

allisaurus: the number of target groups returned by this func (after filtering by account and deduping) is logged on line 75, prior to calling `describeTargetHealth`

clareliguori: No that works

clareliguori:

PR #4437, opened 2020-03-17
fix(titus/instance): Temporarily expose agentId to access IPv6

caseyhebebrand:
The Titus API does not expose IPv6 addresses yet. In the meantime, accessing the task's `agent.instanceId` enables a secondary lookup of the network interfaces so we can obtain the IPv6 address(es). 

This workaround is necessary for some other work around IPv6 support. In the future, the Titus API will be updated and we can directly replace the `agentId` with `ipv6Addresses` from Titus.

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/Task.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/Task.java
@@ -40,6 +40,7 @@ public class Task {
     startedAt = getTimestampFromStatus(grpcTask, TaskStatus.TaskState.StartInitiated);
     finishedAt = getTimestampFromStatus(grpcTask, TaskStatus.TaskState.Finished);
     containerIp = grpcTask.getTaskContextOrDefault("task.containerIp", null);
+    //  The agentId will be used temporarily by deck to lookup IPv6 until Titus API is updated.
     agentId = grpcTask.getTaskContextOrDefault("agent.instanceId", null);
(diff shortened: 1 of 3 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/Task.java
```diff
@@ -40,6 +40,7 @@ public Task(com.netflix.titus.grpc.protogen.Task grpcTask) {
     startedAt = getTimestampFromStatus(grpcTask, TaskStatus.TaskState.StartInitiated);
     finishedAt = getTimestampFromStatus(grpcTask, TaskStatus.TaskState.Finished);
     containerIp = grpcTask.getTaskContextOrDefault("task.containerIp", null);
+    agentId = grpcTask.getTaskContextOrDefault("agent.instanceId", null);
```
jeyrschabu: Guessing deck would be doing the lookup of the IP based on this agent Id?

Hopefully the long term solution comes sooner rather than later

aravindmd: worth adding a comment here about what this field is used for when titus supports it we can clean it up.

caseyhebebrand: Yes deck will do the lookup. They are targeting Q2. It was supposed to be end of Q1 but it is delayed from COVID wfh shifting some priorities.

PR #4441, opened 2020-03-18
fix(provider/cf): Don't get spaces on every call to /credentials

ncknt:
A call to `/credentials` (even without `?expand=true`) currently results in a potentially expensive call to get spaces in the CF account. This uses the same strategy that `KubernetesV2Credentials`  uses for namespaces.


ncknt:
@ezimanyi Perhaps you could take a look? It's pretty much your namespace fix in the v2 credentials but for CloudFoundry's spaces.

ncknt:

ncknt:

PR #4449, opened 2020-03-19
fix(kubernetes/v1): Don't pass force flag to patch operation

ezimanyi:
The patchNamespacedStatefulSet is passing force=false; the Kubernetes API server doesn't allow this flag to be present at all for non-apply patch operations, which is causing a server-side failure.

This only breaks on clusters >=1.14 as the flag was not present before. (I'm not sure if the client library was detecting the old version of the cluster and removing the flag before calling the API, or if the API was just ignoring the unknown flag, but either way the flag had no effect.)


ezimanyi:
Relevant commit introducing the validation we're hitting: https://github.com/kubernetes/apimachinery/commit/de0a49ba28f8bb125d0d9fc277dbe193eff8439a

ezimanyi:

PR #4456, opened 2020-03-21
fix(core): Only log relevant details of description

ezimanyi:
Closes spinnaker/spinnaker#5585.

When we encounter a description that should have application restrictions but doesn't, we log a warning including the full description JSON. In some cases (ex: Kubernetes) this description may contain sensitive information that ideally wouldn't be logged.

Ideally really sensitive values would come from artifacts so they aren't in the pipeline JSON at all, but let's nonetheless cut back on how much we are logging here to reduce the chance of sensitive information ending up in the logs.

In particular, it's not clear how much the full description would help in debugging, as generally what is important is the class of the description (to check whether it implements ApplicationNameable or ResourcesNameable) and and the account, both of which will still be logged after this change.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/deploy/DescriptionAuthorizer.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/deploy/DescriptionAuthorizer.java
@@ -18,7 +18,6 @@ package com.netflix.spinnaker.clouddriver.deploy;
 
 import static java.lang.String.format;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spectator.api.Id;
 import com.netflix.spectator.api.Registry;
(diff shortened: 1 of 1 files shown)


ezimanyi:
@ajordens @ttomsu : Small reduction in how much we're logging from the changes in #3117; I think this should still be enough info to debug if necessary but would appreciate a quick look.  Thanks!

ttomsu:
Lgtm

ajordens:
Hmm ... generally it is nice to have the description source since I'd consider this an anomaly ... but I can see the point.

Is there something about (some) Kubernetes requests where it's expected to have a resource that is account restricted but not associated with an application?

ezimanyi:
@ajordens : It looks like none of the operations for the Kubnernetes V2 provider implement `ApplicationNameable` or `ResourcesNameable`, but none of them override `requiresApplicationRestriction` to `false`.  I'm guessing maybe this was just an oversight when the V2 provider was written?  Is the idea (and motivation for this log message) that all operations should surface both the accounts and applications they apply to, such that we can validate against both?

If that's the case, probably we need to make a conscious decision for the V2 provider if we want to support these application restrictions (and then go through and update all the operations) or decide not to support them and set `requiresApplicationRestriction=false` for them.  (My hesitation is that there are a lot of users using the V2 provider and I've never heard anyone complain about the lack of this permission check so am a bit wary of breaking people with a new permission check...)

Even if we do update the Kubernetes provider to stop logging this on every deploy, I *think* it still makes sense to reduce the scope of what's logged, just to prevent sensitive data in the logs.

ajordens:
> | Is the idea (and motivation for this log message) that all operations should surface both the accounts and applications they apply to, such that we can validate against both?

Generally yes. Failure to implement these interfaces correctly opens the door to some circumvention. Users likely won't know -- years ago it was an audit that pointed some of this out on the AWS (and Titus) providers.

I'm alright with adjusting the log output, 👍. I don't know enough about the intricacies of the v2 Kubernetes provider to understand what may or may not be possible.

ezimanyi:
@ajordens : Awesome, thanks for the explanation!  I've opened #5617 to discuss with Kubernetes users whether we should update the V2 provider to implement that interface; we'll discuss at the Kubernetes SIG.

ezimanyi:

PR #4460, opened 2020-03-24
chore(mergify): release-* branches require release manager approval

ezimanyi:
The mergify config currently doesn't distinguish between merging to master and merging to a release branch.  It currently never merges to release branches because mergify doesn't have permission to merge to these branches (due to branch protection rules).

I don't want to give mergify that permission without updating the config file here, as then it would merge to release branches after any approval, whereas we only want to merge to release branches if a PR is approved by a release manager.

This updates the mergify config to scope the existing rules to only apply to the master branch, and adds a new rule to merge to release branches upon approval by a release manager.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,6 +1,7 @@
 pull_request_rules:
   - name: Automatically merge on CI success and review
     conditions:
+      - base=master
       - status-success=build
       - "label=ready to merge"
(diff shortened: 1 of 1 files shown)


ezimanyi:
If this looks good I'll apply to all the other repos, and add `mergify` to `release-managers` so it's allowed to merge to the release branches.

PR #4463, opened 2020-03-24
chore(mergify): Allow Travis success on release branches

ezimanyi:
Mergify is not working on the release branches because it requires the Github actions build to return success, but we only switched to using Github actions after cutting the 1.19 branch.

To remedy this, add another rule that allows merging if Travis CI passes on a branch.  (There's no way to make an OR condition within a rule, so we need to add another rule.) This way, we'll merge if either Github actions or Travis CI reports success.  Once the branches running Travis are no longer maintained (when Spinnaker 1.23 is released) we can delete this rule.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -23,6 +23,20 @@ pull_request_rules:
         strict: smart
       label:
         add: ["auto merged"]
+  # This rule exists because releases 1.19.x and earlier are configured to build using Travis
+  # instead of Github actions. It can be deleted once we are no longer merging to these branches.
+  - name: Automatically merge release branch changes on Travis CI success and release manager review
+    conditions:
+      - base~=^release-
+      - status-success=continuous-integration/travis-ci/pr
+      - "label=ready to merge"
+      - "approved-reviews-by=@release-managers"
+    actions:
+      merge:
+        method: squash
+        strict: smart
+      label:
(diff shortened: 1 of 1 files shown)


ezimanyi:
Oh, it looks like 1.18 and 1.19 are running both github actions *and* Travis, so this rule would only be necessary to auto-merge into 1.17.  Maybe I shouldn't bother then, as I can probably handle just manually merging to 1.17 for the next couple of months, then when 1.20 is released all of the supported branches will be running github actions.

plumpy:
Wouldn't you want to modify the `.mergify.yml` file on a branch instead of master, or is that not how mergify works?

ezimanyi:
Seems like that's not how it works...I couldn't find any documentation on it, but I tested with a PR against the 1.17 branch and it was pulling in the config I just merged to master.

plumpy:
Huh, okay, well I suspect (but don't know for certain) that only _some_ of the repositories are running GHA builds on 1.18, so you might still be inclined to just merge this everywhere to cover your bases. I guess the downside to that is it'll probably sit there for another year before anyone remembers to take it back out...

ezimanyi:
I opened PRs to put this everywhere, in case you didn't notice the deluge of Github email you probably just got 😁

PR #4467, opened 2020-03-25
feat(gce): Add support for target shape on regional migs.

duftler:


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleDistributionPolicy.java
```diff
@@ -21,10 +21,11 @@
...
 import lombok.NoArgsConstructor;
 
-/** Distribution policy for selecting zones in a regional MIG. */
+/** Distribution policy for selecting zones and target shape in a regional MIG. */
 @Data
 @NoArgsConstructor
 @AllArgsConstructor
 public class GoogleDistributionPolicy {
   List<String> zones;
+  String targetShape;
```
duftler: Went with `String` over `enum` since it is likely additional target shapes will be added in the future. Didn't want to have an error when trying to cache a distribution policy created outside of Spinnaker with a target shape not (yet) supported by Spinnaker.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/AbstractGoogleServerGroupCachingAgent.java
```diff
@@ -667,6 +669,13 @@ private void setRegionConfig(GoogleServerGroup serverGroup, InstanceGroupManager
         .collect(toImmutableList());
   }
 
+  private static String getTargetShape(@Nullable DistributionPolicy distributionPolicy) {
```
maggieneterval: Since this method can return null should we annotate it with `@Nullable`?

duftler: Done.

maggieneterval:
Linking the corresponding Deck changes here for posterity: https://github.com/spinnaker/deck/pull/8089

PR #4468, opened 2020-03-25
fix(saga): Move ConditionalOnMissingBean for EventRepository to eventRepository factory method

jonsie:
On February 19th we rolled out a CloudDriver deploy and our EventRepository switched from SQL to in-memory - I can't find any change that would have caused this.  I'd like to give this a shot and see if this resolves the issue.  I'm unable to reproduce the issue locally, even when running with the exact same spring profiles as our production cluster which is odd.

### DIFF
--- a/clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/config/MemoryEventRepositoryConfig.kt
+++ b/clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/config/MemoryEventRepositoryConfig.kt
@@ -34,7 +34,6 @@ import javax.validation.constraints.Min
 import kotlin.reflect.KClass
 
 @Configuration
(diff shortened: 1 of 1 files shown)


robzienert:
Here's a theory worth trying out: I bet the `@ConditionalOnMissingBean` defined at the class-level is messing things up: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/event/config/MemoryEventRepositoryConfig.kt#L37

My rationale here is that I suspect Spring will load the configuration classes first, _then_ start building up beans. If we moved the `@ConditionalOnMissingBean` to the bean factory method, we would be initializing the in-memory configuration, but not wiring up any of the beans there in. WDYT?

jonsie:
@robzienert I like the theory and I think it's worth trying that first - I'll make the change.

jonsie:
Just deployed this, no luck.

PR #4469, opened 2020-03-25
fix(gce): Revert support for target shape on regional migs

ezimanyi:
The integration tests are failing with errors like:
```
2020-03-25 19:58:17.280  INFO 21942 --- [ionProcessor-11] c.n.s.c.data.task.jedis.JedisTask        : [ORCHESTRATION] Orchestration failed: CopyLastGoogleServerGroupAtomicOperation | GoogleJsonResponseException: [400 Bad Request
{
  "code" : 400,
  "errors" : [ {
    "domain" : "global",
    "message" : "Invalid value for field 'resource.distributionPolicy': ''. Distribution policy cannot be empty.",
    "reason" : "invalid"
  } ],
  "message" : "Invalid value for field 'resource.distributionPolicy': ''. Distribution policy cannot be empty."
}]
```
It seems like maybe we're passing along an empty field to the Google API when it should either be absent or non-empty?  Either way, reverting this so the tests pass in time to investigate.

This reverts commit 9407597200604aa5168dee8101b431f7a1e033f3.


ezimanyi:
@duftler is actively looking at this, so will temporarily remove the `ready-to-merge` label, but if there's no fix in ~1 hour we'll merge this.

duftler:
Fix is here: https://github.com/spinnaker/clouddriver/pull/4471

ezimanyi:
Integration tests passed; I'll close this now.

PR #4472, opened 2020-03-26
refactor(kubernetes): Clean up metrics code

ezimanyi:
* test(kubernetes): Add some tests to metric fetching logic

  Before making a few changes to the call to get pod metrics, add some tests around this to ensure the changes don't break existing functionality.

* refactor(kubernetes): Clean up metrics code

  This started out just trying to remove the noisy log statement
  "No output from `kubectl top` command, no metrics to report", which is logged with severity warn any time we cache an empty namespace. There's no need to log this as it's a perfectly normal situation; we asked for pod metrics and there were none.

  But then I noticed some things that could be cleaned up a bit 😄. So this PR does the following:
  * Extracts the logic to parse the output of kubectl top out of the already large job executor class and into a helper class
  * Replaces use of .split() which compiles a regex every call and requires us to deal in arrays with Guava Splitters
  * Cleans up KubernetesPodMetric, making things immutable and also removing an excessive buidler from ContainerMetric which has only two fields of different types.

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesPodMetric.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesPodMetric.java
@@ -16,26 +16,51 @@
 
 package com.netflix.spinnaker.clouddriver.kubernetes.v2.description;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import java.util.ArrayList;
-import java.util.List;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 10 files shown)


clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/KubernetesPodMetric.java
```diff
@@ -16,26 +16,51 @@
...
+  private final ImmutableList<ContainerMetric> containerMetrics;
 
-  @Data
   @Builder
-  @NoArgsConstructor
-  @AllArgsConstructor
+  @ParametersAreNullableByDefault
+  public KubernetesPodMetric(
+      String podName, String namespace, Iterable<ContainerMetric> containerMetrics) {
+    this.podName = Strings.nullToEmpty(podName);
```
maggieneterval: I had never seen `nullToEmpty` before; how great! ⛔️ ➡️ 🔤

ezimanyi: Yeah, I just learned about it too!  There are a lot of useful things hiding in guava!

PR #4473, opened 2020-03-26
chore(gha): don't run the GHA build when pushing to a fork

plumpy:
These builds will always fail because the secrets don't exist there, but
even if they did exist, we wouldn't want to publish a release from a fork.

Sending this to make sure it looks okay before I put it in 16 other
repositories.


plumpy:
@Mergifyio refresh

PR #4486, opened 2020-04-02
chore(java11): Compile with Java 11 (but targeting Java 8)

plumpy:
Additional differences:

1. The `Dockerfile.compile` is using Ubuntu instead of Alpine. Gradle downloads a `protoc` from Maven that is built against glibc. Alpine uses musl libc, so the downloaded `protoc` won't run. We need a glibc-based distro.

    For consistency, I may backport this change to the other five services I've already updated. I can't decide :/

2. Add `--no-daemon` to the `Dockerfile.compile` for consistency and because otherwise setting `-Xmx` in `GRADLE_OPTS` doesn't do anything useful.

3. `ServerGroupManagerController.java` had to be changed a little because of some tweaks to the type inference in Java 11. I honestly don't understand why the old code wouldn't compile, but this new code is better anyway.

4. `ReconcileClassicLinkSecurityGroupsAgentSpec` had to be changed because `Instant.now()` under Java 11 returns nanosecond precision. This test converts back and forth to `Date` objects, which can't store that level of precision.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,9 +13,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      # Install Java 8 for cross-compilation support. Setting it up before
+      # Java 11 means it comes later in $PATH (because of how setup-java works)
       - uses: actions/setup-java@v1
         with:
           java-version: 8
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 11
       - uses: actions/cache@v1.1.2
         with:
           path: ~/.gradle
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -13,9 +13,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      # Install Java 8 for cross-compilation support. Setting it up before
+      # Java 11 means it comes later in $PATH (because of how setup-java works)
```
jonsie, line 17: Is there a way to ensure the order rather than relying on the implicit behavior of setup-java?

plumpy: I could set `PATH`, but the problem is that GHA installs Java to something version-specific, like `/java/jdk-11-1.11.242.8` or whatever. So it's hard to set it properly. I agree this solution isn't great, but it's only for a few months until we start compiling to Java 11 properly and then I can remove the Java 8 JDK... :confused:

jonsie: Sounds good.  Just figured I'd ask.  I'm OK with how this is for now.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupManagerController.java
```diff
@@ -38,13 +38,14 @@
 @RestController
 @RequestMapping("/applications/{application}/serverGroupManagers")
 public class ServerGroupManagerController {
-  final List<ServerGroupManagerProvider> serverGroupManagerProviders;
+  final List<ServerGroupManagerProvider<ServerGroupManager>> serverGroupManagerProviders;
```
ezimanyi, line 41: Should this instance variable (as well as the constructor parameter) be `List<ServerGroupManagerProvider<? extends ServerGroupManager>>`?

plumpy: Sure, I'm not sure it's adds much, since it's just a private field and we don't really care about any other types besides `ServerGroupManager`, but I guess it doesn't hurt...

PR #4489, opened 2020-04-03
refactor(kubernetes): Clean up KubernetesManifestStrategy

ezimanyi:
We have a class KubernetesManifestStrategy to represent how to deploy a manifest, but it doesn't encapsulate itself well and pushes a lot of logic into callers.

Make the following changes:
* Move the logic for creating this KubernetesManifestStrategy object from annotations into the KubernetesManifestStrategy class; this allows us to move the strings representing the annotations into this class as well, increasing encapsulation.
* Improve the type safety of the member variables so that callers don't need to perform redundant null-unsafe checks. In particular:
  * Add a DeployStrategy enum to replace (no pun intended) the replace and recreate Booleans; these are mutually exclusive as setting both to true will 'kubectl delete' then 'kubectl replace' the manifest which always fails as there's nothing to replace.
  * Change versioned to an enum; we do care if it's set or not because if it's unset we use the value on the deploy operation. I guess this could have just stayed a Boolean, but it seemed clearer to make this an enum with three values (TRUE, FALSE, DEFAULT) though I could be convinced otherwise, I was on the fence here.
  * Make maxVersionHistory an OptionalInt instead of an Integer to get some null-safety in callers
  * Make useSourceCapacity a boolean instead of a Boolean as we always consider null the same as false.
* Make all the member variables private and final, exposing only getters of these new, more convenient return types. This simplifies a lot of the logic in callers.

A few notes:
* The new logic to read annotations avoids the complex logic that was in getAnnotations in KubernetesManifestAnnotater, which used an ObjectMapper instead of parseBoolean or parseInt. ObjectMapper is actually *more* strict here for booleans and would only return true if the value was exactly (case-sensitive) "true"; there is a slight change here that now we'll accept any casing of "true" (ex: "trUe", "TRUE") but I think that's sensible and probably what people would expect anyway.
* A couple of the annotation strings are shared between classes in the same package so I made them package-private instead of private.

Overall this ended up making KubernetesManifestStrategy larger than I might have intended at the beginning but I think this helps a lot with encapsulation, in that consumers need much less of an understanding of its internal workings.

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestAnnotater.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestAnnotater.java
@@ -27,12 +27,13 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.OptionalInt;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 
 @Slf4j
 public class KubernetesManifestAnnotater {
-  private static final String SPINNAKER_ANNOTATION = "spinnaker.io";
+  static final String SPINNAKER_ANNOTATION = "spinnaker.io";
   private static final String TRAFFIC_ANNOTATION_PREFIX = "traffic." + SPINNAKER_ANNOTATION;
   private static final String ARTIFACT_ANNOTATION_PREFIX = "artifact." + SPINNAKER_ANNOTATION;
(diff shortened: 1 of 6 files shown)


ezimanyi:
Somehow this refactor didn't feel as satisfying as some other, possibly because I ended up having to add a lot of code to `KubernetesManifestStrategy`...but I think this at least makes it have a more reasonable public interface and makes it easier in the future to change the implementation of that class if we want.  This should also make it easier to make the couple of upcoming changes to strategies.

maggieneterval:
> Somehow this refactor didn't feel as satisfying as some other, possibly because I ended up having to add a lot of code to KubernetesManifestStrategy...

To that I say:➕ 🔢 ➡️ 📉 🌪
(I think adding the code actually decreased the complexity)

PR #4490, opened 2020-04-06
chore(spotless): reformat code

plumpy:


plumpy:
Perhaps another option is to disable the logic from spinnaker/spinnaker-gradle-project#87 if the `GITHUB_WORKFLOW` environment variable is defined?

PR #4495, opened 2020-04-07
refactor(kubernetes): Clean up runAndRecordMetrics

ezimanyi:
* test(kubernetes): Add tests to metric recording on kubectl calls

  The runAndRecordMetrics funcgtion in KubernetesV2Credentials is very complex and has no tests. Before simplifying it, add some tests to document and validate the current behavior.

  This commit also updates the constructor of KubernetesManifest to set the metadata to an empty map. In real code, we always create a KubernetesManifest by deserializing, so the metadata is set to a map of whatever was on the incoming manifest. But it's often useful in tests to be able to create a manifest and set a few fields on it; before this change, this wasn't possible because metadata was always null. There were functions to set various keys on the metadata, but these would all fail due to the null metadata.

* refactor(kubernetes): Clean up runAndRecordMetrics

  The control flow in runAndRecordMetric is really confusing, in particular in that we re-throw and return from a finally block. Clean this up so that it uses better style.

* fix(kubernetes): Fix some issues with kubectl metrics

  Cleaning up the code in the last commit surfaced a few issues that this commit cleans up.

  (1) We should add a failure tag if a KubectlException is thrown
  (2) Metric tags are intended to be short descriptors, so we shouldn't add the text of an error message. Remove that message from the cause tag and just include the class of the exception.
  (3) There's no reason to separately tag 'kind' and 'kinds' depending if there are more than 1 or not; just always add 'kinds' with the full list of kinds. Also sort it for consistency in the metric tags.

  There's probably a bit we can do to futher clean up the exception handling
  (deciding when to wrap the exception, etc.) and deduplicate the two catch blocks but leaving that s is for now.

### DIFF
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
@@ -627,7 +627,7 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
     } catch (Exception e) {
       tags.put("success", "false");
       tags.put("reason", e.getClass().getSimpleName());
-      throw new KubectlJobExecutor.KubectlException(
+      throw new KubectlException(
           "Failure running " + action + " on " + kinds + ": " + e.getMessage(), e);
     } finally {
       registry
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
@@ -611,12 +611,9 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
       String action, List<KubernetesKind> kinds, String namespace, Supplier<T> op) {
     Map<String, String> tags = new HashMap<>();
     tags.put("action", action);
-    if (kinds.size() == 1) {
-      tags.put("kind", kinds.get(0).toString());
-    } else {
-      tags.put(
-          "kinds", kinds.stream().map(KubernetesKind::toString).collect(Collectors.joining(",")));
-    }
+    tags.put(
+        "kinds",
+        kinds.stream().map(KubernetesKind::toString).sorted().collect(Collectors.joining(",")));
     tags.put("account", accountName);
     tags.put("namespace", StringUtils.isEmpty(namespace) ? "none" : namespace);
     tags.put("success", "true");
@@ -624,11 +621,12 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
     try {
       return op.get();
     } catch (KubectlException e) {
-      // TODO(ezimanyi): We should be setting success to false here; this does not feel like success
+      tags.put("success", "false");
+      tags.put("reason", e.getClass().getSimpleName());
       throw e;
     } catch (Exception e) {
       tags.put("success", "false");
-      tags.put("reason", e.getClass().getSimpleName() + ": " + e.getMessage());
+      tags.put("reason", e.getClass().getSimpleName());
       throw new KubectlJobExecutor.KubectlException(
           "Failure running " + action + " on " + kinds + ": " + e.getMessage(), e);
     } finally {
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
@@ -609,46 +609,32 @@ public class KubernetesV2Credentials implements KubernetesCredentials {
 
   private <T> T runAndRecordMetrics(
       String action, List<KubernetesKind> kinds, String namespace, Supplier<T> op) {
-    T result = null;
-    Throwable failure = null;
-    KubectlException apiException = null;
+    Map<String, String> tags = new HashMap<>();
+    tags.put("action", action);
+    if (kinds.size() == 1) {
+      tags.put("kind", kinds.get(0).toString());
+    } else {
+      tags.put(
+          "kinds", kinds.stream().map(KubernetesKind::toString).collect(Collectors.joining(",")));
+    }
+    tags.put("account", accountName);
+    tags.put("namespace", StringUtils.isEmpty(namespace) ? "none" : namespace);
+    tags.put("success", "true");
     long startTime = clock.monotonicTime();
     try {
-      result = op.get();
+      return op.get();
     } catch (KubectlException e) {
-      apiException = e;
+      // TODO(ezimanyi): We should be setting success to false here; this does not feel like success
+      throw e;
     } catch (Exception e) {
-      failure = e;
+      tags.put("success", "false");
+      tags.put("reason", e.getClass().getSimpleName() + ": " + e.getMessage());
+      throw new KubectlJobExecutor.KubectlException(
+          "Failure running " + action + " on " + kinds + ": " + e.getMessage(), e);
     } finally {
-      Map<String, String> tags = new HashMap<>();
-      tags.put("action", action);
-      if (kinds.size() == 1) {
-        tags.put("kind", kinds.get(0).toString());
-      } else {
-        tags.put(
-            "kinds", kinds.stream().map(KubernetesKind::toString).collect(Collectors.joining(",")));
-      }
-      tags.put("account", accountName);
-      tags.put("namespace", StringUtils.isEmpty(namespace) ? "none" : namespace);
-      if (failure == null) {
-        tags.put("success", "true");
-      } else {
-        tags.put("success", "false");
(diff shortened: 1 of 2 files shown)


clouddriver-kubernetes-v2/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2CredentialsTest.java
```diff
@@ -0,0 +1,376 @@
...
+            new KubernetesSpinnakerKindMap(ImmutableList.of()));
+    ManagedAccount managedAccount = new ManagedAccount();
+    managedAccount.setName("my-account");
+    managedAccount.setProviderVersion(ProviderVersion.v2);
+    return factory.build(managedAccount);
+  }
+
+  private KubernetesManifest getManifest() {
+    KubernetesManifest manifest = new KubernetesManifest();
+    manifest.put("metadata", new HashMap<>());
```
maggieneterval, line 81: > This commit also updates the constructor of KubernetesManifest to
set the metadata to an empty map.

Did you end up deciding not to do this and just manually set it to an empty map here?

clouddriver-kubernetes-v2/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2CredentialsTest.java
```diff
@@ -0,0 +1,376 @@
...
+  }
+
+  @Test
+  void propagatedExceptionForListThrowingOtherException() {
+    KubectlJobExecutor jobExecutor = mock(KubectlJobExecutor.class);
+    Registry registry = new DefaultRegistry();
+    KubernetesV2Credentials credentials = getCredentials(registry, jobExecutor);
+
+    Exception cause = new CustomException("Kubernetes error");
+    when(jobExecutor.list(eq(credentials), any(), any(), any())).thenThrow(cause);
```
maggieneterval, line 354: Probably doesn't matter at all, but is there any reason we have `any()` for the kinds and namespace parameters in these tests when we're already passing them to the credentials call below?

clouddriver-kubernetes-v2/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2CredentialsTest.java
```diff
@@ -0,0 +1,376 @@
...
+
+  // This is an error type that will only ever be thrown by stubs in this test; that way we can
+  // assert that it is thrown and be sure that we aren't accidentally passing due to an unrelated
+  // exception.
+  private static class CustomException extends RuntimeException {
+    CustomException(String message) {
+      super(message);
+    }
+  }
+}
```
maggieneterval, line 374: This is probably the most thorough set of tests I have ever encountered in Spinnaker ⛵️ 💯

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -609,46 +609,32 @@ public void patch(
...
+    tags.put("account", accountName);
+    tags.put("namespace", StringUtils.isEmpty(namespace) ? "none" : namespace);
+    tags.put("success", "true");
     long startTime = clock.monotonicTime();
     try {
-      result = op.get();
+      return op.get();
     } catch (KubectlException e) {
-      apiException = e;
+      // TODO(ezimanyi): We should be setting success to false here; this does not feel like success
```
maggieneterval: 😂

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -611,24 +611,22 @@ public void patch(
...
     } catch (KubectlException e) {
-      // TODO(ezimanyi): We should be setting success to false here; this does not feel like success
+      tags.put("success", "false");
+      tags.put("reason", e.getClass().getSimpleName());
       throw e;
     } catch (Exception e) {
       tags.put("success", "false");
-      tags.put("reason", e.getClass().getSimpleName() + ": " + e.getMessage());
+      tags.put("reason", e.getClass().getSimpleName());
       throw new KubectlJobExecutor.KubectlException(
```
maggieneterval, line 630:
```suggestion
      throw new KubectlException(
```

...or is this different than the un-prefixed `KubectlException` caught above?

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
```diff
@@ -611,24 +611,22 @@ public void patch(
...
           "Failure running " + action + " on " + kinds + ": " + e.getMessage(), e);
```
maggieneterval, line 631: Why do we only wrap non-KubectlException exceptions' messages in this custom "Failure running" message and just re-throw KubectlExceptions with their message as-is?

clouddriver-kubernetes-v2/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2CredentialsTest.java
ezimanyi: Yes, good catch!  It turns out setting it in the constructor broke deserialization so I removed that change (but not the comment).

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
ezimanyi: Oh good call, given that we've imported it already, there's no need to qualify it here.

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2Credentials.java
ezimanyi: To be honest, I also find this a bit confusing and am not 100% sure what would be best here.

I *think* the intent is to ensure that all exceptions leaving this function are a `KubectlException` to make it easier to catch.  On the one hand, that's good as it means we're throwing an exception appropriate to the level of abstraction, so callers will see that a call to `kubectl` (and which call) failed rather than some low-level exception.

I think the presumption is probably that if we already have a `KubectlException` it already has information at the right level of abstraction (and it does largely seem to as it usually contains information about the `kubectl` call that failed).  But it does feel like we should potentially be cleaner about where this exception is thrown and maybe:
* Throw a `KubectlException` in `KubectlJobExecutor` with details about the `kubectl` call that failed
* Wrap *any* exception in a `KubernetesAPIException` in this method providing details on the api operation that failed

I think this will probably be necessary for consistency when we work on using a client library, as `KubectlException` will be an implementation detail of the `kubectl` exception, which is why I didn't overly refactor it here.  In my follow-up PR to make `replace` an upsert, I will slightly improve this, but I think that the real improvement will come when we move to a client library.  (And I definitely want to talk with you in detail about what the exceptions should be as part of that work.)

maggieneterval: 👍 👍 Thank you for the paragraphs. Even more excited for the client library migration now!

clouddriver-kubernetes-v2/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/security/KubernetesV2CredentialsTest.java
ezimanyi: My intent here was to avoid over-specifying the tests and focus on having each test verify the behavior it's testing.  In this case the test is ensuring that an exception thrown by `jobExecutor` gets propagated properly out of the `credentials.list` call, so it seemed best to just have `jobExecutor.list` always throw that exception rather than have it coupled to exactly the arguments that are passed.  (Then, for example, if we changed how `credentials.list` passed things to `jobExecutor.list` the test on exception propagation would break even though exception propagation hasn't actually broken.)

Ideally there would *also* be a test that `credentials.list` properly forwards calls to `kubectl`...in that case we probably would want to assert that the right arguments are forwarded onwards to `jobExecutor.list`, to make that test break if we stopped passing the right arguments.  So you're entirely correct, that these tests don't in any way test that `credentials.list` or `credentials.deploy` do the right thing, but I think that would be belong in other tests (that one of us will probably write one day 😀 )

maggieneterval: 💯 Makes sense; the rule I am deriving from this is to use `any()` instead of explicit argument matchers unless 1) the goal of the test is to assert that the method was called with particular arguments, or 2) multiple `when`s of the same method need to be disambiguated by argument

PR #4501, opened 2020-04-09
refactor(kubernetes): Don't look up default namespace

ezimanyi:
* refactor(kubernetes): Convert DeployManifestOperation test to java

  The next refactor will require some changes to this test that are more difficult than necessary because the test is in groovy and uses more mocking than necessary. Convert the test to Java, and clean up the mocking so it's easier to follow.

* refactor(kubernetes): Operations should return the created manifest

  When creating a manifest, the result is returned by the API, but is currently ignored in most cases by clouddriver. It is useful to use the returned manifest because there are times where it has updated information. One particular example is when we don't specify a namespace on our manifest; in that case, the manifest will be applied to the default namespace that is configured.

  Currently we solve this problem by separately looking up the default namespace before applying the manifest, but it would be easier to just apply it and then check what kubernetes default it to. This change will allow us to do that in a future commit.

* refactor(kubernetes): Don't look up default namespace

  There are two places where we look up the default namespace unnecessarily. The first is on every call to kubectl so that we can pass in a "--namespace" flag; if we just leave that flag out, we'll get the default namespace already. The second is when deploying a manifest without a namespace; in that case, we look up the default namespace so that we can apply it to the manifest before deploying. Similarly, if we just deploy the manifest as is, it will get the default namespace (which we'll see on the result from the API).

  I tested that this works both for the serviceAccount case and for the regular kubeconfig case (testing the serviceAccount by ssh-ing to a pod set up this way to check that kubectl already detected the default).

  This deletes a lot of unecessary code. When we move to use the client library, we may need to solve this again as I'm not sure that it will always detect a default namespace, but this code was complicated enough and coupled pretty tightly to the kubectl implementation that I think it will be better to start from scratch at that point.  (This was also broken, as the command to get the default namespace from the kubeconfig appears to fail every time so you always get 'default' regardless of what is set.)


clouddriver-kubernetes-v2/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/KubernetesDeployManifestOperationTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+  @Test
+  void sendsTrafficWhenEnabledTrafficTrue() {
+    KubernetesDeployManifestDescription description =
+        baseDeployDescription("deploy/replicaset.yml")
+            .setServices(ImmutableList.of("service my-service"))
+            .setEnableTraffic(true);
+    OperationResult result = deploy(description);
+
+    KubernetesManifest manifest = Iterables.getOnlyElement(result.getManifests());
```
maggieneterval, line 114: `Iterables.getOnlyElement` -- this is news to me, love it!

PR #4508, opened 2020-04-13
fix(aws/findimage): Truncate results list after filters have been applied

german-muzquiz:
This would partially fix https://github.com/spinnaker/spinnaker/issues/3355

Currently there's a hard limit of 1000 results read from `namedImages` cache, which causes the Find Image from Tags have erratic behavior when the supplied package name matches more than 1000 results, across all accounts, regions and including all public AMI's.

This change moves the limit after all supplied filters have been applied, particularly tags. The idea is that when orca receives the results and see that there are 1000 matches, display a warning message in the Find Images from Tags stage indicating users that their search criteria matched too many results and is possible that the stage is not behaving in the way they want, prompting them to refine their search criteria.

Orca PR: https://github.com/spinnaker/orca/pull/3594


german-muzquiz:
cc @ajordens

german-muzquiz:
cc @robzienert

cfieber:
@emjburns / @robfletcher PTAL (for the overall feature, but also in light of how managed delivery artifact converging happens - I'm not sure what image lookups happen there)

ajordens:
I suspect the earlier filter application was a performance optimization to avoid fetching some overly large number of objects out of the cache?

I'll let someone with a little more context chime in tho.

robzienert:
Is there any feasibility of bubbling the error from clouddriver to orca instead of passing a payload of 1000 images?

german-muzquiz:
For bubbling up the error, another option is just to fail the request with 4xx error if we see the list will be truncated, that will fail the stage.

Regarding performance impacts: I did a small test looking for package name `all` without any other filter and these were the results:

Before the change:
image

After the change:
image


I think that the fact that the total time spent decreased just means that there wasn't a significant difference between before and after, as every time a request is sent the response time varies slightly. I don't have data of the number of objects being created/destroyed, which certainly affect CPU usage.

On the other hand, having SQL as a backend seems to allow for searches inside the json body of a column, we could improve that implementation.

robzienert:
I was speaking more about network performance - not clouddriver resource utilization performance. I don't see value in passing around large payloads when we know that Orca will just fail the stage anyway.

german-muzquiz:
Yeah we can definitely reduce the 1000 limit. How about a value of 50? 

For the `Find Image by Tags` stage, filtering by region and sorting by creation date to find the "newest" image is done in orca, while filtering by package name and tags is done in clouddriver.

german-muzquiz:
@robzienert is there a prefered way to proceed with this? I agree that the best solution would be to change the API to include an envelope with metadata information like the total number of results, I just wanted to avoid impacting the API. I'm open to do that or to reduce the max results so we don't pass around large amounts of data with no real value.

german-muzquiz:
Friendly reminder for this PR

robzienert:
We can proceed as-is. May as well just get the improvement and look to iterate.

PR #4509, opened 2020-04-13
fix(kubernetes): Run job stage does not always add suffix

ezimanyi:
Fixes spinnaker/spinnaker#5599

* test(kubernetes): Fix class names in test

  I just noticed that a couple of test classes don't match their file name; fix these.

* fix(kubernetes): Run job stage does not always add suffix

  The Kubernetes run job stage currently always adds a random suffix to the job name that it's deploying. This is so that we can re-run the same job and not run into issues trying to replace a job.

  There is already functionality built into Kubernetes for handling this; by setting the generateName, Kubernetes will generate a random suffix and append it.

  This commit removes the logic to automatically add a suffix from all jobs from Spinnaker; users should
instead use a generateName if they want this functionality. This now allows users who do not want a random suffix to just use the name field and deploy the same job name every time. This *does* mean that upon upgrading to this version, users who want the same behavior as before will need to change their manifests
to use generateName instead of name.

  We also force the use of the strategy.spinnaker.io/recreate deploy strategy. This is because neither 'kubectl apply' nor 'kubectl replace' will have the desired effect for a run job stage. In both of these cases, if the job has not changed, the operation will be a successful no-op but the job won't re-run, which will be confusing. If the job has changed, the operation will fail because jobs are immutable. By forcing use of the recreate strategy we'll ensure that a new job is created each time the stage is run, and that there will be a new execution of the job each time the stage is run.


clouddriver-kubernetes-v2/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestStrategyTest.java
```diff
@@ -288,6 +291,35 @@ void useSourceCapacityToAnnotations() {
         .containsOnly(entry("strategy.spinnaker.io/use-source-capacity", "true"));
   }
 
+  @ParameterizedTest
+  @EnumSource(DeployStrategy.class)
```
maggieneterval, line 295: Wow this is amazing, I can't wait to use this!

clouddriver-kubernetes-v2/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubernetesRunJobOperationTest.java
```diff
@@ -0,0 +1,169 @@
...
+    runJobDescription.setCredentials(getNamedAccountCredentials());
+    return runJobDescription;
+  }
+
+  private static KubernetesNamedAccountCredentials<KubernetesV2Credentials>
+      getNamedAccountCredentials() {
+    KubernetesConfigurationProperties.ManagedAccount managedAccount =
+        new KubernetesConfigurationProperties.ManagedAccount();
+    managedAccount.setName("my-account");
+    managedAccount.setProviderVersion(ProviderVersion.v2);
```
maggieneterval, line 114: I think v2 is the [default](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.java#L32) now so you might not need this line!

clouddriver-kubernetes-v2/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/description/manifest/KubernetesManifestStrategyTest.java
ezimanyi: I've known about this for 90 minutes and already used it 3 times!

clouddriver-kubernetes-v2/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/op/job/KubernetesRunJobOperationTest.java
ezimanyi: You are entirely correct, removed it! 🎉

PR #4522, opened 2020-04-15
fix(titus): Fetch ipv4 and ipv6 from titus api

aravindmd:
Today we use `agentId` to infer the `instanceId` and make an additional call to `/instance` endpoint to get the `ipv6Address` from the `networkInterface`

This change is to get the `ipv4` and `ipv6` addresses from Titus API directly . `ipv4` address is transitioning from `titus.containerIp` to `task.containerIPv4`

Todo : once these changes are merged and things are tested , we can remove the agentId and containerIp fields

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/model/TitusInstance.groovy
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/model/TitusInstance.groovy
@@ -108,11 +108,11 @@ class TitusInstance implements Instance {
   }
 
   String getIpv4Address() {
-    ipv4Address
+    return ipv4Address
   }
 
   String getIpv6Address() {
-    ipv6Address
+    return ipv6Address
   }
 
   String getHostIp() {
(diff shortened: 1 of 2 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/model/TitusInstance.groovy
```diff
@@ -102,6 +107,14 @@ class TitusInstance implements Instance {
     agentId
   }
 
+  String getIpv4Address() {
+    ipv4Address
+  }
+
+  String getIpv6Address() {
+    ipv6Address
```
aravindmd, line 115: the earlier issue was due to this getter not being renamed with intellij refactor + groovy !

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/model/TitusInstance.groovy
```diff
@@ -102,6 +107,14 @@ class TitusInstance implements Instance {
     agentId
   }
 
+  String getIpv4Address() {
+    ipv4Address
```
marchello2000, line 111: i personally hate the groovy return-less syntax, but for simple getters i am going to say it's just a nit, up to you

aravindmd: yep agree , added the return statement

aravindmd:
not sure why i added the label and it ended up removing it , it was not showing the label to begin with

PR #4523, opened 2020-04-15
feat(aws): Adding support for caching ec2 launch templates

jeyrschabu:
- Added new Launch template caching agent (feature flagged)
- Updated cluster provider & agent to handle server groups with templates

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/config/AwsProviderConfig.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/config/AwsProviderConfig.groovy
@@ -25,6 +25,7 @@ import com.netflix.spinnaker.clouddriver.aws.AmazonCloudProvider
 import com.netflix.spinnaker.clouddriver.aws.provider.agent.AmazonApplicationLoadBalancerCachingAgent
 import com.netflix.spinnaker.clouddriver.aws.provider.agent.AmazonCertificateCachingAgent
 import com.netflix.spinnaker.clouddriver.aws.provider.agent.AmazonCloudFormationCachingAgent
+import com.netflix.spinnaker.clouddriver.aws.provider.agent.AmazonLaunchTemplateCachingAgent
 import com.netflix.spinnaker.clouddriver.aws.provider.agent.AmazonLoadBalancerCachingAgent
 
 import com.netflix.spinnaker.clouddriver.aws.provider.agent.ReservedInstancesCachingAgent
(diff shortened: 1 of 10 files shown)


jeyrschabu:
This PR introduces the indexing of ec2 launch templates as well as ensuring server groups are augmented with the correct expanded launch template version.
The `aws.features.launch-templates` config flag is used to enable/disable to the lauch template caching agent.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/config/AwsProviderConfig.groovy
```diff
@@ -154,6 +155,10 @@ class AwsProviderConfig {
               amazonClientProvider, credentials, region.name, objectMapper, ctx
             )
           }
+
+          if (dynamicConfigService.isEnabled("aws.features.launch-templates", false)) {
```
cfieber, line 159: I'm not sure we have remote config wired in early enough that this will actually pick up a config change and add the agent (will still read the value from the `aws.featues.launch-templates.enabled` value in the spring environment, so probably worth having here in case we get earlier remote config, or for those that have some other remote config story)

robzienert: The FP property source is weaved in via an `EnvironmentPostProcessor` so its data is loaded plenty early.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonClusterProvider.groovy
```diff
@@ -84,18 +84,29 @@ class AmazonClusterProvider implements ClusterProvider<AmazonCluster>, ServerGro
...
+    def serverGroup = new AmazonServerGroup(serverGroupData.attributes)
+    def serverGroupById = [(serverGroupData.id): new AmazonServerGroup(serverGroupData.attributes)]
+
+    String imageId
 
-    String launchConfigKey = Keys.getLaunchConfigKey(serverGroupData?.attributes['launchConfigName'] as String, account, region)
-    CacheData launchConfigs = cacheView.get(LAUNCH_CONFIGS.ns, launchConfigKey)
+    Map<String, Object> launchTemplateSpec = asg["launchTemplate"] as Map
+    String launchTemplateName = launchTemplateSpec?.get('launchTemplateName')
+    if (launchTemplateName) {
```
robzienert, line 94: Recommend avoiding Groovy truth.

PR #4526, opened 2020-04-16
feat(ecs): Return application load balancer info from LB provider

allisaurus:
Implements https://github.com/spinnaker/spinnaker/issues/4797

Provides application load balancer and target group information for ECS server groups.

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/EcsLoadbalancerCacheClient.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/EcsLoadbalancerCacheClient.java
@@ -25,12 +25,8 @@ import com.netflix.spinnaker.cats.cache.CacheData;
 import com.netflix.spinnaker.cats.cache.RelationshipCacheFilter;
 import com.netflix.spinnaker.clouddriver.aws.data.Keys;
 import com.netflix.spinnaker.clouddriver.ecs.cache.model.EcsLoadBalancerCache;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import com.netflix.spinnaker.clouddriver.ecs.provider.view.EcsAccountMapper;
+import java.util.*;
 import java.util.stream.Collectors;
 import org.springframework.stereotype.Component;
 
@@ -39,10 +35,13 @@ public class EcsLoadbalancerCacheClient {
 
   private final Cache cacheView;
   private final ObjectMapper objectMapper;
+  private final EcsAccountMapper ecsAccountMapper;
 
-  public EcsLoadbalancerCacheClient(Cache cacheView, ObjectMapper objectMapper) {
+  public EcsLoadbalancerCacheClient(
+      Cache cacheView, ObjectMapper objectMapper, EcsAccountMapper ecsAccountMapper) {
     this.cacheView = cacheView;
     this.objectMapper = objectMapper;
+    this.ecsAccountMapper = ecsAccountMapper;
   }
 
   public List<EcsLoadBalancerCache> find(String account, String region) {
@@ -56,6 +55,12 @@ public class EcsLoadbalancerCacheClient {
 
   private Set<Map<String, Object>> fetchFromCache(String account, String region) {
     String accountFilter = account != null ? account : "*";
+    if (!"*".equals(accountFilter)) {
+      String awsAccountName = ecsAccountMapper.fromEcsAccountNameToAwsAccountName(accountFilter);
+      if (awsAccountName != null) {
+        accountFilter = awsAccountName;
+      }
+    }
     String regionFilter = region != null ? region : "*";
 
     String searchKey = Keys.getLoadBalancerKey("*", accountFilter, regionFilter, "*", "*") + "*";
@@ -66,12 +71,16 @@ public class EcsLoadbalancerCacheClient {
     return fetchLoadBalancerAttributes(loadbalancerKeys);
   }
 
-  public Set<EcsLoadBalancerCache> findWithTargetGroups(Set<String> targetGroups) {
-    return findAll().stream()
-        .filter(
-            ecsLoadBalancerCache ->
-                targetGroups.containsAll(ecsLoadBalancerCache.getTargetGroups()))
-        .collect(Collectors.toSet());
+  public List<EcsLoadBalancerCache> findWithTargetGroups(Set<String> targetGroupKeys) {
+    Set<CacheData> targetGroupCacheData =
+        new HashSet<>(
+            cacheView.getAll(
+                TARGET_GROUPS.getNs(),
+                targetGroupKeys,
+                RelationshipCacheFilter.include(LOAD_BALANCERS.getNs())));
+    Set<String> lbKeys = inferAssociatedLoadBalancers(targetGroupCacheData);
+    Set<Map<String, Object>> loadbalancerAttributes = fetchLoadBalancerAttributes(lbKeys);
+    return convertToLoadbalancer(loadbalancerAttributes);
   }
 
   private EcsLoadBalancerCache convertToLoadBalancer(Map<String, Object> targetGroupAttributes) {
@@ -89,8 +98,9 @@ public class EcsLoadbalancerCacheClient {
     return ecsTargetGroups;
   }
 
-  private Set<Map<String, Object>> fetchLoadBalancerAttributes(Collection<String> targetGroupKeys) {
-    Set<CacheData> loadBalancerCache = fetchLoadBalancers(targetGroupKeys);
+  private Set<Map<String, Object>> fetchLoadBalancerAttributes(
+      Collection<String> loadBalancerKeys) {
+    Set<CacheData> loadBalancerCache = fetchLoadBalancers(loadBalancerKeys);
 
     return loadBalancerCache.stream()
         .filter(this::hashTargetGroups)
@@ -108,7 +118,8 @@ public class EcsLoadbalancerCacheClient {
     Map<String, String> parts = Keys.parse(loadbalancerCache.getId());
 
     attributes.put("region", parts.get("region"));
-    attributes.put("account", parts.get("account"));
+    String ecsAccount = ecsAccountMapper.fromAwsAccountNameToEcsAccountName(parts.get("account"));
+    attributes.put("account", ecsAccount);
     attributes.put("loadBalancerType", parts.get("loadBalancerType"));
     attributes.put(
         "targetGroups",
@@ -119,18 +130,12 @@ public class EcsLoadbalancerCacheClient {
     return attributes;
   }
 
-  private Set<Map<String, Object>> retrieveLoadbalancers(
-      Set<String> loadbalancersAssociatedWithTargetGroups) {
-    Collection<CacheData> loadbalancers =
-        cacheView.getAll(LOAD_BALANCERS.getNs(), loadbalancersAssociatedWithTargetGroups);
-    return loadbalancers.stream().map(CacheData::getAttributes).collect(Collectors.toSet());
-  }
-
   private Set<String> inferAssociatedLoadBalancers(Set<CacheData> targetGroups) {
     Set<String> loadbalancersAssociatedWithTargetGroups = new HashSet<>();
 
     for (CacheData targetGroup : targetGroups) {
-      Collection<String> relatedLoadbalancer = targetGroup.getRelationships().get("loadbalancer");
+      Collection<String> relatedLoadbalancer =
+          targetGroup.getRelationships().get(LOAD_BALANCERS.ns);
       if (relatedLoadbalancer != null && relatedLoadbalancer.size() > 0) {
         loadbalancersAssociatedWithTargetGroups.addAll(relatedLoadbalancer);
       }
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/model/loadbalancer/EcsLoadBalancer.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/model/loadbalancer/EcsLoadBalancer.java
@@ -1,31 +1,32 @@
 /*
- * Copyright 2017 Lookout, Inc.
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
  *
- *    http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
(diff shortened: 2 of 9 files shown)


allisaurus:
related: https://github.com/spinnaker/deck/pull/8186

allisaurus:
CC @clareliguori  + @pkandasamy91 for review

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/EcsLoadbalancerCacheClient.java
```diff
@@ -56,6 +55,12 @@ public EcsLoadbalancerCacheClient(Cache cacheView, ObjectMapper objectMapper) {
 
   private Set<Map<String, Object>> fetchFromCache(String account, String region) {
     String accountFilter = account != null ? account : "*";
+    if (!accountFilter.equals("*")) {
```
pkandasamy91, line 58: nit: reverse the check "*".equals(accountFilter). If the logic for accountFilter changes in the future and it's null, the current implementation would throw an NPE.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/model/EcsLoadBalancerCache.java
```diff
@@ -54,6 +54,6 @@ public String getName() {
 
   @Override
   public String getType() {
-    return loadBalancerType;
+    return cloudProvider;
```
pkandasamy91, line 57: Is there a reason why type returns the ecscloudprovider id? Shouldn't the getCloudProvider call be used instead?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/model/loadbalancer/EcsLoadBalancer.java
```diff
@@ -0,0 +1,58 @@
...
+import com.netflix.spinnaker.clouddriver.ecs.EcsCloudProvider;
+import com.netflix.spinnaker.clouddriver.model.LoadBalancer;
+import com.netflix.spinnaker.clouddriver.model.LoadBalancerServerGroup;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import lombok.Data;
+
+@Data
+public class EcsLoadBalancer implements LoadBalancer {
```
pkandasamy91, line 28: The EcsLoadBalancer and the EcsLoadBalancerCache are similar (scheme and targetGroupServices exist in one vs the other). In what scenarios will you use one over the other (EcsLoadBalancerCache vs EcsLoadBalancer)?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsLoadBalancerProvider.java
```diff
@@ -92,27 +106,116 @@ public String getCloudProvider() {
 
   @Override
   public Item get(String name) {
-    return null; // TODO - Implement this.
+    // unused method
+    throw new UnsupportedOperationException("Not implemented.");
```
pkandasamy91, line 110: Not sure if we want to throw an exception here. Previously, this would just return null but allow users to continue, where as this would now throw an exception and potetially break and implementations that use this.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsLoadBalancerProvider.java
```diff
@@ -92,27 +106,116 @@ public String getCloudProvider() {
...
   }
 
   @Override
   public List<Details> byAccountAndRegionAndName(String account, String region, String name) {
-    return null; // TODO - Implement this.  This is used to show the details view of a load balancer
-    // which is not even implemented yet
+    // unused method
+    throw new UnsupportedOperationException("Not implemented.");
```
pkandasamy91, line 116: Same comment as above.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsLoadBalancerProvider.java
```diff
@@ -92,27 +106,116 @@ public String getCloudProvider() {
...
+          targetGroupKeys.addAll(matchingKeys);
+          // associate target groups with services it contains targets for
+          if (targetGroupToServicesMap.containsKey(tgArn)) {
+            Set<String> serviceList = targetGroupToServicesMap.get(tgArn);
+            serviceList.add(service.getServiceName());
+            targetGroupToServicesMap.put(tgArn, serviceList);
+          } else {
+            Set<String> srcServices = new HashSet<>();
+            srcServices.add(service.getServiceName());
+            targetGroupToServicesMap.put(tgArn, srcServices);
```
pkandasamy91, line 165: Since this logic is very similar, it might make more sense to move the common logic out of the if-else?

clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsLoadBalancerProviderSpec.groovy
```diff
@@ -16,33 +16,53 @@
...
 import com.netflix.spinnaker.clouddriver.ecs.security.ECSCredentialsConfig
+import com.netflix.spinnaker.clouddriver.ecs.security.NetflixECSCredentials
 import spock.lang.Specification
 import spock.lang.Subject
 
 class EcsLoadBalancerProviderSpec extends Specification {
-  def client = Mock(EcsLoadbalancerCacheClient)
-  def ecsCredentialsConfig = Mock(ECSCredentialsConfig)
+  public static final String ECS_ACCOUNT = 'ecsAccount'
+  public static final String AWS_ACCOUNT = 'awsAccount'
```
pkandasamy91, line 35: nit: you might want to stick to using def or defining concrete types in both test classes (just for consistency). In EcsLoadBalancerCacheClientSpec you use def, where as here you use the concrete type.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/model/loadbalancer/EcsLoadBalancer.java
allisaurus: `EcsLoadBalancerCache` is the load balancer data we expect to be cached by clouddriver, while `EcsLoadBalancer` is the data we want to be returned (to deck, mostly) about ECS load balancers in an app. 

The latter should include additional things we need to populate the LB view, (e.g. targetGroupServices) and I expect it to diverge further over time as we add things like instance counts and task health which we want to be available to the load balancers view but isn't the responsibility of the caching agent to keep track of.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsLoadBalancerProvider.java
allisaurus: My concern with always returning null is the client will proceed like they have correct information vs. understanding that this method should not be used. But I understand the concern about breaking folks; I'll take another look at where this (and the other such method) may cause problems and add more info here.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/model/loadbalancer/EcsLoadBalancer.java
pkandasamy91: Got it. Would it make sense for one to extend the other or for them to be their own concrete implementations?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/EcsLoadbalancerCacheClient.java
```diff
@@ -25,12 +25,8 @@
...
 import com.netflix.spinnaker.clouddriver.aws.data.Keys;
 import com.netflix.spinnaker.clouddriver.ecs.cache.model.EcsLoadBalancerCache;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import com.netflix.spinnaker.clouddriver.ecs.provider.view.EcsAccountMapper;
+import java.util.*;
```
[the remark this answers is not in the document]
allisaurus: ah, there was [some discussion about this](https://github.com/spinnaker/clouddriver/commit/bcb2a16e191486a21666146e70e8eb65f55b7e6f) during development, but basically the default IntelliJ settings do this after 5 imports from the same lib. I'm inclined to leave it unless there's more updated guidance for the project re: changing it.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/model/EcsLoadBalancerCache.java
allisaurus: > Shouldn't the getCloudProvider call be used instead?

it should! `getType` was actually [deprecated](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/LoadBalancer.java#L52) in favor of `getCloudProvider`, however it's still a part of the the `LoadBalancer` interface which is why the method is still here.

Returning `cloudProvider` instead of `loadBalancerType` I think aligns it better to it's intended purpose (plus `loadBalancerType` should always be "application" in the case of ECS, vs. AWS LBs, which is where we're sourcing the cache data), however @clareliguori might know more.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/model/loadbalancer/EcsLoadBalancer.java
allisaurus: hmm you know it would probably be better to have `EcsLoadBalancer` extend `'cache`. At first I was thinking the data type of `serverGroups` might have to be different (maybe include add'l target health info) but it's part of the parent `LoadBalancer` interface so we wouldn't want to change it anyway. +1

allisaurus: per offline discussion: going to add a todo re: extending and look into refactoring `EcsLoadBalancerCache` in a separate PR.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsLoadBalancerProvider.java
allisaurus: actually I can make the `else` block more concise, that'll differentiate them

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsLoadBalancerProvider.java
allisaurus: Update: going back to returning `null` but adding more descriptive comments. [Other](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/provider/view/CloudFoundryLoadBalancerProvider.java#L61-L70) [providers](https://github.com/spinnaker/clouddriver/blob/29c5c1435ab0c750c4b491208f600647a3e98901/clouddriver-dcos/src/main/groovy/com/netflix/spinnaker/clouddriver/dcos/provider/view/DcosLoadBalancerProvider.groovy#L78-L89) [vary](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/provider/view/AppengineLoadBalancerProvider.groovy#L37-L51) between returning null and `OperationNotSupportedException`, indicating they are not required by deck, however hitting the [associated endpoints](https://github.com/spinnaker/gate/blob/master/gate-web/src/main/groovy/com/netflix/spinnaker/gate/controllers/LoadBalancerController.groovy#L39-L56) directly now results in a 500. Would rather leave existing behavior intact if both have precedence.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsLoadBalancerProvider.java
allisaurus: see [above](https://github.com/spinnaker/clouddriver/pull/4526#discussion_r412619961)

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/model/EcsLoadBalancerCache.java
pkandasamy91: The @Data annotation on the class should already provide the default setters and getters. In this case, I don't think you'll need to override any of the accessors (as long as you update the methods calling getType to call getCloudProvider.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsLoadBalancerProvider.java
pkandasamy91: I'm not opposed to returning the `OperationNotSupportedException`, but we may want to open an issue, and let the community know prior to making the change.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/model/EcsLoadBalancerCache.java
allisaurus: there currently isn't a field just called "type", so we need the override method regardless, and I would rather return the cloud provider ID since this is in line with how [most](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonLoadBalancer.groovy#L39) [other](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/model/AppengineLoadBalancer.groovy#L38) [providers](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/model/CloudFoundryLoadBalancer.java#L111) are currently setting this field.

pkandasamy91: As discussed offline, this makes sense in order to maintain consistency with the other providers. Since the interface only has a getName and getType method, it would potentially break in many places if we add logic to create a new getProvider method.

PR #4205, opened 2019-12-06
feat(auth): Add ResourcesNameable to BasicAmazonDeployDescription, default to no authorization

jonsie:
Additionally, corrected some terminology (authn -> authz) and removed a dangerous `requiresAuthentication` boolean from `DescriptionAuthorizer` that could potentially cause authz to skip on resources.

Authorizing these resources is no doubt backwards incompatible, and not something we can enable at the moment.  However, I would like to collect some data internally to see how often we deploy server groups with resources that fall outside the context of the application frigga name -- that will help inform us how we can go about rolling out a change that authorizes application resources.

PR #4533, opened 2020-04-21
fix(aws): Resolve security group names from list of IDs and names

jonsie:
In https://github.com/spinnaker/clouddriver/pull/4205 I didn't account for the fact that the `securityGroups` list is a mixed list of IDs and names.

This is a bit of a refactor in that I pulled out the useful logic from `DefaultLaunchConfigurationBuilder` and shifted it to `SecurityGroupService`.


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/services/SecurityGroupService.groovy
```diff
@@ -127,4 +130,35 @@ class SecurityGroupService {
...
+      names.removeAll(ids)
+      if (names) {
+        def resolvedIds = nameResolver.call(names.toList())
+        ids.addAll(resolvedIds.values())
+      }
+      return ids.toList()
+    } else {
+      return []
+    }
+  }
```
jonsie, line 163: There's a bit of duplication here since these methods are so similar, but it felt easier to reason about since the fact that SGs are a mix of IDs and names is already somewhat tricky.

PR #4537, opened 2020-04-22
fix(aws): Resolve BasicAmazonDeployDescription NPE when returning names of loadBalancers, targetGroups, and securityGroupNames

jonsie:
This is the safest way to resolve this issue.  Setting an empty list default on the properties would work too, but I don't trust that all usages check for an empty list and my confidence in writing more Groovy at this point is at -1.


robzienert:
> Setting an empty list default on the properties would work too

This assertion is almost assuredly incorrect. There's places in our atomic operations that use Groovy truth where this would be true, but there's also flows that explicitly check these properties for a `null` value to determine what behavior to take - setting the default value from `null` would likely break other things, unfortunately.

jonsie:
> This assertion is almost assuredly incorrect. There's places in our atomic operations that use Groovy truth where this would be true, but there's also flows that explicitly check these properties for a `null` value to determine what behavior to take - setting the default value from `null` would likely break other things, unfortunately.

Yeah, that's what I meant by - I don't want to have to write any more Groovy.  Cause I'd have to go in and look for those places and fix them (if I set the an empty list default).

PR #4538, opened 2020-04-22
refactor(google): Convert GoogleCommonSafeRetry to java

ezimanyi:
* refactor(google): Convert GoogleCommonSafeRetry to java

  In addition to converting this to Java, this commit also allows updates the bare types to be generic, so that the methods return something more strongly-typed than an object, which is very useful for strongly-typing any downstream code without casts.

  It also updates the class to no longer be a superclass of the places that use it; it instead encapsulates its logic and consumers can create one and store it as an instance variable.

  As part of this change, instead of having consumers inject in an exception factory, this class just throws a GoogleAPIException, which consumers can then wrap as they see fit.

  (I came across this while fixing a bug in front50 that uses this, unfortunately by importing clouddriver as a library, and was unable to do so in a typesafe way without this change.)

  This class is pretty robustly tested, as the GCE and appengine operations all run through this class and often assert that the right type of exception is thrown (in addition to the tests directly on this class). For example, when I had a bug in the ResourceInUseException code, it was caught by a GCE test.

  There are two changes to the tests on this class that are worth noting:
  * In one case, the exception type is changed from IllegalStateException to GoogleAPIException. This was just because the tests was using IllegalStateException in its factory.
  * In another case we've changed from returning GoogleJsonResponseException to a GoogleAPIException. I'm fairly certain this was originally a bug as we were incorrectly throwing that exception instead of storing it and processing it through the logic to decide what to return. From reading the places this function is called, it doesn't look like we are catching a GoogleJsonResponseException anywhere so I don't think this should affect any consumers.

  Finally, the isRetryable method has not been used in ~2 years; at some point a refactor removed the call to it (and its logic to retry all 5xx errors). As I don't want to re-introduce that now without a reason, I just removed the function and its tests.

* refactor(appengine): Remove success field from appengine

  Ideally the retry logic would not accept a list of statuses to consider success, as this means that we need it to be nullable because if we get one of those statuses we need to return but don't have anything to return. (This might be more resonable if we had separate calls for operations that returned a value and operations that didn't; in that case the ones that don't return a value could more reasonably accept this parameter.)

  As this isn't actually used anywhere in appengine, let's just delete it to clean up the code a bit.


ezimanyi:
I could probably spend another week trying to clean up this code, but hopefully this is better than the groovy that was there before, and most importantly you can now get back something more specific than `Object` from `doRetry` which was how I started down this path in the first place.

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/AbstractStartStopAppengineAtomicOperation.groovy
```diff
@@ -75,8 +74,8 @@ abstract class AbstractStartStopAppengineAtomicOperation extends AppengineAtomic
       { callApi(project, loadBalancerName, serverGroupName, version) },
       "version",
       task,
-      [409],
-      [],
+      [409]
+      ,
```
plumpy, line 78:
```suggestion
      [409],
```

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DestroyAppengineAtomicOperation.groovy
```diff
@@ -67,11 +67,13 @@ class DestroyAppengineAtomicOperation extends AppengineAtomicOperation<Void> {
...
-      { appengine.apps().services().versions().delete(project, loadBalancerName, serverGroupName).execute() },
+      {
+        appengine.apps().services().versions().delete(project, loadBalancerName, serverGroupName).execute()
+      },
       "version",
       task,
-      [409],
-      [],
+      [409]
+      ,
```
plumpy, line 76:
```suggestion
      [409],
```

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/DisableAppengineAtomicOperation.groovy
```diff
@@ -70,11 +69,13 @@ class DisableAppengineAtomicOperation extends AppengineAtomicOperation<Void> {
...
-      { buildNewLoadBalancerAndCallApi(credentials.project, loadBalancerName, serverGroupName, priorOutputs) },
+      {
+        buildNewLoadBalancerAndCallApi(credentials.project, loadBalancerName, serverGroupName, priorOutputs)
+      },
       "version",
       task,
-      [409],
-      [],
+      [409]
+      ,
```
plumpy, line 78:
```suggestion
      [409],
```

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/UpsertAppengineAutoscalingPolicyAtomicOperation.groovy
```diff
@@ -78,8 +78,8 @@ class UpsertAppengineAutoscalingPolicyAtomicOperation extends AppengineAtomicOpe
       { callApi(projectName, loadBalancerName, serverGroupName, version) },
       "version",
       task,
-      [409],
-      [],
+      [409]
+      ,
```
plumpy, line 82:
```suggestion
      [409],
```

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/ops/UpsertAppengineLoadBalancerAtomicOperation.groovy
```diff
@@ -85,8 +83,8 @@ class UpsertAppengineLoadBalancerAtomicOperation extends AppengineAtomicOperatio
         callApiClosure,
         "service",
         task,
-        [409],
-        [],
+        [409]
+        ,
```
plumpy, line 87:
```suggestion
        [409],
```

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/AppengineSafeRetry.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Map;
+import javax.annotation.Nullable;
+import javax.annotation.ParametersAreNullableByDefault;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@Component
+@NonnullByDefault
+public class AppengineSafeRetry {
```
maggieneterval, line 37: Just curious how you decided which refactored classes should become `final` -- is there any reason this one did not?

clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/deploy/GoogleCommonSafeRetry.java
```diff
@@ -0,0 +1,203 @@
...
+            "{} attempt #{} encountered retryable statusCode={} with error message: {}.",
+            description,
+            tries,
+            jsonException.getStatusCode(),
+            jsonException.getMessage());
+      } catch (SocketTimeoutException toEx) {
+        log.warn("Retryable {} attempt #{} timed out.", description, tries);
+      }
+      tries++;
+      // Sleep with exponential backoff based on the number of retries. Add retry jitter with
```
maggieneterval, line 138: Wow, I had never heard the term "retry jitter" to describe this special case of exponential backoff...love it.
♻️ 🕺 🐞

clouddriver-appengine/src/main/groovy/com/netflix/spinnaker/clouddriver/appengine/deploy/AppengineSafeRetry.java
ezimanyi: Good catch, I should make this `final` too.

ezimanyi:
* refactor(google): Move GoogleAPIException to top level

  Particularly as this starts having subclasses for particular exception types, it makes sense for it to be at the top level, not nested in the GoogleCommonSafeRetry.

* refactor(google): Clean up handling of error details

  The fromGoogleJsonException method was doing two things:
  (1) parsing out the details on the exception to figure out what we know about it
  (2) deciding what type of GoogleApiException to throw

  This was alredy a bit complicated but I'm about to add another subclass of GoogleApiException so things are going to get out of hand if we don't fix it. Create and ErrorDetails private class that figures out what we know about an exception and puts it in a standard format, then let fromGoogleJsonException base its decision on this class.

* refactor(google): Return NotFoundException from API calls

  The code in front50 that uses GoogleCommonSafeRetry was relying on getting back the raw GoogleJSONResposeException, but only because it was digging into the exception to figure out if it was a 404 to respond differently. Let's instead encapsulate this and return a GoogleApiException.NotFoundException in this case so we can catch that.

* refactor(google): Make GoogleApiException checked

  This might be controversial, and definitely interested in feedback on this one, but the expectation at this point is that all callers are catching this and wrapping it in something that describes what they were doing when the exception happened. Because we're expecting them to catch it, having this be a checked exception actually makes it easier to ensure we didn't miss a place...but I also could see the point that there's not much you can do about this beyond wrap it so maybe it shouldn't be checked.

ezimanyi:
@maggieneterval @plumpy : Addressed both of your comments.  I also ran into one issue while incorporating this into where it's called in `front50` where I needed to throw a `NotFoundException` if it was a 404, so added the last 4 commits to address this.  Definitely would like a quick re-review and comments on those before merging.  Thanks!

clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/deploy/GoogleApiException.java
```diff
@@ -19,40 +19,61 @@
...
+  private static class ErrorDetails {
+    private final int statusCode;
+    private final String message;
+    private final String reason;
+
+    static ErrorDetails fromGoogleJsonException(GoogleJsonResponseException e) {
+      Optional<ErrorInfo> optionalErrorInfo =
+          Optional.ofNullable(e.getDetails()).map(GoogleJsonError::getErrors)
+              .orElse(ImmutableList.of()).stream()
+              .findFirst();
```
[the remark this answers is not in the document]
ezimanyi: I'm not sure to be honest why it might have more than one error.  Maybe if it's a batch request with individual components, but I'm really just guessing at this point 🤷 

Given we've only been looking at the first one now I'm probably inclined to leave this until some use case arises where the other messages are useful.

clouddriver-google-common/src/main/groovy/com/netflix/spinnaker/clouddriver/googlecommon/deploy/GoogleApiException.java
```diff
@@ -19,40 +19,61 @@
...
-    if (optionalErrorInfo.isPresent()) {
-      ErrorInfo errorInfo = optionalErrorInfo.get();
-      String message =
-          String.format(
-              "Operation failed. Last attempt returned status code %s with error message %s and reason %s.",
-              e.getStatusCode(), errorInfo.getMessage(), errorInfo.getReason());
-      if ("resourceInUseByAnotherResource".equals(errorInfo.getReason())) {
-        return new ResourceInUseException(message);
+    ErrorDetails errorDetails = ErrorDetails.fromGoogleJsonException(e);
+    if (errorDetails.getReason().equals("resourceInUseByAnotherResource")) {
```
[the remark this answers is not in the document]
ezimanyi: Good catch! I thought I had made this null-safe because we're never passing in null, but I forgot that we're passing `errorDetails.getMessage()` (and similar for reason) to the constructor and they can return null.

I've fixed it by just explicitly writing out the constructor and doing the `nullToEmpty` there rather than marking it nullable, so the calling code doesn't need to worry about nulls.

PR #4540, opened 2020-04-22
feat(api): Allow retrieval of security groups by ID

luispollo:
Allows the retrieval of security groups by ID, as opposed to by name.

This is in support of fixing https://github.com/spinnaker/keel/issues/1060, and resolves an "impedance mismatch" problem in that the clouddriver APIs that return server groups contain only SG IDs, not names, whereas the APIs to return SGs previously only supported names.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonSecurityGroupProvider.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonSecurityGroupProvider.groovy
@@ -140,6 +140,7 @@ class AmazonSecurityGroupProvider implements SecurityGroupProvider<AmazonSecurit
     }
   }
 
+  @Override
   AmazonSecurityGroup getById(String account, String region, String securityGroupId, String vpcId) {
     getAllMatchingKeyPattern(Keys.getSecurityGroupKey('*', securityGroupId, region, account, vpcId), true)[0]
   }
--- a/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2SecurityGroupProvider.java
+++ b/clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2SecurityGroupProvider.java
@@ -144,4 +144,11 @@ public class KubernetesV2SecurityGroupProvider
         .findFirst()
         .orElse(null);
   }
+
+  @Override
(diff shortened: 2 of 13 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonSecurityGroupProvider.groovy
```diff
@@ -140,6 +140,7 @@ class AmazonSecurityGroupProvider implements SecurityGroupProvider<AmazonSecurit
     }
   }
 
+  @Override
```
luispollo, line 143: Turns out this was already here -- I just had to hook into it in the controller.

clouddriver-kubernetes-v2/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/v2/caching/view/provider/KubernetesV2SecurityGroupProvider.java
```diff
@@ -144,4 +144,9 @@ public KubernetesV2SecurityGroup get(
         .findFirst()
         .orElse(null);
   }
+
+  @Override
+  public KubernetesV2SecurityGroup getById(String account, String region, String id, String vpcId) {
+    throw new UnsupportedOperationException("Not currently implemented.");
```
luispollo, line 150: I took the liberty of implementing the overrides for all sub-classes, since they seemed straightforward, but not this one, for the opposite reason.

robzienert: cc @maggieneterval @ezimanyi

luispollo: BTW, I didn't think this was of any concern at this time since the controller change is effectively exposing a new API for all practical purposes (meaning no one would have been using it), but would definitely love a contribution to fill in this gap!

maggieneterval: Thanks for the heads up! Is keel the only anticipated client of this new API (for now)?

luispollo: Yes, it is.

PR #4541, opened 2020-04-23
fix(aws): allow overriding of AWS account Spring beans

claymccoy:
Allowing CredentialsConfig and CredentialsLoader to be overridable with custom Spring components and putting the private synchronizeAmazonAccounts logic into its own AmazonAccountsSynchronizer component that could also be overridden or reused. This allows more flexibility for users that have special AWS account and cred loading needs.


claymccoy:
I didn't write any code in Groovy, I just moved code that was already in Groovy and inaccessible.

jonsie:
Sorry for the delay here.  This looks like a good refactor to me.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonAccountsSynchronizer.groovy
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.aws.security
+
+interface AmazonAccountsSynchronizer {
```
robzienert, line 19: Nooooo! Why the Groovy? 😢

PR #4542, opened 2020-04-23
fix(ecs): make targetgroup service set mutable

allisaurus:
Initializing set of services mapped to a target group as mutable so additional services can be added if found. Previously, EcsLoadBalancerProvider would return `java.lang.UnsupportedOperationException` when trying to add a second service to an existing set.

## testing
ECS target groups mapped to multiple services/server groups display properly
multi_sg_tg


clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsLoadBalancerProvider.java
```diff
@@ -158,7 +158,7 @@ public Item get(String name) {
             serviceList.add(service.getServiceName());
             targetGroupToServicesMap.put(tgArn, serviceList);
           } else {
-            Set<String> srcServices = Collections.singleton(service.getServiceName());
+            Set<String> srcServices = new HashSet<>(Arrays.asList(service.getServiceName()));
```
pkandasamy91, line 161: You could also do something like this `Set<String> srcServices = Sets.newHashSet<>(service.getServiceName());`

PR #4544, opened 2020-04-24
fix(provider/cf): added space caching agent

zachsmith1:
This change was to help large PCF clients with potentially hundreds/thousands of spaces. We created a new caching agent that caches these org/spaces so that they can be retrieved quickly when the call to /credentials is running. This /credentials call runs often so this PR should improve the time it takes to respond and mitigate timeout issues currently seen in production by some clients. 

Some logic was adjusted across the AbstractCloudFoundryCachingAgent so that it wouldn't be duplicated by different implementations. You will see that the abstract class now takes CloudFoundryCredentials through its constructor, instead of account and client, both of which can be retrieved from CloudFoundryCredentials.


kevinawoo:
FYI this supports SQL and Redis because there's some fancyness happening with the data storage layer YAY! 🙌

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/provider/agent/CloudFoundrySpaceCachingAgent.java
```diff
@@ -0,0 +1,102 @@
...
+                SPACES.getNs(),
+                spaces.stream()
+                    .map(
+                        s ->
+                            new ResourceCacheData(
+                                Keys.getSpaceKey(accountName, s.getRegion()),
+                                cacheView(s),
+                                emptyMap()))
+                    .collect(toSet()))
+            .toJavaMap();
```
plumpy, line 78: Try Guava's ImmutableMap.of()... It's much more standard than this vavr thing, plus has the benefit of immutability.

(Looking around I see this vavr thing is used all over CloudFoundry for some reason, but Guava is much, much more common in the rest of Spinnaker/the world, so we should prefer that.)

ZachDev1: @plumpy done!

PR #4545, opened 2020-04-24
fix(aws): fix bug in setting launch config/template on cluster

jeyrschabu:
- fixed a bug where the launch config/template was not read correctly
- also added null checks on setting build info and image

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonClusterProvider.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonClusterProvider.groovy
@@ -84,8 +84,8 @@ class AmazonClusterProvider implements ClusterProvider<AmazonCluster>, ServerGro
     }
 
     def asg = serverGroupData.attributes["asg"]
-    def serverGroup = new AmazonServerGroup(serverGroupData.attributes)
     def serverGroupById = [(serverGroupData.id): new AmazonServerGroup(serverGroupData.attributes)]
+    def serverGroup = serverGroupById.values().first()
 
     String imageId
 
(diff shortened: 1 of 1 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonClusterProvider.groovy
```diff
@@ -564,17 +565,17 @@ class AmazonClusterProvider implements ClusterProvider<AmazonCluster>, ServerGro
...
-          continue
-        }
-
-        if (serverGroup.asg?.launchTemplate) {
-          def launchTemplateSpec = serverGroup.asg.launchTemplate as Map
-          serverGroup.launchTemplate = getLaunchTemplateVersion(ld, launchTemplateSpec.version as String)
-        } else {
-          serverGroup.launchConfig = ld.attributes
+      if (ld.relationships != null && ld.relationships.containsKey(SERVER_GROUPS.ns)) {
+        ld.relationships.serverGroups.each {
```
cfieber, line 569: I'd suggest `ld.relationships[SERVER_GROUPS.ns]` here since that is what the containsKey is checking on

jeyrschabu: Good catch! thanks

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonClusterProvider.groovy
```diff
@@ -589,7 +590,7 @@ class AmazonClusterProvider implements ClusterProvider<AmazonCluster>, ServerGro
     }
 
     imageData.each { ld ->
-      if (ld.relationships.containsKey(SERVER_GROUPS.ns)) {
+      if (ld.relationships != null && ld.relationships.containsKey(SERVER_GROUPS.ns)) {
         def serverGroup = serverGroups[ld.relationships.serverGroups.first()]
```
cfieber, line 594: same here as https://github.com/spinnaker/clouddriver/pull/4545/files#r414892001

PR #4547, opened 2020-04-25
fix(aws): Fixing npe on setting launch settings & build info

jeyrschabu:
- added null check


jeyrschabu:
I notice a couple of npe (Perhaps due to bad data). This is to ensure an npe is not raised.

jeyrschabu:
Self merging this for a quick test. Will revert if I see issues

PR #4548, opened 2020-04-27
chore(kubernetes): Pull latest kubectl patch release

ezimanyi:
We're currently using kubectl 1.16.0 in the docker container; let's bump that to the latest patch release, 1.16.9.

I'm not pulling in a new minor version, as kubectl only allows a skew of +/- 1 minor version from the API, so bumping to 1.17.x could potentially break users on Kubernetes 1.15. Keeping us on 1.16.x allows us to have full support for anyone on Kubernetes 1.15-1.17.


ezimanyi:
Minor bump that would be good to get into 1.20.

PR #4559, opened 2020-04-30
feat(provider/google): Improving support gcp custom machine types

sergeylanzman:
- add support e2/n1/n2-custom types


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/validators/StandardGceAttributeValidator.groovy
```diff
@@ -294,20 +294,26 @@ class StandardGceAttributeValidator {
...
     }
 
-    def ( vCpuCount, memory ) = instanceType.split('-').tail().collect { it.toDouble() }
+    def vCpuCount, memory
+    if (instanceType.startsWith('custom-')) {
+      (vCpuCount, memory) = instanceType.split('-').tail().collect { it.toDouble() }
+    } else{
+      (vCpuCount, memory) = instanceType.split('-').tail().tail().collect { it.toDouble() }
+    }
+
```
plumpy, line 316: As long as we're using a regex here, we may as well actually use it and not switch to string splits halfway through, I suppose...

```
  def customInstanceRegExp = /(\w-)?custom-(\d{1,2})-(\d{4,6})/

  def validateCustomInstanceType(String instanceType, String location, GoogleNamedAccountCredentials credentials) {
    def customTypeMatcher = instanceType =~ customInstanceRegExp
    if (!customTypeMatcher.matches()) {
      errors.rejectValue("instanceType", "${context}.instanceType.invalid", "Custom instance string must match pattern $customInstanceRegExp.")
      return false
    }

    def vCpuCount = customTypeMatcher.group(2).toDouble()
    def memory = customTypeMatcher.group(3).toDouble()
```

sergeylanzman: yes, it's better use regex groups, right regex `/(.*)-?custom-(\d{1,2})-(\d{4,6})/` because `(\w-)` return from `e2` only `2`. 

thanks!

sergeylanzman:
@plumpy 
Add second commit fix changes in validations process.
New custom types available  from 0.5GB to 8GB ram per VCPU
https://cloud.google.com/compute/docs/instances/creating-instance-with-custom-machine-type

plumpy:
Oh boy, those memory checks... Doing a bunch of validation for business logic we don't even own seems like a pretty poor choice. But I guess it's probably safer to just leave it there for now :confused: Thanks for noticing and fixing it.

And thanks for the PR!

erancx:
can this be cherry picked into 1.20?

plumpy:
Cherrypicked!

PR #4560, opened 2020-05-01
fix(provider/cf): Ensure up-to-date data in server group caching agent and force cache

ncknt:
Prior to this change, CF server group caching agent was updating its internal cache in a `ForkJoinPool` without waiting for the result. The agent was then saving stale data in the cache. Similarly, force cache refresh was not doing an actual refresh if the server group had already been loaded. This could lead to stale data overriding good data in the case of force cache running on a different instance of Clouddriver.

This is the first in a series of changes to fix CF provider caching. This one is limited so it can be back-ported.

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
@@ -145,46 +145,50 @@ public class Applications {
     // if the update time doesn't match then we need to update the cache
     // if the app is not found in the cache we need to process with `map` and update the cache
     try {
-      forkJoinPool.submit(
-          () ->
-              newCloudFoundryAppList
-                  .parallelStream()
-                  .filter(
-                      app -> {
-                        CloudFoundryServerGroup cachedApp = findById(app.getGuid());
-                        if (cachedApp != null) {
-                          if (!cachedApp
-                              .getUpdatedTime()
-                              .equals(app.getUpdatedAt().toInstant().toEpochMilli())) {
-                            log.trace(
-                                "App '{}' cached version is out of date on foundation '{}'",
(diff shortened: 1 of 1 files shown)


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
```diff
@@ -145,46 +145,50 @@ public CloudFoundryServerGroup findById(String guid) {
...
+                              log.trace(
+                                  "App '{}' not found in cache for foundation '{}'",
+                                  app.getName(),
+                                  this.account);
+                              return true;
+                            }
+                          })
+                      .map(this::map)
+                      .forEach(sg -> serverGroupCache.put(sg.getId(), sg)))
+          .get();
```
ethanfrogers, line 178: This isn't a blocker since the original code is like this but it'd be _really nice_ if the code within the `filter` was broken out into a different method. This bit of code is hard to reason about.

ethanfrogers: I'm not familiar with `ForkJoinPool`. I assume the absence of the call to `get` in the previous implementation was causing the task to end prematurely and resulting in the stale data?

ncknt: Yeah we'll revisit that part in a later change. I believe it's still going through the health check too many times. The `get()` (could have used `join()`) forces the code to wait for the parallel execution to complete.

ncknt:
@spinnakerbot cherry-pick 1.20

ncknt:
@Mergifyio backport release-1.19.x

ethanfrogers:
@Mergifyio backport release-1.19.x

PR #4567, opened 2020-05-05
fix(ha): Fix deploying RO clouddriver with Halyard

ezimanyi:
Fixes spinnaker/spinnaker#5736

My recent change to push the -ro config for clouddriver down to the base config breaks deploying RO clouddriver with Halyard because it sets the redis endpoint to redisRo.baseUrl, which is not set by Halyard. This should default to redis.baseUrl to handle this case.


ezimanyi:
@Mergifyio backport 1.20

ezimanyi:

PR #4569, opened 2020-05-05
 refactor(kubernetes): remove kubernetes v1 provider

maggieneterval:
* refactor(kubernetes): remove kubernetes-v1 

* refactor(kubernetes): merge kubernetes-v2 into kubernetes 

* refactor(kubernetes): remove clouddriver-kubernetes dependency on clouddriver-docker


maggieneterval:
This PR was already getting huge, so it does not include any of the following cleanups that I'd like to do in a separate PR:
- Simplify classes that are now the only implementation of some previously-shared V1/V2 interface.
- Remove `providerVersion` and `skin`, as there are now no account types with non-default implementations (This will require Deck work first).

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/KubernetesCloudProvider.java
```diff
@@ -1,17 +1,18 @@
 /*
- * Copyright 2015 Google, Inc.
+ * Copyright 2020 Google, Inc.
```
ethanfrogers, line 2: We don't normally update license headers on existing files. Was this on purpose?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesV2JobStatus.java
```diff
@@ -1,7 +1,7 @@
 /*
- * Copyright 2019 Armory
+ * Copyright 2020 Google, Inc.
```
ethanfrogers, line 2: At this point I'm assuming this was the editor making these changes?

ethanfrogers:
😢 
image

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/JobController.groovy
```diff
@@ -45,6 +45,7 @@ class JobController {
   @Autowired
   FiatPermissionEvaluator fiatPermissionEvaluator
 
+  @PreAuthorize("hasPermission(#application, 'APPLICATION', 'READ') and hasPermission(#account, 'ACCOUNT', 'READ')")
```
ethanfrogers, line 48: Is this something we want to do at the same time we remove all of this code?

maggieneterval: Can def pop in a separate PR if that makes more sense to y'all! (Currently it is just in its own commit.)

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesV2JobStatus.java
maggieneterval: Ugh, yeah I think my IDE was overambitious here...this will be fun to fix.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/JobController.groovy
ethanfrogers: IMO, that's a cleaner, more atomic, change.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesV2JobStatus.java
maggieneterval: Just came across one instance of a file you wrote but ended up with the Netflix header, so I'm also changing that to correct history 👼

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/JobController.groovy
maggieneterval: Removed and will be coming soon in a different PR!

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/KubernetesCloudProvider.java
maggieneterval: Thanks for catching this; all fixed!

PR #4570, opened 2020-05-05
fix(provider/gce): Renaming google server group agents to include clustercaching so that it becomes authoritative for clusters in cats-sql

luckoseabraham:
Renaming google server group agents to include clustercaching so that it becomes authoritative for clusters in cats-sql.

When migrating to sql backend for clouddriver, cats_v1_clusters are not getting populated. This is because there are no authoritative agents for CLUSTERS in GCE agents.
For aws, there was a workaround done here to check for agents name containing clustercaching and if present make it authoritative for CLUSTERS.
https://github.com/spinnaker/clouddriver/blob/master/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/SqlProviderCache.kt#L171

For GCE provider agents as well, there are no agents which even has that pattern. So renaming the servergroup agents to fit into the pattern for cluster authoritative.

This is my first pull request to spinnaker. So apologies If I missed anything.

Thank you for your patience.


luckoseabraham:
@spinnaker/reviewers

luckoseabraham:
@plumpy - Would love to hear your thoughts :)

robzienert:
👍 Should work fine.

plumpy:
Thanks for the PR!

I definitely agree with the comment in `SqlProviderCache` that this is a terrible hack :grin: As far as I can tell, it's not even necessary anymore because the TODO items have both been fixed. @robzienert maybe you should just delete that whole block?

For us, I think just changing `AbstractGoogleServerGroupCachingAgent` to be `AUTHORITATIVE` for clusters should be fine. We probably just copy/pasted what AWS did way back in the day. I can't see any reason why we can't just change it.

luckoseabraham:
Thank you so much @plumpy  and @robzienert for your comments. 
I just did a quick test and it looks good on my end. Updated the PR with your suggestions 👍

luckoseabraham:
Thanks a lot for merging this PR @plumpy 👍 
Unfortunately for us, most of the users are using the V1 k8s provider and are not really ready for the v2 k8s provider. And 1.21 release removes the v1 provider code as well.

Do you think it's reasonable to have this change back-ported into any of the older versions ? 1.19.x or 1.20.x ?

PR #4572, opened 2020-05-06
refactor(kubernetes): post-V1-provider removal cleanups

maggieneterval:
* refactor(kubernetes): remove v1-specific logic from JobController

  (cc @AbdulRahmanAlHamali)

* refactor(kubernetes): merge KubernetesV2Configuration into KubernetesConfiguration 

* refactor(kubernetes): remove unnecessary providerVersion from tests 

* refactor(kubernetes): throw error on startup if V1 accounts are configured


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/JobController.groovy
```diff
@@ -45,6 +45,7 @@ class JobController {
   @Autowired
   FiatPermissionEvaluator fiatPermissionEvaluator
```
AbdulRahmanAlHamali, line 46: You might want to remove this dependency since it's not used any more

maggieneterval: Thanks @AbdulRahmanAlHamali, great catch!

PR #4578, opened 2020-05-07
chore(build): gradle 6.4

cfieber:


cfieber:
will be fixed by https://github.com/spinnaker/spinnaker-gradle-project/pull/138

PR #4582, opened 2020-05-11
fix(provider/cf): Adjust lookup of lb's per spinnaker app to avoid retrieving all lb cache

zachsmith1:
This change adjusts the way that the LoadBalancerProvider searches for LB's for a particular spinnaker application. Initially, the logic would search for all loadbalancers for a particular provider and then filter through based on serverGroup/appName matching. This logic works fine for smaller CF users but causes timeouts for larger clients with thousands of loadbalancers. 

The new logic flips how we are doing loadBalancer searching. First, we ensure that the loadBalancerCachingAgent stores a link to the SG it's bound to. Then, we retrieve the loadbalancers tied to SG's which are part of the spinnaker application. The result is that we get the keys specific to the loadbalancers being requested and avoid large lookups.

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/cache/CacheRepository.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/cache/CacheRepository.java
@@ -169,6 +169,36 @@ public class CacheRepository {
         findServerGroupsByKeys(lbData.getRelationships().get(SERVER_GROUPS.getNs()), Detail.NONE));
   }
 
+  public Set<CloudFoundryLoadBalancer> findLoadBalancersByClusterKeys(
+      Collection<String> keys, Detail detail) {
+    Set<String> serverGroupKeys =
+        cacheView.getAll(CLUSTERS.getNs(), keys).stream()
+            .flatMap(cl -> cl.getRelationships().get(SERVER_GROUPS.getNs()).stream())
+            .collect(toSet());
+
+    Set<String> loadBalancerKeys =
+        cacheView.getAll(SERVER_GROUPS.getNs(), serverGroupKeys).stream()
+            .flatMap(
+                sg ->
+                    sg.getRelationships().get(LOAD_BALANCERS.getNs()).stream()
+                        .map(
+                            lb ->
+                                Keys.getLoadBalancerKey(
+                                    objectMapper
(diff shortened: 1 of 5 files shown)


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/provider/agent/CloudFoundryLoadBalancerCachingAgent.java
```diff
@@ -84,12 +84,34 @@ public CacheResult loadData(ProviderCache providerCache) {
           }
         });
 
+    Map<String, CacheData> loadBalancersByServerGroupIds = new java.util.HashMap<>();
```
ncknt, line 87: import `java.util.HashMap`

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/cache/CacheRepository.java
```diff
@@ -169,6 +169,36 @@ private CloudFoundryLoadBalancer loadBalancerFromCacheData(CacheData lbData, Det
         findServerGroupsByKeys(lbData.getRelationships().get(SERVER_GROUPS.getNs()), Detail.NONE));
   }
 
+  public Set<CloudFoundryLoadBalancer> findLoadBalancersByClusterKeys(
+      Collection<String> keys, Detail detail) {
+    Set<String> serverGroupKeys =
```
ncknt, line 174: Could we go directly to `serverGroups` without using `clusters` by using `cacheView.filterIdentifiers(SERVER_GROUPS.getNs(), Keys.getClusterKey("*", app, "*"))`?

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/provider/agent/CloudFoundryLoadBalancerCachingAgent.java
zachsmith1: easier to leave then swap out (for now)

zachsmith1:

ethanfrogers:

ethanfrogers:
@mergifyio backport release-1.19.x

PR #4583, opened 2020-05-11
fix(kubernetes): return kubernetes server group managers

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/5754

"I love bounded wildcards."

-- Winston Churchill

(For the  Java-11 refactor, we parameterized `ServerGroupManagerProvider` with `ServerGroupManager`, but the Kubernetes implementation of this provider is actually parameterized with a subclass of `ServerGroupManager`, so we were failing to pick up that implementation and never returning any Deployment or Stateful Set details, manifesting in the UI errors reported in that GH issue.)


maggieneterval:

PR #4586, opened 2020-05-11
refactor(artifacts): Make ArtifactCredentials null-safe

ezimanyi:
This was mostly a case study in understanding how we handle null vs. empty strings in config parameters to help inform some technical decisions in kleat.  But in order to better understand how things are currently handled I needed to dig into some examples, so chose artifacts.  This PR basically takes all the scattered null vs. empty distinctions and moves them to the API boundary; it's something that we should consider doing everywhere, so hopefully this can be a reasonable example.

It's a bit more code because we don't get free constructors from Lombok given that we have to do null handling...but given that we have to handle that in the constructors, we free ourselves from needing to do that elsewhere so I think this is better.

* refactor(artifacts): Make artifact accounts immutable and null-safe

  There are a lot of config properties that can be set for artifacts. In general, they don't care about the difference between unset and null values, but this is enforced at various points of rather than at the API boundary. In order to make it more clear that this is the case, update the configuration classes to be immutable and to translate null strings to empty strings upon creation. This removes the ability of any consuming code to be able to see a null value in these string properties.

  In particular, we make all fields on artifact accounts immutable, and replace the setters with a constructor. The ConstructorBinding annotation is what tells Spring to bind properties using the constructor (without it, it will look for setters and fail). We also provide a Builder for convenience (primarily for testing) to avoid needing to use the many-arg-constructor.

  As the constructor will be called by Spring, which does allow null properties we need to treat all input as potentially null and convert to non-null in the constructor.

* refactor(artifacts): Make ArtifactCredentials null-safe

  Now that the artifact accounts are immutable, the artifact credentials that read from them can be made NonNullByDefault. This commit also removes a number of now-redundant null-checks on its fields, replacing them with .isEmpty() on a now guaranteed to be non-null string.

  This commit also updates one of the interface methods to return an ImmutableList (which all but one implementation was effectively doing by returning Collections.singletonList), and makes the interface NonnullByDefault.

  There was a single implementation of download() that was not null-safe. I've updated it to eagerly throw an exception; all callers of that method that was updated immediately pass the output to IOUtils.copy which will throw a null pointer exception. This change will simply provide more context around the exception.


clouddriver-artifacts/src/test/java/com/netflix/spinnaker/clouddriver/artifacts/ivy/IvyArtifactProviderPropertiesTest.java
```diff
@@ -1,44 +0,0 @@
...
-
-package com.netflix.spinnaker.clouddriver.artifacts.ivy;
-
-import com.fasterxml.jackson.databind.ObjectMapper;
-import java.io.IOException;
-import org.junit.jupiter.api.Test;
-
-class IvyArtifactProviderPropertiesTest {
-  @Test
-  void deserializeIvySettings() throws IOException {
```
ezimanyi, line 25: The only thing this test was testing was that we could deserialize the account from JSON, but as far as I can tell we never actually deserialize artifact accounts from JSON.  These are classes meant to hold Spring config data and are initialized by Spring; as `IvyArtifactProviderProperties.class` is package-private nobody outside the package should be deserializing it either.  I suspect this was basically added as a placeholder test to have some coverage for the class.

The test breaks because deserializing immutable classes requires a bunch more Jackson annotations, etc...which I could add but given that there seems to be no need to support this I'd rather not include, so I deleted the test.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactCredentials.java
```diff
@@ -17,19 +17,20 @@
...
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
+@NonnullByDefault
 @Slf4j
 final class BitbucketArtifactCredentials
     extends SimpleHttpArtifactCredentials<BitbucketArtifactAccount> implements ArtifactCredentials {
   @Getter private final String name;
-  @Getter private final List<String> types = Collections.singletonList("bitbucket/file");
+  @Getter private final ImmutableList<String> types = ImmutableList.of("bitbucket/file");
```
cfieber, line 33: This doesn't provide any benefit, but adds a guava dependency to this class (Collections.singletonList is already an immutable list, is there any strong reason to expose the details of the collection type?

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/config/TokenAuth.java
```diff
@@ -31,9 +32,9 @@
 
   default Optional<String> getTokenAsString() {
     String token = null;
-    if (StringUtils.isNotEmpty(getTokenFile())) {
+    if (!getTokenFile().isEmpty()) {
```
cfieber, line 35: In this case I'd lean more towards having getTokenFile and getToken return Optionals than never-null Strings where empty has a special meaning - it feels like Optional captures the intent better. WDYT?

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/config/ArtifactCredentials.java
```diff
@@ -18,17 +18,20 @@
...
 import java.util.List;
 import java.util.Optional;
 import org.apache.commons.lang3.NotImplementedException;
 
+@NonnullByDefault
 public interface ArtifactCredentials {
   String getName();
 
-  List<String> getTypes();
+  ImmutableList<String> getTypes();
```
cfieber, line 34: I'm not sure this needs to be on the interface signature unless there are extra methods on ImmutableList that are being used.

cfieber:
I'm bigly in favor of more immutable.

I think there are a bunch of cases in here where the intent would be clearer by translating empty strings to nulls in the value classes and having the getters exposed as Optionals instead

And I hate to be the guy that always craps on guava, but I'm always hesitant about adding more usage of it because it evolves very aggressively and at some point we will hit an unresolvable dependency conflict with it. The cases of replacing Collections.singletonList didn't really seem like they were worth it to me.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactAccount.java
```diff
@@ -17,14 +17,31 @@
...
+  private final String username;
+  private final String password;
+  private final String usernamePasswordFile;
+
+  @Builder
+  @ConstructorBinding
+  @ParametersAreNullableByDefault
+  BitbucketArtifactAccount(
+      String name, String username, String password, String usernamePasswordFile) {
+    this.name = Strings.nullToEmpty(name);
```
cfieber, line 42: is it valid to have an artifact account with no name? (generally asking this question to all the artifact account implementations). It is probably worth a not-null-or-empty check where we can add legitimate constraint checking

ezimanyi: I actually don't think that we prevent a null name anywhere at this point, so would be a bit wary of asserting that as part of this PR.  That also goes for adding more assertions to these constructors for other fields that can't be null...while in general I think we probably should do these assertion checks at the time of object creation, we are currently failing downstream at the time of use.  So moving constraint checks here would have the effect of breaking startup for people who have misconfigured accounts but weren't using them.  I guess we could also catch and warn on failed account creation to mitigate that...but this is getting out of the scope of what I really wanted to bite of with this PR.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/config/TokenAuth.java
ezimanyi: That's a good point, if it really is optional we should just handle that in the constructor; I'll make that change.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactCredentials.java
ezimanyi: I guess the main benefit is that returning a `List` that happens to be immutable doesn't give any hint to the consumer that it is immutable.  This makes it clear to the consumer that the returned list is immutable, which seems like a reasonable detail to expose.  I know @plumpy also has thoughts on this, so interested in his opinion.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/config/ArtifactCredentials.java
ezimanyi: There aren't additional methods, but similar to the above comment, it seemed useful to indicate to consumers (and to implementation) that the list returned from `getTypes()` should be immutable.

ezimanyi:
Are there examples of where Guava has introduced breaking changes to the more stable aspects of it like immutable collections and strings?  The potential of an upgrade to Guava breaking `ImmutableList.of()` seems remote to me, but I also don't know the context for the concern.

In general, I think that the benefit from explicit immutability outweighs the additional dependency on Guava (which is already on the classpath almost everywhere in our codebase).  While obviously the immutability could be done as an implementation detail and we could expose just a `List` or a `Collection` that happens to be immutable, it also seems like the additional information provided to callers is useful.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/config/ArtifactCredentials.java
plumpy: Yeah, generally speaking Google recommends using `Immutable*` as a return type because it gives additional information to people writing clients of this class that the compiler has no way to enforce: namely that if you call any of the mutating methods, you're going to get a runtime exception. Ideally, we'd have a way for the compiler to enforce that, but we don't, so marking the type as immutable is better than nothing. I think that's generally a pretty compelling argument.

Of course internally, there's pretty much no way to compile a binary that doesn't have `com.google.common.collect` linked in, so the additional dependencies aren't a concern.

plumpy:
I think Guava in the past has modified the APIs, but they've taken a stance for a few years now that they absolutely won't make changes to APIs unless they're marked `@Beta`. If we're concerned about that, perhaps we turn on the [Guava Beta checker](https://github.com/google/guava-beta-checker)?

> APIs without @Beta will remain binary-compatible for the indefinite future. (Previously, we sometimes removed such APIs after a deprecation period. The last release to remove non-@Beta APIs was Guava 21.0.)

In practice, Guava feels about as close to a core library of Java as you can get without actually being in the JDK, since it's in the dependency tree of so many libraries already. It makes sense to keep it out of plugin APIs and the like, where we allow essentially _no_ dependencies, but I guess I don't really see the point otherwise. Are you mostly objecting to having it at the API boundaries?

(I guess it is true that `clouddriver-artifacts` needs to add an `api` dependency on it now...)

cfieber:
>  Are you mostly objecting to having it at the API boundaries?
> (I guess it is true that clouddriver-artifacts needs to add an api dependency on it now...)

This would be my primary concern yes.

> I think Guava in the past has modified the APIs, but they've taken a stance for a few years now that they absolutely won't make changes to APIs unless they're marked `@Beta.`

TIL, this is good news.

cfieber:
Thoughts on this?

> I think there are a bunch of cases in here where the intent would be clearer by translating empty strings to nulls in the value classes and having the getters exposed as Optionals instead.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactAccount.java
cfieber: Fair enough

ezimanyi:
Sorry should have added a comment, but the last commit I added makes that change to expose `Optional<String>` instead of a string with `""` having a special meaning.

PR #4587, opened 2020-05-12
feat(provider/aws): Encryption with specific KMS key id

odg0318:
AWS EBS can be encrypted by CMK and it is supported by AWS API.
But it is not implemented in spinnaker. https://github.com/spinnaker/spinnaker/issues/5745 is related issue.

EBS block storage can be configured by editing json object.
I added `kmsKeyId` in `AmazonBlockDevice` class and it is used to request AWS API to create launch configuration.

But I have no idea how to test this code block I added.
Please let me know if you have some ideas.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/DefaultLaunchConfigurationBuilder.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/DefaultLaunchConfigurationBuilder.groovy
@@ -86,7 +86,8 @@ class DefaultLaunchConfigurationBuilder implements LaunchConfigurationBuilder {
           deleteOnTermination: mapping.ebs.deleteOnTermination,
           iops: mapping.ebs.iops,
           snapshotId: mapping.ebs.snapshotId,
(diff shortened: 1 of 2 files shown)


odg0318:
@cfieber 

I was confused with AWS document about between EC2 and Autoscaling

BlockDevice in LaunchConfiguration doesn't support customer KMS key id yet.
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-launchconfig-blockdev-template.html
* https://docs.aws.amazon.com/cli/latest/reference/autoscaling/create-launch-configuration.html

BlockDevice in EC2 supports customer KMS key. 
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-template.html

I think we should wait for supporting AWS API implementation to LaunchConfiguration.

Please close this issue if my thought is right.

cfieber:
have you tried running this with an encrypted block device?

odg0318:
Honestly, I tried just to compile source code.

cfieber:
I'm not set up to verify this at all either. If you can figure out whether it will work or not (the docs are definitely confusing) please reopen and add some details.

Thanks!

PR #4588, opened 2020-05-12
fix(entitytags): Parallelize deletion of entity tags from front50

marchello2000:
This makes it run quite a bit faster, plus adding some logging message so one can follow how far along we are in the process.
(in my tests this makes the operation run about 5x faster)

### DIFF
--- a/clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
+++ b/clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
@@ -516,23 +516,36 @@ public class ElasticSearchEntityTagsProvider implements EntityTagsProvider {
                 entityTags ->
                     executor.submit(
                         () -> {
-                          AuthenticatedRequest.allowAnonymous(
-                              () -> front50Service.batchUpdate(entityTags));
+                          try {
+                            AuthenticatedRequest.allowAnonymous(
+                                () -> front50Service.batchUpdate(entityTags));
 
-                          log.info(
-                              "Deleted {} out of {} tags in namespace {}",
-                              countProcessed.addAndGet(entityTags.size()),
-                              entityTagsForNamespace.size(),
-                              namespace);
+                            log.info(
+                                "Deleted {} out of {} tags in namespace {}",
+                                countProcessed.addAndGet(entityTags.size()),
+                                entityTagsForNamespace.size(),
+                                namespace);
+                          } catch (Exception e) {
+                            log.error(
+                                "Failed to delete a batch of tags from front50 in namespace {}",
+                                namespace,
+                                e);
+                          }
                         }));
 
         try {
           executor.shutdown();
-          executor.awaitTermination(1, TimeUnit.HOURS);
+          executor.awaitTermination(15, TimeUnit.MINUTES);
           results.put("deletedFromSource", true);
         } catch (InterruptedException e) {
-          log.error("Failed to bulk remove tags from front50", e);
-          results.put("error", e);
+          String error =
+              String.format(
+                  "Failed to bulk remove tags from front50 in namespace %s due to timeout, please try again",
+                  namespace);
+
+          log.error(error, e);
+          results.put("error", error);
+          results.put("exception", e);
         }
       }
     }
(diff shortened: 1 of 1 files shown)


clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
```diff
@@ -496,8 +502,38 @@ public Map reconcile(String cloudProvider, String account, String region, boolea
       results.put("deletedFromElasticsearch", true);
 
       if (deleteFromSource) {
-        Lists.partition(entityTagsForNamespace, 50).forEach(front50Service::batchUpdate);
-        results.put("deletedFromSource", true);
+        ExecutorService executor =
+            Executors.newFixedThreadPool(
+                6,
```
cfieber, line 507: should this be a configurable option?

clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
```diff
@@ -496,8 +502,38 @@ public Map reconcile(String cloudProvider, String account, String region, boolea
...
+                    .setNameFormat(ElasticSearchEntityTagsProvider.class.getSimpleName() + "-%d")
+                    .build());
+
+        AtomicLong countProcessed = new AtomicLong();
+
+        Lists.partition(entityTagsForNamespace, 50)
+            .forEach(
+                entityTags ->
+                    executor.submit(
+                        () -> {
```
cfieber, line 518: it is probably worth a try/catch around the whole runnable block with some error logging if things fail

clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
```diff
@@ -496,8 +502,38 @@ public Map reconcile(String cloudProvider, String account, String region, boolea
...
+                          log.info(
+                              "Deleted {} out of {} tags in namespace {}",
+                              countProcessed.addAndGet(entityTags.size()),
+                              entityTagsForNamespace.size(),
+                              namespace);
+                        }));
+
+        try {
+          executor.shutdown();
+          executor.awaitTermination(1, TimeUnit.HOURS);
```
cfieber, line 531: this seems like a long upper bounds time to block on this request, is this close to on the order of how long cleanup takes?

clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
marchello2000: definitely, thanks!

clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
marchello2000: it's how long it used to take, now i think it should be much smaller (in my attempts/estimates <10 mins, i will reduce it to 15 mins and a note to the exception for the operator to retry )

clouddriver-elasticsearch/src/main/java/com/netflix/spinnaker/clouddriver/elasticsearch/model/ElasticSearchEntityTagsProvider.java
marchello2000: i was thinking about it, but since NFLX is pretty much the only company using entity tags and it's an "admin" operation I figured it wasn't worth the added config, what do you think?

cfieber: fair enough

PR #4596, opened 2020-05-14
feat(aws): Support launch templates in createServerGroup operation

jeyrschabu:
- adds support for creating an asg using a launch template

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
@@ -40,6 +40,7 @@ import com.netflix.spinnaker.clouddriver.aws.services.RegionScopedProviderFactor
 import com.netflix.spinnaker.clouddriver.data.task.Task
 import com.netflix.spinnaker.clouddriver.data.task.TaskRepository
 import com.netflix.spinnaker.clouddriver.model.ServerGroup
+import com.netflix.spinnaker.config.AwsConfiguration.DeployDefaults
 import com.netflix.spinnaker.kork.core.RetrySupport
 import groovy.util.logging.Slf4j
 
@@ -106,6 +107,7 @@ class AutoScalingWorker {
   private int maxInstances
   private int desiredInstances
 
+  private DeployDefaults deployDefaults
   private RegionScopedProviderFactory.RegionScopedProvider regionScopedProvider
 
   AutoScalingWorker() {
@@ -170,7 +172,7 @@ class AutoScalingWorker {
       settings = DefaultLaunchConfigurationBuilder.setAppSecurityGroup(
         subnetType,
         application,
-        regionScopedProvider.deployDefaults,
+        regionScopedProvider.getDeploymentDefaults(),
         regionScopedProvider.securityGroupService,
         settings
       )
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
@@ -23,10 +23,12 @@ import com.amazonaws.services.autoscaling.model.CreateAutoScalingGroupRequest
 import com.amazonaws.services.autoscaling.model.DescribeAutoScalingGroupsRequest
 import com.amazonaws.services.autoscaling.model.DescribeAutoScalingGroupsResult
 import com.amazonaws.services.autoscaling.model.EnableMetricsCollectionRequest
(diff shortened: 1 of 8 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
```diff
@@ -160,11 +164,28 @@ class AutoScalingWorker {
       blockDevices: blockDevices,
       securityGroups: securityGroups)
 
-    String launchConfigName = regionScopedProvider.getLaunchConfigurationBuilder().buildLaunchConfiguration(application, subnetType, settings, legacyUdf)
+    LaunchTemplateSpecification launchTemplateSpecification = null
+    String launchConfigName = null
+    if (setLaunchTemplate != null && setLaunchTemplate) {
```
cfieber, line 169:
```suggestion
    if (setLaunchTemplate) {
```
thanks groovy!

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
```diff
@@ -551,20 +563,21 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
   @PackageScope
   Collection<AmazonBlockDevice> buildBlockDeviceMappings(
     BasicAmazonDeployDescription description,
-    LaunchConfiguration sourceLaunchConfiguration
+    Object sourceLaunchSetting
```
cfieber, line 566: This loss of typing and subsequent duck typing of everything in here and in convertBlockDevices isn't that great for our future selfs and their groovy refactoring sanity.

could you make an interface for the things that need do exist on a sourceLaunchSetting (instanceType and List<BlockDeviceMapping> I think is all) and wrap the launchConfiguration/launchTemplate?

jeyrschabu: Agreed! This definitely felt dirty.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
```diff
@@ -97,6 +99,8 @@ class AutoScalingWorker {
   private List<AmazonBlockDevice> blockDevices
   private Map<String, String> tags
   private List<AmazonAsgLifecycleHook> lifecycleHooks
+  private Boolean setLaunchTemplate
```
caseyhebebrand, line 102: I just want to confirm that this flag is being managed by config, and not something that is expected from a user's create request.

jeyrschabu: So this here is request based to allow falling back to launch configs.
Your comment reminded me that this should default to true (overrideable per request).

But to your point, we should have a configurable default that would be flipped to true to enable the functionality for everyone and use this flag for overrides,
I will add that to this PR.

jeyrschabu: Actually, let me do that in a follow-up PR.

caseyhebebrand: ok sounds good, that was my main concern here. 

Sorry I should have been a bit more specific here. We talked about config being managed in one place for defaulting. This means that UI requests will not have this flag, and rely on the default value. API requests can override using `setLaunchTemplate`.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/services/RegionScopedProviderFactory.groovy
```diff
@@ -114,6 +114,14 @@ class RegionScopedProviderFactory {
       new DefaultLaunchConfigurationBuilder(getAutoScaling(), getAsgService(), getSecurityGroupService(), userDataProviders, localFileUserDataProperties, deployDefaults)
     }
 
+    LaunchTemplateService getLaunchTemplateService() {
+      return new LaunchTemplateService(amazonEC2)
+    }
+
+    AwsConfiguration.DeployDefaults getDeploymentDefaults() {
```
jeyrschabu, line 121: renamed because groovy was compiling the method to call itself instead of returning the property. Thanks Groovy

jeyrschabu: And was bombing silently

PR #4602, opened 2020-05-15
fix(aws): validate cfn template

ariadnarouco:
Changes enter a validation of the template before creating a change set or updating a stack. According to AWS SDK, it will check for template syntax only (json or yml). It will throw an exception if format is incorrect. Otherwise, validation just returns a ValidateTemplateResult.


xavileon:
@spinnaker/oss-approvers PTAL

ariadnarouco:
@spinnakerbot add-label "ready to merge"

ariadnarouco:
@spinnakerbot cherry-pick 1.20

PR #4607, opened 2020-05-18
fix(gce): Fix cluster caching on the SQL backend

plumpy:
This is fixed on master as #4570, but I don't think it's safe enough to cherry-pick that. This is a significantly hackier fix, but it's safer since it only affects users using the SQL backend, which apparently isn't working anyway.

This is against the release-1.20.x branch, and then I'll cherry-pick it back to 1.19 and 1.18.


plumpy:
@Mergifyio backport release-1.19.x release-1.18.x

PR #4608, opened 2020-05-18
fix(aws): delete some unnecessary code

plumpy:
It looks like you fixed these two TODOs in #3401, but missed TitusV2ClusterCachingAgent. Is this safe? Happy to roll back if you aren't sure, but I figured I'd clean this up while I was looking at it for #4570.

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
@@ -77,10 +77,10 @@ public class TitusV2ClusterCachingAgent
   static final java.util.Set<AgentDataType> types =
       Collections.unmodifiableSet(
           Stream.of(
+                  AUTHORITATIVE.forType(CLUSTERS.ns),
                   AUTHORITATIVE.forType(SERVER_GROUPS.ns),
                   AUTHORITATIVE.forType(APPLICATIONS.ns),
                   INFORMATIVE.forType(IMAGES.ns),
-                  INFORMATIVE.forType(CLUSTERS.ns),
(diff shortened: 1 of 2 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusV2ClusterCachingAgent.java
```diff
@@ -77,10 +77,10 @@
   static final java.util.Set<AgentDataType> types =
       Collections.unmodifiableSet(
           Stream.of(
+                  AUTHORITATIVE.forType(CLUSTERS.ns),
```
cfieber, line 80: I don't think this is correct

The issue is that no single cluster caching agent can mark itself Authoritative for clusters because clusters span regions/accounts

We should really just have a caching agent that looks at all Server groups in cache and returns the authoritative list of clusters

What this change will cause is strange behaviour if you delete the last server group in a region (the whole cluster will disappear and reappear in cache).

plumpy: Yeah, that's what happens for Applications right now too :(

plumpy: FYI, if you look at the `SqlProviderCache` code that I deleted, this is already functionally true for SQL... this only changes the behavior for Redis users. And since you're the only Titus users and you only use SQL, I don't think this will actually affect anything in reality.

I actually spent a bunch of time several months ago trying to fix this behavior by refactoring CATS, but the lack of tests (and the fact that a ton of it is Groovy) meant I never could convince myself that it wasn't going to break other cloud providers, so I gave up :(

cfieber:
I included the SqlProviderCache cleanup in #4615 - going to close this one in favor of the other

PR #4609, opened 2020-05-19
feat(titus): Adding support for command for titus jobs

aravindmd:
- Adding support for `COMMAND` which is supported in Titus along with `ENTRYPOINT` 
- making changes to deck after this is merged.

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/JobDescription.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/JobDescription.java
@@ -59,6 +59,7 @@ public class JobDescription {
   private Boolean inService;
 
   private String entryPoint;
+  private String cmd;
   private String iamProfile;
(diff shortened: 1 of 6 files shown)


aravindmd:
Getting some clarification from titus before merging this.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/JobDescription.java
```diff
@@ -540,6 +550,10 @@ public JobDescriptor getGrpcJobDescriptor() {
       containerBuilder.addEntryPoint(entryPoint);
     }
 
+    if (cmd != null) {
```
cfieber, line 553: @aravindmd I think this might just need to be a not null or empty check here

aravindmd: Thank @cfieber , looks like that is the case , i'll fix it.

PR #4610, opened 2020-05-19
fix(kubernetes): Restore adding a suffix to jobs

ezimanyi:
Fixes spinnaker/spinnaker#5767.  I will cherry-pick back to 1.20 once this is merged, as per the discussion at the Kubernetes SIG today.

* test(kubernetes): Slightly refactor run job test

  Pull a string to a variable to make the upcoming changes easier.

* fix(kubernetes): Restore adding a suffix to jobs

  In Spinnaker 1.20, we removed the behavior where Spinnaker would automatically add a random suffix to Kubernetes jobs before deploying them, in favor of having users set the metatdata.generateName field to have Kubernetes do this.

  This ended up being more of a breaking change than anticipated, particularly for users with a lot of preconfigured jobs, as by nature these jobs re-use the same name/namespace for jobs in different pipelines.

  While we would like to longer-term still push users towards using the Kubernetes-native API of generateName for this (and avoid introducing more Spinnaker-specific flags/terminology that users need to learn), it is clear that we need to provide a much smoother upgrade path than we initially did.

  This commit puts the new behavior behind a global flag,
  'kubernetes.jobs.appendSuffix'. When the flag is true (which will be the default), a random suffix will be added to all jobs as was the case in 1.19 and earlier.

  I propose the following rollout plan:
  * Remainder of 1.20 and all of 1.21. The flag defaults to 'true', and a suffix is added to all jobs when a name is specified. A warning is printed to the log upon adding this suffix.
  * 1.22: The flag defaults to false. A suffix is no longer added to jobs that are specified by name, but the flag can still be enabled to restore the old behavior.
  * 1.23: The flag is removed.

  Throughout this, jobs which specify a generateName will be passed through to Kubernetes as is. This allows an incremental upgrade process, where users can update all necessary manifests during 1.20 and 1.21 (as well as during 1.22 if they manually update the flag).

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/converter/job/KubernetesRunJobOperationConverter.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/converter/job/KubernetesRunJobOperationConverter.java
@@ -28,20 +28,27 @@ import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation;
 import com.netflix.spinnaker.clouddriver.security.AbstractAtomicOperationsCredentialsSupport;
 import com.netflix.spinnaker.clouddriver.security.ProviderVersion;
 import java.util.Map;
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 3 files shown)


ezimanyi:

PR #4614, opened 2020-05-20
fix(kubernetes/v1): Reloading namespaces during loadData since new namespaces does not reflect in HA enabled clouddriver.

luckoseabraham:
Reloading namespaces during loadData since new namespaces does not reflect in HA enabled clouddriver.

For HA enabled clouddriver, new loadbalancers created for a new namespace is not getting reflected since the namespaces are not reloaded. 
The on demand cache refresh from orca is being invoked to the clouddriver-rw service and the agent instances which are running in clouddriver-caching never get updated and the loadbalancers are not refreshed as well.

For server group agent, the namespace reload is invoked in loadData method here https://github.com/spinnaker/clouddriver/blob/release-1.20.x/clouddriver-kubernetes-v1/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/agent/KubernetesServerGroupCachingAgent.groovy#L257

 I understand v1 is deprecated, But our users are not yet ready to adopt the v2 provider. 
The only other option I can think of for this is disabling ha for clouddriver.

### DIFF
--- a/clouddriver-kubernetes-v1/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/agent/KubernetesLoadBalancerCachingAgent.groovy
+++ b/clouddriver-kubernetes-v1/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/provider/agent/KubernetesLoadBalancerCachingAgent.groovy
@@ -175,6 +175,7 @@ class KubernetesLoadBalancerCachingAgent extends KubernetesV1CachingAgent implem
 
   @Override
   CacheResult loadData(ProviderCache providerCache) {
+    reloadNamespaces()
     Long start = System.currentTimeMillis()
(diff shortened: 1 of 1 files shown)


luckoseabraham:
@lwander - Would love your thoughts on this.

luckoseabraham:
Thank you so much for the approval @ezimanyi  👍 
The perfect fix for this would be to have the ondemand cache refresh being pointed to clouddriver-caching instead of clouddriver-rw. But this would need extensive end to end testing.

And since the serverGroup caching works this way, I did not want to change the entire behavior for a deprecated feature.

Will wait for thoughts from @ethanfrogers

luckoseabraham:
@ethanfrogers  - Could you please take a look?

ethanfrogers:
LGMT! I agree with @ezimanyi - let's be careful about what we choose to bring into the V1 provider.

PR #4615, opened 2020-05-20
fix(cache): adds a proper authoritative clusters source for aws/titus

cfieber:
incorporates the SqlProviderCache clean up from #4608


plumpy:
Interesting... I wonder if you want to do the same thing for applications? It has the same issue here, where if you delete the last server group in a region, the entire application disappears until the next cache refresh. Probably happens a lot less often, though.

cfieber:
ahh yeah this approach would totally work for applications as well

PR #4617, opened 2020-05-20
fix(provider/kubernetes): Allow inheritance of KubectlJobExecutor

ncknt:
Making the constructor public would let the class be extended in a Spinnaker plugin (different class loader) and would allow users to use a different behavior when interacting with kubectl.

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutor.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutor.java
@@ -59,7 +59,7 @@ public class KubectlJobExecutor {
   private final Gson gson = new Gson();
 
   @Autowired
-  KubectlJobExecutor(
+  public KubectlJobExecutor(
       JobExecutor jobExecutor,
       @Value("${kubernetes.kubectl.executable:kubectl}") String executable,
       @Value("${kubernetes.o-auth.executable:oauth2l}") String oAuthExecutable) {


ncknt:
@maggieneterval Would you mind checking it out?

maggieneterval:
Hey @ncknt! Would love to understand a bit more about the specific goals of extending this class in a plugin. I'm nervous about committing to the current version of this class as a public API as @ezimanyi and I plan on [replacing Clouddriver's use of `kubectl` with the client library](https://github.com/spinnaker/spinnaker/issues/5643) within the next several months. That said, would be very interested in the functionality folks hope to add in plugins to help inform this refactor so hopefully it can cover most use cases and obviate the need for this type of open-ended extension.

ncknt:
What's the target release for https://github.com/spinnaker/spinnaker/issues/5643? I think most classes have public constructors so I don't take public constructors as a contract. An alternative may be to use an interface for "KubernetesOperations" in `KubernetesV2Credentials`. It could be KubectlJobExecutor today, Kubernetes Java client tomorrow. That said, I was also hoping to backport that change :)

My immediate use case is to delegate `kubectl` calls to an external service but I can see other use cases such as using different versions `kubectl` for different accounts.

maggieneterval:
> What's the target release for spinnaker/spinnaker#5643?

Would guess within the next 2-3 releases!

>  I think most classes have public constructors so I don't take public constructors as a contract. 

I'm not familiar with the guarantees we're providing to Spinnaker plugins users, but if the intent of this change is to permit inheritance in a plugin, is that not a contract? Or would it be acceptable if the plugin stopped working without notice upon upgrade to 1.23 or whenever we refactor or remove this class? Particularly in the Kubernetes provider, we've become more careful about disallowing inheritance by default, with the understanding that any classes designed for inheritance represent a public contract.

> An alternative may be to use an interface for "KubernetesOperations" in KubernetesV2Credentials. It could be KubectlJobExecutor today, Kubernetes Java client tomorrow. That said, I was also hoping to backport that change :)

> My immediate use case is to delegate kubectl calls to an external service but I can see other use cases such as using different versions kubectl for different accounts.

For what it's worth, I _think_ you can actually already provide your own `kubectl` executable per-account. I don't foresee us being able to backport any refactor adding additional extensibility to `KubectlJobExecutor` or `KubernetesV2Credentials` since these are not fixes, but I'm happy to add an agenda item for next Kubernetes SIG so we can discuss the best path forward here.

ncknt:
> I'm not familiar with the guarantees we're providing to Spinnaker plugins users, but if the intent of this change is to permit inheritance in a plugin, is that not a contract?
There are two modes for Spinnaker plugins: `@ExtensionPoint` which is an official contract and privileged plugins which can do Springy things but from a different classloader (which is the issue here). The former shouldn't change w/o a deprecation notice, but I expect it for the latter. 

> would it be acceptable if the plugin stopped working without notice upon upgrade to 1.23 or whenever we refactor or remove this class?
I'm fine having to change in 1.23 or whenever we have an abstraction for reaching out to Kubernetes.

I also think it won't impact anyone - at least not negatively!

maggieneterval:
> I'm fine having to change in 1.23 or whenever we have an abstraction for reaching out to Kubernetes.

Thanks for clarifying! While it's tempting to just merge this, as you've stated that you have no expectation it won't break down the line, I'd prefer to establish an explicit policy around what we consider extensible (even short-term), so would love to get the input of other contributors and users from the SIG before we merge this and possibly implicitly create a policy.

link108:
@maggieneterval thanks for commenting and sharing your concerns. I think this is an interesting problem with creating good abstractions, the problem being how to create a good abstraction. I don't think it's a good idea to attempt to create a solid abstraction (in the form of extension points) on a first pass. 

Once the extension point is being used, there is an expectation that the extension point will be stable. It's possible to create an abstraction here, but challenging without a deep understanding and experience. I believe its good to try it out and gain some experience, doing that once or twice will then inform on how to create solid abstractions. 

All that being said, it's basically a chicken and egg problem, the ideal state is to have extension points for everything we'd want to extend, but we're not there yet and I think this is how we get to that point. 

We added the ability to extend spinnaker via spring plugins, but that is more of an advanced use case that we try and discourage users from using, given the costs you mentioned earlier (code changing, plugins breaking). Please see https://www.spinnaker.io/guides/developer/plugin-creators/overview/#spring-plugins for more info, particularly the disadvantages section. To your point about plugins not working on a particular version of a service, that is something we're actively working on in the plugins working group (plugin service version constraints). 

Regarding a policy around what we consider extensible, I think it makes sense to  discuss this at the Platform SIG this Thursday, 5/20.

maggieneterval:
> Regarding a policy around what we consider extensible, I think it makes sense to discuss this at the Platform SIG this Thursday, 5/20.

Do you mean tomorrow, 5/21? I'm not sure I'll be able to make it, but wouldn't be opposed to that SIG providing general guidance on designing for extensibility (or not) in light of Spring plugins being one approach to Spinnaker plugin development.

That said, I think a discussion around the appropriate level of abstraction to expose in this particular case would be most useful at the Kubernetes SIG. Would love to better understand what types of external services there's appetite to plug into the provider. The means by which Clouddriver communicates with your cluster (kubectl, client library, some other hand-rolled service) is a pretty low-level aspect of the provider, so I think we'd want to consider more than one use case before we make even a preliminary decision on what level of abstraction, if any, we expose here.

ncknt:
@maggieneterval I'm happy to share my thoughts at the SIG (and change my code to use that abstraction when needed). All I want is to get unblocked without waiting 3 releases :)

link108:
@maggieneterval good catch, it is 5/21. And thank you for the clarification, I was thinking about a more general policy re: extensibility, totally agree, the Kuberentes SIG is the appropriate place to discuss the level of abstraction 👍

maggieneterval:
@ncknt @link108 The amount of back-and-forth for a seemingly small changes likely feels frustrating, so please know that it's not my intention to needlessly block you.

Would love guidance from the Platform SIG on a policy for general extensibility, especially if that policy is different from general Java best practice of only extending classes that were designed and documented for inheritance. I am unsure that adding `public` to a constructor constitutes sufficient design or documentation to consider this class responsibly extendable. If we merge this as-is, I'm afraid we are setting a dangerous precedent and will need to merge other, move extensive refactors in order to unblock plugin development. I'd prefer to have a clearer policy, preferably in line with Java best practices, instead of a handshake deal that you are okay with breaking changes.

Separately, would love to have a discussion at the Kubernetes SIG to assess community interest in extending this part of the provider, to help inform any design decisions we make while refactoring. Would prefer that we make the core provider do the right thing for most people most of the time, and make well-informed decisions about which parts of the provider are meant to be extensible based on a variety of use cases. I've added a note to the agenda for next SIG.

ncknt:
@maggieneterval no worries, it makes sense. I actually would like to discuss a better integration within the Kubernetes provider.

ncknt:
We discussed the change in the SIG this morning. We agreed that only plugin `ExtensionPoints` are contracts. However, it'd be nice to not lock things down and let users extend certain aspects of implementation via interfaces as long as it doesn't hinder future improvements/refactor.

I'm closing that PR and will consider an interface based solution. Thanks @maggieneterval !

PR #4624, opened 2020-05-26
fix(provider/cf): fix buildOnDemandCacheData

zachsmith1:
CFSGcachingAgent was using the buildOnDemandCacheData method from the abstract class it extends from. The object mapper that was used in the abstract class doesn't serialize the object as we need it because it only serializes fields based on the view. This creates a problem when trying to serialize a CacheData object. The fix is to override this method and use the same logic, but use the objectMapper thats currently being used in the serverGroupCachingAgent class. The additional test covers and ensures that the ondemand cache object has the servergroup attributes it should have. This fix will also ensure that on demand cache is actually processed and not just skipped over because the resulting fields are null


zachsmith1:

kevinawoo:
@Mergifyio backport release-1.20.x release-1.19.x

PR #4631, opened 2020-05-28
fix(entity tags): Restore NotFoundException handling

marchello2000:
After https://github.com/spinnaker/clouddriver/pull/4325
Retrofit 404's are now converted to NotFoundException, this made the entitytag cleaner on ASG creation not handle the case when no entity tag exists.


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/DefaultOrchestrationProcessor.groovy
```diff
@@ -118,6 +118,7 @@ class DefaultOrchestrationProcessor implements OrchestrationProcessor {
                   try {
                     it.handle(event)
                   } catch (e) {
+                    log.warn("Error handling event (${event}): ${atomicOperation.class.simpleName}", e)
```
marchello2000, line 121: threw this in here since it was annoying to trace the source of error without a call stack

PR #4632, opened 2020-05-29
fix(aws/cache): index authoritative clusters by application

cfieber:
Adds an application attribute to authoritative clusters so that the lookup by application succeeds.

Also includes relationships from the cluster to the server groups we are crawling

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/agent/CatsClusterCachingAgent.java
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/agent/CatsClusterCachingAgent.java
@@ -29,8 +29,9 @@ import com.netflix.spinnaker.cats.cache.DefaultCacheData;
 import com.netflix.spinnaker.cats.provider.ProviderCache;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/agent/CatsClusterCachingAgent.java
```diff
@@ -93,25 +99,30 @@ public CacheResult loadData(ProviderCache providerCache) {
...
-                  return new DefaultCacheData(
-                      s,
-                      Collections.singletonMap("name", clusterAttributes.get(clusterNameAttribute)),
-                      Collections.emptyMap());
+                entry -> {
+                  Map<String, String> clusterAttributes = keyParser.apply(entry.getKey());
+                  Map<String, Object> cacheAttributes = new HashMap<>();
+                  cacheAttributes.put("name", clusterAttributes.get(clusterNameAttribute));
+                  cacheAttributes.put(
+                      "application", clusterAttributes.get(clusterApplicationAttribute));
```
cfieber, line 119: this specifically was necessary because we use the getAllByApplication lookup to find the cluster objects

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonClusterProvider.groovy
```diff
@@ -203,9 +203,13 @@ class AmazonClusterProvider implements ClusterProvider<AmazonCluster>, ServerGro
       allImages
     )
 
-    Collection<AmazonCluster> clusters = cacheResults[CLUSTERS.ns].collect { clusterData ->
+    Collection<AmazonCluster> clusters = cacheResults[CLUSTERS.ns].findResults { clusterData ->
       Map<String, String> clusterKey = Keys.parse(clusterData.id)
 
+      if (clusterKey == null) {
```
cfieber, line 209: now titus clusters are also indexed by application and end up in the results. The key parser returns null for a key from a different provider so we'll filter them out in here

(the same wasn't necessary in TitusClusterProvider because it doesn't use getAllByApplication - though maybe it should...)

PR #4638, opened 2020-05-30
refactor(artifacts): Remove mutations of artifacts

ezimanyi:
In an upcoming kork release, artifacts will become immutable to prevent confusing bugs where (primarily in orca) we are mutating copies of what we think we are mutating. There are a small number of places in clouddriver where artifacts are mutated, but these can be replaced by returning a copy with the desired changed.


ezimanyi:
There are two places where we call `artifact.getMetadata()` then mutate that resulting map that I'll clean up separately, but this eliminates all places where we directly mutate a field of an artifact.

ezimanyi:
@zachsmith1 : Mind taking a look at the couple of CloudFoundry changes here?  They seem pretty safe but would appreciate a set of eyes from someone more familiar with CloudFoundry.  Thanks!

zachsmith1:
@ezimanyi LGTM!

PR #4642, opened 2020-06-01
feat(saga): Support awaiting cloud state conditions before proceding

robzienert:
Continued effort to remove cloud provider specific logic in Orca. I'm going through Orca trying to find all areas of `orca-clouddriver` where we do conditional logic based on which cloud provider we're dealing with. My end goal is that Orca does not contain any cloud provider logic and just passes the heavy lifting down to clouddriver.

Right now I'm working on [ResizeServerGroupStage](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/servergroup/ResizeServerGroupStage.groovy). I don't really understand `pre/postStatic` or `pre/postDynamic`, but the general flow is when this stage starts, if we're dealing with AWS, wrap the resize operation in suspend / resume scaling processes so that autoscaling doesn't go trampling.

This seems like just the perfect thing for clouddriver to do:

1. Suspend scaling processes.
2. Set the new capacity for the server group.
3. Wait for the desired capacity to equal actual capacity.
4. Resume scaling processes.

In order for Clouddriver to take this entire flow over, it needs to be able to wait for state, so I added `await` to the Saga code to provide this basic utility. Essentially, this would then allow us to express `ResizeAsgAtomicOperation` in the AWS provider as a `Saga` that looks something like so:

```java
new SagaFlow()
  .then(SuspendScalingProcesses.class)
  .then(ResizeServerGroup.class)
  // condition: Function<Saga, Boolean>, ttl: Duration, interval: Duration
  .await(WaitForDesiredCapacityMatch.class, Duration.ofMinutes(5), Duration.ofSeconds(30))
  .then(ResumeScalingProcesses.class);
```

In doing this, the `FlowIterator` would then periodically invoke `WaitForDesiredCapacityMatch` on `interval` (30 seconds) to check the server group state. Once the state matches, we can move on with life.

Doing this work would allow us to have a totally cloud provider agnostic resize stage.

### DIFF
--- a/clouddriver-saga-test/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/AbstractSagaTest.kt
+++ b/clouddriver-saga-test/src/main/kotlin/com/netflix/spinnaker/clouddriver/saga/AbstractSagaTest.kt
@@ -21,6 +21,7 @@ import com.netflix.spinnaker.clouddriver.saga.persistence.SagaRepository
 import dev.minutest.junit.JUnit5Minutests
 import io.mockk.every
 import io.mockk.mockk
+import java.time.Clock
 import org.springframework.context.ApplicationContext
 
 abstract class AbstractSagaTest : JUnit5Minutests {
@@ -52,7 +53,7 @@ abstract class AbstractSagaTest : JUnit5Minutests {
       }
       registerBeans(applicationContext, *options.registerTypes.toTypedArray())
 
-      sagaService = SagaService(sagaRepository, NoopRegistry()).apply {
+      sagaService = SagaService(sagaRepository, NoopRegistry(), Clock.systemDefaultZone()).apply {
         setApplicationContext(applicationContext)
       }
     }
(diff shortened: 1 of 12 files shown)


cfieber:
I believe our resize in orca waits taking into account instance health, and all the logic for that currently lives in orca.

I think we technically have all the information to make that decision in clouddriver, but the duration of the saga execution could become quite long so we need to consider how happy we are about that (right now the failover task stuff @dreynaud was working on kind of assumes a short-ish lived task execution)

We could potentially introduce a 'complete operation' call from orca to river that we issue as a way of offloading cloudprovider specifics back to river, but keep the long term task execution in orca.

e.g.
```
orca                                                            river/aws
resize stage:
cloud provider task (resize server group)             ---->     suspend scaling
                                                                resize asg
monitor kato task

wait for capacity match task

complete cloud provider task (resize server group)    ---->     resume scaling
monitor kato task
```

I don't know that I like it but it does keep the longer running bit orchestrated by orca

what do you think?

robzienert:
Yeah, I think something like this could work. I presume in this scenario, `complete cloud provider task` would be blindly fired regardless of the cloud provider?

cfieber:
yeah I think so, and probably noop most of the time I guess

PR #4644, opened 2020-06-02
test(provider/cf): Add test on cf deploy converter

ezimanyi:
The recent change to the artifact handling in cloudfoundry deploy may have caused a regression. It's still not clear to me exactly what the change is, but adding this (currently passing) unit test as a start so that hopefully we can make it fail and track the bug.


ezimanyi:
@zachsmith1 : In response to the issue you reported about there being a null uid on an artifact, I wrote this test to validate that the logic is correct.  This test passes both before and after my changes from earlier today.  Obviously that's not helpful in actually finding the bug, but hopefully with this passing test as a start it will be easier to update the test to cover whatever specific case is breaking.

ezimanyi:
Chatted with @zachsmith1 : Turns out that my change didn't cause the issue, it was just bad data in the cache unrelated to the artifact change.  I may as well submit this test anyway as extra test coverage can't hurt; as @zachsmith1 is not an approver (yet!), @maggieneterval do you mind taking a quick look?  Thanks!

PR #4648, opened 2020-06-02
fix(tencent): Fix compilation error due to import order

ezimanyi:
This file is not compiling for my version of the JDK (openjdk 1.8.0_242) because we are importing the inner class before we import the Data annotation that the class is annotated with. It looks like this may have been fixed in later versions of the JDK, but let's avoid others running into this issue by fixing it.


ezimanyi:
FYI @shengyu 
This is fixing an interesting import-order issue with #4603 for some versions of Java.

PR #4650, opened 2020-06-03
fix(sql): Store dates as string instead of maps

german-muzquiz:
Fixes https://github.com/spinnaker/spinnaker/issues/5706

Modifies the way dates (like java.time.Instant) are stored in sql, storing them as strings instead of a map.

Example of docker image date stored in redis: `"date": "2017-06-23T00:57:26.841792489Z"`

Example of docker image date stored in sql before this change: `"date": { "nano": 499754581, "epochSecond": 1591115020 }`

After the change in sql: `"date":"2020-06-02T16:23:45.284504313Z"`

### DIFF
--- a/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlCacheConfiguration.kt
+++ b/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlCacheConfiguration.kt
@@ -1,8 +1,6 @@
 package com.netflix.spinnaker.config
 
 import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.databind.SerializationFeature
-import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
 import com.netflix.discovery.EurekaClient
 import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.cats.agent.AgentScheduler
@@ -102,6 +100,7 @@ class SqlCacheConfiguration {
     cacheMetrics: SqlCacheMetrics,
     dynamicConfigService: DynamicConfigService,
     sqlConstraints: SqlConstraints,
+    mapper: ObjectMapper,
     @Value("\${sql.cache.async-pool-size:0}") poolSize: Int,
     @Value("\${sql.table-namespace:#{null}}") tableNamespace: String?
   ): NamedCacheFactory {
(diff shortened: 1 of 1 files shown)


german-muzquiz:
cc @maggieneterval

maggieneterval:
🎉 Thanks so much, @german-muzquiz! LGTM but deferring to SQL master @robzienert for final merge.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlCacheConfiguration.kt
```diff
@@ -124,9 +126,12 @@ class SqlCacheConfiguration {
       log.info("Configured coroutine context with newFixedThreadPoolContext of $poolSize threads")
     }
 
+    // stores dates like java.time.Instant as a String with the format "2017-04-06T16:29:44.947869272Z" instead of a Map
+    val mapper = ObjectMapper().registerModule(JavaTimeModule()).disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
+
     return SqlNamedCacheFactory(
       jooq,
-      ObjectMapper(),
+      mapper,
```
robzienert, line 134: Why not use the object mapper that's already in the application context?

german-muzquiz: I just wanted to avoid other changes pulled from that one that we may not know about into a working sql cache, but I see the point of standardizing on one object mapper config for easier maintenance.

If I read correctly, the mapper in the application context is built from these settings: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/CloudDriverConfig.groovy#L132 . I did a test with that one and the date format is also fixed. Do you think it's safe to use it?

german-muzquiz: Checking in, @robzienert should we use the mapper from the application context?

robzienert: It should be changed. Having different object mappers isn't only inefficient, it can be the source of bugs if one object mapper serializes things while another, differently configured one, deserializes it.

robzienert: Change can be made here: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/CloudDriverConfig.groovy#L132

german-muzquiz:
Using `ObjectMapper` from application context, which already works for storing dates with the same format as redis without any configuration change.

dbyron0:
cherry-pick to 1.19.x since https://github.com/spinnaker/spinnaker/issues/5706 was an issue in 1.19.6?

PR #4652, opened 2020-06-03
feat(aws): Only collect names of security groups in AWS deploy description

jonsie:


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/deploy/DescriptionAuthorizer.java
```diff
@@ -110,7 +114,8 @@ public void authorize(T description, Errors errors) {
     if (description instanceof ResourcesNameable) {
       ResourcesNameable resourcesNameable = (ResourcesNameable) description;
 
-      if (!resourcesNameable.requiresAuthorization()) {
+      if (!resourcesNameable.requiresAuthorization()
+          || dynamicConfigService.isEnabled("aws.fiat.authorize.resources-nameable", false)) {
```
jonsie, line 118: I also added a (hopefully temporary) dynamic config here so we can quickly disable this if we need to.

PR #4653, opened 2020-06-04
fix(provider/cf): fix and finish implementing fiat permissions for accounts and applications

zachsmith1:
CF permissions were never finished, however pieces of it were complete. This fix finishes implementing permissions for account level and application level access. Luckily, the core clouddriver security libraries make it easy to implement by interface so we didn't have to include much logic at all. Cloudfoundry also only supports frigga-based naming conventions today. There is some new logic around deriving the spinnaker app name from servergroups and clusters. We have created a ticket to fix this in a future release and sync the conventions with moniker going forward.


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/description/AbstractCloudFoundryServerGroupDescription.java
```diff
@@ -16,13 +16,30 @@
...
+
+  private String cluster;
+  private Moniker moniker;
+
+  @Override
+  public Collection<String> getApplications() {
+    if (moniker != null) {
+      return Collections.singletonList(moniker.getApp());
+    } else if (cluster != null) {
+      return Collections.singletonList(cluster.split("-")[0]);
```
danielpeach, line 41: Would use Frigga library for this. If for no other reason, it will make it more obvious what's going on here.

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/description/AbstractCloudFoundryServerGroupDescription.java
```diff
@@ -16,13 +16,30 @@
...
+    }
+    return Collections.singletonList(serverGroupName.split("-")[0]);
```
danielpeach, line 43: same here

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/description/AbstractCloudFoundryServerGroupDescription.java
zachsmith1: @danielpeach Thanks for pointing the frigga library out to me! Done

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/description/AbstractCloudFoundryDescription.java
```diff
@@ -18,11 +18,20 @@
...
+public abstract class AbstractCloudFoundryDescription implements AccountNameable {
   @JsonIgnore private CloudFoundryClient client;
 
   private String region;
+
+  @JsonIgnore private CloudFoundryCredentials credentials;
+
+  @Override
+  public String getAccount() {
+    return credentials.getName();
```
german-muzquiz, line 35: Safety check: check if `credentials` is null before using it, to prevent `NullPointerException` if `getAccount` is ever called before setting the `credentials` object.

zachsmith1: @german-muzquiz Added some logic here for the safety check!

maggieneterval:
Hi @zachsmith1! Looks like we need to make a corresponding update in Halyard, which tries to create a `CloudFoundryCredentials`. See failing build on this autobump PR: https://github.com/spinnaker/halyard/pull/1702/checks?check_run_id=739732167. I'll also tag you on the Halyard PR.

PR #4654, opened 2020-06-04
feat(titus): Lookup server group names when converting TitusDeployDescription

jonsie:
-Additionally, removes requiresAuthorization from `ResourcesNameable` and removes `BasicAmazonDeployDescription` implementation of `ResourcesNameable`.  We will instead be requiring authorization on these resources via an implementation of `AllowedAccountsValidator` which runs prior to `DescriptionAuthorizer`.


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/converters/TitusDeployAtomicOperationConverter.groovy
```diff
@@ -21,13 +21,18 @@ import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation
...
 import com.netflix.spinnaker.clouddriver.titus.deploy.description.TitusDeployDescription
+import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 
 @TitusOperation(AtomicOperations.CREATE_SERVER_GROUP)
 @Component
 class TitusDeployAtomicOperationConverter extends AbstractAtomicOperationsCredentialsSupport {
 
+  @Autowired
+  AwsLookupUtil awsLookupUtil
```
robzienert, line 34: Use a constructor instead.

PR #4655, opened 2020-06-04
fix(ecs): set IAMRoleCachingAgent AWS region

allisaurus:
Fixes https://github.com/spinnaker/spinnaker/issues/5710

Previously, the default region was always being used, which results in the wrong endpoint being used for China regions.

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/TaskDefinitionCacheClient.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/TaskDefinitionCacheClient.java
@@ -49,6 +49,7 @@ public class TaskDefinitionCacheClient extends AbstractCacheClient<TaskDefinitio
     taskDefinition.setTaskRoleArn((String) attributes.get("taskRoleArn"));
     taskDefinition.setCpu((String) attributes.get("cpu"));
     taskDefinition.setMemory((String) attributes.get("memory"));
+    taskDefinition.setNetworkMode((String) attributes.get("networkMode"));
 
     if (attributes.containsKey("containerDefinitions")) {
       List<Map<String, Object>> containerDefinitions =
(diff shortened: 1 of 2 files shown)


clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/IamRoleCachingAgent.java
```diff
@@ -83,8 +83,13 @@ public IamRoleCachingAgent(
 
   @Override
   public CacheResult loadData(ProviderCache providerCache) {
-    AmazonIdentityManagement iam =
-        amazonClientProvider.getIam(account, Regions.DEFAULT_REGION.getName(), false);
+    // use a region from the account for correct endpoint resolution
+    String region =
+        !account.getRegions().isEmpty()
+            ? account.getRegions().get(0).getName()
```
pkandasamy91, line 89: Are we assuming that there's only 1 region? Do we only want the first region name?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/IamRoleCachingAgent.java
```diff
@@ -83,8 +83,13 @@ public IamRoleCachingAgent(
 
   @Override
   public CacheResult loadData(ProviderCache providerCache) {
-    AmazonIdentityManagement iam =
-        amazonClientProvider.getIam(account, Regions.DEFAULT_REGION.getName(), false);
+    // use a region from the account for correct endpoint resolution
+    String region =
+        !account.getRegions().isEmpty()
```
pkandasamy91, line 88: you should also check that account.getRegions.get(0) != null (account.getRegions.get(0).getName() will throw an NPE otherwise)

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/IamRoleCachingAgent.java
allisaurus: IAM resources are global on the account, so each regional call should return the same data.

PR #4661, opened 2020-06-05
fix(provider/cf): fix stop server group state checking

zachsmith1:
StopServerGroup operation is only checking that the state is != STARTING. It should also include a != RUNNING since this is one of the outcomes for state. This creates a bug where the operationsPoller completes successfully because the state is != STARTING. When the state is RUNNING, it will proceed and an exception will be thrown. This fix will allow CF enough time to stop the application and correctly show state.


german-muzquiz:

kevinawoo:

PR #4666, opened 2020-06-10
fix(amazon): Don't overwrite the LB target group attribute values during upsert operation

aravindmd:
- Don't overwrite the ppv2 value if its already set , update the value to what is already present.

I don't have the complete context on all the bits here , but this seemed like a quick way to fix it . Please take a look.


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -44,45 +52,72 @@ class LoadBalancerV2UpsertHandler {
...
-        Integer deregistrationDelay = [attributes.deregistrationDelay, deployDefaults?.loadBalancing?.deregistrationDelayDefault].findResult(Closure.IDENTITY)
-        if (deregistrationDelay != null) {
-          targetGroupAttributes.add(new TargetGroupAttribute(key: "deregistration_delay.timeout_seconds", value: deregistrationDelay.toString()))
-        }
-        if (loadBalancer.type == 'application') {
-          if (attributes.stickinessEnabled != null) {
-            targetGroupAttributes.add(new TargetGroupAttribute(key: "stickiness.enabled", value: attributes.stickinessEnabled.toString()))
+
+    //Create Target Group Attributes with values provided in description, set to defaults other wise
+    if(deployDefaults !=null) {
```
cfieber, line 58: why is this conditional on deployDefaults?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -44,45 +52,72 @@ class LoadBalancerV2UpsertHandler {
...
+
+            }
           }
         }
-        if (loadBalancer.type == 'network' ) {
-          if(attributes.proxyProtocolV2 != null) {
-            boolean currentProxyProtocolV2 = currentTargetGroupAttributes.find { it.key == ATTRIBUTE_PROXY_PROTOCOL_V2_ENABLED }?.getValue().toBoolean()
-            if (currentProxyProtocolV2 != null && currentProxyProtocolV2 != attributes.proxyProtocolV2) {
-              targetGroupAttributes.add(new TargetGroupAttribute(key: "proxy_protocol_v2.enabled", value: currentProxyProtocolV2))
+    else {
```
cfieber, line 89: closing `}` should be on this line

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -44,45 +52,72 @@ class LoadBalancerV2UpsertHandler {
...
+      // Modify Target Group Attributes
+      if (attributes) {
```
cfieber, line 91: attributes may be not-null with all the values null, you might need to track that you've actually added something to targetGroupAttributes before issuing the command to update

cfieber:
the new approach seems more in line with what I'd expect to see as behaviour

I'd like to see a test case added that exercises this specifically

I'd also like some confirmation (that will require manual testing) that calling modifyTargetGroupAttributes with just the subset of attributes is non-destructive to the other attributes (otherwise we may have to describe existing values and supply those if the request attribute is null)

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
aravindmd: so from what i see creation of TargetGroup calls `modifyTargetGroupAttributes` with the deployDefaults and updateTargetGroups calls without , i used that to differentiate between create vs modify 
Modify Target Group -> https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy#L201

Create Target Group -> https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy#L148

Do you think that's reasonable  approach ?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
aravindmd: will fix this

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
aravindmd: sounds good , will fix that

aravindmd:
> the new approach seems more in line with what I'd expect to see as behaviour
> 
> I'd like to see a test case added that exercises this specifically
> 
> I'd also like some confirmation (that will require manual testing) that calling modifyTargetGroupAttributes with just the subset of attributes is non-destructive to the other attributes (otherwise we may have to describe existing values and supply those if the request attribute is null)

Thanks for looking Cam, I have manually tested the case you mentioned , updating a few of the attributes does not update the other attributes that are already set , i did that for NLB and ALB type.

So the tests ( create and update operations ) we have here is already excising this path, anything specific that you feel is missing https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/UpsertAmazonLoadBalancerV2AtomicOperationSpec.groovy

aravindmd:
Pleasetake a look and let me know i have addressed all your comments.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -34,6 +34,15 @@ class LoadBalancerV2UpsertHandler {
...
   private static final String ATTRIBUTE_DELETION_PROTECTION = "deletion_protection.enabled"
 
+  //Defaults for Target Group Attributes
+  private static final String DE_REGISTRATION_DELAY = "300"
+  private static Boolean STICKINESS_ENABLED = false
+  private static String STICKINESS_TYPE = "lb_cookie"
+  private static String STICKINESS_DURATION = "86400"
+  private static Boolean PROXY_PROTOCOL_V2 = false
+  /** The following attribute is supported only if the target is a Lambda function. */
+  private static Boolean MULTI_VALUE_HEADERS_ENABLED = false
```
jeyrschabu, line 44:
```suggestion
  private static final Boolean STICKINESS_ENABLED = false
  private static final String STICKINESS_TYPE = "lb_cookie"
  private static final String STICKINESS_DURATION = "86400"
  private static final Boolean PROXY_PROTOCOL_V2 = false
  /** The following attribute is supported only if the target is a Lambda function. */
  private static final Boolean MULTI_VALUE_HEADERS_ENABLED = false
```

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -42,46 +51,81 @@ class LoadBalancerV2UpsertHandler {
...
-        Integer deregistrationDelay = [attributes.deregistrationDelay, deployDefaults?.loadBalancing?.deregistrationDelayDefault].findResult(Closure.IDENTITY)
-        if (deregistrationDelay != null) {
-          targetGroupAttributes.add(new TargetGroupAttribute(key: "deregistration_delay.timeout_seconds", value: deregistrationDelay.toString()))
-        }
-        if (loadBalancer.type == 'application') {
-          if (attributes.stickinessEnabled != null) {
-            targetGroupAttributes.add(new TargetGroupAttribute(key: "stickiness.enabled", value: attributes.stickinessEnabled.toString()))
+
+    //Create Target Group Attributes with values provided in description, set to defaults other wise
+    if(deployDefaults !=null) {
```
jeyrschabu, line 58: The check on `deployDefaults` and the subsequent `else` branch isn't necessary?

You have already removed default values from `UpsertAmazonLoadBalancerV2Description.TargetGroup.Attributes` , shouldn't we just be able to ensure  that those values are accounted for here?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -42,46 +51,81 @@ class LoadBalancerV2UpsertHandler {
...
     }
-
-    try {
-      loadBalancing.modifyTargetGroupAttributes(new ModifyTargetGroupAttributesRequest()
-        .withTargetGroupArn(targetGroup.targetGroupArn)
-        .withAttributes(targetGroupAttributes))
-      task.updateStatus BASE_PHASE, "Modified target group ${targetGroup.targetGroupName} attributes."
-    } catch (AmazonServiceException e) {
-      return handleError("Failed to modify attributes for target group ${targetGroup.targetGroupName} - reason: ${e.toString()}.", e)
+    if(targetGroupAttributes.size()!=0) {
```
jeyrschabu, line 119: Is there a legit case where `targetGroupAttributes` size is 0?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
cfieber: I think the conditional makes sense to account for a create vs an upsert

During a create, we want to supply the default value for all the unspecified attributes
During an upsert, we only want to modify an attribute if it was specifically part of the request, and anything that is null should not be modified

jeyrschabu: Yeah had a chat with @aravindmd this morning. I agree with having a check here but not necessarily `deployDefaults` since it has the potential to be misleading to someone looking at this in the future. I suggested either passing a boolean or split the method into something that would go through the create path and another through the update. It's more of an esthetics opinion. If you don't share that concern, we can keep the check on deplyDefaults as the differentiator between create and update.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
cfieber: I feel like adding a specific flag to indicate that it is an update vs a create would be better - deployDefaults being null doesn't feel like a super reliable contract to base this behaviour off of and feels ripe for a regression

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
aravindmd: 👍 fixed

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
aravindmd: changed the call to two different methods to make it clear

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
aravindmd: just added that for additional safety

aravindmd:
Updated to use two different methods and added some tests to validate the attributes set.

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/UpsertAmazonLoadBalancerV2AtomicOperationSpec.groovy
```diff
@@ -209,6 +343,42 @@ class UpsertAmazonLoadBalancerV2AtomicOperationSpec extends Specification {
...
+    1 * loadBalancing.describeLoadBalancers(new DescribeLoadBalancersRequest(names: ["foo-main-frontend"])) >>
+      new DescribeLoadBalancersResult(loadBalancers: existingLoadBalancers)
+    1 * loadBalancing.setSecurityGroups(new SetSecurityGroupsRequest(
+      loadBalancerArn: loadBalancerArn,
+      securityGroups: ["sg-1234"]
+    ))
+    1 * loadBalancing.describeTargetGroups(new DescribeTargetGroupsRequest(loadBalancerArn: loadBalancerArn)) >> new DescribeTargetGroupsResult(targetGroups: existingTargetGroups)
+    1 * loadBalancing.modifyTargetGroup(_ as ModifyTargetGroupRequest)
+    1 * loadBalancing.modifyTargetGroupAttributes(_) >> { ModifyTargetGroupAttributesRequest request ->
+      assert request.attributes.find { it.key == 'deregistration_delay.timeout_seconds'}.value == "300"
```
aravindmd, line 371: assert only attributes sent in the description set are present during the update call

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -34,18 +34,61 @@ class LoadBalancerV2UpsertHandler {
...
   private static Task getTask() {
     TaskRepository.threadLocalTask.get()
   }
 
-  private static String modifyTargetGroupAttributes(AmazonElasticLoadBalancing loadBalancing, LoadBalancer loadBalancer, TargetGroup targetGroup, UpsertAmazonLoadBalancerV2Description.Attributes attributes) {
-    return modifyTargetGroupAttributes(loadBalancing, loadBalancer, targetGroup, attributes, null)
+  //Create Target Group Attributes with values provided in description, set to defaults other wise
+  static String createTargetGroupAttributes(AmazonElasticLoadBalancing loadBalancing, LoadBalancer loadBalancer, TargetGroup targetGroup, UpsertAmazonLoadBalancerV2Description.Attributes attributes, DeployDefaults deployDefaults) {
+    def targetGroupAttributes = []
+    log.info("Creating target group attributes for targetGroup {} with attributes {} ", targetGroup.targetGroupName, attributes)
```
jeyrschabu, line 53: This won't pretty-print `attributes` if it doesn't have a `toString()` on it or some lombok annotation.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -34,18 +34,61 @@ class LoadBalancerV2UpsertHandler {
...
+
+      }
+    }
+    return updateTargetGroupAttributes(loadBalancing, targetGroup, targetGroupAttributes)
   }
+
+  // Modify target group attributes with attributes that are set in the description , do not update attributes that are not set
   private static String modifyTargetGroupAttributes(AmazonElasticLoadBalancing loadBalancing, LoadBalancer loadBalancer, TargetGroup targetGroup, UpsertAmazonLoadBalancerV2Description.Attributes attributes, DeployDefaults deployDefaults) {
+
+    log.info("Update target group attributes for targetGroup {} with attributes {} ", targetGroup.targetGroupName, attributes)
```
jeyrschabu, line 88: Same printing comment as above.

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/UpsertAmazonLoadBalancerV2AtomicOperationSpec.groovy
```diff
@@ -184,6 +250,74 @@ class UpsertAmazonLoadBalancerV2AtomicOperationSpec extends Specification {
...
+      loadBalancerArn: loadBalancerArn,
+      securityGroups: ["sg-1234"]
+    ))
+    1 * loadBalancing.describeTargetGroups(new DescribeTargetGroupsRequest(loadBalancerArn: loadBalancerArn)) >> new DescribeTargetGroupsResult(targetGroups: existingTargetGroups)
+    1 * loadBalancing.createTargetGroup(_ as CreateTargetGroupRequest) >> new CreateTargetGroupResult(targetGroups: [targetGroup])
+    1 * loadBalancing.describeListeners(new DescribeListenersRequest(loadBalancerArn: loadBalancerArn)) >> new DescribeListenersResult(listeners: existingListeners)
+    1 * loadBalancing.createListener(new CreateListenerRequest(loadBalancerArn: loadBalancerArn, port: 80, protocol: "HTTP", defaultActions: [new Action(targetGroupArn: targetGroupArn, type: ActionTypeEnum.Forward, order: 1)]))
+    1 * loadBalancing.describeLoadBalancerAttributes(_) >> [attributes: loadBalancerAttributes]
+    1 * loadBalancing.modifyTargetGroupAttributes(_) >> { ModifyTargetGroupAttributesRequest request ->
+      assert request.attributes.find { it.key == 'deregistration_delay.timeout_seconds'}.value == "300"
```
jeyrschabu, line 275: nit: these can all be asserted as part of the same `request.attributes.find`

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/UpsertAmazonLoadBalancerV2AtomicOperationSpec.groovy
jeyrschabu: similar nit comment.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
aravindmd: 👍 fixed

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -36,87 +36,90 @@ class LoadBalancerV2UpsertHandler {
 
   //Defaults for Target Group Attributes
   private static final String DE_REGISTRATION_DELAY = "300"
```
cfieber, line 38: nit, but `DEREGISTRATION_DELAY` would align better with the attribute name

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/LoadBalancerV2UpsertHandler.groovy
```diff
@@ -36,87 +36,90 @@ class LoadBalancerV2UpsertHandler {
...
           }
         }
       }
     }
-    if(targetGroupAttributes.size()!=0) {
+    return updateTargetGroupAttributes(loadBalancing, targetGroup, targetGroupAttributes)
+  }
+
+  static String updateTargetGroupAttributes(AmazonElasticLoadBalancing loadBalancing, TargetGroup targetGroup, List<TargetGroupAttribute> targetGroupAttributes) {
+    if (targetGroupAttributes.size() != 0) {
```
cfieber, line 122: `!targetGroupAttributes.isEmpty()`

PR #4673, opened 2020-06-12
fix(cache): OnDemandType in no longer an enum

claymccoy:
Plugins that wanted to add an OnDemandType previously wouldn't be able to add to the enum.


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/OnDemandType.java
```diff
@@ -0,0 +1,57 @@
...
+package com.netflix.spinnaker.clouddriver.cache;
+
+import java.util.Objects;
+import lombok.Data;
+
+@Data
+public class OnDemandType {
+  private final String value;
+
+  public static OnDemandType fromString(String value) {
```
ezimanyi, line 27: The `enum` had the property that for a given string `s`, `fromString(s)` always returned a reference to the same `OnDemandType`.  Now a new object is created every time, so it is longer true that `fromString("abc") == fromString("abc")`.

This will any place where we compare `OnDemandType` objects using `==`, which appears to be in a lot of places.

It's probably a lot easier/safer to update `fromString` to retain the prior behavior than to update all the usages.

claymccoy: @ezimanyi Thank you, great catch! Where `==` is used in Java files it will do an instance comparison and fail. I'd like to discourage instance comparison because it would only work for the build in types that exist in the OnDemandType class. I looked through all the files and only found three cases where instance comparison happens so I just changed those.

ezimanyi: Thanks!  I guess as one more thing I'd love to see a test on `.equals` to make sure it's behaving how it seems...looks correct but if we're now depending on that in more places would be good to have a test on this new function.

claymccoy: @ezimanyi Added a plain java junit test as not to add an Groovy.

ezimanyi: Awesome, thanks! Yeah I've also stopped writing even tests in Groovy, +1 to Java unit tests.

PR #4674, opened 2020-06-15
fix(ecs): set IAMRoleCachingAgent AWS region

allisaurus:
Previously, DEFAULT_REGION was always used, breaking for users in China or GovCloud.
This change uses a configured region, if present, and logs the choice.

Fixes https://github.com/spinnaker/spinnaker/issues/5710

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/IamRoleCachingAgent.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/IamRoleCachingAgent.java
@@ -83,8 +83,7 @@ public class IamRoleCachingAgent implements CachingAgent {
 
   @Override
   public CacheResult loadData(ProviderCache providerCache) {
-    AmazonIdentityManagement iam =
-        amazonClientProvider.getIam(account, Regions.DEFAULT_REGION.getName(), false);
+    AmazonIdentityManagement iam = amazonClientProvider.getIam(account, getIamRegion(), false);
 
     Set<IamRole> cacheableRoles = fetchIamRoles(iam, accountName);
     Map<String, Collection<CacheData>> newDataMap = generateFreshData(cacheableRoles);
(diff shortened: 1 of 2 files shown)


pkandasamy91:
Lgtm, are there any tests that should be updated?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/IamRoleCachingAgent.java
```diff
@@ -134,6 +133,23 @@ private void logUpcomingActions(
...
   }
 
+  protected String getIamRegion() {
+    // sample a region from the account in case the default won't work
+    String testRegion =
+        !account.getRegions().isEmpty() && account.getRegions().get(0) != null
+            ? account.getRegions().get(0).getName()
+            : "";
+
+    if (!testRegion.equals("")
```
pkandasamy91, line 143: Since it's a string, we can also use StringUtils.isBlank(testRegion) here - it checks null, empty or blank spaces.

clouddriver-ecs/src/test/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/IamRoleCachingAgentTest.java
```diff
@@ -152,4 +154,39 @@ public void shouldGenerateFreshData() {
...
+    String expectedRegionName = "cn-north-1";
+    when(netflixAmazonCredentials.getRegions())
+        .thenReturn(
+            Collections.singletonList(new AmazonCredentials.AWSRegion(expectedRegionName, null)));
+
+    // when
+    String actualRegionName = agent.getIamRegion();
+
+    // then
+    assertEquals(
```
pkandasamy91, line 187: Same comment as above

clouddriver-ecs/src/test/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/IamRoleCachingAgentTest.java
```diff
@@ -152,4 +154,39 @@ public void shouldGenerateFreshData() {
...
+    String defaultRegionName = Regions.DEFAULT_REGION.getName();
+    when(netflixAmazonCredentials.getRegions())
+        .thenReturn(Collections.singletonList(new AmazonCredentials.AWSRegion("us-east-1", null)));
+
+    // when
+    String actualRegionName = agent.getIamRegion();
+
+    // then
+    assertEquals(
+        "Expected region to equal " + defaultRegionName + ", but got " + actualRegionName,
```
pkandasamy91, line 170: I think the assert already fails with an error stating what your expected and actual values are. Might be unnecessary

allisaurus: ah, I was going by how the other tests in this file are structured; would prefer to keep them consistent for now

PR #4677, opened 2020-06-15
fix(lang): convert Groovy to Java for cloud provider classes

claymccoy:
Converting these Groovy classes to Java will make it so that cloud provider plugins do not require a Groovy dependency, and it is the first step toward cloud provider extension points.


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/CloudProvider.java
```diff
@@ -1,49 +1,51 @@
 /*
- * Copyright 2015 Netflix, Inc.
+ * Copyright 2020 Armory, Inc.
```
robzienert, line 2: Mind the copyright header changes.

claymccoy: Yeah, I think that happened automatically. I wasn't sure what the protocol is for keeping/changing headers when a file is rewritten. I'm definitely not trying to take credit for existing work.

PR #4678, opened 2020-06-15
fix(provider/cf): fix destroy cf service instance

zachsmith1:
Originally, when deleting a cloudFoundry service it would check for any managed services matching the name, and then delete it. If it was a userProvidedService (not managed), it would delete it but then throw NOT_FOUND and the pipeline stage would crash, even though it was successfully deleted. It should have thrown the correct status which is DELETE -> IN-PROGRESS. This fix does a hard check after the managedServices deletion logic to see if it is a userProvidedService. If found, will then delete it and pass along the correct response. If it can't find a managed or userProvided service it will throw the correct response -> not found. There is already a test that successfully tests this logic, but it had the wrong/incorrect response so it was changed to reflect the correct status.


german-muzquiz:

ethanfrogers:

PR #4680, opened 2020-06-15
refactor(core): Fix warning fetching operation version

ezimanyi:
The code to fetch the version of an operation catches any Exception. The only actual exception that can occur is a null-pointer exception of credentials.getOne returns null. Let's just make the code null-safe and remove all catch and warning logic.

One of the main motivations of this is that ProviderVersion is no longer even relevant now that the Kubernetes V1 provider has been removed from the code. Logging a warning related to being unable to fetch the provider version just confuses debugging; if the account doesn't exist, that itself is the problem, so we should note that in the warning, rather than that we couldn't fetch the version.


ezimanyi:
I came across this because someone reported in Slack an error related to `ProviderVersion`, which at first I feared was related to the Kubernetes V1 deprecation.

But it turns out it was an NPE because the account didn't exist, so we were calling `getProviderVersion` on `null`.  It took too much back-and-forth because of the confusing errors, hence my goal to fix this here.

I know you already have spinnaker/spinnaker#5749 open to completely remove this logic, which will be the better fix, but figured in the meantime this tiny error message fix would be helpful.

PR #4681, opened 2020-06-15
feat(safety): Copy `TrafficGuard` from Orca to Clouddriver

robzienert:
This is an incomplete migration of the `TrafficGuard` class (and supporting classes) from Orca.

This class is 99% the same as the one in Orca, however the `verifyTrafficRemoval` methods have had their signatures simplified, as the complexity is not necessary in Clouddriver.

`TrafficGuard` isn't used anywhere with this PR, it's just copying the code from Orca into Clouddriver - this is in support of my generic cloud operation stage.


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/safety/ClusterMatchRule.java
```diff
@@ -0,0 +1,76 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.clouddriver.safety;
+
+public class ClusterMatchRule {
```
robzienert, line 19: I would've Lombok'd this class, but given the odd handling of null values, I didn't feel comfortable making those changes.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/safety/TrafficGuard.java
```diff
@@ -0,0 +1,377 @@
...
+    //  which would be a much larger refactor to bring over in this commit. I
+    //  would like to postpone such a refactor until it's actually needed.
+    throw new UnsupportedOperationException(
+        "verifyInstanceTermination method has not been migrated from Orca yet");
+  }
+
+  public void verifyTrafficRemoval(
+      String serverGroupName,
+      String account,
+      String location,
```
robzienert, line 86: I was going to port Orca's `Location` class over to Clouddriver, but it looks like we can just use a String here - it'd basically just be `ServerGroup#getRegion()`.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/safety/TrafficGuard.java
```diff
@@ -0,0 +1,377 @@
...
+    Moniker serverGroupMoniker = NamerRegistry.getDefaultNamer().deriveMoniker(serverGroupName);
+
+    ClusterProvider<?> clusterProvider =
+        getClusterProvider(cloudProvider)
+            .orElseThrow(
+                () ->
+                    new TrafficGuardException(
+                        format(
+                            "Could not find ClusterProvider for cloud provider '%s'",
+                            cloudProvider)));
```
robzienert, line 99: Using `ClusterProvider` is the only deviation of internals from Orca.

clouddriver-core/src/test/groovy/com/netflix/spinnaker/clouddriver/safety/TrafficGuardSpec.groovy
```diff
@@ -0,0 +1,615 @@
...
+
+    ServerGroup serverGroup =  new SimpleServerGroup([
+      region   : 'us-east-1',
+      name     : name,
+      disabled : false,
+      instances: instances,
+      capacity : new ServerGroup.Capacity(min: 0, max: 4, desired: 3)
+    ] + overrides)
+    return serverGroup
+  }
```
robzienert, line 87: Tests are all the same except:

1. I `@Ignored` the tests related to `verifyInstanceTermination`
1. I had to refactor `makeServerGroup` because `ServerGroup` has a `Set` for instances which was incompatible with how the tests were originally written. 
1. I also added a `makeCluster` since these things need to typed now instead of maps of maps.
1. I removed one test that was asserting that `namespace` could be used for location. This is unnecessary as the location is explicitly provided via method parameter.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/safety/ClusterMatchRule.java
ezimanyi: Yeah, looks like it could use some `null -> empty` translation in the setters, but agree that's a separate issue from the changes here.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/safety/TrafficGuard.java
```diff
@@ -0,0 +1,377 @@
...
+      String serverGroupName,
+      List<String> instanceIds,
+      String account,
+      String location,
+      String cloudProvider,
+      String operationDescriptor) {
+    // TODO(rz): I opted out of migrating this method because it isn't used in
+    //  my current refactors. This method uses clouddriver search endpoint,
+    //  which would be a much larger refactor to bring over in this commit. I
+    //  would like to postpone such a refactor until it's actually needed.
```
ezimanyi, line 78: How long will this class be both in `clouddriver` and `orca`?  Given that it looks from this comment like there are still usages in `orca`, is the plan for this to live in both places indefinitely, or is there a plan to eventually move the orca usages here?

ajordens: I guess one question would be could we (in some semi short-term) move everything over and just have a shim in `orca` that makes an API call to a traffic guard api backed by this class. Perhaps that's what you mean by move the `orca` usages here @ezimanyi.

We're already making a call across the wire and at least if we can get the business logic all on one side we wouldn't have to worry so much about drift.

I like this logic being in `clouddriver` and see it moving across as a win.

robzienert: Introducing an API is probably the right move - I have no idea how long it will take to migrate all cloud provider logic into Orca -- probably years, unless cloud providers jump on the bandwagon to help the migration process.

It'd be nice to have help on this effort. I'll add creating a Traffic Guard API to clouddriver to my backlog, but getting the initial cloud operation stage in is my higher priority.

PR #4685, opened 2020-06-16
feat(core): Support individually versioned operations

robzienert:
This allows operations to be versioned within the same provider version.
Having this capability will make it easier for us to migrate cloud-specific
logic in Orca stages to Clouddriver, without breaking existing stages.


ezimanyi:
Is there a plan to actually use and support the `ProviderVersion` field?  As far as I'm aware, this was only ever used by the Kubernetes provider, which forked and built a V2 provider alongside the V1 provider.  Now that the Kubernetes V1 provider has been deleted, I don't believe that there are any providers using that field, and there are plans to [delete it](https://github.com/spinnaker/spinnaker/issues/5749) shortly.  Ultimately, our experience with having two versions of the Kubernetes provider was not great, so I'd like to understand better what the plans are around using this field.  CC @maggieneterval

robzienert:
I was talking to Maggie the other day about it. I was going to delete it, but there are still dozens of references to it in the kubernetes provider. It's on the TODO for cleanup, so I'm just going to leave that be so I don't screw anything up with that provider. 😊  I have no interest in using `ProviderVersion` for anything, though - I think versioning on a per-operation basis enables more varied use cases than trying to version an entire provider.

ezimanyi:
Oh, awesome, as long as there are no plans to start using it, we can handle deleting it as part of Kubernetes cleanup.  Thanks for clarifying!

PR #4694, opened 2020-06-18
fix(api): create clouddriver-api and move cloud provider API

claymccoy:
This moves the current cloud provider API to the new clouddriver-api module. Not everything that is API was moved, but it is almost everything I used in the Nomad plugin. Most of what was moved has been left as is for now. The things not yet moved or modified were mainly due to the need to limit external dependencies in clouddriver-api. There are specific notes below.

build.gradle doesn't add any of the cross module stuff to clouddriver-api just like clouddriver-bom.
Several modules now depend on clouddriver-api, but some of thier other interproject depedencies can be removed soon (especially for the cloud providers).

Modifications:
* DefaultCacheData was split into DefaultCacheData and DefaultJsonCacheData to separate out the jackson based construction.
* AbstractAccountCredentials was separated from AccountCredentials to remove the getPermissions method (which requires fiat)
* OnDemandMetricsSupportable is a new interface for OnDemandMetricsSupport which requires spectator, but only for implementation.
* SearchableProvider no longer uses groovy.transform.Canonical. It shouldn't matter since it is a Java class.

I've fix a few places in the code that these changes affect. DefaultJsonCacheData is used for object mappers. AccountCredentials implemenations now extend AbstractAccountCredentials. DefaultAllowedAccountsValidator has added conditional logic around getPermissions() calls because that isn't in the AccountCredentials interface.

Todo:
* DescriptionValidator uses org.springframework.validation.Errors from spring-context in the method contract.
* DeployHandler uses DeploymentResult which is groovy and should be refactored separately.
* A few unit tests need to move to clouddriver. I didn't know if we were cool with Groovy unit tests or if I needed to rewrite them. They work where they are.

This was the least invasive commit I could make. The above todos will happen in follow up PRs.


build.gradle
```diff
@@ -41,7 +41,7 @@ subprojects {
   group = "com.netflix.spinnaker.clouddriver"
   apply plugin: 'io.spinnaker.project'
 
-  if (name != "clouddriver-bom") {
+  if (name != "clouddriver-bom" && name != "clouddriver-api") {
```
jonsie, line 44: Nice.

clouddriver-api/clouddriver-api.gradle
```diff
@@ -0,0 +1,38 @@
...
+
+apply plugin: 'java-library'
+
+dependencies {
+  implementation platform("com.netflix.spinnaker.kork:kork-bom:$korkVersion")
+  annotationProcessor platform("com.netflix.spinnaker.kork:kork-bom:$korkVersion")
+
+  api("com.netflix.spinnaker.kork:kork-plugins-api")
+
+  implementation "com.netflix.spinnaker.moniker:moniker"
```
jonsie, line 25: It's nice that this is implementation and it looks like moniker is pretty thin and not too scary, dependency wise.  This may be ok.

claymccoy: Rob Z said this one was acceptable.

cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/DefaultJsonCacheData.java
```diff
@@ -0,0 +1,36 @@
...
+
+  @JsonCreator
+  public DefaultJsonCacheData(
+      @JsonProperty("id") String id,
+      @JsonProperty("ttlSeconds") int ttlSeconds,
+      @JsonProperty("attributes") Map<String, Object> attributes,
+      @JsonProperty("relationships") Map<String, Collection<String>> relationships) {
+    super(id, ttlSeconds, attributes, relationships, Clock.systemDefaultZone());
+  }
+}
```
robzienert, line 36: Hmm, why not use Jackson mixins? I don't think this class is necessary.

clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/orchestration/AtomicOperation.java
```diff
@@ -1,31 +1,34 @@
 /*
- * Copyright 2017 Netflix, Inc.
+ * Copyright 2020 Netflix, Inc.
```
robzienert, line 2: Mind the copyright changes.

clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountCredentials.java
```diff
@@ -98,43 +96,18 @@ default String getAccountId() {
...
   /**
    * A user in ANY required group should be allowed access to this account.
    *
    * @return the group names that govern access to this account, empty indicates a public account
    *     accessible by all.
    */
   @Deprecated
-  List<String> getRequiredGroupMembership();
-
-  default Permissions getPermissions() {
```
robzienert, line 110: Where'd this stuff go?

clouddriver-security/src/main/groovy/com/netflix/spinnaker/clouddriver/security/AbstractAccountCredentials.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.netflix.spinnaker.fiat.model.Authorization;
+import com.netflix.spinnaker.fiat.model.resources.Permissions;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public abstract class AbstractAccountCredentials<T> implements AccountCredentials<T> {
```
robzienert, line 28: This seems wrong to me. `getCredentials` and `getPermissions` both are used fairly widely in the implementations, but is no longer part of the `AccountCredentials` API - we're basically forcing people to extend from this class, which is quite limiting.

I'm thinking it might make more sense to move the `fiat-api` models into an actual shared models location that we can feel comfortable having as a dependency. I'm happy to bite this work off in support of this PR.

cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/DefaultJsonCacheData.java
claymccoy: Right. I saw the mixins in keel, and that would probably be the long term solution. I didn't see an example of `JsonCreator` and wasn't sure how to ensure those all got applied. I was hoping to get this in before code drifted and then iterate on it further by adding the api-mixin module like keel has. There is a usage of JsonIgnore that would be more straightforward. This was just a straight forward way to avoid the Jackson dependency in the short term.

clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountCredentials.java
claymccoy: I left `getRequiredGroupMembership` out of the interface since it is deprecated. `getPermissions` can't be in api because it requires fiat as part of the contract. It is possible to write a plugin with this interface. And then this membership/permission stuff can be redone separately.

clouddriver-security/src/main/groovy/com/netflix/spinnaker/clouddriver/security/AbstractAccountCredentials.java
claymccoy: `getCredentials` is in AccountCredentials, but also here to separate the jackson annotation. Something that I was going to do with mixins later.
getPermissions can't be in `-api` because of the need to include fiat. I agree with a change to fiat-api in the long term. I was just trying to find a palatable way to move stuff to `-api` without dependencies thinking that we could make those changes later.
This could be an interface with default methods rather than an abstract class if the need to extend it is a concern.

claymccoy:
@rz I totally agree with all the comments. I think the only thing that we are disagreeing about is my strategy of moving things to `-api` and then incrementally refactoring vs getting everything perfect and then moving it to `-api`. My thought was that I had to try to move everything over to get an idea of what would get in the way. Once I did (with some minimal modifications) and saw it working I wanted to go ahead and get that in (as Beta) then progressively improve (with mixins, fiat-api...). My first approach was to move everything over with the unwanted dependencies in api and incrementally remove them. After that got shot down, this is plan B.

So maybe plan C is this. If you want to look into the fiat model, I'll change to using jackson mixins and we can try this again?

It is also worth noting that `DescriptionValidator` api would need to change before it can be moved over, but that class is islolated and optional so didn't have to be moved over at this time.

cats/cats-core/src/main/java/com/netflix/spinnaker/cats/cache/DefaultJsonCacheData.java
robzienert: Ok, that works for me. We can loop back.

Fwiw, I'm 99% sure we should be able to add a constructor on the mixin abstract class and annotate that with `@JsonCreator`: https://github.com/FasterXML/jackson-docs/wiki/JacksonMixInAnnotations

clouddriver-security/src/main/groovy/com/netflix/spinnaker/clouddriver/security/AbstractAccountCredentials.java
robzienert: I see. I'll look into taking on something WRT Fiat, since this appears to be a reoccurring issue in the plugin effort. This is good for now - can you add a TODO to this class so this isn't lost?

robzienert:
Yeah, the approach you're taking sounds legit to me. 

Looks like the build is failing for formatting reasons, once that's fixed up, I'd say we can merge this and move on.

PR #4696, opened 2020-06-19
fix(kubernetes/IRSA): Update aws-iam-authenticator to 0.5.0

AndriMar:
It looks like the aws-iam-authenticator version 0.4.0 currently installed during docker build does not work with IRSA.
After downloading aws-iam-authenticator version 0.5.0, the same as in halyard then it works fine and I can see that when communicating with other EKS cluster that it is using the correct role and not the role on the node.

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -15,7 +15,7 @@ RUN wget https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VER
   chmod +x kubectl && \
   mv ./kubectl /usr/local/bin/kubectl
 
-RUN wget https://amazon-eks.s3-us-west-2.amazonaws.com/1.13.7/2019-06-11/bin/linux/amd64/aws-iam-authenticator && \
+RUN wget https://amazon-eks.s3.us-west-2.amazonaws.com/1.16.8/2020-04-16/bin/linux/amd64/aws-iam-authenticator && \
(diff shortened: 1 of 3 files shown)


AndriMar:
Here is the issue on spinnaker.
https://github.com/spinnaker/spinnaker/issues/5857

ezimanyi:
@AndriMar : Looks like I don't have permission to merge `master` into your branch, so you'll need to do that yourself.  This should automatically merge once you've done that.  Thanks!

AndriMar:
@ezimanyi Great thank you!
One problem though, the fork was created under my company GitHub Org and it looks like it is not possible to **Allow edits from maintainers** when the fork is under Org but it should be possible when it is under my user, see [here](https://github.community/t/how-can-we-enable-allow-edits-from-maintainers-by-default/2847/8).

It looks like it is not enough to merge the master from the origin to the fork.

ezimanyi:
@Mergifyio refresh

ezimanyi:
@AndriMar : I'm not sure why Mergify wasn't working, but I just merged it manually, as the branch was up to date after your merge commit.  I have heard from others that their company doesn't allow them to check "Allow edits from maintainers"; in this case I think all we can really do is approve the PR and add the 'ready to merge' label then when the PR author updates it, then it should merge.

PR #4699, opened 2020-06-23
feat(cloudconfig): Externalize cloud config refresh interval and decouple from cache refresh interval

clanesf:
During operational use of Clouddriver using the Spring Cloud Config backend we've found that having the cloud refresh config interval tied directly to the cache refresh interval problematic. Particularly in a scenario where you have lots of clusters, kicking this off every 30-60 seconds (the default) runs into issues as a refresh of the beans never gets completed before the next refresh beings. This results in slower than desirable addition of new accounts and can also lead to performance problems as clouddriver gets hung up constantly running refresh code. 

This PR creates a separately configurable value that can be included in the spring profile and enabled so that you can set a value for this refresh interval that works for your specific infrastructure requirements. 

It can be enabled by including the following in your clouddriver-local.yml
```
cloud:
  config:
    enabled: true
    refreshInterval: 180   # refresh cloud config every 3 minutes. 
```


ncknt:
LGTM - although 2 comments: 
- you could just depend on `cloud.config.refreshInterval` as it can be confusing to have `cloud.config.enabled: false` but yet still have cloud config refresh.
- `refreshIntervalSeconds` may be a more explicit name.

clanesf:
Adjusted as per feedback from @ncknt as well as slack discussion with @scottfrederick. I removed the part that would tie this to the existing cache refresh interval because as [4937](https://github.com/spinnaker/spinnaker/issues/4937) points out and I discovered as well, the existing implementation didn't work correctly.

This means that going forward users will either have the option of setting their refresh interval as follows in clouddriver-local.yml
```
cloud:
  config:
      refreshIntervalSeconds: 180   # refresh cloud config every 3 minutes. 
```
or it will default to the 60 seconds it's always used previously.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CloudConfigRefreshConfig.java
```diff
@@ -16,37 +16,42 @@
...
 import org.springframework.context.annotation.*;
 
 /**
  * Create a {@link CloudConfigRefreshScheduler} to refresh the Spring Cloud Config Server from an
- * environment repository backend on a schedule that matches the cache refresh schedule.
+ * environment repository backend on a schedule that can be defined in the spring profile or has a
+ * sensible default (60 seconds)..
  */
 @Configuration
 @AutoConfigureAfter({RedisCacheConfig.class})
```
scottfrederick, line 36: This `@AutoConfigureAfter` can be removed now that there is no relationship to the cache config.

clanesf: Removed, thanks!

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CloudConfigRefreshConfig.java
```diff
@@ -16,37 +16,40 @@
...
   @Conditional(RemoteConfigSourceConfigured.class)
   @EnableConfigServer
   static class RemoteConfigSourceConfiguration {
+
     @Bean
-    @ConditionalOnBean(DefaultAgentIntervalProvider.class)
-    public CloudConfigRefreshScheduler intervalProviderConfigRefreshScheduler(
-        ContextRefresher contextRefresher, DefaultAgentIntervalProvider agentIntervalProvider) {
-      return new CloudConfigRefreshScheduler(contextRefresher, agentIntervalProvider.getInterval());
+    @ConditionalOnProperty(prefix = "cloud.config", value = "refreshIntervalSeconds")
```
ncknt, line 44: You can drop the conditional as well as the second bean since you default to 60 in the @configuration itself.

PR #4700, opened 2020-06-23
fix(provider/appengine): Fix for issue spinnaker/spinnaker#5834

guido9j:
Deploying to GAE using GCS throws stderr: ERROR: (gcloud.app.deploy) [...app.yaml] does not exist. The error occurs because GcsStorageService creates directories as files, and later files in the directory fail to created. Later, we determined Google Console UI was creating placeholder objects for folders.


guido9j:
@ezimanyi : Here is the pull request for spinnaker/spinnaker#5834

PR #4710, opened 2020-06-25
feat(aws): Adding support for update asg launch template

jeyrschabu:
- added `ModifyServerGroupLaunchTemplateAtomicOperation` to handle modifying an asg's launch template


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ModifyServerGroupLaunchTemplateAtomicOperation.java
```diff
@@ -0,0 +1,251 @@
...
+            newLaunchTemplateData,
+            Optional.of(blockDevices));
+
+    launchTemplateSpec.setVersion(String.valueOf(newVersion.getVersionNumber()));
+
+    // point asg to new template version
+    autoScaling.updateAutoScalingGroup(
+        new UpdateAutoScalingGroupRequest()
+            .withAutoScalingGroupName(description.getAsgName())
+            .withLaunchTemplate(launchTemplateSpec));
```
robzienert, line 221: Are there reasons why we _aren't_ using sagas for atomic operations that make more than one write operation to AWS?

jeyrschabu: Definitely a saga opportunity.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/ModifyServerGroupLaunchTemplateAtomicOperation.java
```diff
@@ -16,306 +16,59 @@
...
-    return Optional.empty();
+  @Nonnull
+  @Override
+  protected SagaFlow buildSagaFlow(List priorOutputs) {
+    this.priorOutputs = priorOutputs;
+    return new SagaFlow()
+        .then(FindAutoScalingGroup.class)
+        .then(PrepareModifyServerGroupLaunchTemplate.class)
+        .then(ModifyServerGroupLaunchTemplate.class)
+        .then(UpdateAutoScalingGroup.class);
```
aravindmd, line 50: I think a dedicated saga exception handler is required or is the case that exceptions that are of type SpinnakerExceptions will be retried automatically?

jeyrschabu: Yeah this is still in progress. Got a few todos regarding exception handling.

jeyrschabu:
Hey @rz there is a mismatch in serialization/deserialization of commands and their fields.

Been running into a lot of jackson problems and I am getting close to undoing the saga approach for now and figure that out later.
If you know of a quick way to address jackson's issues, I'd like to hear about it.

robzienert:
@jeyrschabu More information needed. At what point is this occurring?

Clouddriver, for the most part, uses the same `ObjectMapper` for everything (see: `Find Usages` on the `ObjectMapper` type). A couple things you could investigate that jumped out immediately at me:

1. `clouddriver-aws`'s `CredentialsLoader` uses its own `ObjectMapper`. I suspect this is just for serde around `clouddriver.yml`, but I don't know why it'd need its own `ObjectMapper` for this task anyway. Could likely be cleaned up.
2. `clouddriver-sql` is creating a new `ObjectMapper` for the two `TaskRepository` factory  methods in `SqlConfiguration`, rather than injecting the object mapper used elsewhere in Clouddriver. Without having more insight into the problem, I'd venture a guess that this is the issue.

jeyrschabu:
Picking this back up. @robzienert thanks for the pointers. I added ```  @JsonProperty
  @Override
  String getAccount() {
    return credentials?.name ?: account
  }``` to `AbstractAmazonCredentialsDescription` and that fixed the ser/de issue.

cfieber (approved):
cool to see this done with the saga framework!

caseyhebebrand (approved):

plumpy:
This [breaks startup when SQL isn't enabled](https://builds.spinnaker.io/job/Validate_BomMultiPlatform/5807/CLOUD_PLATFORM=gce/artifact/logs/service_logs/clouddriver.log):
```
***************************
APPLICATION FAILED TO START
***************************
Description:
Parameter 0 of method sagaRepository in com.netflix.spinnaker.clouddriver.saga.config.SagaAutoConfiguration required a bean of type 'com.netflix.spinnaker.clouddriver.event.persistence.EventRepository' that could not be found.
The following candidates were found but could not be injected:
        - Bean method 'sqlEventRepository' in 'SqlConfiguration' not loaded because @ConditionalOnProperty (sql.enabled) did not find property 'sql.enabled'
Action:
Consider revisiting the entries above or defining a bean of type 'com.netflix.spinnaker.clouddriver.event.persistence.EventRepository' in your configuration.
```

PR #4712, opened 2020-06-25
chore(ecs): Refactor TaskHealthCachingAgent networking check

allisaurus:
Previously, how to check task health forked based on looking for network bindings on each container. Whether the task contains network bindings was already being determined in a previous check, so this surfaces that value and forks based on that instead. 

Also adds test to ensure tasks with missing or null network bindings/interfaces are skipped.

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
@@ -129,27 +129,30 @@ public class TaskHealthCachingAgent extends AbstractEcsCachingAgent<TaskHealth>
             Keys.getTaskDefinitionKey(accountName, region, service.getTaskDefinition());
         TaskDefinition taskDefinition = taskDefinitionCacheClient.get(taskDefinitionCacheKey);
 
-        if (isContainerMissingNetworking(task)) {
+        if (taskDefinition == null) {
           log.debug(
-              "Task '{}' is missing networking. Will not retrieve health.", task.getTaskArn());
+              "Provided task definition '{}' is null for service '{}'",
+              service.getTaskDefinition(),
+              service.getServiceName());
+          continue;
+        }
+
+        boolean lacksNetworkBindings = isTaskMissingNetworkBindings(task);
+        if (task.getContainers().isEmpty()
+            || (lacksNetworkBindings && isTaskMissingNetworkInterfaces(task))) {
+          log.debug(
+              "Task '{}' is missing networking. Will not retrieve load balancer health.",
+              task.getTaskArn());
           continue;
         }
 
         TaskHealth taskHealth = null;
-        Collection<Container> containers = task.getContainers();
-
-        for (Container container : containers) {
-          if (container.getNetworkBindings().size() >= 1) {
-            taskHealth =
-                inferHealthNetworkBindedContainer(
-                    targetHealthCacheClient,
-                    task,
-                    containerInstance,
-                    serviceName,
-                    service,
-                    taskDefinition);
-            break;
-          }
+        // ideally, could determine health check method by looking at taskDef.networkMode,
+        // however this isn't reliably cached yet, so reusing network binding check.
+        if (!lacksNetworkBindings) {
+          taskHealth =
+              inferHealthNetworkBindedContainer(
+                  targetHealthCacheClient, task, containerInstance, serviceName, service);
         }
 
         if (taskHealth == null) {
@@ -157,6 +160,7 @@ public class TaskHealthCachingAgent extends AbstractEcsCachingAgent<TaskHealth>
               inferHealthNetworkInterfacedContainer(
                   targetHealthCacheClient, task, serviceName, service, taskDefinition);
         }
+
         log.debug("Task Health contains the following elements: {}", taskHealth);
 
         if (taskHealth != null) {
@@ -177,11 +181,6 @@ public class TaskHealthCachingAgent extends AbstractEcsCachingAgent<TaskHealth>
       String serviceName,
       Service loadBalancerService,
       TaskDefinition taskDefinition) {
-
-    if (taskDefinition == null) {
-      log.debug("Provided task definition is null for task '{}'.", task.getTaskArn());
-      return null;
-    }
 
     List<LoadBalancer> loadBalancers = loadBalancerService.getLoadBalancers();
     log.debug("LoadBalancerService found {} load balancers.", loadBalancers.size());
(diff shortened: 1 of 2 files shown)


clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
```diff
@@ -129,34 +129,38 @@ public TaskHealthCachingAgent(
...
-              "Task '{}' is missing networking. Will not retrieve health.", task.getTaskArn());
+              "Provided task definition '{}' is null for service '{}'",
+              service.getTaskDefinition(),
+              service.getServiceName());
+          continue;
+        }
+
+        boolean lacksNetworkBindings = isTaskMissingNetworkBindings(task);
+        if (task.getContainers().isEmpty()
+            || (lacksNetworkBindings && isTaskMissingNetworkInterfaces(task))) {
```
pkandasamy91, line 142: nit: we should probably also move isTaskMissingNetworkInterfaces to a variable (seems weird that lacksNetworkBindings is a variable while lacksNetworkInterfaces is not or vice versa)

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
```diff
@@ -129,34 +129,38 @@ public TaskHealthCachingAgent(
             Keys.getTaskDefinitionKey(accountName, region, service.getTaskDefinition());
         TaskDefinition taskDefinition = taskDefinitionCacheClient.get(taskDefinitionCacheKey);
 
-        if (isContainerMissingNetworking(task)) {
+        if (taskDefinition == null) {
```
pkandasamy91, line 132: Out of curiosity, why do we skip getting network bingings if taskDefinition does not exist in the cache? AFAICT NetworkBindings doesn't rely on the taskDefinition being present. Couldn't this lead to issues where taskDefinition hasn't yet been cached, and we skip the networkBindings/Interfaces check?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
pkandasamy91: Also, technically a task should not be able to have both networkingBindings and networkInterfaces (can we add a check that logs an error if this were ever to happen - just so we're made aware of any weird behaviour) 

`          if (!lacksNetworkBindings && !lacksNetworkInterfaces) {`

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
```diff
@@ -129,34 +129,38 @@ public TaskHealthCachingAgent(
...
-          }
+        // ideally, could determine health check method by looking at taskDef.networkMode,
+        // however this isn't reliably cached yet, so reusing network binding check.
+        if (!lacksNetworkBindings) {
+          taskHealth =
+              inferHealthNetworkBindedContainer(
+                  targetHealthCacheClient, task, containerInstance, serviceName, service);
         }
 
         if (taskHealth == null) {
```
pkandasamy91, line 158: We should probably update this to be an else if condition since you can't have both networking bindings and interfaces present 

`else if (!lacksNetworkInterfaces) {`

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
```diff
@@ -178,11 +182,6 @@ private TaskHealth inferHealthNetworkInterfacedContainer(
       Service loadBalancerService,
       TaskDefinition taskDefinition) {
 
-    if (taskDefinition == null) {
```
pkandasamy91, line 181: Might still want to keep this based on my previous comment.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
```diff
@@ -251,12 +249,7 @@ private TaskHealth inferHealthNetworkBindedContainer(
       Task task,
       ContainerInstance containerInstance,
       String serviceName,
-      Service loadBalancerService,
-      TaskDefinition taskDefinition) {
-    if (taskDefinition == null) {
-      log.debug("Provided task definition is null.");
-      return null;
-    }
```
pkandasamy91, line 259: Good catch! We shouldn't be passing taskDefinition if we're not actually using it.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
allisaurus: originally i had both pulled out as vars, but the result of `isTaskMisingNetworkingInterfaces` was never used again, so I decided to just call the func here. Can put it back if there's a benefit to variable-izing it (not sure that's a word...)

pkandasamy91: I think it makes sense if it's used here as well https://github.com/spinnaker/clouddriver/pull/4712#discussion_r445765662

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
allisaurus: hmm you know originally both methods were returning if `taskDefinition` was null, so it made sense to just do once upfront. but since it's actually NOT being used by `inferHealthNetworkBindedContainer` and we can't use the network mode to determine how to health check (which would require it sooner - see comment on line 150), you're right that we can just do that check once below. Will update!

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskHealthCachingAgent.java
allisaurus: I wanted to keep the existing logic (outside removing the loop) mostly intact here, but I agree it's beneficial to log if both are found (see above). Maybe I can make this & the logging change in a separate PR so they're isolated?

pkandasamy91: Sounds good to me!

PR #4715, opened 2020-06-26
fix(titus): Skip security group conversion based on label

aravindmd:
Titus would like to pass SG's that belong to different account / region to support their multi account setup , I made a change to skip validation during cloneServerGroup operation here https://github.com/spinnaker/clouddriver/pull/4307 , But with createServerGroup operation the conversion was failing because the SG's don't belong to the same account / region .

This change should help workaround this case where they would pass a label and we skip the conversion this PR and validation https://github.com/spinnaker/clouddriver/pull/4307 , the assumption here is that the SG ids are required.


aravindmd:
closing this PR for now , we don't need this work around for titus anymore

PR #4716, opened 2020-06-26
fix(kubernetes): Don't delete caching agents of unmodified accounts after a refresh

german-muzquiz:
Fixes https://github.com/spinnaker/spinnaker/issues/5869

Instead of overwriting the collection of caching agents in `KubernetesV2Provider` by one that only has agents for new or changed accounts, new agents are added to the existing collection.

Old agents for changed accounts are removed by `ProviderUtils.unscheduleAndDeregisterAgents`, so there are no duplicate agents for changed accounts.

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/KubernetesV2Provider.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/KubernetesV2Provider.java
@@ -36,24 +36,22 @@ public class KubernetesV2Provider extends AgentSchedulerAware implements Provide
 
   private Collection<Agent> agents = emptyAgentCollection();
 
-  private Collection<Agent> nextAgentSet = emptyAgentCollection();
+  private Collection<Agent> stagedAgents = emptyAgentCollection();
 
   private static Collection<Agent> emptyAgentCollection() {
     return Collections.newSetFromMap(new ConcurrentHashMap<>());
   }
 
-  public void stageAllAgents(Collection<Agent> agents) {
-    nextAgentSet.addAll(agents);
+  public void stageAgents(Collection<Agent> agents) {
+    stagedAgents.addAll(agents);
   }
 
   public void clearStagedAgents() {
-    nextAgentSet.clear();
+    stagedAgents.clear();
   }
 
-  public void addStagedAgents() {
-    Collection<Agent> nextAgentSetCopy = emptyAgentCollection();
-    nextAgentSetCopy.addAll(nextAgentSet);
-    agents.addAll(nextAgentSetCopy);
+  public void promoteStagedAgents() {
+    agents.addAll(stagedAgents);
     clearStagedAgents();
   }
 
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/KubernetesV2Provider.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/KubernetesV2Provider.java
@@ -42,19 +42,19 @@ public class KubernetesV2Provider extends AgentSchedulerAware implements Provide
     return Collections.newSetFromMap(new ConcurrentHashMap<>());
   }
 
-  public void addAllAgents(Collection<Agent> agents) {
+  public void stageAllAgents(Collection<Agent> agents) {
     nextAgentSet.addAll(agents);
   }
 
-  public void clearNewAgentSet() {
+  public void clearStagedAgents() {
     nextAgentSet.clear();
   }
 
-  public void switchToNewAgents() {
+  public void addStagedAgents() {
     Collection<Agent> nextAgentSetCopy = emptyAgentCollection();
     nextAgentSetCopy.addAll(nextAgentSet);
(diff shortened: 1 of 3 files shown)


german-muzquiz:
cc @ezimanyi

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/KubernetesV2ProviderSynchronizable.java
```diff
@@ -178,24 +176,35 @@ private void saveToCredentialsRepository(KubernetesNamedAccountCredentials accou
...
+        kubernetesV2Provider.stageAllAgents(newlyAddedAgents);
       }
     } catch (Exception e) {
       log.warn("Error encountered scheduling new agents -- using old agent set instead", e);
-      kubernetesV2Provider.clearNewAgentSet();
+      kubernetesV2Provider.clearStagedAgents();
     }
 
+    // Remove existing agents belonging to changed accounts
+    ProviderUtils.unscheduleAndDeregisterAgents(accountNames, catsModule);
```
ncknt, line 206: If an exception occurs while building the agent (line 192), it will be caught and all agents will then be unregistered. I think we should 1) try/catch in the loop 2) not unregister/unschedule agents that had been previously scheduled unless they're explicitly deleted.

ncknt:
For context: it looks like a major issue that was present from the beginning. When a new or modified account is detected, all other accounts are removed. They become unavailable for Force Cache Refresh but agents are still scheduled and run as usual.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/KubernetesV2ProviderSynchronizable.java
german-muzquiz: Good catch

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/KubernetesV2Provider.java
```diff
@@ -42,19 +42,19 @@
...
   }
 
-  public void switchToNewAgents() {
+  public void addStagedAgents() {
     Collection<Agent> nextAgentSetCopy = emptyAgentCollection();
     nextAgentSetCopy.addAll(nextAgentSet);
-    agents = nextAgentSetCopy;
-    clearNewAgentSet();
+    agents.addAll(nextAgentSetCopy);
+    clearStagedAgents();
```
ezimanyi, line 57: Do we need the `nextAgentSetCopy` variable anymore?  Isn't this equivalent to:
```
agents.addAll(nextAgentSet);
clearStagedAgents();
```
Or am I missing something?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/KubernetesV2ProviderSynchronizable.java
```diff
@@ -136,12 +125,21 @@ public void synchronize() {
...
         .forEach(this::saveToCredentialsRepository);
 
-    ProviderUtils.unscheduleAndDeregisterAgents(deletedAccounts, catsModule);
-    ProviderUtils.unscheduleAndDeregisterAgents(changedAccounts, catsModule);
-
     return newAndChangedAccounts;
   }
 
+  private void deleteAccounts(List<String> deletedAccounts) {
+    if (deletedAccounts.isEmpty()) {
```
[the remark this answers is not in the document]
german-muzquiz: The `if` was just because of the log line, but the message could be better worded to remove the condition

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/KubernetesV2ProviderSynchronizable.java
```diff
@@ -196,14 +195,17 @@ private void synchronizeKubernetesV2Provider(
         log.info("Adding {} agents for account {}", newlyAddedAgents.size(), credentials.getName());
 
         kubernetesV2Provider.stageAllAgents(newlyAddedAgents);
+        stagedAccountNames.add(credentials.getName());
+      } catch (Exception e) {
+        log.warn(
+            "Error encountered scheduling new agents for account {} -- using old agent set instead",
```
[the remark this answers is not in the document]
german-muzquiz: In this case we're still using the old agent set of the failed account because `ProviderUtils.unscheduleAndDeregisterAgents` will not include the failed account name. So in practice the changed account is still being preiodically cached (and force cache refresh still works for it) if its new agent set fails to build.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/KubernetesV2Provider.java
german-muzquiz: You're right, we don't need the copy anymore

PR #4717, opened 2020-06-26
feat(provider/cf): add command property to manifest

zachsmith1:
The command property will override the existing buildpack command. Some buildpacks require that you specify a custom command so this was needed to satisfy those requirements. Command was partially implemented so there wasn't a ton that was needed in order for this to work. CloudFoundry states that if you want a buildpack to use its original command then you must pass _null_ explicitly as a command. The assumption here is that when you are deploying an application you can either set a custom command or leave the command off the manifest. If command is not included, it will use the default command specified by the buildpack. There was already testing in place to cover the core logic but I did add some minor modifications to ensure the field was being converted correctly.

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
@@ -582,7 +582,7 @@ public class Applications {
     }
     if (command != null && command.isEmpty()) {
       throw new IllegalArgumentException(
-          "Buildpack commands cannot be empty. Either a custom command is provided or set to null to use the original buildpack command.");
+          "Buildpack commands cannot be empty. Please specify a custom command or set it to null to use the original buildpack command.");
     }
 
     safelyCall(() -> api.updateProcess(guid, new UpdateProcess(command, healthCheck)));
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
@@ -580,6 +580,11 @@ public class Applications {
     if (healthCheckEndpoint != null && !healthCheckEndpoint.isEmpty() && healthCheck != null) {
       healthCheck.setData(new Process.HealthCheckData().setEndpoint(healthCheckEndpoint));
     }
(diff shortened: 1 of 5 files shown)


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
```diff
@@ -580,6 +580,11 @@ public void updateProcess(
     if (healthCheckEndpoint != null && !healthCheckEndpoint.isEmpty() && healthCheck != null) {
       healthCheck.setData(new Process.HealthCheckData().setEndpoint(healthCheckEndpoint));
     }
+    if (command != null && command.isEmpty()) {
+      throw new IllegalArgumentException(
+          "Buildpack commands cannot be empty. Either a custom command is provided or set to null to use the original buildpack command.");
```
kevinawoo, line 585:
```suggestion
          "Buildpack commands cannot be empty. Please specify a custom command or set it to null to use the original buildpack command.");
```

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
```diff
@@ -580,6 +580,11 @@ public void updateProcess(
     if (healthCheckEndpoint != null && !healthCheckEndpoint.isEmpty() && healthCheck != null) {
       healthCheck.setData(new Process.HealthCheckData().setEndpoint(healthCheckEndpoint));
     }
+    if (command != null && command.isEmpty()) {
```
kevinawoo, line 583: it might be worth checking for `""`, in this case, it should just default to the original buildpack command

zachsmith1: @kevinawoo I believe "" is being checked by command.isEmpty()

zachsmith1: I feel like an empty string in the manifest is potentially problematic and wanted to throw something if thats found. For example, if there are variables being substituted but in that process of substitution it resolved to an empty string it would be good for the user to know.

kevinawoo: 👍 makes sense, I'll leave this convo open incase someone comes looking to why `""` is throwing that error.

They might have a use case to further improve the error handling/message.

PR #4721, opened 2020-07-06
refactor(core): deprecate ProviderVersion

maggieneterval:
* feat(security): deprecate ProviderVersion enum and public methods that use ProviderVersion 

* feat(core): deprecate public interfaces and methods that use ProviderVersion, and remove ProviderVersion-dependent logic from private methods 

* refactor(alicloud): remove ProviderVersion from AliCloud 

  Although AliCloudCredentials is a public class, I think it should be safe to remove providerVersion because there are no non-default ProviderVersions configurable with Halyard, and this entire module is not even currently part of the OSS build.

* refactor(google): remove ProviderVersion from Google 

* refactor(core): remove acceptsVersion from cloud provider operations 

  Since we no longer filter by ProviderVersion in core, removes all uses of acceptsVersion(ProviderVersion) except those in public (now deprecated) methods and tests, which will be removed next release.

* refactor(kubernetes): remove ProviderVersion from Kubernetes accounts 

* refactor(web): remove providerVersion from CredentialsController

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AnnotationsBasedAtomicOperationsRegistry.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AnnotationsBasedAtomicOperationsRegistry.groovy
@@ -37,6 +37,82 @@ class AnnotationsBasedAtomicOperationsRegistry extends ApplicationContextAtomicO
   List<CloudProvider> cloudProviders
 
   @Override
+  AtomicOperationConverter getAtomicOperationConverter(String description, String cloudProvider) {
+    // Legacy naming convention which is not generic and description name is specific to cloud provider
+    try {
+      AtomicOperationConverter converter = super.getAtomicOperationConverter(description, cloudProvider)
+      if (converter) return converter
+    } catch (NoSuchBeanDefinitionException e) {
+      /**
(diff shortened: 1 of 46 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AnnotationsBasedAtomicOperationsRegistry.groovy
```diff
@@ -37,6 +37,82 @@ class AnnotationsBasedAtomicOperationsRegistry extends ApplicationContextAtomicO
   List<CloudProvider> cloudProviders
 
   @Override
+  AtomicOperationConverter getAtomicOperationConverter(String description, String cloudProvider) {
```
maggieneterval, line 40: This looks like a huge diff but I actually just duplicated the existing body of the now-deprecated `getAtomicOperationConverter` minus the `providerVersion` logic since I'm going to delete the deprecated version next Clouddriver release. Didn't seem worth optimizing for minimal duplicated code for one release.

maggieneterval:
@claymccoy @robzienert Just a heads up that I'll aim to remove the handful of now-deprecated methods that use `ProviderVersion` next Clouddriver release. Let me know if there's anything else I can do that would be helpful from the perspective of a plugin maintainer other than documenting these methods as deprecated for one Clouddriver release.

robzienert:
@maggieneterval Love it. This is good, thanks.

PR #4723, opened 2020-07-06
feat(aws): Support IPv6 for launch template backed ASGs

jeyrschabu:
- assign IPv6 address to instances when `associateIpV6Address` is true

@caseyhebebrand I decided not to overload `associatePublicIpAddress` for IPv6 since that creates a public IPv4 address.
Instead I added `associateIpV6Address` that can be set independently of `associatePublicIpAddress` and avoid coupling the two.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/services/LaunchTemplateService.java
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/services/LaunchTemplateService.java
@@ -138,7 +138,7 @@ public class LaunchTemplateService {
     request.withNetworkInterfaces(
         new LaunchTemplateInstanceNetworkInterfaceSpecificationRequest()
             .withAssociatePublicIpAddress(settings.getAssociatePublicIpAddress())
-            .withIpv6AddressCount(associateIPv6Address != null && associateIPv6Address ? 1 : 0)
+            .withIpv6AddressCount(associateIPv6Address ? 1 : 0)
             .withGroups(settings.getSecurityGroups())
             .withDeviceIndex(0));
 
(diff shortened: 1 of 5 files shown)


caseyhebebrand:
> @caseyhebebrand I decided not to overload `associatePublicIpAddress` for IPv6 since that creates a public IPv4 address.
> Instead I added `associateIpV6Address` that can be set independently of `associatePublicIpAddress` and avoid coupling the two.

Sounds good. I will update the naming in `deck` to `associateIpv6Address`

jeyrschabu:
> _Very_ minor but can we change the name to `associateIPv6Address` instead of `associateIpV6Address` (swapping casing of the `P` and `v`)? This is just more consistent with `IPv6` standalone casing.

Yup, changing that.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/services/LaunchTemplateService.java
```diff
@@ -135,6 +138,7 @@ public LaunchTemplate createLaunchTemplate(
     request.withNetworkInterfaces(
         new LaunchTemplateInstanceNetworkInterfaceSpecificationRequest()
             .withAssociatePublicIpAddress(settings.getAssociatePublicIpAddress())
+            .withIpv6AddressCount(associateIPv6Address != null && associateIPv6Address ? 1 : 0)
```
cfieber, line 141: associateIPv6Address is defaulted above to false so I don't think this will ever be null

jeyrschabu: Good catch. Compulsive null checker

PR #4733, opened 2020-07-09
fix(eureka): Skip instances that are already UP in discovery, or still starting

luispollo:
This PR attempts to address an issue reported at Netflix whereby instances that are still coming up are evaluated for discovery health and so cause overall server group state to be incorrectly determined during deployments.

Changes:
- Based on a suggestion by @dreynaud, modify the behavior of `AbstractEurekaSupport` to skip instances that are not explicitly in an `OUT_OF_SERVICE` status, i.e. instances that are still starting or that already in an `UP` status, when attempting to set the instance status to `UP`.
- Renamed `DiscoveryStatus` enum names to more sensible ones (IMO).


clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/deploy/ops/AbstractEurekaSupport.groovy
```diff
@@ -336,8 +351,8 @@ abstract class AbstractEurekaSupport {
   }
 
   enum DiscoveryStatus {
-    Enable('UP'),
-    Disable('OUT_OF_SERVICE')
+    UP('UP'),
+    DOWN('OUT_OF_SERVICE')
```
dreynaud, line 355: `DiscoveryStatus.UP` makes sense, but `DiscoveryStatus.DOWN` not so much. DOWN and OUT_OF_SERVICE are 2 different Discovery statuses, DOWN means unhealthy (red) whereas OUT_OF_SERVICE means healthy but disabled (gray).

If it's because OUT_OF_SERVICE is too long, `DiscoveryStatus.OOS` might be ok?

luispollo: Oh, thanks for clarifying. I'll rename to `OUT_OF_SERVICE`.

clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/deploy/ops/AbstractEurekaSupport.groovy
```diff
@@ -139,18 +142,26 @@ abstract class AbstractEurekaSupport {
...
-            resp = eureka.resetInstanceStatus(applicationName, instanceId, DiscoveryStatus.Disable.value)
+            // If we're trying to set the status to UP, and the instance is already UP (or hasn't registered yet),
+            // skip it.
+            if (instanceDetails.instance.status != DiscoveryStatus.OUT_OF_SERVICE.value) {
+              log.debug("Instance ${instanceId} is already UP in discovery, skipping.")
+              skipped.add(instanceId)
+              return
+            } else {
+              resp = eureka.resetInstanceStatus(applicationName, instanceId, DiscoveryStatus.OUT_OF_SERVICE.value)
+            }
```
dreynaud, line 156: Minor: I had to read this a couple times to parse the double negative ("if not disabled then skip, else do the thing") so my preference would be to flip the branches.

Also changed the log message a little to provide the actual status.

```suggestion
            if (instanceDetails.instance.status == DiscoveryStatus.OUT_OF_SERVICE.value) {
              resp = eureka.resetInstanceStatus(applicationName, instanceId, DiscoveryStatus.OUT_OF_SERVICE.value)
            } else {
              log.debug("Instance {} is {} in discovery, no override to remove", instanceId, instanceDetails.instance.status)
              skipped.add(instanceId)
              return
            }
```

luispollo: Makes sense. BTW, why do we need to reset the status to `OUT_OF_SERVICE` when the instance is already marked `OUT_OF_SERVICE`? What exactly does that achieve?

dreynaud: resetInstanceStatus *deletes* the `OUT_OF_SERVICE` override

PR #4737, opened 2020-07-10
chore(*): Bump kork; associated refactors

robzienert:


clouddriver-core/clouddriver-core.gradle
```diff
@@ -15,6 +15,7 @@ dependencies {
   // which appears to require consumers to have core groovy on the classpath
   api "org.codehaus.groovy:groovy"
 
+  implementation "javax.inject:javax.inject:1"
```
cfieber, line 18: I think there is a eureka-client dependency in here that should be removable

robzienert: Removed.

PR #4739, opened 2020-07-10
fix(misc): add rxjava dependency which was available before transitively

srekapalli:
rx Api's are used in `NetworkController`. Adding the dependency explicitly which is no longer available transitively.


cfieber:
dupe of #4738

PR #4743, opened 2020-07-13
fix(titus): Add ability to override eurekaName and eurekaRegion for t…

aravindmd:
Titus would like us to connect to a different `region` with a different `eurekaName` for selected regions in few accounts , Adding the ability to override those values per region.
This in turn will be used in to assemble the grpc channel correctly. If both values are present, then we’d be able to use that in preference over the account-defined `eurekaName` and it’s current logic.


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/TitusRegion.java
```diff
@@ -69,6 +73,16 @@ public TitusRegion(
...
+    if (eurekaName == null) {
+      this.eurekaName = null;
+    } else {
+      this.eurekaName = eurekaName;
+    }
+    if (eurekaRegion == null) {
+      this.eurekaRegion = null;
+    } else {
+      this.eurekaRegion = eurekaRegion;
+    }
```
jeyrschabu, line 85: What's the reasoning here. It looks like this should just be set, right? Am i missing anything?

jeyrschabu:
```suggestion
this.eurekaName = eurekaName;
this.eurekaRegion = eurekaRegion;
```

aravindmd: fixed this , i was setting defaults and was handling it this way , not required actually

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/config/TitusConfiguration.groovy
```diff
@@ -63,7 +63,9 @@ class TitusConfiguration {
     List<NetflixTitusCredentials> accounts = new ArrayList<>()
     for (TitusCredentialsConfig.Account account in titusCredentialsConfig.accounts) {
       List<TitusRegion> regions = account.regions.collect {
-        new TitusRegion(it.name, account.name, it.endpoint, it.autoscalingEnabled, it.loadBalancingEnabled, it.applicationName, it.url, it.port, it.featureFlags)
+        String eurekaName = it.eurekaName == null ? null : it.eurekaName
+        String eurekaRegion = it.eurekaRegion == null ? null: it.eurekaRegion
```
jeyrschabu, line 67: It seems like this should just be passed in? Not following the null check if you are setting it to null.

aravindmd: yes fixed it , it was the related to the other comment , had defaults set initially and did not clean up after.

PR #4753, opened 2020-07-21
fix(kubernetes): Fixes to ensure CRD events are reported in deploy stage

harikumarks:
In case of kubernetes accounts with `liveManifestCalls` enabled,  the events of the CRD are not shown in `Deployment ( Manifest) stage` ->`Deploy Status` tab when deploy is in process.  This is because,  in case of CRDs, KubectlJobexecutor.eventsFor() makes kubectl get events command and filter using the field selector `involvedObject.name=kindName.apiGroupName` instead of just `kindName`  (as we are using KubernetesKind.toString()) .  This fix addresses the issue and ensure that we use only kindName . ApiGroup level filtering is done once the available events are fetched by kubectl.

Refer https://github.com/spinnaker/spinnaker/issues/5871 ( issue 1)


harikumarks:
@ezimanyi please review with reference to  https://github.com/spinnaker/spinnaker/issues/5871

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesV2ClusterProvider.java
```diff
@@ -256,7 +256,7 @@ public boolean supportsMinimalClusters() {
       serverGroupToServerGroupManagerKeys.put(
           serverGroupDatum.getId(),
           kindMap.translateSpinnakerKind(SERVER_GROUP_MANAGERS).stream()
-              .map(kind -> serverGroupDatum.getRelationships().get(kind.toString()))
+              .map(kind -> serverGroupDatum.getRelationships().get(kind.getName()))
```
ezimanyi, line 259: Looking at the code more closely, I'm not sure that we should need to make this change.

It looks like the code to store the relationships in the cache is calling `getGroup` on the key:
https://github.com/spinnaker/clouddriver/blob/cfaec246181f207fa46c906d9461c4ae81a6ed6f/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesCacheData.java#L118

And `getGroup` is itself calling `toString` on the `kind`:
https://github.com/spinnaker/clouddriver/blob/cfaec246181f207fa46c906d9461c4ae81a6ed6f/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/Keys.java#L221

So I think we should be using `kind.toString()` as they key when we store the relationships in the cache, and thus should also use that when retrieving.

harikumarks: @ezimanyi agreed. This may not be the best way to fix it. I was not able to figure out the reason this is happening. So for now i have reverted the second fix. Hope you can approve the first fix ( which should be the only one present now)

ezimanyi: Sounds good, we can merge this, and then separately figure out why the cache is not including the fully-scoped name.  Thanks!

ezimanyi:
@harikumarks : While doing some testing around unrelated changes, I noticed that this PR is breaking display of events for some kinds, particularly `Pod`.

I believe the problem is that there are some subtle things to consider around:
* kinds which have moved API groups (ex: `ReplicaSet` from `extensions` to `apps`)
* kinds without an API group (ex: `Pod`); we currently put that in an API group called `core`.  But that means that we're expecting `core` in the `InvolvedObject.ApiGroup` but it's just `v1`.  I suspect this should be fixed by just making our `KubernetesApiGroup.CORE` have string `""` instead of `"core"`.  (Though this is a bit risky because then we'd need to delete the matching `KubernetesApiGroup.NONE`.)

Another possible fix is to do something like this:
```
KubernetesKind.from(
                        x.getInvolvedObject().getOrDefault("kind", ""),
                        KubernetesApiVersion.fromString(
                                x.getInvolvedObject().getOrDefault("apiVersion", ""))
                            .getApiGroup())
                    .equals(kind))
```

The reason that works is that the `equals()` comparator for `KubernetesKind` considers two kinds equal if they are both from built-in API groups.  But this also creates a lot of objects in that filtering stream so maybe we should consider performance?  (Or maybe it's negligible.)

In any case, this is going to be a bit more subtle that it first seemed, and we should also add tests along with any re-apply.  Let's definitely discuss more how to address this, but given that this isn't a really simple fix, I think reverting before 1.22 is the right decision.

harikumarks:
@ezimanyi 
Thanks for taking care of this. We should definitely discuss more.
Regards
Hari

PR #4755, opened 2020-07-23
fix(cats/sql): Fixes CloudDriver SQL Cache from failing due to id or agent being too long to be stored. 

mattsanta:
* fix(cats/sql): Reverting temporary fix when agent is too long for SQL table. 

  This temporary fix can still have issues when the type in the agent is longer than 127 characters, so I'm reverting it in favor of a proper long-term fix. This reverts commit fae18a1ad60582210a8158717aef248698ec817e.

* fix(cats/sql): Adding cats v2 template tables. 

* fix(cats/sql): Fixes CloudDriver SQL Cache from failing due to id or agent being too long to be stored. 

  Fixes github.com/spinnaker/spinnaker/issues/5600. The theoretical max length of an id or agent is much larger than we can store in columns that are primary or index keys. Therefore, in order to solve the issue I moved away from storing the id and agent in key columns. I've introduced new columns that will be used as keys that store hashes of the ids and agents instead. I made no changes to the contracts between the SQL Cache and those who depend on it, all the changes are internal to the SQL Cache.

  I also took the liberty of adding more javadoc style comments to some of the functions in the SQL Cache. Even though some of them are simple, I still think it's beneficial to introduce it since if the functions do become more complex then whoever updates them will need to update the comment.

* fix(cats/sql): Add drop tables by version to CATS SQL Admin Controller.


mattsanta:
DO NOT MERGE! I still need to implement an admin endpoint that operators can use to delete the v1 tables. I put the PR up so that I can get some feedback on the majority of the work while I figure out the admin endpoint stuff.

@ncknt, my PR includes reverting the changes that you made in https://github.com/spinnaker/clouddriver/pull/4704.

@robzienert, something that I discovered while implementing and didn't notice during the design phase is that there is a function whose performance may be impacted depending on how it's used. filterIdentifiers does a query with a like on the id, for obvious reasons we can't use the id hash for this query. Like queries can benefit from indexes depending on where the wildcard is present, if it's at the beginning then an index does nothing but if it's at the end then it an index can increase the performance of the query. Thoughts on how big of an issue this would be? If we really need an index we could make an index on the id just for this case, but then the highest character limit we could configure for the id would be 768 characters with this implementation.

mattsanta:
Added another commit that adds the ability to drop tables by version to the cats sql admin controller. Is this what you were referring to when you recommended adding an admin endpoint, @robzienert ? Still relatively new to Spinnaker so sorry if this isn't the right place and is completely wrong. Also to be honest I can't find how to run the commands in here, do you know how to?

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -649,123 +683,240 @@ class SqlCache(
...
+              it.id_hash,
+              it.id,
+              it.rel_id_hash,
+              it.rel_id,
+              it.rel_agent_hash,
+              it.rel_agent,
+              it.rel_type,
+              now
+            )
+            ulid = ULID().nextMonotonicValue(ulid)
```
ezimanyi, line 871: I think this is pre-existing, but I believe this is creating a new `ULID` object for each iteration of the loop (which from the implementation looks to also create a new `SecureRandom` for each iteration; this likely creates a lot of unecessary garbage.

I'd suggest pulling the `ULID` at least out of the loop, and potentially into an instance variable (if it's thread-safe).

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -893,57 +1022,115 @@ class SqlCache(
...
+   * @param type the type of the relationship table.
+   * @param relAgentHash the rel agent hash of the records whose identifying fields are retrieved.
+   * @return list of relationship entries with uuid, id hash, rel id hash, and rel agent hash
+   * populated.
+   */
+  private fun getRelationshipKeys(type: String, relAgentHash: String): MutableList<RelEntry> {
     return withRetry(RetryCategory.READ) {
       jooq
-        .select(field("uuid"), field("id"), field("rel_id"), field("rel_agent"))
+        .select(field("uuid"), field("id_hash"), field("rel_id_hash"), field("rel_agent_hash"))
```
ezimanyi, line 1100: See comment above for more details, but if we return the full `rel_agent` here instead of just the hash, we can avoid constructing the reverse-lookup table above.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -78,38 +80,14 @@ class SqlCache(
...
-    }
-
-    cacheMetrics.evict(
-      prefix = name,
-      type = type,
-      itemCount = ids.size,
-      itemsDeleted = deletedCount,
-      deleteOperations = opCount
-    )
+    val hashedIds = ids.asSequence().map { getIdHash(it) }.toSet()
```
maggieneterval, line 89: _Potentially dumb kotlin question alert:_ What's the purpose of the `asSequence` and `toSet` calls here if `evictAllInternal` just accepts `Collection<String>`?

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -230,12 +208,13 @@ class SqlCache(
       return mutableListOf()
     }
 
+    val hashedIds = ids.asSequence().map { getIdHash(it) }.toSet()
```
maggieneterval, line 211: Same question here re: `toSet`-ing this when `getDataWith/WithoutRelationships` only requires a `Collection`?

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -1286,16 +1513,79 @@ class SqlCache(
...
+
+    log.info("evicting ${hashedIds.size} $type records")
+
+    var deletedCount = 0
+    var opCount = 0
+    try {
+      hashedIds.chunked(
+        dynamicConfigService.getConfig(
+          Int::class.java,
+          "sql.cache.read-batch-size",
```
ezimanyi, line 1553: Wouldn't this want to use `write-batch-size`?  (Feels like delete is closer to write, in the absence of a separate `delete-batch-size` config.)

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -649,123 +683,240 @@ class SqlCache(
...
 
-    val sourceAgents = items.filter { it.relationships.isNotEmpty() }
+    val relAgentHashToRelAgent = mutableMapOf<String, String>()
+    val relAgentHashes = items.asSequence()
+      .filter { it.relationships.isNotEmpty() }
       .map { it.relationships.keys }
       .flatten()
+      .map {
+        val relAgentHash = getAgentHash(it)
+        relAgentHashToRelAgent[relAgentHash] = it
```
maggieneterval, line 710: Maybe this is not weird in kotlin world, but feels a little strange to have this side effect as part of the map callback.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/controllers/CatsSqlAdminController.kt
```diff
@@ -102,7 +93,39 @@ class CatsSqlAdminController(
     return CleanTablesResult(tableCount = tablesDropped.size, tables = tablesDropped)
   }
 
-  private fun validateParams(currentNamespace: String?, targetNamespace: String) {
+  @PutMapping(path = ["drop_version/{version}"])
+  fun dropTablesByVersion(@PathVariable("version") dropVersion: String): CleanTablesResult {
+    validatePermissions()
+
+    val conn = getConnection()
+    val sql = "show tables like 'cats_v$dropVersion%"
```
ezimanyi, line 101: Two things here:
(1) Even on an admin endpoint I'd avoid directly interpolating user input into a SQL query like this.  Does JOOQ support binding a variable into this query?
(2) What happens when we're on version 10 and someone tries to drop version 1?

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -893,57 +1022,115 @@ class SqlCache(
...
+    return getSha256Hash(id)
+  }
+
+  /**
+   * Returns a hash of the agent given.
+   *
+   * @param agent the agent to hash.
+   * @return hash of the agent.
+   */
+  private fun getAgentHash(agent: String): String {
```
maggieneterval, line 1054: Is this a separate method from `getIdHash(id: String)` just in case someday we want to hash agents differently?

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/controllers/CatsSqlAdminController.kt
```diff
@@ -102,7 +93,39 @@ class CatsSqlAdminController(
     return CleanTablesResult(tableCount = tablesDropped.size, tables = tablesDropped)
   }
 
-  private fun validateParams(currentNamespace: String?, targetNamespace: String) {
+  @PutMapping(path = ["drop_version/{version}"])
```
maggieneterval, line 96: > Also to be honest I can't find how to run the commands in here, do you know how to?

I might be misunderstanding your question from [this comment](https://github.com/spinnaker/clouddriver/pull/4755#issuecomment-663562679) but I think you should be able to just curl a PUT request against `{your_clouddriver}/admin/db/drop_version/{version}` to test this. Sorry if that is not in fact what you were asking!

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -893,57 +1022,115 @@ class SqlCache(
...
 
-  private fun getHashIds(type: String, agent: String?): List<HashId> {
+  /**
+   * Gets a SHA-256 hash value for a given String.
+   *
+   * @param str the String to hash.
+   * @return the hashed value.
+   */
+  private fun getSha256Hash(str: String): String {
+    return Hashing.sha256().hashBytes(str.toByteArray()).toString()
```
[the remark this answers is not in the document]
maggieneterval: Wow, I completely forgot about #3838, great call. 🎉 🕵️

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
maggieneterval: Oh I'm going to just resolve this and my other related comment as I see @ezimanyi already asked the same question! (Sorry, I was halfway done reviewing this when he posted his review and didn't bother looking at his comments first...)

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/controllers/CatsSqlAdminController.kt
```diff
@@ -102,7 +93,43 @@ class CatsSqlAdminController(
...
+  fun dropTablesByVersion(@PathVariable("version") dropVersion: String): CleanTablesResult {
```
robzienert, line 97: You can make this the enum type as a parameter and then Spring will do the validation work for you. I'd feel more comfortable using the enum as the path variable type so we don't have any unwelcome SQL injection regressions in the future.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/controllers/CatsSqlAdminController.kt
```diff
@@ -102,7 +93,43 @@ class CatsSqlAdminController(
...
+
+      while (rs.next()) {
+        val table = rs.getString(1)
+        val dropSql = "drop table `$table`"
+        log.info("Dropping $table")
+
+        jooq.query(dropSql).execute()
+        tablesDropped.add(table)
+      }
+    }
```
robzienert, line 119: Please keep persistence logic out of controllers - move this code to a Repository class (say, `SqlAdminCommandsRepository` or something) and call that instead.

robzienert:
> Added another commit that adds the ability to drop tables by version to the cats sql admin controller. Is this what you were referring to when you recommended adding an admin endpoint, @robzienert ?

Yes it was, thanks!

robzienert:
Deployed this into both test and staging environments; got some stacktraces that need to be fixed prior to this PR being merged.

```json
  {
    "@timestamp": "2020-07-27T23:30:49.564+00:00",
    "@version": 1,
    "message": "Internal Server Error",
    "logger_name": "com.netflix.spinnaker.kork.web.exceptions.GenericExceptionHandlers",
    "thread_name": "https-jsse-nio-7004-exec-9",
    "level": "ERROR",
    "level_value": 40000,
    "stack_trace": "java.lang.NullPointerException: null",
    "X-SPINNAKER-REQUEST-ID": "01EE99S3JS9M0F596A67CGTDP0:01EE99S5W8P5VWWKWS3EG2DTDY:1:86063f31-7942-4fea-8444-2a44f832ae67",
    "X-SPINNAKER-USER": "anonymous",
    "X-SPINNAKER-ACCOUNTS": "REDACTED",
    "X-SPINNAKER-APPLICATION": "spindemo",
    "X-SPINNAKER-EXECUTION-ID": "01EE99S3JS9M0F596A67CGTDP0:01EE99S5W8P5VWWKWS3EG2DTDY:1"
  }
```

These were happening across all of our clouddriver clusters, but predominantly on our readonly clusters, which caused our test pipelines (which are running constantly) to start failing. I'm afraid there's not really any other information here that I can provide, but the rate was high.

```json
{
  "@timestamp": "2020-07-27T23:30:33.644+00:00",
  "@version": 1,
  "message": "Unable to fetch clusters for project 'REDACTED'",
  "logger_name": "com.netflix.spinnaker.clouddriver.core.ProjectClustersService",
  "thread_name": "AgentExecutionAction-141",
  "level": "ERROR",
  "level_value": 40000,
  "stack_trace": "java.lang.NullPointerException: null
    at com.google.common.hash.AbstractHashFunction.hashUnencodedChars(AbstractHashFunction.java:38)
    at com.netflix.spinnaker.cats.sql.cache.SqlCache.getSha256Hash(SqlCache.kt:1057)
    at com.netflix.spinnaker.cats.sql.cache.SqlCache.getIdHash(SqlCache.kt:1037)
    at com.netflix.spinnaker.cats.sql.cache.SqlCache.getAll(SqlCache.kt:211)
    at com.netflix.spinnaker.cats.sql.SqlProviderCache.getAll(SqlProviderCache.kt:78)
    at com.netflix.spinnaker.cats.cache.CompositeCache.getAll(CompositeCache.java:115)
    at com.netflix.spinnaker.cats.cache.CompositeCache.getAll(CompositeCache.java:107)
    at com.netflix.spinnaker.cats.cache.Cache$getAll$6.call(Unknown Source)
    at com.netflix.spinnaker.clouddriver.aws.provider.view.AmazonClusterProvider.translateServerGroups(AmazonClusterProvider.groovy:389)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at org.codehaus.groovy.runtime.callsite.PlainObjectMetaMethodSite.doInvoke(PlainObjectMetaMethodSite.java:43)
    at org.codehaus.groovy.runtime.callsite.PogoMetaMethodSite$PogoCachedMethodSiteNoUnwrapNoCoerce.invoke(PogoMetaMethodSite.java:190)
    at org.codehaus.groovy.runtime.callsite.PogoMetaMethodSite.callCurrent(PogoMetaMethodSite.java:58)
    at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callCurrent(AbstractCallSite.java:176)
    at com.netflix.spinnaker.clouddriver.aws.provider.view.AmazonClusterProvider.translateClusters(AmazonClusterProvider.groovy:239)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at org.codehaus.groovy.runtime.callsite.PlainObjectMetaMethodSite.doInvoke(PlainObjectMetaMethodSite.java:43)
    at org.codehaus.groovy.runtime.callsite.PogoMetaMethodSite$PogoCachedMethodSiteNoUnwrapNoCoerce.invoke(PogoMetaMethodSite.java:190)
    at org.codehaus.groovy.runtime.callsite.PogoMetaMethodSite.callCurrent(PogoMetaMethodSite.java:58)
    at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callCurrent(AbstractCallSite.java:176)
    at com.netflix.spinnaker.clouddriver.aws.provider.view.AmazonClusterProvider.getCluster(AmazonClusterProvider.groovy:515)
    at com.netflix.spinnaker.clouddriver.aws.provider.view.AmazonClusterProvider.getCluster(AmazonClusterProvider.groovy:521)
    at com.netflix.spinnaker.clouddriver.aws.provider.view.AmazonClusterProvider.getCluster(AmazonClusterProvider.groovy)
    at com.netflix.spinnaker.clouddriver.core.ProjectClustersService.lambda$null$4(ProjectClustersService.java:187)
    at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
    at java.util.HashMap$KeySpliterator.forEachRemaining(HashMap.java:1556)
    at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
    at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
    at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
    at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:566)
    at com.netflix.spinnaker.clouddriver.core.ProjectClustersService.lambda$retrieveClusters$5(ProjectClustersService.java:189)
    at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
    at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1382)
    at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
    at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
    at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
    at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:566)
    at com.netflix.spinnaker.clouddriver.core.ProjectClustersService.retrieveClusters(ProjectClustersService.java:194)
    at com.netflix.spinnaker.clouddriver.core.ProjectClustersService.retrieveClusters(ProjectClustersService.java:135)
    at com.netflix.spinnaker.clouddriver.core.ProjectClustersService.getProjectClusters(ProjectClustersService.java:103)
    at com.netflix.spinnaker.clouddriver.core.ProjectClustersService.getProjectClusters(ProjectClustersService.java:73)
    at com.netflix.spinnaker.clouddriver.core.agent.ProjectClustersCachingAgent.loadData(ProjectClustersCachingAgent.java:70)
    at com.netflix.spinnaker.cats.agent.CachingAgent$CacheExecution.executeAgentWithoutStore(CachingAgent.java:87)
    at com.netflix.spinnaker.cats.agent.CachingAgent$CacheExecution.executeAgent(CachingAgent.java:77)
    at com.netflix.spinnaker.cats.sql.cluster.AgentExecutionAction.execute(SqlClusteredAgentScheduler.kt:298)
    at com.netflix.spinnaker.cats.sql.cluster.AgentJob.run(SqlClusteredAgentScheduler.kt:317)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
    at java.util.concurrent.FutureTask.run(FutureTask.java:266)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
    at java.lang.Thread.run(Thread.java:748)
"
}
```

```json
{
  "@timestamp": "2020-07-27T23:29:42.492+00:00",
  "@version": 1,
  "message": "Internal Server Error",
  "logger_name": "com.netflix.spinnaker.kork.web.exceptions.GenericExceptionHandlers",
  "thread_name": "https-jsse-nio-7004-exec-7",
  "level": "ERROR",
  "level_value": 40000,
  "stack_trace": "java.lang.NullPointerException: null
    at com.google.common.hash.AbstractHashFunction.hashUnencodedChars(AbstractHashFunction.java:38)
    at com.netflix.spinnaker.cats.sql.cache.SqlCache.getSha256Hash(SqlCache.kt:1057)
    at com.netflix.spinnaker.cats.sql.cache.SqlCache.getIdHash(SqlCache.kt:1037)
    at com.netflix.spinnaker.cats.sql.cache.SqlCache.getAll(SqlCache.kt:211)
    at com.netflix.spinnaker.cats.sql.SqlProviderCache.getAll(SqlProviderCache.kt:78)
    at com.netflix.spinnaker.cats.cache.CompositeCache.getAll(CompositeCache.java:115)
    at com.netflix.spinnaker.cats.cache.CompositeCache.getAll(CompositeCache.java:107)
    at com.netflix.spinnaker.cats.cache.Cache$getAll$3.call(Unknown Source)
    at com.netflix.spinnaker.clouddriver.aws.provider.view.AmazonClusterProvider.translateServerGroups(AmazonClusterProvider.groovy:389)
    at sun.reflect.GeneratedMethodAccessor1503.invoke(Unknown Source)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at org.codehaus.groovy.runtime.callsite.PlainObjectMetaMethodSite.doInvoke(PlainObjectMetaMethodSite.java:43)
    at org.codehaus.groovy.runtime.callsite.PogoMetaMethodSite$PogoCachedMethodSiteNoUnwrapNoCoerce.invoke(PogoMetaMethodSite.java:190)
    at org.codehaus.groovy.runtime.callsite.PogoMetaMethodSite.callCurrent(PogoMetaMethodSite.java:58)
    at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callCurrent(AbstractCallSite.java:176)
    at com.netflix.spinnaker.clouddriver.aws.provider.view.AmazonClusterProvider.translateClusters(AmazonClusterProvider.groovy:239)
    at sun.reflect.GeneratedMethodAccessor1502.invoke(Unknown Source)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at org.codehaus.groovy.runtime.callsite.PlainObjectMetaMethodSite.doInvoke(PlainObjectMetaMethodSite.java:43)
    at org.codehaus.groovy.runtime.callsite.PogoMetaMethodSite$PogoCachedMethodSiteNoUnwrapNoCoerce.invoke(PogoMetaMethodSite.java:190)
    at org.codehaus.groovy.runtime.callsite.PogoMetaMethodSite.callCurrent(PogoMetaMethodSite.java:58)
    at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callCurrent(AbstractCallSite.java:176)
    at com.netflix.spinnaker.clouddriver.aws.provider.view.AmazonClusterProvider.getCluster(AmazonClusterProvider.groovy:515)
    at com.netflix.spinnaker.clouddriver.aws.provider.view.AmazonClusterProvider.getCluster(AmazonClusterProvider.groovy)
    at com.netflix.spinnaker.clouddriver.model.ClusterProvider$getCluster$0.call(Unknown Source)
    at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:47)
    at com.netflix.spinnaker.clouddriver.model.ClusterProvider$getCluster$0.call(Unknown Source)
    at com.netflix.spinnaker.clouddriver.controllers.ClusterController$_getForAccountAndName_closure4$_closure17.doCall(ClusterController.groovy:123)
    at com.netflix.spinnaker.clouddriver.controllers.ClusterController$_getForAccountAndName_closure4$_closure17.doCall(ClusterController.groovy)
    at sun.reflect.GeneratedMethodAccessor1015.invoke(Unknown Source)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:101)
    at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:323)
    at org.codehaus.groovy.runtime.metaclass.ClosureMetaClass.invokeMethod(ClosureMetaClass.java:263)
    at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1041)
    at groovy.lang.Closure.call(Closure.java:405)
    at groovy.lang.Closure.call(Closure.java:399)
    at com.netflix.spinnaker.clouddriver.requestqueue.RequestQueue$NOOP.execute(RequestQueue.java:97)
    at com.netflix.spinnaker.clouddriver.requestqueue.RequestQueue.execute(RequestQueue.java:80)
    at com.netflix.spinnaker.clouddriver.requestqueue.RequestQueue$execute$0.call(Unknown Source)
    at com.netflix.spinnaker.clouddriver.controllers.ClusterController$_getForAccountAndName_closure4.doCall(ClusterController.groovy:123)
    at sun.reflect.GeneratedMethodAccessor1012.invoke(Unknown Source)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:101)
    at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:323)
    at org.codehaus.groovy.runtime.metaclass.ClosureMetaClass.invokeMethod(ClosureMetaClass.java:263)
    at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1041)
    at groovy.lang.Closure.call(Closure.java:405)
    at groovy.lang.Closure.call(Closure.java:421)
    at org.codehaus.groovy.runtime.DefaultGroovyMethods.collect(DefaultGroovyMethods.java:3574)
    at org.codehaus.groovy.runtime.DefaultGroovyMethods.collect(DefaultGroovyMethods.java:3559)
    at org.codehaus.groovy.runtime.DefaultGroovyMethods.collect(DefaultGroovyMethods.java:3659)
    at org.codehaus.groovy.runtime.dgm$87.invoke(Unknown Source)
    at org.codehaus.groovy.runtime.callsite.PojoMetaMethodSite$PojoMetaMethodSiteNoUnwrapNoCoerce.invoke(PojoMetaMethodSite.java:244)
    at org.codehaus.groovy.runtime.callsite.PojoMetaMethodSite.call(PojoMetaMethodSite.java:53)
    at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:127)
    at com.netflix.spinnaker.clouddriver.controllers.ClusterController.getForAccountAndName(ClusterController.groovy:121)
    at com.netflix.spinnaker.clouddriver.controllers.ClusterController.getForAccountAndNameAndType(ClusterController.groovy:140)
    at com.netflix.spinnaker.clouddriver.controllers.ClusterController$$FastClassBySpringCGLIB$$d4103cfa.invoke(<generated>)
    at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218)
    at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:769)
    at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
    at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:747)
    at org.springframework.security.access.intercept.aopalliance.MethodSecurityInterceptor.invoke(MethodSecurityInterceptor.java:69)
    at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
    at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:747)
    at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:689)
    at com.netflix.spinnaker.clouddriver.controllers.ClusterController$$EnhancerBySpringCGLIB$$34f5027d.getForAccountAndNameAndType(<generated>)
    at sun.reflect.GeneratedMethodAccessor1300.invoke(Unknown Source)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:190)
    at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:138)
    at org.springframework.web.servlet.mvc.method.annotation.ServletInvocableHandlerMethod.invokeAndHandle(ServletInvocableHandlerMethod.java:106)
    at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.invokeHandlerMethod(RequestMappingHandlerAdapter.java:879)
    at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.handleInternal(RequestMappingHandlerAdapter.java:793)
    at org.springframework.web.servlet.mvc.method.AbstractHandlerMethodAdapter.handle(AbstractHandlerMethodAdapter.java:87)
    at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1040)
    at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:943)
    at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1006)
    at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:898)
    at javax.servlet.http.HttpServlet.service(HttpServlet.java:634)
    at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:883)
    at javax.servlet.http.HttpServlet.service(HttpServlet.java:741)
    at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:231)
    at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
    at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:53)
    at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
    at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
    at org.springframework.web.filter.ShallowEtagHeaderFilter.doFilterInternal(ShallowEtagHeaderFilter.java:104)
    at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:119)
    at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
    at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
    at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:320)
    at org.springframework.security.web.access.ExceptionTranslationFilter.doFilter(ExceptionTranslationFilter.java:118)
    at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
    at org.springframework.security.web.authentication.AnonymousAuthenticationFilter.doFilter(AnonymousAuthenticationFilter.java:111)
    at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
    at com.netflix.spinnaker.fiat.shared.FiatAuthenticationFilter.doFilter(FiatAuthenticationFilter.java:65)
    at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
    at org.springframework.security.web.servletapi.SecurityContextHolderAwareRequestFilter.doFilter(SecurityContextHolderAwareRequestFilter.java:158)
    at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
    at org.springframework.security.web.FilterChainProxy.doFilterInternal(FilterChainProxy.java:215)
    at org.springframework.security.web.FilterChainProxy.doFilter(FilterChainProxy.java:178)
    at org.springframework.web.filter.DelegatingFilterProxy.invokeDelegate(DelegatingFilterProxy.java:358)
    at org.springframework.web.filter.DelegatingFilterProxy.doFilter(DelegatingFilterProxy.java:271)
    at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
    at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
    at org.springframework.web.filter.RequestContextFilter.doFilterInternal(RequestContextFilter.java:100)
    at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:119)
    at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
    at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
    at org.springframework.web.filter.FormContentFilter.doFilterInternal(FormContentFilter.java:93)
    at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:119)
    at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
    at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
    at org.springframework.boot.actuate.metrics.web.servlet.WebMvcMetricsFilter.doFilterInternal(WebMvcMetricsFilter.java:109)
    at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:119)
    at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
    at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
    at com.netflix.spinnaker.platform.netflix.security.auth.MetatronAuthenticationFilter.doFilter(MetatronAuthenticationFilter.java:114)
    at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
    at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
    at org.springframework.web.filter.CharacterEncodingFilter.doFilterInternal(CharacterEncodingFilter.java:201)
    at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:119)
    at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
    at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
    at javax.servlet.FilterChain$doFilter$43.call(Unknown Source)
    at com.netflix.spinnaker.filters.AuthenticatedRequestFilter.doFilter(AuthenticatedRequestFilter.groovy:144)
    at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
    at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
    at org.apache.catalina.core.StandardWrapperValve.invoke(StandardWrapperValve.java:202)
    at org.apache.catalina.core.StandardContextValve.invoke(StandardContextValve.java:96)
    at org.apache.catalina.authenticator.AuthenticatorBase.invoke(AuthenticatorBase.java:526)
    at org.apache.catalina.core.StandardHostValve.invoke(StandardHostValve.java:139)
    at org.apache.catalina.valves.ErrorReportValve.invoke(ErrorReportValve.java:92)
    at org.apache.catalina.core.StandardEngineValve.invoke(StandardEngineValve.java:74)
    at org.apache.catalina.connector.CoyoteAdapter.service(CoyoteAdapter.java:343)
    at org.apache.coyote.http11.Http11Processor.service(Http11Processor.java:367)
    at org.apache.coyote.AbstractProcessorLight.process(AbstractProcessorLight.java:65)
    at org.apache.coyote.AbstractProtocol$ConnectionHandler.process(AbstractProtocol.java:860)
    at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1591)
    at org.apache.tomcat.util.net.SocketProcessorBase.run(SocketProcessorBase.java:49)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
    at org.apache.tomcat.util.threads.TaskThread$WrappingRunnable.run(TaskThread.java:61)
    at java.lang.Thread.run(Thread.java:748)
",
  "X-SPINNAKER-REQUEST-ID": "01EE98MJPMDV8EMHY23JK2W26T:01EE99RCPVGGV8G6PK1JXQKD30:1:07757ca5-4d86-48f3-96a7-42e91ee0dd18",
  "X-SPINNAKER-USER": "anonymous",
  "X-SPINNAKER-ACCOUNTS": "REDACTED",
  "X-SPINNAKER-APPLICATION": "spindemo",
  "X-SPINNAKER-EXECUTION-ID": "01EE98MJPMDV8EMHY23JK2W26T:01EE99RCPVGGV8G6PK1JXQKD30:1"
}
```

These last two stacktraces are definitely easier.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/controllers/CatsSqlAdminController.kt
ezimanyi: By binding a variable, I mean that in SQL supports distinguishing what is part of the query syntax from what is a value that is passed to that syntax (and this support is included in most libraries for sending SQL queries, including JOOQ).

So for example, if I wanted to fetch all documents with `owner = 'eric'`, one way to send the query would be:
```
val user = "eric"
jooq.fetch("select * from document where owner = '$user'")
```

Alternatively, you could pull the variable into a bind, and send:
```
jooq.fetch("select * from document where owner = ?", $user)
```

For security reasons, the second of these is much better, because SQL understands that the value is not part of the query and does not consider it when parsing the query.  So even if you passed unsanitized user input as a bind, it could not perform SQL injection because the query itself is fixed, it's only the bind variable that can change.

In the first case, if `user` ends up being set to `eric' or 1=1` then the query will return all rows from the database.  In the second case, it would return all rows whose owner is equal to the string `eric' or 1=1` which is what the code intends (and foils the malicious attempt to dump out the whole database).

The other reason binds are better is that they often lead to better performance, because the database needs to parse fewer queries.  Without binds, the database needs to separately parse the text of `select * from document where owner = 'eric'` and `select * from document where owner = 'matt'` (and presumably the many other versions of this query if it exists in my application logic).  With binds, it only needs to parse the single query `select * from document where owner = ?` and can cache the execution plan for the query (basically compiled database code) and then re-use that compiled plan every time it runs with a different bind variable.

Neither of these are the most important thing in this case, as it's a lightly-used page that only admins can access, but those are the general reasons why using bind variables is good practice.  (And the reasons why I get nervous when I see queries that don't use binds.)

I think here something like `jooq.execute("show tables like", "cats_v${drop_version}_%")` would be the way to bind it.

mattsanta:
Thanks for the stack traces @robzienert. I took a closer look at the [Kotlin null safety documentation ](https://kotlinlang.org/docs/reference/null-safety.html) and it seems like these errors are due to issue with java interoperation, specifically:
```
Generic types used for Java interoperation with incorrect nullability, e.g. a piece of Java code might add null into a Kotlin MutableList<String>, meaning that MutableList<String?> should be used for working with it
```
I wasn't aware that this could happen (first time Kotlin user). I'll make the necessary code changes in the next commit.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/controllers/CatsSqlAdminController.kt
mattsanta: Thank you so much for the informative response. I'll update to make use of the [fetch with bindings](https://www.jooq.org/javadoc/latest/org.jooq/org/jooq/DSLContext.html#fetch(java.lang.String,java.lang.Object...)) and while I'm at it update the other 2 exposed admin endpoints to do the same.

robzienert:
Yeah, there are some rough edges around Java interop; it's why we've been encouraging people to add the `@Nonnull`, `@Nullable`, etc annotations to interfaces written in Java so that IDEs (and Kotlin) can be better informed about nullability.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/controllers/CatsSqlAdminController.kt
mattsanta: Sounds good, done. Thanks!

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/controllers/CatsSqlAdminController.kt
mattsanta: Done. I also went ahead and moved all the other persistence logic that was in this controller over to ```SqlAdminCommandsRepository```.

mattsanta:
Pushed another commit that hopefully fixes the NPEs 🤞 . I had my changes running in my own deployment of Spinnaker with a Kubernetes provider configured for ~20 minutes and there were no NPEs in the Clouddriver logs.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/controllers/CatsSqlAdminController.kt
mattsanta: This was exactly what I was looking for, thank you! Though when I try and run it I get an unauthorized error. Any idea on what else I need to run these endpoints?

Command I ran: ```curl -X PUT http://localhost:7002/admin/db/drop_version/V1```

Response: ```{"error":"Internal Server Error","message":"Unauthorized","status":500,"timestamp":"2020-07-28T21:30:38.334+00:00"}```

maggieneterval: Likely because it is checking whether you are an [admin](https://spinnaker.io/setup/security/admin/#manually-add-configuration-in-fiat)? If you don't already have fiat configured, could also just test it on your LocalGit-deployed dev spinnaker with the perm checks commented out.

mattsanta: Thanks, I don't have fiat configured so I went with commenting out the permission check in my localgit deployed spinnaker to test and it works.

robzienert:
> @robzienert, something that I discovered while implementing and didn't notice during the design phase is that there is a function whose performance may be impacted depending on how it's used. filterIdentifiers does a query with a like on the id, for obvious reasons we can't use the id hash for this query. Like queries can benefit from indexes depending on where the wildcard is present, if it's at the beginning then an index does nothing but if it's at the end then it an index can increase the performance of the query. Thoughts on how big of an issue this would be? If we really need an index we could make an index on the id just for this case, but then the highest character limit we could configure for the id would be 768 characters with this implementation.

Just saw this, sorry. I think we'll need to add an index for `id` alone; 768 char limit seems more than enough.

`filterIdentifiers` is used basically everywhere, so its performance is, unfortunately, fairly critical. The good news is that no use of `filterIdentifiers` uses a glob at the very beginning, but any particular invocation of `filterIdentifiers` has a high chance to use multiple globs.

The existing cache API is pretty hostile towards SQL. When we were initially building out the SQL backend, we made note that we could get way, way better performance if we were to rewrite the cache API to be more friendly to a relational model, but that was never done as it hasn't been high enough priority for us (read: performance today is sufficient).

mattsanta:
Thanks for the response, @robzienert. I've gone ahead and added another commit that introduces an index on id for the resource tables. It's unfortunate that we're still a bit restricted, but I believe we've at least significantly reduced the chances of a user running into this issue. We've more than doubled the acceptable max length of an id and are free to change the agent max length without any schema issues moving forward.

maggieneterval:
@mattsanta, was thinking that this is probably a change that deserves mention in the 1.22 curated changelog: would you mind making a PR against spinnaker.github.io for [this page](https://spinnaker.io/community/releases/next-release-preview/) including instructions on how operators use the admin endpoint you added? We try to add entires here for any notable changes, particularly those that require user intervention to upgrade. Thanks!

mattsanta:
Absolutely, I'll do that. Thanks for the tip!

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlAdminCommandsRepository.kt
```diff
@@ -0,0 +1,126 @@
...
+   * @return String collection of the tables dropped.
+   */
+  fun dropTablesByNamespace(dropNamespace: String): Collection<String> {
+    val conn = getConnection()
+    val tablesDropped = mutableListOf<String>()
+
+    conn.use { c ->
+      val jooq = DSL.using(c, SQLDialect.MYSQL)
+      val rs =
+        jooq.fetch("show tables like ?", "cats_v${SqlSchemaVersion.current()}_${dropNamespace}_%")
```
robzienert, line 76: Same goes here, `dropNamespace` will need some treatment before we use it in a query.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlAdminCommandsRepository.kt
```diff
@@ -0,0 +1,126 @@
...
+  fun truncateTablesByNamespace(truncateNamespace: String): Collection<String> {
+    val conn = getConnection()
+    val tablesTruncated = mutableListOf<String>()
+
+    conn.use { c ->
+      val jooq = DSL.using(c, SQLDialect.MYSQL)
+      val rs =
+        jooq.fetch(
+          "show tables like ?",
+          "cats_v${SqlSchemaVersion.current()}_${truncateNamespace}_%"
```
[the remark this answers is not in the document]
mattsanta: Could you specify what you'd like to see for validation? FYI, this and dropTablesByNamespace were present beforehand, I just moved them out of the controller.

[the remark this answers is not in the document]
robzienert: Oh, I see. Just escaping the input should be fine, or having what you have here, but using a prepared statement.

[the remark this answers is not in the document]
mattsanta: Please correct me if I'm wrong, but the way the DSLContext (jooq variable) is being created results in prepared statements being used at execution. Reason being that ```DSL.using(...)``` defaults to prepared statements when no Settings are provided. See: https://www.jooq.org/doc/3.13/manual/sql-execution/statement-type/

robzienert:
So, good news. We accidentally deployed this into production and the world didn't burn down. Some quick notes: Query latency is slightly (1-3%) improved at the cost of ~5% DB CPU increase - but everything else has stayed roughly the same as far as performance metrics go.

mattsanta:
Well that's a bit terrifying that it was accidentally deployed to prod, but glad nothing went wrong! Can we move forward with merging this PR then?

mattsanta:
I wonder if a part of the increase in CPU could have been due to all the new tables that needed to be created. 🤔

robzienert:
It was a constant CPU increase and the old tables went idle. I'm not sure more tables would account for a constant CPU increase. At any rate, I don't think it moves the needle enough that it needs to worry.

robzienert:
@mattsanta I'm seeing a major spike in clouddriver instance CPU usage with this change that I don't recall seeing before merging.

clouddriver-cpu-util-before-after

And correlated deadlocking on the database.

deadlocks

We're investigating, but wanted to raise this - hopefully you can help!

maggieneterval:
Looping in @german-muzquiz since he is planning to cut the 1.22 release branches today -- not sure if the best path forward here is to revert this for now until we can resolve the CPU and deadlocking spikes, but it's German's call as the release manager.

german-muzquiz:
I agree with @maggieneterval in which this PR would need to be reverted before cutting the new branches.

mattsanta:
I'll send a PR for reverting this change until we can figure out what happened.

mattsanta:
Thanks for the heads up, Rob. Please let me know if you find anything since you have much more to work with then I do in this case.

Initial question for you:
What percentage of the production environment was this change deployed to, by accident, before it was merged? I'm assuming it wasn't 100% and the issue didn't show until merged then deployed to 100%, or some much larger % than initially.

robzienert:
# deployment

That's kinda the odd thing - we didn't see alerts for this fire in `main` - but really should have. That's something we're going to be investigating separately: We were fully deployed, although not for long.

`prestaging` and `main` have the same footprint, but `prestaging` sometimes runs a little hotter, since we have fewer instances allocated. I suspect we tripped our alerts for `prestaging` because we're running fewer instances there. On the other hand, we also get a lot less traffic (`prestaging` is almost all automated traffic).

# SQL

The deadlocking is an issue, so I did a little bit of sleuthing last night. I haven't really gone deep here yet, but I do have two early observations:

1. Relationship queries. For example `select uuid, id_hash, rel_id_hash, rel_agent_hash, rel_agent from cats_v2_a_health_rel where (rel_agent_hash = 'daa88ac042399031090aa5bd06dbe1f332a799e3a480e38e6bb08b1889a7b18e' and rel_type = 'instances')` could use a composite index over `rel_agent_hash` and `rel_type`. The query optimizer will currently use the `rel_agent_hash` index, then do a full scan over that to filter out `rel_type`. For something like `health` (Eureka), that's somewhat costly. Changing `KEY `rel_agent_hash_idx` (`rel_agent_hash`)` so it includes `rel_type` second would do the trick.
2. Pagination. When writing cache data, it looks like we take the entire delta and insert it a single, massive query. I'm noticing some non-trivial lock times for these types of queries. I think it may be better to split them up and submit over a series of queries. I suspect the deadlocks may have something to do with this, but I haven't scanned through your PR again to verify. I think there's pagination on writes already, but maybe it needs to be tuned down.

# Caching agent CPU

As for the caching agents CPU spiking, I ran a [Flame report](https://github.com/Netflix/flamescope) on this. I've attached the report for you if you want to take a poke-about as well, here's the interesting bit:

Screen Shot 2020-08-18 at 10 06 58 AM

We're spending a ton of CPU time hashing: `SqlCache.storeAuthoratative` method: ~54% of CPU time! Granted, this is for body hashes, but, I think we need to start using a faster hash algo across the board in `SqlCache`. Since it's a new schema, we'll be able to make this change easily since everyone will be re-caching everything anyway.

[flamegraph.html.zip](https://github.com/spinnaker/clouddriver/files/5091745/flamegraph.html.zip)

I'll continue looking into this.

robzienert:
We've rolled back `prestaging` just now. I'll be doing another CPU profile and also investigating the delta on the SQL perf insights to see if something else changed. That should help us narrow things down.

mattsanta:
With respect to the **SQL** section:

1. Adding the ```rel_type``` to an index with the ```rel_agent_hash``` makes sense. I do want to point out though that in v1 this query exists with ```rel_agent``` and ```rel_type``` and there is no index with both of those. So adding this index should be an improvement on v1.

2. You're right that there is pagination on writes at the moment. I've gone through each one of them and the amount of entries that we're writing in each execution is the same as it was in v1 except for one instance where we are deleting...
* The amount of entries written per execution is based on ```sql.cache-write-batch-size``` in the config otherwise it defaults to 100. Could the config value have been changed/removed resulting in more entries per write than before?
** Diff storing in resource tables ([link](https://github.com/spinnaker/clouddriver/pull/4755/files#diff-dcf591a5203ab9c7a95960990f550450R580))
** Diff storing in rel tables 1 ([link](https://github.com/spinnaker/clouddriver/pull/4755/files#diff-dcf591a5203ab9c7a95960990f550450R828))
** Diff storing in rel tables 2 ([link](https://github.com/spinnaker/clouddriver/pull/4755/files#diff-dcf591a5203ab9c7a95960990f550450R876))
* With that said, the one instance where the number of entry changes per execution has changed is when we're deleting all entries for a set of ids. Previously the deletes per execution was based on ```sql.cache.read-batch-size``` with a default value of 500. I had it changed to be based on ```sql.cache.write-batch-size``` since deletes are more like writes than reads and set a default of 300, though maybe that should have been 100 to match up with the other write defaults.
** Diff deleting from resource table ([link](https://github.com/spinnaker/clouddriver/pull/4755/files#diff-dcf591a5203ab9c7a95960990f550450R1548))

mattsanta:
One question about changing the hashing algo in SqlCache: From a security perspective is it an issue that the body would be hashed with a non-cryptographic hash function? Not to sure what information could be revealed here.

mattsanta:
Can't get the flamegraph file you uploaded to work for me. I've installed and have FlameGraph running per instructions on https://github.com/Netflix/flamescope. I unzipped the file and added it to the examples directory, but when I try to open it in the UI I get an error. Let me know if I'm doing something wrong.

robzienert:
- Re: SQL indices: Yeah, not going to change anything, but I thought I'd mention it while I saw it. 👍 
- Our pagination hasn't changed. I guess the query I saw just had really big records.
- Hashing algo: No need for cryptographic hashing. The body that is hashed is stored plaintext in the same table; the body hash is used as an optimization so we're not writing records back into the database that have not changed.
- Looks like the flamecommander I uploaded was corrupted somehow. You should just be able to open it, I've re-attached it.
[clouddriver-flamegraph.html.zip](https://github.com/spinnaker/clouddriver/files/5097818/clouddriver-flamegraph.html.zip)

mattsanta:
* I'm going to start a new PR to undo the revert and apply improvements that we discover, including changing the index to include ```rel_type```.
* Out of curiosity, if you look over a larger period of time are you seeing what you've defined as "non-trivial" lock times with the v1 implementations as well? Since it should be similar to v2 from looking at the code.
* Ah right...my bad, that was a dumb question. Thanks for pointing it out :)! We should definitely switch hashing algos then, I'll add this to future PR.
* Awesome, this one works! It may be beneficial if I can compare it to flamegraph of sql cache with v1 implementation. Would you be able to provide that as well? I'll work on getting up to speed with flamegraph.

robzienert:
- deadlocks: No, very infrequently. I've attached two pictures. Prestaging is missing a lot of data because we didn't have performance insights enabled initially. You can see where we rolled back. For comparison, I included main as well.
- flame for v1: I'll get this once I'm done working out!

prestaging

main

robzienert:
Here's the V1 flamegraph. We're still spending a lot of time in hashing in v1. I think a cheaper hash is going to still give us a lot of bang for the buck, but I think the deadlocks are probably our killer. I'm available to deploy to our envs for testing as much as we need to!

Also, it's fairly one-sided right now as far as you doing the work. I'm trying to clear out my current tasks so I can help dig in more.

mattsanta:
Was working on something else the last few days, but I'm back to putting effort into solving this performance issue. I've went ahead and created a draft PR with the improvements we've discussed and will continue to add on as we find other issues:
https://github.com/spinnaker/clouddriver/pull/4834

Also, the v1 flamegraph that you provided is the same as the v2 flamegraph, I think you uploaded the wrong one by accident.

mattsanta:
I've gone through the changes made in this PR and the only significant difference with the DB interaction is how we're writing into the relationship tables in ```storeAuthoritative```.

Previously we iterated through the rel types and stored the forward and reverse relationships for that rel type. I changed this behavior to storing all the forward relationships first because they're all going into the same table. Then iterating through the rel types and storing the reverse relationships since the rel type determines the table used for the reverse relationships. The diff can be seen [here](https://github.com/spinnaker/clouddriver/pull/4755/files#diff-dcf591a5203ab9c7a95960990f550450L752).

This should be a performance improvement since it allows properly batching the forward relationships. With the previous implementation the forward relationships batches could be small thus requiring more db requests to write into the table.

---
@robzienert Are you able to see examples of which queries are deadlocking each other?

robzienert:
Continuing the discussion over at the new PR: https://github.com/spinnaker/clouddriver/pull/4834.

PR #4758, opened 2020-07-27
fix(aws): Skip AMI Sharing and Tag Syncing if Owner is not Managed by…

evansb:
Fixes https://github.com/spinnaker/spinnaker/issues/5920

This changes `AllowLaunchAtomicOperation` which is used to ensure AMI is shared with the target account so that it can be launched.

I this PR there are three accounts to consider

1. Source Account: Clouddriver managing account
2. Target Account: Clouddriver managed account
3. Owner Account: Owner of the AMI

**Previously**, when the credentials of the Owner account is not found, it will throw an exception and fail the task.
However, it is possible that that Owner AMI is not managed by Spinnaker and I believe we should not fail on such cases.

**Now**, if the owner account is not managed by Spinnaker, it will skip both image sharing & tag syncing. If the image can be resolved in target account, sharing is skipped (this is current behavior). Tag syncing will also need because without the owner's credentials, neither of the accounts assumed to have the permission to read the original AMI's tags.

It will still throw an exception if the AMI is found in the source account but we cannot find the owner credentials to share it to the target account.

Credit to @mfreebairn-r7  for analyzing the root cause of the problem.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AllowLaunchAtomicOperation.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AllowLaunchAtomicOperation.groovy
@@ -94,12 +94,12 @@ class AllowLaunchAtomicOperation implements AtomicOperation<ResolvedAmiResult> {
     }
 
     if (amiLocation == ResolvedAmiLocation.TARGET && !ownerCredentials) {
-      task.updateStatus BASE_PHASE, "AMI found in target account and owner not managed: skipping allow launch and tag syncing"
+      task.updateStatus BASE_PHASE, "AMI found in target account, but the AMI owner account is unmanaged: skipping allow launch and tag syncing"
       return resolvedAmi
     }
 
     if (amiLocation == ResolvedAmiLocation.SOURCE && !ownerCredentials) {
-      task.updateStatus BASE_PHASE, "AMI found in source account but owner is not managed, unable to share AMI"
+      task.updateStatus BASE_PHASE, "AMI found in source account, but the owner account is unmanaged: unable to share AMI"
       throw new IllegalArgumentException("Unable to find owner of resolved AMI $resolvedAmi")
     }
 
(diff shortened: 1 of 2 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AllowLaunchAtomicOperation.groovy
```diff
@@ -77,18 +77,30 @@ class AllowLaunchAtomicOperation implements AtomicOperation<ResolvedAmiResult> {
...
-      if (!sourceCredentials) {
-        throw new IllegalArgumentException("Unable to find owner of resolved AMI $resolvedAmi")
+      if (ownerCredentials) {
+        ownerAmazonEC2 = amazonClientProvider.getAmazonEC2(ownerCredentials, description.region, true)
       }
-      sourceAmazonEC2 = amazonClientProvider.getAmazonEC2(sourceCredentials, description.region, true)
+    }
+
+    if (amiLocation == ResolvedAmiLocation.TARGET && !ownerCredentials) {
+      task.updateStatus BASE_PHASE, "AMI found in target account and owner not managed: skipping allow launch and tag syncing"
```
robzienert, line 97:
```suggestion
      task.updateStatus BASE_PHASE, "AMI found in target account, but the AMI owner account is unmanaged: skipping allow launch and tag syncing"
```

This was a little confusing to read.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/AllowLaunchAtomicOperation.groovy
```diff
@@ -77,18 +77,30 @@ class AllowLaunchAtomicOperation implements AtomicOperation<ResolvedAmiResult> {
...
+      return resolvedAmi
+    }
+
+    if (amiLocation == ResolvedAmiLocation.SOURCE && !ownerCredentials) {
+      task.updateStatus BASE_PHASE, "AMI found in source account but owner is not managed, unable to share AMI"
```
robzienert, line 102:
```suggestion
      task.updateStatus BASE_PHASE, "AMI found in source account, but the owner account is unmanaged: unable to share AMI"
```

Adjusting to follow similar changes above.

PR #4759, opened 2020-07-27
test(kubernetes): Add integration tests on caching views

ezimanyi:
There are currently almost no tests on any of the data providers in the Kubernetes provider. Before doing significant work in this area, we should add some test coverage to validate that the upcoming changes don't break anything.

As a lot of the classes are tightly-coupled without clearly defined interfaces between them, I decided to write an integration test that runs the caching agent for a few resources, then calls the data providers to ensure that the resulting data is as expected.

This tests the contract that should mostly stay consistent during the upcoming refactor. Even if we change what data is stored in the cache, we'll need to make sure that these providers still return the same data after a caching cycle is run.

Eventually we should write more unit tests, once there are stronger contracts between these providers and between them and the cache. But at this point I think a test at this level will give the most confidence that the upcoming changes don't break anything.

This is a very large class, as I only wanted to set up the data once, and the various providers all return different view of the same data so they can share some of the assert functions. Maybe there's a better way to split this up, but for the moment this seemed reasonable.

I used SoftAssertions so that you get a report of all the failures for each test at the end (rather than failing at the first issue) which should make debugging easier.

Finally, I had to make a few classes public. These aren't really intended to be used widely, but it's impossible to appropriately create a new KubernetesV2Credentials without them. Eventually we could/should make it easier to create said credentials for testing purposes, but for now this seemed better than making the test less realistic.  (Also as these are Spring components, and Spring can find package-private things, they are in a sense a bit exposed already. In fact the reason we need them exposed is because we're wiring up things as Spring would on startup.)

### DIFF
--- a/clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
+++ b/clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
@@ -239,6 +239,7 @@ final class KubernetesDataProviderIntegrationTest {
 
   @Test
   void getSingleCluster(SoftAssertions softly) {
+    // When not explicitly passing the includeDetails flag, it should default to true.
     KubernetesV2Cluster cluster =
         clusterProvider.getCluster("frontendapp", ACCOUNT_NAME, "deployment frontend");
     assertThat(cluster).isNotNull();
@@ -251,6 +252,14 @@ final class KubernetesDataProviderIntegrationTest {
         clusterProvider.getCluster("frontendapp", ACCOUNT_NAME, "deployment frontend", true);
     assertThat(cluster).isNotNull();
     assertFrontendCluster(softly, cluster, false);
+  }
+
+  @Test
+  void getSingleClusterWithoutDetails(SoftAssertions softly) {
+    KubernetesV2Cluster cluster =
+        clusterProvider.getCluster("frontendapp", ACCOUNT_NAME, "deployment frontend", false);
+    assertThat(cluster).isNotNull();
+    assertFrontendCluster(softly, cluster, true);
   }
 
   @Test
@@ -547,8 +556,9 @@ final class KubernetesDataProviderIntegrationTest {
       // TODO(ezimanyi): The same load balancer is being returned multiple times (likely due to
       // finding all relationships to it). This should be fixed to return it only once.
       softly.assertThat(cluster.getLoadBalancers()).hasSize(4);
-      assertFrontendLoadBalancer(
-          softly, (KubernetesV2LoadBalancer) cluster.getLoadBalancers().iterator().next());
+      if (!cluster.getLoadBalancers().isEmpty())
+        assertFrontendLoadBalancer(
+            softly, (KubernetesV2LoadBalancer) cluster.getLoadBalancers().iterator().next());
     }
   }
 
@@ -633,8 +643,6 @@ final class KubernetesDataProviderIntegrationTest {
     softly.assertThat(serverGroup.getInstanceCounts().getTotal()).isEqualTo(2);
     softly.assertThat(serverGroup.getLoadBalancers()).containsExactly("service frontend");
     softly.assertThat(serverGroup.getUid()).isEqualTo("29630998-bdee-4586-ac64-45223d7ef7d5");
-    // When using a deployment, the prior server group is not disabled as labels aren't changed;
-    // instead this server group is scaled down to 0 instances.
     softly.assertThat(serverGroup.isDisabled()).isFalse();
     softly.assertThat(serverGroup.getRegion()).isEqualTo("frontend-ns");
     softly.assertThat(serverGroup.getServerGroupManagers()).hasSize(1);
--- a/clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
+++ b/clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
@@ -190,9 +190,9 @@ final class KubernetesDataProviderIntegrationTest {
         clusters.stream().collect(toImmutableMap(KubernetesV2Cluster::getName, c -> c));
 
     assertFrontendServiceCluster(softly, clusterLookup.get("service frontend"));
-    assertFrontendCluster(softly, clusterLookup.get("deployment frontend"), false);
+    assertFrontendCluster(softly, clusterLookup.get("deployment frontend"), true);
     assertBackendServiceCluster(softly, clusterLookup.get("service backendlb"));
-    assertBackendCluster(softly, clusterLookup.get("replicaSet backend"), false);
+    assertBackendCluster(softly, clusterLookup.get("replicaSet backend"), true);
   }
 
   @Test
@@ -212,7 +212,7 @@ final class KubernetesDataProviderIntegrationTest {
         clusters.stream().collect(toImmutableMap(KubernetesV2Cluster::getName, c -> c));
 
     assertBackendServiceCluster(softly, clusterLookup.get("service backendlb"));
-    assertBackendCluster(softly, clusterLookup.get("replicaSet backend"), false);
+    assertBackendCluster(softly, clusterLookup.get("replicaSet backend"), true);
   }
 
   @Test
@@ -228,7 +228,7 @@ final class KubernetesDataProviderIntegrationTest {
         clusters.stream().collect(toImmutableMap(KubernetesV2Cluster::getName, c -> c));
 
     assertBackendServiceCluster(softly, clusterLookup.get("service backendlb"));
-    assertBackendCluster(softly, clusterLookup.get("replicaSet backend"), false);
+    assertBackendCluster(softly, clusterLookup.get("replicaSet backend"), true);
   }
 
   @Test
@@ -243,7 +243,7 @@ final class KubernetesDataProviderIntegrationTest {
     KubernetesV2Cluster cluster =
         clusterProvider.getCluster("frontendapp", ACCOUNT_NAME, "deployment frontend");
     assertThat(cluster).isNotNull();
-    assertFrontendCluster(softly, cluster, false);
+    assertFrontendCluster(softly, cluster, true);
   }
 
   @Test
@@ -251,7 +251,7 @@ final class KubernetesDataProviderIntegrationTest {
     KubernetesV2Cluster cluster =
         clusterProvider.getCluster("frontendapp", ACCOUNT_NAME, "deployment frontend", true);
     assertThat(cluster).isNotNull();
-    assertFrontendCluster(softly, cluster, false);
+    assertFrontendCluster(softly, cluster, true);
   }
 
   @Test
@@ -259,7 +259,7 @@ final class KubernetesDataProviderIntegrationTest {
     KubernetesV2Cluster cluster =
         clusterProvider.getCluster("frontendapp", ACCOUNT_NAME, "deployment frontend", false);
     assertThat(cluster).isNotNull();
-    assertFrontendCluster(softly, cluster, true);
+    assertFrontendCluster(softly, cluster, false);
   }
 
   @Test
@@ -287,7 +287,7 @@ final class KubernetesDataProviderIntegrationTest {
         clusters.stream().collect(toImmutableMap(KubernetesV2Cluster::getName, c -> c));
 
     assertBackendServiceCluster(softly, clusterLookup.get("service backendlb"));
-    assertBackendCluster(softly, clusterLookup.get("replicaSet backend"), true);
+    assertBackendCluster(softly, clusterLookup.get("replicaSet backend"), false);
   }
 
   @Test
@@ -536,7 +536,7 @@ final class KubernetesDataProviderIntegrationTest {
   }
 
   private void assertFrontendCluster(
-      SoftAssertions softly, KubernetesV2Cluster cluster, boolean summary) {
+      SoftAssertions softly, KubernetesV2Cluster cluster, boolean includeDetails) {
     softly.assertThat(cluster.getMoniker().getApp()).isEqualTo("frontendapp");
     softly.assertThat(cluster.getMoniker().getCluster()).isEqualTo("deployment frontend");
     softly.assertThat(cluster.getType()).isEqualTo("kubernetes");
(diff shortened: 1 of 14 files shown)


clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
```diff
@@ -0,0 +1,1117 @@
...
+import org.assertj.core.api.SoftAssertions;
+import org.assertj.core.api.junit.jupiter.SoftAssertionsExtension;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.platform.runner.JUnitPlatform;
+import org.junit.runner.RunWith;
+import org.mockito.internal.stubbing.defaultanswers.ReturnsSmartNulls;
+
+@ExtendWith(SoftAssertionsExtension.class)
```
dbyron0, line 94: You no doubt pondered this, but if the goal is to wire things up the way spring does, why not use e.g. `@ExtendWith(SpringExtension.class)` like https://github.com/spinnaker/echo/blob/ac270b219511fb16dd142ec32d2aa8b8a961cd6b/echo-web/src/test/java/com/netflix/spinnaker/echo/telemetry/TelemetrySpec.java does?  Maybe some `@Autowired` simplifies getting the objects required for testing?

ezimanyi: Thanks for the pointer!  In this case I want a bit more control over how each of these beans are built than is possible (or that I've found easy to do with spring integration tests).  For example, to keep things simple I only wanted it to know about 4 of the `*Handler` classes rather than everything on the classpath.

In some sense, this is aspirational as well...in that I don't think I *should* need ~50 lines of code to get a functioning `KubernetesV2Credentials`, so hopefully this will get simpler with future refactors (whereas moving to use Spring would be somewhat admitting defeat).

But I do agree given how complex it is now, it's 50/50 whether wiring this up manually vs. using spring would have been easier.

dbyron0: Oooph.  I just looked at [KubernetesV2Credentials.Factory](https://github.com/spinnaker/clouddriver/blob/36235837d452f251a4f50c6bf3d876e6eabdec9c/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesV2Credentials.java#L690).  No fun.  Yeah, I agree doing it the way you did will it make it more clear if things get simpler...and gives you more control.

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
```diff
@@ -0,0 +1,1117 @@
...
+    Map<String, Set<KubernetesV2Cluster>> results = clusterProvider.getClusters();
+    assertThat(results).hasSize(1);
+    assertThat(results).containsKey(ACCOUNT_NAME);
+
+    Set<KubernetesV2Cluster> clusters = results.get(ACCOUNT_NAME);
+    assertThat(clusters).hasSize(4);
+
+    // TODO(ezimanyi): I don't think it's correct that we are returning services/load balancers as
+    // clusters. I believe that they are just being ignored by deck, and should be removed from
+    // the result of this call (likely with performance benefits). In any case, as seen in the
```
maggieneterval, line 181: 👍  Agree that only entities mapped to Spinnaker server groups should be returned with clusters. We are even [documenting](https://spinnaker.io/concepts/#cluster) this as such already.

However, some users may expect this to work when using dynamic target selection in, for instance, a Delete or Patch (Manifest) stage, which currently permits selection of any registered kind when choosing a target dynamically. The Spinnaker-cluster concept is a bit more intuitive when configuring Enable/Disable/Scale stages, which work for server group types only, but likely pretty unintuitive for Patch/Delete, so perhaps we can rename the "Choose a target dynamically," which implies that one can choose _any_ target dynamically, to "Choose a target dynamically by cluster" or something, and then only hydrate the resulting Kinds list with Kinds that actually have cluster relationships.

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
```diff
@@ -0,0 +1,1117 @@
...
+    assertThat(cluster).isNotNull();
+    assertFrontendCluster(softly, cluster, false);
+  }
+
+  @Test
+  void getSingleClusterWithDetails(SoftAssertions softly) {
+    KubernetesV2Cluster cluster =
+        clusterProvider.getCluster("frontendapp", ACCOUNT_NAME, "deployment frontend", true);
+    assertThat(cluster).isNotNull();
+    assertFrontendCluster(softly, cluster, false);
```
maggieneterval, line 253: Why is this the same assertion as the non-details return on line 245? Is the `summary` bool not actually relevant to whether or not details are included?

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
```diff
@@ -0,0 +1,1117 @@
...
+    assertBackendPriorServerGroup(softly, serverGroup);
+  }
+
+  @Test
+  void getServerGroupWithoutDetails(SoftAssertions softly) {
+    KubernetesV2ServerGroup serverGroup =
+        clusterProvider.getServerGroup(
+            ACCOUNT_NAME, "backend-ns", "replicaSet backend-v014", false);
+    assertThat(serverGroup).isNotNull();
+    // Looks like we ignore the includeDetails flag, so this is the same serverGroup as when we do
```
maggieneterval, line 313: Oh, I think this answers my previous question -- same deal for clusters?

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
```diff
@@ -0,0 +1,1117 @@
...
+    KubernetesV2Application backendApplication = applicationLookup.get("backendapp");
+    softly.assertThat(frontendApplication).isNotNull();
+    if (frontendApplication != null) {
+      assertBackendApplication(softly, backendApplication);
+    }
+  }
+
+  @Test
+  void getApplicationsExpanded(SoftAssertions softly) {
+    // This is the same as the unexpanded test, as it seems like we ignore the flag.
```
maggieneterval, line 342: Any context on why we even have these flags? Curious if other cloud providers use them.

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
```diff
@@ -0,0 +1,1117 @@
...
+    assertFrontendLoadBalancer(softly, results.iterator().next());
+  }
+
+  @Test
+  void getLoadBalancersByName(SoftAssertions softly) {
+    List<Details> results =
+        loadBalancerProvider.byAccountAndRegionAndName(
+            ACCOUNT_NAME, "frontend-ns", "service frontend");
+    // TODO(ezimanyi): This is a bug; we are not properly looking up the load balancer, so always
+    // return null
```
maggieneterval, line 403: 😭  Oh boy, maybe this is why we make a live call for a service by name/namespace [here](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/manifest/KubernetesDeployManifestOperation.java#L262).

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
```diff
@@ -0,0 +1,1117 @@
...
+    softly.assertThat(serverGroup.getAccount()).isEqualTo("my-account");
+    softly.assertThat(serverGroup.getAccountName()).isEqualTo("my-account");
+    softly.assertThat(serverGroup.getKind()).isEqualTo(KubernetesKind.REPLICA_SET);
+    softly.assertThat(serverGroup.getName()).isEqualTo("replicaSet frontend-5c6559f75f");
+    softly.assertThat(serverGroup.getInstanceCounts().getUp()).isEqualTo(2);
+    softly.assertThat(serverGroup.getInstanceCounts().getTotal()).isEqualTo(2);
+    softly.assertThat(serverGroup.getLoadBalancers()).containsExactly("service frontend");
+    softly.assertThat(serverGroup.getUid()).isEqualTo("29630998-bdee-4586-ac64-45223d7ef7d5");
+    // When using a deployment, the prior server group is not disabled as labels aren't changed;
+    // instead this server group is scaled down to 0 instances.
```
maggieneterval, line 637: Can we remove this comment since it only applies to the prior SG?

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
```diff
@@ -0,0 +1,1117 @@
...
+  private static KubernetesV2ClusterProvider clusterProvider =
+      new KubernetesV2ClusterProvider(cacheUtils, kindMap);
+  private static KubernetesV2InstanceProvider instanceProvider =
+      new KubernetesV2InstanceProvider(cacheUtils, accountResolver);
+  private static KubernetesV2LoadBalancerProvider loadBalancerProvider =
+      new KubernetesV2LoadBalancerProvider(cacheUtils, kindMap);
+  private static KubernetesV2SearchProvider searchProvider =
+      new KubernetesV2SearchProvider(cacheUtils, kindMap, objectMapper, accountResolver);
+  private static KubernetesV2ServerGroupManagerProvider serverGroupManagerProvider =
+      new KubernetesV2ServerGroupManagerProvider(cacheUtils);
```
maggieneterval, line 158: Was about to ask if we're explicitly not asserting on entities that map to Spinnaker security groups, then remembered that Deck actually populates the non-details "Firewalls" view using the search endpoint rather than a /securityGroups endpoint. Also, given that this was broken forever and almost no one noticed, likely not worth adding any assertions on NetworkPolicies...

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
ezimanyi: Good catch, it looks like I didn't actually test this function when passing `includeDetails = false`.  This is the same as the test above because the test above doesn't pass the `includeDetails` flag at all (and the behavior is to default to having it `true`).

I've made two changes:
* Added a test when passing `includeDetails = false` so that now there are 3 (`true`, `false`, and omitted)
* I inverted the boolean in `assertFrontendCluster` so instead of passing `summary = false` when you want details you pass `includeDetails = true` so it matches the behavior of the function we're testing.  This hopefully makes it less confusing now that the function and the assertion have a flag with the same meaning.

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
ezimanyi: Added a comment above, but in the above case I was missing a test; for clusters the flag is respected.

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
ezimanyi: Luckily it's [easy to fix](https://github.com/spinnaker/clouddriver/blob/2b78b4693dccce18b87e7fea20dd32100099a2c1/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesV2LoadBalancerProvider.java#L91), I just didn't include that in this PR.

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
ezimanyi: Good question; I'd imagine that some other providers do (or at least one) if they were added to the interface, but I'm actually not sure.  This could also be fertile ground for performance improvements if it turns out that we're returning way more data than we need because we don't respect these flags.

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
ezimanyi: Removed!

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
ezimanyi: The reason I didn't add tests on `KubernetesV2SecurityGroupProvider` is because as far as I can tell security groups are completely independent from the other types.  It probably would be good to add some tests on it but I think those can be a completely separate test that just deploys some `NetworkPolicies` alone and then makes some calls (to avoid adding more logic to this test).  My first round of refactoring doesn't affect security groups much which is why I didn't add the test now, but probably should in the near future.

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
ezimanyi: Great point!  It looks like [getClusterAndSortAscending](https://github.com/spinnaker/clouddriver/blob/2b78b4693dccce18b87e7fea20dd32100099a2c1/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesV2ManifestProvider.java#L116) is expecting a relationship between the cluster and the kind that is being deployed.  We currently only map a few kinds to `cluster` which are (except for `Service` and `NetworkPolicy`) all server-group type things.  So this currently only works for a subset of kinds, and I agree that it probably makes sense to just be clear that this only works for server-group type things.

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
maggieneterval: 👍 The only reason I thought of it was I remembered that the `/clusters` endpoint, in addition to the strange behavior of returning services, also returns network policies.

PR #4760, opened 2020-07-28
chore(aws): Adding iamRole and imageId to userData request for launch templates

jeyrschabu:
- added imageId and role


jeyrschabu:
Self merging this small change. Would like to promote the fix.

PR #4761, opened 2020-07-28
refactor(kubernetes): Clean up data provider classes

ezimanyi:
* refactor(kubernetes): Improve typing in view classes 

  In a number of places we store data in these classes as the interface type rather than as the concrete type, and then need to cast to the Kubernetes-specific implementation. Instead, let's keep everything concrete in the Kubernetes classes to avoid this casting.

  In some cases, using the interface type can be useful (ie, if we don't want certain implementation-specific functions to be visible). But in this case the primary use for these data classes is serializing them and sending them to another microservice (and eventually to deck). Regardless of the type we are storing it in, we are still fully serializing the instance fields, so the effective contract is that callers expect to find Kubernetes-specific fields here. (That's the reason the tests were casting to the specific type, so that they could verify things in the serialized results.)

* refactor(kubernetes): Inline some functions in data provider 

  As a first pass to simplifying the Kubernetes data providers, inline a few functions. While this makes the calling function longer, it will be simpler because we won't be passing as many custom map objects across function boundaries. As the later commits will show, we'll also better be able to simplify things when the logic is all in one function.

* refactor(kubernetes): Inline a variable in cluster provider 

  A small commit, inlining a variable that is only used in one place and can be just created in place rather than stored as a local variable.

  Also removed the needless null filtering in the return from the function; the only things in this set are the results of calling new KubernetesV2Cluster(...) which can never be null.

* refactor(kubernetes): Replace mutating set with collector 

  Small change to update the return of this function to use a collector instead of mutating a set.

* refactor(kubernetes): Clean up null safety in KubernetesCacheUtils 

  There are a lot of unnecessary null checks in KubernetesCacheUtils and in code that calls those functions, because pretty much every function assumes a Collection<CacheData> might contain null elements.

  At first, I had this commit filtering the results from all calls to the cache (on the assumption that the cache might return a collection containing null elements if you requested an id that didn't exist). But I checked all of the current implementations of Cache, and all of them ensure filter nulls from the result.
  (In the SQL case we even know this statically as it returns a MutableCollection<CacheData> rather than a MutableCollection<CacheData?>.

  To avoid the unnecessary duplicate work, just remove filtering of the Collection<CacheData> that we get back from the cache. Then as long we we don't ourselves create a Collection<CacheData> with null elements, we can be sure that all such collections don't have nulls.

  This greatly simplifies the code, and also likely has performance benefits as we no longer need to iterate over every result from calling the cache, just to create a new collection.

  Finally, this commit does not fully update all downstream logic to remove null checks.  We can update those places as we change them.

* refactor(kubernetes): Improve CacheData classes 

  Make these classes shallowly immutable, and null-safe simplifying their use in callers.

  Also change some fields to a generic Collection in cases where we don't care what particular Collection we have, which will make future refactoring easier.

* refactor(kubernetes): Replace full cache data with keys instead 

  KubernetesV2ServerGroupCacheData requires a full Collection<CacheData> of the load balancer data, but only ever uses the associated keys. Instead, just require the actual keys, which means that callers can avoid doing extra lookups to the cache, simplifying the logic and improving performance.

  Note that this commit does not actually change the calling logic to stop fetching the no-longer-needed CacheData objects, it just pushes the CacheData::getId to the caller. That simplification will come later.

  In the case of serverGroupManagerKeys, we're currently requiring a parsed key. Encapsulate this logic better (and match what we now do for load balancers) by accepting the key as a string and parsing the keys as needed.

* refactor(kubernetes): Make model classes and fields final 

* refactor(kubernetes): A few simplifications in cluster provider 

  Simplify the grouper to just use a grouping collector, and rename some functions to more clearly indicate their effects.

* fix(kubernetes): Fix load balancer lookup 

  We were passing the name where the namespace should have been.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ServerGroupManager.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ServerGroupManager.java
@@ -33,7 +33,7 @@ public interface ServerGroupManager {
 
   Moniker getMoniker();
 
-  Set<ServerGroupSummary> getServerGroups();
+  Set<? extends ServerGroupSummary> getServerGroups();
 
   String getRegion();
 
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2ServerGroup.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2ServerGroup.java
@@ -144,24 +144,8 @@ public class KubernetesV2ServerGroup extends ManifestBasedModel implements Serve
   }
 
   public static KubernetesV2ServerGroup fromCacheData(KubernetesV2ServerGroupCacheData cacheData) {
-    CacheData cd = cacheData.getServerGroupData();
-    List<CacheData> instanceData = cacheData.getInstanceData();
-    List<Keys.InfrastructureCacheKey> serverGroupManagerKeys =
-        cacheData.getServerGroupManagerKeys();
-    if (cd == null) {
-      return null;
-    }
-
-    if (instanceData == null) {
-      instanceData = new ArrayList<>();
-    }
-
-    if (serverGroupManagerKeys == null) {
-      serverGroupManagerKeys = new ArrayList<>();
-    }
-
     List<ServerGroupManagerSummary> serverGroupManagers =
-        serverGroupManagerKeys.stream()
+        cacheData.getServerGroupManagerKeys().stream()
             .map(
                 k ->
                     ServerGroupManagerSummary.builder()
@@ -171,15 +155,16 @@ public class KubernetesV2ServerGroup extends ManifestBasedModel implements Serve
                         .build())
             .collect(Collectors.toList());
 
-    KubernetesManifest manifest = KubernetesCacheDataConverter.getManifest(cd);
+    KubernetesManifest manifest =
+        KubernetesCacheDataConverter.getManifest(cacheData.getServerGroupData());
 
     if (manifest == null) {
-      log.warn("Cache data {} inserted without a manifest", cd.getId());
+      log.warn("Cache data {} inserted without a manifest", cacheData.getServerGroupData().getId());
       return null;
     }
 
     List<KubernetesV2Instance> instances =
-        instanceData.stream()
+        cacheData.getInstanceData().stream()
             .map(KubernetesV2Instance::fromCacheData)
             .filter(Objects::nonNull)
             .collect(Collectors.toList());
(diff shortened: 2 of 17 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ServerGroupManager.java
```diff
@@ -33,7 +33,7 @@
 
   Moniker getMoniker();
 
-  Set<ServerGroupSummary> getServerGroups();
+  Set<? extends ServerGroupSummary> getServerGroups();
```
maggieneterval, line 36: Don't worry, Deck won't trust that it's a Kubernetes SGM and [re-cast](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/kubernetes/src/serverGroupManager/details/details.controller.ts#L163) it itself anyway 😂

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2ServerGroup.java
```diff
@@ -171,15 +155,16 @@ public static KubernetesV2ServerGroup fromCacheData(KubernetesV2ServerGroupCache
                         .build())
             .collect(Collectors.toList());
 
-    KubernetesManifest manifest = KubernetesCacheDataConverter.getManifest(cd);
+    KubernetesManifest manifest =
+        KubernetesCacheDataConverter.getManifest(cacheData.getServerGroupData());
 
     if (manifest == null) {
-      log.warn("Cache data {} inserted without a manifest", cd.getId());
+      log.warn("Cache data {} inserted without a manifest", cacheData.getServerGroupData().getId());
```
maggieneterval, line 168: Wow, can't wait until my logs are full of this warning 😉

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2ServerGroup.java
```diff
@@ -231,6 +232,13 @@ public LoadBalancerServerGroup toLoadBalancerServerGroup() {
         .build();
   }
 
+  @Deprecated
+  @Null
+  @Override
+  public ImageSummary getImageSummary() {
```
maggieneterval, line 238: Why do we need this?

ezimanyi: The interface [requires this](https://github.com/spinnaker/clouddriver/blob/924d53a28ebbbb3391ed1fc93ca44a897ec4d092/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/ServerGroup.java#L170) even though it's deprecated in favor of the similarly-named `getImagesSummary()`.

Prior to this change we had a never-initialized field `imageSummary` whose auto-generated getter provided this.  But as we were never setting the field, this seemed clearer than changing the field to `private final ImageSummary imageSummary = null` (which would have been equivalent).

PR #4763, opened 2020-07-30
fix(titus): only get the task we need

marchello2000:
When we disable an ASG in titus we need to get all tasks associated with the job.
However, it appears that we always retried ALL tasks from ALL jobs. Which is neither great for Titus nor Spinnaker since we a) run out of memory and also timeout like crazy trying to get the world

Pretty sure the issue is this erroneous conditional which checks if Titus supports query by job ids


marchello2000:
I have a bunch more changes (namely removing this flag), but I will merge as-is and make another, follow-up with a separate PR

PR #4764, opened 2020-07-30
refactor(kubernetes): Clean up KubernetesCacheUtils class

ezimanyi:
I broke this into tiny commits more than usual because I was starting to confuse myself; the commits are all very small and focused.

* refactor(kubernetes): Extract getHandler to a function 

  This is a minor refactor to make the upcoming changes cleaner; just pull out a function.

* refactor(kubernetes): Fix type safety and null safety 

  resourceModelFromCacheData is very unsafe; it does an unchecked cast and can also return null. The goal of that function is to allow custom overrides for how to convert cache data into a ServerGroup or a ServerGroupManager.  (We have not implemented the ability to customize other types.)

  The problem is that we are checking if the handler is of the generic ModelHandler instead of the specific type we are interested in.  We already have ServerGroupHandler and ServerGroupManagerHandler to handle these two specific cases. Split the generic resourceModelFromCacheData into to specific ones to handle these two cases.  By checking if the handler implements the specific case, we can avoid unchecked casts.

  Finally, rather than return null if the handler doesn't specify an override, just use the default implementation. The goal here is only to allow people to override this if they want; if they don't override, we can just use the default.

* refactor(kubernetes): Clean up CustomKubernetesHandlerFactory 

  This class implemented the (raw) ModelHandler, but only ever implemented the default implementations for the kind in question. As we now fall back to the default implementations in cases where there is no override, we don't need to explicitly override them here.

* refactor(kubernetes): Remove ModelHandler class 

  Now that we're using the more specific handler classes to detect custom implementations of fromCacheData, we don't need the ModelHandler interface at all anymore; remove it.

* refactor(kubernetes): Add relationshipTypes helper function 

  Minor refactor to abstract away translating a Spinnaker kind into its relationship types.

* refactor(kubernetes): Add relationshipKeys function 

  Minor refactor to pull common logic looking up relationships on a CacheData object into a function.

* refactor(kubernetes): Rename to function 

  Rename aggregateRelationshipsBySpinnakerKind as getRelationshipKeys, because it does the same thing as the relationshipKeys function we just created, but for a SpinnakerKind. Also collect to an ImmutableList and update the return type.

  Similarly remove the OfSpinnakerKind from getRelationships, becase the function signature implies what the input is.

* refactor(kubernetes): Update return type of mapByRelationship 

  It will now return a Map<String, Collection<CacheData>> instead of a Map<String, List<CacheData>> because it's not actually important to callers that the values are List<?>. This will make the next commit possible.

* refactor(kubernetes): mapByRelationship returns a MultiMap 

  This is a much better data structure for mapping keys to multiple values. In particular, it has the property that .get(key) returns an empty list rather than null if there are no mappings, so we can avoid a lot of defaulting.

* refactor(kubernetes): Inline a variable 

  Inline the keys variable in loadRelationshipsFromCache; also collect the relationship keys to a Set rather than to a List.

  This cuts down the number of times we look up load balancers from 4 to 2; we'll still need to get that down to 1 but this is progress.

* refactor(kubernetes): Rename getTransitiveRelationships 

  Rename this function to something shorter and clearer; also update it to only accept a single CacheData as all callers are wrapping a single CacheData in a collection when calling it.

* refactor(kubernetes): Require callers of to create the filter 

  Simple refactor to push requiring the cache filter to be created to callers. (Actually caller as this is only used in one place.)

* refactor(kubernetes): Require callers of getRelationships to pass data 

  getRelationships is inherently ineffecient, as it encourages callers to require a lookup of data they already have.  In 2 of the 3 callers we are either looking up the same data in a loop, or already have the data we're looking up.

  Require callers to first look up the cache data, then pass it in to get its relationships.

* refactor(kubernetes): Rename loadRelationshipsFromCache 

  Use getRelationships to be consistent with how we're naming other functions in this class.

  Also update the single-data-consuming version to call the list version with a singleton.

* refactor(kubernetes): Remove loadRelationshipsFromCache 

  This is identical to getRelationships.

  Also the whole class is non-null now, so move the annotation to the class level.

* refactor(kubernetes): Cleanup of KubernetesCacheUtils 

  Make the class package-private as it is only used in the package. Also add Javadoc to some methods.

* refactor(kubernetes): Replace kindMap lookups with getRelationships 

  We already have a function in cacheUtils that looks up all relationships for a SpinnakerKind; replace cases where callers map the SpinnakerKind to KubernetesKinds with just calling this function.


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesCacheUtils.java
```diff
@@ -153,18 +153,24 @@ public KubernetesCacheUtils(
     return result;
   }
 
-  @SuppressWarnings("unchecked")
```
maggieneterval, line 156: The Tempest

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesCacheUtils.java
```diff
@@ -166,15 +164,4 @@ KubernetesHandler getHandler(@Nonnull KubernetesV2CacheData cacheData) {
         .get(kind)
         .getHandler();
   }
-
-  @SuppressWarnings("unchecked")
```
maggieneterval, line 156: The Comedy of Errors

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/CustomKubernetesHandlerFactory.java
```diff
@@ -113,21 +107,5 @@ private KubernetesV2CachingAgent buildCustomCachingAgent(
...
-              (KubernetesV2ServerGroupCacheData) cacheData);
-        case SERVER_GROUP_MANAGERS:
-          return KubernetesV2ServerGroupManager.fromCacheData(
-              (KubernetesV2ServerGroupManagerCacheData) cacheData);
-        default:
-          // TODO(dpeach): finish implementing for other SpinnakerKinds.
-          log.warn("No default cache data model mapping for Spinnaker kind " + spinnakerKind());
-          return null;
-      }
-    }
```
maggieneterval, line 131: Love's Labour's Lost

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/ModelHandler.java
```diff
@@ -1,25 +0,0 @@
...
- */
-
-package com.netflix.spinnaker.clouddriver.kubernetes.op.handler;
-
-import com.netflix.spinnaker.clouddriver.kubernetes.caching.view.model.ManifestBasedModel;
-import com.netflix.spinnaker.clouddriver.kubernetes.caching.view.provider.data.KubernetesV2CacheData;
-
-public interface ModelHandler<T extends KubernetesV2CacheData> {
-  ManifestBasedModel fromCacheData(T cacheData);
-}
```
maggieneterval, line 25: Pericles, Prince of Tyre

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesCacheUtils.java
```diff
@@ -151,6 +153,12 @@ public KubernetesCacheUtils(
     return result;
   }
 
+  /** Given a spinnaker kind, returns a stream of the relationship types representing that kind. */
```
maggieneterval, line 143: The Merchant of Venice

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesCacheUtils.java
```diff
@@ -153,6 +139,17 @@ public KubernetesCacheUtils(
     return result;
   }
 
+  /** Returns a stream of all relationships of a given type for a given CacheData. */
+  @NonnullByDefault
+  private Stream<String> relationshipKeys(CacheData cacheData, String type) {
```
maggieneterval, line 134: The Merry Wives of Windsor

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesCacheUtils.java
```diff
@@ -80,11 +83,12 @@ public KubernetesCacheUtils(
     return Optional.ofNullable(cache.get(type, key, RelationshipCacheFilter.include(to)));
   }
 
-  private Collection<String> aggregateRelationshipsBySpinnakerKind(
-      CacheData source, SpinnakerKind kind) {
-    return relationshipTypes(kind)
-        .flatMap(t -> relationshipKeys(source, t))
-        .collect(Collectors.toList());
+  /** Returns a collection of all relationships of a given SpinnakerKind for a CacheData. */
+  ImmutableCollection<String> getRelationshipKeys(
```
maggieneterval, line 86: The Taming of the Shrew

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2Cluster.java
```diff
@@ -43,8 +43,8 @@ public KubernetesV2Cluster(String rawKey) {
 
   public KubernetesV2Cluster(
       String rawKey,
-      List<KubernetesV2ServerGroup> serverGroups,
-      List<KubernetesV2LoadBalancer> loadBalancers) {
+      Collection<KubernetesV2ServerGroup> serverGroups,
```
maggieneterval, line 46: As You Like It

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2LoadBalancer.java
```diff
@@ -54,7 +54,7 @@ private KubernetesV2LoadBalancer(
   public static KubernetesV2LoadBalancer fromCacheData(
       CacheData cd,
       Collection<CacheData> serverGroupData,
-      Map<String, ? extends Collection<CacheData>> serverGroupToInstanceData) {
+      Multimap<String, CacheData> serverGroupToInstanceData) {
```
maggieneterval, line 57: A Midsummer Night's Dream

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
```diff
@@ -544,7 +544,7 @@ private void assertFrontendCluster(
       assertFrontendServerGroups(softly, cluster.getServerGroups());
       // TODO(ezimanyi): The same load balancer is being returned multiple times (likely due to
       // finding all relationships to it). This should be fixed to return it only once.
-      softly.assertThat(cluster.getLoadBalancers()).hasSize(4);
+      softly.assertThat(cluster.getLoadBalancers()).hasSize(2);
```
maggieneterval, line 547: The Two Gentlemen of Verona

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesV2ManifestProvider.java
```diff
@@ -129,13 +129,10 @@ private KubernetesV2Manifest fromCacheData(
...
     String namespace = manifest.getNamespace();
     KubernetesKind kind = manifest.getKind();
-    String key = data.getId();
 
     List<KubernetesManifest> events =
         includeEvents
-            ? cacheUtils
-                .getRelationships(kind.toString(), key, KubernetesKind.EVENT.toString())
-                .stream()
+            ? cacheUtils.getRelationships(data, KubernetesKind.EVENT.toString()).stream()
```
maggieneterval, line 135: Measure for Measure

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesV2LoadBalancerProvider.java
```diff
@@ -118,17 +118,9 @@ public String getCloudProvider() {
...
-            .collect(Collectors.toList());
-
-    List<CacheData> instanceData =
-        kindMap.translateSpinnakerKind(INSTANCES).stream()
-            .map(kind -> cacheUtils.getRelationships(serverGroupData, kind.toString()))
-            .flatMap(Collection::stream)
-            .collect(Collectors.toList());
+    Collection<CacheData> serverGroupData =
+        cacheUtils.getRelationships(loadBalancerData, SERVER_GROUPS);
+    Collection<CacheData> instanceData = cacheUtils.getRelationships(serverGroupData, INSTANCES);
```
maggieneterval, line 123: All's Well That Ends Well

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesCacheUtils.java
```diff
@@ -39,12 +39,12 @@
 import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import javax.annotation.Nonnull;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 @Component
+@NonnullByDefault
```
maggieneterval, line 47: Much Ado About Nothing

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesCacheUtils.java
```diff
@@ -79,8 +79,8 @@ public KubernetesCacheUtils(
   }
 
   public Optional<CacheData> getSingleEntryWithRelationships(
-      String type, String key, String... to) {
-    return Optional.ofNullable(cache.get(type, key, RelationshipCacheFilter.include(to)));
+      String type, String key, RelationshipCacheFilter cacheFilter) {
```
maggieneterval, line 81: Twelfth Night

(this is the 12th commit)

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesCacheUtils.java
```diff
@@ -114,7 +119,7 @@ public KubernetesCacheUtils(
   /*
    * Builds a map of all keys belonging to `sourceKind` that are related to any entries in `targetData`
    */
-  public ImmutableMultimap<String, CacheData> mapByRelationship(
+  ImmutableMultimap<String, CacheData> mapByRelationship(
```
maggieneterval, line 123: The Winter's Tale

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesCacheUtils.java
```diff
@@ -60,115 +61,100 @@ public KubernetesCacheUtils(
...
 
-  private Collection<String> aggregateRelationshipsBySpinnakerKind(
-      CacheData source, SpinnakerKind kind) {
-    return kindMap.translateSpinnakerKind(kind).stream()
-        .map(g -> source.getRelationships().get(g.toString()))
-        .filter(Objects::nonNull)
-        .flatMap(Collection::stream)
-        .filter(Objects::nonNull)
-        .collect(Collectors.toList());
+  /** Gets the data for all relationships of a given Spinnaker kind for a CacheData items. */
```
maggieneterval, line 85:
```suggestion
  /** Gets the data for all relationships of a given Spinnaker kind for a CacheData item. */
```

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesCacheUtils.java
```diff
@@ -60,115 +61,100 @@ public KubernetesCacheUtils(
...
-    return cache.getAll(
-        to,
-        sourceData.stream()
-            .map(CacheData::getRelationships)
-            .filter(Objects::nonNull)
-            .map(r -> r.get(to))
-            .filter(Objects::nonNull)
-            .flatMap(Collection::stream)
-            .collect(Collectors.toList()));
+  /** Gets the data for all relationships of a given type for a CacheData items. */
```
maggieneterval, line 93:
```suggestion
  /** Gets the data for all relationships of a given type for a CacheData item. */
```

PR #4767, opened 2020-07-31
refactor(kubernetes): Further improvements to Kubernetes data providers

ezimanyi:
* refactor(kubernetes): Add Multimap-returning getRelationships 

  This commit only affects CacheData, in preparation for some changes visible outside the class. Most notably, we'll update getRelationships to return a MultiMap when we pass in a Collection of CacheData items. This allows the caller to see the association between the input data and the output relationships, rather than flattening it to a collection where it's not possible to determine which of the output CacheData items is related to each of the input ones.

  In this commit, we're not taking advantage of this; in fact the two callers have been updated to:
  * Just call .values() to flatten the result
  * Query the cache directly instead of using getRelationships

  But in the next commit we'll push this information outward one more level of calls, which will the allow consumers to use the information to avoid duplicate work.

  Also update getRelationshipKeys to collect to a Set. The general idea is that we'll collect keys (String) to a Set to dedupliate but data (CacheData) to a List as deduplication by reference is not valuable.  (I left one TODO in a case where we are deduplicating by reference but will fix that.)

* refactor(kubernetes): Keep track of relationships when fetching 

  In a number of places, we have a Collection of CacheData items, and we want to fetch all relationships of a particular kind. For example, we might have a List of server groups and want to load all instances.

  Currently we returned a flattened collection of all the related types, forcing the caller to then iterate through it to recompute which instance is associated with which server group.

  Greatly simplify this (and improve performance) by returning a MultiMap from getRelationships. This means, for example, that the caller can then look up the instances related to a given server group id, without needing to build another map.

* refactor(kubernetes): Look up related load balancer keys directly 

  When constructing a server group, we look up the full data for related load balancers, and then only keep the key. Instead let's look up only the keys. As this isn't making any cache calls, we don't need to keep the full map outside the loop, and can just look up the needed data inside the loop.

* refactor(kubernetes): Extract loadServerGroups to a function 

  It returns a Map so we can later get the constructed server group by id, though we are not yet using that feature.

* refactor(kubernetes): Load balancer constructor accepts server groups 

  The fromCacheData method to create a LoadBalancer requires all the cache data to create server groups as well; but in some cases we will already have constructed server groups. It will be simpler and more performant to just pass these existing server groups to fromCacheData.

  Add a (as yet unused) fromCacheData that takes a set of LoadBalancerServerGroups and uses those. Refactor the existing fromCacheData to pass the LoadBalancerServerGroups it constructs to this new constructor.

* refactor(kubernetes): Use new fromCacheData in cluster provider 

  Let's use the new fromCacheData we just created and pass in the server groups that we have already built.

  This also removes the last use of mapByRelationship, so we can remove the function.

* refactor(kubernetes): Minor cleanup in loadClusters 

  Now that only serverGroupToInstances is only used in loadClusters, look up that data in the function rather than require the caller to look it up.

  Similarly, the construction of the stream over load balancers is just a really complex way of getting the values of the MultiMap; use that instead.

* refactor(kubernetes): Remove the old version of fromCacheData 

  It was only used in one place; let's move the logic there.

* refactor(kubernetes): Update getServerGroup to call existing function 

  getServerGroup has its own logic for constructing server groups; instead call the loadServerGroups function we just wrote.

  This commit also fixes a bug; we were only loading load balancer and instance relationships with the cache data, so we never returned any server group managers.

  The integration tests didn't catch this because they were exercising the function only for a non-deployment-managed replica set. Add a test to catch this case.

  It's interesting that we only do this optimization of filtering relationships when loading a single server group, whereas it probably would be much more beneficial in the endpoints that return many. I'll leave this for now, but may revisit.

* refactor(kubernetes): Extract relationshipFilter logic 

  I'm honestly not sure that the benefit of creating this relationship filter for a single server group is worth it, but we'll keep the logic for now.  Let's at least store the cache filter as an instance field so that we can re-use it, and abstract away the logic for creating it.

  I'm not going to include filter logic in other places as I think the correct fix for improving performance here is just to cut down the relationships we store in the cache to those that are actually needed, rather than add logic for ignoring most of them.

* refactor(kubernetes): Fix one more case of manually using kindMap 

  There is now a function for this.

* refactor(kubernetes): Add helper function for looking up a item 

  We made callers accept a MultiMap then look up the one id of interest; let's do that instead for callers.

* fix(kubernetes): Stop returning multiple copies of load balancer 

  The issue was that we were building the load balancer once per attached server group.

  All of the refactors up to now have not fixed it (though we did reduce the number of times it is returned from 4 to 2).

  The reason at this point is because .values() on a multimap returns items multiple times if they map to multiple keys. Let's apply a distinct on the stream that only takes the first for each id.

  We could have just done a regular distinct() because the result of getRelationships always maps a key to the same reference (so we don't have multiple instances of each load balancer, just multiple pointers to the same one).  But I did not want to have to debug this issue if that ever changes, so let's be explicit and keep the first one we encounter with each id.

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2LoadBalancer.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2LoadBalancer.java
@@ -17,6 +17,8 @@
 
 package com.netflix.spinnaker.clouddriver.kubernetes.caching.view.model;
 
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Multimap;
 import com.netflix.spinnaker.cats.cache.CacheData;
@@ -30,7 +32,8 @@ import com.netflix.spinnaker.clouddriver.model.LoadBalancerServerGroup;
 import java.util.Collection;
 import java.util.Objects;
 import java.util.Set;
-import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+import javax.annotation.ParametersAreNonnullByDefault;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
 import lombok.extern.slf4j.Slf4j;
@@ -51,22 +54,14 @@ public final class KubernetesV2LoadBalancer extends ManifestBasedModel
     this.serverGroups = serverGroups;
   }
 
+  @Nullable
+  @ParametersAreNonnullByDefault
   public static KubernetesV2LoadBalancer fromCacheData(
       CacheData cd,
       Collection<CacheData> serverGroupData,
       Multimap<String, CacheData> serverGroupToInstanceData) {
-    if (cd == null) {
-      return null;
-    }
-
-    KubernetesManifest manifest = KubernetesCacheDataConverter.getManifest(cd);
-
-    if (manifest == null) {
-      log.warn("Cache data {} inserted without a manifest", cd.getId());
-      return null;
-    }
-
-    Set<LoadBalancerServerGroup> serverGroups =
+    return fromCacheData(
+        cd,
         serverGroupData.stream()
             .map(
                 d ->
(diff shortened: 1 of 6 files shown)


ezimanyi:
This is, alas, the last batch of commits from my branch full of changes and I can now move on to other things.

PR #4769, opened 2020-08-03
refactor(kubernetes): Remove manifest from view objects

ezimanyi:
* refactor(kubernetes): Push getAccount to implementations 

  First, getAccountName is only needed for security groups; push that down only to the security group provider.

  Second, getAccount is defined in the base class, but depends on the abstract method getKey that is defined by implementations. As this is the only reason we need to have the getKey method, let's just push getAccount itself down to the implementations so that we can just access the key directly. In fact, instead of storing it at all, let's just use the key in the constructor and discard it as we don't need it for anything else.

  Eventually the plan will be to not read this off the key at all in the constructor as we should already be storing the account as part of the cache data we're using to construct this.

* refactor(kubernetes): Push fields down from ManifestBasedModel 

  Most of ManifestBasedModel is lookign up fields on a Manifest; we want to stop using the manifest for all of these lookups because in most cases we already store the parsed data as part of the CacheData item (and in the cases we don't, we should start doing that).

  In order to simplify moving to that model, push down all the manifest logic to the implementing classes. It's a bit more code now, but should make it much easier to understand. We also remove some methods that aren't necessary because they were added to the superclass even though they are only needed by one or two implementations.

  Add a KubernetesResource interface that keeps track of fields we're expecting to find on all kubernetes resources (mapping to the correpsonding interface in deck). This is just helpful to make sure that we get a compile-time error if we accidentally remove one of these fields.

* refactor(kubernetes): Pull ImageSummary to a class 

  This is a fairly large class to be an anonymous class; pull it out to a real class.

* refactor(kubernetes): Remove manifest from view objects 

  This commit removes the manifest from the view object models, instead reading the required fields from the manifest in the constructor, then discarding the manifest.

  This is an intermediate step towards not requiring the manifest in the constructor at all, and instead using the metadata we store on the cache object.


ezimanyi:
@mattsanta : I can't assign you as a reviewer, but this might be a good PR to look at to get to know the Kubernetes code.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2SecurityGroup.java
```diff
@@ -61,12 +61,16 @@
...
 
   private final KubernetesManifest manifest;
-  private final Keys.InfrastructureCacheKey key;
+  private final String account;
   private final String id;
 
   private final Set<Rule> inboundRules;
   private final Set<Rule> outboundRules;
 
+  public String getAccountName() {
```
maggieneterval, line 80: Any reason only security groups need this getter?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2LoadBalancer.java
```diff
@@ -58,4 +62,57 @@ public static KubernetesV2LoadBalancer fromCacheData(
...
+  public KubernetesApiVersion getApiVersion() {
+    return getManifest().getApiVersion();
+  }
+
+  @Override
+  public String getNamespace() {
+    return getManifest().getNamespace();
+  }
+
+  public String getRegion() {
```
mattsanta, line 89: Why is namespace returned for both getNamespace and getRegion? Is the concept of region used by other providers but for Kubernetes it's equivalent to namespace?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2SecurityGroup.java
ezimanyi: All the others look things up by using the auto-generated `getAccount` accessor (and the deck interfaces only required `account` not `accountName`).  For some reason for security groups we've historically  [looked things up](https://github.com/spinnaker/clouddriver/blob/658bac1b8b21121cdb1ac6fbfcb76cec16d5a410/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/SecurityGroupController.groovy#L54) by `accountName`, so it's become [required](https://github.com/spinnaker/clouddriver/blob/658bac1b8b21121cdb1ac6fbfcb76cec16d5a410/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/SecurityGroup.java#L80) on the interface.

That being said, your question made me find a couple of places where `deck` reads `accountName` from a server group/server group manager even though it's not on the interface so I probably need to fix those before merging this ([here](https://github.com/spinnaker/deck/blob/cb71e811c49dc1e92f96045c436b4ce2bb1e167d/app/scripts/modules/kubernetes/src/serverGroup/details/details.html#L96) and [here](https://github.com/spinnaker/deck/blob/cb71e811c49dc1e92f96045c436b4ce2bb1e167d/app/scripts/modules/kubernetes/src/serverGroupManager/details/details.html#L102)).

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2LoadBalancer.java
ezimanyi: Yes, exactly, other providers have a concept of region and zone (where a server group would be in a region, and its instances in particular zones).  For the Kubernetes provider, we basically just return the namespace for both of those.

It's not really a perfect mapping, Kubernetes clusters do also have a concept of region/zone...as the instances in a cluster are a in a particular region/zone, and if you want high-availability you probably run your workloads in >1 cluster in >1 region/zone.  But the underlying infrastructure of a cluster is something that the Spinnaker Kubernetes provider has no knowledge about, so in the absence of that we just use namespace as region or zone.  In a lot of places where we're writing Kubernetes code it's clearer to just call it a namespace, so we return the namespace as `getNamespace` as well to avoid needing callers who know it's a Kubernetes object to need to ask for the region/zone.

ezimanyi:
Fixed a couple of things in deck that need to be merged before this in spinnaker/deck#8464.

PR #4771, opened 2020-08-04
refactor(kubernetes): Improve null safety and remove some generics

ezimanyi:
Just a few things I cleaned up while working on the caching changes.  This starts to remove some redundant generic interfaces that are not necessary now that there is no V1 provider.  It definitely does not come close to removing all of these, just a few that I could do quickly, we can continue to clean these up as we come across them.

* refactor(kubernetes): Move account validation to an instance method 

  The function to validate a managed account is as a default method of an interface; it would be clearer if we just made it an instance method of the account we're trying to validate.

* refactor(kubernetes): Validate account earlier and add annotations 

  Currently we validate a managedAccount when creating the KubernetesV2Credentials, but by that time we've already read a number of fields from it.  Move this validation earlier, to the beginning of the KubernetesNamedAccountCredentials constructor so that we can throw any errors earlier.

  This will not create new exceptions for any workflow, as we were always building the KubernetesV2Credentials at the end of this function (and calling validate) anyway, it just moves the exception to happen earlier.

  Let's also now annotate the two places we read the accountName with NonNull. We'll be extra careful and use Objects.requireNonNull even though we know we've already validated with a better message earlier.
  (This is so that if that validation ever changes/breaks at least we'll still be guaranteed a non-null value as the annotation states, albeit with a less helpful runtime exception.)

* refactor(kubernetes): Remove StubKubernetesCredentials 

  I wrote these to make testing easier, but these implement the generic KubernetesCredentials interface. Now that there's no V1 provider anymore, we have only one implementation of the interface and we should eliminate it for simplicity.

  Given that the interface doesn't really represent a useful level of abstraction, using it in tests is really not that much better than mocking, so use a mock instead.

  Ideally we'll break up KubernetesV2Credentials to make testing easier but for now let's remove the level of abstraction that no longer adds value.

* refactor(kubernetes): Un-generify KubernetesCachingAgentDispatcher 

  Now that there is no Kubernetes V1 provider, there is only one implementation.

* refactor(kubernetes): Remove KubernetesCachingAgent interface 

  Push everything down to KubernetesV2CachingAgent, which is the only implmentation now that the V1 provider is deleted.

* refactor(kubernetes): Un-generify KubernetesAtomicOperationDescription 

  Now that the V1 provider is gone, the only implementations are for the V2 provider, so we can remove the redundant type parameter.


ezimanyi:
@mattsanta: Another one to take a look at if you have a minute!

PR #4772, opened 2020-08-05
fix(saga): Retry INVALID_ARGUMENT exception when message is Titus assume IAM role error

jonsie:

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusExceptionHandler.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusExceptionHandler.java
@@ -46,10 +46,24 @@ public class TitusExceptionHandler implements SagaExceptionHandler {
       case UNKNOWN:
         return true;
       case INVALID_ARGUMENT:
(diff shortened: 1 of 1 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/TitusExceptionHandler.java
```diff
@@ -46,10 +46,23 @@ private boolean isRetryable(StatusRuntimeException statusRuntimeException) {
...
+  private boolean invalidArgumentConditional(String statusRuntimeExceptionMessage) {
+    if (statusRuntimeExceptionMessage == null) {
+      return false;
+    }
+
+    boolean rateExceeded = statusRuntimeExceptionMessage.toLowerCase().contains("rate exceeded");
+    boolean assumeRoleError =
+        statusRuntimeExceptionMessage
+            .toLowerCase()
+            .contains("field: 'jobiamvalidator', description: 'titus cannot assume into role");
```
ajordens, line 64: Might be a little specific and subject to some structural change/ordering of attributes ... ?

```
contains("jobiamvalidator") && contains("titus cannot assume into role")
```

ajordens:
LGTM

PR #4776, opened 2020-08-06
refactor(kubernetes): Further simplifications of cache data logic

ezimanyi:
* refactor(kubernetes): Remove needless ObjectMapper 

  The object is already of the correct type. Also inline declaration/ assignment above so it's easier to see this.

* refactor(kubernetes): Clean up convertAsArtifact 

  convertAsArtifact is doing a lot of null checks on things that can't be null.  If the artifact is present, it's guaranteed to have a non-null type; likewise the getCreationTimestamp never returns null.

  Remove these checks, inline some variables, and move to an
  .ifPresent to greatly simplify the function.

  Finally, we were storing a relationship between the infrastructure item and the artifact; we don't need or use these relationships at all and treat these two areas of the cache separately. It's better for performance on cache writes/reads, as well as simplifies the code, if we remove these relationships. This means we don't need to return the key from convertAsArtifact, or use it to add a relationship in the calling code.

* refactor(kubernetes): Pull convertAsArtifact out of convertAsResource 

  Now that convertAsArtifact and convertAsResource are not dependent on each other (because they don't store reource-artifact relationships) it's clearer if we pull convertAsArtifact up a level rather than have it be inside of convertAsResource.

  Also update catching an Exception to catching a RuntimeException.

* refactor(kubernetes): Clean up null checks in RegistryUtils 

  lookupHandler is really a simple function but is 15 lines long to account for unnecessary null checks. None of the callers ever pass a null kind (as they use KubernetesManifest.getKind which never returns a null). Likewise, then KubernetesResourceProperties in practice can never have a null handler; let's just enforce that at creation and add the annotation.

  Now that we don't need all these null checks, just directly look up the handler where needed instaed of calling lookupHandler.

* refactor(kubernetes): Remove RegistryUtils 

  These two functions are just looking properties up on an object; inline them where they are used so we can delete another static utils class.

* refactor(kubernetes): Move filtering logic to stream 

  Instead of exiting early from convertAsResource, just move the filtering logic to the stream where it's called as it's clearer there.  Also update KubernetesCachingProperties to be immutable and only have the API that we actually use.

  Finally, remove setting the application we got from the KubernetesCachingProperties as an attribute in the cache. We don't use this in any case, and in cases where we do want the applciation, we should use the one on the moniker that we persist in the cache data.

* refactor(kubernetes): Clean up some logging 

  We check a few things on a manifest before caching and log a warning; it's not clear that this is really useful. The check that it's not null is always false; the check for name/namespace I guess could help with debugging, but we're really just checking that Kubernetes is returning a valid manifest, which seems not that useful.

  At points where we actually use these, I guess we could log/fail if they are missing, but I've never found a use for this log message in all my time debugging Kubernetes bugs.

  Similarly, logging every non-Spinnaker-owned manifest does not seem useful; there is a flag to decide if you want to cache these or not. If you have the flag on you want these cached, if not, you don't reach this log statement.

* refactor(kubernetes): Use Moniker from CacheData instead of re-computing 

  As a first step towards using the data we cache instead of re-computing things, let's use the Moniker that we store on the CacheData. We always store this, but are currently throwing it away and re-computing it.

  For now I'm just adding Moniker to the constructor, though my eventual plan (once we build everything from the CacheData) will be to pass in a strongly-typed CacheData object rather than greatly increase the number of constructor parameters.

* refactor(kubernetes): Use a Kubernetes-specific namer registry 

  We currently use a reasonably complex NamerRegistry to handle two things:
  (1) Keeping track of any NamingStrategy components on the classpath and returning them by name when we call namerRegistry.getNamingStrategy.
  (This happens when we construct Kubernetes credentials and have a string representing the naming strategy.)
  (2) Given an account name, asking for the naming strategy for that account.

  In both cases, it's overly complicated that we have general logic for all providers and this logic is not type-safe.  For example getNamingStrategy returns a raw Namer, so we have to hope that is a Namer&lt;KubernetesManifest&gt;, and hope is not a naming strategy.

  Likewise, every time we need to look up the namer for an account, we need to query a global registry across all providers (but first tell it that we want a namer for a Kubernetes account, the name of the account, and that we want to name KubernetesManifest.) While this looks somewhat typesafe, it turns out that the only reason we pass KubernetesManifest.class to these calls is so that withResource can cast the result for us (so we don't need to), so we are only getting the illusion of safety.

  Let's instead:
  (1) Make a KubernetesNamerRegistry that keeps track of any registered NamingStrategy<KubernetesManifest> and return them by name (while guaranteeing typing).
  (2) Once we get the namer, just store it on the credentials object like we do for other things, instead of relying on a global registry.

  I don't like having more and more things on the credentials object, but I still think this is better than global state which is the current state. Also, by having these things all in one place on the object we can start to see patterns and refactor them into better classes.

  We now no longer use the general NamerRegistry at all in the Kubernetes provider; other providers still use it, though they may also want to just implement their own like this.

* refactor(kubernetes): Make KubernetesManifestNamer immutable


ezimanyi:
@mattsanta: Another one to look at if you're interested!

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/names/KubernetesNamerRegistry.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+import java.util.Optional;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * This class handles registering any naming strategies for kubernetes manifests that are on the
+ * classpath, and supports looking these up by name. It is in principle possible for users to add
+ * additional namers in a custom build, but it is not clear how often this is used. The only namer
+ * that exists upstream is {@link KubernetesManifestNamer}.
```
maggieneterval, line 35: This feels like an unnecessary level of abstraction. Do you think we're ever going to register more namers? If we don't have any evidence that this is a useful extension point, feels like we should consider ripping out the registry and just using `KubernetesManifestNamer` as a static utility, similar to `KubernetesManifestAnnotator` or `KubernetesManifestLabeler`.

maggieneterval: Oh, I guess we [document](https://spinnaker.io/guides/developer/extending/crd-extensions/#custom-spinnaker-naming-strategies) this as an extension point...

ezimanyi: Yeah, I completely agree that the benefit of allowing this to be customized is unclear.  In particular, most of what `KubernetesManifestNamer` does is just apply and read annotations, which should work regardless of whether it is naming a built-in object or a CRD.  Maybe someone would want to change the default strategy for naming something that doesn't have the annotations applied (which defaults to frigga now)...but in that case we should probably just allow that to be extended rather than the whole namer.

I did consider just making it a static utility, but at least at this point didn't want to break the extensibility.  Let's discuss at the SIG and if we don't hear of anyone using it I'm on board with making it a static utility, which would greatly simplify things (and make it that much easier to create credentials in a test too).

PR #4780, opened 2020-08-07
chore(build): removed UTF8 "smart quotes"

kevinawoo:
<kbd></kbd>

This was causing some errors with the build


kevinawoo:
here's another Bintry one

<kbd></kbd>

PR #4781, opened 2020-08-08
fix(test): reverts (#4777) Fix intermittent timeout in PooledRequestQueueTest

kevinawoo:
This reverts commit 22feb2a866f96606ebe19be61369bd7a4a155eb6.

I think this might be breaking the build?


ezimanyi:
@kevinawoo @ncknt : What was the symptom of the build failing that led you to believe this was the cause?  I'd like to re-apply this if we've determined it wasn't the cause but am not sure how to validate it.

kevinawoo:
I think it's fine to reapply it since reverting it didn't fix the build.

ezimanyi:
Sounds good, thanks!

PR #4785, opened 2020-08-10
fix(kubernetes): check existing versions in the account we are deploying to

ncknt:
This fixes https://github.com/spinnaker/spinnaker/issues/5964


ncknt:
@ezimanyi could you PTAL? Details are in linked issue 5964 but basically we were computing versions from all artifacts with the same name and location regardless of the account.

PR #4787, opened 2020-08-10
Move providers to api

claymccoy:
This PR moves the providers and model interfaces from cloudriver-core to clouddriver-api.

A few relatively minor changes were needed on some things that were moved.
* Groovy was turned into Java
* All Jackson annotations were removed from the model because that dependency is not allowed in clouddriver-api which is used more like a library.
* The Jackson annotations were added back using mixins in clouddriver-core
* Usages of `NamerRegistry.getDefaultNamer()` was changed to use `new FriggaReflectiveNamer()` directly. Which removed the need to move NamerRegistry and its related classes.
* A usage of org.apache.commons.lang3.ObjectUtils to compare nullable objects was removed and coded directly.
* mergeClusters() was moved out of Application


ezimanyi:
A lot of the models that have been moved have public or package-private fields as well as:
```
@Data
@Builder
@AllArgsConstructor
@NoArgsConstructor
```

This creates an enormous API surface for these classes (hidden from the implementer of the class by Lombok, but fully visible to anyone trying to use it).  How do we expect consumers to construct these?  How do we expect them to access the fields on the objects?

Given that we're going to be moving these to the `clouddriver-api` module and making them public contracts, I think we should decide what the public API should be for these data classes and clean it up before we start having usages that we don't know about (and are prevented from doing this cleanup).

claymccoy:
@ezimanyi
If we insist that it has to be perfect before we move it, then I don't think that it will ever get moved. I'm purposefully making the least invasive changes possible to get these things moved over in this particular PR. The strategy so far has been to get it moved without adding dependencies to clouddriver-api in order to identify the important contracts. Then we can improve them incrementally together. Everything has an Alpha or Beta tag indicating that we expect changes. Alpha has been used for something completely new while Beta is being used on interfaces that have existed for years but may still be reexamined.
It feels like this is just being honest about what is here and making it more convenient to reason about. Moving them to api says these are the current important interfaces that you might implement. It allows your plugin to depend on clouddriver-api only which is a more like a legit library with minimal dependencies. And we have them tagged with Beta to say these could still change. There is nothing preventing us from making changes.
That said, it's just an incremental step in the right direction. I'll be happy to continue to help improve these. Feel free to create issues around specific concerns. Most of the improvement will come from more attempts to use and implement these interfaces in plugins and this makes that easier.
Lastly, I feel like you are exaggerating the problem with Lombok. I think there are two LoadBalancer classes that have these annotations. I'll be happy to remove those classes or try to limit the use of Lombok if that is a concern. But I'd rather do that with a future PR. I feel like this work has to happen to identify what the current API is and start a conversation about how we want to evolve it.

ezimanyi:
I certainly don't expect things to be perfect before making any progress, but do think it's reasonable to be very deliberate about the API that we're going to publicly expose.  While I see that you have now added a commit annotating the classes as `@Beta`, they were not annotated as such when I added my comment above.

I still think that, even for `@Beta` classes, the time to have a discussion about what the API should be is when it is moved to `clouddriver-api`.  If this would involve too many coordinated changes, then at least we can come up with a plan for how to get things in the state that we want.  For example, if we decide that we don't want `@AllArgsConstructor` on things we publicly expose, one solution would be to write the constructor explicitly and mark it `@Deprecated` so we don't break existing uses in clouddriver but API consumers know not to use it as they write plugins.  (This certainly seems less impactful than just changing/deleting it later, even for something in `@Beta`.)

This PR adds 7 uses of `@AllArgsConstructor` to `clouddriver-api`.  Including that annotation on a class makes even adding a new field a breaking change because it changes the public constructor (unless you also explicitly write out the old version of the constructor at the time you add the field, which leads to an even larger API surface).  It seems reasonable to me that we would have a discussion around how we want to handle this as part of the PR moving this to our public API.

I guess, stepping back, what is your expectation for how people should weigh in on what we're adding to our public API?  I don't think replying "If we insist that it has to be perfect before we move it, then I don't think that it will ever get moved" is a helpful or constructive response to feedback.  Is your expectation that PRs moving things to `clouddriver-api` will not generate discussion about what that API should be and will just be approved automatically?  Certainly not every comment needs to be addressed directly in that PR (and maybe some don't need to be addressed at all after discussion), but I would certainly like to see us have these discussions and then decide a path forward (which might include adding an issue to do cleanup later).

claymccoy:
Yes, your comment reminded me about the need for the Beta annotation. To be fair they weren't there when you initially looked just because I forgot. Sorry about that.

claymccoy:
It seems unfair to characterize my point about the perfect getting in the way of the good as unhelpful or unconstructive. It is a question about ordering and priorities. I'm happy to talk about what the bar is to be able to move something to API. So far that bar has been that it must be in Java and the dependencies should be limited. The higher that bar is the more invasive the upfront changes will be. I'd prefer moving stuff with as little change as possible and then incrementing on the API. This is the API whether we want to be transparent about that or not and it was usable in a plugin even from core (just harder to identify).

"Is your expectation that PRs moving things to clouddriver-api will not generate discussion about what that API should be"
No, it should be clear that it is the opposite. There is no reason to do this just to make a plugin. Moving stuff to API like this is only done to make it clear what the real API surface area is and then discuss that. My only disagreement is with the ordering of acting on the feedback and the unclear expectations of an existing class or interface that is moved to -api. It is a bit of a catch 22 to do the work of identifying these important contracts and then having to meet some undetermined bar before the changes can be pushed so that everyone else can easily see the current API.
I'm happy to make more changes and test them before this goes in. I'm also happy to discuss further constraints that must be met before moving to -api.

claymccoy:
@ezimanyi The last commit splits out interfaces and removes all the lombok that was introduced. There is one preexisting use of it that I can get rid of and we may want to remove lombok completely from api, but I'd like to do that in a future PR. This was more change than I wanted in this one.
If you see anything else that concerns you, please leave some actionable feedback and we can discuss if it needs to be done here or in a future PR. Thanks for taking a look at this. I think that this will get easier as we all begin to agree on a standard for what is required to promote something to api.

robzienert:
Ah, one last thing:

I'm of split mind on the idea of making data / value classes interfaces. I don't think forcing plugin developers to write implementations of POJOs is very productive or ergonomic, nor does it make a lot of sense to me what the gain is. If we're worried about Lombok exposing too much, it has annotations to help control that, but otherwise it's pretty trivial to auto-generate constructors, getter/setters, and so-on via IntelliJ.

The part where I'm split, however, is when these models would be extended by cloud providers. In this case, maybe it would be better for them to be interfaces, so that a cloud provider can make the choice of whether they need to `abstract class` logic across models and so-on. Even in this case, I'd still want a default implementation provided by `-api` that I can delegate to (or use directly in cases where I don't need cloud provider-specific data decorated).

claymccoy:
@ezimanyi @robzienert
I'm closing this PR. Though I would have loved to see this change, I don't feel comfortable making this level of modifications to existing code in one PR. I don't really need it, I can use what I want from core as is. This just seemed like a good improvement for the community, but the cost is prohibitive for several reasons.

* I've got a few things that I plan to add to the java docs where I actually implemented the class or where the contracts are bizarre. I am not in a position to document all the interfaces here that the original authors neglected to.
* I tried to remove AllArgsConstructor. It makes sense that this negatively impacts the API. In doing this I found out why it is there. Without it, you cannot have NoArgsConstructor either (generated or explicit). And without a no-arg constructor, things like inheriting from the class and json serialization fail. https://github.com/rzwitserloot/lombok/issues/1389
* With all the Groovy, you can't count on the compiler to tell you about all the places that these lombok class structure changes will impact the code. Unit tests show more of the problems, but the coverage isn't enough to feel confident about.

I may revisit this later and move over a few classes at a time. I'll start with just a PR for the commit where I convert Groovy to Java. Thanks for taking a look and raising concerns. I'll take these into account next time.

ezimanyi:
I definitely find making Jackson work with various flavors of constructors/builders confusing.  Some thoughts, don't know if they will work:
* I thought that if you provided a no-arg constructor and setters Jackson would be able to create an empty object and then use the setters to set any fields, but I could be wrong about that.
* If the above point doesn't work, another option might be to make the `@AllArgsConstructor` private in which case I believe Jackson will still find it.

If we do remove constructors (or other API methods) we could do it in two steps, first deprecating them and finding all usages, then removing them (as an example I did this with `Artifact`/`ExpectedArtifact` in spinnaker/kork#429).  Agree that with Groovy it's much more risky (and maybe in some cases we'd decide not worth the risk).

I think a reasonable path for handling cases like this might be:
* Move things over to `clouddriver-api` as-is (minus dependencies) and mark `@Alpha`
* Deprecate anything we want to remove, and find any usages in the code (agree that this is harder/riskier for groovy)
* After some amount of time to make sure we didn't miss any usages, remove the deprecated functions
* Promote to `@Beta` (potentially also pending some other feedback on the `@Alpha` API)

Anyway, just some thoughts for how we could handle changes we want to make to things as we move them over to `clouddriver-api`.

claymccoy:
@ezimanyi This makes sense to me. I'm giving it another try in smaller chunks. I just got the java conversion and Jackson annotation removal stuff in that you reviewed. Thanks for taking a look at everything.

robzienert:
Converting `clouddriver-core` to all Java is a good first step. I've had that on my TODO list for awhile.

I feel like Lombok makes more problems than it solves, especially when serialization is involved - there's so much boilerplate code that needs to be written that often amounts to more code than the actual code you want serialized 😆 . We can find a better solution, imo. Hopefully.

PR #4788, opened 2020-08-10
fix(build): Reduce Xmx and Xms

ezimanyi:
Our container builds run on a machine with only 7.2 GB of memory; while we do want to give a lot to the JVM we should leave a bit more for the everything else.

I was able to succesfully build the containers with these settings, so let's use them.


plumpy:
@Mergifyio backport release-1.21.x

PR #4725, opened 2020-07-07
feat(gce): Add GCP internal http(s) load balancer. 

takaaki7:
Added GCP Internal HTTP(S) load balancer support. spinnaker/spinnaker#5042
PR for deck: https://github.com/spinnaker/deck/pull/8397

Many parts are copy and pasted from HTTP Load Balancer implementation.
Differences are:
- It is regional 
- It has a network and subnet
- It doesn't have legacy http(s) health checks
- etc

PR #4789, opened 2020-08-11
fix(google): Fix assertion error on unknown target proxy

ezimanyi:
Our code to parse the target proxy type from a URL is overly strict and throws an IllegalArgumentException if it encounter something it doesn't recognize.

While I suppose that it might be helpful for us to know if there are different types of proxies we don't support encountered, this is very disruptive to users who might have these in their deployments as we just fail.

In nearly all cases, we are just checking if the result of getTargetProxyType is a particular type. This logic will still work if we just return an UNKNOWN type if we don't recognize the target proxy type, so let's do that instead.


ezimanyi:
A lot of the integration tests have been failing for GCE operations since #4725 was merged.  The internal load balancer handler is trying to decide whether it should handle a particular target proxy, but is getting an `IllegalArgumentException` because we can't figure out the type of target proxy.  This particular behavior of throwing an exception when we encounter a type we don't know is pre-existing, but I think somehow the code to add internal http(s) load balancers has caused it to get called in some cases where we don't know the type.

Ultimately it seems like the right fix here is just to return `UNKNOWN` rather than fail.  It should really be up to the callers of this function whether they want to fail if we can't parse a known target proxy type out of the URI.  In most cases, callers are just checking if the result is a particular type, so they would not want to fail.  But if we do ever encounter a case where we do need to fail, callers can of course throw their own exception if they get back `UNKNOWN`.

ezimanyi:
Whoops, just realized I had some local changes I forgot to push, added a commit.  If you haven't started reading, please just read the commits together...I intended to submit this all at once.

PR #4790, opened 2020-08-11
 refactor(kubernetes): remove NoopManifestProvider

maggieneterval:
* refactor(kubernetes): remove NoopManifestProvider 

  We are never actually wiring a NoopManifestProvider, since the KubernetesV2ManifestProvider is now always wired and ManifestProvider is now kubernetes-specific.

* refactor(kubernetes): simplify KubernetesManifestProvider callers 

  Now that there is only one ManifestProvider implementation, remove the interface and have callers wire KubernetesManifestProvider directly.

* refactor(kubernetes): rename KubernetesV2ManifestProvider to KubernetesManifestProvider


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/controllers/ManifestController.java
```diff
@@ -124,31 +114,24 @@ Manifest getDynamicManifestFromCluster(
...
+              account,
+              () ->
+                  manifestProvider.getClusterAndSortAscending(
+                      account, location, kind, app, cluster, criteria.getSort()));
+    } catch (Throwable t) {
+      log.warn("Failed to read {}", request, t);
+      return null;
+    }
+
+    if (manifests.isEmpty()) {
```
ezimanyi, line 131: Should this still be `manifests == null` here?  I think that before we were only throwing an exception if all of the manifest providers returned `null` (indicating that none of them could even handle the request), but we didn't throw if at least one returned an empty list (indicating it handled the request but didn't find clusters).  I think this will make us throw an exception if there are no clusters.  But I also might be mis-understanding as the prior logic of `List<List<? extends Manifest>>` as kind of confusing.

maggieneterval: No, I was confused, great catch...added one more commit marking `getClusterAndSortAscending` as nullable & changing this to a null check. We're handling the case of it being empty or containing too few elements below where we're wrapping the criteria check in a try/catch.

PR #4793, opened 2020-08-12
refactor(kubernetes): simplify Kubernetes credentials classes

maggieneterval:
* refactor(kubernetes): remove KubernetesCredentials interface 

  There is now only one implementation of the KubernetesCredentials interface: KubernetesV2Credentials; let's remove KubernetesCredentials to simplify the code.

  Most of this diff is removing type parameters from KubernetesCredentialFactory and KubernetesNamedAccountCredentials, as these are no longer necessary.

* refactor(kubernetes): remove KubernetesCredentialsFactory interface 

  The KubernetesCredentialsFactory has a single implementation, KubernetesV2Credentials.Factory. Move the default getKubeconfigFile method to this implementation and remove the interface. Replace the stub implementation used only by the KubernetesHealthIndicatorTest with mocks in the test itself.

* refactor(kubernetes): rename KubernetesV2Credentials to KubernetesCredentials 

* fix(kubernetes): fix KubernetesHealthIndicatorTest 

  The commit that removed the StubKubernetesCredentialsFactory broke this test because it tried to create a mocked item in a `thenReturn` statement of another mock. I was finding these mocks fairly dificult to reason about when trying to make this test pass, so decided to lift most of the mocks out of static methods and into a setup block (mostly because that is what I've seen Plumpy do when creating mocks, but also because it became a bit easier to see which mocks relied on which other mocks, and which were actually unnecessary). Let me know if you think this actually just made things more confusing...


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/KubernetesV2ProviderSynchronizable.java
```diff
@@ -73,11 +73,10 @@ public void synchronize() {
     Set<String> newAndChangedAccountNames = synchronizeAccountCredentials();
 
     // we only want to initialize caching agents for new or updated accounts
-    Set<KubernetesNamedAccountCredentials<KubernetesV2Credentials>> newAndChangedAccounts =
+    Set<KubernetesNamedAccountCredentials> newAndChangedAccounts =
         ProviderUtils.buildThreadSafeSetOfAccounts(
                 accountCredentialsRepository, KubernetesNamedAccountCredentials.class)
             .stream()
-            .map(c -> (KubernetesNamedAccountCredentials<KubernetesV2Credentials>) c)
```
ezimanyi, line 80: 💯 🥳

PR #4794, opened 2020-08-12
feat(gce): add support for the predictive autoscaler

plumpy:


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleAutoscalingPolicy.groovy
```diff
@@ -39,6 +39,12 @@ class GoogleAutoscalingPolicy {
   @ToString(includeNames = true)
   static class CpuUtilization {
     Double utilizationTarget
+    PredictiveMethod predictiveMethod
+
+    enum PredictiveMethod {
+      NONE,
+      STANDARD
```
maggieneterval, line 46: Do we expect there to be multiple predictive methods in the future, or should the UI simply be a checkbox for now?

plumpy: I think it's relatively likely there will be more in the future, but we can always change the the UI if they ever add that, so I'd say a checkbox for now...

maggieneterval:
@plumpy Two questions about the UI for this:

1. Any opinion on whether the "Enable predictive autoscaling" checkbox should be in the top-level "Basic Settings" section of the scaling policy dialog (which might look the cleanest), or in the CPU utilization collapsible section (which will look a bit clunkier but more accurately reflects the underlying data model), or in its own new section (maybe better if we expect there to be more toggles related to predictive autoscaling in the future)?

FtCfXP4Ua76

2. Should we feature-flag the checkbox or gate it on the account in question being alpha listed?

plumpy:
1. As we discussed in chat, it requires some CPU utilization metrics to be set, so should be in that section (the one that appears after clicking "Add CPU Utilization metric")
2. I think feature-flag is probably best for now. We can remove it once it moves to prod.

PR #4796, opened 2020-08-12
fix(saga): Seek to the lowest index when resuming

robzienert:
We had a failure in a Saga that caused a restart, but the saga seeked to the incorrect command, skipping one command. This was caused by selecting the `max` index returned by the `Seeker`s, which are responsible for figuring out which event to start processing again. I think there may be room to simplify things by removing the concept of `Seeker` entirely, but I want to investigate that separately some other time. The fix here is to select the `min`.

There's a couple things out of this investigation that I think are worth fixing beyond this:

1. Conditions are evaluated each time: If a condition were non-deterministic, the Saga would end up making the flow erratic. I'd like to add a `ConditionEvaluated` event that would cache the result of a condition.
2. The way a `SagaFlow` is incrementally evaluated is plenty useful, but makes it hard for seeking, since all of the logic for how this is done is internal to the iterator; I'd like to move some of the logic out into a separate class that both the iterator and seekers can use independently.
3. I noticed that in some cases, the events are stored into the DB backwards, specifically WRT to `sagaSaved` event. This does not have a functional impact, but it does make reading the log a little confusing.


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/AttachTitusServiceLoadBalancers.java
```diff
@@ -62,9 +63,7 @@ public Result apply(@Nonnull AttachTitusServiceLoadBalancersCommand command, @No
         titusClientProvider.getTitusLoadBalancerClient(
             description.getCredentials(), description.getRegion());
     if (loadBalancerClient == null) {
-      // TODO(rz): This definitely doesn't seem like something to casually skip?
-      saga.log("Unable to create load balancing client in target account/region");
-      return new Result();
+      throw new TitusException("Unable to create load balancing client in target account/region");
```
robzienert, line 66: This is an unrelated change, but while I was poking around, it seemed like a bad thing to keep. I'd rather bomb out the saga and let it retry instead of happily skipping if the load balancer client can't be created (for whatever reason).

I believe this original behavior was carried over from the original, pre-saga, Groovy code.

PR #4798, opened 2020-08-12
fix(api): remove jackson annotations from model

claymccoy:
Not having the jackson annotations directly on the model allows them to more easily be moved to an external library without needing a Jackson dependency.

### DIFF
--- a/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonClusterSpec.groovy
+++ b/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonClusterSpec.groovy
@@ -18,11 +18,12 @@
 package com.netflix.spinnaker.clouddriver.aws.model
 
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.clouddriver.jackson.ClouddriverApiModule
 import spock.lang.Specification
 
 class AmazonClusterSpec extends Specification {
   void "should serialize null loadBalancers and serverGroups as empty arrays"() {
-    def objectMapper = new ObjectMapper()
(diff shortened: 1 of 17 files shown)


clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonClusterSpec.groovy
```diff
@@ -18,11 +18,12 @@
 package com.netflix.spinnaker.clouddriver.aws.model
 
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.clouddriver.jackson.ClouddriverApiModule
 import spock.lang.Specification
 
 class AmazonClusterSpec extends Specification {
   void "should serialize null loadBalancers and serverGroups as empty arrays"() {
-    def objectMapper = new ObjectMapper()
+    def objectMapper = new ObjectMapper().registerModule(new ClouddriverApiModule())
```
ezimanyi, line 26: In `orca` we've created a centralized `OrcaObjectMapper` to [handle registering all the extensions](https://github.com/spinnaker/orca/blob/9d1ff489fd15101c2ddcffe7f654105863086853/orca-core/src/main/java/com/netflix/spinnaker/orca/jackson/OrcaObjectMapper.java#L42) and hide that from consumers
I wonder if we should do something similar for `clouddriver` too, to ensure we don't run into bugs where a particular `ObjectMapper` doesn't have these extensions.

claymccoy: There is a centralized one in CloudDriverConfig that I added the new module too. But some unit tests tend to create their own.

PR #4799, opened 2020-08-13
test(kubernetes): Add some tests on replaceAll

ezimanyi:
* test(kubernetes): Convert kubernetes artifact tests to Java

  There aren't a lot of these, and it will be easier to add new ones
  (and easier to refactor and see at compile-time what tests use functions) if we convert them to Java.

* test(kubernetes): Add some tests on replaceAll

  The functionality to replace artifacts in a manifest before deploying is fairly important but currently completely untested. Give it some good test coverage before making changes.

* refactor(kubernetes): Minor simplifications to artifact replacer

  (1) There are a lot of small functions with unclear names that make the logic hard to follow; inline these where it makes the logic clearer.
  (2) The result of a document.read lets you iterate over the resulting nodes, and convert them to text. This is much clearer (and will perform better) than using an ObjectMapper for this.
  (3) The Artifact passed in to replaceIfPossible is always non-null, because it ultimately comes from the ImmutableCollection created by filterKubernetesArtifactsByNamespaceAndAccount. That function also filters out artifacts without a type set so we don't need to explicitly check again (though we'll still be safe and invert the
  .equals below because we can't really guarantee with an annotation that this particular artifact has a non-null type even though it effectively always does).

* refactor(kubernetes): Move ArtifactProvider to clouddriver-kubernetes

  This is only implmented/used in the kubernetes provider, and it's not clear whether this is an abstraction useful to other providers. Let's move it to clouddriver-kubernetes. If this becomes something useful for other providers we can abstract common functionality to an interface in clouddriver-core at that point.

  We can probably just delete the interface as it has only one implementation, but keeping it for now.


ezimanyi:
This is mostly just converting tests to Java and adding more tests; I have some actual refactor/perf fixes for this code I'm working on but the PR was getting big so wanted to send this part.

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/KubernetesVersionedArtifactConverterTest.java
```diff
@@ -16,52 +16,141 @@
...
+                .build());
+    assertThat(name).isEqualTo("my-other-rs-_--v010");
+  }
+
+  @Test
+  void findsMatchingVersionByEquality() {
+    KubernetesManifest manifest1 = getStubManifest();
+    KubernetesManifest manifest2 = getStubManifest();
+    manifest1.put("data", ImmutableMap.of("key", 1, "value", 2));
+    manifest2.put("data", ImmutableMap.of("key", 3, "value", 2));
```
maggieneterval, line 91: Are we doing this just so we don't have two otherwise-identical stub manifests?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/Replacer.java
```diff
@@ -83,7 +63,8 @@ private String nameFromReference(String s) {
 
   @Nonnull
   ImmutableCollection<Artifact> getArtifacts(DocumentContext document) {
-    return mapper.convertValue(findAll(document), new TypeReference<List<String>>() {}).stream()
+    return Streams.stream(document.<ArrayNode>read(findPath).elements())
```
maggieneterval, line 66: > This is much clearer (and will perform better) than using an ObjectMapper for this.

Agree it is clearer, but curious how you knew it would perform better?

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/KubernetesVersionedArtifactConverterTest.java
ezimanyi: Yeah, this was converted from Groovy.  Ideally it would change something more meaningful in the manifest rather than this random `data` key, but it is indeed to make sure they are different.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/Replacer.java
ezimanyi: Good point; I didn't actually profile this particular call before/after so I can't say for sure.  I probably should have said "likely will perform better".

The reason I think it will perform better is that in general serialization/deserialization/conversion with `ObjectMapper` is pretty slow because it uses reflection to iterate over all properties (or builders/setters, depending on how it's configured) of the object and the target type to do the conversion.

Here since we're replacing that with looping over the found elements and calling a single `getText` function on each I'd expect it to be much faster.   In my profiling this conversion wasn't the bottleneck (compiling the JsonPath was), so it's probably not super important.  But anyway that's why I'd expect this to be faster.

maggieneterval: 💯 thank you for explaining! (wasn't so much worried about actual performance improvements from this, just want to develop a better sense for these things like you have)

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/KubernetesVersionedArtifactConverterTest.java
ezimanyi: Pushed a commit to add a comment as it was kind of confusing.  (Also removed the "value: 2" from both which was confusing.)

PR #4800, opened 2020-08-14
fix(titus): Re-enable saga retry on INVALID_ARGUMENT error due to rate exceeded

jonsie:
In https://github.com/spinnaker/clouddriver/pull/4796 we resolved an issue wherein a saga retry would skip a command under certain scenarios.  The particular failure scenario we had around retrying this error was due to that bug - now that it is fixed we should be able to safely retry here.


jonsie:
We are going to hold off on merging this until https://github.com/spinnaker/clouddriver/pull/4796 has been deployed and verified.

PR #4805, opened 2020-08-17
refactor(kubernetes): Move name parsing logic to KubernetesCoordinates

ezimanyi:
* refactor(kubernetes): Move name parsing logic to KubernetesCoordinates 

  A lot of the Kubernetes logic parses a kind-name String into a Pair<KubernetesKind, String>. This leads to very unclear code as we are always calling Pair#getLeft and Pair#getRight instead of using more clearly-named functions. While Pairs might be reasonable for short-term holding of data in functions or streams, they quickly lead to complicated code when passed across function boundaries.

  We already have a KubernetesCoordinates class that represents a kind and a name (as well as a namespace). (And in most cases where we're parsing a kind-name string we also have a namespace around too.)

  Let's prepare to move clients to instead parse into KubernetesCoordinates. This change does the following:
  * Explicitly write out the auto-generated constructor and have it translate null to empty for name/namespace. This allows us to guarantee that none of the fields are null regardless of how the builder is used. We will explicitly prohibit null KubernetesKind because we are always passing in a non-null kind (as our kind parsing logic returns KubernetesKind.NONE instead of null).
  * Add a method to the builder fullResourceName that takes a "kind name" String and parses it to set both the kind and the name.  (The other auto-generated builder methods will still be injected by Lombok, so you can still call .name() or .kind() separately if you want.)
  * Deprecate KubernetesManifest.fromFullResourceName; update it to call the method we just wrote and wrap the result in a Pair.

* refactor(kubernetes): Add some functions that accept KubernetesCoordinates 

  There are a lot of places where we pass around a separate kind/namespace/name that would be much clearer if we could just directly pass KubernetesCoordinates
  (which avoids unpacking a nice object into individual parameters).

  Let's add a few of these functions where they will be useful, in all cases augmenting a function that accepts these separately to become one that accepts a KubernetesCoordinates. I've deprecated the old versions (and will update usages in a later commit) and rewritten them to call the new version.

  I had to make a small change to topPod, which was only checking if the pod was null (rather than null or empty); as we're going through KubernetesCoordinates now we will never have a null name as that will always be changed to empty string.

* refactor(kubernetes): Update descriptions to use new function 

  These all passed the Pair to a KubernetesCoordinatesBuilder anyway, so we can just do this inline now.

* refactor(kubernetes): Refactor other locations to use new function 

  The various Provider classes as well as the on-demand caching agent can be updated now to create a KubernetesCoordinates class instead of parsing the name into a Pair.

  This also simplifies later code in these functions, which can be updated to pass the KubernetesCoordinates on to other functions rather than extracting out the parts.

  While making this change, update the catch blocks around the name parsing to catch an IllegalArgumentException rather than a generic Exception.

* refactor(kubernetes): Use new functions in Operations 

  Same as the last two commits, but update the Operation classes to use KubernetesCoordinates instead of Pair<KubernetesKind, String>.

  I also removed a null-check in CanLoadBalance; a while back I made getHandler always non-null but didn't remove all the now-unnecessary ones.

* refactor(kubernetes): Extract some common logic in provider code 

  When doing the last refactor I realized that a number of providers have almost identical logic. Make a function in CacheUtils that takes an account/namespace/name and returns the corresponding cache data item so we don't need to keep duplicating this.

  One small change (which I considered doing before and commented on) is that we no longer filter relationships in the server group case. The performance benefits here are not worth the additional complexity of not re-using the code (particularly that we only do this filtering when requesting a single server group and not when requesting them all).

* test(kubernetes): Move tests on name parsing 

  Now that we've removed all uses of fromFullResourceName, let's move those tests to the function that we're recommending (and what the deprecated method itself calls). This just converts the exact same tests to Java and moves them.

  I'll leave this and the other deprecated methods around at least until after we cut the release so there is some time for anyone who might have a custom build to migrate.

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2ServerGroup.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2ServerGroup.java
@@ -31,6 +31,7 @@ import com.netflix.spinnaker.clouddriver.kubernetes.caching.Keys;
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.Keys.InfrastructureCacheKey;
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.agent.KubernetesCacheDataConverter;
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.view.provider.data.KubernetesV2ServerGroupCacheData;
+import com.netflix.spinnaker.clouddriver.kubernetes.description.KubernetesCoordinates;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesApiVersion;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesKind;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesManifest;
@@ -193,13 +194,11 @@ public final class KubernetesV2ServerGroup implements KubernetesResource, Server
     KubernetesManifestTraffic traffic = KubernetesManifestAnnotater.getTraffic(manifest);
     Set<String> explicitLoadBalancers =
         traffic.getLoadBalancers().stream()
-            .map(KubernetesManifest::fromFullResourceName)
-            .map(
-                p ->
-                    KubernetesManifest.getFullResourceName(
-                        p.getLeft(),
-                        p.getRight())) // this ensures the names are serialized correctly when
-            // the get merged below
+            // TODO(ezimanyi): Leaving this logic and the comment below, but I'm not sure that we
+            // still need the logic to parse then re-serialize the name.
+            // this ensures the names are serialized correctly when the get merged below
+            .map(lb -> KubernetesCoordinates.builder().fullResourceName(lb).build())
+            .map(c -> KubernetesManifest.getFullResourceName(c.getKind(), c.getName()))
             .collect(Collectors.toSet());
 
     Set<String> loadBalancers =
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/KubernetesCoordinates.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/KubernetesCoordinates.java
@@ -17,14 +17,57 @@
 
 package com.netflix.spinnaker.clouddriver.kubernetes.description;
 
+import com.google.common.base.Splitter;
+import com.google.common.base.Strings;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesKind;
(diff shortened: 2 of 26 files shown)


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/KubernetesCoordinates.java
```diff
@@ -17,14 +17,57 @@
...
 
-@Builder
+@NonnullByDefault
 @Value
 public class KubernetesCoordinates {
   private final KubernetesKind kind;
   private final String namespace;
   private final String name;
+
+  @Builder(toBuilder = true)
```
maggieneterval, line 39: Do we need a `toBuilder`?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2ServerGroup.java
```diff
@@ -193,13 +194,11 @@ public static KubernetesV2ServerGroup fromCacheData(KubernetesV2ServerGroupCache
...
-                p ->
-                    KubernetesManifest.getFullResourceName(
-                        p.getLeft(),
-                        p.getRight())) // this ensures the names are serialized correctly when
-            // the get merged below
+            // TODO(ezimanyi): Leaving this logic and the comment below, but I'm not sure that we
+            // still need the logic to parse then re-serialize the name.
+            // this ensures the names are serialized correctly when the get merged below
+            .map(lb -> KubernetesCoordinates.builder().fullResourceName(lb).build())
+            .map(c -> KubernetesManifest.getFullResourceName(c.getKind(), c.getName()))
```
maggieneterval, line 201: I am very confused...why do we need the second map?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/KubernetesCoordinates.java
maggieneterval: Never mind, just got to the part where we use this.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesV2ServerGroup.java
ezimanyi: I am also very confused...because there is a comment `this ensures the names are serialized correctly when the get merged below` I was afraid to remove this.  This should be just an operation and its inverse...but presumably there was some condition where it wasn't an exact inverse which is why we need to do this?

That's why I left the comment because I was afraid of breaking some subtle use case and didn't want to spend more time digging into why we need these.  (The original PR adding it #3060 doesn't have any clues beyond that comment.)

maggieneterval: 🆗

PR #4807, opened 2020-08-17
perf(kubernetes): Request metrics and events in parallel

ezimanyi:
* perf(kubernetes): Only deserialize pod status

  Rather than convert the entire manifest to a V1Pod, only convert the status as that's all this function is using.

  Ideally we'd only do a single conversion to the full appropriate type when we get a value back from the cluster, and pass that around, but as long as we're still doing these ad-hoc intercoversions, it's worth avoiding doing more than we need.

* perf(kubernetes): Request metrics and events in parallel

  When deck requests a manifest, it wants the events and metrics as well. This means that we make 3 sequential calls to the cluster, each with a bit of latency, that ends up making the overall call longer than needed.

  Let's speed this up by sending the event and metrics requests to an executor pool that runs them on a separate thread. This allows us to only wait as long as the longest request instead of the sum of all of the request times.

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesManifestProvider.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesManifestProvider.java
@@ -22,6 +22,7 @@ import static com.netflix.spinnaker.clouddriver.kubernetes.caching.Keys.LogicalK
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.Keys;
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.agent.KubernetesCacheDataConverter;
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.view.model.KubernetesV2Manifest;
@@ -52,7 +53,9 @@ import org.springframework.stereotype.Component;
 public class KubernetesManifestProvider {
   private final KubernetesCacheUtils cacheUtils;
   private final KubernetesAccountResolver accountResolver;
-  private final ExecutorService executorService = Executors.newCachedThreadPool();
+  private final ExecutorService executorService =
+      Executors.newCachedThreadPool(
+          new ThreadFactoryBuilder().setNameFormat(getClass().getSimpleName() + "-%d").build());
 
   @Autowired
   public KubernetesManifestProvider(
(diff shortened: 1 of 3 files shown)


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesManifestProvider.java
```diff
@@ -69,26 +77,45 @@ public KubernetesV2Manifest getManifest(
...
+            : Futures.immediateFuture(ImmutableList.of());
+
+    Future<List<ContainerMetric>> metrics =
+        includeEvents
+                && coords.getKind().equals(KubernetesKind.POD)
+                && credentials.isMetricsEnabled()
+            ? executorService.submit(() -> getPodMetrics(credentials, coords))
+            : Futures.immediateFuture(ImmutableList.of());
+
     KubernetesManifest manifest = credentials.get(coords);
```
maggieneterval, line 92: Is there any reason to not also request the manifest itself on a separate thread?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesManifestProvider.java
```diff
@@ -45,6 +52,7 @@
 public class KubernetesManifestProvider {
   private final KubernetesCacheUtils cacheUtils;
   private final KubernetesAccountResolver accountResolver;
+  private final ExecutorService executorService = Executors.newCachedThreadPool();
```
plumpy, line 55: If you do this, the threads will have names, which is helpful for debugging... they'll show up in stack traces, for example:
```
Executors.newCachedThreadPool(
    ThreadFactoryBuilder().setNameFormat(getClass().getSimpleName() + "-%d").build()
  );
```

(I forgot to mention that on your last PR, sorry.)

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesManifestProvider.java
ezimanyi: I hesitated between the two; it seems like the advice on the internet is that often it's more clear to just send all your tasks to the executor so there's no asymmetry around one of them happening on the main thread and the others happening in the executor.

But in this case since we do have something like a "parent task" (getting the manifest) and ancillary tasks (getting the metrics/events) it seemed reasonable to get the manifest on the main thread.  (This also completely avoids the need to use the executor in the case where we don't want metrics/events.).

Anyway, that was my thinking and why I leaned towards fetching the manifest on the main thread.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesManifestProvider.java
ezimanyi: Oh, right, I've seen you do that before.  I'll update both places to do this.

PR #4817, opened 2020-08-19
refactor(core): Add type bounds to AtomicOperation classes

ezimanyi:
These interfaces use raw types, which forces all implementations to do the same as well (as you can't override a raw type with a parametrized type).

I've added a few type bounds so that implementations can avoid using raw types; this should in no way affect clients that are still using raw types (as evidenced by the fact that I'm not changing any of the implementations here).

Obviously Map<String, Object> is not great as a parameter type, but that's what the Map actually is, and it's better than a raw Map.


ezimanyi:
After this merges, implementations can updated to stop using raw types at their own pace; I plan to do this for the Kubernetes provider.

PR #4819, opened 2020-08-19
refactor(kubernetes): Enable error-prone and compiler warnings

ezimanyi:
* fix(kubernetes): Fix errorprone TypeParameterUnusedInFormals 

  [TypeParameterUnusedInFormals] Declaring a type parameter that is only used in the return type is a misuse of generics: operations on the type parameter are unchecked, it hides unsafe casts at invocations of the method, and it interacts badly with method overload resolution.

  This check is 100% correct here; the only advantage of using the generic here is that we can hide the casting in getRequiredField. In fact, let's just get rid of this method and force callers to do the casting; keeping the method just to throw the exception seems more confusing than inlining.

* fix(kubernetes): Fix error-prone issues 

  This fixes a few error-prone issues that were trivial to fix, and were easier than suppressing the whole warning class.

  [EmptyBlockTag] A block tag (@param, @return, @throws, @deprecated) has an empty description. Block tags without descriptions don't add much value for future readers of the code; consider removing the tag entirely or adding a description.

  In this case just delete the tag as the name is enough.

  [UnusedNestedClass] This nested class is unused, and can be removed.

  We were indeed not using this nested class.

  [PublicConstructorForAbstractClass] Constructors of on an abstract class can be declared protected as there is never a need for them to be public.

  [UnnecessaryLambda] Returning a lambda from a helper method or saving it in a constant is unnecessary; prefer to implement the functional interface method directly and use a method reference instead.

  Turns out we weren't using this lambda saved to a variable at all.

  [UnusedMethod] Private method is never used.

  error-prone doesn't seem to be able to detect that JUnit is using these as a MethodSource for parametrized tests; just suppress the warnings here.

  Finally, this isn't from error-prone, but there's a compiler warning that KubernetesNamedAccountCredentials overrides equals but doesn't call the super class. We don't want to do that here (because there's nothing in the super class and because we only want some explicitly-included fields) so add callSuper = false to stop the warning.

* refactor(kubernetes): Enable compilation warnings 

  This commit enables compilation warnings for clouddriver-kubernetes. It turns off a few categories that either generate too many false positives for our code (with a detailed comment) or that have too many actual issues we need to fix first.

  The plan is to fix the categories of issues one-by-one and re-enable them.

* feat(kubernetes): Add error-prone to clouddriver-kubernetes 

  This commit adds the error-prone static analysis checker to clouddriver-kubernetes. It disables any of the checks that have actual issues.

  The plan is that we can fix the issues in a given category then enable that category, until all of the issues are fixed.

### DIFF
--- a/clouddriver-kubernetes/clouddriver-kubernetes.gradle
+++ b/clouddriver-kubernetes/clouddriver-kubernetes.gradle
@@ -1,3 +1,7 @@
+plugins {
+  id("net.ltgt.errorprone") version "1.2.1"
+}
+
 tasks.compileGroovy.enabled = false
 sourceSets.main.java.srcDirs = ['src/main/java']
 
@@ -21,9 +25,21 @@ tasks.withType(JavaCompile) {
     '-Xlint:-rawtypes',
     '-Xlint:-unchecked',
   ]
+
+  options.errorprone.disable(
+    "DefaultCharset",
+    "EmptyCatch",
+    "ImmutableEnumChecker",
+    "JdkObsolete",
+    "MissingOverride",
+    "StringSplitter",
+    "UnnecessaryParentheses",
+    "UnusedVariable"
+  )
 }
 
 dependencies {
+  errorprone("com.google.errorprone:error_prone_core:2.4.0")
   implementation project(":clouddriver-api")
   implementation project(":clouddriver-artifacts")
   implementation project(":clouddriver-core")
--- a/clouddriver-kubernetes/clouddriver-kubernetes.gradle
+++ b/clouddriver-kubernetes/clouddriver-kubernetes.gradle
@@ -1,6 +1,28 @@
 tasks.compileGroovy.enabled = false
 sourceSets.main.java.srcDirs = ['src/main/java']
 
+tasks.withType(JavaCompile) {
+  options.compilerArgs += [
+    '-Xlint:all',
+
+    // This check ensures that all annotations are handled at compile time, but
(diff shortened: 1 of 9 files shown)


robzienert:
Awesome! What's blocking adding this elsewhere?

ezimanyi:
> What's blocking adding this elsewhere (if anything)?

The one thing I had to do for this to work in `clouddriver-kubernetes` is remove `@Slf4j` annotations.  All the default checks seem to have no problem with the other Lombok annotations (`@Data`, `@Getter`, etc.) but `@Slf4j` was breaking at least one of them.  I got around that by just replacing the annotation with a field in `clouddriver-kubernetes` (#4816) which is probably better anyway as the annotation is only replacing one line of code.  So we'd have to do that everywhere (or disable whichever check was erroring).

I wanted to use this in `clouddriver-kubernetes` for a bit to see how it goes/any issues we ran into, but would definitely be on board with running it globally if all goes well.  We'd have to think about rollout, but probably each sub-project could just exclude everything that has failures, then let the maintainer of the sub-project remove exclusions as they fix issues.

I also want to figure out how to more prominently surface these warnings other than in the compilation logs before rolling out beyond `clouddriver-kubernetes` as otherwise I'm afraid we'll just generate a lot of noise that nobody sees.  (Though I guess actual errors would fail the build.)

I think there was a static analysis item on the Core Infrastructure Best Practices Badge list (for Spinnaker graduation), so this would definitely help us with that if we can roll it out everywhere.

PR #4822, opened 2020-08-19
fix(core): Remove extraneous semicolons

ezimanyi:
The upgrade to google-java-format pulled semicolons after methods to be on the next line (probably to make them more visible because they are in general not necessary); let's just remove them.


ezimanyi:
Makes sense, thanks for the insight into the parser!

PR #4825, opened 2020-08-20
perf(kubernetes): live lookups for dynamic target selection (part 1 of 2!)

maggieneterval:
* perf(kubernetes): live lookups for dynamic target selection 

* perf(kubernetes): avoid creating unnecessary KubernetesV2Manifest objects 

* perf(kubernetes): getDynamicManifestFromCluster returns only KubernetesCoordinates instead of entire manifest 

* test(kubernetes): add getClusterAndSortAscending to KubernetesDataProviderIntegrationTest 

  In a followup PR, will likely refactor getClusterAndSortAscending in order to:
  (1) move business logic out of the ManifestController and into the ManifestProvider; the controller shouldn't need to know anything about what largest/smallest/newest mean
  (2) expose a new endpoint for Orca to ask for all cluster siblings of a given resource (non-sorted) to support rollout strategies not relying on an up-to-date cache

  (So will add better test coverage once both contracts fully live in the provider)

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/KubernetesCoordinates.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/KubernetesCoordinates.java
@@ -20,6 +20,7 @@ package com.netflix.spinnaker.clouddriver.kubernetes.description;
 import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesKind;
+import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesManifest;
 import com.netflix.spinnaker.kork.annotations.FieldsAreNullableByDefault;
 import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
 import java.util.List;
@@ -69,5 +70,20 @@ public class KubernetesCoordinates {
       this.name = parts.get(1);
       return this;
     }
+  }
+
+  /**
+   * Given a full KubernetesManifest object, parses out the kind, namespace, and name to create a
+   * corresponding KubernetesCoordinates object.
+   *
+   * @param manifest the manifest to parse
+   * @return the KubernetesCoordinates object
+   */
+  public static KubernetesCoordinates fromManifest(KubernetesManifest manifest) {
+    return KubernetesCoordinates.builder()
+        .kind(manifest.getKind())
+        .namespace(manifest.getNamespace())
+        .name(manifest.getName())
+        .build();
   }
 }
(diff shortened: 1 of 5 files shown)


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/manifest/KubernetesManifestAnnotater.java
```diff
@@ -299,4 +299,8 @@ public static KubernetesManifest getLastAppliedConfiguration(KubernetesManifest
         new TypeReference<KubernetesManifest>() {},
         null);
   }
+
+  public static String getManifestCluster(KubernetesManifest manifest) {
+    return manifest.getAnnotations().get(CLUSTER);
+  }
```
ezimanyi, line 305: This function should probably either be annotated `@Nullable` or do a `Strings.nullToEmpty` on the result.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/KubernetesCoordinates.java
```diff
@@ -70,4 +71,19 @@ public KubernetesCoordinatesBuilder fullResourceName(String fullResourceName) {
...
   }
+
+  /**
+   * Given a full KubernetesManifest object, parses out the kind, namespace, and name to create a
+   * corresponding KubernetesCoordinates object.
+   *
+   * @param manifest the manifest to parse
+   * @return the KubernetesCoordinates object
+   */
+  public static KubernetesCoordinates fromManifest(KubernetesManifest manifest) {
```
ezimanyi, line 82: 💯 I kind of wished this function existed in one of my refactors last week, but didn't actually write it.

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
```diff
@@ -517,6 +521,20 @@ void getArtifactsWrongAccount(SoftAssertions softly) {
     softly.assertThat(artifacts).isEmpty();
   }
 
+  @Test
+  void getClusterAndSortAscending(SoftAssertions softly) {
+    List<KubernetesManifest> manifests =
+        manifestProvider.getClusterAndSortAscending(
+            ACCOUNT_NAME, "backend-ns", "replicaSet", "replicaSet backend", Sort.AGE);
+    softly.assertThat(manifests).isNotNull();
```
ezimanyi, line 529: In this case I'd actually just do a regular `assertThat` because if `manifests` is `null` the next assertion will just NPE anyway.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/manifest/KubernetesManifestAnnotater.java
maggieneterval: 👍 👍 Good call!

PR #4826, opened 2020-08-20
fix(*): Fix gradle deprecation warnings

ezimanyi:
Because I'm in a warning-fixing mood this week...this fixes the warning when running builds that deprecated features are used.

These were all pretty easy to fix:
* Instead of force, need to use version constraints
* Instead of testCompile/testRuntime, need to use testCompileOnly/testRuntimeOnly.

### DIFF
--- a/clouddriver-docker/clouddriver-docker.gradle
+++ b/clouddriver-docker/clouddriver-docker.gradle
@@ -5,7 +5,7 @@ dependencies {
   implementation project(":cats:cats-core")
 
   compileOnly "org.projectlombok:lombok"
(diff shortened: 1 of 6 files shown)


ezimanyi:
Oh no, I broke something...will investigate.

ezimanyi:
Arg, this might actually be impossible...as I should have predicted the two `strict` requriements for elasticsearch are fine when compiling each separate module but lead to an unresolvable version conflict in `clouddriver-web`.

Presumably it was picking one before (so one of the two `force` commands was not taking effect), so maybe we should just pick the one it was using?

PR #4827, opened 2020-08-21
feat(test): Clouddriver integration tests

german-muzquiz:
~~This is a work in progress for introducing spring integration tests into clouddriver.~~

This PR introduces spring integration tests for clouddriver.

The idea is to have these tests run by explicitly providing a flag (`-PintegrationTests`), so that usual build times during development are not affected, but include them during PR builds to catch possible bugs before they are merged.
These tests can complement unit tests and end to end testing.

Tech stack:
- `@SpringBootTest` to start a real clouddriver server on a random port
- `Testcontainers` to use Mysql as backend storage (requires docker, already used by cats-sql module)
- ~~Spock framework for writing BDD like tests~~
- Rest-assured (new dependency) for interacting with and testing the rest api
- K3s is used as a test container for kubernetes provider tests

Tests ran in github action successfully:
image


robzienert:
The language choice has no material difference on logs, so I think we're safe there.

german-muzquiz:
cc @ezimanyi @maggieneterval

allisaurus:
@ezimanyi @german-muzquiz  Re: project organization, would we prefer to have a separate top-level `integration-tests` project that contains all the tests, or have provider-specific tests live in their respective sub-projects as an additional sourceset and define them as separate tasks programmatically like we do for [unit tests](https://github.com/spinnaker/clouddriver/blob/master/build.gradle#L65)?

In general, I'm curious about the extensibility and potential brittleness of having them all in a top-level package that is being frequently updated by many contributors (esp. if `spring.config` properties need to be shared or frequently updated between them), but maybe there's some performance benefits or app context magic in the top-level approach that mitigates these concerns.

german-muzquiz:
@allisaurus that's a good concern. I preferred to isolate kubernetes integration tests in a separate module, just so that they're not mixed with unit tests and are easy to identify. For example, somebody could argue that all integration tests need to live in `clouddriver-web`. I see that having a different source set in each individual provider module could also work.

Regarding isolation between providers in the same `integration-tests` module, it can be achieved by:
1. Separation just with folder names, there's a `kubernetes` folder under resources as well as a package under src.
2. Separation with different source sets in `integration-tests` module (kubernetesTests, ecsTest, etc.)

About the time it will take to run them, I think it's better to have different clouddrivers started for each different provider, than one shared across all providers. It could spend more time but we would not step on each other's toes having a single `clouddriver.yml`.

integration-tests/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/it/requests/DeployManifestRequest.java
```diff
@@ -0,0 +1,89 @@
...
+import static io.restassured.RestAssured.given;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.netflix.spinnaker.clouddriver.kubernetes.it.helpers.TestManifest;
+import io.restassured.response.Response;
+import java.util.*;
+
+public class DeployManifestRequest extends BaseRequest<String> {
+
+  public Map<String, Object> deployManifest = new HashMap<>();
```
maggieneterval, line 28: Is this meant to simulate an instance of [`KubernetesDeployManifestDescription`](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/manifest/KubernetesDeployManifestDescription.java#L29)? Any reason we are not using that class directly?

german-muzquiz: This was done intentionally, the reasoning is that by not relying on the existing implementation classes, test code is more resilient to any refactors or changes in the internal implementation. If the contracts of clouddriver with external dependencies doesn't change, then integration tests doesn't need to change either. Another benefit is that if someone changes one of these contracts inadvertently, then having to update integration tests will make it more clear that probably orca would need to also be updated.

german-muzquiz:
> Is there an issue or doc I missed outlining the plan for the scope and maintenance of these integration tests? My biggest concern is that the tests seem coupled to the existence and ordering of certain Orca tasks in the Deploy Manifest stage. If I change the order of or remove one of these tasks in Orca, it won't break these integration tests, but they'll no longer reflect reality. How should we plan to keep these tests as useful as possible given that they only test Clouddriver? I'd love to have a better understanding of exactly which types of bugs we are trying to prevent with these tests to help answer that question.

There's not an issue or doc, but probably I need to create one for that type of discussion. The types of bugs that these tests intent to catch are the ones that can only be spotted when starting clouddriver and doing real requests to it. For example:
1. Bugs related with the wrong wiring of bean dependencies
2. Faulty startups with different configurations in `clouddriver.yml`
3. Inadvertently changing the contract of the rest API: at least this will make it obvious that for example orca would need to be also updated

There's a bit of overlap between end to end tests and these integration tests. I think the main focus of end to end tests is to exercise that all services play well together and are closer to how real users interact with Spinnaker. In your example of changing orca tasks not breaking clouddriver tests, I think that should be catched either by orca integration tests or end to end tests.

german-muzquiz:
Re: Having the tests live in each cloudprovider specific module works for me.

integration-tests/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/it/requests/ForceCacheRefreshRequest.java
```diff
@@ -0,0 +1,73 @@
...
+  public String getName() {
+    return name;
+  }
+
+  @Override
+  public Response executeAndValidate() {
+    Response resp =
+        given()
+            .contentType("application/json")
+            .body(this)
```
ezimanyi, line 67: Similar to my comment on `DeployManifestRequest` I think that the level of indirection we have from an additional class that we serialize does not outweigh the benefits.  It feels like for a simple request like this we could probably just have clients create the body of the request themselves as a `Map` rather than introduce a test-specific class for serializing this.

integration-tests/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/it/helpers/TestManifest.java
```diff
@@ -0,0 +1,72 @@
...
+public class TestManifest {
+
+  private final List<Map<String, Object>> contents;
+  private ObjectMapper objectMapper = new ObjectMapper();
+
+  public TestManifest(String path) {
+    ResourceLoader resourceLoader = new DefaultResourceLoader();
+    try {
+      InputStream is = resourceLoader.getResource(path).getInputStream();
+      Yaml yaml = new Yaml();
```
ezimanyi, line 39: We should probably use `new Yaml(new SafeConstructor())` here, both to match what's in `orca` and to prevent deserialization attacks.

integration-tests/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/it/helpers/TestManifest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.springframework.core.io.DefaultResourceLoader;
+import org.springframework.core.io.ResourceLoader;
+import org.yaml.snakeyaml.Yaml;
+
+public class TestManifest {
+
+  private final List<Map<String, Object>> contents;
+  private ObjectMapper objectMapper = new ObjectMapper();
+
+  public TestManifest(String path) {
```
ezimanyi, line 35: Is there a reason we need to return a class `TestManifest` here instead of loading the resources and returning a `List<Map<String, Object>>` directly?

Obviously for production code (that has lots of tests on it) it might make sense to encapsulate this, but here it feels like the indirection again makes it a bit harder to see what's going on when trying to debug the tests.

german-muzquiz:
@ezimanyi I totally agree that tests need to make super obvious what's being done, and there's a balance between making everything explicit and not having test cases with 100+ lines. I'll revisit the helper classes and make some changes

german-muzquiz:
@ezimanyi I'm looking into the best way to provide request data. This is how it currently looks like:
```
String taskId =
         new DeployManifestRequest(baseUrl())
             .withAccount(ACCOUNT1_NAME)
             .withNamespace(overrideNamespace)
             .withSpinApplication(APP1_NAME)
             .withManifest(manifest)
             .executeAndValidate();
```
As you pointed out, this hides a few things in the request. Now this is without any helper class:
```
    List<Map<String, Object>> body =
        ImmutableList.of(
            ImmutableMap.of(
                "deployManifest",
                Stream.of(
                        new Object[][] {
                          {"cloudProvider", "kubernetes"},
                          {"enableTraffic", true},
                          {"account", ACCOUNT1_NAME},
                          {"namespaceOverride", overrideNamespace},
                          {"source", "text"},
                          {"skipExpressionEvaluation", false},
                          {"requiredArtifacts", new ArrayList<>()},
                          {"manifests", manifest},
                          {"moniker", ImmutableMap.of("app", APP1_NAME)}
                        })
                    .collect(Collectors.toMap(d -> (String) d[0], d -> d[1]))));
    Response resp =
        given().contentType("application/json").body(body).post(baseUrl() + "/kubernetes/ops");
    resp.then().log().ifValidationFails().statusCode(200);
    String taskId = resp.jsonPath().get("id");
```
It doesn't hide anything but triples the amount of lines used and I think it could be better.
This is yet another approach in which request body is loaded from a json file:
```
    List<Map<String, Object>> body =
        KubeTestUtils.loadJsonList("classpath:requests/deploy_manifest.json")
            .withValue("deployManifest.account", ACCOUNT1_NAME)
            .withValue("deployManifest.namespaceOverride", overrideNamespace)
            .withValue("deployManifest.moniker.app", APP1_NAME)
            .withValue("deployManifest.manifests", manifest)
            .asList();
    Response resp =
        given().contentType("application/json").body(body).post(baseUrl() + "/kubernetes/ops");
    resp.then().log().ifValidationFails().statusCode(200);
    String taskId = resp.jsonPath().get("id");
```
It's not as lengthy as the second option and the base full request is in a valid json file. What are your thoughts? Any other pattern we can use?

german-muzquiz:
I took the liberty to delete the request helper classes in favor of explicit request/response code.

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/containers/KubernetesCluster.java
```diff
@@ -0,0 +1,152 @@
...
+  private static final String DOCKER_IMAGE = "rancher/k3s:v1.17.11-k3s1";
+  private static final String KUBECFG_IN_CONTAINER = "/etc/rancher/k3s/k3s.yaml";
+
+  private static final Map<String, KubernetesCluster> instances = new HashMap<>();
+
+  private Path kubecfgPath;
+  private final String accountName;
+
+  public static KubernetesCluster getInstance(String accountName) {
+    if (instances.get(accountName) == null) {
```
ezimanyi, line 51: Really minor, feel free to ignore, but could be done as:
```
return instances.computeIfAbsent(accountName, KubernetesCluster::new);
```
which might be a tiny bit clearer.

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/containers/KubernetesCluster.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import org.springframework.util.FileCopyUtils;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.yaml.snakeyaml.Yaml;
+
+public class KubernetesCluster extends GenericContainer {
```
ezimanyi, line 40: This is throwing a compile warning:
```
warning: [rawtypes] found raw type: GenericContainer
public class KubernetesCluster extends GenericContainer {
                                       ^
  missing type arguments for generic class GenericContainer<SELF>
  where SELF is a type-variable:
    SELF extends GenericContainer<SELF> declared in class GenericContainer
```

I think the way this is supposed to be extended is by
```
public class KubernetesCluster extends GenericContainer<KubernetesCluster>
```

If that doesn't work we should at least `@SuppressWarnings("rawtypes")` but I think my suggestion should work and is how that class is expecting to be extended.

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/containers/KubernetesCluster.java
```diff
@@ -0,0 +1,152 @@
...
+    return myKubeconfig;
+  }
+
+  private void fixKubeEndpoint(Path kubecfgPath) throws IOException {
+    String kubeEndpoint = "https://" + getHost() + ":" + getMappedPort(6443);
+    Yaml yaml = new Yaml();
+    InputStream inputStream = Files.newInputStream(kubecfgPath);
+    Map<String, Object> obj = yaml.load(inputStream);
+    List<Map<String, Map<String, String>>> clusters =
+        (List<Map<String, Map<String, String>>>) obj.get("clusters");
```
ezimanyi, line 148: Agree that the unchecked cast here makes sense, but we should add a `@SuppressWarnings("unchecked")` to the method with a brief comment so that we don't get a warning.

You *could* do `KubeConfig obj = KubeConfig.loadKubeConfig(new FileReader(kubecfgPath.toFile()));` but it looks like the `KubeConfig` fields are just `Object` anyway so you'd still need to do the unchecked cast (and suppress it).

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/utils/KubeTestUtils.java
```diff
@@ -0,0 +1,147 @@
...
+      Iterable<Object> contentIterable = yaml.loadAll(is);
+      List<Map<String, Object>> content =
+          StreamSupport.stream(contentIterable.spliterator(), false)
+              .filter(Objects::nonNull)
+              .map(KubeTestUtils::coerceManifestToList)
+              .flatMap(Collection::stream)
+              .collect(Collectors.toList());
+      return new TestResourceFile(content);
+    } catch (IOException e) {
+      throw new RuntimeException("Unable to load manifest from file " + file, e);
```
ezimanyi, line 54: Not really important for a test, so feel free to ignore here...but often `UncheckedIOException` is the preferred way to wrap an `IOException` that we're just catching to avoid dealing with checked exceptions.

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/utils/KubeTestUtils.java
```diff
@@ -0,0 +1,147 @@
...
+    public List<Map<String, Object>> asList() {
+      return content;
+    }
+
+    public Map<String, Object> asMap() {
+      return content.get(0);
+    }
+
+    public TestResourceFile withValue(String path, Object value) {
+      String[] parts = path.split("\\.");
```
ezimanyi, line 120: This is throwing the warning:
```
warning: [StringSplitter] String.split(String) has surprising behavior
      String[] parts = path.split("\\.");
                                 ^
    (see https://errorprone.info/bugpattern/StringSplitter)
  Did you mean 'List<String> parts = Splitter.on('.').splitToList(path);'?
```

ezimanyi (approved):
This looks great, thanks for making those updates!

I took a final look through and just found a few minor one-line things that I've added inline.  The main thing is just that we've enabled more compiler warnings in the `clouddriver-kubernetes` module, so there are some warnings that should be fixed or suppressed.

Ideally this would no longer emit any warnings:
```
./gradlew -p clouddriver-kubernetes clean integrationClasses
```

I wish there were a better place to surface these warnings other than the compilation output...but this is where they are for now.  I think for the `unchecked` warnings given these are tests just adding `@SuppressWarnings("unchecked")` to the relevant methods is probably fine; the other warnings are probably worth fixing.

I added my approval now as I'll be away for ~1 week so feel free to merge after fixing/suppressing warnings; you don't need to wait for me to be back to look again.

PR #4831, opened 2020-08-24
fix(kubernetes): Do a live lookup in the artifact provider

ezimanyi:
* refactor(kubernetes): Update getArtifacts to accept kind

  As we'll be directly querying the cluster, it makes more sense to accept the actual kind rather than the artifact kind, which is "kubernetes/kind". For now we'll just re-construct the kind in the function, but that will shortly be unnecessary when we move to directly queryin the cluster.

* fix(kubernetes): Do a live lookup in the artifact provider

  This is only used during deploys, so latency is not a big issue here. This will fix race conditions with live manifest calls enabled where operations later in a pipeline don't account for operations that happened earlier in the pipeline.

  For now I am just replacing the exact same logic with a live call, including:
  * Building the whole artifact (even though we only need a few fields)
  * Sorting by creation timestamp instead of by version

  This unblocks deleting the code that caches artifacts, which then unblocks some futher simplifications.

* test(kubernetes): Move a test to Java

  As part of my goal to incrementally move tests to Java, if I break a test and need to fix it, I'm converting it to Java. The next commit breaks this test, so first convert it to Java.

* refactor(kubernetes): Stop caching artifacts

  Now that we always do a live lookup of artifacts, we can stop caching them, as no code reads artifacts from the cache.

  I've left the enum constant for artifact cache keys, and left the core caching agent authoritative, so that existing entries get evicted. We can remove this code after we've done a release with it present.

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesV2ArtifactProvider.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesV2ArtifactProvider.java
@@ -22,7 +22,9 @@ import static com.google.common.collect.ImmutableList.toImmutableList;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableList;
 import com.netflix.spinnaker.cats.cache.CacheData;
+import com.netflix.spinnaker.clouddriver.kubernetes.KubernetesCloudProvider;
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.Keys;
+import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesKind;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
 import java.util.Comparator;
 import javax.annotation.Nonnull;
@@ -42,7 +44,8 @@ public class KubernetesV2ArtifactProvider implements ArtifactProvider {
 
   @Override
   public ImmutableList<Artifact> getArtifacts(
-      String type, String name, String location, @Nonnull String account) {
+      KubernetesKind kind, String name, String location, @Nonnull String account) {
+    String type = String.join("/", KubernetesCloudProvider.ID, kind.toString());
     String key = Keys.ArtifactCacheKey.createKey(type, name, location, "*");
     return cacheUtils.getAllDataMatchingPattern(Keys.Kind.ARTIFACT.toString(), key).stream()
         .sorted(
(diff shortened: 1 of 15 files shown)


clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
```diff
@@ -160,7 +160,7 @@
   private static KubernetesV2ServerGroupManagerProvider serverGroupManagerProvider =
       new KubernetesV2ServerGroupManagerProvider(cacheUtils);
   private static KubernetesV2ArtifactProvider artifactProvider =
-      new KubernetesV2ArtifactProvider(cacheUtils, objectMapper);
+      new KubernetesV2ArtifactProvider(accountResolver);
```
ezimanyi, line 163: One of my favorite things in the past week is that this was the only change I needed to make to the integration test.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesV2ArtifactProvider.java
```diff
@@ -19,44 +19,40 @@
...
-    return cacheUtils.getAllDataMatchingPattern(Keys.Kind.ARTIFACT.toString(), key).stream()
-        .sorted(
-            Comparator.comparing(
-                cd -> (String) cd.getAttributes().getOrDefault("creationTimestamp", "")))
-        .map(this::cacheDataToArtifact)
-        .filter(a -> account.equals(a.getMetadata("account")))
+    return accountResolver
+        .getCredentials(account)
+        .map(credentials -> credentials.list(kind, location).stream())
+        .orElseGet(Stream::empty)
```
maggieneterval, line 50: If we can't find the credentials, do we want to just return an empty list or throw some kind of error?

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesDataProviderIntegrationTest.java
maggieneterval: 💯 💯 💯

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesV2ArtifactProvider.java
ezimanyi: Good question; I guess this is returning an empty list now.  (Also let me know if this mixing of Optional/Stream code is confusing...I'm not sure what i think of it.).

I picked this for consistency with the Manifest provider...but maybe we do want to throw an error, as presumably you can't reasonably pick a next version or decide what to delete if you don't have an account?

Though I also suspect that when I do a minor upcoming refactor I'll be able to pass in the credentials anyway so this might be moot.

maggieneterval: Yeah I was actually thinking of lifting the credentials resolution out of the ManifestProvider (into the controller, I guess) so it can fail as quickly as possible with an error appropriate to each endpoint, but we can leave this for now and see if we like that pattern any better 🤷‍♀️

PR #4848, opened 2020-08-28
refactor(kubernetes): Fix short circuit boolean error prone warning.

mattsanta:
* refactor(kubernetes): Fix short circuit boolean error prone warning. 

  The right hand side of the & has no side effect, so we can simply fix by switching to &&.


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesManifestProvider.java
```diff
@@ -125,7 +125,8 @@ public KubernetesV2Manifest getManifest(
                     .filter(
                         m ->
                             cluster.equals(KubernetesManifestAnnotater.getManifestCluster(m))
-                                & app.equals(KubernetesManifestAnnotater.getManifestApplication(m)))
+                                && app.equals(
```
maggieneterval, line 128: Wow this was just bad typing on my part, thanks for fixing!

PR #4852, opened 2020-08-28
refactor(kubernetes): add missing @Overrides to non-generated code

maggieneterval:
We cannot yet enable the error-prone warnings as there are a few dozen classes with Lombok-generated getter overrides that I'll handle in a followup PR.


mattsanta:
I personally prefer the annotation above the function, but from Google's java style guide it's ok for an annotation to be on the same line as the method signature if there is only one annotation and the annotation is parameterless.

maggieneterval:
That's so weird that the precommit spotlessApply didn't catch that, I wonder if it's because these are from the error prone generated patch?! Should be fixed now.

PR #4853, opened 2020-08-28
fix(kubernetes): Fix ImmutableEnumChecker and rawtypes warnings

ezimanyi:
* fix(kubernetes): Make enum fields final

  [ImmutableEnumChecker] enums should be immutable: 'PodPhase' has non-final field 'message'.

* fix(kubernetes): Fix use of raw types and enable warning

  I suppressed one warning because the interface requires using a raw type.

  Also the AccountCredentialsRepository interface returns raw types, which causes one issue where we try to directly use a lambda on the stream of results. I updated that to not use a lambda (which I think implicitly implies (AccountCredentials<?> c) -> c.getName() avoiding the warning, but we should also consider fixing AccountCredentials itself.

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/validator/KubernetesValidationUtil.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/validator/KubernetesValidationUtil.java
@@ -27,7 +27,6 @@ import com.netflix.spinnaker.clouddriver.kubernetes.security.KubernetesCredentia
 import com.netflix.spinnaker.clouddriver.security.AccountCredentials;
 import com.netflix.spinnaker.clouddriver.security.AccountCredentialsProvider;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import org.slf4j.Logger;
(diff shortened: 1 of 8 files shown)


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/validator/KubernetesValidationUtil.java
```diff
@@ -65,24 +64,6 @@ public boolean validateNotEmpty(String attribute, Object value) {
     return true;
   }
 
-  public boolean validateSizeEquals(String attribute, Collection items, int size) {
```
ezimanyi, line 68: This was unused, so I deleted it.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/validator/KubernetesValidationUtil.java
```diff
@@ -65,24 +64,6 @@ public boolean validateNotEmpty(String attribute, Object value) {
...
-    if (items.size() != size) {
-      reject("size!=" + size, attribute);
-      return false;
-    }
-
-    return true;
-  }
-
-  private boolean validateNotEmpty(String attribute, String value) {
```
ezimanyi, line 77: I just inlined this to the one place it was used (below)

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/validator/KubernetesValidationUtil.java
maggieneterval: 🥳

PR #4858, opened 2020-08-31
refactor(kubernetes): Split artifact creation from version determination

ezimanyi:
This PR is the first part of splitting artifact creation from version determination.  There are still a few improvements to be made (notably reading the existing versions from the moniker annotation instead of parsing the artifact version annotation) but this PR at least encapsulates the version determination into its own class (even if that class happens to look at the artifact annotations).

* refactor(kubernetes): Pull most ArtifactConverter logic to the base

  We have two implementations of ArtifactConverter that differ only in how they derive a version for the resource. Let's pull everything that doesn't involve determining a version up to the base, and leave the concrete classes to only handle versioning.

  Also remove converter.getDeployedName(artifact), which re-computes what we already stored in the artifact's reference field; just have callers look there.

* refactor(kubernetes): Delete tautological tests

  After the last refactor, some of the tests on the artifact converters are just testing Artifact itself; there's no need for us to ensure that creating an artifact with a reference returns that reference.

  We should replace these with better tests, but I'm going to wait until a later commit when logic is further consolidated.

* refactor(kubernetes): Return OptionalInt from getVersion

  The version parsing logic should return an OptionalInt instead of a String; this doesn't help us yet as we're just going to convert it to a String in getArtifact, but this will enable us in a later commit to directly access the int.

  Also, unrelatedly, move a few variable declarations closer to where they are used in KubernetesDeployManifestOperation.

* refactor(kubernetes): Split versioning and artifact creation

  Create a new ResourceVersioner class that is effectively the logic that was in KubernetesVersionedArtifactConverter (in fact the diff should even show it as a move). We'll make this a component so we don't need to always pass in the ArtifactProvider and can simplify the API. This class is responsible for taking a KubernetesManifest and telling you what version should be appended to its name.

  Have KubernetesArtifactConverter require a version as input; this gets it out of the business of figuring out what version should be applied and with the core competency of just creating an artifact. It now just has the single toArtifact function, which we'll make static and remove the (unecessary) implementation classes.

  The remaining commits are just changing from passing around an ArtifactProvider to passing around a ResourceVersioner.

* test(kubernetes): Add some tests on ArtifactConverter

  First rename KubernetesArtifactConverter to ArtifactConverter as it already has kuberetes in the package name. (Maybe I've been writing too much go.)

  Second add some tests on this class.

  Finally, add a few missing test cases to improve the coverage on the ResourceVersioner (mainly testing what happens when versions don't parse correctly).


dbyron0:
Another fabulous PR no doubt :)  Any chance it paves the way for a fix to https://github.com/spinnaker/spinnaker/issues/4442 ?

ezimanyi:
@dbyron0 : Short answer is no, this PR won't help but there is other work going on that might; I've added a comment on the issue.

dbyron0:
Thank you @ezimanyi !

PR #4874, opened 2020-09-08
fix(kubernetes/artifacts): Update Clouddriver's artifact replacement logic to replace docker images even if the image specified in the manifest includes a tag.

mattsanta:
* fix(kubernetes/artifacts): Update Clouddriver's artifact replacement logic to replace docker images even if the image specified in the manifest includes a tag. 

  Previous implementation didn't allow overwriting a image in a manifest with a bound artifact if the input manifest had a tag on it. This was problematic when using helm since most helm charts have a tag.

  BREAKING CHANGE: Can no longer rely on Spinnaker ignoring any matching image artifacts available in the pipeline and using the tag specified in the manifest image.

* Use nameFromReference to extract the artifact name in the input manifest rather than using regex. 

* Use JsonPath in replace filter predicate. 

* Change artifact replacement in KubernetesDeployManifestOperation from replacing required then optional artifacts to replacing optional then required artifacts. This change is being done to prevent optional artifacts from replacing required artifacts.

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/Replacer.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/Replacer.java
@@ -23,7 +23,6 @@ import static com.jayway.jsonpath.Filter.filter;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.google.common.collect.ImmutableCollection;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Streams;
 import com.jayway.jsonpath.DocumentContext;
@@ -31,12 +30,11 @@ import com.jayway.jsonpath.Filter;
 import com.jayway.jsonpath.JsonPath;
 import com.jayway.jsonpath.PathNotFoundException;
 import com.jayway.jsonpath.Predicate;
+import com.jayway.jsonpath.internal.filter.ValueNode;
 import com.netflix.spinnaker.clouddriver.artifacts.kubernetes.KubernetesArtifactType;
 import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
 import java.util.Collection;
-import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Function;
@@ -62,7 +60,7 @@ public final class Replacer {
    *     representing a filter
    * @param findFilter a filter that should be applied to the path when finding any artifacts in a
    *     manifest; defaults to a filter matching all nodes
-   * @param replacePathFromPlaceholder a list that represents the path from the [?] placeholder to
+   * @param replacePathFromPlaceholder a string that represents the path from the [?] placeholder to
    *     the replaced field.
    * @param nameFromReference a function to extract an artifact name from its reference; defaults to
    *     returning the reference
@@ -72,7 +70,7 @@ public final class Replacer {
       KubernetesArtifactType type,
       String path,
       @Nullable Filter findFilter,
-      List<String> replacePathFromPlaceholder,
+      String replacePathFromPlaceholder,
       @Nullable Function<String, String> nameFromReference) {
     this.type = Objects.requireNonNull(type);
     Objects.requireNonNull(path);
@@ -114,22 +112,11 @@ public final class Replacer {
     return replacedArtifacts.build();
   }
 
-  private Predicate createReplaceFilterPredicate(List<String> replacePath, String name) {
+  private Predicate createReplaceFilterPredicate(String replacePath, String name) {
     return ctx -> {
-      Map<String, Object> ctxMap = ctx.item(Map.class);
-      String replace = "";
-      for (String field : replacePath) {
-        if (!ctxMap.containsKey(field)) {
-          return false;
-        }
-
-        if (ctxMap.get(field) instanceof String) {
-          replace = (String) ctxMap.get(field);
-        } else {
-          ctxMap = (Map<String, Object>) ctxMap.get(field);
-        }
-      }
-      return nameFromReference.apply(replace).equals(name);
+      ValueNode node = ValueNode.toValueNode("@." + replacePath).asPathNode().evaluate(ctx);
+      String value = node.asStringNode().getString();
+      return nameFromReference.apply(value).equals(name);
     };
   }
 
@@ -160,7 +147,7 @@ public final class Replacer {
   private static final Replacer DOCKER_IMAGE =
       builder()
           .path("$..spec.template.spec['containers', 'initContainers'].[?].image")
-          .replacePathFromPlaceholder(ImmutableList.of("image"))
+          .replacePathFromPlaceholder("image")
           .nameFromReference(
               ref -> {
                 // @ can only show up in image references denoting a digest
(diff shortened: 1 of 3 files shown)


mattsanta:
Fix for https://github.com/spinnaker/spinnaker/issues/5948

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/Replacer.java
```diff
@@ -136,7 +137,7 @@ private boolean replaceIfPossible(DocumentContext obj, Artifact artifact) {
   private static final Replacer DOCKER_IMAGE =
       builder()
           .path("$..spec.template.spec['containers', 'initContainers'].[?].image")
-          .replaceFilter(a -> filter(where("image").is(a.getName())))
+          .replaceFilter(a -> filter(where("image").regex(Pattern.compile(a.getName() + "(:.*)?"))))
```
maggieneterval, line 140: Is there a way for us to use `nameFromReference` to extract the name excluding the colon + tag sequence, or do we need to use a regex to match?

maggieneterval (commented):
☁️ 🚗 🏺 ♻️ 🐳 🏷️ 💯 
Thanks so much for doing this, @mattsanta!

maggieneterval:
> BREAKING CHANGE: Can no longer rely on Spinnaker ignoring any matching image artifacts available in the pipeline and using the tag specified in the manifest image.

👍 Although as we discussed in the Kubernetes SIG we don't think anyone is intentionally relying on the existing behavior, let's still document it in the [1.23 changelog](https://spinnaker.io/community/releases/next-release-preview/) as a breaking change.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/Replacer.java
mattsanta: Regex match was the only way I could see for us to find the image, with or without tag, in the manifest given the name of the image that is replacing.

mattsanta:
I'll put up another PR to update the 1.23 changelog and mention this breaking change.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/Replacer.java
maggieneterval: Okay! Just felt it was worth calling out as it feels a little strange at first glance to use a regex in `replaceFilter` when we have kind of similar logic in the `nameFromReference` method to extract just the artifact name from name + optional tag that doesn't use a regex, but I trust your judgment 👍 

One weird edge case that came to mind is an image that contains both a colon-delimited tag and a colon-delimited port (in which case we want to replace), or a colon-delimited digest and a colon-delimited port (in which case we do not want to replace); might be worth adding test coverage if we don't have already.

maggieneterval (approved):

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/Replacer.java
mattsanta: Please feel free to correct me if I'm wrong though! From my understanding, ```nameFromReference``` is used to get the name of an image in a manifest, it's not useful here because we already know the image name that we want to replace from the artifact that is replacing (```a.getName()```). Instead we're looking to find an image in a manifest that has the same name as the artifact that is replacing, but with the caveat that the image in the manifest may have a tag on it.

With respect to the edge case, could you provide some more context or examples? I'm not familiar with the image formatting.

ezimanyi: +1 to Maggie's comment, doing this with a regex seems pretty fragile to me.  It also looks like this regex is just re-implementing `nameFromReference` though actually slightly differently (as `nameFromReference` handles images with digests not just tags).

At the very least we need much more robust tests (similar to [these](https://github.com/spinnaker/clouddriver/blob/faf0dfee31ae50b2e76808fa89cad068f1825448/clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/ArtifactReplacerTest.java#L93)) as the single test case we have now is not sufficient to handle the added complexity of a regex here.

But ultimately I'm confident there's a way to lightly refactor this so that we can re-use `nameFromReference` instead of needing to re-implement it.  Obviously this isn't possible directly at the call site because we're just building an object.

But what if `replaceFilter` were actually be a `Function<String, JsonPath>` as it looks like we always only pass in the `artifact.getName()`.  Then we could call `nameFromReference` on `artifact.getName()` before passing it in.  I think that (or something similar) would work; while it's a tiny bit more code it would avoid having the existing logic re-implemented as a regex.

ezimanyi: I wrote my comment before seeing your comment...the tests I linked should have good examples of the image formatting.  What `nameFromReference` does is remove any digest/tag from an image, so we just have the actual image name.  It's used when reading artifacts from a manifest because we want the name to not have the tag (while the reference does).

But it could also be used when replacing because the manifest contains a full reference (though maybe it doesn't have a tag).  We effectively want to take that full reference in the manifest and figure out what the name is, then replace based on that name.  The regex is one solution (by ignoring anything after `:`) but `nameFromReference` is a well-tested and more robust way of doing that; if the manifest lists `gcr.io/my-project/my-image:tag` then `nameFromReference` will give us back `gcr.io/my-project/my-image` which is exactly what we want.

mattsanta: > But what if replaceFilter were actually be a Function<String, JsonPath> as it looks like we always only pass in the artifact.getName(). Then we could call nameFromReference on artifact.getName() before passing it in

If my understanding is correct, we would want to  call ```nameFromReference``` on what's in the manifest image path not on ```artifact.getName()``` because ```artifact``` is the replacement and we already know the name, it's what ```artifact.getName()``` returns.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/Replacer.java
```diff
@@ -109,6 +114,25 @@ private Replacer(
     return replacedArtifacts.build();
   }
 
+  private Predicate createReplaceFilterPredicate(List<String> replacePath, String name) {
+    return ctx -> {
+      Map<String, Object> ctxMap = ctx.item(Map.class);
```
ezimanyi, line 119: I think here we can have JSONPath do this lookup for us so we don't need to handle looking up the various keys at each level.

Something like this should directly get you the value at the input replace path (where the `@.` that's prepended is JSONPath for "relative to the current node"):
```
String path = String.join(".", replacePath);
ValueNode node = ValueNode.toValueNode("@." + path).asPathNode().evaluate(ctx);

// Then you can check if it's a String with
node.isStringNode()
// And get the value with
String value = node.asStringNode().getString()
```
(Given we're re-joining the `List<String> replacePath` it might make sense if you go with the above to leave it as it was initially, as a dot-separated JsonPath query.)

The main reason I suggest this is just that given we're invested in JsonPath in this class I think we should try to encapsulate as much as possible behind it; by doing this we can avoid some manual iteration and unchecked casts that the library handles.

But overall I like this solution a lot better than using a regex as re-using `nameFromReference` will keep behavior consistent with any future changes. 💯

mattsanta: Thanks for the response. Your suggestion is much cleaner, I've went ahead and updated the PR.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/Replacer.java
```diff
@@ -109,6 +112,14 @@ private Replacer(
     return replacedArtifacts.build();
   }
 
+  private Predicate createReplaceFilterPredicate(String replacePath, String name) {
+    return ctx -> {
+      ValueNode node = ValueNode.toValueNode("@." + replacePath).asPathNode().evaluate(ctx);
+      String value = node.asStringNode().getString();
```
ezimanyi, line 118: We should check `node.isStringNode()` before calling `node.asStringNode()` to avoid propagating an exception.  (I think in that case we'd just want to return `false` early as the node didn't have what we expected.)

mattsanta: Updated!

ezimanyi: You're checking `node.isStringNode()` after already calling `node.asStringNode()` by which time the exception will already have happened---you'll need to move the `node.isStringNode()` to happen before we call `node.asStringNode()`.

mattsanta: Sorry about that, fixed now.

ezimanyi (approved):
Looks great!

I wish the logic for whether to prioritize required or optional artifacts was a bit more explicit, but that was not introduced in this PR so don't think we should expand the scope here to make changes there.

💯 ☁️ 🚗 🏺 ✅

PR #4877, opened 2020-09-08
fix(aws): Fix startup when SQL isn't enabled

plumpy:
Reverts spinnaker/clouddriver#4710

This breaks clouddriver startup if AWS is enabled but not SQL.


jeyrschabu (approved):
Will reintroduce later once I figure out what’s happening.

plumpy:
Thanks for the quick review!

PR #4879, opened 2020-09-09
fix(provider/gce): Fix failed to deploy when backendservice with no b…

takaaki7:
spinnaker/spinnaker#6028


plumpy:
What if you changed `GoogleBackendService` so the `backends` field has a default value:
```
  List<GoogleLoadBalancedBackend> backends = ImmutableList.of()
```

Does that work? I think in general removing the possibility of a `null` everywhere feels like a more robust change. Then you could also remove a few places where you check that value for `null` (like `UpsertGoogleInternalHttpLoadBalancerAtomicOperation:944`).

takaaki7:
It does work in my environment. It seems there is no mutations for the list.
And I added null check.

plumpy:
Sorry, I think I didn't do a good job explaining. What I meant is to make sure that `GoogleBackendService#backends` is _never_ `null`. So initialize the value to be an empty list inside `GoogleBackendService`. Then you can remove all the `null` checks everywhere and remove the `else` block that you added.

Does that make sense? So basically the only real change you make should be in `GoogleBackendService.groovy`. And for bonus, you could remove the various null checks since it shouldn't ever be null anymore.

takaaki7:
Ah, I got it. I'll fix.

takaaki7:
I've fixed

plumpy:
Great, thanks!

plumpy (approved):

PR #4882, opened 2020-09-09
fix(cache): OnDemandCacheStatus no longer an inner enum

german-muzquiz:
`OnDemandCacheStatus` was recently moved to its own class instead of being an inner enum of `OnDemandCacheUpdater`.

@robzienert this was caught by the soon-to-be-merged integration tests from https://github.com/spinnaker/clouddriver/pull/4827, so once those are merged we can catch these before reaching `master` :)


german-muzquiz:
It seems there's a duplicate PR to fix this https://github.com/spinnaker/clouddriver/pull/4881

PR #4884, opened 2020-09-10
feat(aws): Adding support for update asg launch template

jeyrschabu:
- modify server group launch template saga support
- this operation will be used by rollingpush and the editing of asg's security groups

### DIFF
--- a/clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/config/EventConfiguration.kt
+++ b/clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/config/EventConfiguration.kt
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


jeyrschabu:
Reintroduce https://github.com/spinnaker/clouddriver/commit/b5efd0d90ff25eb153caf9e849ae7203f94ba1d3 with a fix to the issue that led to the revert.

When sql.enabled=false and aws is the provider.

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/config/EventConfiguration.kt
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.config
+
+import org.springframework.context.annotation.ComponentScan
+import org.springframework.context.annotation.Configuration
+
+@Configuration
+@ComponentScan(basePackages = ["com.netflix.spinnaker.clouddriver.event"])
```
jeyrschabu, line 23: The in-memory-event repository was previously not seen by Spring. This fixes that.

jeyrschabu:
Can I get a thumbs up on this? Would like to test this out this am.

caseyhebebrand (approved):
The launch template logic looks good to me, but I think someone who is more familiar with sagas should also take a look.

clouddriver-event/src/main/kotlin/com/netflix/spinnaker/clouddriver/config/EventConfiguration.kt
jeyrschabu: FYI @robzienert

jeyrschabu:
Saga parts are the same as last PR with the only addition being the fix for spring to be able to find the in memory event repo.

Merging for now but feel free to comment on things and I will address them in separate PRs.

PR #4887, opened 2020-09-14
feat(test): Add initial ECS integration tests

allisaurus:
Adds new `integration` sourceSet to `clouddriver-ecs`, which contains setup and initial test cases for Amazon ECS provider integration tests. Package structure builds off precedents set in https://github.com/spinnaker/clouddriver/pull/4827. 

Intention is to have these run during the ["Integration Tests"](https://github.com/spinnaker/clouddriver/blob/master/.github/workflows/pr.yml#L29) step of the PR approval workflow.

### DIFF
--- a/clouddriver-ecs/clouddriver-ecs.gradle
+++ b/clouddriver-ecs/clouddriver-ecs.gradle
@@ -1,3 +1,17 @@
+sourceSets {
+  integration {
+    java.srcDirs = ["src/integration/java"]
+    resources.srcDirs = ["src/integration/resources"]
+    compileClasspath += main.output + test.output
+    runtimeClasspath += main.output + test.output
+  }
+}
+
+configurations {
+  integrationImplementation.extendsFrom testImplementation
+  integrationRuntime.extendsFrom testRuntime
+}
+
 dependencies {
   implementation project(":cats:cats-core")
   implementation project(":clouddriver-api")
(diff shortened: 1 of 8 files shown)


allisaurus:
FYI `PR Build / build (pull_request)` check failure anticipated to be fixed by https://github.com/spinnaker/clouddriver/pull/4889

Run of ECS Integ tests in isolation succeeds here: https://github.com/allisaurus/clouddriver/runs/1114862660

clouddriver-ecs/clouddriver-ecs.gradle
```diff
@@ -1,3 +1,17 @@
+sourceSets {
+  integration {
+    java.srcDirs = ["src/integration/java"]
+    resources.srcDirs = ["src/integration/resources"]
+    compileClasspath += main.output + test.output
```
pkandasamy91, line 5: Out of curiosity, what happens if you don't include main.output+test.output on the compileClasspath?

clouddriver-ecs/clouddriver-ecs.gradle
```diff
@@ -37,4 +51,24 @@ dependencies {
...
+  integrationImplementation "org.testcontainers:mysql"
+  integrationImplementation "mysql:mysql-connector-java"
+  integrationImplementation ("io.rest-assured:rest-assured:4.0.0") {
+    exclude group: "org.codehaus.groovy", module: "groovy" // use kork's version
+  }
+}
+
+task integrationTest(type: Test) {
+  description = "Runs Amazon ECS provider integration tests."
+  group = 'verification'
```
pkandasamy91, line 66: What other groups are there? Can we make this an enum?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/EcsSpec.java
```diff
@@ -0,0 +1,107 @@
...
+
+  @DisplayName(".\n===\n" + "Assert AWS and ECS providers are enabled" + "\n===")
+  @Test
+  public void configTest() {
+    assertTrue(awsEnabled);
+    assertTrue(ecsEnabled);
+  }
+
+  protected String generateStringFromTestFile(String path) throws IOException {
+
```
pkandasamy91, line 89: nit: extra newline

pkandasamy91 (approved):
Ship it! Had a couple minor nits but looks good overall!

clouddriver-ecs/clouddriver-ecs.gradle
allisaurus: AFAICT the only other places task groups are mentioned in `clouddriver` are in `build.gradle` and `clouddriver-kubernetes.gradle`, and this usage exactly matches the latter (also an integrationTest task). The [gradle docs](https://guides.gradle.org/writing-gradle-tasks/#add_a_task_description) also show using a string to specify task group, so I think this is the way to go?

clouddriver-ecs/clouddriver-ecs.gradle
allisaurus: so w/o the "compileClasspath" line the integ tests won't build, but I tried it w/o the "runtimeClasspath" and it seems to build/run fine! will remove that one (iirc I originally added it based on reference docs but the specific usage must have been different).

PR #4889, opened 2020-09-14
fix(test): Changed cache refreshed check, precreated ns

german-muzquiz:
There have been some one-off failures in kubernetes integration tests: https://github.com/spinnaker/clouddriver/runs/1103123724?check_suite_focus=true
The above job was successful after a re-run.

Failure message:
```
===
Given a secret manifest
When sending deploy manifest request
  And sending force cache refresh request
  And waiting on manifest stable
Then secret is deployed with a version suffix name
=== FAILED
    org.opentest4j.AssertionFailedError: GET /cache/kubernetes/manifest did not returned processedTime > -1 for object "secret mysecret-v000" after 5 minutes
```

This PR introduces these changes:
* List filter code for the above request was rewritten from rest-assured jsonpath to plain java code
* More logging of relevant response fragments
* Test namespaces are created on startup, rather than on each test case. This allows the namespaces to be already available in the cache when the test cases start, as oposed of doing deployments to not-yet-cached namespaces
* A gradle plugin is added for printing better test reports: this allows to print stdout and stderr only for the failed tests instead of all tests, when run without the `--info` flag

### DIFF
--- a/clouddriver-kubernetes/clouddriver-kubernetes.gradle
+++ b/clouddriver-kubernetes/clouddriver-kubernetes.gradle
@@ -1,5 +1,6 @@
 plugins {
   id("net.ltgt.errorprone") version "1.2.1"
+  id 'com.adarshr.test-logger' version '2.1.0'
 }
 
 tasks.compileGroovy.enabled = false
@@ -116,4 +117,12 @@ task integrationTest(type: Test) {
   testClassesDirs = sourceSets.integration.output.classesDirs
   classpath = sourceSets.integration.runtimeClasspath
   shouldRunAfter test
+
+  testlogger {
+    theme 'standard'
+    showStandardStreams true
+    showPassedStandardStreams false
+    showFailedStandardStreams true
+    showPassed false
+  }
 }
--- a/clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeployManifestIT.java
+++ b/clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeployManifestIT.java
@@ -17,8 +17,7 @@
 package com.netflix.spinnaker.clouddriver.kubernetes.it;
 
 import static io.restassured.RestAssured.given;
-import static org.hamcrest.Matchers.anyOf;
-import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.*;
 import static org.junit.jupiter.api.Assertions.*;
 
 import com.google.common.collect.ImmutableMap;
@@ -26,11 +25,10 @@ import com.netflix.spinnaker.clouddriver.kubernetes.it.utils.KubeTestUtils;
 import io.restassured.path.json.JsonPath;
 import io.restassured.response.Response;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.apache.logging.log4j.util.Strings;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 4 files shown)


clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeployManifestIT.java
```diff
@@ -826,17 +824,24 @@ private void deployAndWaitStable(
...
-                  jsonPath.get(
-                      "findAll { it -> it.details.account == \""
-                          + ACCOUNT1_NAME
-                          + "\" && it.details.location == \""
-                          + targetNs
-                          + "\" && it.details.name == \""
-                          + objectName
-                          + "\" && it.processedTime > -1 }");
-              return list != null && !list.isEmpty();
+                  Stream.of(fcrWaitResp.as(Map[].class))
```
maggieneterval, line 829: > List filter code for the above request was rewritten from rest-assured jsonpath to plain java code

Was this part of the fix or just rewritten to improve debuggability?

clouddriver-kubernetes/src/integration/resources/clouddriver.yml
```diff
@@ -28,6 +29,7 @@ kubernetes:
       metrics: false
       kubeconfigFile: ${PROJECT_ROOT}/clouddriver-kubernetes/build/kubeconfigs/kubecfg-${account1_containername}.yml  # File is automatically created during initialization
     - name: account2
+      cacheIntervalSeconds: 30
```
maggieneterval, line 32: What was the rationale behind having different `cacheIntervalSeconds` for account1 vs account2?

maggieneterval (commented):
> Test namespaces are created on startup, rather than on each test case. This allows the namespaces to be already available in the cache when the test cases start, as oposed of doing deployments to not-yet-cached namespaces

Just making sure I understand -- the (intermittent) absence of a test namespace from the cache was the only cause of the flakiness, and the other changes in this PR were just to improve debuggability, right?

clouddriver-kubernetes/clouddriver-kubernetes.gradle
```diff
@@ -116,4 +117,12 @@ task integrationTest(type: Test) {
...
   classpath = sourceSets.integration.runtimeClasspath
   shouldRunAfter test
+
+  testlogger {
+    theme 'standard'
+    showStandardStreams true
+    showPassedStandardStreams false
+    showFailedStandardStreams true
+    showPassed false
+  }
```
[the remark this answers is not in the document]
german-muzquiz: Yes they are the defaults. I just wanted to explicitly define flags interesting for these tests, although now that I see it, the theme is not that interesting :)

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeployManifestIT.java
german-muzquiz: This is the actual change that hopefully prevents intermittent failures. For some reason, when filtering the on demand cache results looking for the desired object using rest-assured json path, the object was not always returned by the filter even if it was present in the response. I assume it might be related to the `processedTime` comparison having to be treated as a `long`.

So I changed the filter to use `Object.equals` for the other fields and treating `processedTime` as a java `Number` to guarantee number comparisons in all cases. On my debugging, if `processedTime` was `-1` it was returned as `Integer` and if it had an actual time it was a `Long`.

clouddriver-kubernetes/src/integration/resources/clouddriver.yml
german-muzquiz: I reduced the interval for `account1` just so that it runs the cache refresh more frequently to speed up tests. `account2` is actually not directly used, it's just there so that there is more than one account present, but it could be deleted and added later when there are tests actually needing it.

german-muzquiz:
> Just making sure I understand -- the (intermittent) absence of a test namespace from the cache was the only cause of the flakiness, and the other changes in this PR were just to improve debuggability, right?

The absence of test namespaces from the cache is not related to the flakiness we saw, because in those cases the force cache refresh returns `200 OK` (no agent handled the request) instead of `202`, and therefore there's no pending OnDemand cache entries to wait for. It could cause problems though on `getManifest` request if we're querying a namespace that doesn't exist yet in the cache, I didn't see any errors for that but it's better to make sure that namespaces are cached.

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeployManifestIT.java
maggieneterval: >  On my debugging, if processedTime was -1 it was returned as Integer and if it had an actual time it was a Long.

Oh, so strange! Thanks for the explanation.

PR #4891, opened 2020-09-15
fix(aws): Fetch launch template versions at once

jeyrschabu:
- added proxy method for describe launch templates
- updated caching agent to fetch versions in one call
- TODO(not supported): Add proxy method for describe versions


jeyrschabu:
It looks like `describeLaunchTemplateVersion` isn't quite there, I will follow up with Edda. Instead I updated to use the new API using the versions in the request. This allows to fetch all items in a single call rather than per launch template.

Will circle back to add the proxy for `describeLaunchTemplateVersion` but this should greatly reduce rate limit concerns.

jeyrschabu:
will fix test later today I guess.

robzienert (approved):

caseyhebebrand (approved):

ajordens (approved):

PR #4892, opened 2020-09-15
fix (aws/lambda):  patch broken translation of uuid fields by object mapper

shyamsfo:
fix (aws/lambda):  patch broken translation of uuid fields by object mapper


kevinawoo (approved):
LGTM!

kevinawoo (approved):
LGTM to as a workaround the library.

PR #4898, opened 2020-09-16
feat(test/int): Run int tests only on pushes to master

german-muzquiz:
There are two concerns when running integration tests to block PR builds:
1. Builds take a long time to complete. With the few tests added so far in clouddriver, build times were increased by about 13 minutes. This is only going to take longer as more tests are added to the service.
2. Test flakiness. Given the inherent complexity of integration tests, there could be sometimes when they fail in a particular execution without any code change.

The goal of integration tests is to find bugs before doing releases, the sooner the better. A good compromise would be to run them on pushes to `master` branch, so that PR builds are not delayed or blocked by the tests, but yet there's a clear indication of exactly which PR broke the tests, which is not possible when they are run on a scheduled basis like the current end to end tests.

This PR introduces a new workflow for running integration tests only on pushes to `master` branch. It makes use of gradle's [build cache](https://docs.gradle.org/current/userguide/build_cache.html) and github's [cache action](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows), so that only gradle tasks that were affected by a commit are executed. This allows to save a lot of time and unnecessary test runs on changes that could not affect test output. As an example, build time for changing a workflow file goes down from ~23 minutes to ~3 minutes, because gradle doesn't detect any changes and marks all tasks as up-to-date.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -1,0 +1,33 @@
+name: Integration Tests
+
+on:
+  push:
+    branches:
+      - master
+
+env:
+  GRADLE_OPTS: -Dorg.gradle.daemon=false -Xmx6g -Xms6g
+
+jobs:
+  it-test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    # Install Java 8 for cross-compilation support. Setting it up before
+    # Java 11 means it comes later in $PATH (because of how setup-java works)
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 8
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 11
+    - uses: actions/cache@v2
+      with:
+        path: ~/.gradle
+        key: ${{ runner.os }}-cd-build-${{ github.sha }}
+        # Restore build outputs from the previous commit (if successful), if current commit hasn't run successfully yet
+        restore-keys: |
+          ${{ runner.os }}-cd-build-${{ github.event.before }}
+    # Separating integration tests by provider allows to have separate logs
+    - name: Kubernetes Provider Integration Tests
+      run: ./gradlew --build-cache :clouddriver-kubernetes:integrationTest
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+  pull_request:
 
 env:
   GRADLE_OPTS: -Dorg.gradle.daemon=false -Xmx6g -Xms6g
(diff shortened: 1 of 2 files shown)


german-muzquiz:
cc @allisaurus @maggieneterval @jonsie

maggieneterval:
> A good compromise would be to run them on pushes to master branch, so that PR builds are not delayed or blocked by the tests, but yet there's a clear indication of exactly which PR broke the tests

I think this is a reasonable compromise, although I don't love that in the case where regressions are introduced, the burden of identifying them is essentially being pushed from the PR author to whoever is on build rotation (no worse than the current end-to-end tests, though, and as you mention, better because we can at least identify the exact PR). I don't have a better suggestion in mind, but might be worth calling out this decision in the #build-process channel as I'm not sure build rotation members are very actively monitoring the build statuses page in addition to the Jenkins server.

> It makes use of gradle's build cache and github's cache action, so that only gradle tasks that were affected by a commit are executed. 

Is it feasible use these on PR build as well?

german-muzquiz:
> Is it feasible use these on PR build as well?

Yes! cache can also be used in PR builds to compile only the code that was changed instead of the whole project.

I'm thinking that running integration tests can also be done as part of a PR status checks, just a separate one that doesn't block the merge. Then the approver could see if it's ok to merge, wait, etc. I'll ask feedback in #build-process

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,33 @@
...
+      with:
+        java-version: 11
+    - uses: actions/cache@v2
+      with:
+        path: ~/.gradle
+        key: ${{ runner.os }}-cd-build-${{ github.sha }}
+        # Restore build outputs from the previous commit (if successful), if current commit hasn't run successfully yet
+        restore-keys: |
+          ${{ runner.os }}-cd-build-${{ github.event.before }}
+    # Separating integration tests by provider allows to have separate logs
```
allisaurus, line 31: 🔥

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,33 @@
+name: Integration Tests
+
+on:
+  push:
+    branches:
+      - master
```
jonsie, line 6: I think we possibly could run on this on PR so long as it's not a build the blocks merge.

maggieneterval: In addition to running on PR and master builds, should we run on release branch builds as well?

german-muzquiz: I'm fine running or not running in release branch builds.
As long as every commit to release branches is done through PRs, the tests will run in those PRs checks before the merge. 
Moreover, not running in release branches will make that the full test suite is executed at least once in each PR regardless of any code change, because there will be no cache to inherit from. 

In the case of `master` its PRs will inherit the cache that was saved in `master` runs, at least only on the first commit in the PR. The logic is literally "get the cache for this commit sha, if not found, get the cache from the previous commit sha (if exists)".

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Cache on push
+      if: github.event_name == 'push'
+      uses: actions/cache@v2
+      with:
+        path: ~/.gradle
+        key: ${{ runner.os }}-cd-it-${{ github.sha }}
+        # Restore build outputs from the previous commit (if successful), if current commit hasn't run successfully yet
+        restore-keys: |
+          ${{ runner.os }}-cd-it-${{ github.event.before }}
+    - name: Cache on pull_request
```
allisaurus, line 34: How are we specifying that this check is non-blocking to PRs? If that has to be done later/separately, that's cool, just wondering if I missed something.

allisaurus (approved):
Had a question for my own understanding, but I am down with 1) running this as a status blocker on merges to master 2) running this as a non-blocking check on PRs. Thanks for drafting this!

Approving on the condition other folks with opinions on build/PR CI are also cool with things.

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,44 @@
+name: Integration Tests
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
allisaurus, line 7: How are we specifying that this check is non-blocking to PRs? If that has to be done later/separately, that's cool, just wondering if I missed something.

german-muzquiz: PR status checks can be configured on the branch rules of the repo, under `Settings` page, there you can indicate which ones are required. For this repo you can see it in all PRs:

image

Also I see the following relevant rules of Mergify bot:

image

I guess those refer to the workflow job name, which is `build` for the PR build workflow and `it-test` for the integration tests workflow.

maggieneterval (approved):

jonsie (approved):

PR #4901, opened 2020-09-17
fix(kubernetes): Support for unregistered owners

ncknt:
When an unregistered owner is discovered, its kind is added directly as a relationship - ignoring its potential api group. Since we store unregistered owner with their group, the relationship is lost. This change ensures the api group is considered for owner references. It's basically the same computation as on the manifest itself.

The value is not memoized because as far as I can tell it's called only once.


german-muzquiz (approved):
Looks good to me

PR #4902, opened 2020-09-17
fix(servergroups): Add fields server group instanceType and instance availabilityZone fields

jonsie:
This way, when a request for `expanded` results comes in it includes fields that we also return on non-expanded requests.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonInstance.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonInstance.groovy
@@ -88,6 +88,10 @@ class AmazonInstance implements Instance, Serializable {
     getExtraAttributes().get("placement")?.availabilityZone
   }
 
+  String getAvailabilityZone() {
+    return this.getZone()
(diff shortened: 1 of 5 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonInstance.groovy
```diff
@@ -88,6 +88,10 @@ class AmazonInstance implements Instance, Serializable {
     getExtraAttributes().get("placement")?.availabilityZone
   }
 
+  String getAvailabilityZone() {
```
marchello2000, line 91: this just gets included as part of `ServertGroup`?

jonsie: No, this was missing from the `instance` object when expanded vs. not expanded because the non expanded returns a `InstanceViewModel` which maps `zone` -> `availabilityZone` whereas expanded just returns the entire `instance` object. 🤷‍♂️

marchello2000 (approved):
thank you!

PR #4903, opened 2020-09-17
refactor(kubernetes): Remove KubernetesOnDemandCachingAgent

ezimanyi:
* refactor(kubernetes): Remove force cache refresh

  As all operations during a pipeline execution are doing live lookups, we no longer need to support force cache refresh (or on-demand cache updates at all) in the Kubernetes provider.

  This commit updates the force-cache refresh to be a no-op; an upcoming commit will make the caching agent completely stop implementing the on-demand interface.

* refactor(kubernetes): Kubernetes caching agents don't implment on-demand

  Now that we don't do on-demand cache refreshes at all, we don't need to implement this interface.

* refactor(kubernetes): Delete on-demand logic from caching agent

  While the last commit made KubernetesOnDemandCachingAgent no longer implment OnDemandCachingAgent, there is still logic that looks for on-demand entries and merges them in when loading data. Remove this logic as we'll no longer be creating said entries.

* refactor(kubernetes): Remove KubernetesOnDemandCachingAgent

  This agent no longer adds any value; classes that extend it can just directly extend KubernetesCachingAgent.


maggieneterval (approved):
🚫 ☁️ 💰 💵 🕵️ 💯 🥳 🎉 🎈 🤗

PR #4904, opened 2020-09-17
refactor(kubernetes): Remove liveManifestCalls flag

ezimanyi:
The performance improvements of the past few months have ultimately led to this flag no longer being read anywhere. Spinnaker now decides whether to read from the cache or to make a live call based on the context of the call rather than based on a flag.

From a practical perspective this means that:
* Users who had liveManifestCalls enabled will see the same fast deploys as always; bugs and race conditions with dynamic target selection have now been fixed.
* Users who had liveManifestCalls disabled will notice significantly faster deployments.


dbyron0:
Yay! 🎉 🎉

maggieneterval (approved):
👋 🎸 🎤 📜 📲

PR #4907, opened 2020-09-18
feat(security/credentials): AccountCredentials as kork/Credentials

ncknt:
This is the second part of the change described here: https://github.com/spinnaker/governance/blob/master/rfc/account-management.md

`AccountCredentials` is now also a kork `Credentials`. We wire the `DefaultAccountCredentialsProvider` to also look for credentials in the future `CredentialsRepository` we'll create. When all providers are migrated this class should go away in favor of `CompositeCredentialsRepository` which we also create here.

Finally we wire pollers to reload credentials repositories with https://github.com/spinnaker/kork/blob/master/kork-credentials/src/main/java/com/netflix/spinnaker/credentials/poller/Poller.java#L27.


ncknt:
@cfieber @robzienert would you mind taking a look? It's the continuation of last week's PR. After this is merged, work should be per-provider.

clouddriver-api/clouddriver-api.gradle
```diff
@@ -21,6 +21,7 @@ dependencies {
   annotationProcessor platform("com.netflix.spinnaker.kork:kork-bom:$korkVersion")
 
   api("com.netflix.spinnaker.kork:kork-plugins-api")
+  api("com.netflix.spinnaker.kork:kork-credentials")
```
robzienert, line 24: No can do. `api` modules must be _very_ limited in the dependency graph that it exposes: `kork-credentials` has a dependency on Spring. If you want to expose `kork-credentials`, you'll need to remove the Spring dependency, e.g. by creating a `kork-credentials-spring` or similar that is included in services, but not included within `api` modules.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/security/AbstractAtomicOperationsCredentialsConverter.java
```diff
@@ -0,0 +1,63 @@
...
+  @Getter private ObjectMapper objectMapper;
+
+  @Autowired
+  public void setObjectMapper(ObjectMapper objectMapper) {
+    // TODO(rz): This is a bad pattern, we should be using the object mapper customizer bean, rather
+    //  than modifying a singleton, global object mapper after injecting it somewhere.
+    this.objectMapper =
+        objectMapper
+            .configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false)
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
robzienert, line 46: Can we take the opportunity to not propagate this pattern?

robzienert (changes requested):
The `api` module needs addressing.

clouddriver-api/clouddriver-api.gradle
ncknt: I split `kork-credentials-api` here: https://github.com/spinnaker/kork/pull/787. Will update that PR afterwards.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/security/AbstractAtomicOperationsCredentialsConverter.java
ncknt: Sure. Any reason you didn't just create a new objectMapper? These are in `@Component` so the instantiation cost is pretty negligible overall.

clouddriver-api/clouddriver-api.gradle
ncknt: Updated with `clouddriver-api` -> `kork-credentials-api`, `clouddriver-core` -> `kork-credentials`. Marking as resolved.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/security/AbstractAtomicOperationsCredentialsConverter.java
ncknt: Just used a new instance. The mapper is instantiated once per component.

robzienert (approved):
Much more bester.

PR #4909, opened 2020-09-18
feat(aws): Adding delete launch template operation

jeyrschabu:
- adding support for lt deletion
- updated to delete lt on asg deletion


jeyrschabu:
This will also be used by swabbie.

Will PR the orca part later.

aravindmd (approved):

caseyhebebrand (approved):

PR #4911, opened 2020-09-18
chore(dependencies): Autobump korkVersion

spinnakerbot:


robzienert:
Looks like this was already applied: https://github.com/spinnaker/clouddriver/blob/master/gradle.properties#L2

PR #4916, opened 2020-09-22
feat(titus): Wrap `DestroyTitusJob` operation in a simple saga

ajordens:
This PR address one of a small handful of Titus operations that have yet
to receive the saga treatment.

A subsequent PR will handle the remainders.


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/converters/DestroyTitusJobAtomicOperationConverter.groovy
```diff
@@ -34,27 +33,25 @@ import org.springframework.stereotype.Component
...
   AtomicOperation convertOperation(Map input) {
-    new DestroyTitusJobAtomicOperation(titusClientProvider, convertDescription(input))
+    new DestroyTitusJobAtomicOperation(convertDescription(input))
   }
 
   @Override
   DestroyTitusJobDescription convertDescription(Map input) {
     def converted = objectMapper.convertValue(input, DestroyTitusJobDescription)
     converted.credentials = getCredentialsObject(input.credentials as String)
+    converted.account = converted.credentials.name
```
ajordens, line 54: @robzienert Added this as `credentials` was `@JsonIgnore` and for some reason, I saw it getting lost during serialization.

Looking at the code that _shouldn't_ be the case right?

ajordens: Apparently a lie, I just re-tested and all is good.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/ResolveTitusJobId.java
```diff
@@ -0,0 +1,102 @@
...
+    AccountCredentials accountCredentials =
+        accountCredentialsProvider.getCredentials(command.description.getAccount());
+
+    TitusClient titusClient =
+        titusClientProvider.getTitusClient(
+            (NetflixTitusCredentials) accountCredentials, command.description.getRegion());
+
+    Job job = titusClient.findJobByName(command.description.getServerGroupName());
+    if (job != null) {
+      DestroyTitusJobDescription destroyTitusJobDescription = new DestroyTitusJobDescription();
```
ajordens, line 69: This worked first shot ... is it legitimate? 

@robzienert Is this what you were referring to with context needing to be explicitly passed forward?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/ops/DestroyTitusServerGroupAtomicOperation.groovy
```diff
@@ -16,58 +16,54 @@
...
-      events << new DeleteServerGroupEvent(
-        TitusCloudProvider.ID, description.credentials.name, description.region, description.serverGroupName
-      )
-      task.updateStatus PHASE, "Successfully issued terminate job request to titus for ${job.id} which corresponds to ${description.serverGroupName}"
-    } else {
-      task.updateStatus PHASE, "No titus job found for ${description.serverGroupName}"
-    }
+  protected void configureSagaBridge(@NotNull @Nonnull SagaAtomicOperationBridge.ApplyCommandWrapper.ApplyCommandWrapperBuilder builder) {
+    builder.initialCommand(
+      ResolveTitusJobId.ResolveTitusJobIdCommand.builder().description(description).build()
```
ajordens, line 55: I could avoid the `.description(description)` and actually pass in account/region/serverGroupName/user explicitly.

ajordens:
I need to update the failing spec here as it's now relevant in a different way.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/sagas/AbstractSagaAtomicOperation.java
```diff
@@ -35,7 +35,7 @@
  * @param <T> The AtomicOperation description
  * @param <R> The operation result type
  */
-public abstract class AbstractSagaAtomicOperation<T, R>
+public abstract class AbstractSagaAtomicOperation<T, SR, R>
```
robzienert, line 38: Would like to see this type documented up top.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/ResolveTitusJobId.java
robzienert: > .. is it legitimate?

Which part, creating a description inside of the action? Yeah, that's legit.

> @robzienert Is this what you were referring to with context needing to be explicitly passed forward?

Sort of. This is actually what we want, actions that can be composed together. The thing where the design is flawed is in the `Result` object, requiring a `Command` for the next action to run - this makes each action tightly coupled, which isn't desired.

We can change things such that actions return _only_ events, and a command is only used for kicking off a Saga. Actions, then, would just specify the events they require, and those events would be injected in `apply()` instead of a `Command`.

So, in this case instead of returning a `Result(DestroyTitusJobCommand)` we'd return a `Result(TitusJobIdResolved)` and then `DestroyTitusJobAction` would require `TitusJobIdResolved`. The only thing linking the two actions together would be the Flow.

Does that make sense?

robzienert (approved):
Looks right!

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/DestroyTitusJob.java
```diff
@@ -0,0 +1,109 @@
...
+        command.description.getJobId());
+
+    AccountCredentials accountCredentials =
+        accountCredentialsProvider.getCredentials(command.description.getAccount());
+
+    TitusClient titusClient =
+        titusClientProvider.getTitusClient(
+            (NetflixTitusCredentials) accountCredentials, command.description.getRegion());
+
+    Job job = titusClient.getJobAndAllRunningAndCompletedTasks(command.description.getJobId());
```
jeyrschabu, line 71: This call could fail. Could handle the exception and make it retryable?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/ResolveTitusJobId.java
jeyrschabu: Looks legit to me. Though I suppose the description could also just be passed forward.

jeyrschabu (approved):
minor comments but lgtm.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/DestroyTitusJob.java
ajordens: Whole point of the saga right? The `TitusExceptionHandler` is going to flag the action as retryable based on its criteria and `orca` will automatically retry.

ajordens:
@jeyrschabu It doesn't make sense to pass the original description/command context forward. In this case the `DestroyTitusJob` is usable independently from `DestroyTitusServerGroup`.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/DestroyTitusJob.java
jeyrschabu: Yup, didn’t notice the exception handler at first glance.

PR #4917, opened 2020-09-22
fix(aws): Fixing launch template flag

jeyrschabu:
- handle empty list of allowed accounts/regions


jeyrschabu:
Merging to calm our automated pipelines

PR #4919, opened 2020-09-22
chore(java11): Target Java 11

plumpy:
After this commit, we can use Java 11 language features and APIs.

I added a single var declaration to make it more obvious if any of our various CI systems are misconfigured.

I had to remove `final` from a bunch of Spock tests because of an issue with Spock: spockframework/spock#1011

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -10,11 +10,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    # Install Java 8 for cross-compilation support. Setting it up before
-    # Java 11 means it comes later in $PATH (because of how setup-java works)
-    - uses: actions/setup-java@v1
-      with:
-        java-version: 8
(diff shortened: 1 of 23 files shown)


.github/workflows/pr.yml
```diff
@@ -25,4 +20,6 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-gradle-
     - name: Build
-      run: ./gradlew -PenableCrossCompilerPlugin=true build
+      run: ./gradlew build
+    - name: Integration Tests
```
maggieneterval, line 24: Were lines 24 & 25 not meant to be part of this commit?

maggieneterval (approved):
🎯 ☕ 1️⃣ 1️⃣ 🚀 💯

.github/workflows/pr.yml
plumpy: Oops, there was a merge conflict and I got confused and resolved it backwards! Thanks for catching that!

robzienert (approved):

PR #4927, opened 2020-09-23
fix(sql): Add config for limiting amount of aggregate cleanup

robzienert:
Agent isn't completing in time. Not a big deal, just kinda annoying. This will be good safety to have anyway.


jonsie (approved):

clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlEventCleanupAgentConfigProperties.kt
```diff
@@ -39,4 +39,14 @@ class SqlEventCleanupAgentConfigProperties {
...
   var maxAggregateAgeDays: Long = 7
+
+  /**
+   * The max number of events to cleanup in each agent invocation.
+   */
+  @Positive
+  var cleanupLimit: Int = EVENT_CLEANUP_LIMIT
+
+  companion object {
+    const val EVENT_CLEANUP_LIMIT = 100_000
```
marchello2000, line 50: 100k still takes a decent time for me to return ~7secs
but maybe a reasonable default 🤷

PR #4929, opened 2020-09-24
feat(extensibility): Use ExceptionSummaryService in DefaultOrchestrationProcessor to provide custom user messages on exceptions

jonsie:
Depends on: https://github.com/spinnaker/kork/pull/788


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/DefaultOrchestrationProcessor.groovy
```diff
@@ -213,7 +217,8 @@ class DefaultOrchestrationProcessor implements OrchestrationProcessor {
    * TODO(rz): Not 100% sure we should keep these two methods.
    */
   Map<String, Object> extractExceptionSummary(Throwable e, String userMessage) {
-    def summary = ExceptionSummary.from(e)
+    ExceptionSummary summary = exceptionSummaryService.summary(e)
```
jonsie, line 220: Same basic behavior as before, except it'll check the exception type against a list of `UserMessageAppender` objects which can optionally append messages to the user message.

robzienert (approved):

PR #4931, opened 2020-09-24
fix(saga): Support for multiple atomic operations in a single task

robzienert:
The idea is simple: Tasks can have more than one `AtomicOperation` submitted for it. Changes recently went in that made each saga ID unique to correctly support this use case, but it broke resumability of sagas in the process, and kind of caused some cascading failures. This PR fixes the resume capabilities. Resume requests will now search through all sagas attached to a Task and filter them down to a unique list of sagas (which it should already be, actually) based on the snapshotted description information: Each unique description is a unique atomic operation that has its own saga. The saga's ID is captured and added to the SagaContext, which is already used for bridge data between sagas and atomic operations - we use this context data if present, or create a new UUID if absent (which will happen any time the operation has never run).

I need a shower. 😆  This doesn't include tests, I'm a little fried so I won't be doing that tonight. I'm going to be out (or in meetings) for most of the day tomorrow, so if someone needs to expedite this into an environment and I'm unavailable, they at least have something to start from. This _should_ work as-is, but due to reasons, I can't run clouddriver on my laptop anymore so I haven't run it through any spot-checks.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
@@ -24,7 +24,9 @@ import com.netflix.spinnaker.clouddriver.deploy.DescriptionAuthorizer;
 import com.netflix.spinnaker.clouddriver.deploy.DescriptionValidationErrors;
 import com.netflix.spinnaker.clouddriver.deploy.DescriptionValidationException;
 import com.netflix.spinnaker.clouddriver.deploy.DescriptionValidator;
+import com.netflix.spinnaker.clouddriver.orchestration.sagas.AbstractSagaAtomicOperation;
 import com.netflix.spinnaker.clouddriver.orchestration.sagas.SnapshotAtomicOperationInput.SnapshotAtomicOperationInputCommand;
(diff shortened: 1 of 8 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
```diff
@@ -204,27 +207,59 @@ public OperationsService(
...
-                        Collections.singletonMap(
-                            it.getDescriptionName(), it.getDescriptionInput()))))
-        .flatMap(Collection::stream)
-        .map(this::atomicOperationOrError)
+            saga ->
+                new SagaAndSnapshot(saga, saga.getEvent(SnapshotAtomicOperationInputCommand.class)))
+        .filter(
+            it -> {
+              // Reduce the list of sagas attached to the task to one for each uniquely submitted
+              // description. This is probably unnecessary long-term.
```
robzienert, line 229: Unnecessary long-term, because we'll be back to keeping a unique list of saga IDs on a task once this is rolled out, like the old behavior once was. This is really only needed for currently executing sagas that are currently appending new saga IDs.

ajordens (approved):
This jives with my expectations. I'll give it a whirl.

PR #4932, opened 2020-09-24
feat(aws): Add ability to specificy externalId to AWS account for assuming roles via STS

fieldju:
My goal is to enable the use of the externalId parameter when assuming a role as outlined here:
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html

I am opening the PR now as a draft, while I figure out how to test this.
Digging through the code, this looks like what needs to be changed though.


dbyron0:
Not related to this PR, but I've been staring at this part of the code trying to figure out how to make spinnaker work when the STS global endpoint token version is set to `Valid only in AWS Regions enabled by default` (see [here](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html#sts-regions-manage-tokens) for more).

A couple of related slack threads: https://spinnakerteam.slack.com/archives/CB78MFQPP/p1600364065024900, https://spinnakerteam.slack.com/archives/CB78MFQPP/p1600373491026100.

Any chance you're working with an AWS account set up that way and found a way to make it work for you?

fieldju:
https://github.com/spinnaker/clouddriver/pull/4968

PR #4936, opened 2020-09-24
 feat(ecs): Moniker support for ECS server groups

deverton:
Enable support for using moniker for ECS server groups implementing https://github.com/spinnaker/spinnaker/issues/6062

Service names are constructed as before, however if moniker support is enabled, the moniker values are represented as tags on the ECS Service and are propagated to tasks. The tag keys are the same as the Kubernetes V2 provider:

| Key | Value |
|----|-----|
| `moniker.spinnaker.io/application` | The application name |
| `moniker.spinnaker.io/cluster ` | Cluster the application belongs to |
| `moniker.spinnaker.io/stack` | Stack the application belongs to |
| `moniker.spinnaker.io/detail` | Additional detail |
| `moniker.spinnaker.io/sequence` | The version sequence number |

With a standard configuration the default Spinnaker naming strategy will be used and no tags will be applied.

```yaml
ecs:
  enabled: true
  accounts:
    - name: "ecs-cicddev-test"
      awsAccount: "ec2-cicddev-test"
```

To have moniker tags applied at the account level you can set the `namingStrategy` field to `tags` like:

```yaml
ecs:
  enabled: true
  accounts:
    - name: "ecs-cicddev-test"
      awsAccount: "ec2-cicddev-test"
      namingStrategy: "tags"
```

Or to enable it for all accounts

```yaml
ecs:
  enabled: true
  defaultNamingStrategy: "tags"
  accounts:
    - name: "ecs-cicddev-test"
      awsAccount: "ec2-cicddev-test"
```

You can also mix strategies between accounts

```yaml
ecs:
  enabled: true
  defaultNamingStrategy: "tags"
  accounts:
    - name: "ecs-cicddev-test"
      awsAccount: "ec2-cicddev-test"
    - name: "ecs-cicddev-test"
      awsAccount: "ec2-cicddev-test"
      namingStrategy: "default"
```

Based on #3258 with the tag key convention taken from the Kubernetes V2 provider.


pkandasamy91:
@deverton thanks for taking the time to add moniker support to ECS! 

Would you mind providing some more detail around what users would see differently from a service that is deployed prior to this change vs with this change (specifically in ECS)?

If you could also add some screenshots of how these changes in clouddriver would be viewed in deck that would be great!

deverton:
So there isn't actually much change in the UI when this is enabled as everything works as it did before. The only change is in how the the applications and server groups are grouped and presented.

In my testing I've deployed a service to ECS called `deverton-spin-test` which has one instance running. By default it shows like this in the application view in Deck

deck-before

If I then go in to the ECS console and add tags like this

ecs-tags

Wait for `clouddriver` to refresh the account and you get this

deck-after

After on-boarding the application as usual, everything works as you would expect. The big change is that Spinnaker now recognises the stack and detail values from the tags rather than parsing from the name. It even treats the existing server group properly

copy-settings

deverton:
Oh, and in the ECS console it ends up looking something like this

ecs-console-view

The test case here is to have Spinnaker adopt an existing service deployed through other means. So the service without the `-v002` was already there and I had Spinnaker copy that and deploy a second version shifting traffic as expected. It keeps the old service around and can be rolled back to.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsServerClusterProvider.java
```diff
@@ -171,11 +170,11 @@ public EcsServerClusterProvider(
...
 
       if (clusterMap.containsKey(applicationName)) {
-        String escClusterName = StringUtils.substringBeforeLast(ecsServerGroup.getName(), "-");
+        String ecsServerClusterName = moniker.getCluster();
         boolean found = false;
 
         for (EcsServerCluster cluster : clusterMap.get(applicationName)) {
-          if (cluster.getName().equals(escClusterName)
+          if (cluster.getName().equals(ecsServerClusterName)
               && cluster.getAccountName().equals(credentials.getName())) {
```
pkandasamy91, line 178: nit: since we're using StringUtils in most places, would you mind using StringUtils.equals here?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsServerClusterProvider.java
```diff
@@ -534,13 +535,13 @@ public ServerGroup getServerGroup(
...
-    // TODO - remove the application filter.
-    String application = StringUtils.substringBefore(serverGroupName, "-");
+
     Map<String, Set<EcsServerCluster>> clusterMap = new HashMap<>();
 
     try {
       AmazonCredentials credentials = getEcsCredentials(account);
-      clusterMap = findClusters(clusterMap, credentials, application);
+      // Can't filter by application as there's not enough information in the serverGroupName
+      clusterMap = findClusters(clusterMap, credentials);
```
pkandasamy91, line 544: Would this be a regression?

pkandasamy91 (changes requested):
@deverton this is incredible! I love it and wanted to say thanks for adding support for monikers in ECS!

One thing to consider though is that tagging is not supported for all aws account users, and so this would break functionality to use spinnaker for other customers who have opted out/have not opted into using the new arn format. 

What do you think about doing a lookup by making a ListAccountSettings request: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ListAccountSettings.html. From what I can tell we could make an sdk call to see if a role is opted in to the ServiceLongArnFormat/TaskLongArnFormat (https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Setting.html#API_Setting_Contents)?

If the account isn't opted in(has explicitly opted out), use the existing/legacy logic, and if the account has opted in, then use your new logic with moniker support (this would require some aggregated logic for lookups though).

pkandasamy91:
To add some more context please see: https://github.com/spinnaker/clouddriver/pull/4936#discussion_r504295263

deverton:
Thanks for the feedback, I'll take a look.

As for the tag support, there's a feature gate with the `ecs.applyTags` setting. So rather than trying to guess the user intention, I think that code just needs to be updated to not tag if that's not enabled. Though, I guess it would need to move down to being an account level setting rather than something for the whole provider.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsServerClusterProvider.java
deverton: It shouldn't be as the `clusterMap` collection is still filtered to find the matching `serverGroupName` a little further down.

https://github.com/spinnaker/clouddriver/pull/4936/files/4d5938d458e864ffaadcf14223f3195a91c5e3e4#diff-08ef4737fffbc8ef45383c06e16b3a2f1a9a2fef2d4b4b314665b4f134c0b5a4R553

Given that the server group name still includes the application name this should still meet the requirements of the interface.

clouddriver-ecs/src/integration/resources/testoperations/createServerGroup-inputs-ecsCreateFails.json
```diff
@@ -20,6 +20,9 @@
   "ecsClusterName": "integInputsErrorFromEcs-cluster",
   "launchType": "EC2",
   "networkMode": "bridge",
+  "moniker": {
+    "app": "ecs"
+  },
```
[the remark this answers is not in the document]
deverton: The `moniker.app` field is required in the body of the server group create call as the `EcsCreateServerGroupDescriptionValidator` now validates it's there.

https://github.com/spinnaker/clouddriver/pull/4936/files/4d5938d458e864ffaadcf14223f3195a91c5e3e4#diff-98df6fef9d39f2021cad2a3b31a12e52ade2acfddb206c542d16bbf64bf1b9bdR104

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -806,33 +847,37 @@ private String getServerGroupName(Service service) {
...
+
+    Moniker m = newServerGroupName.getMoniker();
 
-    targetEnv.add(new KeyValuePair().withName("SERVER_GROUP").withValue(newServerGroupName));
-    targetEnv.add(new KeyValuePair().withName("CLOUD_STACK").withValue(description.getStack()));
     targetEnv.add(
-        new KeyValuePair().withName("CLOUD_DETAIL").withValue(description.getFreeFormDetails()));
+        new KeyValuePair().withName("SERVER_GROUP").withValue(newServerGroupName.getServiceName()));
+    targetEnv.add(new KeyValuePair().withName("CLOUD_STACK").withValue(m.getStack()));
+    targetEnv.add(new KeyValuePair().withName("CLOUD_DETAIL").withValue(m.getDetail()));
```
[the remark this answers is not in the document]
deverton: That I'm not sure on. The existing code would have added the environment variables regardless of the values so I've kept that behaviour I think.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsServerClusterProvider.java
deverton: Should be fixed.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/names/EcsServerGroupName.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+    if (StringUtils.isEmpty(stack) && StringUtils.isNotEmpty(detail)) {
+      sb.append("-");
+    }
+    if (StringUtils.isNotEmpty(detail)) {
+      sb.append("-").append(detail);
+    }
+    return sb.toString();
+  }
+}
```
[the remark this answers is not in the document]
deverton: Can't quite structure it like that, but I did end up with something simpler I think in the new `MonikerHelper` class.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/names/EcsResourceTagger.java
```diff
@@ -0,0 +1,131 @@
...
+      sb.append("-").append(stack);
+    }
+    if (StringUtils.isEmpty(stack) && StringUtils.isNotEmpty(detail)) {
+      sb.append("-");
+    }
+    if (StringUtils.isNotEmpty(detail)) {
+      sb.append("-").append(detail);
+    }
+    return sb.toString();
+  }
```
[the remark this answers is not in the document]
deverton: Can't quite structure it like that, but I did end up with something simpler I think in the new `MonikerHelper` class.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/model/EcsServerGroup.java
```diff
@@ -43,12 +44,13 @@
   Capacity capacity;
   ImagesSummary imagesSummary;
   ImageSummary imageSummary;
-  Map<String, Object> tags;
```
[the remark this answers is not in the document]
deverton: I think that was left over from one of my earlier attempts. I've restored it.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -806,33 +847,37 @@ private String getServerGroupName(Service service) {
   }
 
   private Collection<KeyValuePair> setSpinnakerEnvVars(
-      Collection<KeyValuePair> targetEnv, String newServerGroupName) {
+      Collection<KeyValuePair> targetEnv, EcsServerGroupName newServerGroupName) {
+
+    Moniker m = newServerGroupName.getMoniker();
```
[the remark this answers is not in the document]
deverton: Good idea. I think I've fixed all the cases of short names.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -806,33 +847,37 @@ private String getServerGroupName(Service service) {
...
-      newLabels.put(DOCKER_LABEL_KEY_STACK, description.getStack());
+    Moniker m = newServerGroupName.getMoniker();
+
+    if (StringUtils.isNotEmpty(m.getStack())) {
+      newLabels.put(DOCKER_LABEL_KEY_STACK, m.getStack());
     }
 
-    if (description.getFreeFormDetails() != null) {
-      newLabels.put(DOCKER_LABEL_KEY_DETAIL, description.getFreeFormDetails());
+    if (StringUtils.isNotEmpty(m.getDetail())) {
```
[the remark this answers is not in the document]
deverton: The existing behaviour was to add the labels regardless of value, same as the environment variables. I'm not sure it's a good idea to do that, but fixing it would be a breaking change.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/description/AbstractECSDescription.java
```diff
@@ -23,8 +23,5 @@
 @Data
 @EqualsAndHashCode(callSuper = false)
 public abstract class AbstractECSDescription extends AbstractAmazonCredentialsDescription {
-  String application;
   String region;
-  String stack;
-  String freeFormDetails;
```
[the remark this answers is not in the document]
deverton: As far as I can tell, there's no impact as those operations all make use of the server group name rather than any of the other details.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/description/CreateServerGroupDescription.java
```diff
@@ -30,6 +31,16 @@
...
+  /** @deprecated This field is deprecated in favour of [moniker.app] */
+  @Deprecated String application;
+
+  /** @deprecated This field is deprecated in favour of [moniker.stack] */
+  @Deprecated String stack;
+
+  /** @deprecated This field is deprecated in favour of [moniker.detail] */
+  @Deprecated String freeFormDetails;
+
+  Moniker moniker;
```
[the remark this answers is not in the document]
deverton: These fields seemed to be specific to the create server group action. The other ones all operate on the server group name.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -466,9 +484,12 @@ private Service createService(
...
+      String taskDefinitionArn,
+      EcsServerGroupName newServerGroupName,
+      Integer desiredCount,
+      Namer<EcsResource> namer) {
     Collection<LoadBalancer> loadBalancers =
-        retrieveLoadBalancers(EcsServerGroupNameResolver.getEcsContainerName(newServerGroupName));
+        retrieveLoadBalancers(newServerGroupName.getContainerName());
 
     Collection<ServiceRegistry> serviceRegistries = new LinkedList<>();
     if (description.getServiceDiscoveryAssociations() != null) {
```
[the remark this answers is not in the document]
deverton: Done.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -806,33 +847,37 @@ private String getServerGroupName(Service service) {
...
     Map<String, String> newLabels = new HashMap<>();
     if (targetMap != null) {
       newLabels.putAll(targetMap);
     }
 
-    if (description.getStack() != null) {
-      newLabels.put(DOCKER_LABEL_KEY_STACK, description.getStack());
+    Moniker m = newServerGroupName.getMoniker();
+
+    if (StringUtils.isNotEmpty(m.getStack())) {
```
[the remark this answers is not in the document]
deverton: The existing behaviour was to add the labels regardless of value, same as the environment variables. I'm not sure it's a good idea to do that, but fixing it would be a breaking change.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/names/EcsServerGroupNameResolver.java
```diff
@@ -0,0 +1,142 @@
...
+    List<String> allServices = listAllServices(ecsClusterName);
+
+    // 2. Get the details of the services in the same server group
+    List<List<String>> serviceBatches = Lists.partition(allServices, 10);
+    for (List<String> serviceBatch : serviceBatches) {
+      DescribeServicesRequest request =
+          new DescribeServicesRequest()
+              .withCluster(ecsClusterName)
+              .withServices(serviceBatch)
+              .withInclude("TAGS");
```
deverton, line 66: @pkandasamy91 it's not clear from the AWS ECS API documentation if it's safe to call `ecs:DescribeServices` with include set to `TAGS`. I'm assuming this is safe but I've got no easy way to test as all accounts I have access to have the new ARN format enabled.

deverton:
So the latest change to this PR delegates the decision to tag or not to the naming strategy. If no naming strategy is configured, then no tags are applied. If the tagging naming strategy `ecsTags` is set at either the account level or at the provider level, then it will be used and will add tags.

This way you can mix accounts that have the new ARN enabled and disabled at the account level and still have a simple feature toggle at the provider level.

deverton:
I'm also hoping to (eventually) extend this approach to the AWS provider, though that seems like it will be more difficult. It would at least enable spinnaker/spinnaker#6120 by keying off the presence of the application tag.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/description/AbstractECSDescription.java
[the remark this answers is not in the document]
piradeepk: I'm a bit hesitant on removing the property from the top level description since we don't know for certain that there's no deserialization that they're used for. Since there's not really any added benefit (other than code cleanliness), do you mind leaving them there for now?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/description/CreateServerGroupDescription.java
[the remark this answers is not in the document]
piradeepk: Please see previous comment.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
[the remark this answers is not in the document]
piradeepk: That's a fair point. I think we should fix it, but I agree it's outside the scope of this PR.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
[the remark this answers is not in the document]
piradeepk: See previous comment.

piradeepk:
> So the latest change to this PR delegates the decision to tag or not to the naming strategy. If no naming strategy is configured, then no tags are applied. If the tagging naming strategy `ecsTags` is set at either the account level or at the provider level, then it will be used and will add tags.
> 
> This way you can mix accounts that have the new ARN enabled and disabled at the account level and still have a simple feature toggle at the provider level.

@deverton Thanks for addressing the previous feedback. I think we should still look at adding logic that checks that the particular account is using the new long arn format. It's possible for users to not realize that they need an account opted into using the long arn format in order to add tags, and we should handle that use case to provide them with feedback.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -495,25 +515,46 @@ protected CreateServiceRequest makeServiceRequest(
...
             .withPlacementStrategy(description.getPlacementStrategySequence())
             .withServiceRegistries(serviceRegistries)
-            .withDeploymentConfiguration(deploymentConfiguration);
+            .withDeploymentConfiguration(deploymentConfiguration)
+            .withTags(taskDefTags)
+            .withEnableECSManagedTags(true)
+            .withPropagateTags("SERVICE");
+
+    // Apply moniker to tags which are subsequently recorded in the task tags
+    namer.applyMoniker(
```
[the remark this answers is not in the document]
deverton: Would you expect Spinnaker to be able to handle this changing at runtime? Or should this be done at the account level so only checked at startup?

I don't think we should be calling `ListAccountSettings` that often so checking it during account creation makes the most sense to me, but then that requires restarting Spinnaker to pick up changes.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/description/AbstractECSDescription.java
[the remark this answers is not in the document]
deverton: No worries, that should be fixed with the latest commit.

deverton:
In the end, I've added a query inline to the create server group operation to run `ecs:ListAccountSettings` to check if tagging is enabled. If it's not and there are tags specified (either manually or via the namer) then the deploy will fail with an exception.

piradeepk:
This is awesome 💯 @deverton! I'll take a look at your PR and hope to have some feedback/approval for you by early next week!

piradeepk (commented):
Sorry for the delay @deverton, it's taken substantially longer than expected due to the sheer size of this PR. 

In the future, for a substantial feature, such as Moniker support, we would really appreciate it if you could submit a design proposal as an issue as it'll allow us to work through the intent and design the implementation prior to writing any logic. 

Also, it looks like there are incremental changes in the PR that are not directly related to supporting Monikers. I'd appreciate it if you could move those to a separate PR.

If you could test using an account that hasn't opted-in/has opted-out of the long arn format. I would suggest logging into the AWS console with one of your test accounts and opting out of the long arn format (in a particular region - i believe opt-in is account and region specific). 
- Kick off a new Spinnaker deployment/create resources in Spinnaker
- Verify that everything continues to function in Spinnaker. 
- Verify that the arn format in the AWS console for the resources that were just created after you opted out now uses the short arn format (if you're still using the long arn format, then you have not opted out successfully).

clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/view/EcsApplicationProviderSpec.groovy
```diff
@@ -61,8 +61,8 @@ class EcsApplicationProviderSpec extends Specification {
       desiredCount: 1,
       createdAt: new Date()
     )
-    def attributes = ServiceCachingAgent.convertServiceToAttributes(credentials.getName(),
-      credentials.getRegions()[0].getName(), service)
+    def attributes = TestServiceCachingAgent.create(credentials,
+      credentials.getRegions()[0].getName()).convertServiceToAttributes(service)
```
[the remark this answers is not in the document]
deverton: The `convertServiceToAttributes()` method used to be a static method on the `ServiceCachingAgent` class. As we need to pass in the naming strategy to be able to derive the correct moniker values, it can no longer be static.

This meant that the tests needed a way to create service caching agent instances for testing and hence the creation of the test caching agent.

clouddriver-ecs/src/test/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TestServiceCachingAgent.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ */
+
+package com.netflix.spinnaker.clouddriver.ecs.provider.agent;
+
+import com.netflix.spectator.api.NoopRegistry;
+import com.netflix.spinnaker.clouddriver.aws.security.NetflixAmazonCredentials;
+import com.netflix.spinnaker.clouddriver.names.NamerRegistry;
+
+public class TestServiceCachingAgent {
```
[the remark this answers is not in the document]
deverton: See below comment, but it's needed for testing the caching with different naming strategies.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/ServiceCacheClient.java
```diff
@@ -79,6 +81,25 @@ protected Service convert(CacheData cacheData) {
 
     service.setCreatedAt((Long) attributes.get("createdAt"));
 
+    if (attributes.containsKey("moniker")) {
+      service.setMoniker(objectMapper.convertValue(attributes.get("moniker"), Moniker.class));
+    }
+
+    if (attributes.containsKey("tags")) {
+      List<Map<String, Object>> tags = (List<Map<String, Object>>) attributes.get("tags");
+      List<Tag> deserializedTags = new ArrayList<>(tags.size());
```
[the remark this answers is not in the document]
deverton: Turns out we don't need to cache the tags anyway for the moniker support to work. So I think I'll remove this and other tag caching in the agent.

deverton:
If you have configured the ECS account naming strategy to be `ecsTags` and the new ARN format is disabled in the target account, you will get the following error message when trying to create a server group.

image

The stack trace is what you would expect

```
2020-11-09 13:21:33.157  INFO 17284 --- [tionProcessor-0] c.n.s.c.data.task.jedis.JedisTask        : [ORCHESTRATION] Processing op: CreateServerGroupAtomicOperation
2020-11-09 13:21:33.170  INFO 17284 --- [tionProcessor-0] c.n.s.c.data.task.jedis.JedisTask        : [CREATE_ECS_SERVER_GROUP] Initializing Create Amazon ECS Server Group Operation...
2020-11-09 13:21:34.206  INFO 17284 --- [tionProcessor-0] c.n.s.c.data.task.jedis.JedisTask        : [CREATE_ECS_SERVER_GROUP] Creating Amazon ECS Task Definition...
2020-11-09 13:21:34.455  INFO 17284 --- [tionProcessor-0] c.n.s.c.data.task.jedis.JedisTask        : [CREATE_ECS_SERVER_GROUP] Done creating Amazon ECS Task Definition...
2020-11-09 13:21:34.701  INFO 17284 --- [tionProcessor-0] c.n.s.c.data.task.jedis.JedisTask        : [ORCHESTRATION] Orchestration failed: CreateServerGroupAtomicOperation | IllegalArgumentException: [Tagging is not enabled on ECS account ecs-test.]
2020-11-09 13:21:34.715 ERROR 17284 --- [tionProcessor-0] c.n.s.c.o.DefaultOrchestrationProcessor  : java.lang.IllegalArgumentException: Tagging is not enabled on ECS account ecs-test.
	at com.netflix.spinnaker.clouddriver.ecs.deploy.ops.CreateServerGroupAtomicOperation.makeServiceRequest(CreateServerGroupAtomicOperation.java:542)
	at com.netflix.spinnaker.clouddriver.ecs.deploy.ops.CreateServerGroupAtomicOperation.createService(CreateServerGroupAtomicOperation.java:442)
	at com.netflix.spinnaker.clouddriver.ecs.deploy.ops.CreateServerGroupAtomicOperation.operate(CreateServerGroupAtomicOperation.java:131)
	at com.netflix.spinnaker.clouddriver.ecs.deploy.ops.CreateServerGroupAtomicOperation.operate(CreateServerGroupAtomicOperation.java:74)
...
```

The same error will show up if you are using the default naming strategy but set tags in the "Create Server Group" action/stage.

If you have disabled the new ARN format and don't use tags or the moniker naming strategy then ECS resources are created as before.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/names/EcsServerGroupNameResolver.java
deverton: Tested this as part of some other testing and it appears to work fine on accounts with the new ARN format disabled.

clouddriver-ecs/src/test/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/ServiceCachingAgentTest.java
```diff
@@ -114,7 +114,8 @@ public void shouldGenerateFreshData() {
                       .withMaximumPercent(100))
               .withLoadBalancers(Collections.emptyList())
               .withDesiredCount(1)
-              .withCreatedAt(new Date()));
+              .withCreatedAt(new Date())
+              .withTags(Collections.emptyList()));
```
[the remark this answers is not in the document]
deverton: This was just a unit test, but tag caching is removed now anyway.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsServerClusterProvider.java
```diff
@@ -85,7 +81,8 @@ public EcsServerClusterProvider(
       ScalableTargetCacheClient scalableTargetCacheClient,
       EcsLoadbalancerCacheClient ecsLoadbalancerCacheClient,
       TaskDefinitionCacheClient taskDefinitionCacheClient,
-      EcsCloudWatchAlarmCacheClient ecsCloudWatchAlarmCacheClient) {
+      EcsCloudWatchAlarmCacheClient ecsCloudWatchAlarmCacheClient,
+      Namer<EcsResource> namer) {
```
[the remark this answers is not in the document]
deverton: Good catch, removed.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/names/MonikerHelper.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    if (StringUtils.isNotEmpty(stack)) {
+      return String.join("-", appName, stack);
+    }
+
+    return appName;
+  }
+
+  public static Moniker friggaToMoniker(String friggaName) {
```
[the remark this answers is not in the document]
deverton: Done.

deverton:
> In the future, for a substantial feature, such as Moniker support, we would really appreciate it if you could submit a design proposal as an issue as it'll allow us to work through the intent and design the implementation prior to writing any logic.

I'm happy to write up a design document based on my findings from this PR. Originally I worked on this to see if it was even possible within the existing design and figured submitting it was a way to start the discussion on whether this is the right approach.

clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperationSpec.groovy
```diff
@@ -129,6 +134,10 @@ class CreateServerGroupAtomicOperationSpec extends CommonAtomicOperation {
     def result = operation.operate([])
 
     then:
+    ecs.listAccountSettings(_) >> new ListAccountSettingsResult().withSettings(
+      new Setting(name: SettingName.TaskLongArnFormat, value: "enabled"),
+      new Setting(name: SettingName.ServiceLongArnFormat, value: "enabled")
+    )
```
piradeepk, line 140: Can you also add tests to account for only Task or Service long arns being enabled?

clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperationSpec.groovy
```diff
@@ -1234,4 +1261,107 @@ class CreateServerGroupAtomicOperationSpec extends CommonAtomicOperation {
...
+    then:
+    assert request.enableECSManagedTags == true
+    assert request.propagateTags == 'SERVICE'
+    def tags = request.getTags()
+    assert tags.size() == 2
+    tags.contains(new Tag(key: 'label1', value: 'value1'))
+    tags.contains(new Tag(key: 'fruit', value: 'tomato'))
+  }
+
+  def 'should create custom tags and moniker tags if moniker enabled'() {
```
piradeepk, line 1335: Can you also add a test for the exception case - when moniker support is enabled (enableTags) but the account is not configured to use the long arn format?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/security/EcsCredentialsInitializer.java
```diff
@@ -76,6 +84,20 @@ public ECSCredentialsConfig ecsCredentialsConfig() {
...
+            String namingStrategy =
+                StringUtils.firstNonBlank(
+                    ecsAccount.getNamingStrategy(),
+                    ecsCredentialsConfig.getDefaultNamingStrategy());
+            if (namingStrategy != null) {
+              Namer<EcsResource> namer = namerRegistry.getNamingStrategy(namingStrategy);
+              NamerRegistry.lookup()
+                  .withProvider(EcsCloudProvider.ID)
+                  .withAccount(ecsAccount.getName())
+                  .setNamer(EcsResource.class, namer);
```
piradeepk, line 98: Did you mean to store this and use it later?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/names/EcsServerGroupNameResolver.java
```diff
@@ -0,0 +1,142 @@
...
+        new DescribeServicesRequest().withCluster(ecsClusterName).withServices(newServiceName);
+    DescribeServicesResult result = ecs.describeServices(request);
+
+    // an active or draining ECS service with this name was not found
+    return result.getServices().isEmpty()
+        || result.getServices().get(0).getStatus().equals("INACTIVE");
+  }
+
+  private List<String> listAllServices(String ecsClusterName) {
+    List<String> allServices = new ArrayList<>();
```
piradeepk, line 125: nit: can we make this a set instead of a list (to remove duplicates)? `Set<String> allServices = new HashSet<String>();`

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/names/EcsServerGroupNameResolver.java
```diff
@@ -0,0 +1,142 @@
...
+  public EcsServerGroupNameResolver(
+      String ecsClusterName, AmazonECS ecs, String region, Namer<EcsResource> namer) {
+    this.ecsClusterName = ecsClusterName;
+    this.ecs = ecs;
+    this.region = region;
+    this.naming = namer;
+  }
+
+  public EcsServerGroupName resolveNextName(String application, String stack, String detail) {
+    Moniker m = Moniker.builder().app(application).detail(detail).stack(stack).build();
```
piradeepk, line 48: nit: can you rename the variable to something more descriptive than `m` possibly something like `moniker`?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/names/EcsServerGroupNameResolver.java
```diff
@@ -0,0 +1,142 @@
...
+      DescribeServicesRequest request =
+          new DescribeServicesRequest()
+              .withCluster(ecsClusterName)
+              .withServices(serviceBatch)
+              .withInclude("TAGS");
+      DescribeServicesResult result = ecs.describeServices(request);
+      for (Service service : result.getServices()) {
+        Moniker moniker = naming.deriveMoniker(new EcsResourceService(service));
+
+        if (Objects.equals(currentName.getApp(), moniker.getApp())
```
piradeepk, line 71: If this is a string, might make more sense to use StringUtils.equals instead of Objects.equals

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -507,12 +503,44 @@ protected CreateServiceRequest makeServiceRequest(
             .withServiceRegistries(serviceRegistries)
             .withDeploymentConfiguration(deploymentConfiguration);
 
+    List<Tag> taskDefTags = new LinkedList<>();
```
piradeepk, line 506: nit: can you make this a set to avoid adding duplicate tags

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -507,12 +503,44 @@ protected CreateServiceRequest makeServiceRequest(
...
+    if (taggingEnabled) {
+      request
+          .withTags(taskDefTags)
+          .withEnableECSManagedTags(true)
+          .withPropagateTags(PropagateTags.SERVICE.toString());
+    } else {
+      if (!taskDefTags.isEmpty()) {
+        throw new IllegalArgumentException(
+            "Tagging is not enabled on ECS account " + description.getAccount() + ".");
+      }
```
piradeepk, line 543: It might make sense to do this prior to creating the request (fail fast).

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -507,12 +503,44 @@ protected CreateServiceRequest makeServiceRequest(
             .withServiceRegistries(serviceRegistries)
             .withDeploymentConfiguration(deploymentConfiguration);
 
+    List<Tag> taskDefTags = new LinkedList<>();
     if (description.getTags() != null && !description.getTags().isEmpty()) {
-      Collection<Tag> taskDefTags = new LinkedList<>();
       for (Map.Entry<String, String> entry : description.getTags().entrySet()) {
         taskDefTags.add(new Tag().withKey(entry.getKey()).withValue(entry.getValue()));
       }
-      request.withTags(taskDefTags).withEnableECSManagedTags(true).withPropagateTags("SERVICE");
```
piradeepk, line 515: Might make sense to leave this logic here (easier to follow)

piradeepk:
> > In the future, for a substantial feature, such as Moniker support, we would really appreciate it if you could submit a design proposal as an issue as it'll allow us to work through the intent and design the implementation prior to writing any logic.
> 
> I'm happy to write up a design document based on my findings from this PR. Originally I worked on this to see if it was even possible within the existing design and figured submitting it was a way to start the discussion on whether this is the right approach.

Thank you so much for providing all of the screenshots, testing the different scenarios manually, adding all of the unit/integ tests, and addressing the feedback so quickly!!! 💯 

At this point, I don't think it really adds much value for you to spend time writing up a proposal (if you'd like to write up something quick, feel free!), it was more around something to consider when looking into future features. It looks like it's really close! There are a couple conflicts that need to be resolved.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/names/EcsServerGroupNameResolver.java
deverton: Maybe, but the calling method relies on being able to chunk the list in to lots of 10 to optimise the calls to `ecs:DescribeServices` efficiently. Changing the return type to a `Set` would make that more complicated. Are you aware of any circumstances where AWS would return duplicate service names from `ecs:ListServices`?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/names/EcsServerGroupNameResolver.java
deverton: The contract between `Objects.equals(String, String)` and `StringUtils.equals(String, String)` is the same as they both end up just calling `String.equals()` but it's probably clearer to use the `StringUtils` one here.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/names/EcsServerGroupNameResolver.java
piradeepk: It shouldn't but was thinking about it from the perspective that it could possibly happen. That being said, agree with you here, not sure if the effort is necessary.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/names/EcsServerGroupNameResolver.java
piradeepk: Agreed, but it is also possible that the hashcode is different, causing the equals operator on two string objects to return false (even if both strings are identical)

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
deverton: The AWS API expects a list of `Tag` objects so changing this to a set increases the complexity. The `CreateServerGroupDescription.tags` field is already a `Map<String, String>` so it's not possible to pass in tags with duplicate keys. I've added a test to verify this.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
deverton: It has to be checked after the call to `namer.applyMoniker()` as that may have added tags even if the original operation description didn't include tags.

I don't think there's any large cost to checking after the `CreateServiceRequest` is created. The only thing I can see in the code that might be affected is the call to describe load balancers to attach. That should be moved after the tag setting check.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/security/EcsCredentialsInitializer.java
deverton: The `NamerRegistry` is essentially a singleton within the `clouddriver` context. This call is here to register the configure naming strategy for the given account.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
piradeepk: It's fairly straightforward in that you could do `Set<String> hSet = new HashSet<String>(taskDefTags);`, but that being said, I don't see the need at the moment.

deverton:
Okay, I think that's all the feedback addressed? Let me know if I've missed anything.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -497,24 +490,57 @@ protected CreateServiceRequest makeServiceRequest(
...
+    // Only add tags if they're set as it's an optional feature for ECS
+    if (taggingEnabled) {
+      request
+          .withTags(taskDefTags)
+          .withEnableECSManagedTags(true)
+          .withPropagateTags(PropagateTags.SERVICE.toString());
+    } else {
+      if (!taskDefTags.isEmpty()) {
+        throw new IllegalArgumentException(
+            "Tagging is not enabled on ECS account " + description.getAccount() + ".");
```
piradeepk, line 538: Can you make this exception a bit more descriptive? Users might not be aware that they can't tag without enabling the new (long) arn format.

piradeepk:
> Okay, I think that's all the feedback addressed? Let me know if I've missed anything.
> 
> This is really going to need to be squash merged 😄

Sounds great! I'll do a final pass tomorrow morning! Thanks for all of your hard work! 💯

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
deverton: > ECS account settings for account Test do not allow tagging as `serviceLongArnFormat` and `taskLongArnFormat` are not enabled.

How about that?

deverton:
Just remembered the `defaultNamingStrategy` setting doesn't work after the account loader changes. I'll see if I can fix that as well today.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
piradeepk: Looks great! Thanks! 💯

deverton:
Should be done now.

piradeepk (commented):
@deverton would using an existing template, without the moniker properties defined successfully allow me to deploy a service (there's some validation logic that's been removed, so could it possibly fail)?

deverton:
That's a fair point. I'm pretty sure the UI flow always adds the moniker property https://github.com/spinnaker/deck/blob/master/app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts#L573 but it's possible there will be configurations that are missing it.

I'll update the tests and logic.

clouddriver-ecs/src/integration/resources/testoperations/createServerGroup-artifact-EC2-targetGroupMappings.json
```diff
@@ -19,6 +19,11 @@
   "ecsClusterName": "integArtifactEC2TgMappings-cluster",
   "launchType": "EC2",
   "networkMode": "bridge",
+  "moniker": {
+    "app": "ecs",
+    "stack": "integArtifactsEC2TgMappingsStack",
+    "detail": "detailTest"
+  },
```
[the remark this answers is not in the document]
deverton: I'm pretty sure most of the UI flow in Deck will add `moniker` as a property. It doesn't mean that the tags will be applied though as that's triggered by the configuration.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/validators/EcsCreateServerGroupDescriptionValidator.java
```diff
@@ -101,8 +101,8 @@ public void validate(List priorDescriptions, Object description, ValidationError
       rejectValue(errors, "placementStrategySequence", "not.nullable");
     }
 
-    if (createServerGroupDescription.getApplication() == null) {
-      rejectValue(errors, "application", "not.nullable");
+    if (createServerGroupDescription.getMoniker().getApp() == null) {
+      rejectValue(errors, "moniker.app", "not.nullable");
     }
```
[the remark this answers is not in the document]
deverton: As mentioned, the moniker property is added already by Deck but it's worth handling the case where it's missing.

piradeepk:
Agreed, but not all users will be using the UI. Some users use pipeline templates, the SpinCLI, etc.

deverton:
Yeah, the code should be fully backwards compatible now with the integration tests in place to confirm that.

piradeepk (approved):
Ship it! 💯  Thanks for pushing this forward and getting this in @deverton!

allisaurus:
@piradeepk @deverton thank you both for this substantial contribution and review! 
I will put a second (less invasive) pair of eyes on this in the very near future so we can work on getting this merged prior to 1.24

allisaurus (commented):
This is a super exciting feature, @deverton ! I haven't quite yet reviewed everything, but added 2 initial questions and 1 comment re: correcting the license attribution so far. Will take a final look later today and any remaining thoughts.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -169,8 +150,20 @@ public DeploymentResult operate(List priorOutputs) {
     return makeDeploymentResult(service);
   }
 
+  private EcsServerGroupName buildEcsServerGroupName(AmazonECS ecs, Namer<EcsResource> namer) {
+    EcsServerGroupNameResolver serverGroupNameResolver =
+        new EcsServerGroupNameResolver(description.getEcsClusterName(), ecs, getRegion(), namer);
+
+    if (description.getMoniker() != null) {
```
allisaurus, line 157: this seems to be resolving the next SG name based on the presence of a `Moniker` in the description vs. the account-level naming strategy. Is it possible (esp if the request is hitting `gate` directly vs. via `deck`) a moniker could be present but differ from the app/stack/detail settings also provided in the description (in which case we would be naming this incorrectly)?

piradeepk:
@deverton based on the number of commits, would you mind rebasing all of your changes into a single commit before addressing @allisaurus feedback (makes it easier for us to see the diff)?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/names/EcsDefaultNamer.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright 2020 Netflix, Inc.
```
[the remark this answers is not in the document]
deverton: This should be fixed in the rebased commit.

clouddriver-ecs/src/integration/resources/clouddriver.yml
```diff
@@ -24,10 +24,16 @@ aws:
 ecs:
   enabled: true
   primaryAccount: ecs-account
+  defaultNamingStrategy: tags
```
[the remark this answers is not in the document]
deverton: There's two naming strategies included in this PR `default` (see `ECSDefaultNamer`) and `tags` (see `ECSTagNamer`). If there's no value for `defaultNamingStrategy` it will have the value of `default` which is set at a field level in `ECSCredentialsConfig`.  This will then be used in the `EcsCredentialsParser` naming strategy lookup.

The lookup will check for a non-blank account level value in `namingStrategy` and then use the value from the provider level `defaultNamingStrategy`

If the value for `defaultNamingStrategy` or `namingStrategy` are blank or otherwise invalid (i.e. don't map to a registered `Namer`) then the account parsing will fail with the following exception:

```Caused by: java.lang.IllegalArgumentException: Could not find naming strategy 'invalid'
	at com.netflix.spinnaker.clouddriver.names.NamerRegistry.lambda$getNamingStrategy$1(NamerRegistry.java:58)
	at java.base/java.util.Optional.orElseThrow(Optional.java:408)
	at com.netflix.spinnaker.clouddriver.names.NamerRegistry.getNamingStrategy(NamerRegistry.java:56)
	at com.netflix.spinnaker.clouddriver.ecs.security.EcsCredentialsParser.parse(EcsCredentialsParser.java:70)
```

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
deverton: Good point. There's nothing validating that the values must match between the current name fields and the `Moniker` fields. We could update the `EcsCreateServerGroupDescriptionValidator` to verify that the values match if they're set and reject the request.

deverton: Though one thing that helps here is that the `EcsTagNamer` will first parse the service name using the Frigga naming convention, and then overwrite the values for stack, detail, sequence etc with values from the tags if present.

deverton: So without validation the scenario is something like a `CreateServerGroupDescription` with values like

```json
{
  "application": "foo",
  "moniker": {
    "app": "bar"
  }
}
```

| Naming Strategy | Current Service Name | New Service Name |
| - | - | - |
| `default` | | `bar-v001` |
| `tags` | | `bar-v001` |
| `default` | `foo-v001` | `bar-v001` |
| `tags` | `foo-v001` | `bar-v001` |
| `default` | `bar-v001` | `bar-v002` |
| `tags` | `bar-v001` | `bar-v002` |

So I think adding validation would be the best approach as it would prevent a mismatch in outcomes.

piradeepk: Would it make more sense to have validation that says both properties cannot be defined? Rather than validating that they're the same?

allisaurus: @deverton validating that they match is good, though would it matter as much if we were sure which was the correct strategy to use? Is the account's naming strategy available here and could we use it to make a decision?

@piradeepk I believe deck will populate `moniker` values by default when you specify app/stack/detail so enforcing the presence of only one or the other would likely break a bunch of existing pipelines.

deverton: Maybe I'm misunderstanding, but should the source of the data change the naming strategy though? As in, if values are set in the `moniker` object, that doesn't mean the code should use tags. And vice versa, just because tags are used as the source of truth for the name, doesn't mean the deployment request needs to use the `moniker` object. The `application`, `freeFormDetails`, and `stack` fields still work.

allisaurus: > just because tags are used as the source of truth for the name, doesn't mean the deployment request needs to use the moniker object. The application, freeFormDetails, and stack fields still work.

ok, I did not fully understand this. So if the idea is that these fields should match (enforced through your new validation) and be correctly set during the deploy operation based on lookup of the existing services, then I am not concerned. Your chart of the diff name results based on mismatched input is exactly what I was trying to avoid and it seems we are now protected against it. Thanks for the additional context!

deverton: Hm, that suggests maybe this PR should be split in to two. One part to handle the addition of tag based name storage, and one to change the deployment request object to prefer the `moniker` structure over the top-level `application`, `freeFormDetails`, and `stack` fields.

deverton: I'm not likely to get time to address this until next week at the earliest. Would you rather I split the patch up or are you happy to accept this as is?

allisaurus (approved):

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
piradeepk: Thanks @deverton for all your hard work! Based on the limited time available, we'd like to see this feature go into the next release, as many in the community would benefit from this! 😃  I've added the merge label

PR #4938, opened 2020-09-25
fix(titus): Update error handling

aravindmd:
Throwing an exception for deck to read the better error message.


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/ops/AbstractEnableDisableAtomicOperation.groovy
```diff
@@ -83,9 +83,7 @@ abstract class AbstractEnableDisableAtomicOperation implements AtomicOperation<V
       if (disable && description.desiredPercentage && loadBalancingClient && job.labels.containsKey("spinnaker.targetGroups")) {
         def errorMessage = "Could not ${verb} ServerGroup '$serverGroupName' in region $region! " +
           "Disabling by percentage for server groups with target groups is not supported by Titus"
-        log.error(errorMessage)
-        task.updateStatus phaseName, errorMessage
-        return false
+        throw new IllegalArgumentException(errorMessage)
```
marchello2000, line 86: nit: maybe 
```suggestion
        throw new UnsupportedOperationException(errorMessage)
```
instead

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/ops/AbstractEnableDisableAtomicOperation.groovy
```diff
@@ -83,9 +83,7 @@ abstract class AbstractEnableDisableAtomicOperation implements AtomicOperation<V
       if (disable && description.desiredPercentage && loadBalancingClient && job.labels.containsKey("spinnaker.targetGroups")) {
         def errorMessage = "Could not ${verb} ServerGroup '$serverGroupName' in region $region! " +
           "Disabling by percentage for server groups with target groups is not supported by Titus"
-        log.error(errorMessage)
-        task.updateStatus phaseName, errorMessage
-        return false
+        throw new UnsupportedOperationException(errorMessage)
```
robzienert, line 86: Even more bester, having the exception be part of the `SpinnakerException` family would be great.

aravindmd: I can take a look into that.

aravindmd: looked at `kork-exceptions` and figured i can just use the `SystemException` for this case, let me know if you think adding a new type of exception is required vs this change ?

robzienert: It's an `IntegrationException`. You can probably just use [`TitusException`](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/TitusException.java) directly.

A `SystemException` is for core service code, so, if something in the code that enables atomic operations to work broke, that would be a `SystemException`, but since Titus is just an integration, its exceptions will be of the `IntegrationException` family.

robzienert: Curious to know if I should update the docs on these exception types if it needs more clarity or examples?

aravindmd: No i think its adequate, i think i missed looking at the `IntegrationException` , i will update this.

robzienert (approved):

PR #4946, opened 2020-09-25
feat(k8s/runJob): add pod status and detail collection limits

kevinawoo:
What happens?
When running kubernetes jobs that set a large `completions` value, for example 500+, this will take down Orca due to growing pipeline context size. 💥

How does this happen?
Clouddriver returns the pod status for each pod and puts the result in `$stage(...).outputs.jobStatus.pods`. Since it's part of the pipeline context now, it'll get dragged across stages and potentially parent/child executions too. 

The fix:
Add an adjustable limit to how many `jobs` and `jobStatus`s to store in the pipeline. The limit allows users to use SPeL. The defaults for `job`s and `jobStatus`s are set to something reasonable. My hunch is users will not likely try to use SPeL expressions when they have more `job`s or `jobStatus`s since it'll be a nightmare to count.


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.java
```diff
@@ -55,6 +55,8 @@
     private List<String> omitKinds = new ArrayList<>();
     private boolean onlySpinnakerManaged = false;
     private Long cacheIntervalSeconds;
+    private Integer runJobPodCollectionLimit = 5;
+    private Integer runJobPodStatusCollectionLimit = 10;
```
ezimanyi, line 59: If these can never be null (which appears to be the case), they should be `int`s.  (If they ever were null we'd get an NPE when doing comparisons such as `pods.size() <= credentials.getRunJobPodCollectionLimit()`; by choosing the correct data type we can eliminate that failure mode.)

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesJobProvider.java
```diff
@@ -73,14 +73,17 @@ public KubernetesJobStatus collectJob(String account, String location, String id
...
 
-    jobStatus.setPods(
-        pods.stream()
-            .map(
-                p -> {
-                  V1Pod pod = KubernetesCacheDataConverter.getResource(p, V1Pod.class);
-                  return new KubernetesJobStatus.PodStatus(pod);
-                })
-            .collect(Collectors.toList()));
+    if (pods.size() <= credentials.getRunJobPodCollectionLimit()) {
```
ezimanyi, line 76: This is a bit confusing to me.

What is the use case for needing separate limits on the number of pods to collect, and the number of statuses to collect?   It feels like this is going to be complex for users to configure (adding to already complex configuration for kubernetes accounts).

Even with the default values, it's not clear to me what's happening.  If you have 5 pods, we'll collect them all with a status.  But if you have 6 pods, we won't collect any of them because we skip this block entirely.  (So what is `runJobPodStatusCollectionLimit = 10` controlling then?)

I think we can solve the problem with only a single configurable parameter controlling the number of jobs to collect (without the separate parameter for status), which would make it much easier for users to understand.

ezimanyi (commented):
As noted in the inline comment, I think this can be simplified a bit while still solving the problem.  Also, I'd like to see some test coverage on the added functionality.

Is there a reason we're passing the statuses back at all?  Is that just to support SpEL?  By changing the defaults like this we will likely break users using SpEL anyway if their pipelines have more than the default number of jobs.

Finally---given that we store entire manifests it the pipeline context (and people often deploy dozens of these at a time), I'm surprised that only ~50 tiny JSON objects with name/status are causing this issue.  In the cases you looked into were there also manifests in the pipeline context?  What percentage of the total context size did these represent?

german-muzquiz:
Closed in favor of https://github.com/spinnaker/clouddriver/pull/5360

PR #4947, opened 2020-09-25
fix(titus): Improve handling for a non-existent Titus Job

ajordens:
Rather than fail the action, no-op the delete and return an empty `DeleteServerGroupEvent`.

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/DestroyTitusJob.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/DestroyTitusJob.java
@@ -31,6 +31,8 @@ import com.netflix.spinnaker.clouddriver.titus.client.model.Job;
 import com.netflix.spinnaker.clouddriver.titus.client.model.TerminateJobRequest;
 import com.netflix.spinnaker.clouddriver.titus.credentials.NetflixTitusCredentials;
 import com.netflix.spinnaker.clouddriver.titus.deploy.description.DestroyTitusJobDescription;
+import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
 import javax.annotation.Nonnull;
 import lombok.Builder;
 import lombok.Value;
@@ -68,7 +70,7 @@ public class DestroyTitusJob implements SagaAction<DestroyTitusJob.DestroyTitusJ
         titusClientProvider.getTitusClient(
             (NetflixTitusCredentials) accountCredentials, command.description.getRegion());
 
-    Job job = titusClient.getJobAndAllRunningAndCompletedTasks(command.description.getJobId());
+    Job job = fetchJob(titusClient, command.description.getJobId());
     if (job != null) {
       titusClient.terminateJob(
           (TerminateJobRequest)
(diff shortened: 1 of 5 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/DestroyTitusJob.java
```diff
@@ -68,7 +70,7 @@ public Result apply(@NotNull DestroyTitusJob.DestroyTitusJobCommand command, @No
         titusClientProvider.getTitusClient(
             (NetflixTitusCredentials) accountCredentials, command.description.getRegion());
 
-    Job job = titusClient.getJobAndAllRunningAndCompletedTasks(command.description.getJobId());
+    Job job = fetchJob(titusClient, command.description.getJobId());
```
ajordens, line 73: The titus client has differing behaviors depending on method called. 

Some will return null, some will raise the `StatusRuntimeException`. 

It's unlikely to happen all that often in practice but this was a gap that existed before this was converted to a `SagaAction`.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/converters/DestroyTitusJobAtomicOperationConverter.groovy
```diff
@@ -58,6 +61,20 @@ class DestroyTitusJobAtomicOperationConverter extends AbstractAtomicOperationsCr
       converted.requiresApplicationRestriction = !converted.applications.isEmpty()
       converted.serverGroupName = job.name
     } catch (Exception e) {
+      if (e instanceof StatusRuntimeException) {
+        def statusRuntimeException = (StatusRuntimeException) e
+        if (statusRuntimeException.status.code == Status.NOT_FOUND.code) {
+          throw new NotFoundException(
```
ajordens, line 67: It seemed weird to just continue when we know up front that the job doesn't exist. That said, I have now made the downstream behavior handle a missing job.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/DestroyTitusJobCompletionHandler.java
```diff
@@ -22,25 +22,31 @@
...
 
 @Component
 public class DestroyTitusJobCompletionHandler
-    implements SagaCompletionHandler<DeleteServerGroupEvent> {
+    implements SagaCompletionHandler<Optional<DeleteServerGroupEvent>> {
 
   @Nullable
   @Override
-  public DeleteServerGroupEvent handle(@Nonnull Saga completedSaga) {
+  public Optional<DeleteServerGroupEvent> handle(@Nonnull Saga completedSaga) {
```
ajordens, line 36:
```
  private fun <T> invokeCompletionHandler(saga: Saga, flow: SagaFlow): T? {
    return flow.completionHandler
      ?.let { completionHandler ->
        val handler = applicationContext.getBean(completionHandler)
        val result = sagaRepository.get(saga.name, saga.id)
          ?.let { handler.handle(it) }
          ?: throw SagaNotFoundException("Could not find Saga to complete by ${saga.name}/${saga.id}")
```

The `SagaService` will raise `SagaNotFoundException` if a completion handler returns a null object.

That does seem suspect (and easily fixed) but it's also nice to indicate when a result object is optional.

ajordens:
I noticed this behavior when implementing the saga and had a task to revisit. 

It's rather small and inconsequential.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/handlers/DestroyTitusJobCompletionHandler.java
robzienert: 🤔  Definitely a bug.

robzienert (approved):

PR #4954, opened 2020-09-28
fix(titus): simplify retry logic in SubmitTitusJob

dreynaud:
* avoid the (probably accidental) bitwise xor in Thread.sleep
* in fact, don't call Thread.sleep at all, this is in a retry block
* we don't need no retryCount, retries are handled by RetrySupport
* rely on SpinnakerException with setRetryable(false) to stop retries


robzienert (approved):
excellent!

jeyrschabu (approved):
More betta

aravindmd (approved):

PR #4959, opened 2020-09-29
feat(cloudfoundry): Migrate to kork-credentials

ncknt:
This is the implementation of https://github.com/spinnaker/governance/blob/master/rfc/account-management.md for the CloudFoundry provider.
 
- Use `CredentialsRepository`: this lets us be type safe and use a repository dedicated to CloudFoundry accounts. We can now define new credential sources or hook into the lifecycle of a credentials.
- `CredentialsInitializerSynchronizable` is now just running a `com.netflix.spinnaker.credentials.poller.Poller`.
- Use `BaseProvider` and avoid static calls to ProviderUtils (groovy)
- Test to not throw NPE when account does not exist in service provider.


ncknt:
@zachsmith1 could you PTAL?

zachsmith1:
LGTM!

zachsmith1 (approved):

kevinawoo (approved):

PR #4967, opened 2020-10-01
fix(provider/cf): only perform ondemand caching when account equals the credentials account

zachsmith1:
Small but very important fix. Without checking to see if the `agent.getAccountName() equals data.account`, every caching agent that matches the handle for provider `cloudfoundry` and type `OnDemandType.ServerGroup or OnDemandType.LoadBalancer`, will proceed and create false evictions and potentially other harmful data inconsistencies. This fix will remove those data inconsistencies from happening and in theory, greatly speed up the Force Cache Refresh in the deploy stages for serverGroups and loadBalancers. Instead of N-CloudFoundry agents all processing onDemandCache and making subsequent API calls, only a single agent will now run to process the cache.

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/provider/agent/CloudFoundryLoadBalancerCachingAgent.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/provider/agent/CloudFoundryLoadBalancerCachingAgent.java
@@ -165,6 +165,11 @@ public class CloudFoundryLoadBalancerCachingAgent extends AbstractCloudFoundryCa
     if (account == null || region == null || loadBalancerName == null) {
       return null;
     }
+
+    if (!this.getAccountName().equals(account)) {
+      return null;
+    }
+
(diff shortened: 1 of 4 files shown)


zachsmith1:
@kevinawoo @ncknt plz review and backport to 1.22, 1.21 if possible. Tests included.

ncknt:
LGTM!

kevinawoo (approved):

zachsmith1:
@Mergifyio backport 1.22.x

zachsmith1:

german-muzquiz:
@zachsmith1 did this work before and a regression was introduced, or this has been the usual behavior? Although this seems an obvious bug, only PRs fixing regressions that comply with the [backport rules](https://spinnaker.io/community/contributing/releasing/#release-branch-patch-criteria) can be backported.

zachsmith1:
@german-muzquiz It seems a regression was not introduced but yes, it is an obvious bug. Are they any exceptions that can be made? This is one that I would advocate for because its consequences could introduce data corruption or other nasty things. Plus, this only touches CF so I would think we should have a little more say to make exceptions for things like this.

german-muzquiz:
Given that this has been the usual behavior so far and there are no known reported issues, I think this cannot be backported.

PR #4970, opened 2020-10-01
fix (aws/lambda) : Send back results from invoke lambda atomic operation

shyamsfo:
fix (aws/lambda) : Send back results from invoke lambda atomic operation


kevinawoo (approved):
LGTM, talked to @shyamsfo offline and tests are coming for lambda in a separate PR

PR #4972, opened 2020-10-02
feat(test): extend ECS CreateServerGroup integ tests

allisaurus:
Previously, these tests would succeed after successful submission of an ECS createServerGroup task. This change adds further checks, including:
* checking the task status after creation to ensure the ECS Service has been created
* ensures the expected ECS services calls were made
* ensures the content of the task definition and service reflect expected values given test input.

### DIFF
--- a/clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/EcsSpec.java
+++ b/clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/EcsSpec.java
@@ -15,7 +15,8 @@
 
 package com.netflix.spinnaker.clouddriver.ecs;
 
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.*;
 
@@ -32,18 +33,16 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.*;
-import org.junit.Test;
+import java.util.function.BooleanSupplier;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayName;
-import org.junit.runner.RunWith;
+import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.boot.web.server.LocalServerPort;
 import org.springframework.test.context.TestPropertySource;
-import org.springframework.test.context.junit4.SpringRunner;
 
-@RunWith(SpringRunner.class)
 @SpringBootTest(
     classes = {Main.class},
     webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
@@ -62,7 +61,7 @@ public class EcsSpec {
 
   @LocalServerPort private int port;
 
-  @MockBean AmazonClientProvider mockAwsProvider;
+  @MockBean protected AmazonClientProvider mockAwsProvider;
 
   @MockBean AmazonAccountsSynchronizer mockAccountsSyncer;
 
@@ -102,5 +101,17 @@ public class EcsSpec {
     dataMap.put(namespace, dataPoints);
 
     return new DefaultCacheResult(dataMap);
+  }
+
+  protected void retryUntilTrue(BooleanSupplier func, String failMsg, int retrySeconds)
+      throws InterruptedException {
+    for (int i = 0; i < retrySeconds; i++) {
+      if (!func.getAsBoolean()) {
+        Thread.sleep(1000);
+      } else {
+        return;
+      }
+    }
+    fail(failMsg);
   }
 }
--- a/clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
+++ b/clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
@@ -15,31 +15,95 @@
 
 package com.netflix.spinnaker.clouddriver.ecs.test;
 
+import static io.restassured.RestAssured.get;
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.notNullValue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.*;
 
+import com.amazonaws.services.ecs.AmazonECS;
+import com.amazonaws.services.ecs.model.*;
+import com.amazonaws.services.elasticloadbalancingv2.AmazonElasticLoadBalancing;
+import com.amazonaws.services.elasticloadbalancingv2.model.DescribeTargetGroupsRequest;
+import com.amazonaws.services.elasticloadbalancingv2.model.DescribeTargetGroupsResult;
+import com.amazonaws.services.elasticloadbalancingv2.model.TargetGroup;
 import com.netflix.spinnaker.clouddriver.aws.security.AmazonCredentials;
 import com.netflix.spinnaker.clouddriver.aws.security.NetflixAmazonCredentials;
+import com.netflix.spinnaker.clouddriver.aws.security.NetflixAssumeRoleAmazonCredentials;
 import com.netflix.spinnaker.clouddriver.ecs.EcsSpec;
 import com.netflix.spinnaker.clouddriver.security.AccountCredentialsRepository;
 import io.restassured.http.ContentType;
 import java.io.IOException;
 import java.util.Collections;
-import org.junit.Test;
+import java.util.HashMap;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+import org.mockito.stubbing.Answer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 
-/**
- * TODO (allisaurus): Ideally these tests would go further and actually assert that the resulting
- * ecs:create-service call is formed as expected, but for now, it asserts that the given operation
- * is correctly validated and submitted as a task.
- */
 public class CreateServerGroupSpec extends EcsSpec {
+
+  private final Logger log = LoggerFactory.getLogger(getClass());
 
   @Autowired AccountCredentialsRepository accountCredentialsRepository;
 
+  private final int TASK_RETRY_SECONDS = 3;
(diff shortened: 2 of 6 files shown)


clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/EcsSpec.java
```diff
@@ -62,7 +61,7 @@
 
   @LocalServerPort private int port;
 
-  @MockBean AmazonClientProvider mockAwsProvider;
+  @MockBean protected AmazonClientProvider mockAwsProvider;
```
pkandasamy91, line 64: Curios, why are we reducing the scope here? Would it make sense to also reduce the scope of all public mocked beans?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
```diff
@@ -15,57 +15,167 @@
...
+            .post(url)
+            .then()
+            .statusCode(200)
+            .contentType(ContentType.JSON)
+            .body("id", notNullValue())
+            .body("resourceUri", containsString("/task/"))
+            .extract()
+            .path("id");
+
+    retryUntilTrue(
```
pkandasamy91, line 138: Do we need retry logic in tests, you wouldn't be making calls outside the scope of clouddriver, so should any call take too long?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/EcsControllersSpec.java
```diff
@@ -30,8 +30,8 @@
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
-import org.junit.Test;
 import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
```
pkandasamy91, line 34: Is this intentional?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
```diff
@@ -15,57 +15,167 @@
...
 
+import static io.restassured.RestAssured.get;
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.notNullValue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.*;
 
+import com.amazonaws.services.ecs.AmazonECS;
+import com.amazonaws.services.ecs.model.*;
```
pkandasamy91, line 26: nit: Can we add specific imports here instead of *?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
```diff
@@ -15,57 +15,167 @@
 
 package com.netflix.spinnaker.clouddriver.ecs.test;
 
+import static io.restassured.RestAssured.get;
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.notNullValue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.*;
```
pkandasamy91, line 23: nit: Can we add specific imports here instead of *?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
allisaurus: for context: these tests run against a live/running instance of clouddriver, which begins execution of the `createServerGroupAtomicOperation` workflow _after_ successful submission of the task. 

Asserting against the mocks directly after submitting the operation led to errors b/c often the mocked calls had not yet been invoked. It's still pretty fast, so adding a brief `Thread.sleep()` before verifying could potentially solve this as well (it did during testing), but by retrying on the provided task ID we're more closely mocking how `Orca` would poll for task completion  + implicitly testing that the provided ID matches the intended operation.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/EcsControllersSpec.java
allisaurus: Yes! ToC [guidance](https://github.com/spinnaker/clouddriver/pull/4827#issuecomment-678526751) is to use Junit5 for these tests. See [K8s integ tests](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeployManifestIT.java#L33-L34) for and existing example.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/EcsSpec.java
allisaurus: AFAIK the default scope is package-private, so this is _increasing_ scope to allow the tests in `CreateServerGroupSpec` to setup mocked behavior against this bean. Maybe it would be better to explicitly set the other bean in this package to `private` to make this distinction more explicit?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
allisaurus: this was done automatically by intellij (default is to import * after 5 from the same package). I've got my IDE configured [as instructed](https://github.com/spinnaker/clouddriver#developing-with-intellij) for this package so I'm inclined to leave it unless there's some competing community/Java guidance I'm unaware of?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
allisaurus: see previous comment

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
pkandasamy91: Makes sense, I'm just hesitant in adding additional retry logic that's tightly coupled to another service (orca in this case) as that can change.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
pkandasamy91: Usually it's best to explicitly define the different imports (so that there aren't any accidental usages of the wrong import). That being said, feel free to choose which ever one that works for you.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/EcsSpec.java
pkandasamy91: Ah that's fair (I just assumed that anything that extended this was in the same package, but it's not necessarily true).

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
allisaurus: well to be clear, we're not mocking anything Orca-specific here, it's just that the operation is asynchronous so anything submitting one (aka Orca) would have to query the task ID later to figure out if it succeeded or not.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
```diff
@@ -15,57 +15,167 @@
...
 import org.springframework.beans.factory.annotation.Autowired;
 
-/**
- * TODO (allisaurus): Ideally these tests would go further and actually assert that the resulting
- * ecs:create-service call is formed as expected, but for now, it asserts that the given operation
- * is correctly validated and submitted as a task.
- */
 public class CreateServerGroupSpec extends EcsSpec {
 
+  private final Logger log = LoggerFactory.getLogger(getClass());
```
paragbhingre, line 53: Is it okay to keep the log statements in the committed code? 
Or just keeping it for the tracking purpose?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
```diff
@@ -15,57 +15,167 @@
...
+
   @DisplayName(
       ".\n===\n"
           + "Given description w/ inputs, EC2 launch type, and legacy target group fields, "
           + "successfully submit createServerGroup operation"
           + "\n===")
   @Test
-  public void createServerGroupOperationTest() throws IOException {
+  public void createServerGroup_InputsEc2LegacyTargetGroupTest()
+      throws IOException, InterruptedException {
```
paragbhingre, line 115: Where is the possibility of this InterruptedException to occur?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
allisaurus: AHH this should have been removed, will update.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
allisaurus: `InterruptedException` may be thrown by the call to `Thread.sleep()` in `retryUntilTrue` (called in this method) in the event that the thread is interrupted while the `sleep` is active. Some more context is available in the [docs for sleep](https://docs.oracle.com/javase/tutorial/essential/concurrency/sleep.html). 

FWIW we're not defining or expecting any thread interruptions here, but bubbling it up here as part of the method sig of `sleep`.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
pkandasamy91: Since the retry count is 3, I'm okay with this (I'm hesitant in hardcoding retry logic that could potentially get the tests into a weird state).

pkandasamy91 (approved):
Lgtm! Thanks for adding these @allisaurus!

PR #4984, opened 2020-10-07
fix(provider/aws): encode security group names

deverton:
Security group names in AWS can contain `:` which breaks the cache key parsing. This patch adds a simple encode/decode step
to the security group name using percent encoding when it encounters either `:` or `%`.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/cache/Keys.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/cache/Keys.groovy
@@ -88,8 +88,9 @@ class Keys implements KeyParser {
 
     switch (result.type) {
       case Namespace.SECURITY_GROUPS.ns:
-        def names = Names.parseName(parts[2])
-        result << [application: names.app, name: parts[2], id: parts[3], region: parts[4], account: parts[5], vpcId: parts[6] == "null" ? null : parts[6]]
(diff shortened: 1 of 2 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/cache/Keys.groovy
```diff
@@ -120,12 +121,21 @@ class Keys implements KeyParser {
     result
   }
 
+  static String encode(String value) {
+    return value.replace([':' : '%3A', '%' : '%25'])
+  }
+
+  static String decode(String value) {
+    return value.replace(['%3A' : ':', '%25' : '%'])
+  }
```
robzienert, line 130: Since this is a backwards-incompatible change, it needs to be put behind a feature flag that is disabled by default.

robzienert (changes requested):

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/cache/Keys.groovy
deverton: Good point. Though, while I'm looking at this, I'm wondering why the group name is even part of the key? The security group ID and the VPC ID should be enough to ensure uniqueness.

deverton:
Closing in favor of #5009

PR #4987, opened 2020-10-07
fix(cf): Build sg names using existing methods

german-muzquiz:
Reusing existing methods to build server group names.


zachsmith1 (approved):
LGTM!

kevinawoo (approved):

PR #4968, opened 2020-10-01
feat(aws): Adding property externalId to AWS account config for assuming roles.

ivanphdz:
Enabling the use of the externalId parameter when assuming a role prevents the "confused deputy" problem.

https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html#external-id-purpose

PR #4988, opened 2020-10-07
fix(aws): revert change that is breaking clouddriver on Jenkins

plumpy:
Reverts spinnaker/clouddriver#4968


fieldju:
Cant we just set the default env var for jenkins that is what we did for github actions

fieldju:
https://github.com/spinnaker/clouddriver/blob/master/build.gradle#L72

This needs to be set as an env var for the AWS sdk if you are not running on an AWS instance

plumpy:
Yeah, we can do that, but is this likely to bite customers too? Why wasn't this an issue before?

fieldju:
> 
> 
> Yeah, we can do that, but is this likely to bite customers too? Why wasn't this an issue before?

Because you don't have unit tests, we can delete the unit test and things will be happy :(

fieldju:
Basically if you instantiate an AWS client it needs to be configured for a region, were are doing that in the unit test.

plumpy:
Discussed this on Slack, @fieldju is going to send a small PR to fix the issue.

PR #4989, opened 2020-10-07
fix(aws): use the Spinnaker region provider with the sts client builder, so that it defaults to the default region, when no region is defined

fieldju:
fix(aws): use the Spinnaker region provider with the sts client builder, so that it defaults to the default region, when no region is defined


plumpy (approved):
Thanks for the quick fix!

PR #4990, opened 2020-10-07
feat(titus): Converted a few more Titus operations to sagas

ajordens:
* `detachInstances`
* `resizeServerGroup`
* `terminateInstances`

I opted to use Titus terminology when naming the actions (Tasks vs Instances and
Job vs Server Group).

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/DetachTitusTasks.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/DetachTitusTasks.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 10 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/DetachTitusTasks.java
```diff
@@ -0,0 +1,165 @@
...
+    AccountCredentials accountCredentials =
+        accountCredentialsProvider.getCredentials(command.description.getAccount());
+
+    TitusClient titusClient =
+        titusClientProvider.getTitusClient(
+            (NetflixTitusCredentials) accountCredentials, command.description.getRegion());
+
+    Job job = titusClient.findJobByName(command.description.getAsgName(), true);
+    if (job == null) {
+      saga.log("Job not found");
```
ajordens, line 85: This was pre-existing behavior. 

Seems potentially suspect to "successfully" complete an operation yet not actually do the work.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/ResizeTitusJob.java
```diff
@@ -0,0 +1,122 @@
...
+    AccountCredentials accountCredentials =
+        accountCredentialsProvider.getCredentials(command.description.getAccount());
+
+    TitusClient titusClient =
+        titusClientProvider.getTitusClient(
+            (NetflixTitusCredentials) accountCredentials, command.description.getRegion());
+
+    Job job = titusClient.findJobByName(command.description.getServerGroupName());
+    if (job == null) {
+      throw new IllegalArgumentException(
```
ajordens, line 73: Related to the previous comment, we're raising an exception here.

I should flip this to a `UserException`, I'll do the same in both cases.

ajordens:
These operations were all idempotent to begin with, so the conversion was straight forward.

cfieber (approved):

jeyrschabu (approved):

aravindmd (approved):

PR #4991, opened 2020-10-07
fix(provider/cf): fix regression caused by clusterSummaries

zachsmith1:
Fix a regression caused by clusterSummaries. ClusterSummaries should not fully hydrate the servergroups, hence needs to be changed to NONE for details. The underlying method that is ran got reverted to its original state until we can revisit this set of features at a later date. A test was added to ensure the functionality works as expected.


zachsmith1:

german-muzquiz (approved):

german-muzquiz:
This was the PR that introduced the regression: https://github.com/spinnaker/clouddriver/pull/4749 , so the only version affected is `1.22`

german-muzquiz:
@Mergifyio release-1.22.x

german-muzquiz:
@Mergifyio backport release-1.22.x

PR #4999, opened 2020-10-08
feat(test): extend ECS create server group tests for artifacts

paragbhingre:
This PR contains integration tests for Create Server Group Operation for the task definition type as artifacts.

There are total 2 tests:

- Create server group operation for the task definition type as artifacts, launch type as EC2 with target group mappings.

- Create server group operation for the task definition type as artifacts, launch type as FARGATE target group mappings.

This change also adds further checks for 2 tests mentioned above, including:

- checking the task status after creation to ensure the ECS Service has been created

- ensures the expected ECS services calls were made

- ensures the content of the task definition and service reflect expected values given test input.


clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -0,0 +1,311 @@
...
+            if (taskHistory
+                .toString()
+                .contains(String.format("Done creating 1 of %s", expectedServerGroupName))) {
+              return true;
+            }
+            return false;
+          },
+          String.format("Failed to detect service creation in %s seconds", TASK_RETRY_SECONDS),
+          TASK_RETRY_SECONDS);
+    } catch (InterruptedException e) {
```
allisaurus, line 284: I'd prefer we not catch this but instead let it bubble up/fail, since we can't proceed with validating the test if it occurs. 
(NOTE: did eclipse auto-generate this block for you? it can't always be trusted.)

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -0,0 +1,311 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.stubbing.Answer;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.mock.mockito.MockBean;
+
+/**
+ * TODO (allisaurus): Ideally these tests would go further and actually assert that the resulting
```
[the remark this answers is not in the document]
paragbhingre: Ohh yes.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -0,0 +1,311 @@
...
+    when(mockECS.listServices(any(ListServicesRequest.class))).thenReturn(new ListServicesResult());
+
+    when(mockECS.describeServices(any(DescribeServicesRequest.class)))
+        .thenReturn(new DescribeServicesResult());
+
+    when(mockAwsProvider.getAmazonApplicationAutoScaling(
+            any(NetflixAmazonCredentials.class), anyString(), anyBoolean()))
+        .thenReturn(mockAWSApplicationAutoScalingClient);
+
+    when(mockAWSApplicationAutoScalingClient.describeScalableTargets(
```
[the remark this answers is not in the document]
paragbhingre: Yes, if we don't mock this call then we get NPE in getSourceScalableTarget() of CreateServerGroupAtomicOperation as it checks for following condition.
```
if (result.getScalableTargets() != null && !result.getScalableTargets().isEmpty()) {
        return result.getScalableTargets().get(0);
      }
```

So if we don't mock this call then result will null and the test will fail.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
paragbhingre: It relates to my comment on your PR where I had asked why have you used `**throws InterruptedException**` for the tests. For me, eclipse auto-generated this try-catch but even I would also prefer using `** throws InterruptedException**` for my tests. I will make that change.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -0,0 +1,300 @@
...
+
+  private final int TASK_RETRY_SECONDS = 3;
+
+  private static final String CREATE_SG_TEST_PATH = "/ecs/ops/createServerGroup";
+
+  @BeforeEach
+  public void setup() {
+
+    // mocking calls
+    when(mockAwsProvider.getAmazonEcs(
```
allisaurus, line 81: we're mocking calls to several services/beans here, but they're not grouped together. Can we organize this func so that all the calls to a specific mock (e.g., `mockECS`) are grouped together, with calls to `mockAwsProvider` at the end? (since this is the mock bean which is returning all the other AWS service mocks). that way it's easier to see everything a specific service/mock needs to do, and its organized more topologically.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -0,0 +1,300 @@
...
+        .thenReturn(mockELB);
+  }
+
+  @DisplayName(
+      ".\n===\n"
+          + "Given description w/ task def artifacts, EC2  launch type, and new target group fields, "
+          + "successfully submit createServerGroup operation"
+          + "\n===")
+  @Test
+  public void createSGOWithArtifactsEC2TargetGroupMappingsTest()
```
allisaurus, line 143: can we rename the test functions in this file to match the [format of the other atomic operation tests](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java#L110)?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -0,0 +1,300 @@
...
+                });
+
+    when(mockAwsProvider.getAmazonElasticLoadBalancingV2(
+            any(NetflixAmazonCredentials.class), anyString(), anyBoolean()))
+        .thenReturn(mockELB);
+  }
+
+  @DisplayName(
+      ".\n===\n"
+          + "Given description w/ task def artifacts, EC2  launch type, and new target group fields, "
```
allisaurus, line 139: nit: extra while space between "EC2" and "launch" (same for other test display text)

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -0,0 +1,300 @@
...
+import org.mockito.ArgumentCaptor;
+import org.mockito.stubbing.Answer;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.mock.mockito.MockBean;
+
+public class CreateServerGroupWithArtifactsSpec extends EcsSpec {
+
+  @Autowired AccountCredentialsRepository accountCredentialsRepository;
+
+  @Autowired protected AmazonClientProvider mockAwsProvider;
```
allisaurus, line 59: I don't believe we need this, we can use the existing one in the parent `EcsSpec`

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -0,0 +1,300 @@
...
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.stubbing.Answer;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.mock.mockito.MockBean;
+
+public class CreateServerGroupWithArtifactsSpec extends EcsSpec {
+
+  @Autowired AccountCredentialsRepository accountCredentialsRepository;
```
allisaurus, line 57: this appears to be unused?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -0,0 +1,300 @@
...
+
+  @Autowired protected AmazonClientProvider mockAwsProvider;
+
+  @MockBean ArtifactDownloader artifactDownloader;
+
+  @MockBean AWSApplicationAutoScalingClient mockAWSApplicationAutoScalingClient;
+
+  @MockBean ArtifactCredentialsRepository mockArtifactCredentialsRepository;
+
+  @MockBean ArtifactCredentials mockArtifactCredentials;
```
allisaurus, line 67: similar to above, I think this can be a simple mock, not a `@MockBean`

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -0,0 +1,300 @@
...
+
+public class CreateServerGroupWithArtifactsSpec extends EcsSpec {
+
+  @Autowired AccountCredentialsRepository accountCredentialsRepository;
+
+  @Autowired protected AmazonClientProvider mockAwsProvider;
+
+  @MockBean ArtifactDownloader artifactDownloader;
+
+  @MockBean AWSApplicationAutoScalingClient mockAWSApplicationAutoScalingClient;
```
allisaurus, line 63: no need to use `@MockBean` for this, let's use more basic mocks for service clients (like mockECS on line 69). Generally, mock beans would be used to mock classes that are autowired in the app context, not classes we can inject and return through other means (in this case via the `mockAwsProvider`, which is wired into the context.)

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -0,0 +1,300 @@
...
+          }
+          return false;
+        },
+        String.format("Failed to detect service creation in %s seconds", TASK_RETRY_SECONDS),
+        TASK_RETRY_SECONDS);
+
+    ArgumentCaptor<RegisterTaskDefinitionRequest> registerTaskDefArgs =
+        ArgumentCaptor.forClass(RegisterTaskDefinitionRequest.class);
+    verify(mockECS).registerTaskDefinition(registerTaskDefArgs.capture());
+    RegisterTaskDefinitionRequest seenTaskDefRequest = registerTaskDefArgs.getValue();
```
allisaurus, line 198: since we're providing more task def fields via the artifact, can we add assertions for them? specifically:

1. executionRoleArn
2. taskRoleArn
3. "name" in the containerDefinition
4. "logConfiguration" details in the container definition

allisaurus (changes requested):

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
paragbhingre: Done

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
paragbhingre: Done

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
paragbhingre: Done

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
paragbhingre: Done

clouddriver-ecs/src/integration/resources/testartifacts/createServerGroup-artifact-Fargate-targetGroup-artifactFile.json
```diff
@@ -0,0 +1,42 @@
...
+    }
+  ],
+  "cpu": "256",
+  "memory": "512",
+  "requiresCompatibilities": [
+    "FARGATE"
+  ],
+  "executionRoleArn": "arn:aws:ecs:::task-definition/SpinnakerManaged:1",
+  "networkMode": "aws-vpc",
+  "taskRoleArn" : "arn:aws:ecs:::task-definition/PLACEHOLDER:1"
```
allisaurus, line 41: for both roles (and in general for all resource types), please make the value look like an actual IAM ARN (e.g., `arn:aws:iam:::role/testTaskRole`) or an obviously-named tests value (e.g., `arn:testRole`). using the ARN format for a different kind of resource is confusing, esp. in assertions (context: I discovered this b/c above we are asserting "executionRoleArn" equal "...task-definition..." ?)

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -0,0 +1,328 @@
...
```
allisaurus, line 291: can we please add the extra assertions [from this comment](https://github.com/spinnaker/clouddriver/pull/4999#discussion_r502711948) to this test as well?

allisaurus: oh, I see they are below the `createService` arg captor - let's just move then up here then, closer to the call being calidated

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
```diff
@@ -28,14 +28,11 @@
 import com.amazonaws.services.elasticloadbalancingv2.model.DescribeTargetGroupsRequest;
 import com.amazonaws.services.elasticloadbalancingv2.model.DescribeTargetGroupsResult;
 import com.amazonaws.services.elasticloadbalancingv2.model.TargetGroup;
-import com.netflix.spinnaker.clouddriver.aws.security.AmazonCredentials;
 import com.netflix.spinnaker.clouddriver.aws.security.NetflixAmazonCredentials;
-import com.netflix.spinnaker.clouddriver.aws.security.NetflixAssumeRoleAmazonCredentials;
 import com.netflix.spinnaker.clouddriver.ecs.EcsSpec;
 import com.netflix.spinnaker.clouddriver.security.AccountCredentialsRepository;
```
allisaurus, line 33: I believe this (and where its autowired on line 48) are unused and can be removed from this file as well

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -0,0 +1,326 @@
...
+  private AWSApplicationAutoScalingClient mockAWSApplicationAutoScalingClient =
+      mock(AWSApplicationAutoScalingClient.class);
+
+  private AmazonECS mockECS = mock(AmazonECS.class);
+
+  private AmazonElasticLoadBalancing mockELB = mock(AmazonElasticLoadBalancing.class);
+
+  private final int TASK_RETRY_SECONDS = 3;
+
+  private static final String CREATE_SG_TEST_PATH = "/ecs/ops/createServerGroup";
```
pkandasamy91, line 69: nit, can we move these to somewhere centralized (since they'll be used by more than just this spec file and it's unnecessary to duplicate them?)

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -0,0 +1,326 @@
...
+      ".\n===\n"
+          + "Given description w/ task def artifacts, FARGATE launch type, and new target group fields, "
+          + "successfully submit createServerGroup operation"
+          + "\n===")
+  @Test
+  public void createServerGroup_ArtifactsFARGATETgMappingsTest()
+      throws IOException, InterruptedException {
+
+    // given
+    String url = getTestUrl(CREATE_SG_TEST_PATH);
```
pkandasamy91, line 239: Can we set this in Setup() rather than in each test?

allisaurus (approved):
😎

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
paragbhingre: resolved

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
paragbhingre: Yes, makes sense. Moved this code to EcsSpec that is getting extended by the other Spec files.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
paragbhingre: I am not able to move this to setup as variables defined in the setup() is not accessible in other test methods. We can rather extract it from method level to the class level.

pkandasamy91: You could make it a global property that's set in setup and used across methods

pkandasamy91 (approved):
Ship it! 💯 Great work @paragbhingre!

PR #5003, opened 2020-10-09
fix(aws): Fixing NPE while retrieving imageId in launch templates

jeyrschabu:
- updated to get image id from launch template data
- made lookup null-safe


jeyrschabu:
Avoids the occasional `npe` due to the eventual consistency nature between the indexing of launch templates and server groups. Switched the lookup to the cached launch template (It's there so use it).

cfieber (approved):

PR #5005, opened 2020-10-09
fix(discovery): Make 404s non fatal for enabling ASG in discovery

marchello2000:
This is a follow up PR for https://github.com/spinnaker/clouddriver/pull/4273 and an undo of https://github.com/spinnaker/clouddriver/pull/4733.

Because Discovery is not guaranteed to be always consistent, skipping enabling instances in discovery because they appear to be enabled already is dangerous.
Imagine this situation:
1. Rolling/monitored deploy starts and gets to ~50% and a failure is detected that prompts an automatic rollback
2. At this point half of the old ASG is disabled so to perform rollback, we need to enable the old ASG and disable the half baked new ASG.
3. We try to enable the instances in the old ASG but we happen to hit a new Discovery instance that hasn't fully synced yet so it thinks a paricular instance is already UP
4. We skip enabling the instance, the operation completes with some number of instances skipped due to inconsistent Discovery state.
5. Deploy proceeds with WaitForUpInstances which fails because the skipped instances will never come up.

It is better to always ask Discovery to bring enable an instance and let it reconcile if the request happens to be a no-op.

However, we need to address the issue that #4733 originally tried to address - how do deal with failures from Discovery? (e.g we tried to bring up an instance in discovery that was already up so the operation wasn't needed but Discovery failed)
Here, we mirror the logic that was implemented in #4273 which was the exact same issue but for taking instances out of Discovery:
We retry failures from Discovery while trying to enable an instance and, unless strict mode is specified, we ignore 404's from Discovery.

NOTE: Strict mode is specified when performing operation on a single instance (e.g. user click an instance in UI and disables/enables it). When opearting on an ASG strict mode is OFF


dreynaud (approved):
That approach seems reasonable to me, I can't come up with obvious corner cases 👍

marchello2000:
> That approach seems reasonable to me, I can't come up with obvious corner cases 👍

Thank you!

PR #5008, opened 2020-10-10
Revert "perf(cats/sql): Introduce V2 SQL cache schema"

robzienert:
Reverts spinnaker/clouddriver#4998

Rolling this back, again, unfortunately. We didn't have a very successful rollout and we're not sure we can roll it out. We're now observing about a 30% decrease in performance, but I can't really see how that's directly related to this change, however. I suspect it's more to do with the deployment doubling the amount of DB connections, and for awhile, doubling write traffic. We did notice some issues with our prod Aurora topology, which we fixed, but it didn't affect the outcome of our DB seeing a sustained increase of load - increase in clouddriver CPU utilization is expected with the hashing, but not the DB CPU. **We're rolling this back because 1.23 is right around the corner: I thought I could merge this in and be confident with our usage it would be good enough for the rest of the community. Since that is not the case, I do not feel comfortable keeping it in `master` for the community to consume in 1.23.**

As for my investment in this issue specifically, I'm going to bow out. I've spent a tremendous amount of time on this issue over the last couple months and need to re-focus on my own internal priorities, as they've been slipping. I'm open to review PRs and help talk things through with whomever takes this on, though. An environment to validate the changes before / after is crucial, because this change passes the eyeball test.

I think a major contributing factor is that V1 schema (which V2 inherits the same design decisions) is exacerbating all issues. In V1, we were under serious pressure to get off of Redis onto something that would perform better - so we took some large shortcuts, specifically maintaining the `id` structure (e.g. `{cloud}:{resourceType}:{account}:{resourceName}`, etc) and used `regexp` SQL queries to filter in table results in the same way that we did in Redis. Regexp queries are terrible for read performance, obviously, because they cannot fully utilize indexes. A "V3", as it were, would see each resource be able to define its own schema for each table and index as needed, where each part of the compound `id` structure would be a separate column that could be used in conditionals. This was always our plan, but it hadn't been needed yet. I suspect the solution lies somewhere in here.

Go-ahead plan? Maybe it's diving deep on this and fixing the problem, or the V3 schema, or re-introducing @ncknt's original change before this PR.

@ajordens @cfieber - feel free to add any additional points should you have any

cc @mattsanta


ajordens (approved):

robzienert:
Related: https://github.com/spinnaker/clouddriver/pull/5041

PR #5011, opened 2020-10-12
feat(aws): Some additional logging around upsert security group operations

ajordens:
Short of rewriting this operation as a Saga, I'm looking to get a better indication
of how often the eventual consistency of AWS causes a problem.

This PR also increases the amount of time we'll wait for the upsert security group
tags operation to complete. This is a proxy for waiting for consistency around the
security group creation.


aravindmd (approved):
LGTM

PR #5014, opened 2020-10-12
Fix (OCI) : Bugfix 5911 Spinnaker - Oracle Load Balancer Regional Subnets don't work

anupmpatil:
Due to this bug, while creating OCI public load balancer using Spinnaker, cloud driver had additional check that ensured that  2 regional subnets have been chosen.
But the actual logic should be, if chosen subnet is regional, allow load balancer creation using 1 subnet, else chose 2 ad-specific subnets.

Before fix:

before fix

After fix:

after fix


anupmpatil:
@maggieneterval cloud driver side fix for issue:
https://github.com/spinnaker/spinnaker/issues/5911

Deck fix has been merged. Whom should I add for reviewing this fix?

anupmpatil:
@robzienert @plumpy @ttomsu @kevinawoo Could you please help us in getting this reviewed and merged? Thanks

plumpy (approved):

PR #5017, opened 2020-10-13
fix(upgrade): Docker image bump to 3.12 to fix potential JVM error on Git handling

jasonmcintosh:
https://bugs.openjdk.java.net/browse/JDK-8229872

We encountered this bug in java 11.0.5 which is the jdk on alpine 3.11.  Alpine 3.12 uses 11.0.8 which fixes it.  Of course other updates, but this is a safeguard.  Requires having enough overlays to go over 1024 in total overalay size which unlikely to be an issue on most builds but can't be entirely ruled out.

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -1,4 +1,4 @@
-FROM alpine:3.11
+FROM alpine:3.12
 LABEL maintainer="sig-platform@spinnaker.io"
 
 ENV KUBECTL_VERSION v1.17.6


dogonthehorizon (approved):

german-muzquiz (approved):

cfieber (approved):

skaji:
Can anyone take a look at https://github.com/spinnaker/spinnaker/issues/6226 ?

skaji:
In alpine:3.12, py-pip is actually py3-pip so that awscli is installed as a python3 library.
https://github.com/spinnaker/clouddriver/blob/c8b609e6f8a37e6e1a11489ae5195902d0def7d7/Dockerfile.slim#L23-L24

After installing awscil, `apk -v --purge del py-pip` also deletes python3; so awscli does not work.

skaji:
python2 is no longer supported. Can we use python3 thoroughly in Dockerfile.slim?

PR #5018, opened 2020-10-13
fix(aws/asg): Launch template security groups from network interfaces

caseyhebebrand:
**Problem:** When cloning a launch template-backed ASG via pipelines that need to `copyLastAsgAtomicOperation`, security groups were missing.

**Solution:** Security group IDs for launch templates are stored in the network interfaces. If the stage context does not explicitly have `launchTemplate.securityGroups`, the `networkInterfaces` need to be the source of truth for setting the security groups.


aravindmd (approved):
LGTM , worth adding a comment that the SG will be in network interface section for LT

PR #5021, opened 2020-10-14
feat(aws): Implement Kork credentials and provider specific CredentialsRepository

nabuskey:
This is the AWS implementation of `CredentialsRepository` introduced in https://github.com/spinnaker/governance/blob/master/rfc/account-management.md 
Once ECS implements its own `CredentialsRepository`, `CredentialsLoader` and `AmazonAccountSynchronizer` files and beans should be removed. 

* Keep `CredentialsLoader` and `AmazonAccountSynchronizer` due to clouddriver-ecs' dependency on them
* Replace uses or `AccountCredentialsProvider` with `CredentialsRepository`
* Update `AgentProvider` and its implementatons to take `Credentials` in agents() since agents must be generated before credentials are added to repository
* Move agent generation tasks to `AmazonCredentialsLifecycleHandler` to generate/remove agents when credentials are added/removed.

@ajordens @nimakaviani @ncknt

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/InstanceTerminationLifecycleWorkerProvider.java
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/InstanceTerminationLifecycleWorkerProvider.java
@@ -21,7 +21,7 @@ import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.clouddriver.aws.deploy.ops.discovery.AwsEurekaSupport;
 import com.netflix.spinnaker.clouddriver.aws.security.AmazonClientProvider;
 import com.netflix.spinnaker.clouddriver.aws.security.NetflixAmazonCredentials;
-import com.netflix.spinnaker.clouddriver.security.AccountCredentialsProvider;
+import com.netflix.spinnaker.credentials.CredentialsRepository;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
@@ -47,7 +47,7 @@ public class InstanceTerminationLifecycleWorkerProvider {
 
   private final ObjectMapper objectMapper;
   private final AmazonClientProvider amazonClientProvider;
-  private final AccountCredentialsProvider accountCredentialsProvider;
+  private final CredentialsRepository<NetflixAmazonCredentials> credentialsRepository;
   private final InstanceTerminationConfigurationProperties properties;
   private final Provider<AwsEurekaSupport> discoverySupport;
   private final Registry registry;
@@ -56,13 +56,13 @@ public class InstanceTerminationLifecycleWorkerProvider {
   InstanceTerminationLifecycleWorkerProvider(
       @Qualifier("amazonObjectMapper") ObjectMapper objectMapper,
       AmazonClientProvider amazonClientProvider,
-      AccountCredentialsProvider accountCredentialsProvider,
+      CredentialsRepository<NetflixAmazonCredentials> credentialsRepository,
       InstanceTerminationConfigurationProperties properties,
       Provider<AwsEurekaSupport> discoverySupport,
       Registry registry) {
     this.objectMapper = objectMapper;
     this.amazonClientProvider = amazonClientProvider;
-    this.accountCredentialsProvider = accountCredentialsProvider;
+    this.credentialsRepository = credentialsRepository;
     this.properties = properties;
     this.discoverySupport = discoverySupport;
     this.registry = registry;
@@ -71,8 +71,7 @@ public class InstanceTerminationLifecycleWorkerProvider {
   @PostConstruct
   public void start() {
     NetflixAmazonCredentials credentials =
-        (NetflixAmazonCredentials)
-            accountCredentialsProvider.getCredentials(properties.getAccountName());
+        credentialsRepository.getOne(properties.getAccountName());
     ExecutorService executorService =
         Executors.newFixedThreadPool(
             credentials.getRegions().size(),
@@ -89,7 +88,7 @@ public class InstanceTerminationLifecycleWorkerProvider {
                   new InstanceTerminationLifecycleWorker(
                       objectMapper,
                       amazonClientProvider,
-                      accountCredentialsProvider,
+                      credentialsRepository,
                       new InstanceTerminationConfigurationProperties(
                           properties.getAccountName(),
                           properties
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/InstanceTerminationLifecycleWorkerProvider.java
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/InstanceTerminationLifecycleWorkerProvider.java
@@ -33,9 +33,11 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.context.annotation.DependsOn;
 import org.springframework.stereotype.Component;
 
 @Component
+@DependsOn("amazonCredentialsLoader")
 @ConditionalOnExpression(
     "${aws.lifecycle-subscribers.instance-termination.enabled:false} && ${caching.write-enabled:true}")
 public class InstanceTerminationLifecycleWorkerProvider {
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservationReportCachingAgent.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservationReportCachingAgent.groovy
@@ -50,6 +50,7 @@ import com.netflix.spinnaker.clouddriver.aws.security.AmazonClientProvider
 import com.netflix.spinnaker.clouddriver.aws.security.AmazonCredentials
 import com.netflix.spinnaker.clouddriver.aws.security.NetflixAmazonCredentials
 import com.netflix.spinnaker.clouddriver.cache.CustomScheduledAgent
+import com.netflix.spinnaker.credentials.CredentialsRepository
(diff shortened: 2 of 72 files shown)


nimakaviani:
@ajordens any thoughts on when you get a chance to look into this? thanks!

ajordens:
Looking at this today should have feedback by end-of-day/tomorrow morning.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/lifecycle/InstanceTerminationLifecycleWorkerProvider.java
```diff
@@ -71,8 +71,7 @@
   @PostConstruct
   public void start() {
     NetflixAmazonCredentials credentials =
-        (NetflixAmazonCredentials)
-            accountCredentialsProvider.getCredentials(properties.getAccountName());
+        credentialsRepository.getOne(properties.getAccountName());
     ExecutorService executorService =
```
ajordens, line 75: I got an NPE running this locally.

The `properties.getAccountName()` should be a legitimate account, something at play with the @PostConstruct ordering?

nimakaviani: looking into it, thanks!

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/config/ProviderHelpers.java
```diff
@@ -209,14 +211,32 @@ public static BuildResult buildAwsProviderAgents(
         }
       }
     }
+    if (reservationReportPool.isPresent() && awsProvider.getAgentScheduler() == null) {
```
ajordens, line 214: What makes awsProvider.getAgentScheduler() not null ... I'm seeing the reservation report caching agent not running but haven't had time to dig yet.

nabuskey: I wanted to leave as much code as it was written originally but I now realize checking for null is not a good idea here due to timing. I will do an explicit check for the presence of `ReservationReportCachingAgent` once, since it needs to run across all accounts in repository.

nabuskey: Fixed this issue and `InstanceTerminationLifeCycleWorker` issue. Please verify and thank you for your help.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservationReportCachingAgent.groovy
```diff
@@ -319,7 +317,7 @@ class ReservationReportCachingAgent implements CachingAgent, CustomScheduledAgen
             def osType = operatingSystemType(it.productDescription)
             def reservation = getReservation(region.name, it.availabilityZone, osType.name, it.instanceType)
             reservation.totalReserved.addAndGet(it.instanceCount)
-
+            def vpcOnlyAccounts = determineVpcOnlyAccounts()
```
ajordens, line 320: We can't really run this in a tight loop since it makes a describe call to ec2 and is going to get throttled (it certainly 

I played around locally with:
`final AtomicReference<Set<String>> vpcOnlyAccountsReference = new AtomicReference<>()`

and 

```
  private Set<String> determineVpcOnlyAccounts() {
    if (vpcOnlyAccountsReference.get() != null) {
      return vpcOnlyAccountsReference.get();
    }

    def vpcOnlyAccounts = new HashSet<>()

    accounts.each { credentials ->
      def amazonEC2 = amazonClientProvider.getAmazonEC2(credentials, credentials.regions[0].name)
      def describeAccountAttributesResult = amazonEC2.describeAccountAttributes(
        new DescribeAccountAttributesRequest().withAttributeNames("supported-platforms")
      )
      if (describeAccountAttributesResult.accountAttributes[0].attributeValues*.attributeValue == ["VPC"]) {
        vpcOnlyAccounts << credentials.name
      }
    }

    log.info("VPC Only Accounts: ${vpcOnlyAccounts.join(", ")}")
    vpcOnlyAccountsReference.set(vpcOnlyAccounts)

    return vpcOnlyAccounts
  }
```

and

```    def vpcOnlyAccounts = determineVpcOnlyAccounts()

    credentials.regions.each { AmazonCredentials.AWSRegion region ->
        log.info("Fetching reservation report for ${credentials.name}:${region.name}")
        long startTime = System.currentTimeMillis()
```

nabuskey: One of aims of the account registration RFC is to allow for dynamic account loading, so I think it makes sense for `vpcOnlyAccounts` to be updated to reflect account changes. Would something like this work? I will also need to update `AmazonCredentialsLifecycleHandler` to update these fields in case of credentials deletion and updates.

```
final Set<String> vpcOnlyAccounts = ConcurrentHashMap.newKeySet()
final Set<String> checkedAccounts = ConcurrentHashMap.newKeySet()
```
```
private synchronized Set<String> determineVpcOnlyAccounts() {
    accounts.each { credentials ->
      if (!checkedAccounts.contains(credentials.getName())) {
        def amazonEC2 = amazonClientProvider.getAmazonEC2(credentials, credentials.regions[0].name)
        def describeAccountAttributesResult = amazonEC2.describeAccountAttributes(
          new DescribeAccountAttributesRequest().withAttributeNames("supported-platforms")
        )
        if (describeAccountAttributesResult.accountAttributes[0].attributeValues*.attributeValue == ["VPC"]) {
          vpcOnlyAccounts.add(credentials.getName())
        }
        checkedAccounts.add(credentials.getName())
      }
    }
    log.info("VPC Only Accounts: ${vpcOnlyAccounts.join(", ")}")
    return vpcOnlyAccounts
  }
```

ajordens: Ah yes, of course.

Tbh I don't know if we even effectively support non-vpc accounts anymore. It's also unlikely that this particular caching agent is widely run.

I think even maintaining a Map<String, Boolean> for whether an account is vpc or not would suffice.

ajordens: I did roll out my earlier quick change here and it's looking better and not causing excessive rate limiting.

nimakaviani: I wonder if it makes sense to determine VPC only accounts in `AmazonCredentialsLifecycleHandler` and when these credentials get added rather than delaying it to be done inside an agent? 

The property is inherent to the accounts anyway and the extra call shouldn't introduce a crazy amount of overhead.

ajordens:
Unclear if there was an expectation that I would make these tweaks, I haven't done so yet but happy to give it another test. 

It does seem close.

nabuskey:
There is no expectation for you to make these changes. I apologize I have been busy tracing and fixing bugs in the account registration plugin. ETA is end of the day today. Thank you for fixing the bug in ProviderHelpers.

nimakaviani:
Thank you, you two. it does seem close indeed.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ReservationReportCachingAgent.groovy
nabuskey: I have moved `determineVpcOnlyAccounts()` functionality to `AmazonCredentialsLifecycleHandler`. It will now determine if a given account is VPC only or not when adding accounts. The account is removed from map when account is deleted.

nimakaviani:
hey @ajordens looks like this is ready for another round of testing. thanks!

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonCredentialsLifecycleHandler.java
```diff
@@ -0,0 +1,273 @@
...
+      credentialsRepository; // Circular dependency.
+  protected Set<String> publicRegions = new HashSet<>();
+  protected Set<String> awsInfraRegions = new HashSet<>();
+  protected boolean reservationReportCachingAgentScheduled = false;
+
+  @Override
+  public void credentialsAdded(@NotNull NetflixAmazonCredentials credentials) {
+    scheduleAgents(credentials);
+    scheduleReservationReportCachingAgent();
+    if (reservationReportCachingAgentScheduled) {
```
ajordens, line 91: It strikes me as weird to have widespread knowledge of specific caching agents, and having to manipulate local state in an agent even more so.

If we want to track whether an account is VPC or not, I'd think we could add an attribute to `NetflixAmazonCredentials`.

ajordens (approved):
I'm going to merge this, but likely do a pass and just remove the VPC account handling as I don't think its necessary.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonCredentialsLifecycleHandler.java
nimakaviani: yeah that's what I meant also in the previous comment, ie to have a flag in `NetflixAmazonCredentials` to check if the account is vpc only. maybe we can fix this in the next PR.

ajordens: Yes, I do think that is the better approach.

In this case, I'm quite sure we don't actually need to worry about vpc/non-vpc in this particular agent (I'd be super surprised if anyone but Netflix uses it).

https://github.com/spinnaker/clouddriver/pull/5045 drops the non-vpc handling and cleans up the describes we were making in the lifecycle handler.

Should something like this come up again, we can enhance `NetflixAmazonCredentials`.

PR #5022, opened 2020-10-15
feat(titus): `upsertScalingPolicy` is now implemented as a saga for Titus

ajordens:
Fairly straightforward conversion into two idempotent actions:
1) `UpsertTitusScalingPolicy` is responsible for creating or updating the scaling policy.
2) `MonitorTitusScalingPolicy` is responsible for ensuring the create/update is applied.

State between the two actions is passed via a new `TitusScalingPolicyModified` event.

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/MonitorTitusScalingPolicy.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/MonitorTitusScalingPolicy.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.titus.deploy.actions;
+
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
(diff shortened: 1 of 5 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/MonitorTitusScalingPolicy.java
```diff
@@ -0,0 +1,125 @@
...
+    TitusAutoscalingClient titusClient =
+        titusClientProvider.getTitusAutoscalingClient(
+            (NetflixTitusCredentials) accountCredentials, event.getRegion());
+
+    if (titusClient == null) {
+      throw new UserException("Autoscaling is not supported for this account/region");
+    }
+
+    // make sure the new policy was applied
+    retrySupport.retry(
```
ajordens, line 84: I've not done it yet but in theory, the saga conversation would let us drop this `retrySupport` usage.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/UpsertTitusScalingPolicy.java
```diff
@@ -0,0 +1,191 @@
...
+
+    ScalingPolicy.Builder builder = command.description.toScalingPolicyBuilder();
+
+    PutPolicyRequest.Builder requestBuilder =
+        PutPolicyRequest.newBuilder()
+            .setScalingPolicy(builder)
+            .setJobId(command.description.getJobId());
+
+    ScalingPolicyID result =
+        retrySupport.retry(
```
ajordens, line 120: Same comment as earlier about the potential to remove this `retrySupport`.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/MonitorTitusScalingPolicy.java
jonsie: I think a few network retries are OK.  We could potentially drop the total number of retries down to 5 though - if it's failing after 5 times due to a retryable exception it may be best to just give up and let the task run again at a later time.

jonsie (approved):
Feel like we could probably ratchet down the total number of retries but otherwise looks good. 👍

aravindmd (approved):

ajordens:
Going to merge this and will circle around and lower the retry intervals in a subsequent PR.

PR #5024, opened 2020-10-15
feat(core): Store saga identifier as an MDC attribute (`X-SAGA`)

ajordens:
`Saga.log()` will now also log to the configured `Logger` in addition to storing a log message event.

The logging occurs at 'info' level to remain consistent with `DefaultTask.updateStatus()`.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/sagas/SagaAtomicOperationBridge.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/sagas/SagaAtomicOperationBridge.java
@@ -27,6 +27,7 @@ import java.util.UUID;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import lombok.Builder;
+import org.slf4j.MDC;
 
 /**
  * A helper class to reduce boilerplate code while integrating Sagas into existing AtomicOperations.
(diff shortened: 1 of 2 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/sagas/SagaAtomicOperationBridge.java
```diff
@@ -55,18 +56,23 @@ public SagaAtomicOperationBridge(SagaService sagaService, String sagaId) {
...
-        SnapshotAtomicOperationInput.SnapshotAtomicOperationInputCommand.builder()
-            .cloudProvider(sagaContext.getCloudProvider())
-            .descriptionName(sagaContext.getDescriptionName())
-            .descriptionInput(sagaContext.getOriginalInput())
-            .description(applyCommand.inputDescription)
-            .priorOutputs(applyCommand.priorOutputs)
-            .nextCommand(applyCommand.initialCommand)
-            .build());
+    try {
+      MDC.put("X-SAGA", sagaName + "/" + sagaId);
```
jonsie, line 60: Minor -- In case we have sagas in other services do we want to identify the saga by the service (clouddriver in this case)?

ajordens: We have enough additional context when we log to be able to differentiate between services, I'm not sure we need more than that in this case.

PR #5027, opened 2020-10-16
feat(titus): Converted the remaining discovery-related operations into sagas

ajordens:
Some slight complication due to re-writing 2016 Groovy in Java but otherwise a 1:1
AtomicOperation to Saga Action conversion.

`EnableTitusJob` and `DisableTitusJob` do interact with both
discovery and load balancer apis.

I did think twice about splitting these two activities, but the original
implementation was idempotent (by my eyes) and I didn't think it was
necessary to have two actions.


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/AbstractTitusEnableDisableAction.java
```diff
@@ -0,0 +1,281 @@
...
+      TitusLoadBalancerClient titusLoadBalancerClient =
+          titusClientProvider.getTitusLoadBalancerClient(
+              description.getCredentials(), description.getRegion());
+
+      String serverGroupName = description.getServerGroupName();
+      String region = description.getRegion();
+
+      Job job = titusClient.findJobByName(serverGroupName, true);
+      if (job == null) {
+        saga.log("No job named '%s' found in %s", serverGroupName, region);
```
aravindmd, line 97: should this sufficient for debugging or do we also need `log.error()` ?

ajordens: `saga.log()` does end up being logged to the standard logger.

jonsie (approved):

PR #5032, opened 2020-10-20
refactor(aws): Update launch template rollout flags and revert regex

caseyhebebrand:
Updating `shouldSetLaunchTemplate` to help with account and region rollout
- `aws.features.launch-templates.allowed-accounts` to enable by account
- `aws.features.launch-templates.excluded-accounts` to exclude by accounts
- `aws.features.launch-templates.all-applications` to fully enable launch templates on everything but what lies in an exclusion list

Reverting regex support
- Application level exclusion/inclusion should be fairly narrow and transient as part of the rollout process
- For most cases excluding/allowing accounts and regions provide enough control for the rollout while keeping these properties simple

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
@@ -472,7 +472,7 @@ class AutoScalingWorker {
             return true
           }
         } catch (Exception e) {
-          log.warn("Unable to verify if application is allowed in shouldSetLaunchTemplate: ${appAccountRegion}")
+          log.error("Unable to verify if application is allowed in shouldSetLaunchTemplate: ${appAccountRegion}")
           return false
         }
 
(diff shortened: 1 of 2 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
```diff
@@ -445,14 +467,7 @@ class AutoScalingWorker {
     for (appAccountRegion in applicationAccountRegions) {
       if (appAccountRegion && appAccountRegion.contains(":")) {
         def (app, account, regions) = appAccountRegion.split(":")
```
ajordens, line 469: I still think we need to wrap all this in a try/catch ... in this case we're looking for a `:` but really needing there to be 2 (otherwise `def (app, account, regions) = appAccountRegion.split(":")` is going to fail).

caseyhebebrand: Yes, sorry I forgot to add this and will update to include this

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
```diff
@@ -444,17 +466,17 @@ class AutoScalingWorker {
...
-        String regex = null
-        if (app.startsWith("regex=")) {
-          regex = ((String) app).substring(((String) app).indexOf("=") + 1)
+        try {
+          def (app, account, regions) = appAccountRegion.split(":")
+          if (app == application && account == accountName && region in (regions as String).split(",")) {
+            return true
+          }
+        } catch (Exception e) {
+          log.warn("Unable to verify if application is allowed in shouldSetLaunchTemplate: ${appAccountRegion}")
```
aravindmd, line 475: log.error() would be better.

aravindmd (approved):

PR #5033, opened 2020-10-20
fix(cache/aws): Add authoritative type for on demand update

aufarg:
This PR make sure that only authoritative data types from `ClusterCachingAgent` will gets applied during on demand cache update. 

Fixes https://github.com/spinnaker/spinnaker/issues/6122

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ClusterCachingAgent.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ClusterCachingAgent.groovy
@@ -249,7 +249,10 @@ class ClusterCachingAgent implements CachingAgent, OnDemandAgent, AccountAware, 
 
     log.info("onDemand cache refresh (data: ${data}, evictions: ${evictions}, cacheResult: ${cacheResultAsJson})")
     return new OnDemandAgent.OnDemandResult(
-      sourceAgentType: getOnDemandAgentType(), cacheResult: cacheResult, evictions: evictions
+      sourceAgentType: getOnDemandAgentType(),
+      cacheResult: cacheResult,
+      evictions: evictions,
+      authoritativeTypes: types.findAll { it.authority == AgentDataType.Authority.AUTHORITATIVE }.collect { it.typeName }
     )
   }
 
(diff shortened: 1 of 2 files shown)


aufarg:
I'll add the necessary tests if the change is okay

cfieber:
This looks reasonable to me - nice job investigating.

Please do add the tests

aufarg:
Hi @cfieber, it appears that my previous change is not quite sufficient to fix the issue. I tried several more tests and the issue is still happening.

I investigate further and found that [here](https://github.com/spinnaker/clouddriver/blob/ab7eee1aa2ea2cb30bd8e30f1277a411885caf0e/cats/cats-core/src/main/java/com/netflix/spinnaker/cats/provider/DefaultProviderCache.java#L148) `cacheDataType`, which is the method that updates the cache, is not guarded with authoritative check. This means even if we set `authoritativeTypes`, types that is not authoritative will still replace the cache.

So I added a new commit to add the authoritative check and issue that I faced is resolved, but I'm afraid this might potentially break other agents that might miss assigning authoritative value but still works as intended because there wasn't any check. Here are the places that calls `putCacheResult` and how they pass authoritative value (i.e. determine if agent will misbehave if the change went through):

- In [CachingAgent](https://github.com/spinnaker/clouddriver/blob/ab7eee1aa2ea2cb30bd8e30f1277a411885caf0e/clouddriver-api/src/main/java/com/netflix/spinnaker/cats/agent/CachingAgent.java#L138-L141), the method is called by using the previously computed authoritative value based on agent's `providedTypes`. That means that any agents that doesn't define `providedTypes` with proper authoritative/informative value will be affected.

- In [CatsOnDemandCacheUpdater](clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.java), the method is called based on `OnDemandResult` returned by `handle`. That means any agents that doesn't set the correct `authoritativeTypes` value on `handle` return value will be affected.

Let me know what you think about this.

aufarg:
I see that for [SqlProviderCache](https://github.com/spinnaker/clouddriver/blob/c7fc9df844e8f297cf1e341347879824d3576ce6/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/SqlProviderCache.kt#L212-L220) the logic uses the authoritative types to guard the cache replacement. So there was a difference in how the two different ProviderCache handle authoritative types informations.

aufarg:
I'll make a separate PR for the fix on `DefaultProviderCache` because my first change should at least fixes the issue for `SqlProviderCache`.

aufarg:
Hi added the test for the change in `ClusterCachingAgent`. For the issue with `DefaultProviderCache`, I made a new PR [here](https://github.com/spinnaker/clouddriver/pull/5038).

cfieber (approved):

aufarg:
Hi, is it possible to port this to 1.23?

PR #5034, opened 2020-10-20
fix(aws): Set launch template security group ids and rollup to asg data

caseyhebebrand:
- Security groups were only being set on launch templates within the network interfaces. This PR sets `securityGroupIds` attribute when a launch template is being created.
- ASGs with launch templates were not rolling up `securityGroups` to the `ServerGroup` model. This PR also fixes a bug that sets the ASG security groups based on `securityGroupIds`, with a fallback to the `networkInterfaces` for safety.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonServerGroup.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonServerGroup.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.clouddriver.aws.model
 
+import com.amazonaws.services.ec2.model.RequestLaunchTemplateData
 import com.fasterxml.jackson.annotation.JsonAnyGetter
 import com.fasterxml.jackson.annotation.JsonAnySetter
 import com.fasterxml.jackson.annotation.JsonIgnore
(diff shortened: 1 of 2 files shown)


lorin (approved):
LGTM, thanks for turning this around so quickly.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonServerGroup.groovy
```diff
@@ -113,8 +113,20 @@ class AmazonServerGroup implements ServerGroup, Serializable {
       securityGroups = (Set<String>) launchConfig.securityGroups
     }
 
-    if (launchTemplate && launchTemplate.containsKey("securityGroups")) {
-      securityGroups = (Set<String>) launchConfig.securityGroups
+    if (launchTemplate) {
```
ajordens, line 116:
```
if (launchTemplate) {
  def launchTemplateData = launchTemplate.get("launchTemplateData")
  def securityGroups = (Set<String>) launchTemplateData?.get("securityGroupIds") ?: []

  if (!securityGroups) {
    // attempt to lookup security group identifiers off the first network interface
    def networkInterface = launchTemplateData?.get("networkInterfaces")?.find { it["deviceIndex"] == 0 }
    securityGroups = (Set<String>) networkInterace?.get("groups") ?: []
  }
}
```

I think you might be missing a few null checks.

caseyhebebrand: I updated the checks. I had to import `RequestLaunchTemplateData` model in order to move away from `ltData["something"]` syntax, but it still will not let me use the getter methods.

aravindmd (approved):

PR #5035, opened 2020-10-21
fix(titus): Add additional image attributes to Titus Server Group

aravindmd:
`cats_v1_taggedImage` table now has additional bits of information around image build metadata , like `jenkins host` , `build number` etc being populated from the image registry as part of one of the caching agents. 

This change fetches these additional bits of information and attaches to `buildInfo.jenkins` in the server group response object, This should make the Titus Server group response on par with Aws Server Group response w.r.t to `buildInfo`.

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/TitusUtils.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/TitusUtils.java
@@ -51,4 +51,14 @@ public class TitusUtils {
               accountCredentials.getName()));
     }
   }
+
+  /** Get registry details for a particular Titus account. */
+  @Nonnull
+  public static String getRegistry(
+      @Nonnull AccountCredentialsProvider accountCredentialsProvider, @Nonnull String credentials) {
+    AccountCredentials accountCredentials = accountCredentialsProvider.getCredentials(credentials);
+    // Assert that AccountCredentials of NetflixTitusCredentials type
+    TitusUtils.assertTitusAccountCredentialsType(accountCredentials);
+    return (((NetflixTitusCredentials) accountCredentials).getRegistry());
+  }
 }
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/TitusUtils.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/TitusUtils.java
@@ -57,7 +57,6 @@ public class TitusUtils {
   public static String getRegistry(
       @Nonnull AccountCredentialsProvider accountCredentialsProvider, @Nonnull String credentials) {
     AccountCredentials accountCredentials = accountCredentialsProvider.getCredentials(credentials);
-    // Assert that AccountCredentials of NetflixTitusCredentials type
     TitusUtils.assertTitusAccountCredentialsType(accountCredentials);
     return (((NetflixTitusCredentials) accountCredentials).getRegistry());
   }
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
@@ -191,16 +191,18 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
     serverGroup.scalingPolicies = serverGroupData.attributes.scalingPolicies
     serverGroup.targetGroups = serverGroupData.attributes.targetGroups
 
-    serverGroup.buildInfo.jenkins = [:]
-    def imageAttributes = getTaggedImageAttributes(account, job)
-    if (imageAttributes != null && !imageAttributes.isEmpty()) {
-      imageAttributes = imageAttributes.first()
-      Map<String, String> jenkins = ["name"  : Optional.ofNullable(imageAttributes.jenkinsJob).orElse("NA"),
-                                     "number": Optional.ofNullable(imageAttributes.buildNumber).orElse("NA"),
-                                     "host"  : Optional.ofNullable(imageAttributes.jenkinsHost).orElse("NA")
-      ]
-      serverGroup.buildInfo.jenkins = jenkins
-    }
+    getTaggedImageAttributes(account, job).ifPresentOrElse(
+      { it ->
+        serverGroup.buildInfo.jenkins =
+          [
+            "name"  : it.jenkinsJob ?: "NA",
+            "number": it.buildNumber ?: "NA",
+            "host"  : it.jenkinsHost ?: "NA"
+          ]
+
+      },
+      { serverGroup.buildInfo.jenkins = [:] }
+    )
 
     if (includeDetails) {
       serverGroup.instances = translateInstances(resolveRelationshipData(serverGroupData, INSTANCES.ns), Collections.singletonList(serverGroupData)).values()
@@ -213,7 +215,7 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
     serverGroup
   }
 
-  private List<Map<String, String>> getTaggedImageAttributes(String account, Job job) {
+  private Optional<Map<String, String>> getTaggedImageAttributes(String account, Job job) {
     try {
       // Infer registry from account configuration
       String registry = TitusUtils.getRegistry(accountCredentialsProvider, account)
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
@@ -22,13 +22,16 @@ import com.netflix.spinnaker.cats.cache.CacheData
 import com.netflix.spinnaker.cats.cache.CacheFilter
 import com.netflix.spinnaker.cats.cache.RelationshipCacheFilter
 import com.netflix.spinnaker.clouddriver.core.provider.agent.ExternalHealthProvider
+import com.netflix.spinnaker.clouddriver.docker.registry.provider.DockerRegistryProviderUtils
+import com.netflix.spinnaker.clouddriver.docker.registry.cache.Keys as DockerRegistryCacheKeys
 import com.netflix.spinnaker.clouddriver.model.ClusterProvider
 import com.netflix.spinnaker.clouddriver.model.ServerGroupProvider
+import com.netflix.spinnaker.clouddriver.security.DefaultAccountCredentialsProvider
 import com.netflix.spinnaker.clouddriver.titus.TitusCloudProvider
+import com.netflix.spinnaker.clouddriver.titus.TitusUtils
 import com.netflix.spinnaker.clouddriver.titus.caching.Keys
 import com.netflix.spinnaker.clouddriver.titus.caching.TitusCachingProvider
 import com.netflix.spinnaker.clouddriver.titus.caching.utils.AwsLookupUtil
-import com.netflix.spinnaker.clouddriver.titus.caching.utils.CachingSchema
 import com.netflix.spinnaker.clouddriver.titus.caching.utils.CachingSchemaUtil
 import com.netflix.spinnaker.clouddriver.titus.client.model.Job
 import com.netflix.spinnaker.clouddriver.titus.client.model.Task
@@ -39,9 +42,7 @@ import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
-
 import javax.inject.Provider
-
 import static com.netflix.spinnaker.clouddriver.core.provider.agent.Namespace.HEALTH
 import static com.netflix.spinnaker.clouddriver.titus.caching.Keys.Namespace.*
 
@@ -56,6 +57,8 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
 
   private final Provider<AwsLookupUtil> awsLookupUtil
   private final Provider<CachingSchemaUtil> cachingSchemaUtil
+  private final DefaultAccountCredentialsProvider accountCredentialsProvider
+
 
   @Autowired
   TitusClusterProvider(TitusCloudProvider titusCloudProvider,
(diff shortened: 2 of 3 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/TitusUtils.java
```diff
@@ -51,4 +51,14 @@ public static void assertTitusAccountCredentialsType(AccountCredentials accountC
               accountCredentials.getName()));
     }
   }
+
+  /** Get registry details for a particular Titus account. */
+  @Nonnull
+  public static String getRegistry(
+      @Nonnull AccountCredentialsProvider accountCredentialsProvider, @Nonnull String credentials) {
+    AccountCredentials accountCredentials = accountCredentialsProvider.getCredentials(credentials);
+    // Assert that AccountCredentials of NetflixTitusCredentials type
```
robzienert, line 60: nit: I don't see value in code comments like this, since the code itself is documenting the same thing.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
```diff
@@ -195,6 +213,20 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
     serverGroup
   }
 
+  private List<Map<String, String>> getTaggedImageAttributes(String account, Job job) {
```
robzienert, line 216: This method signature could use some love. Either [L197](https://github.com/spinnaker/clouddriver/pull/5035/files#diff-8898a6601d27549d1304e49f079d5827142b15e673ddd34794461efee1482460R197) isn't correctly handling this method's return type and has bugs, or this method is returning the wrong type.

Inferring what I can from your current usage, this method returns either 0 or 1 list elements. If that is the case, then this method should probably return `Optional<Map<String, String>>` so you could do:

```
serverGroup.buildInfo.jenkins = getTaggedImage.ifPresentOrElse(
  { it -> 
    [
      "name": it.jenkinsJob ?: "NA", 
      "number": it.buildNumber ?: "NA", 
      "host": it.jenkinsHost ?: "NA"
    ] 
  },
  { [:] }
)
```

Doing so would make the method usage not only much clearer, but less prone to bugs.

robzienert (changes requested):

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
aravindmd: 👍 made some updates based on your input.

aravindmd:
Going to verify if deck needs any changes before merging this in.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/TitusUtils.java
aravindmd: fixed this

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
robzienert: Did you test this? You changed the signature, but didn't change the method body at all.

aravindmd: Yes , and it seems to work fine , do you see any other concerns ?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
```diff
@@ -195,6 +215,20 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
...
+      String registry = TitusUtils.getRegistry(accountCredentialsProvider, account)
+      String key = DockerRegistryCacheKeys.getTaggedImageKey(registry, job.applicationName, job.version)
+      Set<CacheData> images = DockerRegistryProviderUtils
+        .getAllMatchingKeyPattern(cacheView, DockerRegistryCacheKeys.Namespace.TAGGED_IMAGE.getNs(), key)
+      return images.collect { it.attributes }
+    }
+    catch (Exception e) {
+      log.error("Failed to fetch tagged image attributes ", e)
+    }
+  }
```
robzienert, line 230: Continuing from: https://github.com/spinnaker/clouddriver/pull/5035#discussion_r509855514

The method signature was changed to `Optional<Map<String, String>>` but the method does not do anything to make the method return an optional, and the one place that it returns a value, it's of type `List<Map<String, String>>`. I haven't tested this code that I'm about to write, but I would have expected to see something more like this:

```
  private Optional<Map<String, String>> getTaggedImageAttributes(String account, Job job) {
    try {
      // Infer registry from account configuration
      String registry = TitusUtils.getRegistry(accountCredentialsProvider, account)
      String key = DockerRegistryCacheKeys.getTaggedImageKey(registry, job.applicationName, job.version)
      Set<CacheData> images = DockerRegistryProviderUtils
        .getAllMatchingKeyPattern(cacheView, DockerRegistryCacheKeys.Namespace.TAGGED_IMAGE.getNs(), key)
      List<Map<String, String>> allAttributes = images.collect { it.attributes }
      if (allAttributes.size() > 1) {
        throw new TItusException("Expected 1 tagged image while retrieving attributes, but got ${allAttributes.size()}")
      }
      return Optional.of(allAttributes.first())
    }
    catch (Exception e) {
      log.error("Failed to fetch tagged image attributes ", e)
    }
    return Optional.empty()
  }
```

The key changes being:

1. The end of the method body returns `Optional.empty()` if we are in an exceptional case, instead of implicitly returning `null`.
2. Validate that we're indeed only getting one image from `getAllMatchingKeyPattern`. Since this method returns a `Set`, I would assume that getting more than 1 image is an error condition, since `Set`s are unordered and thus this method becomes non-deterministic.
3. Returns `Optional.of(allAttributes.first())` if we indeed get only one image from the query.

I ask if you tested this because running this method as it exists today would cause a class cast exception in the case where you're returning a result, or an NPE in the case where no image was found.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
```diff
@@ -56,20 +57,25 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
...
 
   @Autowired
   TitusClusterProvider(TitusCloudProvider titusCloudProvider,
                        TitusCachingProvider titusCachingProvider,
                        Cache cacheView,
                        ObjectMapper objectMapper,
                        Provider<AwsLookupUtil> awsLookupUtil,
-                       Provider<CachingSchemaUtil> cachingSchemaUtil) {
+                       Provider<CachingSchemaUtil> cachingSchemaUtil,
+                       DefaultAccountCredentialsProvider accountCredentialsProvider) {
```
ajordens, line 70: Why `DefaultAccountCredentialsProvider` and not `AccountCredentialsProvider`?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
```diff
@@ -184,6 +190,20 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
     serverGroup.placement.region = region
     serverGroup.scalingPolicies = serverGroupData.attributes.scalingPolicies
     serverGroup.targetGroups = serverGroupData.attributes.targetGroups
+
+    getTaggedImageAttributes(account, job).ifPresentOrElse(
+      { it ->
+        serverGroup.buildInfo.jenkins =
+          [
+            "name"  : it.jenkinsJob ?: "NA",
```
ajordens, line 198: Curious for the "NA" value as if the attributes aren't there weren't going to return an empty map.

Why not just leave them null?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
```diff
@@ -184,6 +190,20 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
...
+            "number": it.buildNumber ?: "NA",
```
ajordens, line 199: Is the key name 'number' obvious enough here, is it better than just leaving as `buildNumber`?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
```diff
@@ -184,6 +190,20 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
     serverGroup.placement.region = region
     serverGroup.scalingPolicies = serverGroupData.attributes.scalingPolicies
     serverGroup.targetGroups = serverGroupData.attributes.targetGroups
+
+    getTaggedImageAttributes(account, job).ifPresentOrElse(
+      { it ->
+        serverGroup.buildInfo.jenkins =
```
ajordens, line 196: I'm assuming that `serverGroup.buildInfo` is never null here right?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
aravindmd: i just followed the same naming convention we have for AWS Server groups , might be easier for deck in this case.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
aravindmd: yes , it either has some details or will be an empty map

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
aravindmd: yeah makes sense , i'll leave them null.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
aravindmd: Makes sense what you mention here , from testing it locally it is returning an Optional , i'll validate a few more cases and update the code based on your inputs here.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
aravindmd: Updated to use `AccountCredentialsProvider` instead

robzienert (approved):

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
```diff
@@ -22,13 +22,17 @@ import com.netflix.spinnaker.cats.cache.CacheData
 import com.netflix.spinnaker.cats.cache.CacheFilter
 import com.netflix.spinnaker.cats.cache.RelationshipCacheFilter
 import com.netflix.spinnaker.clouddriver.core.provider.agent.ExternalHealthProvider
+import com.netflix.spinnaker.clouddriver.docker.registry.provider.DockerRegistryProviderUtils
+import com.netflix.spinnaker.clouddriver.docker.registry.cache.Keys as DockerRegistryCacheKeys
```
cfieber, line 26: are there tabs in here or something?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
```diff
@@ -195,6 +216,25 @@ class TitusClusterProvider implements ClusterProvider<TitusCluster>, ServerGroup
...
+      Set<CacheData> images = DockerRegistryProviderUtils
+        .getAllMatchingKeyPattern(cacheView, DockerRegistryCacheKeys.Namespace.TAGGED_IMAGE.getNs(), key)
+      List<Map<String, String>> allAttributes = images.collect {it.attributes }
+      if(allAttributes.size() >1) {
+        throw new TitusException("More than 1 tagged image found, Expected 1, but got ${allAttributes.size()}")
+      }
+      else return Optional.of(allAttributes.first())
+    }
+    catch (Exception e) {
+      log.error("Failed to fetch tagged image attributes ", e)
```
cfieber, line 233: is this the correct behavior on a failure here? what happens if we deploy with no attributes, we just see no build information?

aravindmd: The idea here was if we can fetch the image attributes we populate `buildInfo.jenkins.number` , `buildInfo.jenkins.name` and `buildInfo.jenkins.host`. In case of an error or no attributes it would be `buildInfo.jenkins : {}` . There could be a fair amount of images without these attributes. Do you see any concern with this approach ?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
aravindmd: no tabs here `import com.netflix.spinnaker.clouddriver.docker.registry.cache.Keys as DockerRegistryCacheKeys` , not sure why its showing up this way

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
cfieber: I'd at least like to see a differentiation between an ERROR (something unexpected happened looking things up) vs an expected case that there are no tags available (probably DEBUG)

aravindmd: sounds good , i'll address this.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusClusterProvider.groovy
robzienert: Github always renders aliased imports like this. It's indeed odd.

cfieber (approved):

PR #5004, opened 2020-10-09
feat(aws): Use Kork's BaseProvider

nabuskey:
This is the first step for AWS provider to use `CredentialsRepository`  introduced in https://github.com/spinnaker/governance/blob/master/rfc/account-management.md. 

- Use the new `BaseProvider` introduced in #4930 
- Move provider configuration to helper functions

PR #5036, opened 2020-10-21
fix(aws): Always index private images for an account

ajordens:
Corrects a regression from spinnaker/clouddriver#5004 where we did not consistently
index private images across all accounts.


ajordens:
FYI @nabuskey

PR #5038, opened 2020-10-22
fix(cats-core): Consult authorative type before update

aufarg:
During my [attempt]( https://github.com/spinnaker/clouddriver/pull/5033) to fix problem when on-demand cache update is requested for `ClusterCachingAgent`, I managed to discover that a problem from the underlying update method which should affect all non-SQL provider cache.

I investigate further and found that the call to [cacheDataType](https://github.com/spinnaker/clouddriver/blob/ab7eee1aa2ea2cb30bd8e30f1277a411885caf0e/cats/cats-core/src/main/java/com/netflix/spinnaker/cats/provider/DefaultProviderCache.java#L148), which is the method that updates the cache, is not guarded with authoritative check. This means even if we set `authoritativeTypes`, types that is not authoritative will still replace the cache.

So I added a new commit to add the authoritative check and issue that I faced is resolved, but I'm afraid this might potentially break other agents that might miss assigning authoritative value but still works as intended because there wasn't any check. Here are the places that calls `putCacheResult` and how they pass authoritative value (i.e. determine if agent will misbehave if the change went through):

- In [CachingAgent](https://github.com/spinnaker/clouddriver/blob/ab7eee1aa2ea2cb30bd8e30f1277a411885caf0e/clouddriver-api/src/main/java/com/netflix/spinnaker/cats/agent/CachingAgent.java#L138-L141), the method is called by using the previously computed authoritative value based on agent's `providedTypes`. That means that any agents that doesn't define `providedTypes` with the correct value will be affected.

- In [CatsOnDemandCacheUpdater](clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsOnDemandCacheUpdater.java), the method is called based on `OnDemandResult` returned by `handle`. That means any agents that doesn't set the correct `authoritativeTypes` value on `handle` return value will be affected.

### DIFF
--- a/cats/cats-core/src/main/java/com/netflix/spinnaker/cats/provider/DefaultProviderCache.java
+++ b/cats/cats-core/src/main/java/com/netflix/spinnaker/cats/provider/DefaultProviderCache.java
@@ -145,7 +145,9 @@ public class DefaultProviderCache implements ProviderCache {
         previousSet = new HashSet<>();
       }
       if (cacheResult.getCacheResults().containsKey(type)) {
-        cacheDataType(type, sourceAgentType, cacheResult.getCacheResults().get(type));
+        if (authoritativeTypes.contains(type)) {
+          cacheDataType(type, sourceAgentType, cacheResult.getCacheResults().get(type));
+        }
         for (CacheData data : cacheResult.getCacheResults().get(type)) {
           previousSet.remove(data.getId());
         }


robzienert:
Tests are failing for non-transient reasons.

aufarg:
> Tests are failing for non-transient reasons.

Yes, I'm waiting for a review on the change first to see if adding the conditional there makes sense before fixing the tests.

cfieber (commented):
~~looks good to me, I approved the other PR where the authoritative types are populated~~

scratch that - I think this might be too heavy handed for the non onDemand case

cats/cats-core/src/main/java/com/netflix/spinnaker/cats/provider/DefaultProviderCache.java
```diff
@@ -145,7 +145,9 @@ public void putCacheResult(
         previousSet = new HashSet<>();
       }
       if (cacheResult.getCacheResults().containsKey(type)) {
-        cacheDataType(type, sourceAgentType, cacheResult.getCacheResults().get(type));
+        if (authoritativeTypes.contains(type)) {
```
cfieber, line 148: this will never populate relationship data when we are dealing with a full cache result from a caching agent (not the on demand refresh of a single entity)

I think we need to rethink this or we may be missing some of the relationship links for entities (especially for the redis cache implementation)

aufarg: IIUC, full cache result comes from https://github.com/spinnaker/clouddriver/blob/91515774dcfa43835efcdcfa993a0ed9efa1bffa/clouddriver-api/src/main/java/com/netflix/spinnaker/cats/agent/CachingAgent.java#L90

which later computes authoritative types:
https://github.com/spinnaker/clouddriver/blob/91515774dcfa43835efcdcfa993a0ed9efa1bffa/clouddriver-api/src/main/java/com/netflix/spinnaker/cats/agent/CachingAgent.java#L94-L99

then passed to `putCacheResult`/`addCacheResult`:
https://github.com/spinnaker/clouddriver/blob/91515774dcfa43835efcdcfa993a0ed9efa1bffa/clouddriver-api/src/main/java/com/netflix/spinnaker/cats/agent/CachingAgent.java#L138-L142

So we should still have agents populating the data, although now only authoritative data from these agents are used. Are there other sources that I missed?

aufarg: I did a quick look at `SqlProviderCache`, it seems it also cache only authoritative data types.  https://github.com/spinnaker/clouddriver/blob/53f9d33362c02b0bfe4b9ca86a1ea9e6cf9131b7/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/SqlProviderCache.kt#L212-L220

cfieber: FWIW the `isPartialResult` doesn't seem to be used much (at least in the cloud providers that I have enabled currently anyway) and defaults to false so I'm not sure it is 100% the right thing to expect we rely on. In the majority of cases we'll be going through the putCacheResult for all cache updates

In the SQL cache there isn't a need to cache non-authoritative results because SQL gives you the ability to join

In the Redis implementation we store a pointer each direction (e.g. from cluster -> serverGroup and from serverGroup -> cluster) and without populating the non-authoritative we may be left with the inability to link from object to another

PR #5041, opened 2020-10-23
fix(amazon): Fix application name matching in load balancer cache keys

christopherthielen:
This stops the previous behavior where it would match any segment, e.g.: account, region, vpc, type

Previously if you queried  `api.myspinnaker.com/applications/{someaccount/someregion/somevpc}/loadBalancers`, you would get all load balancers in that account/region/vpc.  We match the load balancer key and look for the application name.  The `.*:` portion was too loose and would match the wrong segments from `aws:loadBalancers:account:region:loadbalancer-name:vpc:type`

This tightens up the regex to look only in the application portion of the key.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonLoadBalancerProvider.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonLoadBalancerProvider.groovy
@@ -66,7 +66,8 @@ class AmazonLoadBalancerProvider implements LoadBalancerProvider<AmazonLoadBalan
   }
 
   private static boolean applicationMatcher(String key, String applicationName) {
-    return key ==~ 'aws:.*:' + applicationName + '-.*' || key ==~ 'aws:.*:' + applicationName || key ==~ 'aws:.*:' + applicationName + ':.*'
+    // aws:loadBalancers:account:region:loadbalancer-name:vpc:type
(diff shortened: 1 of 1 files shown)


christopherthielen:
Not sure if CI will pass.  `./gradlew test` failed on my workstation because of something docker docker

robzienert (approved):

robzienert:
Wonder if we do this other places...?

Publicly thinking to myself, not feedback for your PR: It may make sense to create a name scheme utility that would let us do something like:

```kotlin
 matchesApplication = resourceSchemeMatcher
  .get<AmazonResourceScheme.LoadBalancer>()
  .matches {
    application = "myapp"
  }

sealed class AmazonResourceScheme<T : AmazonResourceScheme> : ResourceScheme<T> {

  data class LoadBalancer(
    // would use spring to validate that resource name is complete in other cases
    @Nonnull val application: String? = null,
    val account: String? = null,
    // ...
  ) : AmazonResourceScheme<LoadBalancer>

}

interface ResourceSchemeMatcher<T : ResourceScheme> {
  fun matches((T) -> T): Boolean
}
```

This would actually help quite a few other areas of clouddriver, as well, especially in making the cache layer more better.

christopherthielen:
it would be doubly nice if the key format didn't leak out to the caller, i.e. this part: ```  .matches("aws:loadBalancers:0123456789:...") {
```

robzienert:
Oh, hm. Yeah, I suppose you could just pass in `Any`, having that object be say... `AmazonLoadBalancer` and the matcher would know how to tease out the right information from the amazon object.

PR #5048, opened 2020-10-26
feat(titus): Validate that `jobId` provided to `UpsertTitusScalingPolicy` operation

ajordens:
Also added an additional `hasValidationErrors` tag to the `authorization.missingApplication`
metric.

This tag will allow us to silence some alerting. The presence of a validation error will
prevent execution (and this metric is meant to indicate operations that are executing but
lack application details).


ajordens:
Cleans up a bit of noise on our side.

aravindmd (approved):

PR #5051, opened 2020-10-27
fix(aws lb): Parse LBs with dashes in them

marchello2000:
Follow up to https://github.com/spinnaker/clouddriver/pull/5041
Some load balancers wouldn't get parsed because the regex required `:` at the end and didn't have the `-stack-detail` as optional

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonLoadBalancerProvider.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonLoadBalancerProvider.groovy
@@ -67,7 +67,7 @@ class AmazonLoadBalancerProvider implements LoadBalancerProvider<AmazonLoadBalan
 
   private static boolean applicationMatcher(String key, String applicationName) {
     // aws:loadBalancers:account:region:loadbalancer-name:vpc:type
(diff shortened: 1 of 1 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonLoadBalancerProvider.groovy
```diff
@@ -67,7 +67,7 @@ class AmazonLoadBalancerProvider implements LoadBalancerProvider<AmazonLoadBalan
 
   private static boolean applicationMatcher(String key, String applicationName) {
     // aws:loadBalancers:account:region:loadbalancer-name:vpc:type
-    return key ==~ 'aws:[^:]*:[^:]*:[^:]*:' + applicationName + '(-[^:]*):'
+    return key ==~ 'aws:[^:]*:[^:]*:[^:]*:' + applicationName + '(-[^:]*)?:.*'
```
ajordens, line 70: Don't want to get in the way of this bug fix _but_ it'd be nice to have some unit tests here.

ajordens (approved):

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonLoadBalancerProvider.groovy
marchello2000: good point, will add some!

PR #5057, opened 2020-10-29
feat(kubernetes): Add raw resources endpoint

julienduchesne:
**/applications/{application}/rawResources**

This allows a Spinnaker user to query every kind of resource for an application, so you could have a nice view (which we're working on) that shows:
secrets, hpas, config maps, webhooks, roles, pvcs, etc

The raw resource only has a provider for Kubernetes now but it's a generic format so it could very well support other providers

Let me know if anything is missing here. I'm not familiar with this repo. I'm open to making changes

Also, I didn't find any tests for other providers like load balancer and server groups. Do I need any here?

This was discussed a bit here: https://spinnakerteam.slack.com/archives/C2W6RE0N4/p1603738949056400 (not with any of you though)


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesCacheDataConverter.java
```diff
@@ -127,15 +123,12 @@ public static void convertAsResource(
...
-    if (logicalRelationshipKinds.contains(spinnakerKind)
-        && !Strings.isNullOrEmpty(moniker.getApp())) {
-      addLogicalRelationships(
-          kubernetesCacheData,
-          key,
-          account,
-          moniker,
-          clusterRelationshipKinds.contains(spinnakerKind));
-    }
+    addLogicalRelationships(
```
julienduchesne, line 126: This change adds application relationships for every resource. Not only the ones in the `logicalRelationshipKinds` list

julienduchesne:
I also want to mention that I'm very open to redoing this if you guys have a better way. This serves as a POC, to see if we had all the data we need in Spinnaker
With this endpoint, we managed to very easily create a new view in Deck that showed all of our resources, regardless of their Spinnaker kind

rbadillo:
Awesome work, I'm happy to help @julienduchesne . I think this Raw Resources View can benefit a lot.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesRawResource.java
```diff
@@ -0,0 +1,90 @@
...
+  @Nullable
+  @ParametersAreNonnullByDefault
+  public static KubernetesRawResource fromCacheData(CacheData cd) {
+    KubernetesManifest manifest = KubernetesCacheDataConverter.getManifest(cd);
+    if (manifest == null) {
+      log.warn("Cache data {} inserted without a manifest", cd.getId());
+      return null;
+    }
+    Moniker moniker = KubernetesCacheDataConverter.getMoniker(cd);
+    return new KubernetesRawResource(manifest, cd.getId(), moniker);
```
rbadillo, line 74: Should we filter resources like:

- Deployments
- StatefulSets
- DaemonSets
- Services
- IngressControllers

Basically Resources that already have a designated tab on Spinnaker?

julienduchesne: I think it should be possible. We could expose the Spinnaker Kind (load balancer, firewalls, server groups) in an attribute. That way, it'd be possible to filter them out on the UI side

jasonmcintosh: Another thought would be "sensitive" resource handling, aka secrets.  Might be careful on what's exposed/how/etc.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/RawResourceProvider.java
```diff
@@ -0,0 +1,69 @@
...
+import com.netflix.spinnaker.clouddriver.documentation.Empty;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * A rawResourceProvider is an interface for the application to retrieve {@link RawResource}
+ * objects. The interface provides a common contract for which one or many providers can be queried
+ * for their knowledge of raw resources at a given depth of specificity.
+ */
+public interface RawResourceProvider {
```
robzienert, line 28: If these interfaces are intended for implementation by cloud providers, they belong in `clouddriver-api`.

As such, the interfaces will require additional documentation (each method, nested interface, etc), to implement `SpinnakerExtensionPoint`, and be annotated with `@Alpha`.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/RawResourceController.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping(produces = MediaType.APPLICATION_JSON_VALUE)
+class RawResourceController {
+
+  @Autowired(required = false)
+  List<RawResourceProvider> rawResourceProviders = new ArrayList<>();
```
robzienert, line 39: Use constructor injection.

robzienert: Also, as this is a `List` type, `required = false` is unnecessary, as Spring will auto-populate an empty list if no beans are found implementing this type.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesRawResource.java
julienduchesne: From what I saw, secret values are not stored in the cache because they are not returned by the regular `kubectl get -o json` call. Don't know if that depends on the Kubernetes version

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesCacheDataConverter.java
```diff
@@ -166,6 +159,9 @@ private static void addLogicalRelationships(
       Moniker moniker,
       boolean hasClusterRelationship) {
     String application = moniker.getApp();
+    if (Strings.isNullOrEmpty(application)) {
+      return;
+    }
```
robzienert, line 164: In what case would this happen? Seems like something worth logging _at least_ a WARN about?

julienduchesne: Hmm, I moved that condition down from where it was (outside the function itself). I'm not super familiar with the Frigga naming, we use the moniker annotations ourselves. Is it possible that resources could have names that don't result in them having an app name?

If so, a WARN would probably be too verbose, if we're reading all resources in a namespace. What do you think?

german-muzquiz:
I'm wondering about the best way to obtain this information from kubernetes. Maybe today the kubernetes provider caches the world and this new endpoint is feasible, but this approach has come with so many performance problems, that we started to remove code that relies in the cache (like when doing deployments) so that the end goal is to not cache anything but things shown in UI.
If we want to see the full contents of the cluster in the UI then we need to think in a better way than caching everything. @ezimanyi thoughts? maybe a shallow cache that only does discovery and stores object coordinates (account, namespace, kind, name) and doing live calls for retrieving the full object details.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/RawResourceProvider.java
julienduchesne: clouddriver-api is pretty empty. Is it a new thing?

julienduchesne:
Only having a skeleton would work for this feature
Show the full list of resources (account, namespace, kind and name, as you say)
Then from the UI, you can open them up to see the full details (which would result in a cluster call)

julienduchesne:
For now though, what would be best? Should the endpoint be changed to not return the full data?

ezimanyi:
Definitely agree with @german-muzquiz's comment.  We've encountered a lot of scalability issues due to the way that clouddriver caches everything in the cluster, and have been moving towards reducing what we cache to only what is needed to support the UI.  Some examples of this are:
* Doing a live lookup to check resource status during a deployment instead of using the cache (already done)
* Instead of caching the entire Kubernetes manifest in clouddriver's cache, only store fields necessary to support the views we have (some progress, but not done)
  * This both reduces the amount of data we send/receive to the persistence store (Redis/SQL) and the amount of post-processing we need to do when rendering the UI.

Ultimately, the root cause of these performance issues are that, as currently designed, Spinnaker is replicating what is already in `etcd` for each cluster, but doing this for all configured clusters (often hundreds/thousands).  The idea of having the full manifest JSON of every object in all of your clusters in Spinnaker's single data store isn't really scalable, so we've been trying to restrict clouddriver to only cache things where we are supporting what is rendered in the UI (where doing live lookups would be expensive either because of the number of calls or because of post-processing that is needed).

If we move forward with this, we're effectively guaranteeing that we will always need to cache the full Kubernetes manifest of every resource going forward, which will limit a lot of performance improvements.  I think that we should think about a way to support this without requiring the full manifest for every kind to be stored.  If we intend to look up object by application, could we instead store that link, and then live lookup any requested objects if we need the full manifest?

On a related note, in the past there have been a number of performance improvements from reducing the set of kinds where we store relationships, as computing and storing these relationships is a reasonably expensive part of the caching cycle.  (Ex: #3758, #4861).  I think that the changes here will likely cause performance regressions so should be carefully profiled against large clusters to understand the effect.

I think this is large enough change to be discussed at the [Kubernetes SIG](https://github.com/spinnaker/governance/blob/master/sig-kubernetes/README.md) to understand better the goals and how to implement them.  The next meeting is this coming Tuesday if you'd like to present there, and we can figure out next steps from there.

I think that we can support the goals of this endpoint, but want to make sure that we do that in a way that doesn't cause performance regressions (or prevent future optimizations), and possibly also that allows people to opt in/out if there are some unavaoidable performance costs.

ezimanyi:
> Only having a skeleton would work for this feature
> Show the full list of resources (account, namespace, kind and name, as you say + labels for filtering if possible)
> Then from the UI, you can open them up to see the full details (which would result in a cluster API call)

This sounds like a good approach, and would avoid needing to cache the full manifest.  I would still prevent us from later cutting down the kinds that Spinnaker caches at all (ie, we'd still need to ask the cluster for every possible kind) but that might be a trade-off that users are willing to accept for this feature (and there's always the ability to configure specific kinds for users who want that).

julienduchesne:
Alright, we can discuss the way forward at the Kubernetes SIG. I know this feature has a lot of considerations but I also think it adds a lot of visibility to users of community Spinnaker. Thanks for helping me out with it

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/RawResourceProvider.java
robzienert: It is a new thing, yes.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesCacheDataConverter.java
robzienert: Ah, I missed that this was just pulled down from the method above. In a Frigga world, it's impossible to have a resource that is not associated with an application, which is why I asked. In a Moniker world, I don't believe it makes any guarantees about the presence of any data from a resource name, so it's probably fine as you have it.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/RawResourceProvider.java
julienduchesne: I had trouble moving these interfaces to `clouddriver-api`. For example, the NamerRegistry utility isn't available. What's the best way to do this? Only provide the function definitions in `clouddriver-api` and implement the NamerRegistry call in the Kubernetes provider?

julienduchesne:
I made some changes:
1. The manifest is no longer included in the data. For the kubernetes provider, it will be fetched using the `/manifests/` endpoint (like other pages do)
2. The added caching is now toggled on with a `cacheAllApplicationRelationships` flag in the kubernetes account config
3. A new `kubernetes.rawResourcesEndpoint` was added with the `kinds` and `omitKinds` attributes. This config controls what resources will show up when calling the endpoint

Let me know what you think!

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/NoopRawResourceProvider.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.model;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+
+public class NoopRawResourceProvider implements RawResourceProvider {
```
ezimanyi, line 23: Do we need this `NoopRawResourceProvider` as it looks like the `RawResourceController` handles the absence of a `RawResourceProvider` just fine.  (I think these exist for other providers in cases where the auto-wiring isn't robust to a missing provider.)

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/RawResource.java
```diff
@@ -0,0 +1,59 @@
...
+   * @return name
+   */
+  String getName();
+
+  /**
+   * This resource's moniker
+   *
+   * @return moniker
+   */
+  default Moniker getMoniker() {
```
ezimanyi, line 36: Do we need a default implementation of this (and thus to pull in the `Moniker` dependency)?  So far the only implementation is for Kubernetes which overrides this, so it may be better to just omit the default implementation and only add a default if needed in the future.

ezimanyi (commented):
This looks great to me!  A couple of general comments:
* I'd like to see some tests on the new classes (particularly in the `clouddriver/kubernetes/caching` directory); there should be some good examples already for similar classes.
* I wonder if it might be better to keep this in the `clouddriver-kubernetes` module for now given that only Kubernetes implements it.  (That's where the `ManifestController` lives now since other providers likely would not need to implement it.)  The reason I suggest this is that presumably there will be some iteration/changes while this is new and being tested out behind the flag; it will be easier if this is only in `kubernetes` (and will be harder if some of it is in `clouddriver-api`).  I also am generally hesitant to write these interfaces when only one cloud provider implements them; once a second cloud provider is added, the parts that live in the interface will be more clear than when we have only one interface.

Other than that a few comments inline, but this looks good to merge after addressing comments.  I definitely do want @german-muzquiz to take a look at these updates too!

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesRawResourceProvider.java
```diff
@@ -0,0 +1,76 @@
...
+        .getSingleEntry(APPLICATIONS.toString(), ApplicationCacheKey.createKey(application))
+        .map(
+            applicationData ->
+                fromRawResourceCacheData(cacheUtils.getAllRelationships(applicationData)))
+        .orElseGet(ImmutableSet::of);
+  }
+
+  private Set<KubernetesRawResource> fromRawResourceCacheData(
+      Collection<CacheData> rawResourceData) {
+    List<String> kinds = configuration.getKinds();
```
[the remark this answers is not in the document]
julienduchesne: I changed it to a set in the config class directly and it's handled correctly. So, it's better indeed :+1:

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/NoopRawResourceProvider.java
julienduchesne: Removed

julienduchesne:
@ezimanyi I contained the feature in the `clouddriver-kubernetes` project. I also added some tests for the raw resource provider as well as some caching tests (Using `cacheAllRelationships` to cache or not CRD relations). Let me know what you think!

ezimanyi (approved):
Changes look great, and overall a great PR, thanks for this! 💯

german-muzquiz (approved):
This looks good! I only left a very minor comment.

I remember that Deck is constantly making requests to get load balancers, server groups and so on while you are seeing `Infrastructure` or even `Pipelines` screens. For large clusters this makes clouddriver cry (I have seen `/serverGroups` returning in > 30 sec), so it would be good if this new endpoint is not queried automatically so often like the others.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/model/KubernetesRawResource.java
```diff
@@ -0,0 +1,78 @@
...
+  private final String account;
+  private final String name;
+  private final String namespace;
+  private final String displayName;
+  private final KubernetesApiVersion apiVersion;
+  private final KubernetesKind kind;
+  private final Map<String, String> labels;
+  private final Moniker moniker;
+  private final Long createdTime;
+  private final String cloudProvider;
```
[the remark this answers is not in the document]
julienduchesne: Indeed, removed

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/controllers/RawResourceController.java
```diff
@@ -0,0 +1,49 @@
...
+  private final KubernetesRawResourceProvider rawResourceProvider;
+
+  @Autowired
+  public RawResourceController(KubernetesRawResourceProvider rawResourceProvider) {
+    this.rawResourceProvider = rawResourceProvider;
+  }
+
+  @PreAuthorize("hasPermission(#application, 'APPLICATION', 'READ')")
+  @PostAuthorize("@authorizationSupport.filterForAccounts(returnObject)")
+  @RequestMapping(value = "/applications/{application}/rawResources", method = RequestMethod.GET)
```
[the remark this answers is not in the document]
julienduchesne: Since caching app relationship for all resources will be optional initially (and probably the page in deck too), there shouldn't be any issues. Once people try it, we'll see if there are any performance issues. However, if we do face performance issues, will pagination be the right thing to do? In order to populate the filters in the UI, we need to fetch the whole list of objects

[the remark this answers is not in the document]
german-muzquiz: Which are the filters? if they are kubernetes `kinds` or `namespaces` that information could be supplied by different endpoints. We can check in case there are performance issues.

ezimanyi:
@julienduchesne : Looks like `mergify` doesn't have permission to update the base branch; if you update it manually it should run CI then auto-merge.

julienduchesne:
Oops, I think I messed up. I rebased it onto master. It's rightfully not merging it automatically

ezimanyi:
I think it should be OK.  It's running CI now and (presuming someone doesn't merge another change during the CI run) it will auto-merge once that is done.

PR #5058, opened 2020-10-29
fix(core): Default equals()/hashCode() implementation for `AbstractAccountCredentials`

ajordens:
Addresses a problem with the `DefaultAccountCredentialsProvider` wherein duplicate `AccountCredentials`
were being returned.


ajordens:
https://github.com/spinnaker/clouddriver/blob/master/clouddriver-security/src/main/groovy/com/netflix/spinnaker/clouddriver/security/DefaultAccountCredentialsProvider.java#L44

Since that change, we've been returning duplicate results from the `CredentialsController` due to none of the `AccountCredentials` actually implementing equals/hashCode.

cfieber (approved):

robzienert (approved):

PR #5060, opened 2020-10-30
fix(aws/loadbalancing): mark draining instances as out of service

programminh:
Marking draining instances as "UP" would falsely finish a pipeline successfully.

More details about this bug is documented here: https://github.com/spinnaker/spinnaker/issues/5867
As well as a slack conversation here https://spinnakerteam.slack.com/archives/CB78MFQPP/p1588774672064600

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/InstanceTargetGroupState.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/InstanceTargetGroupState.groovy
@@ -40,16 +40,17 @@ class InstanceTargetGroupState {
   private HealthState deriveHealthState() {
     //ELBv2 has concrete states: unused -> initial -> healthy    -> draining
     //                                            \-> unhealthy -/
-    if (state == 'healthy') {
-      return HealthState.Up
+    switch (state) {
+      case 'healthy':
+        return HealthState.Up
+      case 'initial':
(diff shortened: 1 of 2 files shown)


cfieber:
There are tradeoffs either direction we take with this as it stands but I think you have to go more out of your way to shoot yourself in the foot with the current implementation than with the proposed change

Definitely interested in @dreynaud opinion here though

dreynaud:
With the current `draining == Up` model, we have the problem that unhealthy/dying instances are considered up (by orca tasks), potentially leading to an incident where we would end up with no instances in service.

With this change (`draining == OutOfService`), we have the problem that they would pass a waitForDown check in orca. So I guess the worst case here is that we might terminate `draining` instances, causing errors for the clients that are still on the line and maybe data loss.

My take is that `draining` instances should:
- NOT pass an up check (i.e. they should not be considered part of the normal rotation)
- but also NOT pass a down check (i.e. they are not taking new traffic but they are maybe still doing stuff)

My recommendation would be to introduce a new `HealthState` constant, like `Draining`, or `Stopping` to be a bit more vendor agnostic.

dreynaud (changes requested):
See https://github.com/spinnaker/clouddriver/pull/5060#issuecomment-722712064, I believe it would be better to return something like `HealthState.Stopping`

dreynaud:
Thanks for the change, LGTM! 👍

gardleopard:
Thank you!

PR #5061, opened 2020-10-30
feat(kubernetes): Send SIGKILL to kubectl

german-muzquiz:
Spawning `kubectl` jobs has a timeout controlled by `jobs.local.timeout-minutes` (default: 10 minutes), which is used for aborting the job.

The default implementation of apache commons exec uses `Process.destroy` for sending a SIGTERM signal to `kubectl`. Usually this works fine, but there may be some situations in which `kubectl` doesn't respond to that signal and may remain running for an unknown period of time.

This PR introduces a way for sending a SIGKILL signal (`kill -9`) to `kubectl` if it ignores the SIGTERM signal, by using `Process.destroyForcibly`.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/jobs/local/ForceDestroyWatchdog.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/jobs/local/ForceDestroyWatchdog.java
@@ -50,7 +50,7 @@ public class ForceDestroyWatchdog extends ExecuteWatchdog {
     try {
       Thread.sleep(GRACE_PERIOD_MS);
     } catch (InterruptedException e) {
-      // nothing to do
+      Thread.currentThread().interrupt();
     }
 
     if (process.isAlive()) {
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/jobs/local/ForceDestroyWatchdog.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/jobs/local/ForceDestroyWatchdog.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2020 Armory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.jobs.local;
+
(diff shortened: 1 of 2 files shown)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/jobs/local/ForceDestroyWatchdog.java
```diff
@@ -0,0 +1,54 @@
...
+  @Override
+  public synchronized void start(Process processToMonitor) {
+    super.start(processToMonitor);
+    this.process = processToMonitor;
+  }
+
+  @Override
+  public synchronized void timeoutOccured(final Watchdog w) {
+    super.timeoutOccured(w);
+    if (process.isAlive()) {
```
ezimanyi, line 47: Would we want to wait after sending the `SIGTERM` in the super method before sending the `SIGKILL` here?   (I'm not sure if we need to but maybe it would avoid sending too many unecessary SIGKILLs?)

ezimanyi (approved):
Looks good, great find determining that there were still `kubectl` processes running after being sent `SIGTERM`!

For a while I've wanted to convert this to [ProcessBuilder](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ProcessBuilder.html) because I think the API is quite a bit better.  But given that the real plan is to move to use the client library, it's probably not worth the effort to re-write this logic, and we should keep it using `org.apache.commons.exec` until we can delete it in favor of the client library.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/jobs/local/ForceDestroyWatchdog.java
german-muzquiz: That's a good point, I think we can wait a few fixed milliseconds before SIGKILL 👍

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/jobs/local/ForceDestroyWatchdog.java
```diff
@@ -44,10 +46,17 @@ public synchronized void start(Process processToMonitor) {
   @Override
   public synchronized void timeoutOccured(final Watchdog w) {
     super.timeoutOccured(w);
+
+    try {
+      Thread.sleep(GRACE_PERIOD_MS);
+    } catch (InterruptedException e) {
+      // nothing to do
```
ezimanyi, line 53: Usually when ignoring an `InterrputedException` it's good to re-set the interrupt status on the thread via `Thread.currentThread().interrupt()` so that we don't lose the information that the thread was interrupted.  (This lets callers further up in the stack handle or ignore the exception, and prevents cases where we ignore interruption.)

I learned that [here](https://github.com/spinnaker/clouddriver/pull/3531#discussion_r271794353) when I also made an empty catch block in this case 😄 

In this case it probably doesn't make a big difference, but (as I learned on that linked PR) I think it's generally good practice.

german-muzquiz: The more you know... I haven't seen that technique but totally makes sense

PR #5062, opened 2020-10-30
fix(aws/lambda): Support Artifact as Lambda payload

jasoncoffman:
When invoking a Lambda via the Invoke Lambda Atomic Operation, the input only accepts payloads formatted as strings. 

_fix_: allow for either string payload or Spinnaker Artifact payload when invoking Lambda


kevinawoo (approved):
👍

PR #5068, opened 2020-11-03
fix(aws): Catch not found errors in destroyASGAtomicOperation

caseyhebebrand:
Fixing `deleteLaunchSetting` to catch not found errors for launch configs and launch templates separately.


ajordens:
This looks a little more reasonable, thanks.

ajordens (approved):

PR #5042, opened 2020-10-24
feat(test): Add integration test cases for ECS CreateServerGroup for task definition inputs and artifacts

paragbhingre:
This PR contains integration tests for Create Server Group Operation for the task definition type as input as well artifacts

This PR covers following three tests:
1. Create server group operation for the task definition type as input, launch type as EC2, legacy target group without any load balancer
2. Create server group operation for the task definition type as input, launch type as EC2, legacy target group with service discovery registry
3. Create server group operation for the task definition type as artifacts, launch type as EC2, target group mappings with no container definition(i.e. zero containers)

PR #5069, opened 2020-11-03
feat(test): Add integration tests for ECS CreateServerGroup for task definition inputs and artifacts

paragbhingre:
This PR contains integration tests for Create Server Group Operation for the task definition type as input as well artifacts

This PR covers following three tests:

1. Create server group operation for the task definition type as input, launch type as EC2, legacy target group without any load balancer
2. Create server group operation for the task definition type as input, launch type as EC2, legacy target group with service discovery registry
3. Create server group operation for the task definition type as artifacts, launch type as EC2, target group mappings with no container definition(i.e. zero containers)

This PR links to the [Add Integration tests PR](https://github.com/spinnaker/clouddriver/pull/5042). Previous PR has been merged to the spinnaker master but then reverted out as it passed during PR check but failed during merge check.
Failure Condition - Bad Assertion
Reason - As the bad assertion is happening in a retry block it's possible that clouddriver is running slower during 2nd check and hence it failed to find the expected exception text.
This PR fix the same issue.


clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupSpec.java
```diff
@@ -420,15 +420,31 @@ public void createServerGroup_InputsEc2WithoutLoadBalacingTest()
         ArgumentCaptor.forClass(RegisterTaskDefinitionRequest.class);
     verify(mockECS).registerTaskDefinition(registerTaskDefArgs.capture());
     RegisterTaskDefinitionRequest seenTaskDefRequest = registerTaskDefArgs.getValue();
-    assertEquals(expectedServerGroupName, seenTaskDefRequest.getFamily());
-    assertEquals(1, seenTaskDefRequest.getContainerDefinitions().size());
+    assertEquals(
+        expectedServerGroupName,
+        seenTaskDefRequest.getFamily(),
+        "Expected server group name is ecs-integInputsEc2NoLoadBalancing. Server group name : "
```
allisaurus, line 426: (general comment) I like that we're including the expected _and_ actual value in the error msgs, but would prefer if we used `String.format()` and provided them as params. Like:

```
String.format(
    "Expected server group name to be '%s' but saw '%s'", 
    expectedServerGroupName, seenTaskDefRequest.getFamily())
```

paragbhingre:
This PR links to the [Add Integration tests PR](https://github.com/spinnaker/clouddriver/pull/5042). Previous PR has been merged to the spinnaker master but then reverted out as it passed during PR check but failed during merge check.
Failure Condition - Bad Assertion
Reason - As the bad assertion is happening in a retry block it's possible that clouddriver is running slower during 2nd check and hence it failed to find the expected exception text.
This PR fix the same issue.

allisaurus (approved):
LGTM, thank you for fixing this!

PR #5072, opened 2020-11-04
chore(test/ecs): Update integration README

allisaurus:
Adding guidance section.

### DIFF
--- a/clouddriver-ecs/src/integration/README.md
+++ b/clouddriver-ecs/src/integration/README.md
@@ -11,7 +11,7 @@ $> ./gradlew :clouddriver-ecs:integrationTest
 
 ## Guidance for modifying this package
 
-### Adding new tests
+### When to add a new test
 
 New Amazon ECS provider features of significant scope should include an integration test which exercises new functionality.
 Examples of qualifying changes include (but are not limited to):
--- a/clouddriver-ecs/src/integration/README.md
+++ b/clouddriver-ecs/src/integration/README.md
@@ -8,3 +8,22 @@ To manually run the gradle task for these tests:
 ```bash
 $> ./gradlew :clouddriver-ecs:integrationTest
 ```
+
+## Guidance for modifying this package
+
+### Adding new tests
+
+New Amazon ECS provider features of significant scope should include an integration test which exercises new functionality.
+Examples of qualifying changes include (but are not limited to):
+
+* Implementing a new atomic operation
+* New forking logic in how `CreateServerGroup` functions or is validated (especially re: broadly impactful settings like `networkMode`, launch type, load balancing, or application autoscaling)
+* Adding a new controller for a new type of resource
+
+### Changing existing tests
+
+In general, existing test cases should function as-is after new contributions to ensure existing features continue to function as expected.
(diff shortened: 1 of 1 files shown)


clouddriver-ecs/src/integration/README.md
```diff
@@ -8,3 +8,22 @@ To manually run the gradle task for these tests:
...
+
+### Adding new tests
+
+New Amazon ECS provider features of significant scope should include an integration test which exercises new functionality.
+Examples of qualifying changes include (but are not limited to):
+
+* Implementing a new atomic operation
+* New forking logic in how `CreateServerGroup` functions or is validated (especially re: broadly impactful settings like `networkMode`, launch type, load balancing, or application autoscaling)
+* Adding a new controller for a new type of resource
+
```
paragbhingre, line 22: Can we mention about mocking of some of the main clients like AmazonClientProvider and AmazonElasticLoadBalancing.  Also, preparing the JSON file for inputs and artifacts with significant values.

clouddriver-ecs/src/integration/README.md
```diff
@@ -8,3 +8,22 @@ To manually run the gradle task for these tests:
...
+### Changing existing tests
+
+In general, existing test cases should function as-is after new contributions to ensure existing features continue to function as expected.
+Possible exceptions to this guidance may include:
+
+* Updates to internal implementation details (required `@Beans`, etc.) that don't effect operation success or API response content
+* Adding and asserting on *new* data in a `clouddriver` API response
```
paragbhingre, line 29: Can we specify an example for how JSON modification might be needed to accommodate new changes after this line?
e.g. If we add support for target group mappings then legacy target group parameter should be updated to start accepting new target group mappings which has few extra parameters.

paragbhingre (commented):
Looks great to me, just added a couple of comments to add up few guidelines that personally helped me to work through it.

clouddriver-ecs/src/integration/README.md
allisaurus: > e.g. If we add support for target group mappings then legacy target group parameter should be updated to start accepting new target group mappings which has few extra parameters.

I'm a little confused by what you mean. Adding new fields/info to a test's request payload (which we are storing as JSON) is not something we should be doing on _existing_ tests (which this section is referring to), b/c that means we are requiring new information to fulfill old test cases and that would break people with existing deployments/pipelines.

paragbhingre: Correct, makes sense. Then we can just specify this example as a part of the 1st section for new tests.

clouddriver-ecs/src/integration/README.md
allisaurus: So this section is supposed to answer the question: "what kinds of new features require an integration test?" - so how to mock required clients and request payloads/responses isn't quite relevant here, but if you think we could use a section in this doc about how to author a test case, that would be a good addition to make in a future PR

allisaurus: actually i can see that the section title isn't super clear re: this intention so i will update

PR #5073, opened 2020-11-04
chore(test/ecs): Add integ test for existing server group update

paragbhingre:
Integration tests for create server group with an existing service.
Create server group operation for the task definition input, launch type as EC2, target group mappings with an existing service.


paragbhingre:
reopening to fix the issue that caused this PR to fail.

piradeepk:
Rerunning the build since it was stopped

paragbhingre:
@allisaurus @piradeepk Please review the PR.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupForExistingServiceSpec.java
```diff
@@ -0,0 +1,323 @@
...
+    // mocking calls
+    when(mockECS.describeServices(any(DescribeServicesRequest.class)))
+        .thenReturn(
+            new DescribeServicesResult()
+                .withServices(
+                    Collections.singletonList(
+                        new Service()
+                            .withServiceName("createService")
+                            .withCreatedAt(new Date())
+                            .withStatus("INACTIVE"))));
```
allisaurus, line 81: does it have to be `INACTIVE`? what happens if it's running?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupForExistingServiceSpec.java
```diff
@@ -0,0 +1,323 @@
...
+        .thenReturn(
+            new CreateServiceResult().withService(new Service().withServiceName("createdService")));
+
+    when(mockECS.registerTaskDefinition(any(RegisterTaskDefinitionRequest.class)))
+        .thenAnswer(
+            (Answer<RegisterTaskDefinitionResult>)
+                invocation -> {
+                  RegisterTaskDefinitionRequest request =
+                      (RegisterTaskDefinitionRequest) invocation.getArguments()[0];
+                  String testArn = "arn:aws:ecs:::task-definition/" + request.getFamily() + ":1";
```
allisaurus, line 93: I don't think this will have downstream effects on other parts of this operation, but if a task definition with this server group name (what would be provided as the "family" param) already exists, then the newly registered and returned task def should have an incremented revision ("2" in this case). Let's update that so it's in line with what is expected in this situation.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupForExistingServiceSpec.java
```diff
@@ -0,0 +1,323 @@
...
+
+    // given
+    String url = getTestUrl(CREATE_SG_TEST_PATH);
+    String requestBody =
+        generateStringFromTestFile(
+            "/createServerGroup-input-EC2-targetGroupMappings-existingService.json");
+    String expectedServerGroupName =
+        "ecs-integInputEC2TgMappingsExistingServiceStack-detailTest-v001";
+
+    setEcsAccountCreds();
```
allisaurus, line 254: this method no longer exists in `/master`. Please rebase to pull in the latest and update as needed.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupForExistingServiceSpec.java
```diff
@@ -0,0 +1,323 @@
...
+        .thenReturn(mockELB);
+  }
+
+  @DisplayName(
+      ".\n===\n"
+          + "Given description w/ task def artifacts, EC2 launch type, and new target group "
+          + "fields with the existing service, successfully submit createServerGroup operation"
+          + "\n===")
+  @Test
+  public void createServerGroup_ArtifactsEC2TgMappingsExistingServiceTest()
```
allisaurus, line 136: after looking at the assertions we're doing in both tests, I don't believe this test case adds any value (we are already asserting on the incremented server group name wherever it would appear with the inputs test). Can we remove this test case and it's associated artifacts/mocks?

clouddriver-ecs/src/integration/resources/testoperations/createServerGroup-input-EC2-targetGroupMappings-existingService.json
```diff
@@ -0,0 +1,37 @@
...
+  "computeUnits": 256,
+  "credentials": "ecs-account",
+  "dockerImageAddress": "nginx",
+  "ecsClusterName": "integInputEC2TgMappingsExistingService-cluster",
+  "launchType": "EC2",
+  "networkMode": "bridge",
+  "placementStrategySequence": [],
+  "reservedMemory": 512,
+  "stack": "integInputEC2TgMappingsExistingServiceStack",
+  "freeFormDetails" : "detailTest",
```
allisaurus, line 25: this field is not required and not consistently used between tests. Can we remove it?

allisaurus (changes requested):
Thanks for this! Mostly minor changes and a question.

allisaurus:
oh, and also can we update the commit (and PR) title to reflect the purpose of the test we're adding? E.g., "feat(test/ecs): Add integr test for existing server group update"

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupForExistingServiceSpec.java
paragbhingre: Yeah, actually make sense.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupForExistingServiceSpec.java
paragbhingre: Good question @allisaurus 
So we are mocking the call here and setting the status to INACTIVE in order to get the service version as 2. If we keep this status as ACTIVE or DRAINING then [this particular code](https://github.com/paragbhingre/clouddriver/blob/f86b76ae24ef05ad100ccac372b4051164cfceb8/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/EcsServerGroupNameResolver.java#L95-L115) would run max 5 times and nextServerGroupName would become v005 or v006. So in order to keep it simple we considered that the existing service is in INACTIVE state in order to get the nextServerGroupName in terms of v002.
Does that make sense?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupForExistingServiceSpec.java
```diff
@@ -0,0 +1,197 @@
...
+        generateStringFromTestFile(
+            "/createServerGroup-input-EC2-targetGroupMappings-existingService.json");
+    String expectedServerGroupName = "ecs-integInputEC2TgMappingsExistingServiceStack-v001";
+
+    when(mockECS.listServices(any(ListServicesRequest.class)))
+        .thenReturn(
+            new ListServicesResult()
+                .withServiceArns(
+                    Collections.singletonList(
+                        "arn:aws:ecs:ecs-integInputEC2TgMappingsExistingServiceStack")));
```
allisaurus, line 135: the first service in a server group created by spinnaker will have the "-v000" suffix. Can we add that to the returned string here so it better matches reality?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupForExistingServiceSpec.java
allisaurus: thanks for the code link! it looks like that logic will [actually do the same thing](https://github.com/paragbhingre/clouddriver/blob/f86b76ae24ef05ad100ccac372b4051164cfceb8/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/EcsServerGroupNameResolver.java#L105-L106) if this describe call returns empty (vs. with one inactive service). What do you think about doing that instead, since it's simpler?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupForExistingServiceSpec.java
```diff
@@ -0,0 +1,197 @@
...
+      throws IOException, InterruptedException {
+
+    // given
+    String url = getTestUrl(CREATE_SG_TEST_PATH);
+    String requestBody =
+        generateStringFromTestFile(
+            "/createServerGroup-input-EC2-targetGroupMappings-existingService.json");
+    String expectedServerGroupName = "ecs-integInputEC2TgMappingsExistingServiceStack-v001";
+
+    when(mockECS.listServices(any(ListServicesRequest.class)))
```
allisaurus, line 130: AFAICT mocking this non-zero result is what actually triggers the "increment"/existing service workflow. Can we move it to the `setup()` func along with the rest of the mocked responses?

allisaurus (commented):
thanks for the updates, @paragbhingre ! A couple more minor comments, but otherwise LGTM

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupForExistingServiceSpec.java
paragbhingre: Hey @allisaurus ,
If I try to return empty describe call (i.e. withServices but without any other properties) as you recommend then [taken slots code](https://github.com/paragbhingre/clouddriver/blob/611940b7676ee5b43f3066dfd78bc08796d01e8c/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/EcsServerGroupNameResolver.java#L71-L82) successfully returns a slot for server group with the next sequence i.e. -v001. 
But then [this particular code](https://github.com/paragbhingre/clouddriver/blob/f86b76ae24ef05ad100ccac372b4051164cfceb8/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/EcsServerGroupNameResolver.java#L95-L115) fails to consider service as empty and gives NPE.

paragbhingre: Hello @allisaurus , may I know your thoughts on this one so that I will push the remaining changes and we are good to go with this test?

allisaurus: @paragbhingre which part of the code is returning the NPE? [Line 105](https://github.com/paragbhingre/clouddriver/blob/f86b76ae24ef05ad100ccac372b4051164cfceb8/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/EcsServerGroupNameResolver.java#L105)?

paragbhingre: @allisaurus  `result.getServices().get(0).getStatus().equals("INACTIVE")` is throwing NPE as service is not empty so it tries to check the status property and we are not setting it's status property.

allisaurus: oh! so it's not recognizing the service list as empty - but aren't we just returning an empty `DescribeServicesRequest` ? That was the [suggestion](https://github.com/spinnaker/clouddriver/pull/5073#discussion_r526356346)

allisaurus: OK so I think I understand what's happening (finally). First time we call **DescribeServices** it needs to return something to indicate that we need to increment the server group version. But the 2nd time it's called it needs to return empty (or `INACTIVE`) to indicate that the `v001` "slot" is free to use.

What you have allows the same mock response to be returned both times and results in the expected SG name. I am OK with this! Though it is, realistically, a less likely scenario than the first one existing & the second one not. If you are inclined to try and mock the 2 sequential responses go ahead, though I will not continue to block on this in order to wrap up this PR.

paragbhingre: So after final trial as per your suggestion on this, the outcome is, even if we try to mock 2 sequential responses, code doesn't know which one to pick 1st and that's why the problem remains. Moreover, as per your suggestion when we send empty response in **DescribeServices** then [getTokenSlots code](https://github.com/spinnaker/clouddriver/blob/611940b7676ee5b43f3066dfd78bc08796d01e8c/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/EcsServerGroupNameResolver.java#L75-L78) fails to find a service and does not create a slot for it which means in return we get the service with v000 and not v001. 
So I guess we need to keep this INACTIVE condition as is to pass this particular test case.

clouddriver-ecs/src/integration/resources/testoperations/createServerGroup-input-EC2-targetGroupMappings-existingService.json
```diff
@@ -1,6 +1,6 @@
 {
   "account": "ecs-account",
-  "application": "ecs",
+  "application": "createService",
```
allisaurus, line 3: did we need to change the app name? if not i'd rather we kept it consistent (ok with other stuff with an app prefix having this)

allisaurus (commented):
LGM, save one questions re: app name. If you can resolve that and squash/prefix commit with "chore(ecs)" as discussed offline, this should be good to go.

clouddriver-ecs/src/integration/resources/testoperations/createServerGroup-input-EC2-targetGroupMappings-existingService.json
paragbhingre: @allisaurus  So apparently, moniker considers 1st part of service name as app name.  So if you see the service name which is "createService-integInputEC2TgMappingsExistingServiceStack-v000" and 1st part of it before hyphen "-" is createService. So, To avoid failing one of the conditions where we check moniker.app() equals descriptions.app() I changed the application name as createService.  Does that make sense?

allisaurus: yes! I guess i'm saying that i'm OK with prefixing the server group name with "ecs" instead, then, since this matches the other tests and what users will more often be doing.

paragbhingre: Okay. Got your point.

allisaurus (approved):

PR #5074, opened 2020-11-04
feat(aws): Allow for AWS account comparison

nabuskey:
This PR aims to allow for AWS account definitions to be compared with each other. Main use is for [Credentials Loaders](https://github.com/spinnaker/kork/blob/6d465ad350a8f83d96824d888e1661f0651dd9e1/kork-credentials-api/src/main/java/com/netflix/spinnaker/credentials/definition/BasicCredentialsLoader.java#L81) to correctly identify already loaded credentials.


kevinawoo (changes requested):
@nabuskey wdyt about `@EqualsAndHashCode` [lombok annotation](https://projectlombok.org/features/EqualsAndHashCode)?

Here's an [example in k8s](https://github.com/spinnaker/clouddriver/blob/c761925896edc6ccc014c38ee7b3ffc8c002a60f/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesCredentials.java#L76). Notice the `explicit` include and the `@Include` annotation on each field compare equality.

nabuskey:
@kevinawoo Did not know lombok had the annotation. Thanks!

kevinawoo (approved):
💯

PR #5083, opened 2020-11-06
feat(amazon): Add Edda proxy for LaunchTemplateVersions

caseyhebebrand:
- Add a proxy to fetch launch tempalte version collectors
- extend support for constructing an `idKey` of the format `launchTemplateId:version`


ajordens (approved):
LGTM!

PR #5086, opened 2020-11-09
feat(ecs): add more integration test for Amazon ECS createServerGroup

paragbhingre:
This PR contains integration test for Create Server Group Operation for the task definition artifact, EC2 launch type, and target group mappings with multiple load balancers and multiple containers.


paragbhingre:
@allisaurus @piradeepk Please review the PR.

clouddriver-ecs/src/integration/resources/testartifacts/createServerGroup-artifact-EC2-TGMappings-multipleLBsAndContainers-artifactFile.json
```diff
@@ -0,0 +1,67 @@
...
+          "awslogs-region": "us-west-2",
+          "awslogs-stream-prefix": "spinnaker"
+        }
+      },
+      "portMappings": [
+        {
+          "protocol": "tcp",
+          "containerPort": 84
+        }
+      ],
```
piradeepk, line 47: Shouldn't we also be specifying hostPort?

piradeepk (commented):
Looks good overall, no major comments!

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -367,4 +367,131 @@ public void createServerGroup_errorIfNoContainersTest() throws IOException, Inte
         String.format("Failed to detect task failure, in %s seconds", TASK_RETRY_SECONDS),
         TASK_RETRY_SECONDS);
   }
+
+  @DisplayName(
+      ".\n===\n"
+          + "Given description w/ task def artifacts, EC2 launch type, and target group mappings, "
+          + "with multiple load balancers and multiple containers successfully submit createServerGroup operation"
```
allisaurus, line 374: this description (and test name) are getting quite long. I'm ok with shortening this to just "multiple target group mappings" or "multiple load balancers" if you'd prefer.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -367,4 +367,131 @@ public void createServerGroup_errorIfNoContainersTest() throws IOException, Inte
...
+        ArgumentCaptor.forClass(RegisterTaskDefinitionRequest.class);
+    verify(mockECS).registerTaskDefinition(registerTaskDefArgs.capture());
+    RegisterTaskDefinitionRequest seenTaskDefRequest = registerTaskDefArgs.getValue();
+    assertEquals(expectedServerGroupName, seenTaskDefRequest.getFamily() + "-v000");
+    assertEquals(2, seenTaskDefRequest.getContainerDefinitions().size());
+    assertEquals(
+        80,
+        seenTaskDefRequest
+            .getContainerDefinitions()
+            .get(0)
```
allisaurus, line 439: I'm not sure we can rely on the ordering of these to be the same every time. instead of getting the container def by index, can we find the container in the list by name and assert on all its props then?

allisaurus: let's also assert on the name of each container (in addition to port)

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -367,4 +367,131 @@ public void createServerGroup_errorIfNoContainersTest() throws IOException, Inte
...
+    verify(mockELB, times(2)).describeTargetGroups(elbArgCaptor.capture());
+
+    ArgumentCaptor<CreateServiceRequest> createServiceArgs =
+        ArgumentCaptor.forClass(CreateServiceRequest.class);
+    verify(mockECS).createService(createServiceArgs.capture());
+    CreateServiceRequest seenCreateServRequest = createServiceArgs.getValue();
+    assertEquals("EC2", seenCreateServRequest.getLaunchType());
+    assertEquals(expectedServerGroupName, seenCreateServRequest.getServiceName());
+    assertEquals(2, seenCreateServRequest.getLoadBalancers().size());
+    LoadBalancer serviceLB1 = seenCreateServRequest.getLoadBalancers().get(0);
```
allisaurus, line 480: same comment re: list ordering above.

clouddriver-ecs/src/integration/resources/testartifacts/createServerGroup-artifact-EC2-TGMappings-multipleLBsAndContainers-artifactFile.json
allisaurus: +1 `bridge` network mode uses both (and they can be different)

paragbhingre: I was reading through the document and it says "If you are using containers in a task with the bridge network mode, you can specify a non-reserved host port for your container port mapping, or you can omit the hostPort (or set it to 0) while specifying a containerPort and your container automatically receives a port in the ephemeral port range for your container instance operating system and Docker version."
Hence I was trying that scenario in this case. Do I till have to mention the host port?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
paragbhingre: @allisaurus, AFAIK we can get container definitions from getContainerDefinitions() in the form of list. I tried to find the container in the list by name but that's not possible because there is no function available to find the container name from the list of container objects.
Do you think there is any other way by which we can figure that out?

let's also assert on the name of each container (in addition to port) -> I'm already asserting on container name on line 455-456.

allisaurus: > that's not possible because there is no function available to find the container name from the list of container objects

you're right there's no built in function, but we can use an old fashion loop OR a fancy [filter expression w/ a lambda](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java#L365-L368).

> I'm already asserting on container name on line 455-456

true! i'm just asking we moving it up so that all assertions on a given container are happening in the same place.

clouddriver-ecs/src/integration/resources/testartifacts/createServerGroup-artifact-EC2-TGMappings-multipleLBsAndContainers-artifactFile.json
piradeepk: It would be great to have a test that specifically tests host port being a different value.

paragbhingre:
@allisaurus  please verify the final changes.

clouddriver-ecs/src/integration/resources/testoperations/createServerGroup-artifact-EC2-TGMappings-multipleLBsAndContainers.json
```diff
@@ -0,0 +1,53 @@
...
+    "customKind" :  true,
+    "name" : "applications",
+    "location" : "us-west-2",
+    "reference" : "refernce",
+    "metadata" : null,
+    "artifactAccount" : "my-github",
+    "provenance" : "prov",
+    "uuid" : "uid-123"
+  },
+  "containerToImageMap": {}
```
allisaurus, line 52: this almost always contains one or more docker images the user wants to insert into the task def artifact. Can we update this test to supply this map and assert that the correct container image is inserted into each container definition of the `registerTaskDefinition` call?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/controllers/EcsClusterController.java
```diff
@@ -37,4 +39,10 @@ public EcsClusterController(EcsClusterProvider ecsClusterProvider) {
   public Collection<EcsCluster> getAllEcsClusters() {
     return ecsClusterProvider.getAllEcsClusters();
   }
+
+  @RequestMapping(value = {"/ecs/getEcsClusterDescriptions/{account}/{region}"})
```
allisaurus, line 43: i think this is a change meant for another PR?

allisaurus:
@paragbhingre  one comment re: the test, plus we should remove commit `240a1d1` from this PR. Otherwise looks good

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/controllers/EcsClusterController.java
paragbhingre: Yes, but something is up with my commit history. Let me remove those commits.

paragbhingre:
Reopening, as closed accidentally

clouddriver-ecs/src/integration/resources/testoperations/createServerGroup-artifact-EC2-TGMappings-multipleLBsAndContainers.json
paragbhingre: make sense, I have added the container to image map and asserting on the image value that container holds.

paragbhingre:
@allisaurus I have made changes please let me know if it looks fine to you so that I will rebase and push the final commit again.

clouddriver-ecs/src/integration/resources/testoperations/createServerGroup-artifact-EC2-TGMappings-multipleLBsAndContainers.json
```diff
@@ -0,0 +1,57 @@
...
+    "reference" : "refernce",
+    "metadata" : null,
+    "artifactAccount" : "my-github",
+    "provenance" : "prov",
+    "uuid" : "uid-123"
+  },
+  "containerToImageMap":
+  {
+    "application1": "PLACEHOLDER",
+    "application2": "PLACEHOLDER"
```
allisaurus, line 55: it's not helpful to keep these values as "PLACEHOLDER" since:
1) they're the same, so we can't tell if they've been mis-mapped to the wrong container definitions when we assert
2) this is the same text which typically appears in the task def artifact (I see you removed it from the artifact used for this test, but the behavior we want to test is that this "PLACEHOLDER" is _overwritten_ by the correct image ID). 

Please update to "app1/image" / "app2/image" or something similar.

allisaurus (changes requested):

clouddriver-ecs/src/integration/resources/testoperations/createServerGroup-artifact-EC2-TGMappings-multipleLBsAndContainers.json
paragbhingre: Aahhh, okay I understood now.

clouddriver-ecs/src/integration/resources/testartifacts/createServerGroup-artifact-EC2-TGMappings-multipleLBsAndContainers-artifactFile.json
```diff
@@ -0,0 +1,69 @@
...
+        "logDriver": "awslogs",
+        "options": {
+          "awslogs-group": "spinnaker-ecs-demo",
+          "awslogs-region": "us-west-2",
+          "awslogs-stream-prefix": "spinnaker"
+        }
+      },
+      "portMappings": [
+        {
+          "hostPort": 80,
```
allisaurus, line 45: sorry, I should have noticed this before - host ports can't collide (won't fail this test but ECS will reject the request). Can we update this to be 84 like the container port?

paragbhingre:
Reopening, as closed accidentally.

paragbhingre:
@allisaurus  Addressed all the concerns, please take a final look and then we are good to merge this one.

allisaurus (approved):

PR #5089, opened 2020-11-10
feat(aws): Add filtering of AutoScalingGroups to include/exclude based on Tags

gavinbunney:
This provides a filtering mechanism to include/exclude AutoScalingGroups based on configured name/value pairs. The intent of naming the filter configuration `AmazonCachingAgentFilterConfiguration` is to allow it to be reused for different agent filtering (load balancers etc).

This is needed to allow for a subset of AWS resources to be managed by Spinnaker (for our environment, exclude is enough, but added include for other use cases).

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonCachingAgentFilterConfiguration.java
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonCachingAgentFilterConfiguration.java
@@ -1,0 +1,57 @@
+package com.netflix.spinnaker.clouddriver.aws.provider.agent;
+
+import java.util.List;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+@Component
+@ConfigurationProperties("aws.caching.filter")
+public class AmazonCachingAgentFilterConfiguration {
+
+  List<TagFilterOption> includeTags;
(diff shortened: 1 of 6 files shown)


ajordens (commented):
I like it. 

One minor point, we're trying to cut down on the amount of net-new Groovy code. Mind swapping that small configuration component over to Java.

gavinbunney:
> I like it.
> 
> One minor point, we're trying to cut down on the amount of net-new Groovy code. Mind swapping that small configuration component over to Java.

Done! Thanks

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonCachingAgentFilterConfiguration.java
```diff
@@ -0,0 +1,57 @@
+package com.netflix.spinnaker.clouddriver.aws.provider.agent;
+
+import java.util.List;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+@Component
+@ConfigurationProperties("aws.filter")
```
ajordens, line 8: Let's move this to `aws.caching.filter` and we'll be good to go.

gavinbunney: Sure thing - done.

ajordens:
Awesome, thanks for the contribution!

PR #5097, opened 2020-11-14
feat(sql): add support for PostgreSQL

ncknt:
This adds support for PostgreSQL 10+ (and likely supports older versions as well).

As part of this, table constraints now default per dialect. Postgres does not have limits on indices and table names are only up to 63 bytes.

From a user point of view, nothing to do except use the right jdbc URL.


ncknt:
Note that the SqlCache tests now run with Postgres so it adds time to the build. I also had to change the initial schema that was using MySQL `bigint(20)` to just `bigint`. It should not change how things are stored as the `20` is just a hint for the client.

I did not use `jsonb` for PostgreSQL b/c it would incur a cost when writing and the code still explicitly parse the `body` fields anyway. It could be a future improvement.

ncknt:
Fixes https://github.com/spinnaker/spinnaker/issues/5648

zachsmith1 (approved):
LGTM

cats/cats-sql/src/test/groovy/com/netflix/spinnaker/cats/sql/PostgreSqlCacheSpec.groovy
```diff
@@ -0,0 +1,64 @@
...
+      "test",
+      context,
+      mapper,
+      null,
+      clock,
+      sqlRetryProperties,
+      "test",
+      Mock(SqlCacheMetrics),
+      dynamicConfigService,
+      new SqlConstraintsInitializer().getDefaultSqlConstraints(SQLDialect.MYSQL)
```
german-muzquiz, line 60: `SQLDialect.POSTGRES`?

clouddriver-sql/src/main/resources/db/changelog/20190822-initial-event-schema.yml
```diff
@@ -2,6 +2,7 @@ databaseChangeLog:
   - changeSet:
       id: create-event-aggregates-table
       author: robzienert
+      validCheckSum: 8:99ad0fa01e0a2f5a2d66e015a58dccfd
```
german-muzquiz, line 5: What happens for users running MySQL and then upgrading to this new version? Is this `initial-event-schema` applied again? I guess it should not be applied and existing installations will still have the `bigint(20)` constraint, but new ones will not be constrained.

clouddriver-sql/src/main/resources/db/changelog/20201110-field-type-postgres.yml
```diff
@@ -0,0 +1,105 @@
...
+            newDataType: varchar(36)
+            tableName: task_results
+        - modifyDataType:
+            columnName: task_id
+            newDataType: varchar(36)
+            tableName: task_results
+#        - modifyDataType:
+#            columnName: body
+#            newDataType: text
+#            tableName: task_results
```
german-muzquiz, line 33: Is this entry because of tests during development? If it should not be applied it would be better to be removed.

german-muzquiz (commented):
Overall looks nice, I only had a few questions

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -583,12 +586,25 @@ class SqlCache(
...
           }
-
-          onDuplicateKeyUpdate()
-            .set(field("application"), MySQLDSL.values(field("application")) as Any)
-            .set(field("body_hash"), MySQLDSL.values(field("body_hash")) as Any)
-            .set(field("body"), MySQLDSL.values(field("body")) as Any)
-            .set(field("last_updated"), MySQLDSL.values(field("last_updated")) as Any)
+            .run {
+              when (jooq.dialect()) {
+                SQLDialect.POSTGRES ->
```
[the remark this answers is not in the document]
ncknt: The list is endless. I can't see how this would scale without making this hard to read.

clouddriver-sql/src/main/resources/db/changelog/20190822-initial-event-schema.yml
ncknt: Correct. `bigint(20)` is the same as `bigint` with a hint of the size of the data for the client.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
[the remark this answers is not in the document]
german-muzquiz: I meant to move this dialect specific code to `SqlUtil`, not to implement all possible dialects. Having one path for postgres and other for everything else is good.

german-muzquiz (approved):

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -583,12 +586,25 @@ class SqlCache(
           chunk.forEach {
             values(it, sqlNames.checkAgentName(agent), apps[it], hashes[it], bodies[it], now)
           }
-
-          onDuplicateKeyUpdate()
-            .set(field("application"), MySQLDSL.values(field("application")) as Any)
-            .set(field("body_hash"), MySQLDSL.values(field("body_hash")) as Any)
-            .set(field("body"), MySQLDSL.values(field("body")) as Any)
-            .set(field("last_updated"), MySQLDSL.values(field("last_updated")) as Any)
+            .run {
```
robzienert, line 589: I'm confused what we're attaching `.run` to here and why? `forEach`? That returns a `Unit`. Looks like we can remove that entirely.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlConstraints.kt
```diff
@@ -16,12 +16,41 @@
...
+  fun getDefaultSqlConstraints(dialect: SQLDialect): SqlConstraints =
+    when(dialect) {
+      SQLDialect.POSTGRES ->
+        // https://www.postgresql.org/docs/current/limits.html
+        SqlConstraints(63, Int.MAX_VALUE, Int.MAX_VALUE)
+      else ->
+        SqlConstraints(
+          64,
+          // 352 * 2 + 64 (max rel_type length) == 768; 768 * 4 (utf8mb4) == 3072 == Aurora's max index length
+          352, 127
```
robzienert, line 53: nit: odd formatting here.

clouddriver-sql-postgres/clouddriver-sql-postgres.gradle
```diff
@@ -0,0 +1,6 @@
+dependencies {
+  implementation project(":cats:cats-sql")
+  implementation project(":clouddriver-sql")
+
+  runtimeOnly "org.postgresql:postgresql:42.2.18"
```
robzienert, line 5: Now that pg is being added into other services, we should probably look to move this to `spinnaker-dependencies` so the various services can be aligned.

robzienert (approved):
Diggin it. Great work.

clouddriver-sql-postgres/clouddriver-sql-postgres.gradle
ncknt: Yeah. There's a bit more follow up work to this. For instance I can't see `SqlTaskRepositoryTest` being executed - at least on build. I wonder if that's on purpose.

robzienert:
Whenever starting clouddriver, I'm now seeing these logs emitted:

```
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.608  WARN 64417 --- [           main] liquibase.changelog.DatabaseChangeLog    : [] modifyDataType will lose primary key/autoincrement/not null settings for mysql.  Use <sql> and re-specify all configuration if this is the case
2020-11-20 11:15:18.609  INFO 64417 --- [           main] liquibase.changelog.ChangeSet            : [] Continuing past: classpath:db/changelog/20201110-field-type-postgres.yml::change-field-types-postgresql::ncknt despite precondition failure due to onFail='CONTINUE': 
          classpath:db/changelog-master.yml : DBMS Precondition failed: expected postgresql, got mysql
```

It seems weird to me that these logs would be emitted, and then the changeset is continued. Two things:

1. Investigate these logs -- they are kind of concerning.
2. We should mark this changeset as completed so it's not continually re-processed by non-postgres deployments.

srekapalli:
@ncknt : We are seeing DeadLock's in our test environments with this change.

```com.mysql.cj.jdbc.exceptions.MySQLTransactionRollbackException: Deadlock found when trying to get lock; try restarting transaction
\tat org.jooq_3.13.2.MYSQL.debug(Unknown Source)
\tat org.springframework.jdbc.support.SQLExceptionSubclassTranslator.doTranslate(SQLExceptionSubclassTranslator.java:73)
\tat org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:72)
\tat org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:81)
\tat com.netflix.spinnaker.kork.sql.JooqToSpringExceptionTransformer.exception(JooqToSpringExceptionTransformer.kt:36)
\tat org.jooq.impl.ExecuteListeners.exception(ExecuteListeners.java:274)
\tat org.jooq.impl.AbstractQuery.execute(AbstractQuery.java:386)
\tat org.jooq.impl.AbstractDelegatingQuery.execute(AbstractDelegatingQuery.java:119)
\tat com.netflix.spinnaker.cats.sql.cache.SqlCache$storeAuthoritative$5.invoke(SqlCache.kt:602)
\tat com.netflix.spinnaker.cats.sql.cache.SqlCache$storeAuthoritative$5.invoke(SqlCache.kt:51)```

ncknt:
@srekapalli sorry I missed your message. The deadlock was addressed by https://github.com/spinnaker/clouddriver/pull/5114.

PR #5098, opened 2020-11-14
fix(docker): Support sortTagsByDate for SQL cache

KeisukeYamashita:
## WHAT
I know it's not recommended(ref: [Clouddriver: docker-registry --sort-tags-by-date bad performance, spinnaker/spinnaker](https://github.com/spinnaker/spinnaker/issues/5538)) to use this flag, but it's left to be used for small use cases.(Ref: https://github.com/spinnaker/spinnaker/issues/5538#issuecomment-617884816)

> After discussing this issue in Slack and at the Kubernetes SIG, we learned that users with smaller registries still find this flag valuable and are not hindered by these performance issues.

And I'm one of the users.However, it only supports Redis Cache. This time we supported Caching with SQL. Until this will get deprecated, I want this flag to support SQL.

## WHY
When I run with Clouddriver with SQL, I ran into this error. It didn't happen to me when I used with Redis.

```log
clouddriver java.lang.IllegalArgumentException: Cannot compare java.util.LinkedHashMap with value '{nano=338479230, epochSecond=1569038084}' and java.util.LinkedHashMap with value '{nano=143822191, epochSecond=15533792320}'
```

I think we can use `epochSecond` to compare.

## REF
* [perf(docker): Use parallel streams for caching docker images, spinnaker/clouddriver](https://github.com/spinnaker/clouddriver/pull/3926)


mattgogerly:
@Mergifyio update

mattgogerly (approved):

PR #5099, opened 2020-11-16
fix(kubernetes): k8s event timestamp should be null safe

titirigaiulian:
In some cases k8s api returns `lastTimestamp` with null value causing the `ManifestBuilder` to throw nullPointerException.
Made `lastEventTimestamp` optional and set default value for usecases when the key exists in the map but is null

Issue: https://github.com/spinnaker/spinnaker/issues/6181

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesManifestContainerBuilder.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/KubernetesManifestContainerBuilder.java
@@ -17,6 +17,8 @@
 
 package com.netflix.spinnaker.clouddriver.kubernetes.caching.view.provider;
 
+import static java.util.Comparator.*;
+
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.view.model.KubernetesManifestContainer;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.KubernetesPodMetric;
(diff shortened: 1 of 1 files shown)


ezimanyi (approved):
Thanks for fixing this!

One suggestion that might make the fix more robust.  It looks like the issue with the `null` is because of the `Comparator` below.

Another way to fix this would be to change below:
```
.sorted(Comparator.nullsLast(Comparator.comparing(lastEventTimestamp)))
```
which makes the comparator null-safe.  This might be a bit more clear than the "n/a" (which I realize is pre-existing already).

Definitely consider this an `Optional<Suggestion>`, this is fine to merge too!

titirigaiulian:
Hey @ezimanyi  thanks for the quick review. Changed a bit to use `Comprator` `nullsLast`, but I had to use it twice to catch also the nulls returned by the `lastEventTimestamp`. Using only `Comparator.nullLast...` will handle the `KubernetesManifest` nulls. 
Can you please take another look?

andreeagheorghescu (approved):

ezimanyi (approved):
This looks good to me!  Thanks!

titirigaiulian:
Can this be backported to 1.23?

rbadillo:
We need this fix in 1.23.x please

ezimanyi:
The release branch patch criteria are [here](https://spinnaker.io/community/contributing/releasing/#release-branch-patch-criteria); this would need to fix a documented regression and include tests demonstrating the regression.  If that's the case then there are instructions at that link for requesting a cherry-pick from the current release manager.

PR #5106, opened 2020-11-18
fix(kubernetes): Ensure overridden namespace is used for deploy manifest validation

nabuskey:
This fixes an issue where namespace override value is not used when deploying manifests. Fixes the first point mentioned in [this issue](https://github.com/spinnaker/spinnaker/issues/5992). This also fixes a case where deployment to overridden namespace fails when namespace specified in manifests is not in namespaces defined in target account. 

Note that this does not address the second point mentioned in the issue. Namespace validation passes if namespace is not specified in manifests.


german-muzquiz (commented):
The validator doesn't seem like a good place for mutating the manifest for setting the namespace override, because its responsibility is just to validate that the manifest is already correct.

Instead there is a [KubernetesDeployManifestConverter](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/converter/manifest/KubernetesDeployManifestConverter.java#L70 ) whose responsibility is to transform between the input request and the object used downstream to apply the manifest that seems like a better place to set the namespace override before validations are done.

Looking at [OperationsService](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java#L155) the converter is first called for `convertDescription`, then validations are run, and finally `convertOperation` is called. Probably `convertDescription` would need to mutate the input Map if the output description is not used for doing the deployment.

nabuskey:
Thanks for the feedback and detailed pointers. Highly appreciated. 
I was a little confused by how namespace override is implemented in `KubernetesDeployManifestOperation` but not in validator. I do not believe namespace mutation is needed in deploy operation any longer.

german-muzquiz (approved):
Thanks for the changes, looks good to me

PR #5107, opened 2020-11-18
fix(serverGroup): Set instanceType in view model for asgs with launch…

caseyhebebrand:
The `ServerGroupViewModel` only set the instance type if the ASG was backed by a launch config. This affected filtering by `instanceType`, since the type was undefined for any asg backed by a launch template.

Note: The `instanceType` was set in AWS at the launch template level, just not on Spinnaker's view model.

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
@@ -142,6 +142,9 @@ class ServerGroupController {
         sg.instanceType = serverGroup.launchConfig.instanceType
       }
     }
+    if (serverGroup.launchTemplate && serverGroup.launchTemplate.launchTemplateData) {
+      instanceType = serverGroup.launchTemplate.launchTemplateData.instanceType
+    }
     sg.account = cluster.accountName
 
     return sg
@@ -304,6 +307,11 @@ class ServerGroupController {
           instanceType = serverGroup.launchConfig.instanceType
         }
       }
+
+      if (serverGroup.launchTemplate && serverGroup.launchTemplate.launchTemplateData) {
+        instanceType = serverGroup.launchTemplate.launchTemplateData.instanceType
+      }
+
       if (serverGroup.tags) {
         tags = serverGroup.tags
       }


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -142,6 +142,9 @@ class ServerGroupController {
         sg.instanceType = serverGroup.launchConfig.instanceType
       }
     }
+    if (serverGroup.launchTemplate && serverGroup.launchTemplate.launchTemplateData) {
```
luispollo, line 145: Should this be an `else if` with the block above? Are launch config and launch templates mutually exclusive?

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -304,6 +307,11 @@ class ServerGroupController {
           instanceType = serverGroup.launchConfig.instanceType
         }
       }
+
+      if (serverGroup.launchTemplate && serverGroup.launchTemplate.launchTemplateData) {
```
luispollo, line 311: Same question.

luispollo (approved):
Small question inline, otherwise LGTM! Thanks for fixing this so quickly!

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
caseyhebebrand: Yes, the serverGroup will have either a launch config or a launch template, but never both.

ajordens: I believe we could add a `getInstanceType()` to `AmazonServerGroup` and make the determination there.

As it stands now, this is referencing an attribute `launchTemplate` that is not defined on the `ServerGroup` interface. Of course because it's groovy, it'll work for AmazonServerGroup but not for any other providers.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -142,6 +142,9 @@ class ServerGroupController {
         sg.instanceType = serverGroup.launchConfig.instanceType
       }
     }
+    if (serverGroup.launchTemplate && serverGroup.launchTemplate.launchTemplateData) {
+      instanceType = serverGroup.launchTemplate.launchTemplateData.instanceType
```
ajordens, line 146: `instanceType` isn't actually defined anywhere, I think you actually want `sg.instanceType` (but if there was a `getInstanceType()` on `AmazonServerGroup`, you wouldn't need this).

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ServerGroupController.groovy
```diff
@@ -304,6 +307,11 @@ class ServerGroupController {
           instanceType = serverGroup.launchConfig.instanceType
         }
       }
+
+      if (serverGroup.launchTemplate && serverGroup.launchTemplate.launchTemplateData) {
+        instanceType = serverGroup.launchTemplate.launchTemplateData.instanceType
```
ajordens, line 312: Slightly more complicated.

You could also do a `objectMapper.convertValue(serverGroup, Map)` here and just reference attributes out of a map. 

That would at least be consistent across providers.

PR #5108, opened 2020-11-19
feat(aws): Add filtering of LoadBalancers to include/exclude based on Tags

gavinbunney:
This provides a filtering mechanism to include/exclude LoadBalancers (both classic and ALB/NLBs) based on configured name/value pairs. This expands on the previous ASG filtering done in #5089.

Tags are fetched via a separate `DescribeTag` call, which has a limit of 20 arns/names.

This is needed to allow for a subset of AWS resources to be managed by Spinnaker (for our environment, exclude is enough, but added include for other use cases).


okoye:
👍

ajordens (approved):

PR #5109, opened 2020-11-19
fix(saga): In reality, the argument to `parseSagaResult` can be `null`

ajordens:
Uncovered this while implementing a saga in Kotlin which is a little more sticky about
arguments and return types that are marked `@Nonnull`.

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/sagas/AbstractSagaAtomicOperation.java
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/sagas/AbstractSagaAtomicOperation.java
@@ -64,7 +64,7 @@ public abstract class AbstractSagaAtomicOperation<T, SR, R>
    * Provides the opportunity to convert a {@link SagaAction.Result} into the expected result type
    * of the AtomicOperation.
    */
-  protected abstract R parseSagaResult(@Nonnull SR result);
+  protected abstract R parseSagaResult(SR result);
 
   @Override
   public R operate(List priorOutputs) {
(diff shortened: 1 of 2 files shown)


clouddriver-eureka/src/main/groovy/com/netflix/spinnaker/clouddriver/eureka/provider/agent/EurekaCachingAgent.groovy
```diff
@@ -93,7 +94,9 @@ class EurekaCachingAgent implements CachingAgent, HealthProvidingCachingAgent, C
   @Override
   CacheResult loadData(ProviderCache providerCache) {
     log.info("Describing items in ${agentType}")
-    EurekaApplications disco = retry.retry({ eurekaApi.loadEurekaApplications() }, 3, 100, false)
+    EurekaApplications disco = AuthenticatedRequest.allowAnonymous({
```
ajordens, line 97: Unrelated but was clogging up my logs.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/sagas/AbstractSagaAtomicOperation.java
```diff
@@ -64,7 +64,7 @@ public AbstractSagaAtomicOperation(T description) {
    * Provides the opportunity to convert a {@link SagaAction.Result} into the expected result type
    * of the AtomicOperation.
    */
-  protected abstract R parseSagaResult(@Nonnull SR result);
+  protected abstract R parseSagaResult(SR result);
```
ajordens, line 67: This can totally be null right? (I think more often than not it's actually going to be null)

robzienert: Ah, yeah. That can be null, yes.

robzienert (approved):

PR #5110, opened 2020-11-19
fix(serverGroup): Fix instanceType assignment

caseyhebebrand:


ajordens:
Duplicated by spinnaker/clouddriver#5111

PR #5112, opened 2020-11-20
fix(saga): Shift placement of `@JsonIgnore` so that the attribute is actually ignored

ajordens:


ajordens:
Hmm. 

Worked in actual running code, looking at specs now.

cfieber (approved):

PR #5114, opened 2020-11-20
fix(sql): Remove accidental double execute on insert

robzienert:
As written, cache inserts will deadlock themselves.

Actual `execute` is called at L606.


ncknt (approved):

christopherthielen (approved):
nice catch

srekapalli (approved):

PR #5119, opened 2020-11-25
fix(cf): fix error description with v2 routes/route-mappings

zachsmith1:
This problem caused server groups to not deploy and error out if you had a route in your manifest. It also rendered the map/unmap route stages useless. This change needs to be backported to 1.23, 1.22. I spent a decent amount of time trying to figure out why this worked in previous versions and was unable to find the causation. I've added tests to cover the new logic and ensure proper mapping. This class will eventually go away as we move towards okhttp3 and dynamic api binding for services based on the supported cf version.


kevinawoo (approved):

dogonthehorizon:
@mergifyio backport release-1.22.x release-1.23.x

zachsmith1:
Fixes this -> https://github.com/spinnaker/spinnaker/issues/6163

PR #5124, opened 2020-11-30
feat(ecs): Improve task definition cache

allisaurus:
* cache only task definitions associated with ECS services

Addresses https://github.com/spinnaker/spinnaker/issues/6063

This change reduces the number of unnecessary cache entries held by Spinnaker and, as a result, the number of `ecs:*TaskDefinition` API calls needed to populate the cache. 

### Testing

Testing in an account with ~1000 task definitions (not necessarily associated w/ services) shows a reduction in the # of calls made to ECS during regular caching agent execution.

* 0 `DescribeTaskDefinition` calls were made during registrations of task defs NOT associated w/ a service (vs. ~1000 w/ previous version)
* 0 `ListTaskDefinition` calls made during a 30 window where no deployments were happening (vs. ~300 w/ previous version).
* Pipeline deployments succeed in slightly shorter (maybe negligible) amount of time (NOTE: test account only held ~10 services)

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/AbstractCacheClient.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/AbstractCacheClient.java
@@ -24,8 +24,11 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 abstract class AbstractCacheClient<T> {
+  private final Logger log = LoggerFactory.getLogger(getClass());
 
   private final String keyNamespace;
   protected final Cache cacheView;
@@ -89,6 +92,7 @@ abstract class AbstractCacheClient<T> {
    * @return
    */
   private Collection<CacheData> fetchFromCache(String account, String region) {
+    log.debug("fetching all for account '{}' and region '{}'", account, region);
     String accountFilter = account != null ? account + Keys.SEPARATOR : "*" + Keys.SEPARATOR;
     String regionFilter = region != null ? region + Keys.SEPARATOR : "*" + Keys.SEPARATOR;
     Set<String> keys = new HashSet<>();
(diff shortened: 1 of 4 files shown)


clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/AbstractCacheClient.java
```diff
@@ -89,6 +92,7 @@ public T get(String key) {
    * @return
    */
   private Collection<CacheData> fetchFromCache(String account, String region) {
+    log.debug("fetching all for account '{}' and region '{}'", account, region);
```
piradeepk, line 95: not a necessity of this PR, but it would be cool if we could add some more debug logging below for things like the number of items in allData or nameMatches?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskDefinitionCachingAgent.java
```diff
@@ -90,71 +81,55 @@ public String getAgentType() {
...
-      for (String taskDefinitionArn : newTaskDefArns) {
-        TaskDefinition taskDefinition =
-            ecs.describeTaskDefinition(
-                    new DescribeTaskDefinitionRequest().withTaskDefinition(taskDefinitionArn))
+      // TaskDefinitions are immutable, there's no reason to
+      // make a describe call on existing ones.
+      TaskDefinition cacheEntry = retrieveFromCache(arn, providerCache);
+
+      if (cacheEntry != null) {
+        taskDefinitions.add(cacheEntry);
```
piradeepk, line 106: Is it possible that the cacheEntry is stale?

piradeepk (commented):
This is awesome!!!

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/AbstractCacheClient.java
allisaurus: I wanted to be conservative re: adding more here, since they'll apply to every caching agent (which also have logging of their own), but I agree some more strategic debug statements (here or elsewhere in the caching layer) would be good.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskDefinitionCachingAgent.java
allisaurus: Meaning, is it possible for the cached task def to no longer match the task def in the account w/ the same ARN? 
It is not! Task defs are immutable. Even if you deregister a revision in a family, and then register a new one in the same family, it will be created with a revision number incremented higher than the latest revision, active or not.

piradeepk: Good point! So different revisions will be considered different task defs!

piradeepk (approved):
Ship it! 💯

PR #5125, opened 2020-12-01
fix(kubernetes): fix for clouddriver's high loading time with SCC

kirangodishala:
Fix for issue #[6200](https://github.com/spinnaker/spinnaker/issues/6200)

With dynamic accounts using Spring cloud config and large number of accounts, it is well observed that the clouddriver is taking a lot of time to start. And the reason for this is the way SpringBoot binds thousands of properties to KubernetesConfigurationProperties object. Per-account-binding-time is increasing with the increase in number of kubernetes accounts.  So this fix is to avoid the SpringBoot's binding process and go with a custom constructor which takes a map of all accounts' properties, parses the keys and bind then to the KubernetesConfigurationProperties object.

This fix extends support to existing "configserver:" feature to fetch kubeconfig file but doesn't improve the time taken in this process which is approximately 3-4 seconds per fetch-call. However, a caching mechanism is added for an already fetched kubeconfig file path so that any other account using the same kubeconfig file doesn't go through the whole fetching process again.
 
This is an opt-in fix through a property: `custom.kubernetes.constructor.enabled`


emagana:
This is very important for Salesforce's use cases. We have over 2000 provider accounts and it takes hours for clouddriver to boot and even longer to dynamically load new provider accounts. Asking for your kind review @robzienert  @ezimanyi @maggieneterval @german-muzquiz  Thank you in advaced!

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/config/KubernetesCustomBinderConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+import com.netflix.spinnaker.clouddriver.kubernetes.config.KubernetesPropertiesMapExtractor;
+import com.netflix.spinnaker.clouddriver.model.PropertiesMapExtractor;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.cloud.context.config.annotation.RefreshScope;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConditionalOnProperty({"kubernetes.enabled", "custom.kubernetes.constructor.enabled"})
```
german-muzquiz, line 30: We put all properties which are exclusive to kubernetes under `kubernetes` key, so probably a more appropriate name would be something like `kubernetes.manualPropertyBindingEnabled`, which indicates that binding of properties is done manually compared to spring's default behavior.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.java
```diff
@@ -75,4 +85,93 @@ public void validate() {
...
+          if (propertyName.equals("kubeconfigFile")) {
+            if (((String) value).startsWith("configserver:")) {
+              value = propertiesMapExtractor.resolveConfigServerFilePath((String) value);
+            }
+          }
+          accountPropertiesMap.put(
+              entry.getKey().substring(entry.getKey().indexOf(propertyName)), value);
+        }
+      } else {
+        log.debug("Ignoring the properties having blank values. Key:" + entry.getKey());
```
german-muzquiz, line 169: Seems that this code ignores all properties not belonging to accounts. What happens for properties at the `kubernetes` top level like `rawResourcesEndpointConfig`?

clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationPropertiesSpec.groovy
```diff
@@ -0,0 +1,144 @@
...
+    def ACC_1_OMIT_NS_1 = "acc-1-ns-1"
+    def ACC_1_KIND_0 = "role"
+    def ACC_1_KIND_1 = "apiService"
+    def ACC_1_KUBECONFIG = "/tmp/config1"
+    def ACC_1_CRD_0_K8S_KIND = "deploymentConfig"
+    def ACC_1_CRD_0_NAMESPACED = false
+    def ACC_1_CRD_1_K8S_KIND = "Route"
+    def ACC_1_CRD_1_VERSIONED = true
+    def map = new ImmutableMap.Builder()
+      .put("kubernetes.accounts[0].name", ACC_0_NAME)
```
german-muzquiz, line 61: Can you add `rawResourcesEndpointConfig` section as well?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesPropertiesMapExtractor.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+    if (cacheContainsKey(key)) {
+      filePath = configServerCache.get(key);
+      if (resourceExist(filePath)) {
+        return filePath;
+      }
+    }
+    filePath = configResourceService.getLocalPath(key);
+    addToCache(key, filePath);
+    return configResourceService.getLocalPath(key);
```
german-muzquiz, line 89: Can be simplified to `return filePath`.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesPropertiesMapExtractor.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private boolean resourceExist(String filePath) {
+    return Path.of(filePath).toFile().isFile();
```
german-muzquiz, line 93: Seems that this needs to check for existence (`exists()`) as well

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.java
```diff
@@ -75,4 +85,93 @@ public void validate() {
       }
     }
   }
+
+  public KubernetesConfigurationProperties() {}
+
+  /**
+   * For larger number of Kubernetes accounts, SpringBoot based properties binding is taking tens of
```
german-muzquiz, line 92: I think it's worth to document the limitations of this approach (in case they apply):
* Property placeholders don't work (for example you cannot do `kubeconfigFile: ${KUBECONFIGS_HOME}/kubecfg-dev`)
* `encryptedFile:` and `encrypted:` notation doesn't work
* `configserver:` only works for kubeconfig files

german-muzquiz (commented):
I see that this way of property binding involves manually inspecting account properties and having different logic for special cases like `permissions` and `kubeconfigFile`, which could make the code a bit fragile if the custom logic needs to be adjusted for changes in account configuration.

Have you considered binding properties with `Yaml` and/or `ObjectMapper`? The essential problem remains on how to transform a `yaml` text file into a java object without using spring expensive property binding, and those libraries have already been battle tested to handle different scenarios. Creating a `KubernetesConfigurationProperties` from a `.yaml` file can be done using `Yaml` to convert to a nested `Map<String, Object>` and then `ObjectMapper` to convert from that to the final java object. I don't know how that compares with spring binding or with this solution regarding performance, which is the main concern.

I'm just trying to exhaust all possibilities before going with manual binding and custom logic per property. Whatever the final solution is, I think it's worth to add an integration test case to ensure we notice if the code breaks unintentionally in the future (a new class like [BaseTest](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/BaseTest.java)  loading a different `clouddriver.yml`). The test can just make a call to `/credentials?expand=true` and verify that the account has the right properties.

kirangodishala:
Thank you @german-muzquiz for the detailed review comments. I will fix them and update the PR.  
I would have very well preferred a yml parser if my input had been a yml file. But SCC converts yml to a map in the bootstrap context itself. So we get only the map to deal with.

ezimanyi:
I definitely agree with @german-muzquiz here; while I think the performance improvements are 💯 , I worry a bit about the maintainability of the solution.

I like @german-muzquiz 's idea of using an `ObjectMapper`.  If I'm understanding correctly, the main performance issue is that Spring boot performs very poorly when parsing the long list of accounts (but actually deserializing each account itself is reasonably fast).  Would there be some way to have this custom code just handle parsing the list, then call an `ObjectMapper` to handle each individual account?

kirangodishala:
> I like @german-muzquiz 's idea of using an `ObjectMapper`. If I'm understanding correctly, the main performance issue is that Spring boot performs very poorly when parsing the long list of accounts (but actually deserializing each account itself is reasonably fast). Would there be some way to have this custom code just handle parsing the list, then call an `ObjectMapper` to handle each individual account?

Thank you @ezimanyi. I will try this ObjectMapper-approach.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/config/KubernetesCustomBinderConfiguration.java
kirangodishala: renamed the property to kubernetes.customPropertyBindingEnabled as suggested.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.java
kirangodishala: Now rawResourcesEndpointConfig is included.

clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationPropertiesSpec.groovy
kirangodishala: added `rawResourcesEndpointConfig` section is added in the unit test case.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesPropertiesMapExtractor.java
kirangodishala: corrected now.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesPropertiesMapExtractor.java
kirangodishala: I verified again, it's working fine as-is.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.java
kirangodishala: Included all the limitations in the BootstrapKubernetesConfigurationProvider class.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.java
```diff
@@ -17,13 +17,23 @@
...
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import lombok.Data;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.BeanWrapper;
+import org.springframework.beans.BeanWrapperImpl;
 
 @Data
+@Slf4j
```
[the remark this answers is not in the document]
kirangodishala: Removed @Slf4j annotation and no logger is used in KubernetesConfigurationProperties class now.

kirangodishala:
@ezimanyi  @german-muzquiz  I have replaced the regex based custom logic with ObjectMapper and json unflattener. Now the code should be maintainable better. I have renamed couple of files to match the new changes. I request you to review the code again and provide your comments.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/BootstrapKubernetesConfigurationProvider.java
```diff
@@ -0,0 +1,148 @@
...
+      }
+    }
+
+    throw new RuntimeException("No BootstrapPropertySource found!!!!");
+  }
+
+  private String resolveConfigServerFilePath(String key) {
+    String filePath;
+
+    if (configResourceService == null) {
```
ezimanyi, line 112: Why are we manually fetching this bean here when we access it instead of autowiring as we do above for `ConfigurableApplicationContext applicationContext`?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/BootstrapKubernetesConfigurationProvider.java
```diff
@@ -0,0 +1,148 @@
...
+            JsonUnflattener.unflattenAsMap(kubernetesPropertiesMap).get("kubernetes");
+
+    KubernetesConfigurationProperties kubernetesConfigurationProperties =
+        objectMapper.convertValue(propertiesMap, KubernetesConfigurationProperties.class);
+
+    kubernetesConfigurationProperties
+        .getAccounts()
+        .forEach(
+            acc -> {
+              if (acc.getKubeconfigFile().startsWith("configserver:")) {
```
ezimanyi, line 80: I'm not sure I understand why we need to post-process only this field on the account?  Could you please briefly explain (and maybe add a comment for this).

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/BootstrapKubernetesConfigurationProvider.java
```diff
@@ -0,0 +1,148 @@
...
+            });
+    return kubernetesConfigurationProperties;
+  }
+
+  @SuppressWarnings("unchecked")
+  private Map<String, Object> getPropertiesMap() {
+    ConfigurableEnvironment environment = applicationContext.getEnvironment();
+    Map<String, Object> map;
+
+    for (PropertySource<?> propertySource : environment.getPropertySources()) {
```
ezimanyi, line 92: If the environment has multiple property sources, it looks like we will be returning immediately on the first here.  What happens in the case that there is a `spinnaker.yml`, a `clouddriver.yml`, and `clouddriver-local.yml`.  It seems like this logic will not be properly merging these (and depending on the order will just pick one).

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/config/KubernetesCustomBinderConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+import com.netflix.spinnaker.clouddriver.kubernetes.config.KubernetesConfigurationProperties;
+import com.netflix.spinnaker.clouddriver.kubernetes.config.KubernetesConfigurationProvider;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.cloud.context.config.annotation.RefreshScope;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConditionalOnProperty({"kubernetes.enabled", "kubernetes.customPropertyBindingEnabled"})
```
ezimanyi, line 30: this needs to be `kebab-case`, `kubernetes.custom-property-binding-enabled`

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.java
```diff
@@ -75,4 +77,21 @@ public void validate() {
...
+  public KubernetesConfigurationProperties() {}
+
+  /**
+   * For larger number of Kubernetes accounts, SpringBoot based properties binding is taking tens of
+   * minutes, hence this custom constructor for KubernetesConfigurationProperties is needed.
+   *
+   * @param kubernetesConfigurationProvider This provides the KubernetesConfigurationProperties
+   *     object
+   */
+  public KubernetesConfigurationProperties(
```
ezimanyi, line 90: I'm not sure I understand this constructor...it looks like it is never setting the accounts so the resulting properties will always have an empty list of accounts?

In general, I'm not sure why we need this constructor, which creates another `KubernetesConfigurationProperties` inside the constructor.  Couldn't we just replace every call to `KubernetesConfigurationProperties(kubernetesConfigurationProvider)` with kubernetesConfigurationProvider. getKubernetesConfigurationProperties() and avoid the extra call?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProvider.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.clouddriver.kubernetes.config;
+
+public interface KubernetesConfigurationProvider {
```
ezimanyi, line 20: Do we need this interface with only one implementation?

ezimanyi (commented):
Thanks for the updates!  It's definitely a lot clearer and less fragile to use `ObjectMapper` than the custom logic from before.  That being said, I still feel a bit uneasy that there are all those limitations documented in `BootstrapKubernetesConfigurationProvider` (which may also become out of date as we add more logic and if Spring changes things).

Have you looked into calling the Spring [Binder](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/context/properties/bind/Binder.html) class to do the binding, which should remove some of the limitations?  (Of course it may also remove some of the performance benefit in which case it's not a great solution.)

Maybe it would be helpful to include (either here or on the linked issue) details of where exactly the slowness is coming from Spring (if you did profiling) so that the fix here can minimally avoid the slow logic while still re-using as much of what Spring gives us for free as possible.

ezimanyi:
(To clarify the above comment...you'll definitely need to re-implement parts of Spring to fix the issue, so definitely the solution will require that.  I'm just trying to make sure this re-implements the minimal part.  Great work so far!)

ncknt:
To add a bit of context, we've added the ability to make [credentials loading pluggable](https://github.com/spinnaker/governance/blob/master/rfc/account-management.md) for that use case. It is currently supported in 1.23 for Kubernetes and CloudFoundry. AWS and Appengine have been added for 1.24. As an example, you can achieve what you want today with a plugin (e.g. https://github.com/armory-plugins/external-accounts or just write your own - the code is very simple).

Trying to rewrite Spring boot loaders doesn't seem like the best route as it's not designed to load tons of data. If you want to do so, I'd recommend doing it in a plugin where the necessary interfaces are public and extensible and not make the code less readable.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/BootstrapKubernetesConfigurationProvider.java
kirangodishala: Now I autowired it.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/BootstrapKubernetesConfigurationProvider.java
kirangodishala: I though only kubeconfigFile would need the configserver feature but now I extended the logic to all the fields.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/BootstrapKubernetesConfigurationProvider.java
kirangodishala: This is the same behaviour as existing clouddriver. If there are multiple files having kubernetes accounts, only the one having highest priority wins. The accounts don't get merged from different files.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/config/KubernetesCustomBinderConfiguration.java
kirangodishala: I updated with kabab casing

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.java
kirangodishala: Agreed and removed the constructor from KubernetesConfigurationProperties.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProvider.java
kirangodishala: Removed the interface.

kirangodishala:
> (To clarify the above comment...you'll definitely need to re-implement parts of Spring to fix the issue, so definitely the solution will require that. I'm just trying to make sure this re-implements the minimal part. Great work so far!)

@ezimanyi I have implemented the binding using SpringBoot's Binder class. I could see the performance gain when processing the properties at ManagedAccount level rather than at KubernetesConfigurationProperties level. All the limitations are overcome now and code should be good to go. Please review and provide the feed back.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/BootstrapKubernetesConfigurationProvider.java
ezimanyi: Sounds good as long as this has been tested for cases where the kubernetes accounts live in any of the above files.  Definitely I realize currently there's no merging behavior, but currently I could put my accounts in any of `spinnaker.yml`, `clouddriver.yml`, or `clouddriver-local.yml` as long as I only pick one, and it will work.  What I was unsure of is whether after these changes only the highest-priority source would be used.  For example, if `clouddriver-local.yml` is highest priority and does not define kubernetes accounts, will we still use the ones from `clouddriver.yml`?

ezimanyi:
@kirangodishala : Thanks for the updates, particularly to scope down the amount that this is re-implementing logic that exists in Spring boot to the minimum.  As I noted in the SIG today, given that I'm stepping down as SIG lead and away from the Spinnaker project I'll leave it to the current SIG leads to move forward with this PR (and discuss whether an approach like @ncknt suggests might be better).

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.java
```diff
@@ -19,8 +19,7 @@
 import com.google.common.base.Strings;
 import com.netflix.spinnaker.credentials.definition.CredentialsDefinition;
 import com.netflix.spinnaker.fiat.model.resources.Permissions;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
```
german-muzquiz, line 22: Was the import change unintended?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/BootstrapKubernetesConfigurationProvider.java
```diff
@@ -0,0 +1,206 @@
...
+    // unflatten
+    Map<String, Object> propertiesMap =
+        (Map<String, Object>)
+            JsonUnflattener.unflattenAsMap(kubernetesPropertiesMap).get("kubernetes");
+
+    // loop through each account and bind
+    for (Map<String, Object> unflattendAcc :
+        ((List<Map<String, Object>>) propertiesMap.get("accounts"))) {
+      result =
+          bind(getFlatMap(unflattendAcc), KubernetesConfigurationProperties.ManagedAccount.class);
```
german-muzquiz, line 87: If I uderstand the code correctly, this is aproximately the sequence of transformations taking place:
```
Flattened Map<String, Object> --> Unflattened Map<String, Object> -->  (per account) String --> Flattened Map<String, Object> --> Binder --> Target class
```
Seems that the intermediate transformations are only used for handling `encrypted|encryptedFile|configserver` values by the `resolveSpecialCases` function.

In this case it's possible to use `ObjectMapper` to handle the special cases, and the transformation sequence could be simplied to:

```
Flattened Map<String, Object> --> Unflattened Map<String, Object> --> Target class
```

By only using `ObjectMapper` instead of `Binder` the code could look like:

```
Map<String, Object> unflattenedMap =
         (Map<String, Object>)
             JsonUnflattener.unflattenAsMap(kubernetesPropertiesMap).get("kubernetes");
KubernetesConfigurationProperties result = mapper.convertValue(unflattenedMap, KubernetesConfigurationProperties.class);
```

Of course individual account deserialization can be done with the same approach if that's a performance issue, but I haven't seen performance problems deserializing hundreds of accounts by `ObjectMapper` alone.

You can find an example of how to configure `ObjectMapper` to resolve special cases like `encrypted` and environment variables placeholders [here](https://github.com/armory-plugins/external-accounts/blob/master/clouddriver/src/main/java/io/armory/plugin/eap/loaders/URLCredentialsLoader.java#L81), the same approach can be done for `configserver`.

.idea/gradle.xml
```diff
@@ -1,11 +1,54 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="GradleMigrationSettings" migrationVersion="1" />
   <component name="GradleSettings">
     <option name="linkedExternalProjectsSettings">
       <GradleProjectSettings>
         <option name="distributionType" value="DEFAULT_WRAPPED" />
         <option name="externalProjectPath" value="$PROJECT_DIR$" />
-        <option name="useAutoImport" value="true" />
```
german-muzquiz, line 8: Not sure if the changes in `.idea` folder are intended but usually these files are not modified. For example this one could indicate that gradle auto import feature will be disabled in all developer's environments. I think it's safer to add `.idea` to `.gitignore` in general but that's another discussion.

german-muzquiz (commented):
@kirangodishala the code starts looking better, thanks!
There's still a few comments pending. The main one being around using intermediate transformations to handle special cases, which could be simplified by registering custom deserializers to `ObjectMapper`.

As Nico said, with the introduction of `kork-credentials` in newer spinnaker versions now it's possible to implement dynamic accounts in different ways than using configserver, which we know has a _lot_ of performance problems. The linked plugin is an example of replacing configserver entirely and still get the same and more features.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/BootstrapKubernetesConfigurationProvider.java
kirangodishala: >  you'll definitely need to re-implement parts of Spring to fix the issue, so definitely the solution will require that. I'm just trying to make sure this re-implements the minimal part 

@german-muzquiz I received the above review comment earlier hence I changed the code from using `ObjectMapper` to Spring Binder. The problems with SpingBoot binding surface when the whole `KubernetesConfigurationProperties` is supplied to SpringBoot instead of `ManagedAccount`s. So by using `Binder` class itself, we are achieving the performance just by changing the way we supply objects to it. This way I am able to keep the re-implementation to the minimum.

.idea/gradle.xml
kirangodishala: I agree that the files under `.idea` should be out of the repo. Anyway, I reverted the line added to this file by IntelliJ.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.java
kirangodishala: Yes. Reverted now.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/BootstrapKubernetesConfigurationProvider.java
```diff
@@ -0,0 +1,206 @@
...
+      throw new RuntimeException(
+          "Error occurred while building KubernetesConfigurationProperties object: "
+              + e.getMessage());
+    }
+  }
+
+  @SuppressWarnings("unchecked")
+  private Map<String, Object> getPropertiesMap() {
+    ConfigurableEnvironment environment = applicationContext.getEnvironment();
+    Map<String, Object> map;
```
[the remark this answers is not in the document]
kirangodishala: Now that variable is in use to check whether the remote file is in fact having the k8s accounts.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/BootstrapKubernetesConfigurationProvider.java
```diff
@@ -0,0 +1,211 @@
...
+
+    for (PropertySource<?> propertySource : environment.getPropertySources()) {
+
+      if (propertySource instanceof BootstrapPropertySource) {
+        map = (Map<String, Object>) propertySource.getSource();
+        if (map.containsKey(FIRST_ACCOUNT_NAME_KEY)) {
```
german-muzquiz, line 150: The code returns on the first property source that has defined an account, instead of looking through the priority of `clouddriver-local.yml`, `clouddriver.yml`, `spinnaker-local.yml`, `spinnaker.yml`. How do you know that the right file is parsed first?

There may be existing spinnaker installations having accounts in `.hal/config` and accounts in `default/profiles/clouddriver-local.yml`, for which this code seems to have 50% chances of picking the right set of accounts.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/BootstrapKubernetesConfigurationProvider.java
german-muzquiz: I agree on the neccesity to reimplement parts of spring as part of this PR.
The issue I'm seeing is that you could have a big file having 1,000 k8s accounts, and you're doing several transformations of all the accounts that are not really needed, so instead of transforming from A to B you're transforming between intermediate types using the `getFlatMap` and `bind` methods.

`ObjectMapper` is able to transform between any given 2 types, so it very well can transform from `Map<String, Object>` of individual accounts to `ManagedAccount` like this:

```
Map<String, Object> unflattenedMap =
         (Map<String, Object>)
             JsonUnflattener.unflattenAsMap(kubernetesPropertiesMap).get("kubernetes");
List<Map<String, Object>> accountsMap = unflattenedMap.get("accounts");
List<ManagedAccount> acounts = accountsMap.stream()
  .map(a -> mapper.convertValue(a, ManagedAccount.class))
  .collect(Collectors::toList);
```

In this way we're still processing individual accounts instead of the full `KubernetesConfigurationProperties`, and we save the processing time of converting from unflatenned map to string and back to flatenned map that happens in `getFlatMap`.

german-muzquiz (commented):
Thanks for the changes, there are still a couple of comments to address.

kirangodishala:
> The code returns on the first property source that has defined an account, instead of looking through the priority of clouddriver-local.yml, clouddriver.yml, spinnaker-local.yml, spinnaker.yml. How do you know that the right file is parsed first?
> 
> There may be existing spinnaker installations having accounts in .hal/config and accounts in default/profiles/clouddriver-local.yml, for which this code seems to have 50% chances of picking the right set of accounts.

The code behaves in the same way as the existing. The getPropertySources() on environment provides the sources in the order of priority so the code checks for accounts in that order and returns the first property source having accounts.

kirangodishala:
> german-muzquiz 16 days ago Contributor
> I agree on the neccesity to reimplement parts of spring as part of this PR.
> The issue I'm seeing is that you could have a big file having 1,000 k8s accounts, and you're doing several transformations of all the accounts that are not really needed, so instead of transforming from A to B you're transforming between intermediate types using the getFlatMap and bind methods.
> 
> ObjectMapper is able to transform between any given 2 types, so it very well can transform from Map<String, Object> of individual accounts to ManagedAccount like this:
> 
> Map<String, Object> unflattenedMap =
>          (Map<String, Object>)
>              JsonUnflattener.unflattenAsMap(kubernetesPropertiesMap).get("kubernetes");
> List<Map<String, Object>> accountsMap = unflattenedMap.get("accounts");
> List<ManagedAccount> acounts = accountsMap.stream()
>   .map(a -> mapper.convertValue(a, ManagedAccount.class))
>   .collect(Collectors::toList);
> In this way we're still processing individual accounts instead of the full KubernetesConfigurationProperties, and we save the processing time of converting from unflatenned map to string and back to flatenned map that happens in getFlatMap.

Using Spring Binder we get the features like relaxed binding which is not possible with ObjectMapper. I verified the processing time with 1500 accounts and it's still taking seconds (not minutes) and the time time taken for intermediate conversions is negligible.

german-muzquiz (approved):

PR #5128, opened 2020-12-02
fix(artifacts/gitRepo): use a try-with-resources block so Git objects…

dbyron0:
Easier to see when [ignoring whitespace](https://github.com/spinnaker/clouddriver/pull/5128/files?w=1)


ezimanyi (approved):
Look good, thanks!

 I guess one possible alternative way of solving this might have been to pull the declaration of `Git localrepository...` up into the existing `try` to reduce the nested try blocks:

```
try (Closeable ignored = ...; Git localRepository = ...) {
    // ...
    // return ...
} catch (GitAPIException e) {
    // throw ...
}
```

But just a minor style point, approving so feel free to ignore.

dbyron0:
Fair point.  I like reducing the nested try blocks, but I like limiting the catch of GitAPIException to the code that actually throws it.  I was half-tempted to move the catching of GitAPIException into the clone and archiveToOutputStream methods to get more specific error messages...and perhaps now I'm more tempted...

dbyron0:
Gonna leave this as it is for now...and will revisit if I end up back here.

PR #5135, opened 2020-12-04
feat(artifacts): dynamic artifact credentials via kork-credentials

ncknt:
This implements dynamic credentials for artifacts. Each type of credentials gets its own `CredentialsRepository` that can be adapted (dynamically reloaded, reading non Spring resources, etc.). `ArtifactCredentialsRepository` is now a simple `CompositeCredentialsRepository`.

Credentials types are defined via `CredentialsTypeProperties` and processed via a `CredentialsTypeBaseConfiguration` that creates all the required beans.

As a side note, credential types are `artifacts-<type>` because they can be dynamically reloaded via `credentials.poller.types.artifacts-<type>.reloadFrequencyMs` - using a dash works better than dot or slash even though slash seems more natural.


robzienert (approved):
Looks like a nice improvement in code complexity, do like. :)

PR #5136, opened 2020-12-05
refactor(exceptions): use exception handling in kork now that it's there

dbyron0:
see https://github.com/spinnaker/kork/pull/821 and https://github.com/spinnaker/spinnaker/issues/5473


ezimanyi (approved):
💯

PR #5137, opened 2020-12-06
feat(artifacts): throw a non-retryable exception when an artifact is …

dbyron0:
…misconfigured

Introduce non-retryable exception ArtifactCredentialsException and use it in
ArtifactCredentialsRepository, so there's enough information in the response to the fetch
endpoint for callers to not retry.


dbyron0:
I've got more digging to do here.  Please don't merge.

PR #5138, opened 2020-12-07
feat(authz): Authorize descriptions from list of DescriptionAuthorizer objects; refactor authz logic out to DescriptionAuthorizerService

jonsie:
This allows for additional `DescriptionAuthorizer` implementations for any necessary custom description authz.

Refactors authz logic out to `DescriptionAuthorizerService ` and adds ability to perform authz check against specific resource types.


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
```diff
@@ -180,8 +180,7 @@ public OperationsService(
                           allowedAccountValidators.forEach(
                               it -> it.validate(username, allowedAccounts, description, errors));
 
-                          // TODO(rz): Assert `description` is T
-                          descriptionAuthorizer.authorize(description, errors);
+                          descriptionAuthorizers.forEach(it -> it.authorize(description, errors));
```
ajordens, line 183: It would be nice if a `DescriptionAuthorizer` had a better means of indicating successful auth, rather than just mutating the passed in `Errors` object.

_or_

Perhaps this changes to `findFirst(it.supports(description)).authorize(description, errors)`. For this to work, we'd need to mark the existing `DefaultDescriptionAuthorizer` with the lowest priority.

That would avoid a need to explicitly clean any errors up?

cfieber: I like the findFirst approach

clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/deploy/ValidationErrors.java
```diff
@@ -33,4 +33,7 @@
   void rejectValue(String field, String errorCode, String defaultMessage);
 
   void rejectValue(String field, String errorCode, Object[] errorArgs, String defaultMessage);
+
+  /** Remove error by error code. */
+  void remove(String errorCode);
```
cfieber, line 38: I'm not seeing any use of this aside from in a test - what is the use case here?

jonsie: The idea was we would use it a -nflx DescriptionAuthorizer.  However, I'm trying out the idea of having the DescriptionAuthorizer return an errors object so we don't have to remove an `ObjectError` error.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/OperationsService.java
jonsie: Yeah.  findFirst sounds good to me :+1:

cfieber (approved):

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/deploy/DescriptionAuthorizerService.java
```diff
@@ -0,0 +1,155 @@
...
+    this.skipAuthorizationId = registry.createId("authorization.skipped");
+    this.missingApplicationId = registry.createId("authorization.missingApplication");
+    this.authorizationId = registry.createId("authorization");
+  }
+
+  public void authorize(Object description, Errors errors) {
+    authorize(description, errors, null);
+  }
+
+  public void authorize(Object description, Errors errors, ResourceType resourceType) {
```
ajordens, line 54: Thoughts on this taking a list of resource types to authorize rather than just a single one/null?

jonsie: Yep, dig it.

ajordens (approved):
LGTM -- Thanks for the revisions here.

PR #5141, opened 2020-12-07
chore(artifacts): add test to demonstrate current behavior when fetch…

dbyron0:
…ing a misconfigured artifact

A subsequent PR to change the response code to 400 is on its way.  Ideally that would also indicate to code that uses [RetrySupport](https://github.com/spinnaker/kork/blob/c58b660cfebd225a5c2a779b45ce1be3808ec11f/kork-exceptions/src/main/java/com/netflix/spinnaker/kork/core/RetrySupport.java#L37) that there's no point in retrying when artifacts are misconfigured, but I don't have that worked out just yet.

### DIFF
--- a/clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/controllers/ArtifactControllerSpec.java
+++ b/clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/controllers/ArtifactControllerSpec.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2020 Avast Software, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


clouddriver-web/src/test/groovy/com/netflix/spinnaker/clouddriver/controllers/ArtifactControllerSpec.java
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.http.MediaType;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.test.web.servlet.MvcResult;
+import org.springframework.web.context.WebApplicationContext;
+
+@RunWith(SpringJUnit4ClassRunner.class)
```
ezimanyi, line 43: For new Java tests we'd generally prefer JUnit5.  I think the only changes would be replacing this with:
```
@ExtendWith(SpringExtension.class)
@WebAppConfiguration
```
and changing the `org.junit` imports to the `org.junit.jupiter.api` ones.

dbyron0: Done in https://github.com/spinnaker/clouddriver/pull/5141/commits/a7637c76a65b65448e600617dddf973375dbd5fb

ezimanyi (approved):

PR #5142, opened 2020-12-08
fix(exceptions): enable the IllegalArgumentException handler, which r…

dbyron0:
…esponds with 400

Because clouddriver-web brings in kork-web, and kork-web contains
GenericExceptionHandlers, there is a precedence issue between the two classes annotated
with @ControllerAdvice: WebConfig and GenericExceptionHandlers.

See
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/bind/annotation/ControllerAdvice.html
for background, but currently GenericExceptionHandlers takes precedence and so its
handleException method handles IllegalArgumentExceptions and responds with a 500 response
code.


emagana (approved):
LGTM

emagana:
@german-muzquiz @ezimanyi  can you folks bless this PR?  Thanks!

ezimanyi (commented):
Thanks for this!

To make sure I'm understanding here...without the `@Order` annotation, the `kork` `@ControllerAdvice` has priority, and within that `@ControllerAdvice` there is a handler for `Exception` so that gets called (even though there is a more specific handler here)?

So we want to make this handler higher priority so we first check if the exception is a `IllegalArgumentExceptionHandler` and then fall back to the `kork` handlers otherwise?  (And if so, are we sure the `kork` ones will still have effect and aren't completely ignored?)

I guess my only question here is whether it just makes sense to put the `IllegalArgumentExceptionHandler` in `GenericExceptionHandlers` in `kork` instead of having it here.  That would avoid this seemingly-fragile ordering and also would probably fix the return status in our other microservices too.

dbyron0:
> To make sure I'm understanding here...without the `@Order` annotation, the `kork` `@ControllerAdvice` has priority, and within that `@ControllerAdvice` there is a handler for `Exception` so that gets called (even though there is a more specific handler here)?

Yes, that's how I understand it.

> So we want to make this handler higher priority so we first check if the exception is a `IllegalArgumentExceptionHandler` and then fall back to the `kork` handlers otherwise? (And if so, are we sure the `kork` ones will still have effect and aren't completely ignored?)

Sort of.  Yes, we make this one higher priority, and then spring finds a handler for IllegalArgumentException, so it uses it.  In some other tests I did (by changing the exception from IllegalArgumentException to something else), it was clear that the kork exception handler is still in play.

> I guess my only question here is whether it just makes sense to put the `IllegalArgumentExceptionHandler` in `GenericExceptionHandlers` in `kork` instead of having it here. That would avoid this seemingly-fragile ordering and also would probably fix the return status in our other microservices too.

Good point.  My reflex is to say yes, except that I don't know if the IllegalArgumentException convention is specific to clouddriver,

dbyron0:
Looking through [the spinnaker org on github](https://github.com/search?q=org%3Aspinnaker+ExceptionHandler+IllegalArgumentException&type=code) I see that kayenta has [its own exception handler](https://github.com/spinnaker/kayenta/blob/1f43abda552afa77a11b293a87a354c6f74ebc9e/kayenta-web/src/main/java/com/netflix/kayenta/controllers/ControllerExceptionHandler.java#L32) ... and that [kayenta-web uses kork-web](https://github.com/spinnaker/kayenta/blob/1f43abda552afa77a11b293a87a354c6f74ebc9e/kayenta-web/src/main/java/com/netflix/kayenta/controllers/ControllerExceptionHandler.java#L32) ... but then kayenta uses `@RestControllerAdvice` 

and then there's [this in keel](https://github.com/spinnaker/keel/blob/9db12ff2949db3c38501747e6cae3de3ec11fc6f/keel-web/src/main/kotlin/com/netflix/spinnaker/keel/rest/ExceptionHandler.kt) that does its own thing as well, and  [keel-web uses kork-web](https://github.com/spinnaker/keel/blob/9db12ff2949db3c38501747e6cae3de3ec11fc6f/keel-web/keel-web.gradle.kts#L24)....so I see dangerous waters ahead making a more global change.

ezimanyi:
Thanks for the explanation!  Yeah I can see how it might be more impactful to change this in `kork`...and ultimately it might depend on the context what we want the return value of `IllegalArgumentException` to be.  (Probably a 4xx error if the argument eventually comes from the user or config, but a 5xx if somehow it's a server-side issue.).

Not to throw too many ideas out here...but would another way of solving this be to change the code in [getCredentials](https://github.com/spinnaker/clouddriver/blob/cf6cc6dc7b5ba35c40cf19fac8e47d5bbff077ea/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/ArtifactCredentialsRepository.java#L53) to return an `InvalidRequestException` so the `kork` handler picks it up?  (Or catch and rethrow as a `InvalidRequestException` in `ArtifactDownloader` or `ArtifactController` if we don't want that exception too deep in library code?)

I wouldn't mind @rz weighing in here as he wrote all the `SpinnakerException` stuff and I think has a better idea of how it's intended to be used, and this feels like a reasonable example.

ezimanyi:
Thanks for the explanation!  Yeah I can see how it might be more impactful to change this in `kork`...and ultimately it might depend on the context what we want the return value of `IllegalArgumentException` to be.  (Probably a 4xx error if the argument eventually comes from the user or config, but a 5xx if somehow it's a server-side issue.).

Not to throw too many ideas out here...but would another way of solving this be to change the code in [getCredentials](https://github.com/spinnaker/clouddriver/blob/cf6cc6dc7b5ba35c40cf19fac8e47d5bbff077ea/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/ArtifactCredentialsRepository.java#L53) to return an `InvalidRequestException` so the `kork` handler picks it up?  (Or catch and rethrow as a `InvalidRequestException` in `ArtifactDownloader` or `ArtifactController` if we don't want that exception too deep in library code?)

I wouldn't mind @robzienert weighing in here as he wrote all the `SpinnakerException` stuff and I think has a better idea of how it's intended to be used, and this feels like a reasonable example.

dbyron0:
> Not to throw too many ideas out here...but would another way of solving this be to change the code in [getCredentials](https://github.com/spinnaker/clouddriver/blob/cf6cc6dc7b5ba35c40cf19fac8e47d5bbff077ea/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/ArtifactCredentialsRepository.java#L53) to return an `InvalidRequestException` so the `kork` handler picks it up? (Or catch and rethrow as a `InvalidRequestException` in `ArtifactDownloader` or `ArtifactController` if we don't want that exception too deep in library code?)

That would certainly work, though the existence of the exception handler for IllegalArgumentException makes me think it was supposed to be active.  Plus also, I can see a whole truckload of `throw new IllegalArgumentException`s in the code that look like the same kind of thing as `getCredentials` -- where it's a user-level thing, not an internal, something's-wrong-with-the-code kind of thing.

ezimanyi:
Ok, that makes sense!  This definitely underscores that it might be good to have a Spinnaker guide on a standardized way to handle exceptions so that we move towards something more consistent.

This PR seems like a reasonable incremental move towards that...at least let's return 4xx for `IllegalArgumentException` in `clouddriver` (as it seems the intent always was and used to happen) then as a next step we could consider moving it to `kork`.

ezimanyi (approved):

dbyron0:
Thank you!

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/WebConfig.groovy
```diff
@@ -92,6 +93,7 @@ public class WebConfig extends WebMvcConfigurerAdapter {
   }
 
   @ControllerAdvice
+  @Order(Ordered.HIGHEST_PRECEDENCE)
   static class IllegalArgumentExceptionHandler {
     @ExceptionHandler(IllegalArgumentException)
     public void handle(HttpServletResponse response, IllegalArgumentException ex) {
```
german-muzquiz, line 99: Would it make sense to decorate the returned error message like how it's done in [kork](https://github.com/spinnaker/kork/blob/3288c4b87be1f71178e3282d131dd815e08d3574/kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java#L80) in the GenericExceptionHandler?

I'm also curious if the exception handler in [kayenta](https://github.com/spinnaker/kayenta/blob/1f43abda552afa77a11b293a87a354c6f74ebc9e/kayenta-web/src/main/java/com/netflix/kayenta/controllers/ControllerExceptionHandler.java#L32) actually works, because it doesn't have the `@Ordered` annotation.

Anyway seems that the only difference for calling code is to receive a 400 instead of 500, with the same body structure as before. This will potentially impact other services but I think this a right step forward.

dbyron0: > Would it make sense to decorate the returned error message like how it's done in [kork](https://github.com/spinnaker/kork/blob/3288c4b87be1f71178e3282d131dd815e08d3574/kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java#L80) in the GenericExceptionHandler?

Could be.  I haven't stared at the differences in responses by doing that.  My bigger picture goal with all of this is to try to get rosco to not retry when [this call](https://github.com/spinnaker/rosco/blob/03e4e26a6e19fd26a62842c12f93aeb8d4275d8c/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloaderImpl.java#L33) fails due to a misconfigured artifact.  If the decoration helps with that, I'm all for it...I'm planning to bring this up with sig-platform since it looks like there are lots of possible ways to handle this.

ajordens:
Ya, I'm alright with this going into `clouddriver` initially but it should be ported over to `kork-exceptions`. 

I'd like to avoid ad-hoc exception handlers sprinkled through the service for general exceptions.

ajordens:
Ported to `kork` in https://github.com/spinnaker/kork/pull/831

PR #5149, opened 2020-12-09
feat(ecs): add API support for Describe Clusters

paragbhingre:
Add support for getting capacity providers using Describe Cluster API.

Backend changes to support: [spinnaker/spinnaker#5400](https://github.com/spinnaker/spinnaker/issues/5400)


clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -36,4 +46,46 @@ public EcsClusterProvider(Cache cacheView) {
...
+          }
+          clusterNames.clear();
+        }
+      }
+      if (clusterNames.size() % EcsClusterDescriptionMaxSize != 0) {
+        List<Cluster> describeClusterResponse = getDescribeClusters(client, clusterNames);
+        if (describeClusterResponse.size() > 0) {
+          clusters.addAll(describeClusterResponse);
+        }
+      }
```
piradeepk, line 71: This logic is identical. To improve maintainability, we should avoid copying and pasting duplicate code. Instead it would be beneficial to move this logic into a separate method that can be called in both places.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -36,4 +46,46 @@ public EcsClusterProvider(Cache cacheView) {
   public Collection<EcsCluster> getAllEcsClusters() {
     return ecsClusterCacheClient.getAll();
   }
+
+  public Collection<Cluster> getAllEcsClustersDescription(String account, String region) {
```
piradeepk, line 50: Might be better to rename this as getAllEcsClusterDetails (rather than description as that's a bit misleading)

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/controllers/EcsClusterController.java
```diff
@@ -37,4 +39,10 @@ public EcsClusterController(EcsClusterProvider ecsClusterProvider) {
   public Collection<EcsCluster> getAllEcsClusters() {
     return ecsClusterProvider.getAllEcsClusters();
   }
+
+  @RequestMapping(value = {"/ecs/ecsDescribeClusters/{account}/{region}"})
```
piradeepk, line 43: Can we keep the path to be the same as the method name? (for instance if you're using getAllEcsClustersDescription, then the path would be `/ecs/ecsClustersDescription/{account}/{region}`

piradeepk (changes requested):

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -36,4 +46,46 @@ public EcsClusterProvider(Cache cacheView) {
...
+    List<String> clusterNames = new ArrayList<>();
+    List<Cluster> clusters = new ArrayList<>();
+    Collection<EcsCluster> ecsClusters = ecsClusterCacheClient.getAll();
+    AmazonECS client = getAmazonEcsClient(account, region);
+    if (ecsClusters.size() > 0 && ecsClusters != null) {
```
allisaurus, line 55: these 2 conditions should be reversed, otherwise if `ecsClusters` is null then I believe `ecsClusters.size()` will throw an NPE.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -36,4 +46,46 @@ public EcsClusterProvider(Cache cacheView) {
   public Collection<EcsCluster> getAllEcsClusters() {
     return ecsClusterCacheClient.getAll();
   }
+
+  public Collection<Cluster> getAllEcsClustersDescription(String account, String region) {
+    List<String> clusterNames = new ArrayList<>();
+    List<Cluster> clusters = new ArrayList<>();
+    Collection<EcsCluster> ecsClusters = ecsClusterCacheClient.getAll();
```
allisaurus, line 53: we need to filter this list by region & account, don't we?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -36,4 +46,46 @@ public EcsClusterProvider(Cache cacheView) {
...
+    AmazonECS client = getAmazonEcsClient(account, region);
+    if (ecsClusters.size() > 0 && ecsClusters != null) {
+      for (EcsCluster ecsCluster : ecsClusters) {
+        clusterNames.add(ecsCluster.getName());
+        if (clusterNames.size() % EcsClusterDescriptionMaxSize == 0) {
+          List<Cluster> describeClusterResponse = getDescribeClusters(client, clusterNames);
+          if (describeClusterResponse.size() > 0) {
```
piradeepk, line 60: Would it make sense to use !isEmpty rather than size > 0 (whichever is your preference). Can you do a null check here as well? 

`if (describeClusterResponse != null and !describeClusterResponse.isEmpty()) `

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -36,4 +46,46 @@ public EcsClusterProvider(Cache cacheView) {
...
+  private AmazonECS getAmazonEcsClient(String account, String region) {
+    NetflixECSCredentials credentials = credentialsRepository.getOne(account);
+    if (!(credentials instanceof NetflixECSCredentials)) {
+      throw new IllegalArgumentException("Invalid credentials:" + account + ":" + region);
+    }
+    return amazonClientProvider.getAmazonEcs(credentials, region, true);
+  }
+
+  private List<Cluster> getDescribeClusters(AmazonECS client, List<String> clusterNames) {
+    DescribeClustersRequest describeClustersRequest =
```
piradeepk, line 85: What happens if you pass an empty list of clusters? Does this return an error?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
paragbhingre: Added in the new commit.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/controllers/EcsClusterController.java
```diff
@@ -37,4 +39,10 @@ public EcsClusterController(EcsClusterProvider ecsClusterProvider) {
   public Collection<EcsCluster> getAllEcsClusters() {
     return ecsClusterProvider.getAllEcsClusters();
   }
+
+  @RequestMapping(value = {"/ecs/getAllEcsClusterDetails/{account}/{region}"})
```
piradeepk, line 43: instead of getAll you just want to call it ecsClusterDetails (see ecsClusters and getAllEcsClusters for reference.

`"/ecs/ecsClusterDetails/{account}/{region}"`

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
piradeepk: This comment still needs to be addressed.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -36,4 +49,71 @@ public EcsClusterProvider(Cache cacheView) {
...
+                    cluster.getRegion().equals(region) && cluster.getAccount().equals(account))
+            .collect(Collectors.toList());
+    AmazonECS client = getAmazonEcsClient(account, region);
+    if (filteredEcsClusters != null && filteredEcsClusters.size() > 0) {
+      for (EcsCluster ecsCluster : filteredEcsClusters) {
+        clusterNames.add(ecsCluster.getName());
+        if (clusterNames.size() % EcsClusterDescriptionMaxSize == 0) {
+          List<Cluster> describeClusterResponse =
+              makeCallToGetDescribeClusters(client, clusterNames);
+          if (!describeClusterResponse.isEmpty()) {
```
piradeepk, line 71: You should check that describeClusterResponse is not null as well. Please see comment from previous revision for more details.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
paragbhingre: In that case we will get error in failures something like :

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -36,4 +49,71 @@ public EcsClusterProvider(Cache cacheView) {
...
+          clusterNames.clear();
+        }
+      }
+      if (clusterNames.size() % EcsClusterDescriptionMaxSize != 0) {
+        List<Cluster> describeClusterResponse = makeCallToGetDescribeClusters(client, clusterNames);
+        if (!describeClusterResponse.isEmpty()) {
+          clusters.addAll(describeClusterResponse);
+        }
+      }
+    }
```
piradeepk, line 83: This logic contains duplicated code, please refactor it into separated methods.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -16,17 +16,30 @@
...
 import org.springframework.stereotype.Component;
 
+@Slf4j
 @Component
 public class EcsClusterProvider {
 
   private EcsClusterCacheClient ecsClusterCacheClient;
+  @Autowired private CredentialsRepository<NetflixECSCredentials> credentialsRepository;
+  @Autowired private AmazonClientProvider amazonClientProvider;
+  private static final int EcsClusterDescriptionMaxSize = 100;
```
piradeepk, line 42: Can you add a comment here to explain why you've set a static count of 100? (for future reference of the next person who looks at this codebase)

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -36,4 +49,71 @@ public EcsClusterProvider(Cache cacheView) {
...
+              makeCallToGetDescribeClusters(client, clusterNames);
+          if (!describeClusterResponse.isEmpty()) {
+            clusters.addAll(describeClusterResponse);
+          }
+          clusterNames.clear();
+        }
+      }
+      if (clusterNames.size() % EcsClusterDescriptionMaxSize != 0) {
+        List<Cluster> describeClusterResponse = makeCallToGetDescribeClusters(client, clusterNames);
+        if (!describeClusterResponse.isEmpty()) {
```
piradeepk, line 79: Same comment as above.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -36,4 +49,71 @@ public EcsClusterProvider(Cache cacheView) {
...
+    return amazonClientProvider.getAmazonEcs(credentials, region, true);
+  }
+
+  private List<Cluster> makeCallToGetDescribeClusters(AmazonECS client, List<String> clusterNames) {
+    List<Cluster> describeClusterResponse = getDescribeClusters(client, clusterNames);
+    if (describeClusterResponse.size() > 0) {
+      return describeClusterResponse;
+    }
+    return Collections.emptyList();
+  }
```
piradeepk, line 101: You should check that describeClusterResponse is not null (since describeClusterResponse.size will throw an NPE if getClusters is null)

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
piradeepk: In this case, it seems like the wrong approach to knowingly send an erroneous describeClusterRequest and expect a list of failures. It might make more sense to validate your inputs before you make the call in the first place (prevent unnecessary request/rate limiting, and it shouldn't be an error from the describe call if the input you provide this method is invalid)

piradeepk (changes requested):

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
piradeepk: You can just return an emptyList if clusterNames is null or empty

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
paragbhingre: resolved in the latest commit, please take a look at it and let me know if that addresses your concern.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
paragbhingre: Resolved

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/EcsControllersSpec.java
```diff
@@ -18,26 +18,95 @@
...
+    Collection<Cluster> clusters =
+        Arrays.asList(objectMapper.readValue(response.asString(), Cluster[].class));
+    // then
+    assertNotNull(clusters);
+    Cluster clusterDescription =
+        (clusters.stream().filter(cluster -> cluster.getClusterName().equals(testClusterName)))
+            .findAny()
+            .get();
+    assertTrue(clusterDescription.getClusterArn().contains(testClusterName));
+    assertTrue(clusterDescription.getCapacityProviders().size() == 2);
```
piradeepk, line 105: nit: Since you're comparing 2 values, you might want to use assertEquals here (it provides the user with a better error "expected * actual *")

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/EcsControllersSpec.java
```diff
@@ -18,26 +18,95 @@
...
+    assertTrue(clusterDescription.getStatus().equals("ACTIVE"));
```
piradeepk, line 106: nit: Since you're comparing 2 values, you might want to use assertEquals here (it provides the user with a better error "expected * actual *")

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/controllers/EcsClusterController.java
```diff
@@ -37,4 +39,10 @@ public EcsClusterController(EcsClusterProvider ecsClusterProvider) {
   public Collection<EcsCluster> getAllEcsClusters() {
     return ecsClusterProvider.getAllEcsClusters();
   }
+
+  @RequestMapping(value = {"/ecs/getEcsClusterDescriptions/{account}/{region}"})
```
piradeepk, line 43: Since this is already define this as a GET operation, typically we don't add the operation type to the path. For instance, instead of adding getEcsClusterDescriptions to the path you might want to update it to be ecsClusterDescriptions like below:

Current: ("/ecs/getEcsClusterDescriptions/{account}/{region}")

to

Updated: ("/ecs/ecsClusterDescriptions/{account}/{region}")

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -36,4 +51,52 @@ public EcsClusterProvider(Cache cacheView) {
   public Collection<EcsCluster> getAllEcsClusters() {
     return ecsClusterCacheClient.getAll();
   }
+
+  // TODO include[] input of Describe Cluster is not part of this implementation, need to implement
+  // in the future.
```
piradeepk, line 56: Not a comment but rather curious, what inputs did you have in mind?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -36,4 +51,52 @@ public EcsClusterProvider(Cache cacheView) {
...
+  public Collection<Cluster> getEcsClusterDescriptions(String account, String region) {
+    Collection<Cluster> clusters = new ArrayList<>();
+    List<String> filteredEcsClusters =
+        ecsClusterCacheClient.getAll().stream()
+            .filter(
+                cluster ->
+                    cluster.getRegion().equals(region) && cluster.getAccount().equals(account))
```
piradeepk, line 63: It could be possible for cluster.getRegion and getAccount to be null. in this case it would lead to potential NPEs. One way to avoid this is to reverse your comparison:

For instance, you could update the

Current: `cluster.getRegion().equals(region) && cluster.getAccount().equals(account))`

to 

Updated: `region.equals(cluster.getRegion()) && account.equals(cluster.getAccount())`

In this case, if cluster.getRegion or cluster.getAccount is null, this won't thrown an NPE.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -16,17 +16,32 @@
...
 
+@Slf4j
 @Component
 public class EcsClusterProvider {
 
   private EcsClusterCacheClient ecsClusterCacheClient;
+  @Autowired private CredentialsRepository<NetflixECSCredentials> credentialsRepository;
+  @Autowired private AmazonClientProvider amazonClientProvider;
+  private static final int EcsClusterDescriptionMaxSize =
+      100; // Describe Cluster API accepts only 100 cluster Names at a time as an input.
```
piradeepk, line 44: nit: Usual styling in this case would be to add a comment above the property itself.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
paragbhingre: Good question. So, Describe Cluster API responds with ATTACHMENTS | SETTINGS | STATISTICS | TAGS only when they are included in the include[] input. e.g. if you want description of TestCluster with it's SETTINGS then you would send TestCluster inside` Clusters["TestCluster"] `and include settings like this `include["SETTINGS"].`

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/EcsControllersSpec.java
```diff
@@ -16,28 +16,98 @@
...
+        Arrays.asList(objectMapper.readValue(response.asString(), Cluster[].class));
+    // then
+    assertNotNull(clusters);
+    Cluster clusterDescription =
+        (clusters.stream().filter(cluster -> cluster.getClusterName().equals(testClusterName)))
+            .findAny()
+            .get();
+    assertTrue(clusterDescription.getClusterArn().contains(testClusterName));
+    assertEquals(clusterDescription.getCapacityProviders().size(), 2);
+    assertEquals(clusterDescription.getStatus(), "ACTIVE");
```
piradeepk, line 107: You'll want to reverse your arguments (assertEquals(expected, actual)). Right now your expected value is in the actual place holder.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
piradeepk: Got it! That's awesome! Do you see a need for any of those at the moment?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -36,4 +51,52 @@ public EcsClusterProvider(Cache cacheView) {
...
+  public Collection<Cluster> getEcsClusterDescriptions(String account, String region) {
+    Collection<Cluster> clusters = new ArrayList<>();
+    List<String> filteredEcsClusters =
+        ecsClusterCacheClient.getAll().stream()
+            .filter(
+                cluster ->
+                    account.equals(cluster.getAccount()) && region.equals(cluster.getRegion()))
+            .map(cluster -> cluster.getName())
+            .collect(Collectors.toList());
+    List<List<String>> batchClusterList =
```
piradeepk, line 66: Not a blocker, but could you add some debug logging here? It would be great to be able to enable debugging and be able to determine the size of the batchClusterList (how many partitions did we split the original list into?)

piradeepk:
Lgtm, will hold off approving until unit tests are added.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
paragbhingre: No, we do not need them for now.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/EcsControllersSpec.java
paragbhingre: oops, gotcha. I will do it.

paragbhingre:
Added unit test. Please review @piradeepk @allisaurus

clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProviderSpec.groovy
```diff
@@ -30,9 +36,136 @@ class EcsClusterProviderSpec extends Specification {
...
+    attributes.put("region", "us-east-1")
+    attributes.put("clusterArn", "arn:aws:ecs:::cluster/" + clusterNames[2])
+    attributes.put("clusterName", clusterNames[2])
+
+    cacheData.add(new DefaultCacheData(clusterKey, attributes, Collections.emptyMap()))
+
+
+    cacheView.getAll(_) >> cacheData
+
+    //Adding onnly two clusters in the response which belongs to the expected region.
```
allisaurus, line 132: nit: "only"

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
```diff
@@ -36,4 +51,54 @@ public EcsClusterProvider(Cache cacheView) {
   public Collection<EcsCluster> getAllEcsClusters() {
     return ecsClusterCacheClient.getAll();
   }
+
+  // TODO include[] input of Describe Cluster is not part of this implementation, need to implement
+  // in the future.
```
allisaurus, line 56: nit: as phrased, makes it sound like we're missing something we intend to add, when it's really that we don't need the additional info yet. Suggest amending to say "...in the future **if additional properties are needed**"

allisaurus (approved):
LGTM! Approved pending @piradeepk 's comments have been addressed.

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/EcsControllersSpec.java
```diff
@@ -16,28 +16,98 @@
...
+
+    // when
+    String testUrl =
+        getTestUrl("/ecs/ecsClusterDescriptions/" + ECS_ACCOUNT_NAME + "/" + TEST_REGION);
+
+    Response response =
+        get(testUrl).then().statusCode(200).contentType(ContentType.JSON).extract().response();
+
+    ObjectMapper objectMapper = new ObjectMapper();
+    objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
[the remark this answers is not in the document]
paragbhingre: Right now there are no extra values that we are ignoring, but in the future if API responds with any extra value in that case our code might fail to deserialize the response. So it is already future proof.

clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProviderSpec.groovy
paragbhingre: Resolved.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsClusterProvider.java
paragbhingre: make sense.

piradeepk (approved):
Ship it! 💯

PR #5150, opened 2020-12-09
refactor(cloudfoundry): use CredentialsType

ncknt:
Use credentials type configuration to use strict typing on credentials sources.


zachsmith1 (approved):
LGTM! Good catch with the EqualsAndHashCode

PR #5152, opened 2020-12-11
feat(cloudfoundry): bind services before building droplet

zachsmith1:
This change includes some refactoring for create service bindings that will allow us to use this same logic in a new stage specifically for binding to services with optional parameters. 

We've also moved the createServiceBindings() logic ahead of buildDroplet(). This allows the droplet creation to take advantage of environment variables/service bindings that are present now, rather than doing a restage after we've added service bindings.


kevinawoo (approved):

kevinawoo:
@mergifyio backport release-1.24.x

zachsmith1:
@fieldju this is technically a "performance" improvement. I wasn't aware that we could tag this as "perf". Ideally, since performance improvements are not blocked backport candidates, we should merge this into 1.23, 1.22, if possible.

kevinawoo:
@mergifyio backport release-1.23.x release-1.22.x

PR #5166, opened 2020-12-17
feat(cloudfoundry): add service-binding stage

zachsmith1:
This is the completed clouddriver work for the new cf stage that will bind applications to services. This functionality allows users to optionally provide a parameter artifact (valid json) with the service binding request. We've also added a few edge cases where users can pick between restarting an application vs restaging after the binding takes place. We also added the ability to override existing bindings or continue on if they already exist.


kevinawoo (commented):
LGTM

PR #5169, opened 2020-12-21
feat(amazon/lb): Allow external load balancers to set IpAddressType

caseyhebebrand:
In an effort to expand Ipv6 support to load balancers, clouddriver must support setting the ALB/NLB `IpAddressType`. 

Currently, clouddriver does not set this so it defaults to `ipv4` in AWS. With this PR, the value can be either `ipv4` or `dualstack` for _external_ load balancers (AWS has does not allow dual-stacking of internal load balancers since all IPv6 are public).


aravindmd:
I don't see any change to the caching agent, do we want to return this detail?

caseyhebebrand:
> I don't see any change to the caching agent, do we want to return this detail?

@aravindmd This is already returned and visible in deck as the `IP Type`, users just haven't been able to change it before. 

Screen Shot 2021-01-04 at 10 08 36 AM

aravindmd (approved):

PR #5176, opened 2021-01-05
fix(aws): Support china and rest-of-world pricing hosts

minkjj:
fixes https://github.com/spinnaker/spinnaker/issues/6064


ajordens (approved):
Re-formatted (tabs to spaces) but otherwise this looks good to me.

PR #5180, opened 2021-01-06
feat(aws): Allow user supplied user data to override user provider user data (typically, udf templates)

jonsie:
- Additionally allows user supplied user data to use the standard UDF tokens
- Adds parameter `userDataOverride ` to the `BasicAmazonDeployDescription` and `ModifyAsgLaunchConfigurationDescription`
-  Moves UDF tokenization code out to util `UserDataTokenizer`
- Consolidates the user data aggregation logic via `UserDataProviderAggregator`
- Cleans up `UserDataProvider` (requires a few small -nflx changes)
- Introduces spinnaker extension point `UserDataTokenizer`.


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/UserDataProviderAggregator.java
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * Aggregates all user data. First iterates through all providers and joins user data with a
+   * newline. Then, adds the user supplied base64 encoded user data and again joins with a newline.
+   * The result is such that the user supplied base64 encoded user data is always appended last to
+   * the user data.
+   *
+   * <p>Note, if the {@link UserDataProvider.UserDataRequest} parameter overrideDefaultUserData is
+   * true, then the user data from the providers is skipped and the user supplied base64 encoded
+   * user data is used as the override. If this is the case, the standard set of user data format
+   * tokens ("%%app%%, for example) are replaced in the user data - effectively processing the user
```
robzienert, line 33:
```suggestion
   * tokens ("%%app%%", for example) are replaced in the user data - effectively processing the user
```

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/UserDataProviderAggregator.java
```diff
@@ -0,0 +1,77 @@
...
+   * data as a UDF template.
+   *
+   * @param userDataRequest {@link UserDataProvider.UserDataRequest}
+   * @return String
+   */
+  public String aggregate(UserDataProvider.UserDataRequest userDataRequest) {
+    List<String> allUserData = new ArrayList<>();
+    if (providers != null && (!userDataRequest.isOverrideDefaultUserData())) {
```
robzienert, line 41:
```suggestion
    if (providers != null && !userDataRequest.isOverrideDefaultUserData()) {
```

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/UserDataTokenizer.java
```diff
@@ -0,0 +1,73 @@
...
+            .replace("%%countries%%", countries)
+            .replace("%%devPhase%%", devPhase)
+            .replace("%%hardware%%", hardware)
+            .replace("%%zone%%", zone)
+            .replace("%%cluster%%", cluster)
+            .replace("%%stack%%", stack)
+            .replace("%%detail%%", detail)
+            .replace("%%tier%%", "");
+
+    if (userDataRequest.getLaunchTemplate() != null && userDataRequest.getLaunchTemplate()) {
```
robzienert, line 55:
```suggestion
    if (userDataRequest.getLaunchTemplate() != null) {
```

Odd conditional, isn't the second expression is redundant?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/UserDataTokenizer.java
```diff
@@ -0,0 +1,73 @@
...
+          result
+              .replace("%%launchconfig%%", userDataRequest.getLaunchSettingName())
+              .replace("%%launchtemplate%%", "");
+    }
+
+    return result;
+  }
+
+  private static boolean isPresent(String value) {
+    return value != null && !value.isBlank();
```
robzienert, line 71: nit, but we could use `!Strings.isNullOrEmpty()` here.

clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/UserDataProviderAggregatorSpec.groovy
```diff
@@ -0,0 +1,105 @@
+package com.netflix.spinnaker.clouddriver.aws.deploy.userdata
+
+import spock.lang.Specification
+
+class UserDataProviderAggregatorSpec extends Specification {
```
robzienert, line 5: nit: Would prefer to see the base64-encoded test inputs as test resources instead and then base64 the file contents inside the test class, that way if the test inputs need to change, it'll be more readily obvious what specific part changed. Further, we won't need to worry about these comments drifting from the actual values.

robzienert (approved):

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/UserDataProviderAggregator.java
jonsie: Weird.  Must've been an artifact from some change I made and then deleted.  Nice eye.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/UserDataTokenizer.java
jonsie: In this case `getLaunchTemplate` is a `Boolean` object which can be null and we only want to run this code if it's both not null and true.

cfieber (approved):
nice cleanup, a couple minor suggestions

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/UserDataTokenizer.java
```diff
@@ -0,0 +1,73 @@
+package com.netflix.spinnaker.clouddriver.aws.deploy.userdata;
+
+import com.netflix.frigga.Names;
+
+/** Utility class to replace tokens in user data templates. */
+public class UserDataTokenizer {
```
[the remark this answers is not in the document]
jonsie: I like that.  Making this an extension point could potentially greatly simplify the use case too!

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/userdata/UserDataTokenizer.java
robzienert: I didn't realize what value it was. Poor method / property naming, that one. Oh well. Thanks!

jonsie: Yeah that's the default name you get from lombok when you use the `Boolean` object.  👎

clouddriver-api/clouddriver-api.gradle
```diff
@@ -23,6 +23,7 @@ dependencies {
   api("com.netflix.spinnaker.kork:kork-plugins-api")
   api("com.netflix.spinnaker.kork:kork-credentials-api")
 
+  implementation "com.netflix.frigga:frigga"
```
jonsie, line 26: I think it's ok to add frigga to clouddriver-api - it is lightweight (believe it is just Java) and stable.

cfieber (approved):

PR #5184, opened 2021-01-11
fix(kubernetes): update kubernetes jobStatus with relevant failure details

apoorvmahajan:
## Purpose
- Presently, when a runjob fails its execution, we see this message in the UI:
Screen Shot 2021-01-11 at 10 26 16 AM

We can view the logs from the `Console Output` link. But if the pod gets deleted, we see the following message:
Screen Shot 2021-01-11 at 10 28 42 AM

For those pods that have containers with the property 
```
terminationMessagePolicy: FallbackToLogsOnError
```
it should be possible for them to see the last few lines of the container in the UI itself. This PR allows that to happen. 

Output with the changes in this PR:

Screen Shot 2021-02-02 at 10 40 31 AM

PS: this requires some changes in orca which is linked [here](https://github.com/spinnaker/orca/pull/4177)

List of changes:
- update KubernetesJobStatus to capture why a job failed
- update the pods to capture details about its containers and their executions. 
- using the container execution details, provide a summary of what went wrong with the job


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesJobStatus.java
```diff
@@ -64,6 +71,45 @@ public KubernetesJobStatus(V1Job job, String account) {
     details.put("signal", this.signal != null ? this.signal.toString() : "");
     details.put("message", this.message != null ? this.message : "");
     details.put("reason", this.reason != null ? this.reason : "");
+
+    // find the execution details of the first pod that has failed. The Job completion details will
```
clanesf, line 75: it's a bit of a nit, but is there a reason not to return all of the pods statuses. If they're drilling down to that level of detail in the UI, we might as well give them everything we can about why a pod failed.

clanesf (approved):
Other than the one small nit above this is great. Anything that can help bubble up errors inside of containers is a big help to ops. We've had a few people ask for the ability to bubble up failure messages inside of containers, so this is great.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/model/KubernetesJobStatus.java
apoorvmahajan: I went with the thought that if the job has failed, then we should simply find the first pod that has containers which exit with a non-zero code and use its details in the spinnaker context. We could look at all pod failures, but it would start looking like a big wall of text in the ui. Atleast with one pod failure error, it'll be a good starting point for users to start debugging what went wrong.

clanesf: That makes sense. We discussed this at the Kubernetes SIG meeting on 1-26-21 and feedback from the community was very positive. This is a good place to start and if additional use cases bring up a need to retrieve more specific k8s objects then we can use this as a model.

clanesf:
There are additional optimizations that have been discovered as part of running this code in our environments. @apoorvmahajan or myself will be submitting updates in the next couple of weeks sometime.

dbyron-sf (approved):

PR #5185, opened 2021-01-12
fix(titus): Set env variable to map to spinnaker account

aravindmd:
Set env variable to map to spinnaker account , this has to be accessible within the task(container) for some automated work flows.


ajordens:
Curious if this would be confusing given other udf fields are typically `NETFLIX_` prefixed.

aravindmd:
> Curious if this would be confusing given other udf fields are typically `NETFLIX_` prefixed.

Good point , i think it would be ok to add the `NETFLIX_` prefix here.

aravindmd:
> > Curious if this would be confusing given other udf fields are typically `NETFLIX_` prefixed.
> 
> Good point , i think it would be ok to add the `NETFLIX_` prefix here.

Any other concerns other than the naming , i'll add the prefix.

ajordens:
Nothing from me, I think it'll be fine.

ajordens (approved):

PR #5186, opened 2021-01-12
feat(clouddriver/Azure): Assign user/system managed identity during vmscaleset creation

harikumarks:
When we deploy VMSS using azure cloud provider, have an option to specify managed identity ( User assigned or System assigned). This allows the process inside the VM to gain access to necessary azure resources with RBAC of the assigned identities without need for maintaining separate credentials for authentication/authorization. https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/qs-configure-cli-windows-vmss

Refer issue: https://github.com/spinnaker/spinnaker/issues/5638

The changes are  to add two additional attributes which are optional in nature for server group deploy

- useSystemManagedIdentity ( if this is selected a system assigned identity is automatically assigned)
- userAssignedIdentities ( we can provide a comma separated list of user assigned identity over here to be assigned to the VMs inside the scaleset)
This is a backend change only, attributes could be added to pipeline json directly. Corresponding UI changes could be added separatetely


harikumarks:
@spinnakerbot add-label sig/azure

ajordens:
Confirmed with @neil-yechenwei

neil-yechenwei:
@ajordens , after checked CI build status, seems there is a [CI failure](https://github.com/spinnaker/clouddriver/runs/2195269132) while merging this PR. I am not sure what the root cause is. Could you help to take a look? Thanks in advance.

image

PR #5189, opened 2021-01-14
feat(provider/azure): Support termination notification profile

EasyAsABC123:
resolves https://github.com/spinnaker/spinnaker/issues/6135
bump api version from 2018-10-01 to 2019-03-01 for azure
updated to have unit tests for termination notification profile
updated all unit tests for server group to have the newly updated api version

### DIFF
--- a/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/model/AzureServerGroupDescription.groovy
+++ b/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/model/AzureServerGroupDescription.groovy
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.clouddriver.azure.resources.servergroup.model
 
 import com.google.common.collect.Sets
+import com.microsoft.azure.management.compute.TerminateNotificationProfile
 import com.microsoft.azure.management.compute.VirtualMachineScaleSetDataDisk
 import com.microsoft.azure.management.compute.implementation.VirtualMachineScaleSetInner
 import com.netflix.frigga.Names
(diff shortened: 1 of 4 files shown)


EasyAsABC123:
@grayzu please have your team review for Azure SIG.  Anyone else also welcome to review.

EasyAsABC123:
@spinnakerbot add-label sig/azure

kfrank2017 (approved):
This looks great!

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/model/AzureServerGroupDescription.groovy
```diff
@@ -193,6 +194,15 @@ class AzureServerGroupDescription extends AzureResourceOpsDescription implements
     azureSG.region = scaleSet.location()
     azureSG.upgradePolicy = getPolicyFromMode(scaleSet.upgradePolicy().mode().name())
 
+    def termProfile = scaleSet.virtualMachineProfile()?.scheduledEventsProfile()?.terminateNotificationProfile()
+    if (termProfile)
+    {
+      String[] str = termProfile.notBeforeTimeout().findAll( /\d+/ )
+      if (str.size() > 0) {
+        azureSG.terminationNotBeforeTimeout = str[0].toInteger()
```
harikumarks, line 202: Again since we are assuming that this in minutes we probably have to the conversion if other representations are possible. What happens in scenario where PT600S was provided during install, does the saved config give us back as PT10M ? then this is ok

EasyAsABC123:
```
--terminate-notification-time    : Length of time (in minutes, between 5 and 15) a notification to be sent to the VM on the instance metadata server till the VM gets deleted.
```

So az cli only accepts minutes, i’m fine with forcing Spinnaker to also only accept minutes, i’ll update the PR to state minutes.  This also has a limit but i haven't seen this type of limit implemented anywhere else in clouddriver, if you have an opinion on it.

PR #5196, opened 2021-01-20
fix(docker): Fix docker.slim builds with python3 support

jasonmcintosh:
Fixes an issue where with 3.12 of alpine, python3 is now defaulted and that broke aws cli installation.

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -24,7 +24,7 @@ RUN apk update \
 RUN pip install --upgrade awscli==${AWS_CLI_VERSION} s3cmd==${AWS_CLI_S3_CMD} python-magic \
   && pip uninstall -y pip
 
-# Google cloud SDK with anthos removed for CVE and because we don't need it
+# Google cloud SDK 
 RUN wget -nv https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${GOOGLE_CLOUD_SDK_VERSION}-linux-x86_64.tar.gz \
   && mkdir -p /opt && cd /opt \
   && tar -xzf /google-cloud-sdk-${GOOGLE_CLOUD_SDK_VERSION}-linux-x86_64.tar.gz \
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -1,9 +1,8 @@
 FROM python:3.7-alpine3.12
 LABEL maintainer="sig-platform@spinnaker.io"
 
-# KUBECTL_VERSION kept one minor version behind latest to maximise compatibility overlap
-ENV KUBECTL_VERSION v1.18.10
-ENV KUBECTL_RELEASE=1.17.7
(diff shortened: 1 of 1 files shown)


Dockerfile.slim
```diff
@@ -1,33 +1,47 @@
...
-ENV PATH "$PATH:/usr/local/bin/"
-
-ENV PATH=$PATH:/opt/google-cloud-sdk/bin/
+# KUBECTL_RELEASE kept one minor version behind latest to maximise compatibility overlap
+ENV KUBECTL_RELEASE=1.18.10
+ENV AWS_CLI_VERSION=1.18.152
+ENV AWS_CLI_S3_CMD=2.0.2
+ENV AWS_AIM_AUTHENTICATOR_VERSION=0.4.0
+ENV GOOGLE_CLOUD_SDK_VERSION=313.0.1
+ENV ECR_TOKEN_VERSION=v1.0.2
```
allisaurus, line 10: Nice move, surfacing these higher up!

allisaurus (commented):
Thanks so much for drafting this fix! Re: testing, are you able to run and [execute AWS CLI](https://github.com/spinnaker/spinnaker/issues/6226#issuecomment-761675966) cmds within the container? That should confirm that this addresses https://github.com/spinnaker/spinnaker/issues/6226

Changes otherwise LGTM, but would like insight from others more familiar with this file & non-AWS contents prior to merging (maybe @german-muzquiz , @plumpy ? Pls redirect if not)

jasonmcintosh:
Yeah was able to exec into the container run some basic aws commands ;) STILL want a better end to end test... hoping we see more of this in the Jenkins layer!

kevinawoo (approved):
👍 from me for everything, I think everything is good to go. Merging!

kevinawoo:
Had a chat with @Chinikins and it seems like it's a regression. Adding the backport label for 1.24.x
<kbd>
)</kbd> source [slack](https://spinnakerteam.slack.com/archives/C091CCWRJ/p1611358860026900)

fieldju:
@Mergifyio backport release-1.24.x

Dockerfile.slim
battlecow: Arrggg. AWS_AIM_AUTHENTICATOR_VERSION variable is the older version which doesn't support IRSA and was* fixed previously in #4696

jasonmcintosh: This is intentional.  We encountered some rather nasty bugs in 0.5.0 that cause slow-downs in large environments.  0.5.2 SLIGHTLY fixed this but was substantial enough of an impact that we recommended rolling back for all our customers.  The recommendation was been "DON'T" use aws-iam-authenticator but instead the aws cli for those cases.  See: https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/347 for reference.  The AWS CLI should support this natively instead of using aws-iam-authenticator

PR #5197, opened 2021-01-20
chore(*): Make jackson modules injectable, tweaks to models for serialization/deserialization

jonsie:


ajordens (approved):

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/CloudDriverConfig.groovy
```diff
@@ -137,15 +138,19 @@ class CloudDriverConfig {
   }
 
   @Bean
-  Jackson2ObjectMapperBuilderCustomizer defaultObjectMapperCustomizer() {
+  Jackson2ObjectMapperBuilderCustomizer defaultObjectMapperCustomizer(List<Module> modules) {
```
robzienert, line 141: I am _a little_ worried about this change: I'd like to see a Spring integration test that verifies we're not adding some modules from Spring that we previously were not. If I were betting, I'd say we're not, but it's better safe than sorry.

robzienert (approved):

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/CloudDriverConfig.groovy
jonsie: Good point.  Yeah - I think it's adding `JsonComponentModule` and `ParameterNamesModule`.  `JsonComponentModule` allows us to register beans annotated with `@JsonComponent` (not used in clouddriver). `ParameterNamesModule` adds default introspection of method and constructor parameters (just reading the maven repo description).

robzienert: Neat, I didn't know about those. I'm good with it as-is, then.

PR #5198, opened 2021-01-20
patch(clouddriver): threaded healthcheck (#12)

karlomendozaarmory:
This PR Focus on Clouddriver health-check Accounts.
It does 2 things.
1: Run in parallel the health-check for all the accounts on a given provider you have.
2: Clouddriver health-check is run on the default scheduler which uses a thread pool of size 1. Another health-check that runs on this same thread pool is the SQL one. 
This causes Clouddriver to be marked as healthy after all your accounts are checked and then the SQL health is checked.
The change allows the user to set a property and increase the pool size so checks can be done in parallel. 

Both of this changes allow the user to set a property to activate the changes and if not set everything runs as usual.
(Would gladly push a doc change with this info if pointed to the best place to add this.)


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/AccountHealthIndicator.java
```diff
@@ -45,6 +49,11 @@
   @Nonnull private Health health = new Health.Builder().up().build();
   @Nonnull private final AtomicLong unhealthyAccounts = new AtomicLong(0);
 
+  @Value("${clouddriver.account-threads:-1}")
+  private int threads;
+
+  private ForkJoinPool customThreadPool;
```
german-muzquiz, line 55: Given that this health check always returns "up" and the change of `ThreadPoolTaskScheduler` will unblock parallel `@Scheduled` executions, probably creating another thread pool in clouddriver doesn't add much value.

german-muzquiz (approved):

PR #5200, opened 2021-01-21
chore(dependencies): upgrade of jGit for performance improvements

galuszkak:
This PR is including changes from jGit that may speedup performance 
```
JGit 5.9.0 supports git repositories using git index version 4. In version 4 entry path names are prefix-compressed relative to the path name for the previous entry which reduces the size of the git index and improves performance.
```


mattgogerly:
I think this can probably be closed now that Clouddriver is shelling out to Git rather than using JGit?

PR #5201, opened 2021-01-22
fix(core): Fix job executor stuck indefinitely

aufarg:
This commit introduces a timeout value for streaming job to prevent the agent thread to stuck indefinitely when something goes wrong with the job.

Fixes: https://github.com/spinnaker/spinnaker/issues/6288

### DIFF
--- a/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/jobs/local/JobExecutorLocal.java
+++ b/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/jobs/local/JobExecutorLocal.java
@@ -26,6 +26,8 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.exec.*;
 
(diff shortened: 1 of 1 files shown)


aufarg:
I don't think there's a test file for this class, should I add a new test file for `JobExecutorLocal`?

aufarg:
Right now, I'm using the existing timeout value so if user haven't configure timeout yet, they will notice a change in behavior where work that can actually complete with more than 10m will start to time out.
If this is not desired, we can make a separate timeout for streaming jobs. Let me know which one you think is better!

ajordens:
Can you rebase this PR, it's hard to follow with the other commits pulled in.

aufarg:
Done

mitali-sahoo:
@aufarg @ajordens  I am facing the similar caching issue on spinnaker. We are hosting v1.20.7 spinnaker version currently. Could you please consider merging the changes to the mentioned version too ?

rampatnaik:
weare facing the similar caching issue on spinnaker. We are hosting v1.20.7 spinnaker version currently. Could you please consider merging the changes to the mentioned version too ?

sethev:
This can also cause monitorDeploy to hang indefinitely if this occurs during a deploy task: https://github.com/spinnaker/spinnaker/issues/6410#issuecomment-923440166

sethev:
@aufarg want to update and resolve the conflict?

deverton (approved):

dbyron-sf (approved):

jkschneider (approved):

PR #5208, opened 2021-01-26
feat(cloudfoundry): support docker deployments

zachsmith1:
This change allows docker artifacts to be deployed to cloudfoundry. The original logic was constrained to deployments packages with types buildpack/bits. This was refactored so that packages can be independently created for either type. We also refactored the lifecycle object so that it can be independently created for the docker or buildpack lifecycle. Tests were added and adjusted to cover the new logic and support the existing logic in place. 

https://github.com/spinnaker/deck/pull/8864


zachsmith1:
Fixed -> https://github.com/spinnaker/spinnaker/issues/6191

kevinawoo (approved):

PR #5209, opened 2021-01-27
feat(provider/azure): Support timeZone for windows Configuration

EasyAsABC123:
Adds support for `timeZone` to VMSS creation.

Resolves https://github.com/spinnaker/spinnaker/issues/6313


bm7150:
Screen Shot 2021-01-27 at 5 26 11 PM

bm7150:
Screen Shot 2021-01-27 at 5 26 11 PM

bm7150 (approved):
Uploaded screenshot with info showing this change deployed correctly.

kfrank2017 (approved):
This is great!

clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/templates/AzureServerGroupResourceTemplate.groovy
```diff
@@ -458,6 +458,23 @@ class AzureServerGroupResourceTemplate {
...
+      super(description)
+      windowsConfiguration = new OsProfileWindowsConfiguration(description)
+    }
+  }
+
+  static class OsProfileWindowsConfiguration {
+    String timeZone
+
+    OsProfileWindowsConfiguration(AzureServerGroupDescription description) {
+      timeZone = description.timeZone
```
harikumarks, line 474: @EasyAsABC123 lets us chat on this , i feel that if timezone is the only factor that can be used to decide if this is a windows config we should probably also rename the attribute to reflect that this is a specific to windows ( for example: description.windowsConfig.timeZone)

EasyAsABC123:
```
"windowsConfiguration": {
        "provisionVMAgent": "boolean",
        "enableAutomaticUpdates": "boolean",
        "timeZone": "string",
        "additionalUnattendContent": [
          {
            "passName": "OobeSystem",
            "componentName": "Microsoft-Windows-Shell-Setup",
            "settingName": "string",
            "content": "string"
          }
        ],
        "winRM": {
          "listeners": [
            {
              "protocol": "string",
              "certificateUrl": "string"
            }
          ]
        }
      },
```

harikumarks: Right, so these are specific to windowsConfig. In future we may need to add other config as well so it would help group it under a sub bean class

EasyAsABC123: after discussion we decided to migrate this to `windowsTimeZone`

neil-yechenwei:
LGTM

PR #5211, opened 2021-01-27
feat(kubernetes/clouddriver): sharding logic added for kubernetes accounts for clouddriver caching

kirangodishala:
Addressing the issue [#6315](https://github.com/spinnaker/spinnaker/issues/6315)

This feature is added on top of custom binding of kubernetes accounts for bootstrap sources.

The following configuration enables and provides **account name pattern** based on which clouddriver caching services filter the accounts. The assumption is that there will be more than one clouddriver caching deployments each having a unique name(say clouddriver-caching, clouddriver-caching-1 etc). 

```
---
spring:
  profiles: local

kubernetes:
  customPropertyBindingEnabled: true
  shardingEnabled: false

---
spring:
  profiles: caching-local

kubernetes:
  sharding:
    config:
      selectorType: AccountName
      pattern: k8s-acc-v1.*

---
spring:
  profiles: caching-1-local

kubernetes:
  sharding:
    config:
      selectorType: AccountName
      pattern: k8s-acc-v2.*|.*acc-v3.*

``` 

This way, every caching service is ensured to look for only a limited and same set of accounts in every run of caching agents resulting in optimal memory utilization.

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/BootstrapKubernetesConfigurationProvider.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/BootstrapKubernetesConfigurationProvider.java
@@ -21,6 +21,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.wnameless.json.flattener.JsonFlattener;
 import com.github.wnameless.json.unflattener.JsonUnflattener;
+import com.netflix.spinnaker.clouddriver.kubernetes.sharding.KubernetesShardingFilter;
 import com.netflix.spinnaker.kork.configserver.CloudConfigResourceService;
 import com.netflix.spinnaker.kork.secrets.EncryptedSecret;
 import com.netflix.spinnaker.kork.secrets.SecretManager;
@@ -30,6 +31,8 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.boot.context.properties.bind.BindResult;
 import org.springframework.boot.context.properties.bind.Bindable;
 import org.springframework.boot.context.properties.bind.Binder;
@@ -49,20 +52,26 @@ import org.springframework.core.env.PropertySource;
  * KubernetesConfigurationProperties object.
  */
 public class BootstrapKubernetesConfigurationProvider {
+  private static final Logger logger =
+      LoggerFactory.getLogger(BootstrapKubernetesConfigurationProvider.class);
   private final ConfigurableApplicationContext applicationContext;
   private CloudConfigResourceService configResourceService;
   private SecretSession secretSession;
   private Map<String, String> configServerCache;
   private ObjectMapper objectMapper = new ObjectMapper();
   private final String FIRST_ACCOUNT_NAME_KEY = "kubernetes.accounts[0].name";
+  private KubernetesShardingFilter kubernetesShardingFilter;
 
   public BootstrapKubernetesConfigurationProvider(
       ConfigurableApplicationContext applicationContext,
       CloudConfigResourceService configResourceService,
-      SecretManager secretManager) {
+      SecretManager secretManager,
+      KubernetesShardingFilter kubernetesShardingFilter) {
     this.applicationContext = applicationContext;
     this.configResourceService = configResourceService;
     this.secretSession = new SecretSession(secretManager);
+    this.kubernetesShardingFilter = kubernetesShardingFilter;
+    logger.info("custom-property-binding-enabled for Kubernetes accounts");
   }
 
   public KubernetesConfigurationProperties getKubernetesConfigurationProperties() {
@@ -74,6 +83,7 @@ public class BootstrapKubernetesConfigurationProvider {
       Map<String, Object> kubernetesPropertiesMap) {
     KubernetesConfigurationProperties k8sConfigProps = new KubernetesConfigurationProperties();
     BindResult<?> result;
+    KubernetesConfigurationProperties.ManagedAccount account;
 
     // unflatten
     Map<String, Object> propertiesMap =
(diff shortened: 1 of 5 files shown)


german-muzquiz:
I think having to define different `clouddriver.yml` files and having different clouddriver `Deployment` is going to increase operational complexity a lot. Another consideration is that this solution relies on manually assingning accounts to CD instances, so in case of accounts growing up and the need to deploy a new CD instance, either the new instance will have just the 1 account recently added or a manual rebalancing of the load needs to be done by renaming accounts, so that all instances share the same load.

It would be much better if we can scale CD in a kubernetes native way by just by increasing its replicas, and the service does itself the balancing of the number of accounts, so that we still have only one `clouddriver.yml` and one `Deployment`.

kirangodishala:
> I think having to define different `clouddriver.yml` files and having different clouddriver `Deployment` is going to increase operational complexity a lot. Another consideration is that this solution relies on manually assingning accounts to CD instances, so in case of accounts growing up and the need to deploy a new CD instance, either the new instance will have just the 1 account recently added or a manual rebalancing of the load needs to be done by renaming accounts, so that all instances share the same load.

I agree it's a bit overhead but compared to not using clouddriver-caching because of frequent crashes, this solution seemed a good alternate. For the second consideration, any new account just like existing accounts follow certain naming pattern so by adding it to the clouddriver-local.yml (a single file which has all the accounts), the new account gets loaded into the appropriate caching instance automatically.

german-muzquiz:
There are ways for achieving automatic sharding without the operational burden of manually managing different clouddriver deployments, and I think we should strive for that, Spinnaker is already complex enough to manage.

For example, one way could be to make each clouddriver instance have a unique id and register it in the database in order to know at any given time how many replicas are available (which also has the benefit of knowing when one goes bad). Then when the [agent scheduler](https://github.com/spinnaker/clouddriver/blob/master/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt#L152) determines which caching agents to execute, grab only the accounts for that replica based on something like [the namespace hashCode technique](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesCachingAgent.java#L265) that is already used to spread the load of caching namespaces across different threads.

That's only an example, but any idea for accomplishing sharding without managing different config files for the same service is welcome. At the end users shouldn't need to know scaling internals to operate Spinnaker, just provide a single config file and the number of replicas to use.

kirangodishala:
@german-muzquiz  we are not only looking for automatic sharding but also a user controlled sharding. We have a use case where there are a large number of accounts and we want some control on prioritizing certain applications over others. With the changes in clouddriver deployments to perform live manifest checks without the caching requirement, we want to be able to keep the state of applications higher priority applications and lower priority on applications that are not critical. 

We also want to extend the sharding capability to other clouddriver instances like ro and rw. The reason to do this is to improve performance, particularly with clouddriver-ro as the GUI requests. These instances will require a rule based approach as Gate and Orca will connect with the instance that can serve the data. Using a similar scheme across all clouddriver is useful. 
For users not using HA and want to achieve sharding, we need to be able to configure rules. Having a consistent scheme for default users is simpler to maintain. 
Like Orca which has the pattern based configuration for sharding, I am proposing for clouddriver caching which can be extended to other clouddriver profiles going forward.

german-muzquiz:
When the problem is performance and scalability, I still don't see how statically assigning account patterns to specific deployments could be better than automatically balance the number of accounts across stateless replicas, which would make a more efficient use of available resources. Maybe you're trying to solve another problem? I don't fully understand the purpose of prioritizing some application above others or how it works.

kirangodishala:
Some accounts have more namespaces/resources associated with them, some accounts can be more dynamic than others (more deployments), we are looking at deployment with 100s of accounts.
Same strategy will work with non-HA clouddriver instances as we look at the sharding for clouddriver access from Orca and Gate.
Automatically balancing is also an option, we can extend the implementation for HA use case with caching with no additional option to user. However, the current proposal is also legitimate for the reasons mentioned above.

german-muzquiz:
If I understand there's a use case where you can have one account with a disproportionate amount of resources compared with smaller accounts, so doing an automatic load balance treating all accounts as equal would not be ideal because one pod would need more resources than the others.

I see that another potential use case of url traffic sharding is to split clouddriver by cloud providers, for example by having a clouddriver `Deployment` which exclusively deploys to kubernetes and another to AWS.

The current solution is too specific for a single use case (kubernetes accounts pulled in from configserver using the manual constructor, maybe only works for HA mode?). It would be good if we can implement traffic url sharding in a generic way, so it can work for any provider, even if not using configserver, and in HA and non-HA modes. Something similar to the sharding logic done in [kork](https://github.com/spinnaker/kork/blob/master/kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByApplicationServiceSelector.java#L22) where there are small classes each implementing different sharding logic, so everything is generic and decoupled.

With that being said, how do you plan to actually deploy different clouddriver `Deployment` and manage their individual configurations?

kirangodishala:
> If I understand there's a use case where you can have one account with a disproportionate amount of resources compared with smaller accounts, so doing an automatic load balance treating all accounts as equal would not be ideal because one pod would need more resources than the others.

Yes, that is the case (need not be one account but a set of accounts), hence account name based sharding.

kirangodishala:
We are seeing issues when using a large number of Kubernetes accounts so the solution is also specific to Kubernetes accounts with dynamic configuration. This is a pressing issue so I want to address this first. Having said that, I am planning to extend the custom binding feature to all the sources next. Similarly we can enhance the  sharding logic to address going forward. 

For deployment, we will work on a separate PR for halyard. But for testing this feature we will provide scripts.

german-muzquiz:
Given that this potentially changes how spinnaker is deployed I think a broader audience might be interested in this. @clanesf already has a few ideas for this type of sharding and will probably be a RFC, so you two might want to get in touch to discuss it.

kirangodishala:
Created another [PR](https://github.com/spinnaker/clouddriver/pull/5295) to address sharding without need to additional deployments. Hence closing this one.

PR #5212, opened 2021-01-27
fix(titus): Avoid fetching the job before terminating it

ajordens:
This can be a costly lookup should the job have many thousands of associated tasks.

If the job doesn't exist, Titus will return a `NOT_FOUND` exception that can be appropriately handled.


robzienert (approved):

christopherthielen (approved):
Neat.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/DestroyTitusJob.java
```diff
@@ -70,24 +70,31 @@ public Result apply(@NotNull DestroyTitusJob.DestroyTitusJobCommand command, @No
...
+          saga.log("No titus job found");
+          return new Result();
+        }
+      }
+
+      throw e;
+    }
   }
 
   private Job fetchJob(TitusClient titusClient, String jobId) {
```
jonsie, line 100: I suppose this fetchJob method could be deleted.

PR #5213, opened 2021-01-27
fix(aws/titus cache): fix cluster and application cache eviction

cfieber:
Clusters and Applications are global entities and can not be
authoritatively cached by a regionally scoped caching agent.

This moves them to INFORMATIVE and adds a new ClusterCleanupAgent
to handle eviction of clusters with no serverGroups.

Also includes changes in the Application and Cluster providers to
account for inconsistent reads (particularly when doing
titus streaming updates) by looking for the presence of cluster
or server groups rather than assuming the application -> cluster
relationship is valid.

### DIFF
--- a/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/SqlProviderCache.kt
+++ b/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/SqlProviderCache.kt
@@ -7,12 +7,10 @@ import com.netflix.spinnaker.cats.cache.DefaultCacheData
 import com.netflix.spinnaker.cats.cache.WriteableCache
 import com.netflix.spinnaker.cats.provider.ProviderCache
 import com.netflix.spinnaker.cats.sql.cache.SqlCache
-import com.netflix.spinnaker.clouddriver.core.provider.agent.Namespace.CLUSTERS
-import com.netflix.spinnaker.clouddriver.core.provider.agent.Namespace.NAMED_IMAGES
-import com.netflix.spinnaker.clouddriver.core.provider.agent.Namespace.ON_DEMAND
-import kotlin.contracts.ExperimentalContracts
+import com.netflix.spinnaker.clouddriver.core.provider.agent.Namespace.*
 import org.slf4j.LoggerFactory
 import org.slf4j.MDC
+import kotlin.contracts.ExperimentalContracts
 
 @ExperimentalContracts
 class SqlProviderCache(private val backingStore: WriteableCache) : ProviderCache {
@@ -166,27 +164,16 @@ class SqlProviderCache(private val backingStore: WriteableCache) : ProviderCache
     try {
       MDC.put("agentClass", "$source putCacheResult")
 
-      // TODO every source type should have an authoritative agent and every agent should be authoritative for something
-      // TODO terrible hack because no AWS agent is authoritative for clusters, fix in ClusterCachingAgent
-      // TODO same with namedImages - fix in AWS ImageCachingAgent
-      if (
-        source.contains("clustercaching", ignoreCase = true) &&
-        !authoritativeTypes.contains(CLUSTERS.ns) &&
-        cacheResult.cacheResults
-          .any {
-            it.key.startsWith(CLUSTERS.ns)
-          }
-      ) {
-        authoritativeTypes.add(CLUSTERS.ns)
-      } else if (
-        source.contains("imagecaching", ignoreCase = true) &&
-        cacheResult.cacheResults
-          .any {
-            it.key.startsWith(NAMED_IMAGES.ns)
-          }
-      ) {
-        authoritativeTypes.add(NAMED_IMAGES.ns)
-      }
+      // This is a hack because some types are global and a single agent
+      // can't be authoritative for cleanup but can supply enough
+      // information to create the entity. For those types we need an out
+      // of band cleanup so they should be cached as authoritative but
+      // without cleanup
+      //
+      // TODO Consider adding a GLOBAL type for supported data types to
+      //  allow caching agents to explicitly opt into this rather than
+      //  encoding them in here..
+      val globalTypes = getGlobalTypes(source, authoritativeTypes, cacheResult)
 
       cacheResult.cacheResults
         .filter {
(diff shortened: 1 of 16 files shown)


robzienert (commented):
How often do these cleanup agents run? Are they just going to be scheduled as quickly as possible?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/ClusterCleanupAgent.java
```diff
@@ -0,0 +1,71 @@
...
+import com.netflix.spinnaker.cats.provider.ProviderCache;
+import com.netflix.spinnaker.clouddriver.titus.TitusCloudProvider;
+import com.netflix.spinnaker.clouddriver.titus.caching.Keys;
+import com.netflix.spinnaker.clouddriver.titus.caching.TitusCachingProvider;
+import com.netflix.spinnaker.clouddriver.titus.caching.utils.CachingSchema;
+import java.util.*;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class ClusterCleanupAgent implements CachingAgent {
```
cfieber, line 19: This is super similar to the ClusterCleanupAgent in aws - I could probably not have these basically be C&P of each other

cfieber:
They are on the default interval as the rest of the agents. They complete super quick so I'd expect them to run pretty consistently right on that interval.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonApplicationProvider.groovy
```diff
@@ -36,48 +30,53 @@ import static com.netflix.spinnaker.clouddriver.core.provider.agent.Namespace.*
...
+    // rather than the application cacheData which is not reliably updated or evicted
+    Map<String, Map<String, Set<String>>> appClusters = getAppClustersByAccount(allAwsGlob)
+    return appClusters.findResults {translate(it.key, appClusters) }
   }
 
   @Override
   Application getApplication(String name) {
-    translate(cacheView.get(APPLICATIONS.ns, Keys.getApplicationKey(name)))
+    name = name.toLowerCase()
+    String glob = Keys.getServerGroupKey("${name}*", "*", "*", "*")
```
ajordens, line 53: Should this be `${name}-*` ... would it match on `clouddriver` and `clouddriverproxy` otherwise?

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/providers/TitusApplicationProvider.groovy
```diff
@@ -29,51 +25,55 @@ import com.netflix.spinnaker.clouddriver.titus.model.TitusApplication
...
+    // rather than the application cacheData which is not reliably updated or evicted
+    Map<String, Map<String, Set<String>>> appClusters = getAppClustersByAccount(allTitusGlob)
+    return appClusters.findResults {translate(it.key, appClusters) }
   }
 
   @Override
   Application getApplication(String name) {
-    translate(cacheView.get(APPLICATIONS.ns, Keys.getApplicationKey(name)))
+    name = name.toLowerCase()
+    String glob = Keys.getServerGroupV2Key("${name}*", "*", "*", "*")
```
ajordens, line 54: Potentially same comment as above here.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/SqlProviderCache.kt
```diff
@@ -166,27 +164,12 @@ class SqlProviderCache(private val backingStore: WriteableCache) : ProviderCache
...
-      } else if (
-        source.contains("imagecaching", ignoreCase = true) &&
-        cacheResult.cacheResults
-          .any {
-            it.key.startsWith(NAMED_IMAGES.ns)
-          }
-      ) {
-        authoritativeTypes.add(NAMED_IMAGES.ns)
-      }
+      // TODO terrible hack because some types are global and a single agent can't be authoritative
```
ajordens, line 167: Do you have some weird ide code formatting ... seems like your first line `//` is whitespace differently than what follows.

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/caching/agents/TitusStreamingUpdateAgent.java
```diff
@@ -142,9 +142,9 @@
       unmodifiableSet(
           Stream.of(
                   AUTHORITATIVE.forType(SERVER_GROUPS.ns),
-                  AUTHORITATIVE.forType(CLUSTERS.ns),
-                  AUTHORITATIVE.forType(APPLICATIONS.ns),
                   AUTHORITATIVE.forType(INSTANCES.ns),
+                  INFORMATIVE.forType(CLUSTERS.ns),
```
ajordens, line 146: Might be worth a comment here for our future selfs.

ajordens:
I think I'm following and it looks sensible. 

I think it'd be worth taking a pass through with an eye towards inline comments so that anyone with slightly less historical context can understand what's going on.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonApplicationProvider.groovy
cfieber: this is intentional - the globbing doesn't really support regex syntax. The case where it matches multiple applications is handled by grouping all the server groups in getAppClustersByAccount. There may be extra apps in there but they will not end up in the result

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/SqlProviderCache.kt
cfieber: I think it is an idea default but I can fix

cfieber: this is an intellij feature where a TODO includes all the indented lines

I've just marked this as not a TODO and rather a (normally formatted) comment about why it is here

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonApplicationProvider.groovy
cfieber: (if we did `${name}-` we would require all server groups had a stack or detail)

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/SqlProviderCache.kt
cfieber: I did end up leaving a multi-line (funky indented) TODO in there but separated out a bunch of the comment into a more conventional format

ajordens (approved):

PR #5214, opened 2021-01-28
feat(titus): Preliminary support for configuring placement and license details on launch templates

ajordens:
The API is effectively a pass through.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
@@ -131,6 +131,16 @@ class BasicAmazonDeployDescription extends AbstractAmazonCredentialsDescription 
   Source source = new Source()
   Map<String, String> tags
 
+  /**
+   * Launch template placement details, see {@link com.amazonaws.services.ec2.model.LaunchTemplatePlacementRequest}.
+   */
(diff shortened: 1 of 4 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
```diff
@@ -131,6 +131,16 @@ class BasicAmazonDeployDescription extends AbstractAmazonCredentialsDescription
   Source source = new Source()
   Map<String, String> tags
 
+  /**
+   * Launch template placement details, see {@link com.amazonaws.services.ec2.model.LaunchTemplatePlacementRequest}.
+   */
+  LaunchTemplatePlacement placement
```
ajordens, line 137: The sheer number of attributes on this description is getting a little overwhelming. 

It would have perhaps been nice to bucket some of these launch template-specific attributes into their own nested object but I'm afraid that ship may have sailed.

Not trying to rock the boat too much -- just passing through attributes as expected by AWS.

robzienert: It's not too late, if we do really want to change things up, yet it would be non-trivial work. We'd have to create a custom deserializer for this description which would be able to convert the old format to the new type structure. We did this in Orca a few times with queue messages and Keel has this pattern baked-in for the various resource types it supports for API versioning.

robzienert (approved):

PR #5215, opened 2021-01-28
feat(credentials): Use GitHub APIs to download artifacts in a single request

ncknt:
Add 2 new settings to GitHub accounts (`credentials.github.accounts`) to download files directly without a second request:
- `useContentAPI` (defaults to `false`): when true, clouddriver will send the [proper media type request](https://docs.github.com/en/rest/overview/media-types) and download files directly.
- `githubAPIVersion` (defaults to `v3`)

It only needs a single request and as far as I know, it is the recommended way and works in all cases. We could default to t his method if it runs fine for some time.


kevinawoo (approved):
Lgtm!

PR #5216, opened 2021-01-28
fix(titus): Add find by job ID and use in UpsertTitusScalingPolicyAtomicOperationConverter

jonsie:


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/converters/UpsertTitusScalingPolicyAtomicOperationConverter.groovy
```diff
@@ -53,7 +53,7 @@ class UpsertTitusScalingPolicyAtomicOperationConverter extends AbstractAtomicOpe
 
     try {
       def titusClient = titusClientProvider.getTitusClient(converted.credentials, converted.region)
-      def titusJob = titusClient.getJobAndAllRunningAndCompletedTasks(converted.jobId)
+      def titusJob = titusClient.findJobById(converted.jobId, false)
```
jonsie, line 56: No need to get a job and all running and completed tasks when we're simply looking for an app name in a converter - this is needlessly expensive.

aravindmd (approved):

ajordens (approved):

PR #5218, opened 2021-01-29
fix(ecs): check for already suspended autoscaling on disable service op

allisaurus:
Fixes https://github.com/spinnaker/spinnaker/issues/6222

Previously, "DisableServiceAtomicOperation" would always call `RegisterScalableTarget` to suspend app autoscaling on the service. However, if the server group is already disabled, this call re-registers the scalable target which prompts the ECS service to be updated with the existing `MinCapacity` as the `DesiredCount` - which then causes a scale-out action (ECS tries to launch that number of tasks).

Since disabling the server group is done prior to deletion, destroying a disabled server group means Spinnaker will try to "re-disable" it before deletion and this seemed to be causing the problem. This change adds a check in the "disable" operation which skips re-registering the scalable target if scaling is already suspended. 

## Testing

**1. have disabled server group**
image

**2. attempt destruction**
image

**3. ECS doesn't see a scale-out action!**
image

**4. Profit (Server Group is destroyed)**
image

**Bonus: task status msgs indicate autoscaling is already suspended on target service**
image

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/DisableServiceAtomicOperation.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/DisableServiceAtomicOperation.java
@@ -17,10 +17,7 @@
 package com.netflix.spinnaker.clouddriver.ecs.deploy.ops;
 
 import com.amazonaws.services.applicationautoscaling.AWSApplicationAutoScaling;
-import com.amazonaws.services.applicationautoscaling.model.RegisterScalableTargetRequest;
-import com.amazonaws.services.applicationautoscaling.model.ScalableDimension;
-import com.amazonaws.services.applicationautoscaling.model.ServiceNamespace;
-import com.amazonaws.services.applicationautoscaling.model.SuspendedState;
+import com.amazonaws.services.applicationautoscaling.model.*;
 import com.amazonaws.services.ecs.AmazonECS;
 import com.amazonaws.services.ecs.model.UpdateServiceRequest;
 import com.netflix.spinnaker.clouddriver.ecs.deploy.description.ModifyServiceDescription;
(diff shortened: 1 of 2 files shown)


clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/DisableServiceAtomicOperationSpec.groovy
```diff
@@ -40,7 +43,42 @@ class DisableServiceAtomicOperationSpec extends CommonAtomicOperation {
...
+      .withScalableTargets(new ScalableTarget()
+        .withSuspendedState(new SuspendedState()
+          .withScheduledScalingSuspended(false)
+          .withDynamicScalingInSuspended(false)
+          .withDynamicScalingOutSuspended(false)))
     1 * autoscaling.registerScalableTarget(_)
     1 * ecs.updateService(_)
   }
+
+  void 'should not suspend autoscaling if it is already suspended'() {
```
piradeepk, line 56: Can you also add a check in the `then` clause to verify that registerScalableTarget api call is not invoked?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/DisableServiceAtomicOperation.java
```diff
@@ -48,26 +45,53 @@ private void disableService() {
...
 
     updateTaskStatus(String.format("Disabling %s server group for %s.", service, account));
     UpdateServiceRequest request =
         new UpdateServiceRequest().withCluster(cluster).withService(service).withDesiredCount(0);
     ecs.updateService(request);
     updateTaskStatus(String.format("Server group %s disabled for %s.", service, account));
   }
+
+  private boolean isSuspended(DescribeScalableTargetsResult describeResult) {
+    if (describeResult.getScalableTargets() != null
```
piradeepk, line 86: Might want to check if describeResult is null (if the api call fails for some reason, it's possible that describeResult is null)

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/DisableServiceAtomicOperation.java
```diff
@@ -48,26 +45,53 @@ private void disableService() {
...
+        && describeResult.getScalableTargets().size() > 0) {
+      ScalableTarget target = describeResult.getScalableTargets().get(0);
```
piradeepk, line 88: Is it possible to have more than one? always curious about a get(0)

piradeepk (commented):
Looks great overall! Had a couple minor comments.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/DisableServiceAtomicOperation.java
allisaurus: since the source call is specifying the resource ID and the name of the scaling dimension there can only be one result AFAICT; but your right `get(0)` is suspicious. Might be a way to do this with a filter lambda.

piradeepk: No worries, I was just curious. If we're sure that we only need to consider the first, I'm okay with getting this in, but if you can add a comment above it mentioning why (and ideally a FIXME).

piradeepk (approved):
Ship it! 💯

PR #5220, opened 2021-01-29
fix(titus): Handle constraint violation errors during retry

aravindmd:
- We have been handling the job pending case, but not the case where the job is already running and we retry , adding the additional check.

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/SubmitTitusJob.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/SubmitTitusJob.java
@@ -96,8 +96,10 @@ public class SubmitTitusJob extends AbstractTitusDeployAction
                 if (isServiceExceptionRetryable(description, e)) {
                   String statusDescription = e.getStatus().getDescription();
                   if (statusDescription != null
-                      && statusDescription.contains(
-                          "Job sequence id reserved by another pending job")) {
+                      && (statusDescription.contains(
+                              "Job sequence id reserved by another pending job")
+                          || statusDescription.contains(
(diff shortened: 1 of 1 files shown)


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/SubmitTitusJob.java
```diff
@@ -96,8 +96,10 @@ public Result apply(@Nonnull SubmitTitusJobCommand command, @Nonnull Saga saga)
                 if (isServiceExceptionRetryable(description, e)) {
                   String statusDescription = e.getStatus().getDescription();
                   if (statusDescription != null
-                      && statusDescription.contains(
-                          "Job sequence id reserved by another pending job")) {
+                      && (statusDescription.contains(
+                              "Job sequence id reserved by another pending job")
+                          || statusDescription.contains(
+                              "Constraint violation - job with group sequence"))) {
```
jonsie, line 102: I don't think we want to get the "next" server group name.  Instead, we want to return the one that was just created.  This will end up creating two server groups with one of them orphaned.

jonsie: I think the solution would be, if we aren't on the first `submitJob` try then we should handle this error and instead of reving the sequence number we should return the running job.

aravindmd: Agree , makes sense , i was assuming the resolve name method is handling the sequence number resolution , going to close this PR out for now.

ajordens:
+1 to what @jonsie called out.

PR #5221, opened 2021-01-29
fix(aws/loadbalancing): Fix draining instances appearing as healthy

dmart:
https://github.com/spinnaker/clouddriver/pull/5060 changed
deriveHealthState to return Draining however deriveInstanceHealthState
does not handle Drain which results in HealthState.Up


dmart:
Perhaps

dreynaud (approved):
LGTM, thanks for the fix!

PR #5222, opened 2021-01-29
fix(titus): When retrying job submission, if submitted job exists then return the job ID from known server group name

jonsie:


clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/deploy/actions/SubmitTitusJob.java
```diff
@@ -102,6 +105,23 @@ public Result apply(@Nonnull SubmitTitusJobCommand command, @Nonnull Saga saga)
...
+                  // Can't do an exact match on the server group name because the error from Titus
+                  // adds dashes if the server group name does not contain a stack or free form
+                  // details, i.e. testapp---v001
+                  else if (statusDescription != null
+                      && statusDescription.contains(
+                          "Constraint violation - job with group sequence")
+                      && statusDescription.contains("exists")
+                      && submissionAttempts.intValue() > 1) {
+
+                    String jobId = titusClient.findJobByName(nextServerGroupName[0]).getId();
```
jonsie, line 117: Thought about parsing the ID out of the exception message but decided against relying too heavily on the format of the message.  If this call fails it will be retried as part of this retry support lambda.

cfieber (approved):

ajordens (approved):

PR #5224, opened 2021-01-31
fix(clouddriver/appengine): Fix Validation errors when multiple services have same named versions

skalle:
Fixes spinnaker/spinnaker#6305

Uses the ServerGroup `loadBalancer` set to validate Appengine versions with identical names in differerent services(Load Balancers).


skalle:
@spinnaker/reviewers  ... ping.

plumpy (approved):

PR #5226, opened 2021-02-02
feat(aws/deploy): Auto-enable ipv6 in test by config

caseyhebebrand:
For apps with launch templates enabled, IPv6 assignment can now be automated by configuration. In this phase of the rollout, only asgs in the `test` environment with the config enabled will be automated.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
@@ -171,6 +171,12 @@ class AutoScalingWorker {
     String launchConfigName = null
 
     if (shouldSetLaunchTemplate(asgConfig)) {
+      if(asgConfig.getAssociateIPv6Address() == null) {
+        def asgConfigEnv = asgConfig.getCredentials().getEnvironment()
+        def autoEnableIPv6 = dynamicConfigService.getConfig(Boolean.class, "aws.features.launch-templates.ipv6.${asgConfigEnv}", false)
+        asgConfig.setAssociateIPv6Address(autoEnableIPv6)
+      }
+
       def asgConfigWithSecGroups = AsgConfigHelper.setAppSecurityGroups(
               asgConfig,
               regionScopedProvider.getSecurityGroupService(),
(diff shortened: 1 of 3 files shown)


aravindmd (approved):
nit : If you want more control to test this without code change the env/region can be read from a dynamic config

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
```diff
@@ -171,6 +171,10 @@ class AutoScalingWorker {
     String launchConfigName = null
 
     if (shouldSetLaunchTemplate(asgConfig)) {
+      if (asgConfig.getCredentials().getEnvironment() == 'test' && dynamicConfigService.isEnabled("aws.features.launch-templates.ipv6", false)) {
```
ajordens, line 174: It'd be kinda nice to not bind the oss code directly against our environments. Perhaps a configuration option.

Even ... `aws.features.launch-templates.ipv6.${enviroment}` might work.

caseyhebebrand: Long term it will be enabled in all environments by default, but this is not something we want to force oss to do. This this is a good suggestion to prevent that.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/AutoScalingWorker.groovy
```diff
@@ -171,6 +171,12 @@ class AutoScalingWorker {
     String launchConfigName = null
 
     if (shouldSetLaunchTemplate(asgConfig)) {
+      if(asgConfig.getAssociateIPv6Address() == null) {
+        def asgConfigEnv = asgConfig.getCredentials().getEnvironment()
+        def autoEnableIPv6 = dynamicConfigService.getConfig(Boolean.class, "aws.features.launch-templates.ipv6.${asgConfigEnv}", false)
+        asgConfig.setAssociateIPv6Address(autoEnableIPv6)
+      }
```
caseyhebebrand, line 178: @ajordens I refactored a bit so this wouldn't override any inputs from the UI/API request. This value will now be `null` if the user hasn't explicitly set it. Now this auto-assignment will only occur if launch templates are enabled, the user has not opted into IPv6, **_and_** IPv6 is auto enabled. 

Let me know if that makes sense. Basically, instead of defaulting in the `BasicAmazonDeployDescription`, I am intercepting with the config and setting here.

ajordens: This looks good to me.

PR #5230, opened 2021-02-04
fix(aws): Update only spinnaker specific tags during security group u…

aravindmd:
- Previously we were updating all available tags during `upsert` operation , this fails with `UnauthorizedOperation` error if some additional tags are present on the resource which have more restricted permissions. This fix will filter tags which are spinnaker specific while updating tags.


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/securitygroup/SecurityGroupLookupFactory.groovy
```diff
@@ -293,8 +293,12 @@ class SecurityGroupLookupFactory {
         DescribeTagsResult tagsResult = amazonEC2.describeTags(describeTagsRequest)
         List<TagDescription> tags1 = tagsResult.getTags()
```
luispollo, line 294: nit: Consider renaming to something more meaningful, like `currentTags`.

luispollo (approved):
LGTM! Thanks for investigating and fixing this!

ajordens (approved):

PR #5232, opened 2021-02-05
fix(titus): Support for entrypoint and cmd list

srekapalli:
-- Accept entrypoint and command as an array and pass it along to titus.
-- Currently users cannot specify either entry point or cmd in exec form(Json Array Format) and can only give one string and this change will allow users to specify entry and cmd in multiple parts.

### DIFF
--- a/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/Job.java
+++ b/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/Job.java
@@ -38,7 +38,9 @@ public class Job {
   private String user;
   private String version;
   private String entryPoint;
+  private List<String> entryPointList;
   private String cmd;
+  private List<String> cmdList;
   private String iamProfile;
   private String capacityGroup;
   private Boolean inService;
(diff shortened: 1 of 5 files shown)


aravindmd (approved):

clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/client/model/Job.java
```diff
@@ -131,12 +133,14 @@ public Job(
...
     version = grpcJob.getJobDescriptor().getContainer().getImage().getTag();
     digest = grpcJob.getJobDescriptor().getContainer().getImage().getDigest();
-    entryPoint =
-        grpcJob.getJobDescriptor().getContainer().getEntryPointList().stream()
-            .collect(Collectors.joining(" "));
-    cmd =
-        grpcJob.getJobDescriptor().getContainer().getCommandList().stream()
-            .collect(Collectors.joining(" "));
+
+    entryPointArray =
```
ajordens, line 137: The `*Array` naming is throwing me a bit, Titus is calling it `entryPointList`?

Is there a reason not to have `entryPoint` and `entryPoints`? `cmd` and `cmds`?

srekapalli: Yh. I struggled with that name. `entryPoints` or `entryPointList` sounded like there can be more than one entry point where as in the real world those are kind of parts of what makes an entryPoint and same is true for cmd. The final payload that gets sent to titus is a list/array of strings that defines the full entryPoint and cmd.
Docker refers to those as exec form array so I ended up using that
```The exec form is parsed as a JSON array, which means that you must use double-quotes (“) around words not single-quotes (‘).```

srekapalli: @ajordens :  made it as `entryPointList` and `cmdList` similar to titus api.

PR #5233, opened 2021-02-08
fix(entitytags): include required module

deverton:
The removal of the AWS integration with Entity Tags as part of #4923 mean that the required modules for core Entity Tags functions was not included in the `clouddriver` build.

This change changes the dependency on `clouddriver-elasticsearch` in `clouddriver-web` from test only to implementation. This ensure that the ES module is included in the build.

Fixes spinnaker/spinnaker#6331


deverton:
@spinnaker/reviewers any chance you could take a look at this?

deverton:
@cfieber @ajordens would you mind taking a look at this?

PR #5237, opened 2021-02-09
feat(aws): Support for automatically tagging volumes created as a result of autoscaling activity

ajordens:
Implementations of the `AmazonResourceTagger` interface will be used to generate one or more
`LaunchTemplateTagSpecificationRequest` that will be attached to the generated `LaunchTemplate`.

When enabled, `DefaultAmazonResourceTagger` will generate two tags:
* `spinnaker:application`
* `spinnaker:cluster`

Configuration:
```
aws:
  defaults:
    resourceTagging:
      enabled: true
      # clusterTag: spinnaker:cluster
      # applicationTag: spinnaker:application
```


aravindmd (approved):
LGTM

PR #5238, opened 2021-02-09
(spinnaker#6334) Fix to re up the aws-iam-authenticator binary version

battlecow:
(fix spinnaker#6334): Reup the version of aws-iam-authenticator to support IRSA

Fix the version of the aws-iam-authenticator and typo of the environment variable in the dockerbuild.

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -5,7 +5,7 @@ LABEL maintainer="sig-platform@spinnaker.io"
 ENV KUBECTL_RELEASE=1.18.10
 ENV AWS_CLI_VERSION=1.18.152
 ENV AWS_CLI_S3_CMD=2.0.2
-ENV AWS_AIM_AUTHENTICATOR_VERSION=0.4.0
+ENV AWS_IAM_AUTHENTICATOR_VERSION=0.5.2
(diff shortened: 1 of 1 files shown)


jasonmcintosh:
See post I made referencing issues with performance on this.
https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/347
and https://github.com/spinnaker/clouddriver/pull/5196#discussion_r572543644 I'd recommend using the AWS cli instead of aws-iam-authenticator  due to some of the issues.

battlecow:
Thanks @jasonmcintosh for the quick response! This is really good to know as the change seemed to be a mistake and not purposeful revert of the version.  Modifying all of my kube configs now...

jasonmcintosh:
np - yeah we hit this on armory side... :) We pulled the docker file we used for ours into OSS to help the community but we didn't exactly document some of the "WHY" on the versions we used on these ;) If you need help let me know - I'd not dug into the binary, and .. 0.5.2 WAS better, but with the number of k8s calls clouddriver makes, it is still pretty nasty :( AWS CLI also can do things like caching credentials & other stuff.

PR #5243, opened 2021-02-11
fix(aws): Additional checks during updates to security group tags

aravindmd:
In https://github.com/spinnaker/clouddriver/pull/5230 , logic to filter original `tags` based on `description.tags` was added 
This change adds a couple of additional checks 

- Adds null safe checks for `description.tags` 
- Handle a case where all additional tags are removed completely during an upsert, which would make the `description.tags` empty and will result in not removing the additional tag from the resource, introducing a dynamic config with these additional tags names as a failsafe ensures these additional tags are always part of the `deleteTag` request.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/securitygroup/SecurityGroupLookupFactory.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/securitygroup/SecurityGroupLookupFactory.groovy
@@ -24,6 +24,7 @@ import com.netflix.spinnaker.clouddriver.aws.security.AmazonClientProvider
 import com.netflix.spinnaker.clouddriver.aws.security.NetflixAmazonCredentials
 import com.netflix.spinnaker.credentials.CredentialsRepository
 import com.netflix.spinnaker.kork.core.RetrySupport
+import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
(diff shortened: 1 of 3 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/securitygroup/SecurityGroupLookupFactory.groovy
```diff
@@ -294,9 +295,10 @@ class SecurityGroupLookupFactory {
         List<TagDescription> currentTags = tagsResult.getTags()
         Collection<Tag> oldTags = new HashSet()
         // Filter Spinnaker specific tags, update to other tags might result in permission errors
+        def additionalTags = dynamicConfigService.getConfig(String.class, "aws.features.security-group.additional-tags","")
         currentTags.each {
           it ->
-            if (it.key.equals("Name") || description.tags.keySet().contains(it.key)) {
+            if (it.key.equals("Name") || description.tags?.keySet()?.contains(it.key) || additionalTags.contains(it.key)) {
```
ajordens, line 301: Can we get a real-world example of what would be in `additionalTags`? Is it really something that should be backed by the `DynamicConfigService` or should it be some config object that wraps `List<String>` (rather than just a plain `String`?

Just so I'm understanding this and the previous PR, the goal here is to only remove a subset of all potential tags on the security group. 

Effectively "Name" and any others that have been supplied as part of the description?

I assume that if there are policies governing special tags that would prevent removal, that they might also prevent any change?

PR #5244, opened 2021-02-11
feat(blockDevices/tags): allows specify Tags for all the block devices

jorgebee65:
This feature allows specify Tags for all the block devices using **blockDevicesTags** on the Pipeline JSON format.

```
      "clusters": [
        {
          "account": "aws-psdev",
          "application": "awsk8",
          "blockDevices": [
            {
              "deleteOnTermination": true,
              "deviceName": "/dev/sda1",
              "size": 8,
              "volumeType": "gp2"
            }
          ],
          "blockDeviceTags": {
            "Creator": "Armory"
          }
        ...
        }
```
Image 2021-02-11 at 1 47 01 PM


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/services/LaunchTemplateService.java
```diff
@@ -297,6 +301,16 @@ private RequestLaunchTemplateData buildLaunchTemplateData(
                   .collect(Collectors.toList()));
     }
 
+    if (!asgConfig.getBlockDevicesTags().isEmpty()) {
```
ajordens, line 304: Might be preferable if this was an implementation of the `AmazonResourceTagger` interface. 

I recently added that to also support ebs volume tagging. The interface may need to change to support passing in the `AsgConfig`.

jorgebee65: I created a `CustomAmazonResourceTagger` class and use it in `AsgConfiguration` (AsgConfig)

ajordens:
Sorry for the delay here ... I pulled this PR locally and tried to make a few changes but for one reason or another it didn't work to push the changes back here.

I think this (https://github.com/spinnaker/clouddriver/pull/5289/files) would be a little cleaner, it extends the existing block device tagger but I think does so in a reasonable way.

jorgebee65:
@ajordens I made the suggested changes.

ajordens (approved):

kinnonyee-armory:
The following change does not work for 1.26.x.  AWS advises using Launch Templates to define volumes instead. 
https://spinnaker.io/docs/setup/other_config/server-group-launch-settings/aws-ec2/launch-templates/

PR #5246, opened 2021-02-11
fix(cloudfoundry): add equals and hash code to managed account

zachsmith1:
When using Spring Cloud Config, or another dynamic account solution, at some point the credentials definition will be compared against the existing definitions. This ensures that we only update credentials when they actually are updated/changed. Prior to this fix, during the parsing/updating phase it would see every credential as not equal to any other definitions, so it would remove the old and add the new. If the refresh interval is set to something like 30 seconds, this means every 30 seconds the credentials loader will remove all agents and re-create them for the new credentials. This leads to unexpected caching and numerous downstream problems. I.e. some agents may never run. 

This fix adds the annotation to ensure credentials are compared correctly with the equals operation and a test is included to ensure the regression doesn't happen again.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,8 +8,6 @@ on:
 
 env:
   GRADLE_OPTS: -Dorg.gradle.daemon=false -Xmx6g -Xms6g
-  TESTCONTAINERS_RYUK_DISABLED: true
-  TESTCONTAINERS_CHECKS_DISABLE: true
 
 jobs:
   branch-build:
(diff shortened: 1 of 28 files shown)


german-muzquiz:
For context: This is a regression introduced by [this PR](https://github.com/spinnaker/clouddriver/pull/4959) on version `1.23`. Versions `1.24` and later were fixed by [this PR](https://github.com/spinnaker/clouddriver/pull/5150).

Before `kork-credentials` refactor, account equality check was done at [CloudFoundryCredentials](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/security/CloudFoundryCredentials.java#L171) level instead of `CloudFoundryConfigurationProperties.ManagedAccount`.

german-muzquiz (approved):

link108 (approved):

PR #5247, opened 2021-02-12
fix(test): Avoid namespace reuse in integration tests

german-muzquiz:
Current kubernetes infrastructure integration tests are sometimes failing: https://github.com/spinnaker/clouddriver/actions/runs/559938714

Those tests reuse a single namespace in order to speed up execution by not doing deployments on every test case. However that's likely the culprit for test flakiness, so this PR isolates test cases to different namespaces.


ncknt:
Have you been able to measure how much time and memory it adds to the integration tests? I don't see a cleanup phase in the tests, perhaps it could help with memory a bit? Also nginx is not big but perhaps busybox would be a tag lighter?

Otherwise LGTM.

german-muzquiz:
I tried to run again the job but cache took place. I saw the time increase roughly from 12 to 15 minutes the first run.

Regarding cleanup I'm worried that doing it will introduce unnecessary time in the build, specially having the k3s container (and everything in there) getting killed when tests finish anyway. However the memory usage of that k3s container is a good concern as it will have a lot of nginx containers running around. I opted for more reliable tests instead of trying to save execution time. I'll change to a lighter image anyway for reducing exeuction memory footprint.

german-muzquiz:
Identified that test flakiness was due to caching agents not finishing within the time tests were waiting. After increasing the max waiting time from 1 minute to 5 minutes, tests have not been failing after 6 consecutive runs.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
```diff
@@ -175,6 +175,9 @@ class SqlClusteredAgentScheduler(
       .filterNot { disabledAgents.contains(it.key) }
       .toMutableMap()
 
+    log.debug("Agents running: {}, agents disabled: {}. Picking next agents to run from: {}",
```
ncknt, line 178: I like the additional log. Do you want to keep the fix/shuffle separate?

ncknt (approved):

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
german-muzquiz: I'll make another PR for the shuffle

PR #5248, opened 2021-02-12
fix(kube): Redblack strategy uses service manifest from stage

german-muzquiz:
Fixes https://github.com/spinnaker/spinnaker/issues/6215

When using `Deploy (Manifest)` stage with Red/Black rollout strategy, a requirement was for a `Service` to exist before running the stage in order to read its `selector` and apply it to the newly deployed replicaset. There are use cases where the `Service` gets deployed in that same stage, for example when deploying a helm chart, so the first time the stage is executed it fails because of a mising deployed `Service`, and subsequent runs are successful.

With this change, now the `Service` will first be read from the stage manifest list, and if not found it will proceed with the same behavior as before, reading it live from the cluster. Most of the modified files are for integration tests to take into account this scenario.


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/KubernetesServiceHandler.java
```diff
@@ -181,6 +182,11 @@ private String podLabelKey(String namespace, Map.Entry<String, String> label) {
   @Override
   @ParametersAreNonnullByDefault
   public void attach(KubernetesManifest loadBalancer, KubernetesManifest target) {
+    // avoid loops
```
ncknt, line 185: How we can we reach that state? Does it result in an actual loop? I feel that if we end up attempting to attach a load balancer onto a load balancer something is wrong with our abstraction.

german-muzquiz: [This line](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/manifest/KubernetesDeployManifestOperation.java#L186) applies traffic to every manifest in the request, and "applying traffic" means adding labels to the manifests so that the kubernetes `Service` selector can locate them.

I think the right way to do things is to only label traffic related objects like `ReplicaSet` and `Pod`. Since today _all_ objects are labeled with the service selector (even the same service being deployed) I wonder if changing it would provide some backwards compatibility issues, it's not unusual that users inadvertently rely on spinnaker bugs for their pipelines, we fix the bugs and pipelines break. Maybe a good compromise is to keep labeling everything but show a deprectation warning and in a later spinnaker version only label traffic related objects?

And to answer your question I put in place that "avoid loops" protection because I faced an issue without it, but now I ran tests without it successfully and I don't remember what was the issue...

ncknt: Yeah at worst, I see the service changing its own label if it has the `traffic.spinnaker.io/load-balancers` annotation but I can't see how this would result in a loop. I'd remove that special case unless we have a plan to address what load balancers can target. And I don't think we should assume it's only the built-in types as users could define their own handlers.

+1 on the warning. AFAIK, services can also target `StatefulSet` and `DaemonSet`.

ncknt (approved):
LGTM

PR #5249, opened 2021-02-12
feat(core): add namespace provider

zachsmith1:
The new NamespaceProvider will support querying accounts for namespaces, spaces, etc. There are places across spinnaker, like Deck, that need this information to populate fields for users. Before this change, this data is gathered by calling the /credentials endpoint and then parsing the namespaces/etc from that account. It doesn't make sense to continue to parse this information from a credentials object directly. Allowing this data to be fetched independently allows for better scalability and control.


clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/NamespaceProvider.java
```diff
@@ -0,0 +1,36 @@
...
+public interface NamespaceProvider {
+
+  /**
+   * Looks up all of the namespaces for a particular account. Keyed on account name.
+   *
+   * @param account name
+   * @return map of namespace-names to namespaces objects.
+   */
+  @Empty
+  Map<String, Object> getNamespaces(String account);
```
robzienert, line 32: `Map<String, Object>` is not a helpful return type - we shouldn't be propagating this type of method signature to new places in the codebase wherever possible. Is there a common interface we can implement here?

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/NamespaceController.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.security.access.prepost.PreAuthorize;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/namespaces")
+public class NamespaceController {
+
+  @Autowired List<NamespaceProvider> namespaceProviders;
```
robzienert, line 33: Avoid `@Autowired` properties: Prefer constructor injection.

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/NamespaceProvider.java
zachsmith1: I agree, and yes, I hope we could think about a common interface.

If we created a Namespace interface, to begin it could have a single method signature ->
 `String getName()`
 
This would satisfy the most basic interactions and leave room for further development.

robzienert: Sounds good.

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/model/NamespaceProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.clouddriver.model;
+
+import com.netflix.spinnaker.clouddriver.documentation.Empty;
+import java.util.Collection;
+
+public interface NamespaceProvider {
```
ncknt, line 23: We're using the term `region` in many places (coming from AWS), `location` in others. I'm hesitant to add one more term. I think it only matches Kubernetes (?)

zachsmith1: Yes, Namespace in its literal sense is a k8 object but the abstraction is more or less what I was trying to accomplish. Namespace being a logical grouping around some resource. This happens to be a 1-1 with k8 and very close to 1-1 with CF (Org + Space combo). These are also dynamic in nature and can be created/updated/deleted. The tight coupling between the credentials and these "namespaces" is the main reason for this PR. Regions being more static might not make sense in this light

I'm open to calling it something different if you have a better abstraction in mind?

ncknt: I was thinking `location` to be consistent. It's not a very strongly held feeling on my side, I just wanted you to be aware that it's yet another naming scheme.  Alternative idea: we could define: `/credentials/kubernetes/[account name]/namespaces` for Kubernetes, `/credentials/cloudfoundry/[account name]/spaces` for CF, etc. I guess it depends how you plan to use this in Deck.

PR #5250, opened 2021-02-13
fix(titus): Set default disruption budget based on job type

aravindmd:
Titus batch jobs submitted through a Run Job pipeline stage defaults to a global default policy, This policy is well suited for services, but inappropriate for batch jobs as it allows to kill running batch jobs, Setting it to `self-managed` is a better option since this has the setting on Titus to not disrupt the jobs.


aravindmd:
Confirmed with Titus , this should work fine.

robzienert (approved):

robzienert:
Would do well to have a test for this.

aravindmd:
> Would do well to have a test for this.

Agree , will look at adding one

PR #5254, opened 2021-02-19
fix(sql): Sql agent scheduler shuffling

german-muzquiz:
When the number of caching agents needed to run exceeds the value in `sql.agent.max-concurrent-agents`, some agents need to be dropped and wait for some agents to complete before running.

The sql scheduler iterates over the collection of pending agents and only executes the first ones on that collection until reaching the max limit. It turns out that the order of iteration is always the same, so there are agents that are never executed in situations of passing the max limit.

This PR addresses this issue by shuffling the collection before iterating over it.


cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
```diff
@@ -211,15 +211,21 @@ class SqlClusteredAgentScheduler(
     }
 
     log.debug("Next agents to run: {}, max: {}", candidateAgentLocks.keys, availableAgents)
+    val shuffledCandidates = mutableMapOf<String, AgentExecutionAction>()
```
ncknt, line 214: I don't think it does what you want as it's relying on the map iteration sequence for the randomness. You could just move the `shuffled()` below instead of having that intermediate map.

german-muzquiz: Tests before and after the change showed that agents were randomly executed with the shuffled map, and they were executed always in the same order before. I agree that the intermediate map is not needed and I can use `shuffled` in the original map though 👍

ncknt (approved):

PR #5258, opened 2021-02-24
fix(ecs): Malformed ECS Task Execution Role Arn in China partition

paragbhingre:
This PR fixes issue [#6053](https://github.com/spinnaker/spinnaker/issues/6053)

Previously, in case of china and gov-cloud partition ECS used to send malformed ARN (e.g. `arn:aws:iam::123123123:arn:aws-cn:iam:123123123:role/roleName`) to AWS which was causing this issue.

 GovCloud and China regions need to have the full arn passed because of differing formats

- Govcloud: arn:aws-us-gov:iam 
- China: arn:aws-cn:iam

This change allows clouddriver to send full valid ARN (e.g. `arn:aws-cn:iam:123123123:role/roleName`) in case of china and gov-cloud and not the malformed one.


clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperationSpec.groovy
```diff
@@ -1760,4 +1760,131 @@ class CreateServerGroupAtomicOperationSpec extends CommonAtomicOperation {
     ex.message == "ECS account settings for account Test do not allow tagging as `serviceLongArnFormat` and `taskLongArnFormat` are not enabled."
   }
 
+  def 'should return valid execution task def arn for non-china and non-gov-cloud partition'() {
```
allisaurus, line 1763: I think you mean "task **execution role** arn" (same goes for other added test cases)

allisaurus (commented):
Awesome, thanks for fixing this @paragbhingre! I just have one small comment on the change. 

If you can address that and fix the commit message (matching the PR title would work), that should fix the [spinnakerbot complaint](https://github.com/spinnaker/clouddriver/pull/5258#issuecomment-784719358) and then I think this is good to go.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -743,8 +743,11 @@ private String inferAssumedRoleArn(AmazonCredentials credentials) {
           "The given kind of credentials is not supported, "
               + "please report this issue to the Spinnaker project on Github.");
     }
-
-    return String.format("arn:aws:iam::%s:%s", credentials.getAccountId(), role);
+    if (!role.startsWith("arn:")) {
+      return String.format("arn:aws:iam::%s:%s", credentials.getAccountId(), role);
+    } else {
```
allisaurus, line 748: since we're returning in the previous `if` statement, we don't need this `else` block. We can just do:
```
...
if (!role.startsWith("arn:")) {
      return String.format("arn:aws:iam::%s:%s", credentials.getAccountId(), role);
}
return role;
...
```

clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperationSpec.groovy
paragbhingre: Ohh, yes.

allisaurus (approved):

zweizhang (commented):
Hi @paragbhingre .
Actually, I got the same issue when I deploy clouddriver in `EKS` china region. (Not ECS)
I checked your PR and found the arn defind with `arn:aws-cn:iam:123123123123:role/test-role` ,
But the real full ARN should be `arn:aws-cn:iam::123123123123:role/test-role`, which two `:` between `iam` and `123123123123` ,not just one `:` .
So I wonder if you can help me to know which one is right ? Thanks in advance.

paragbhingre:
> Hi @paragbhingre .
> Actually, I got the same issue when I deploy clouddriver in `EKS` china region. (Not ECS)
> I checked your PR and found the arn defind with `arn:aws-cn:iam:123123123123:role/test-role` ,
> But the real full ARN should be `arn:aws-cn:iam::123123123123:role/test-role`, which two `:` between `iam` and `123123123123` ,not just one `:` .
> So I wonder if you can help me to know which one is right ? Thanks in advance.

Hello @zweizhang, to answer your question `::` between `iam` and `123123123` is the right format. Thank you for catching that there is only one `:` in my PR. 
This PR works perfectly fine as code does provide `::` between `iam` and `123123123` but my explanation in the PR missed it. I will change the description so that it will not mislead anyone else. Again Thank you for your question.

PR #5259, opened 2021-02-24
feat(gitrepo): Git pull for downloading repo updates

german-muzquiz:
This PR introduces the ability to do `git pull` in gitrepo artifacts for downloading repository updates, in case the repo has been already cloned by a previous call.

Instead of cloning to a random directory and deleting it upon return of the fetch artifact call, now repositories are cloned to fixed directories that depend on the artifact coordinates. A cleanup mechanism is also introduced for deleting clones that have not been accessed after some time. This time is controlled by the new configuration setting

```
artifacts.git-repo.clone-retention-minutes
```

Which defaults to 0, which disables retention by default and deletes the clone immediately after each usage. Setting the value to -1  disables the cleanup mechanism and will keep clones for the lifetime of the pod.

There's also a new setting

```
artifacts.git-repo.clone-retention-max-bytes
```

Defaulting to 100MB, which controls the max disk usage that clones will consume in the pod.


dbyron0:
Thanks for this!

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitJobExecutor.java
```diff
@@ -75,6 +76,29 @@ public GitJobExecutor(
     askPassBinary = initAskPass();
   }
 
+  public void cloneOrPull(String repoUrl, String branch, Path localPath, String repoBasename)
+      throws IOException {
+    if (!Files.exists(localPath)
+        || localPath.toFile().listFiles() == null
+        || localPath.toFile().listFiles().length == 0) {
```
ncknt, line 83: Avoid calling `localPath.toFile().listFiles()` twice and might (unlikely) return different results.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitJobExecutor.java
```diff
@@ -75,6 +76,29 @@ public GitJobExecutor(
...
+      clone(repoUrl, branch, localPath);
+      return;
+    }
+    if (!Files.isDirectory(localPath)) {
```
ncknt, line 87: If you have `localPath.toFile()` above you can use `isDirectory()` on it.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitJobExecutor.java
```diff
@@ -110,6 +134,34 @@ public void clone(String repoUrl, String branch, Path destination) throws IOExce
     }
   }
 
+  public void pull(String repoUrl, Path localPath) throws IOException {
```
ncknt, line 137: This could be `private` (just like `clone` above - although it may be too late for that) unless you think it's useful to expose.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitJobExecutor.java
```diff
@@ -75,6 +76,29 @@ public GitJobExecutor(
...
+    }
+    if (!Files.isDirectory(localPath)) {
+      throw new IllegalArgumentException(
+          "Local path " + localPath.toString() + " is not a directory");
+    }
+    Path clonePath = Paths.get(localPath.toString(), repoBasename);
+    File[] files = clonePath.toFile().listFiles();
+    if (files == null
+        || files.length == 0
+        || Arrays.stream(files).noneMatch(f -> f.getName().equals(".git"))) {
```
ncknt, line 95: The files are ignored so just testing on `clonePath.resolve(".git").toFile().exists()` without listing the files be better. At the same time, if `clonePath` has files shouldn't we give an exception and fallback on pulling?

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoFileSystem.java
```diff
@@ -0,0 +1,80 @@
...
+  private String hashCoordinates(String repoUrl, String branch) {
+    String coordinates =
+        String.format(
+            "%s-%s",
+            Optional.ofNullable(repoUrl).orElse("unknownUrl"),
+            Optional.ofNullable(branch).orElse("defaultBranch"));
+    return Hashing.sha256().hashString(coordinates, Charset.defaultCharset()).toString();
+  }
+
+  @Scheduled(fixedDelay = 1000 * 60)
```
ncknt, line 59: This seems like it should be a setting (wish list)

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoFileSystem.java
```diff
@@ -0,0 +1,80 @@
...
+  private void deleteExpiredRepos() {
```
ncknt, line 60: It seems prone to conflict if the scheduled task tries to clean a repo that another thread is trying to use or clone to.

ncknt (commented):
Overall, I like the idea of the optimization. It will make a difference for users who keep their clouddriver instances running for a while.

If I got that right, the path is predetermined by the request so two requests in a short amount of time will cause concurrency issues. I'd like to see some type of locking or fallback mechanism.

It also means that potentially sensitive data will hang out on instances. We should make that explicit in the documentation.

Finally, I'd like to see a cap on the amount of data kept. If I leave it on, I'll fill the disk and cause runtime errors after some time depending on usage.

So I'd prefer to turn that feature off by default so people can choose if they're comfortable running this optimization.

clouddriver-artifacts/clouddriver-artifacts.gradle
```diff
@@ -56,6 +56,7 @@ dependencies {
   testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine"
 
   integrationImplementation project(":clouddriver-web")
+  integrationImplementation project(":clouddriver-artifacts")
```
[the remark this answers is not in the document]
german-muzquiz: It's already like that on line 17 of this file, for some reason integration tests didn't load `src/main` classes of their own module.

clouddriver-artifacts/src/integration/java/com/netflix/spinnaker/clouddriver/artifacts/GitRepoTest.java
```diff
@@ -251,4 +295,15 @@ private void assertBytesHaveFile(byte[] bytes, String... files)
           "File " + file + " not found in response");
     }
   }
+
+  private static void deleteTmpClone(Map<String, Object> artifact) throws IOException {
+    Path localClone =
+        new GitRepoFileSystem(0)
+            .getLocalClonePath(
+                String.valueOf(artifact.get("reference")), String.valueOf(artifact.get("version")));
```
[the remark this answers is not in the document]
german-muzquiz: `null` values are valid but not used yet in these tests. I'll change to cast to `String` instead of `String.valueOf` to preserve true nullness.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitJobExecutor.java
german-muzquiz: If `clonePath` has `.git` we can pull, if not we fallback to clone, which nukes the directory to start fresh again. 
Since we already tested that `localPath` has files above, I'll add a warning in that case indicating that the directory has files but there's no `.git`, so they will be deleted to start with a fresh clone.

german-muzquiz:
@ncknt thanks for the feedback, I've addressed all comments and this functionality is more robust!

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
```diff
@@ -57,24 +61,62 @@ public String getType() {
...
+  private FileInputStream getInputStream(
+      String repoUrl,
+      String subPath,
+      String branch,
+      Path stagingPath,
+      String repoBasename,
+      Path outputFile)
+      throws IOException {
+    try {
+      if (gitRepoFileSystem.tryLock(repoUrl, branch, MAX_WAIT_LOCK_SEC, TimeUnit.SECONDS)) {
```
ncknt, line 92: Isn't 10s too short? We've seen some checkout take minutes. Can we make it user configurable?

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoFileSystem.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+        try {
+          log.info("Deleting expired git clone {}", r.getName());
+          FileUtils.forceDelete(r);
+        } finally {
+          unlock(r.getName());
+        }
+      }
+    } catch (IOException e) {
+      log.error("Error deleting expired git clones, ignoring", e);
```
ncknt, line 125: If we ignore, shouldn't it be a warning?

ncknt:
Looks great! Thanks for making the changes. Just a couple of nits but otherwise LGTM.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoFileSystem.java
german-muzquiz: I think `error` level is appropriate because if the expectation is that clones are deleted and that doesn't work, it could potentially lead to running out of disk and the pod getting killed.

ncknt (approved):

PR #5267, opened 2021-03-02
fix(kubernetes): K8s creds synchronizer not configured when using loader

aufarg:
This PR makes `CredentialsInitializerSynchronizable` to be configured when `CredentialsDefinitionSource<KubernetesConfigurationProperties.ManagedAccount>` is available, not the other way around. This way we can refresh credentials through `/refresh` actuator when we have `CredentialsDefinitionSource`.

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/config/KubernetesConfiguration.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/config/KubernetesConfiguration.java
@@ -30,6 +30,7 @@ import com.netflix.spinnaker.credentials.definition.BasicCredentialsLoader;
 import com.netflix.spinnaker.credentials.definition.CredentialsDefinitionSource;
 import com.netflix.spinnaker.credentials.poller.Poller;
 import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


ncknt:
The "synchronizable" is currently configured when there is no custom `CredentialsDefinitionSource` defined (which means it is using Spring's `KubernetesConfigurationProperties::getAccounts`). This is when we want to reload on Spring context refresh.

If you changed it to `@ConditionalOnBean` then out of the box, that would not work. Also if I add my own `CredentialsDefinitionSource` a Spring context refresh would trigger my custom source when it should not (e.g. to query an external endpoint).

If you have defined your own `CredentialsDefinitionSource` (e.g. via a plugin) and still want the actuator to work then add your own `CredentialsInitializerSynchronizable` or just listen to Spring's `RefreshScopeRefreshedEvent` to reload.

aufarg:
Thanks for the explanation and suggestion! I was under the impression that the value returned by `getAccounts`  is static, but now I realize that's not the case when we're using SCC (right?).

I'll close this this PR then.

ncknt:
Right. If you use SCC, it will update the Spring values (via a context refresh). Just beware that Spring context refresh does not perform well if you have a significant number of clusters. That's why we added these extension points.

PR #5273, opened 2021-03-04
fix(artifacts/helm): look up artifact credentials by supported type

ncknt:
There are 2 types for artifacts: the type of artifacts credentials (e.g. github, gitlab, helm) and the type of artifacts (helm/chart, git/repo). The two endpoints in `ArtifactController` are called from Deck - just for helm artifacts - and pass an artifact type.

This should fix https://github.com/spinnaker/spinnaker/issues/6349.

### DIFF
--- a/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ArtifactController.java
+++ b/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ArtifactController.java
@@ -69,7 +69,7 @@ public class ArtifactController {
   List<String> getNames(
       @PathVariable("accountName") String accountName, @RequestParam(value = "type") String type) {
     ArtifactCredentials credentials =
-        artifactCredentialsRepository.getCredentials(accountName, type);
+        artifactCredentialsRepository.getCredentialsForType(accountName, type);
     return credentials.getArtifactNames();
   }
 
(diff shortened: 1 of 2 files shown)


german-muzquiz:
Since this is a regression, can you add tests to assert that this doesn't break in the future without notice? There's already a `ArtifactControllerSpec` test class that could be used.

clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/ArtifactController.java
```diff
@@ -67,19 +69,24 @@ StreamingResponseBody fetch(@RequestBody Artifact artifact) {
...
       @RequestParam(value = "artifactName") String artifactName) {
     ArtifactCredentials credentials =
-        artifactCredentialsRepository.getCredentialsForType(accountName, type);
+        artifactCredentialsRepository.getCredentialsForType(accountName, artifactType);
     return credentials.getArtifactVersions(artifactName);
   }
+
+  @ExceptionHandler(MissingCredentialsException.class)
+  @ResponseStatus(HttpStatus.NOT_FOUND)
+  public void handleMissingCredentials() {}
```
dbyron0, line 91: Do you figure this belongs in kork eventually?  Like [here](https://github.com/spinnaker/kork/blob/3214d577916fef40de909db7cc9da2ed828d63a3/kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java)?

ncknt: Yes but it's a regression that needs to be backported so no kork.

dbyron0: Aaah, ok.  Makes sense.

link108:
@mergifyio backport release-1.25

link108:
@Mergifyio backport release-1.25.x

PR #5274, opened 2021-03-05
Spinnaker cannot fetch Load Balancers for Azure accounts Exception ra…

dancb10:
…ised in Clouddriver

Spinnaker throws Exception in Clouddriver when fetching VMSS. By default, Spinnaker expects all VMSS to use a LoadBalancer attached and any VMSS which doesn't comply with this rule will end up throwing this Exception. This affects the way Load Balancers are displayed in the UI (If these exceptions are thrown the LB page will not work).
https://github.com/spinnaker/spinnaker/issues/6294

In accounts where there are multiple teams there may be VMSS which have or not an ELB. In such cases, the Azure provider throws errors.
The solution is to tag all VMSS with the necessary tags and when caching VMSS to look for a specific tag. This is to ensure Spinnaker will cache only VMSS which are managed by it.
In case a VMSS without ELB is detected, the isDisabled property is set to false


EasyAsABC123:
@dancb10 need assistance getting this working?

PR #5275, opened 2021-03-06
feat(kubernetes): Breaking, Update API versions

sikevux:
This change makes Spinnaker incompatible with Kubernetes prior to 1.14.

The reason for this change is to enable the now General Available
Ingress API. This was added in Kubernetes 1.19. To make this change the
client library had to be updated. Since 1.16 removed a bunch of APIs
that change now affects us, and a few of the old APIs got removed.

As far as I can tell the oldest supported version of Kubernetes is now
1.14. It was released in 2019, almost two years ago to date. It's been
End of Life since December 2019.

BREAKING CHANGE: This breaks Kubernetes prior to 1.14.

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/KubernetesDaemonSetHandler.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/KubernetesDaemonSetHandler.java
@@ -17,15 +17,18 @@
 
 package com.netflix.spinnaker.clouddriver.kubernetes.op.handler;
 
+import static com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesApiVersion.APPS_V1;
 import static com.netflix.spinnaker.clouddriver.kubernetes.op.handler.KubernetesHandler.DeployPriority.WORKLOAD_CONTROLLER_PRIORITY;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.netflix.spinnaker.clouddriver.kubernetes.artifact.Replacer;
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.Keys.InfrastructureCacheKey;
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.agent.KubernetesCacheDataConverter;
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.agent.KubernetesCachingAgentFactory;
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.agent.KubernetesCoreCachingAgent;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.SpinnakerKind;
+import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesApiVersion;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesKind;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesManifest;
 import com.netflix.spinnaker.clouddriver.kubernetes.model.Manifest.Status;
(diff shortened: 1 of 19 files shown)


german-muzquiz:
Thanks for the contribution!, since this is a breaking change we'll present it in tomorrow's kubernetes SIG meeting for awarness.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/KubernetesDaemonSetHandler.java
```diff
@@ -91,9 +91,8 @@ protected KubernetesCachingAgentFactory cachingAgentFactory() {
 
   @Override
   public Status status(KubernetesManifest manifest) {
-    V1beta2DaemonSet v1beta2DaemonSet =
-        KubernetesCacheDataConverter.getResource(manifest, V1beta2DaemonSet.class);
-    return status(v1beta2DaemonSet);
+    V1DaemonSet v1DaemonSet = KubernetesCacheDataConverter.getResource(manifest, V1DaemonSet.class);
```
german-muzquiz, line 94: Just to be consistent with the other handlers, I think we can add the collection of `SUPPORTED_API_VERSIONS` with only `apps/v1` here, and throwing the `UnssuportedVersionException` if something else is passed in.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/KubernetesStatefulSetHandler.java
```diff
@@ -97,9 +97,9 @@ protected KubernetesCachingAgentFactory cachingAgentFactory() {
 
   @Override
   public Status status(KubernetesManifest manifest) {
-    V1beta2StatefulSet v1beta2StatefulSet =
-        KubernetesCacheDataConverter.getResource(manifest, V1beta2StatefulSet.class);
-    return status(v1beta2StatefulSet);
+    V1StatefulSet v1StatefulSet =
```
german-muzquiz, line 100: Also here please introduce the `SUPPORTED_API_VERSIONS` collection with `apps/v1` as the only entry, like in `Deployment`, `ReplicaSet` and `DaemonSet` handlers.

german-muzquiz (commented):
I just have a couple of minor comments, but everything looks good.
Question: as far as I know we don't have an official list of kubernetes versions supported by spinnaker, so kubernetes versions are basically determined by which API versions individual resources support. Moreover, this seems to be determined by which versions the kubernetes api java client supports.
I cannot find official docs for the kubernetes API prior to 1.16, how do you know that 1.14 would be oldest working version with spinnaker?

sikevux:
> I just have a couple of minor comments, but everything looks good.
> Question: as far as I know we don't have an official list of kubernetes versions supported by spinnaker, so kubernetes versions are basically determined by which API versions individual resources support. Moreover, this seems to be determined by which versions the kubernetes api java client supports.
> I cannot find official docs for the kubernetes API prior to 1.16, how do you know that 1.14 would be oldest working version with spinnaker?

This change could actually work with some older APIs present in 1.13, based on the version matrix for the java client. But given the changes announced in 1.16 that took effect in 1.18, 1.19, and 1.20 there's fewer of those APIs available. The reason why I chose to clearly mark it as breaking for versions prior to 1.14 is that most of the newer APIs where added in that version, if I'm reading the release notes correctly.

Even with all this, the only users that could potentially be negatively affected by this change are the once that deploy Spinnaker on systems other than k8s, or that build their own containers. This because the version of `kubectl` bundled in the containers is 1.18. Given that the only supported versions would be 1.17, 1.18, and 1.19.

I could remove the big warning, just figured it's better to be cautious.

german-muzquiz:
Thanks for bringing attention to this, there are integration tests that use Rancher's k3s cluster but the oldest version available is 1.16, so it's difficult to know what's the behavior in older versions.

As far as I know you can deploy to kubernetes old API versions if they are supported by the server, and when you read them back they get automatically converted to the latest version. So if you apply a `ReplicaSet` manifest with version `apps/v1beta2` in 1.15, when you read it back it will show as `apps/v1`, which is good because spinnaker expects the latest. According to this [Deprecation API Migration Guide](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-16), deprecated APIs in this PR (besides `Ingress`) have their latest version (`apps/v1`) available  since kubernetes 1.9, so it's possible that spinnaker can still work with 1.9 but not earlier.

german-muzquiz (approved):

german-muzquiz:
@sikevux approved, you just need to merge latest changes from `master` to get this PR merged.

PR #5278, opened 2021-03-09
feat(cfn): prefer templateURL over templateBody if provided

jaredstehler:
The `templateBody` field has a size limit of 51200 bytes; larger CFN templates need to be hosted in S3. This PR updates the CloudFormation deploy task to use `templateURL` if provided in the task input.

There will be a subsequent PR on orca to send `templateURL` if the given template artifact has a reference starting with `s3://`.


jaredstehler:
I based this off the `release-1.25.x` branch; I can close and rebase from `master` instead.

PR #5288, opened 2021-03-11
Adding integration tests for AWS EC2 provider

pdk27:
* Adding integration tests for AWS EC2 provider for CreateServerGroup operation (with launch configuration / launch template settings enabled)
* Adding AWS provider integration tests to Github workflow

#### Sample integ test run:
Look for `AWS EC2 Provider Integration Tests` in https://github.com/pdk27/clouddriver/runs/2090227653?check_suite_focus=true

#### Successful run (shows tests summary):
image

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -46,3 +46,5 @@ jobs:
       run: ./gradlew --build-cache :clouddriver-ecs:integrationTest
     - name: Artifacts Integration Tests
       run: ./gradlew --build-cache :clouddriver-artifacts:integrationTest
+    - name: AWS EC2 Provider Integration Tests
+      run: ./gradlew --build-cache :clouddriver-aws:integrationTest
--- a/clouddriver-aws/clouddriver-aws.gradle
+++ b/clouddriver-aws/clouddriver-aws.gradle
@@ -1,3 +1,20 @@
+plugins {
+  id 'com.adarshr.test-logger' version '2.1.0'
+}
+
+sourceSets {
+  integration {
+    java.srcDirs = ["src/integration/java"]
+    resources.srcDirs = ["src/integration/resources"]
+    compileClasspath += main.output + test.output
+  }
+}
+
+configurations {
+  integrationImplementation.extendsFrom testImplementation
+  integrationRuntime.extendsFrom testRuntime
+}
+
 dependencies {
   implementation project(":cats:cats-core")
   implementation project(":clouddriver-api")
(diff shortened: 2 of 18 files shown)


pdk27:
@cfieber @ajordens

german-muzquiz (commented):
Thanks for making the effort to increase spinnaker quality! Overall I only have some comments based on the experience writing kubernetes integration tests, maybe they're of help.

The focus of each test case was to validate some functionality at a level that a user can understand with the least amount of context as possible. For example, looking at `Deploy (Manifest)` stage it is composed of several tasks, where each task may or may not result in a request to clouddriver:
image
I decided that each test case covers _all_ requests to clouddriver done to successfully deploy a manifest, instead of testing a single request. This has the effect that test cases can be designed at a business/acceptance level, making it easy to write them with clear given/when/then sections according to behavior driven development. It also has the effect of making it obvious when a test case is needed: basically every field in the stage configuration can be a separate test case. On the other hand, test cases are longer and may fail at different points.

ajordens (approved):
I don't have a tremendous amount of experience with the existing integration tests for other providers, not against having similar tests for aws. 

Don't see a blocker to starting with these.

ajordens:
Are you happy with the state of these @pdk27 and would you like them merged?

pdk27:
> Are you happy with the state of these @pdk27 and would you like them merged?

Thanks Adam. I do have the permissions to merge now (yay!) 
Once I address German's comments, I will take this forward.

pdk27:
> Thanks for making the effort to increase spinnaker quality! Overall I only have some comments based on the experience writing kubernetes integration tests, maybe they're of help.
> 
> The focus of each test case was to validate some functionality at a level that a user can understand with the least amount of context as possible. For example, looking at `Deploy (Manifest)` stage it is composed of several tasks, where each task may or may not result in a request to clouddriver:
> image
> I decided that each test case covers _all_ requests to clouddriver done to successfully deploy a manifest, instead of testing a single request. This has the effect that test cases can be designed at a business/acceptance level, making it easy to write them with clear given/when/then sections according to behavior driven development. It also has the effect of making it obvious when a test case is needed: basically every field in the stage configuration can be a separate test case. On the other hand, test cases are longer and may fail at different points.

Makes sense. I have little context into pipelines and its stages. I thought until I gain more context in that area / other contributors who are already familiar want to add more tests, AWS provider could use some integ tests for each supported operation. These tests can definitely be improved and made more focused on customer's view point in the future. Thanks for the insight.

german-muzquiz (approved):

PR #5291, opened 2021-03-15
fix(gitrepo): Support for SHA references

german-muzquiz:
Fixes https://github.com/spinnaker/spinnaker/issues/6363

`git clone` only supports branch names and tag names for references. For fetching a specific SHA the following sequence is used:

```
git init
git remote add origin <repo url>
git fetch origin <sha>
git reset --hard FETCH_HEAD
```

Not all git servers support directly fetching a SHA, if the server doesn't support that the following error may be shown:

```
error: Server does not allow request for unadvertised object
```

In case that happens, the code fallbacks to fetching all the repo and then the specific SHA:

```
git fetch origin
git fetch origin <sha>
git reset --hard FETCH_HEAD
```

Additionally, this PR introduces the following changes:

* Git binary upgraded to `2.26.3-r0`. Without this change the docker image cannot be built because of an error like:

```
#5 9.282 ERROR: unsatisfiable constraints:
#5 9.294   git-2.26.3-r0:
#5 9.294     breaks: world[git=2.26.2-r0]
```

* Fixed an issue checking free disk space if the clones home directory didn't exist

A separate PR for backporting to 1.25 will be done.


clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitJobExecutor.java
```diff
@@ -105,71 +107,99 @@ public void cloneOrPull(String repoUrl, String branch, Path localPath, String re
...
     if (!isValidReference(repoUrl)) {
       throw new IllegalArgumentException(
           "Git reference \""
               + repoUrl
               + "\" is invalid for credentials with auth type "
               + authType);
     }
-    FileUtils.deleteDirectory(destination.toFile());
+
+    if (destination.toFile().exists()) {
```
ncknt, line 128: Wasn't it fine before without the `if`? The first step of `FileUtils.deleteDirectory()` is to check if the directory exists. You could use a local variable for `destination.toFile()`.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitJobExecutor.java
```diff
@@ -41,6 +42,7 @@
 public class GitJobExecutor {
 
   private static final String SSH_KEY_PWD_ENV_VAR = "SSH_KEY_PWD";
+  private static final Pattern SHA_PATTERN = Pattern.compile("[0-9a-f]{40}");
```
ncknt, line 45: Should we also support shorter SHA (like the default 7 characters)?

german-muzquiz: I actually couldn't find official docs from jgit on how it supports cloning SHA. The method used before only talks about ref name, branch name or tag name: https://download.eclipse.org/jgit/site/5.10.0.202012080955-r/apidocs/org/eclipse/jgit/api/CloneCommand.html#setBranch-java.lang.String-

Besides `git fetch` only supporting the full SHA, I'll check with the user that reported the issue.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitJobExecutor.java
german-muzquiz: I only looked at the javadoc of that method and found this:

```
Throws IllegalArgumentException – if directory does not exist or is not a directory
```

So I just wanted to be sure. I'll use the local variable for `destination.toFile()`.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitJobExecutor.java
ethanfrogers: +1 on supporting shorter SHAs. As a matter of fact, we have an internal Run Job stage that depends on Clouddriver/Git repo support and short SHAs so preserving this behavior would be great.

german-muzquiz: That's an interesting problem, we cannot use short SHAs in git fetch commands:
```
git fetch origin 9904d5e
fatal: couldn't find remote ref 9904d5e
```
It's possible that the only way to make short SHAs to work is to actually download the entire repo (with full history) and then checkout that SHA.

german-muzquiz: I added support for short SHAs as a fallback mechanism in case regular cloning of branch/tag doesn't work, this is to account for cases where a branch or tag matches the same pattern of a short SHA. The full repo needs to be cloned for this to work though.

ethanfrogers (approved):

german-muzquiz:
Opened backport PR fixing merge conflicts: https://github.com/spinnaker/clouddriver/pull/5306

PR #5292, opened 2021-03-15
fix(aws): Ensure that the cleanup agents are scheduled

ajordens:
Due to (potential) changes in bean ordering, `awsCleanupProvider.getAgentScheduler()`
may not always belonger null.

In these cases, `CleanupAlarmsAgent` and `CleanupDetachedInstancesAgent` would
not be added to the `AwsCleanupProvider`.


ajordens:
This has been discussed offline and rolled out at Netflix

PR #5308, opened 2021-03-22
fix(git): Remove git version constraints in distro package

german-muzquiz:
Removing git pinned version. The original intent of pinning was to avoid regressions in case a newer git version was used that is incompatible with how clouddriver uses it, like by removing flags or options. However, Alpine and Ubuntu make no longer available previous versions when a new one is released (at least for security patches), breaking spinnaker releases:
```
Version '1:2.17.1-1ubuntu0.7' for 'git' was not found
```
And
```
#5 9.282 ERROR: unsatisfiable constraints:
#5 9.294   git-2.26.3-r0:
#5 9.294     breaks: world[git=2.26.2-r0]
```


link108 (approved):

link108:

PR #5312, opened 2021-03-23
feat(cloudfoundry): added metric event listener for outbound requests to cf

zachsmith1:
Added OkHttpMetricsEventListener for outbound requests to cloud foundry. This will give operators some insight across cf foundations and latency. As of right now, we are not mapping the URI's, but this could be requested/developed assuming we avoid tag cardinality explosion due to tagging the URI.


clouddriver-cloudfoundry/clouddriver-cloudfoundry.gradle
```diff
@@ -51,15 +51,14 @@ dependencies {
   implementation "io.spinnaker.kork:kork-credentials"
   implementation "io.spinnaker.kork:kork-annotations"
   implementation "io.spinnaker.kork:kork-moniker"
+  implementation "io.spinnaker.kork:kork-web"
```
german-muzquiz, line 54: Didn't `clouddriver-web` depend on `clouddriver-cloudfoundry`? Usually web module depends on all the others, not the other way around

zachsmith1: yes, clouddriver-web depends on clouddriver-cloudfoundry but not sure about kork-web. Either way, this was actually a mistake that slipped in during testing. We actually need to import micrometer directly because micrometer core isn't in kork. I think it was getting brought indirectly through kork-web because of the spring web dependencies.

german-muzquiz (approved):

PR #5314, opened 2021-03-24
feat(ecs): Override Task Def Artifacts

paragbhingre:
Add a support to override task definitions. 

Backend changes to support [#6374](https://github.com/spinnaker/spinnaker/issues/6374)

Related PRs - 
Deck - [#9092](https://github.com/spinnaker/deck/pull/9092)
ORCA - [#4117](https://github.com/spinnaker/orca/pull/4117)


piradeepk:
@paragbhingre can you add screenshots and details of the different scenarios that you manually tested?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -337,14 +337,17 @@ protected RegisterTaskDefinitionRequest makeTaskDefinitionRequest(
...
+      File artifactFile =
+          downloadTaskDefinitionArtifact(description.getResolvedTaskDefinitionArtifact());
+      try {
+        requestTemplate = mapper.readValue(artifactFile, RegisterTaskDefinitionRequest.class);
+      } catch (IOException e) {
+        throw new UncheckedIOException(e);
+      }
+    } else {
+      requestTemplate = description.getSpelProcessedTaskDefinitionArtifact();
     }
```
piradeepk, line 351: Would you be able to point to a different spot where we have split logic like this for processing spel logic? I'm a bit hesitant in having more forked logic. For instance, what if enable SPEL processing is true but the artifact doesn't actually have any expressions (in that case, the logic btwn true and false would have to function the same). 

Is there a way we can share the logic instead of having forked logic?

paragbhingre: Your concern is valid, but no where in the code we have this situation where in we need to either download server group info from somewhere else or get it from pipeline JSON. In this particular scenario we have that scenario due to which we need to have this forked logic. And it is hard to figure out if SPEL processing is true but the artifact doesn't actually have any expressions.

piradeepk: In either case, could we have the same logic apply across both scenarios, and merge the artifact with the processed artifact (for instance, have orca provide the processed artifact if it uses spel expressions)?

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
```diff
@@ -602,4 +601,121 @@ public void createServerGroup_ArtifactsEC2WithMultipleLBsTest()
...
+            .getContainerDefinitions()
+            .get(0)
+            .getLogConfiguration()
+            .getOptions()
+            .get("awslogs-group"));
+
+    assertEquals(
+        80,
+        seenTaskDefRequest.getContainerDefinitions().stream()
+            .filter(container -> container.getName().equals("application"))
```
allisaurus, line 674: since there's only one container definition in this test, instead of filtering it out of the request each time (and incurring the cost of iterating over the list) we should retrieve it once and then do the assertions on that variable.

clouddriver-ecs/src/integration/resources/testoperations/createServerGroup-spelProcessedArtifact-EC2-targetGroupMappings.json
```diff
@@ -0,0 +1,99 @@
...
+
+    "location" : "us-west-2",
+    "reference" : "refernce",
+    "metadata" : null,
+    "artifactAccount" : "my-github",
+    "provenance" : "prov",
+    "uuid" : "uid-123"
+  },
+  "spelProcessedTaskDefinitionArtifact" : [{
+  "family": "PLACEHOLDER",
```
allisaurus, line 56: please indent the contents of the artifact so it appears as a child of `spelProcessedTaskDefinitionArtifact` (they're on the same vertical line right now)

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -337,14 +337,26 @@ protected RegisterTaskDefinitionRequest makeTaskDefinitionRequest(
...
-    File artifactFile =
-        downloadTaskDefinitionArtifact(description.getResolvedTaskDefinitionArtifact());
-
-    RegisterTaskDefinitionRequest requestTemplate;
-    try {
-      requestTemplate = mapper.readValue(artifactFile, RegisterTaskDefinitionRequest.class);
-    } catch (IOException e) {
-      throw new UncheckedIOException(e);
+    RegisterTaskDefinitionRequest requestTemplate = null;
+    if (!description.isEvaluateTaskDefinitionArtifactExpressions()) {
```
allisaurus, line 341: this would be more readable if we evaluated the `true` condition first and extracted the logic of each block to separate functions. then we could end up with something like:
```java
RegisterTaskDefinitionRequest requestTemplate = null;
if (description.isEvaluateTaskDefinitionArtifactExpressions()) {
    requestTemplate = getSpelProcessedArtifact();
} else {
    requestTemplate = getArtifactFromFile();
}
```

allisaurus (changes requested):
Thanks @paragbhingre ! I've requested a couple changes re: simplification but overall looks good.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/description/CreateServerGroupDescription.java
```diff
@@ -93,7 +93,9 @@
   List<ServiceDiscoveryAssociation> serviceDiscoveryAssociations;
 
   boolean useTaskDefinitionArtifact;
+  boolean evaluateTaskDefinitionArtifactExpressions;
   Artifact resolvedTaskDefinitionArtifact;
+  List<Map<Object, Object>> spelProcessedTaskDefinitionArtifact;
```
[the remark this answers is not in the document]
paragbhingre: Yes

clouddriver-ecs/src/integration/java/com/netflix/spinnaker/clouddriver/ecs/test/CreateServerGroupWithArtifactsSpec.java
paragbhingre: make sense, changed.

clouddriver-ecs/src/integration/resources/testoperations/createServerGroup-spelProcessedArtifact-EC2-targetGroupMappings.json
paragbhingre: oops, fixed.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
paragbhingre: done.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/description/CreateServerGroupDescription.java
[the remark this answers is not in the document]
allisaurus: why is this a list of `Map<Object,Object>`  and not just a map of object:object?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -334,18 +334,42 @@ protected RegisterTaskDefinitionRequest makeTaskDefinitionRequest(
     return request;
   }
 
-  protected RegisterTaskDefinitionRequest makeTaskDefinitionRequestFromArtifact(
-      String ecsServiceRole, EcsServerGroupName newServerGroupName) {
+  private RegisterTaskDefinitionRequest getSpelProcessedArtifact() {
+    RegisterTaskDefinitionRequest requestTemplate = null;
+    for (Map<Object, Object> serializedSpelProcessedTaskDefArtifact :
```
allisaurus, line 339: I'm confused why `serializedSpelProcessedTaskDefArtifact` is a list - i'm not sure it needs to be, but if it does, then as written this func will loop through and convert each item in the collection but only return the last one. If we're doing this for future proofing reasons wouldn't we want to retain the results of each conversion?

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
```diff
@@ -334,18 +334,42 @@ protected RegisterTaskDefinitionRequest makeTaskDefinitionRequest(
...
+        description.getSpelProcessedTaskDefinitionArtifact()) {
+      if (serializedSpelProcessedTaskDefArtifact != null) {
+        requestTemplate =
+            mapper.convertValue(
```
allisaurus, line 343: I think we can just return this line, no need for the separate `requestTemplate` variable

paragbhingre: Make sense. Updated.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperation.java
paragbhingre: So I have changed the type of spelProcessedTaskDefArtifact from List of Maps to Map of Objects which solves this issue. The code has been updated.

allisaurus:
@paragbhingre changes look good, though it appears to be missing [unit tests](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperationSpec.groovy)? We should add ones at least for the happy case and the null processed artifact error case.

paragbhingre:
@allisaurus  I have added the the unit test for Spel processed artifact only as we already do have the unit test that goes to the other side of the forked logic and test the code flow.

clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperationSpec.groovy
```diff
@@ -1853,6 +1853,39 @@ class CreateServerGroupAtomicOperationSpec extends CommonAtomicOperation {
...
 
+  def 'should return valid task def from spelProcessedArtifact'() {
+    given:
+    def credentials = Mock(NetflixAssumeRoleAmazonCredentials) {
+      getName() >> { "test" }
+      getRegions() >> { [new AmazonCredentials.AWSRegion('us-west-1', ['us-west-1a', 'us-west-1b'])] }
+      getAssumeRole() >> { 'arn:aws-us-gov:iam:123123123123:role/test-role' }
+      getAccountId() >> { 'test' }
+    }
+    def resolvedArtifact = createResolvedArtifact()
```
allisaurus, line 1864: what happens if we don't mock this? the resolved artifact is only used when we need to download the file, so running this test w/o would better ensure we're actually using the spelProcessedTaskDefinitionArtifact

clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/deploy/ops/CreateServerGroupAtomicOperationSpec.groovy
```diff
@@ -1853,6 +1853,39 @@ class CreateServerGroupAtomicOperationSpec extends CommonAtomicOperation {
...
+    def containerDef = createContainerDef()
+    def registerTaskDefRequest = createRegisterTaskDefRequest(containerDef)
+    def description = createDescription(resolvedArtifact);
+    description.isEvaluateTaskDefinitionArtifactExpressions() >> true
+    Map<Object, Object> map = new HashMap<>()
```
allisaurus, line 1869: I know it's small/simple, but can we name this something more descriptive ("spelArtifactMap", etc.)

allisaurus: also: maybe we can put a different execution role in here and then assert on it below to make sure the valued from this object is what's being used.

allisaurus (approved):
🔥 🔥 🔥  Approved! Soooooo stoked for this feature @paragbhingre ! Nice work! 

Keeping the 'do not merge' label on for now to ensure the Orca changes get approved/settled, but should be good to go after that (and after squashing your commits :P )

PR #5316, opened 2021-03-25
refactor(cloudfoundry): move processes into their own services

zachsmith1:
This is a refactor to move processes and their operations into process specific classes for code cleanliness and readability. This also maps more closely with the cloud foundry api model. No new logic and no changes to existing logic.


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/api/ApplicationService.java
```diff
@@ -78,19 +77,6 @@
...
-
-  @PATCH("/v3/processes/{guid}")
-  Call<Process> updateProcess(@Path("guid") String guid, @Body UpdateProcess updateProcess);
-
-  @GET("/v3/processes/{guid}")
-  Call<Process> findProcessById(@Path("guid") String guid);
-
-  @GET("/v3/processes/{guid}/stats")
-  Call<ProcessResources> findProcessStatsById(@Path("guid") String guid);
-
```
rvazquezglez, line 93: Is the plan to move other methods to different interfaces (like one interface for routes, one for packages, etc)?

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/HttpCloudFoundryClient.java
```diff
@@ -192,49 +197,4 @@ public void checkServerTrusted(X509Certificate[] x509Certificates, String s) {}
...
-
-  @Override
-  public Tasks getTasks() {
-    return tasks;
-  }
-
-  @Override
-  public Logs getLogs() {
-    return logs;
-  }
```
rvazquezglez, line 239: Nice removal.

rvazquezglez (approved):
LGTM

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/api/ApplicationService.java
zachsmith1: I think as they get big enough it makes sense. Overall I think our nomenclature should be as similar to the cloud foundry api as possible. Another example, I renamed "scaleApplication" to "scaleProcess" since thats actually whats happening (v3/processes/:guid/actions/scale). That could cause confusion for someone looking at it for the first time if they don't have the context

german-muzquiz (approved):

PR #5319, opened 2021-03-27
feat(lambda): additional properties for stream event source mappings

jaredstehler:
Adds additional properties used in stream-based lambda event source mappings (i.e. kinesis)


jaredstehler:
closed as dupe of #5318

jaredstehler:
this was errantly closed

zachsmith1:
@jaredstehler any kind of testing we can do here? Looks simple enough but not sure if any of those additional fields can cause issues if they are explicitly null?

jaredstehler:
I'm not sure there's much we can do in a unit test, since the aws-sdk seems to just pass the values on to the API. That said, looking into the `UpdateEventSourceMappingRequest` and `CreateEventSourceMappingRequest` classes, these fields are all defined as nullable default null.

zachsmith1 (approved):
LGTM

jaredstehler:
@Mergifyio refresh

zachsmith1:
@Mergifyio refresh

jaredstehler:
Do i need to do anything to resolve this mergify failure? I enabled it to access that repo through my account.

zachsmith1:
@mergify refresh

PR #5320, opened 2021-03-29
refactor(clusters): Add typing and convert to java syntax.

Bijnagte:
In the future full java conversion can be done but it will require additional changes so I want to break it up into more commits.


mattgogerly:
@Mergifyio update

mattgogerly (approved):

PR #5323, opened 2021-03-31
chore(cloudfoundry): Extracts configuration variables for retries.

rvazquezglez:
chore(cloudfoundry): Extracts configuration variables for retries.

Related PR: https://github.com/spinnaker/halyard/pull/1862

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/HttpCloudFoundryClient.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/HttpCloudFoundryClient.java
@@ -32,7 +32,6 @@ import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
 import java.security.cert.X509Certificate;
 import java.util.concurrent.ForkJoinPool;
-import java.util.concurrent.TimeUnit;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.TrustManager;
 import javax.net.ssl.X509TrustManager;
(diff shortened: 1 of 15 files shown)


zachsmith1 (changes requested):

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/config/CloudFoundryConfigurationProperties.java
```diff
@@ -75,7 +75,7 @@ public void destroy() {
   }
 
   @Data
-  public class ClientConfig {
+  public static class ClientConfig {
```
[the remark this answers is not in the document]
rvazquezglez: Yes, it's to avoid tying it to an object instance, from Effective Java (third edition), item 24:

> Each instance of a nonstatic member class is implicitly associated with an enclosing instance of its containing class. Within instance methods of a nonstatic member class, you can invoke methods on the enclosing instance or obtain a reference to the enclosing instance using the qualified this construct. If an instance of a nested class can exist in isolation from an instance of its enclosing class, then the nested class must be a static member class: *it is impossible to create an instance of a nonstatic member class without an enclosing instance.*

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/HttpCloudFoundryClient.java
```diff
@@ -106,9 +109,12 @@ public HttpCloudFoundryClient(
     okHttpClient =
         okHttpClient
             .newBuilder()
+            .connectTimeout(clientConfig.getConnectionTimeout(), TimeUnit.MILLISECONDS)
```
zachsmith1, line 112: Looks like we are setting these twice. Once in CloudFoundryProviderConfig for the bean cloudFoundryOkHttpClient and then here as well. I don't think I have a preference as to which one we use

zachsmith1 (changes requested):

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/HttpCloudFoundryClient.java
rvazquezglez: Removed.

german-muzquiz (approved):

PR #5324, opened 2021-04-01
feat(provider/kubernetes): Added remaning infrastructure tests.

danielGz:
Add the infrastructure integration tests to OSS clouddriver pertaining to the following spinnaker objects:

securityGroups

applications

serverGroupManagers


clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/InfrastructureIT.java
```diff
@@ -838,4 +838,484 @@ public void shouldGetRawResources() throws InterruptedException, IOException {
...
+                      + name);
+
+          // ------------------------- then --------------------------
+          System.out.println(resp.asString());
+          if (resp.statusCode() == 404) {
+            return false;
+          }
+          resp.then().statusCode(200);
+          List securityGroupList = resp.jsonPath().getList(ns);
+          return securityGroupList != null && securityGroupList.size() > 0;
```
german-muzquiz, line 1164: Since you're querying for name, this list should return exactly one item right? Otherwise you wouldn't discover bugs where the response doesn't filter properly by name.

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/InfrastructureIT.java
```diff
@@ -838,4 +838,484 @@ public void shouldGetRawResources() throws InterruptedException, IOException {
...
+          System.out.println("> Sending get clusters request");
+          Response resp = get(baseUrl() + "/applications/" + appName + "/serverGroupManagers");
+
+          // ------------------------- then --------------------------
+          System.out.println(resp.asString());
+          if (resp.statusCode() == 404) {
+            return false;
+          }
+          resp.then().statusCode(200);
+          List serverGroupList = resp.jsonPath().getList("[0].serverGroups");
```
german-muzquiz, line 1268: Would this fail with an unwanted error if the returned list is empty (i.e. there's no index `0`)? Probably it's better to check the root list instead of the nested `serverGroups`.

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/InfrastructureIT.java
```diff
@@ -829,4 +831,345 @@ public void shouldGetRawResources() throws InterruptedException, IOException {
...
+        () -> {
+          // ------------------------- when --------------------------
+          System.out.println("> Sending get securityGroups request");
+          Response resp = get(baseUrl() + "/securityGroups");
+
+          // ------------------------- then --------------------------
+          System.out.println(resp.asString());
+
+          resp.then().statusCode(200);
+          List securityGroupList = resp.jsonPath().getList(ACCOUNT1_NAME + ".kubernetes." + ns);
```
german-muzquiz, line 863: Using raw types shows compilation warning like these (there are many):

```
clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/InfrastructureIT.java:863: warning: [rawtypes] found raw type: List
          List securityGroupList = resp.jsonPath().getList(ACCOUNT1_NAME + ".kubernetes." + ns);
```

Let's use `List<Object>` instead of raw types

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/InfrastructureIT.java
```diff
@@ -829,4 +831,345 @@ public void shouldGetRawResources() throws InterruptedException, IOException {
...
+    // ------------------------- given --------------------------
+
+    String appName = "getmanifest";
+    System.out.println("> Using namespace " + ns);
+
+    KubeTestUtils.deployIfMissing(
+        baseUrl(), ACCOUNT1_NAME, ns, "networkPolicy", NETWORK_POLICY_1_NAME, appName, kubeCluster);
+
+    KubeTestUtils.deployIfMissing(
+        baseUrl(), ACCOUNT1_NAME, ns, "networkPolicy", NETWORK_POLICY_2_NAME, appName, kubeCluster);
```
german-muzquiz, line 851: This test case is exactly the same as `shouldListSecurityGroupsByAccount`, so let's test here that `GET /securityGroups` returns data for both accounts instead of the same account.

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/InfrastructureIT.java
```diff
@@ -829,4 +831,345 @@ public void shouldGetRawResources() throws InterruptedException, IOException {
...
+      throws InterruptedException, IOException {
+    // ------------------------- given --------------------------
+
+    String appName = "getmanifest";
+    System.out.println("> Using namespace " + ns);
+
+    KubeTestUtils.deployIfMissing(
+        baseUrl(), ACCOUNT1_NAME, ns, "networkPolicy", NETWORK_POLICY_1_NAME, appName, kubeCluster);
+    KubeTestUtils.deployIfMissing(
+        baseUrl(), ACCOUNT1_NAME, ns, "networkPolicy", NETWORK_POLICY_2_NAME, appName, kubeCluster);
```
german-muzquiz, line 931: Also this test is similar to the other ones for security groups, let's test here with the second deployment done to account2 and `default` namespace, so that `GET /securityGroups/account1?region=<ns>` only returns one item

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/InfrastructureIT.java
```diff
@@ -829,4 +831,345 @@ public void shouldGetRawResources() throws InterruptedException, IOException {
...
+                      + NETWORK_POLICY_1_NAME);
+
+          // ------------------------- then --------------------------
+          System.out.println(resp.asString());
+          if (resp.statusCode() == 404) {
+            return false;
+          }
+          resp.then().statusCode(200);
+          List securityGroupList = resp.jsonPath().getList(ns);
+          return securityGroupList != null && securityGroupList.size() != 0;
```
german-muzquiz, line 1037: If you're requesting security groups by network policy name then the returned list should be exactly one item

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/InfrastructureIT.java
```diff
@@ -829,4 +831,345 @@ public void shouldGetRawResources() throws InterruptedException, IOException {
...
+          Response resp = get(baseUrl() + "/applications/" + appName + "/serverGroupManagers");
+
+          // ------------------------- then --------------------------
+          System.out.println(resp.asString());
+          if (resp.statusCode() == 404) {
+            return false;
+          }
+          resp.then().statusCode(200);
+          List serverGroupList = resp.jsonPath().getList("$");
+          return serverGroupList != null & serverGroupList.size() == 1;
```
german-muzquiz, line 1125: Should be `&&` instead of single `&`

german-muzquiz (approved):

PR #5325, opened 2021-04-06
feat(cloudfoundry): add support for manifest processes

zachsmith1:
This features supports processes which can be defined in cloud foundry manifests. For more information about processes see [here](https://docs.cloudfoundry.org/devguide/multiple-processes.html#about-processes). Buildpacks typically define the processes for an application however a Procfile can also be used to define these processes and their starting commands. See [here](https://docs.cloudfoundry.org/devguide/multiple-processes.html#push-with-multiple-processes) for more information regarding Procfiles and processes.

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/api/ProcessesService.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/api/ProcessesService.java
@@ -16,8 +16,11 @@
 
 package com.netflix.spinnaker.clouddriver.cloudfoundry.client.api;
 
-import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.*;
+import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.Pagination;
 import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.Process;
+import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.ProcessResources;
+import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.ScaleProcess;
+import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.UpdateProcess;
 import okhttp3.ResponseBody;
 import retrofit2.Call;
 import retrofit2.http.*;
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/api/ProcessesService.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/api/ProcessesService.java
@@ -16,15 +16,17 @@
 
 package com.netflix.spinnaker.clouddriver.cloudfoundry.client.api;
 
+import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.*;
 import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.Process;
-import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.ProcessResources;
-import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.ScaleProcess;
-import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.UpdateProcess;
 import okhttp3.ResponseBody;
 import retrofit2.Call;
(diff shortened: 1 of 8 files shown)


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/api/ProcessesService.java
```diff
@@ -16,16 +16,18 @@
 
 package com.netflix.spinnaker.clouddriver.cloudfoundry.client.api;
 
+import com.netflix.spinnaker.clouddriver.cloudfoundry.client.model.v3.*;
```
jorgebee65, line 19: I think we prefer to import just the necessary classes rather than wildcard

jorgebee65 (commented):
Just check the imports on ProcessesService

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/api/ProcessesService.java
zachsmith1: Ya I think my IntelliJ converts them when there are more than 5 from that package. Either way, made the change. Do you know if there is an agreed upon limit?

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/ops/DeployCloudFoundryServerGroupAtomicOperation.java
```diff
@@ -472,28 +473,75 @@ private void scaleApplication(
...
+      for (ProcessRequest req : description.getApplicationAttributes().getProcesses()) {
+        String processGuid =
+            processes.stream()
+                .filter(p -> p.getType().equalsIgnoreCase(req.getType()))
+                .map(p -> p.getGuid())
+                .findFirst()
+                .orElseThrow(
+                    () ->
+                        new CloudFoundryApiException(
+                            "Unable to find a process with type: " + req.getType()));
```
rvazquezglez, line 496: Should this fail for _all_ processes when one of them is not found?

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/ops/DeployCloudFoundryServerGroupAtomicOperation.java
```diff
@@ -472,28 +473,75 @@ private void scaleApplication(
...
             description.getApplicationAttributes().getCommand(),
             description.getApplicationAttributes().getHealthCheckType(),
             description.getApplicationAttributes().getHealthCheckHttpEndpoint());
+
+    if (!description.getApplicationAttributes().getProcesses().isEmpty()) {
+      List<Process> processes = client.getProcesses().getAllProcessesByAppId(serverGroupId);
+
+      for (ProcessRequest req : description.getApplicationAttributes().getProcesses()) {
+        String processGuid =
+            processes.stream()
```
rvazquezglez, line 526: This stream looks very similar to the one from line 489. Is it worth extracting it to a method?

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/ops/DeployCloudFoundryServerGroupAtomicOperation.java
zachsmith1: I think so, yes. In the scenario you mention, if someone is deploying an app to production and it requires two processes to have specific commands, if one of them gets skipped because it wasn't found it could be problematic and they might not even know that it was skipped. In this case, we want to enforce correct manifests and specifically correct process requests.

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/ops/DeployCloudFoundryServerGroupAtomicOperation.java
zachsmith1: Its pretty small and not doing anything logic-wise but I like where your head is at 👍 . Done!

rvazquezglez (approved):
LGTM

german-muzquiz (approved):

PR #5327, opened 2021-04-07
fix(kubernetes): Removed PodPreset references

kirangodishala:
Issue#[6398](https://github.com/spinnaker/spinnaker/issues/6398)

This PR removes all the references of PodPreset kind as it is not longer supported by Kubernetes starting from v1.20.


dbyron0:
What about people running versions of kubernetes before 1.20?  If they're using spinnaker to deploy PodPresets, this is pretty painful.  It seems like a small enough amount of code that it's not so bad to leave it around, no?

german-muzquiz (changes requested):
Let's not remove podPreset code because this will break pipelines for clusters that use it today.

From the sig meeting discussion it appeared the problem was that clouddriver tried to access this kind in clusters that didn't have it, so it was needed to be blacklisted in `kinds/omitKinds`. Can you explain in more detail the exact problem? Until there's a better way to discover available kinds in the cluster we would need to support all of them for all versions of kubernetes we want to support.

jaredstehler:
FWIW i see errors in caching agent with 1.20:
```
2021-06-02 08:51:52.910  WARN 1 --- [tionAction-3060] c.n.s.c.cache.LoggingInstrumentation     : kubernetes:spinnaker-ops/KubernetesCoreCachingAgent[1/1] completed with one or more failures in  13.349s

com.netflix.spinnaker.clouddriver.kubernetes.op.job.KubectlJobExecutor$KubectlException: Failed to read [configMap, controllerRevision, cronJob, daemonSet, deployment, event, horizontalpodautoscaler, ingress, job, limitRange, networkPolicy, persistentVolumeClaim, pod, podPreset, podDisruptionBudget, replicaSet, role, roleBinding, secret, service, serviceAccount, statefulSet] from monitoring: error: the server doesn't have a resource type "podPreset"

        at com.netflix.spinnaker.clouddriver.kubernetes.op.job.KubectlJobExecutor.list(KubectlJobExecutor.java:411) ~[clouddriver-kubernetes.jar:na]
```

I wasn't aware of the workaround, so will be putting that in place.

kirangodishala:
> FWIW i see errors in caching agent with 1.20:
> 
> ```
> 2021-06-02 08:51:52.910  WARN 1 --- [tionAction-3060] c.n.s.c.cache.LoggingInstrumentation     : kubernetes:spinnaker-ops/KubernetesCoreCachingAgent[1/1] completed with one or more failures in  13.349s
> 
> com.netflix.spinnaker.clouddriver.kubernetes.op.job.KubectlJobExecutor$KubectlException: Failed to read [configMap, controllerRevision, cronJob, daemonSet, deployment, event, horizontalpodautoscaler, ingress, job, limitRange, networkPolicy, persistentVolumeClaim, pod, podPreset, podDisruptionBudget, replicaSet, role, roleBinding, secret, service, serviceAccount, statefulSet] from monitoring: error: the server doesn't have a resource type "podPreset"
> 
>         at com.netflix.spinnaker.clouddriver.kubernetes.op.job.KubectlJobExecutor.list(KubectlJobExecutor.java:411) ~[clouddriver-kubernetes.jar:na]
> ```
> 
> I wasn't aware of the workaround, so will be putting that in place.

Keeping podPreset under omitKinds in the hal config should fix this issue.

jaredstehler:
it did indeed, thanks!

PR #5328, opened 2021-04-08
fix(kube/test): Reducing execution time of integration tests

german-muzquiz:
Kubernetes integration tests now share namespaces to speed execution time. Test cases use separate spinnaker application names to keep responses isolated from other test cases, which simplifies debugging a failed test as if a separate namespace was used.


clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/utils/KubeTestUtils.java
```diff
@@ -355,4 +348,54 @@ public static void deployAndWaitStable(
...
+                + "\")]"
+                + path
+                + "}'");
+    if (!output.isEmpty()) {
+      if (image == null || output.contains(image)) {
+        return;
+      }
+    }
+
+    deployAndWaitStable(baseUrl, account, namespace, kind, name, app, image);
```
danielGz, line 399: Do you think we should also support ```
deployAndWaitStable(
      String baseUrl, List<Map<String, Object>> reqBody, String targetNs, String... objectNames)
``` signature method, or to have a separate one that does? Since it is much more flexible.

german-muzquiz: We can add it when we need it, today it wouldn't be used

danielGz (approved):

PR #5329, opened 2021-04-08
test(aws): Adding integration tests for mixed instances features

pdk27:
- Adding integration tests for recently added support for multiple instance types
 - simplifying some tests


cfieber (approved):
thanks!

PR #5331, opened 2021-04-12
feat(aws/enclave): Enable Nitro Enclave option

iandelahorne:
[AWS Nitro Enclaves](https://docs.aws.amazon.com/enclaves/latest/user/nitro-enclave.html) is a feature that allows for isolated execution environments, or _enclaves_, to be launched alongside of normal instances.

This PR adds a property, `enableEnclave`, to the deploy description. Restrict
this feature to launch templates since it can only be enabled via
Launch Templates

Fixes spinnaker/spinnaker#6396

## Verification

I created a clouddriver container off this branch by first compiling the JAR:
```
./gradlew clouddriver-web:installDist -x test
```
and built a container with `Dockerfile.ubuntu` and pushed it to dockerhub. I then launched a [Minnaker](https://github.com/armory/minnaker) with OSS spinnaker, and applied the following patchfile:

```
apiVersion: spinnaker.io/v1alpha2
kind: SpinnakerService
metadata:
  name: spinnaker
spec:
  spinnakerConfig:
    service-settings:
      clouddriver:
        artifactId: lflux/clouddriver:nitro-ubuntu
    profiles:
      clouddriver:
        aws:
          features:
            launch-templates:
              enabled: true
              all-applications:
                enabled: true
```

I then created a pipeline which launched an Amazon Linux AMI with the option ` "enableEnclave": true` option set in it.

Launching the instance, I can see that the Enclave option is set

```
$ aws ec2 describe-launch-template-versions --launch-template-id lt-xxxx --query LaunchTemplateVersions[0].LaunchTemplateData.EnclaveOptions
{
    "Enabled": true
}
```
 and on the instance I can see that the device is available:

```
$ ls -la /dev/nitro_enclaves
crw-rw---- 1 root ne 10, 57 Apr 12 04:21 /dev/nitro_enclaves
```

Modifying the pipeline to remove `"enableEnclave": true" line, the AMI deploys with a launch template with the enclave option disabled:

```
$ aws ec2 describe-launch-template-versions --launch-template-id lt-yyy --query LaunchTemplateVersions[0].LaunchTemplateData.EnclaveOptions
{
    "Enabled": false
}
```

and the instance does not have the device:

```
$ ls /dev/nitro_enclaves
ls: cannot access /dev/nitro_enclaves: No such file or directory
```

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
@@ -137,7 +137,7 @@ class BasicAmazonDeployDescription extends AbstractAmazonCredentialsDescription 
    * This is a Launch Template only feature
    * * https://docs.aws.amazon.com/enclaves/latest/user/nitro-enclave.html
    */
-  Boolean enableEnclave = false
+  Boolean enableEnclave
 
   /**
    * Launch template placement details, see {@link com.amazonaws.services.ec2.model.LaunchTemplatePlacementRequest}.
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
@@ -133,6 +133,13 @@ class BasicAmazonDeployDescription extends AbstractAmazonCredentialsDescription 
   Boolean unlimitedCpuCredits
 
   /**
+   * When set to true, the created server group will be configured with Nitro Enclaves enabled
+   * This is a Launch Template only feature
+   * * https://docs.aws.amazon.com/enclaves/latest/user/nitro-enclave.html
+   */
+  Boolean enableEnclave = false
+
+  /**
    * Launch template placement details, see {@link com.amazonaws.services.ec2.model.LaunchTemplatePlacementRequest}.
    */
   LaunchTemplatePlacement placement
@@ -195,7 +202,7 @@ class BasicAmazonDeployDescription extends AbstractAmazonCredentialsDescription 
     return ["requireIMDSv2", "associateIPv6Address", "unlimitedCpuCredits",
             "placement", "licenseSpecifications", "onDemandAllocationStrategy",
             "onDemandBaseCapacity", "onDemandPercentageAboveBaseCapacity", "spotAllocationStrategy",
-            "spotInstancePools", "launchTemplateOverridesForInstanceType"].toSet()
+            "spotInstancePools", "launchTemplateOverridesForInstanceType", "enableEnclave"].toSet()
   }
 
   static Set<String> getMixedInstancesPolicyFieldNames() {
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/CopyLastAsgAtomicOperation.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/CopyLastAsgAtomicOperation.groovy
@@ -169,8 +169,9 @@ class CopyLastAsgAtomicOperation implements AtomicOperation<DeploymentResult> {
           iamInstanceProfile = launchTemplateData.iamInstanceProfile?.name
           instanceMonitoring = launchTemplateData.monitoring?.enabled
           spotMaxPrice = launchTemplateData.instanceMarketOptions?.spotOptions?.maxPrice
+          newDescription.enableEnclave = launchTemplateData.enclaveOptions != null ? launchTemplateData.enclaveOptions.getEnabled() : false
           newDescription.requireIMDSv2 = description.requireIMDSv2 != null ? description.requireIMDSv2 : launchTemplateData.metadataOptions?.httpTokens == "required"
-          newDescription.associateIPv6Address = description.associateIPv6Address 
+          newDescription.associateIPv6Address = description.associateIPv6Address
           if (!launchTemplateData.networkInterfaces?.empty && launchTemplateData.networkInterfaces*.associatePublicIpAddress?.any()) {
             associatePublicIpAddress = true
           }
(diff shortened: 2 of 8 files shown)


zachsmith1 (approved):
LGTM pending an additional AWS approval (if possible)

zachsmith1:
Note: as of right now enabling enclaves is a pretty niche feature. However if in the future more spinnaker users would like to take advantage of this with a flag or similar in the UI, we can potentially add those to deck as well.

ajordens (approved):

jonsie (approved):
FYI @marctrem

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/services/LaunchTemplateService.java
```diff
@@ -247,6 +247,10 @@ private RequestLaunchTemplateData buildLaunchTemplateData(
     networkInterfaceRequest.setDeviceIndex(0);
     networkInterfaceRequest.setGroups(description.getSecurityGroups());
 
+    // Nitro Enclave options
+    request.setEnclaveOptions(
```
pdk27, line 251: Instead of setting this property with every create launch template request, it might make sense to set enclave options only when specified (i.e. the default in the description will be null)?
Thinking of (current / future) features that might be incompatible with it. One such feature is `hibernationOptions` which is not currently supported in Spinnaker, but just throwing it out there for forward compatibility :)

pdk27 (changes requested):
Except the minor change I requested, everything else looks good.

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/services/LaunchTemplateService.java
pdk27: @iandelahorne `associateIPv6Address` and  `unlimitedCpuCredits` are two launch template only fields that have a null default. Note that there might some extra checks for `unlimitedCpuCredits` in other files, which will not be necessary for `enableEnclave`.

For the change I requested, wrapping the `request.setEnclaveOptions` in a null check like on L#235 or L#353 should do.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
```diff
@@ -132,6 +132,13 @@ class BasicAmazonDeployDescription extends AbstractAmazonCredentialsDescription
    */
   Boolean unlimitedCpuCredits
 
+  /**
+   * When set to true, the created server group will be configured with Nitro Enclaves enabled
+   * This is a Launch Template only feature
+   * * https://docs.aws.amazon.com/enclaves/latest/user/nitro-enclave.html
+   */
+  Boolean enableEnclave = false
```
pdk27, line 140: @iandelahorne remove the default here in order to set enclave options only when specified in request.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/CopyLastAsgAtomicOperation.groovy
```diff
@@ -169,8 +169,9 @@ class CopyLastAsgAtomicOperation implements AtomicOperation<DeploymentResult> {
           iamInstanceProfile = launchTemplateData.iamInstanceProfile?.name
           instanceMonitoring = launchTemplateData.monitoring?.enabled
           spotMaxPrice = launchTemplateData.instanceMarketOptions?.spotOptions?.maxPrice
+          newDescription.enableEnclave = launchTemplateData.enclaveOptions != null ? launchTemplateData.enclaveOptions.getEnabled() : false
```
pdk27, line 172: @iandelahorne 
Let's modify this line so that the clone request description takes precedence over source ASG's config.

```
newDescription.enableEnclave = description.enableEnclave != null ? description.enableEnclave :  launchTemplateData.enclaveOptions?.getEnabled()
```

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/services/LaunchTemplateService.java
```diff
@@ -247,6 +247,13 @@ private RequestLaunchTemplateData buildLaunchTemplateData(
     networkInterfaceRequest.setDeviceIndex(0);
     networkInterfaceRequest.setGroups(description.getSecurityGroups());
 
+    // Nitro Enclave options
+    if (asgConfiguration.getEnableEnclave() != null) {
```
pdk27, line 251: this should be `description.getEnableEnclave()` instead of `asgConfiguration.getEnableEnclave()` since the launch template data being built here is for modifyLaunchTemplate operation. 
(FYI, this method shouldn't have a `AsgConfiguration` object - I will fix that in an upcoming PR)

And you also need to add this new field in [ModifyServerGroupLaunchTemplateDescription](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/description/ModifyServerGroupLaunchTemplateDescription.java#L20).

pdk27 (changes requested):

iandelahorne:
@pdk27 Thanks for the suggestions - I've pushed up a commit that hopefully addresses them!

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/services/LaunchTemplateService.java
```diff
@@ -247,6 +247,13 @@ private RequestLaunchTemplateData buildLaunchTemplateData(
     networkInterfaceRequest.setDeviceIndex(0);
     networkInterfaceRequest.setGroups(description.getSecurityGroups());
 
+    // Nitro Enclave options
+    if (description.getEnableEnclave() != null) {
+      request.setEnclaveOptions(
+          new LaunchTemplateEnclaveOptionsRequest()
+              .withEnabled(asgConfiguration.getEnableEnclave()));
```
pdk27, line 254: `description.getEnableEnclave()` here too.

pdk27: Everything else looks good. Thank you!

iandelahorne:
@pdk27 sorry for the long turnaround, it's ready for review now!

pdk27:
> @pdk27 sorry for the long turnaround, it's ready for review now!

LGTM. Thanks for the changes.

pdk27 (approved):

PR #5332, opened 2021-04-13
fix(cats/sql): Zombie agents cleanup

titirigaiulian:
Add cleanup mechanism for zombie caching agents. 
Add `sql.agent.zombie-threshold-ms` with a default of 1h -- after 1 hour consider the agent is zombie and remove it from activeAgents
 
Issue: https://github.com/spinnaker/spinnaker/issues/6405

### DIFF
--- a/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
+++ b/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
@@ -153,6 +153,7 @@ class SqlClusteredAgentScheduler(
   }
 
   private fun findCandidateAgentLocks(): Map<String, AgentExecutionAction> {
+    cleanupZombieAgents()
     val skip = HashMap(activeAgents).entries
     val maxConcurrentAgents = dynamicConfigService.getConfig(Int::class.java, "sql.agent.max-concurrent-agents", 100)
     val availableAgents = maxConcurrentAgents - skip.size
@@ -231,6 +232,16 @@ class SqlClusteredAgentScheduler(
       }
 
     return trimmedCandidates
+  }
+
+  private fun cleanupZombieAgents() {
+    val zombieAgentThreshold = dynamicConfigService.getConfig(Long::class.java, "sql.agent.zombie-threshold-ms", 3600000)
+    activeAgents
+      .filter { it.value.currentTime < System.currentTimeMillis() - zombieAgentThreshold }
+      .forEach {
+        log.warn("Found zombie agent {}, removing it", it.key)
+        activeAgents.remove(it.key, it.value)
+    }
   }
 
   private fun tryAcquireSingle(agentType: String, now: Long, timeout: Long): Boolean {


titirigaiulian:
@german-muzquiz @ncknt can you take a look?

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
```diff
@@ -233,6 +234,16 @@ class SqlClusteredAgentScheduler(
     return trimmedCandidates
   }
 
+  private fun cleanupZombieAgents() {
+    val zombieAgentThreshold = dynamicConfigService.getConfig(Long::class.java, "sql.agent.zombie-threshold-ms", 3600000)
+    activeAgents
+      .filter { it.value.currentTime < System.currentTimeMillis() - zombieAgentThreshold }
+      .forEach {
+        log.warn("Found zombie agent {}, removing it", it.key)
+        activeAgents.remove(it.key, it.value)
```
german-muzquiz, line 243: Instead of `activeAgents.remove` let's call `agentCompleted` to make sure the database lock is also released.

german-muzquiz (commented):
Thank you for discovering such an obscure issue! I can think of at least one situation where caching can be doomed and this PR will fix it:
Having clouddriver replicas 1 and 2 and number 1 starts a batch of caching agents, it will lock those agents in db so that replica 2 is not able to run them. If during that run replica 1 is killed/restarted, the lock will never be released and no replica will run those agents again until the db lock is released.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
titirigaiulian: @german-muzquiz, thanks for the detailed review. If I get it right, I see there is already a mechanism for cleaning expired locks from the db [here](https://github.com/spinnaker/clouddriver/blob/12abd4d778d22829707137673717466e5334511a/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt#L185). The problem was that the local list is not cleanup never.

titirigaiulian:
I think this could be even a bigger issue combined with the [sharding logic](https://github.com/spinnaker/clouddriver/pull/5295) -- if `agent1` will become a zombie on the replica in charge will impact all the deployments for that acc.  As a side note, I think would be great to have also a replication factor like approach, e.g. `agent1` scheduled on at least 2 replicas.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
german-muzquiz: Ah you're right, it's already cleared when trying to acquire other locks

german-muzquiz (approved):

PR #5337, opened 2021-04-20
fix(aws): Fixes name resolver when stack is empty.

rvazquezglez:
fix(aws): Fixes name resolver when stack is empty.

With the current name comparison, if some names (like stack) are empty it would compare them to a `null` value, not adding them to the currently taken sequences. 

The functionality was working on 1.22 and the bug was introduced on https://github.com/spinnaker/clouddriver/commit/fc08d9a75c0498a5e56ac8e9851c8125d8dca7af.


rvazquezglez:

zachsmith1 (approved):
LGTM

zachsmith1:
cc: @allisaurus

german-muzquiz (approved):

allisaurus:
@german-muzquiz @rvazquezglez @zachsmith1  It looks like this has already been merged, but historically we annotate ECS changes with "ecs" (not "aws") in the semantic commit prefix so it's appropriately identified in release changelogs, since they're essentially separate providers w/r/t deployment primatives and logic.

Otherwise no concerns since the existing tests pass and new test cases have been added to cover the bug 👍

link108:
@mergifyio backport release-1.24.x release-1.25.x

PR #5341, opened 2021-04-28
fix(gitrepo): Download artifact content instead of tar.gz

st8f:
Hi! I'm faced with YAML parser error in orca while trying to use gitrepo artifact for Deploy stage (k8s provider). After some research I found that orca is trying to download artifact's yaml from clouddriver and expect to get plain text yaml. But clouddriver pack that yaml into tar.gz instead.

All other ArtifactCredentials implementations seems to return plain text as expected, but not GitRepoArtifactCredentials. So I removed zip code from there. It fixed errors on our spinnaker installation.

### DIFF
--- a/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
+++ b/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactCredentials.java
@@ -64,36 +64,29 @@ public class GitRepoArtifactCredentials implements ArtifactCredentials {
     String branch = artifactVersion(artifact);
     Path stagingPath = gitRepoFileSystem.getLocalClonePath(repoUrl, branch);
     String repoBasename = getRepoBasename(repoUrl);
-    Path outputFile = Paths.get(stagingPath.toString(), repoBasename + ".tgz");
 
     if (!gitRepoFileSystem.canRetainClone()) {
       // delete clone before returning
       try (Closeable ignored = () -> FileUtils.deleteDirectory(stagingPath.toFile())) {
-        return getInputStream(repoUrl, subPath, branch, stagingPath, repoBasename, outputFile);
+        return getInputStream(repoUrl, subPath, branch, stagingPath, repoBasename);
       }
     } else {
       // clones are deleted by gitRepoFileSystem depending on retention period
-      return getInputStream(repoUrl, subPath, branch, stagingPath, repoBasename, outputFile);
+      return getInputStream(repoUrl, subPath, branch, stagingPath, repoBasename);
     }
   }
 
   @NotNull
   private FileInputStream getInputStream(
-      String repoUrl,
-      String subPath,
-      String branch,
-      Path stagingPath,
-      String repoBasename,
-      Path outputFile)
+      String repoUrl, String subPath, String branch, Path stagingPath, String repoBasename)
       throws IOException {
     try {
       if (gitRepoFileSystem.tryTimedLock(repoUrl, branch)) {
         try {
           executor.cloneOrPull(repoUrl, branch, stagingPath, repoBasename);
-          log.info("Creating archive for git/repo {}", repoUrl);
-          executor.archive(
-              Paths.get(stagingPath.toString(), repoBasename), branch, subPath, outputFile);
-          return new FileInputStream(outputFile.toFile());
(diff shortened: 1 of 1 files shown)


dbyron-sf:
Pretty sure this will break bake manifest stages, as [this code](https://github.com/spinnaker/rosco/blob/master/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestEnvironment.java#L80) expects a tarball.

st8f:
> Pretty sure this will break bake manifest stages, as [this code](https://github.com/spinnaker/rosco/blob/master/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestEnvironment.java#L80) expects a tarball.

So, I guess it will be better to haldle tar on orca's side?

dbyron-sf:
There was a related conversation [here](https://github.com/spinnaker/spinnaker/issues/6142#issuecomment-734999533).  I'd love to know why you'd like to use a git/repo artifact with deploy manifest, as opposed to an artifact type that correponds with a single file.

st8f:
> I'd love to know why you'd like to use a git/repo artifact with deploy manifest, as opposed to an artifact type that correponds with a single file.

Well, we wanted to use k8s manifests from private bitbucket installation and gitrepo looks like a solution. And, like already said in conversation you mentioned, it's really not obvious that gitrepo is not supported for Deploy stage.

Is it still possible to support (from an architectural point of view)? Or maybe there are some well known workarounds to use manifest from git for Deploy stage?

dbyron-sf:
> Is it still possible to support (from an architectural point of view)?

It seems possible, but complicated...as now each place that uses the git/repo artifact needs to specify a path for a particular file.  How feasible is it to improve the support for bitbucket artifacts?

st8f:
Is private bitbucket artifact is supported for this stage? As I can see from doc's there only basic auth is suported, which is ok. Access token support will be better of course. I will try to use it with basic auth.

st8f:
> I'd love to know why you'd like to use a git/repo artifact with deploy manifest, as opposed to an artifact type that correponds with a single file.

It turns out that git/repo is [the only option](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/kustomize/BakeKustomizeConfigForm.tsx#L32) for Bake/Kustomize stage (which i'm trying to use).

Also [documentation here](https://spinnaker.io/guides/user/kubernetes-v2/kustomize-manifests/) demonstrate that git/repo **can** be used for this stage. But in fact it doesn't work.

Probably git/repo support still needs be fixed somehow? Or maybe more artifact types can be enabled for Bake/Kustomize stage? Bitbucket for example.

mattgogerly:
I'm a little confused by this issue - we have an almost identical use case (private Bitbucket + Kustomize bake) and it's working fine for us.

As mentioned bake only supports git/repo, so you'll have to add a git/repo account and a bitbucket account (can use the same credentials).

Then use the git/repo account for bake (and deploy the artifact generated), and bitbucket account for pulling individual files in deploy stages.

jasonmcintosh:
IF you're wanting the artifact content, not a rendered manifest, you're looking for a different artifact type (aka http or similar).  git/repo is for baking helm/kustomize files where there are multiple files to handle concurrently.  SO it'll download them and tar.gz them for transport to rosco which does the bakes.

st8f:
> I'm a little confused by this issue - we have an almost identical use case (private Bitbucket + Kustomize bake) and it's working fine for us.

@mattgogerly 
That's interesting. Maybe you're using different spinnaker version (can you tell which is it?). Because threre clearly was a [problem](https://github.com/spinnaker/spinnaker/issues/6142) with this combination (git/repo + kustomize bake).

For now we successfully using bitbucket account + kustomize bake but it required some fix [here](https://github.com/spinnaker/deck/blob/a3f61c63e55527b7ee0631756ecd73121d37d47b/packages/core/src/pipeline/config/stages/bakeManifest/kustomize/BakeKustomizeConfigForm.tsx#L31) since git/repo was the only option allowed.

Anyway this PR is incorrect, closing it.

jasonmcintosh:
Same stuff..  
https://github.com/spinnaker/spinnaker/issues/6142#issuecomment-734999533

Git/repo type artifacts are ONLY for "baking" - you can't use them for "Deploys" DIRECTLY.  git/file is what you're looking at
https://spinnaker.io/docs/reference/artifacts-legacy/types/github-file/

Each artifact type has slightly different handling/use case.

PR #5343, opened 2021-04-28
feat(kubernetes): Opt-out of artifact binding

german-muzquiz:
In "Deploy (Manifest)" stage there's a new configuration parameter called "enableArtifactBinding" defaulting to "true", which controls whether binding of required or optional artifacts should be done for the manifests being deployed. Default behavior is backwards compatible with existing pipelines (automatic binding enabled), and the parameter allows for users to opt-out of automatic binding. Automatic binding of versioned objects (ConfigMap, Secrets) that are deployed as part of a single multidoc yaml document is still done regardless of the config parameter.

Closes https://github.com/spinnaker/spinnaker/issues/6327 and will close https://github.com/spinnaker/spinnaker/issues/6388 once changes in deck are implemented.


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/manifest/KubernetesDeployManifestOperation.java
```diff
@@ -218,6 +158,85 @@ public OperationResult operate(List<OperationResult> _unused) {
     return result;
   }
 
+  private List<KubernetesManifest> bindArtifacts(
```
ncknt, line 161: It looks like the order of operations has changed from bind then version to version then bind (in `bindArtifacts`). At the same time we still try to version (line 102). Was it done on purpose? Looking up the version is a costly operation.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/manifest/KubernetesDeployManifestOperation.java
```diff
@@ -218,6 +158,85 @@ public OperationResult operate(List<OperationResult> _unused) {
...
+      List<KubernetesManifest> inputManifests, Set<Artifact> boundArtifacts) {
+    List<Artifact> requiredArtifacts = description.getRequiredArtifacts();
+    if (requiredArtifacts == null) {
+      requiredArtifacts = new ArrayList<>();
```
ncknt, line 165: Could you move this further down where it is actually used? We could also avoid creating an empty list if all we do is iterate over elements. Same for `optionalArtifacts`

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/manifest/KubernetesDeployManifestOperation.java
```diff
@@ -218,6 +158,85 @@ public OperationResult operate(List<OperationResult> _unused) {
...
+    }
+
+    List<Artifact> optionalArtifacts = description.getOptionalArtifacts();
+    if (optionalArtifacts == null) {
+      optionalArtifacts = new ArrayList<>();
+    }
+
+    List<Artifact> manifestArtifacts = new ArrayList<>();
+    for (KubernetesManifest manifest : inputManifests) {
```
ncknt, line 174: Per comment above, does it really belong here since we already version in `operate`? How does it work for a manifest that can be versioned and can have bound artifacts (`Job` perhaps?)? There are many lists and map so it's a little hard to follow.

german-muzquiz: Today artifact binding works like this:
* First, optional artifacts are binded to manifests
* Second, required artifacts are binded to manifests, overwriting optional artifacts
* Third, main loop for actually deploying the manifests is done. During this loop we bind versioned objects deployed in the same stage (like for a multidoc yaml having `Deployment` + `ConfigMap`)

Binding versioned manifests during deployment loop produces the issue https://github.com/spinnaker/spinnaker/issues/6327 , because once optional/required artifacts are binded, then versioned objects cannot be bind because the match coordinates were changed. This means that if you're deploying a `Deployment+ConfigMap` but already have another `ConfigMap` in the pipeline context, spinnaker binds the one from the context.

Here I try to make it more explicit to follow this binding sequence:
1. Versioned objects deployed in the same stage take the highest precedence
2. Required artifacts are bound if target is not already bounded
3. Optional artifacts are bound if target is not already bounded

Thanks for letting me know that we're getting the version twice and this results in two live calls. I need the version during initial binding because that's part of the artifact information, so I'll make changes to not duplicate this work.

ncknt (approved):
LGTM!

PR #5345, opened 2021-04-30
fix(git/repo): Fixed concurrency issues

german-muzquiz:
* Deletion of clones should also be protected by the lock, so that no threads try to access the clone while it's being deleted
* Synchronize creation of lock instances to avoid having multiple locks for the same coordinates
* Don't delete lock instances after unlocking, to prevent threads waiting on a deleted lock to timeout
* Added debug logging


german-muzquiz:
Backport should be to `1.26` and `1.25` versions.

ethanfrogers (approved):

link108 (approved):

link108:

link108:

PR #5350, opened 2021-05-05
Fix convert block devices

jorgebee65:
Adding a method to convert com.amazonaws.services.ec2.model.BlockDeviceMapping into AmazonBlockDevice and use it in **BasicAmazonDeployHandler** class to avoid the error:
```
class com.amazonaws.services.ec2.model.BlockDeviceMapping cannot be cast to class com.amazonaws.services.autoscaling.model.BlockDeviceMapping (com.amazonaws.services.ec2.model.BlockDeviceMapping and com.amazonaws.services.autoscaling.model.BlockDeviceMapping are in unnamed module of loader 'app')
```


rvazquezglez (approved):

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/asg/AsgConfigHelper.java
```diff
@@ -332,6 +332,35 @@ private static String createDefaultSuffix() {
         .collect(Collectors.toList());
   }
 
+  protected static List<AmazonBlockDevice> convertBlockDevices(
+      List<com.amazonaws.services.ec2.model.BlockDeviceMapping> blockDeviceMappings) {
+    blockDeviceMappings.forEach(b -> log.info(b.getDeviceName()));
```
ajordens, line 337: Likely don't need this logging anymore.

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/asg/AsgConfigHelper.java
```diff
@@ -332,6 +332,35 @@ private static String createDefaultSuffix() {
         .collect(Collectors.toList());
   }
 
+  protected static List<AmazonBlockDevice> convertBlockDevices(
```
ajordens, line 335: Javadoc would be helpful here (we have it for the other variant).

ajordens (commented):
Small nits but otherwise LGTM

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
```diff
@@ -241,7 +241,7 @@ class BasicAmazonDeployHandler implements DeployHandler<BasicAmazonDeployDescrip
       }
 
       if (description.useAmiBlockDeviceMappings) {
```
zachsmith1, line 243: Looks like we are setting description.blockDevices in two areas. Once here, and once on line 226. Does it make sense to have one method or area that the logic to set blockDevices can be constrained? It looks like there could be a scenario where a user could have blockDevices in the description and also have useAmiBlockDevice selected. This could be problematic. If it should only be 1 or the other, maybe we throw an error?

zachsmith1 (changes requested):

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/asg/AsgConfigHelper.java
```diff
@@ -332,6 +332,35 @@ private static String createDefaultSuffix() {
...
+        .map(
+            bdm -> {
+              AmazonBlockDevice amzBd =
+                  new AmazonBlockDevice.Builder()
+                      .deviceName(bdm.getDeviceName())
+                      .virtualName(bdm.getVirtualName())
+                      .build();
+
+              if (bdm.getEbs() != null) {
+                final EbsBlockDevice ebs = bdm.getEbs();
```
zachsmith1, line 348: Looks like AmazonBlockDevice has a new field called kmsKeyId. We may need to do a null check/set setKmsKeyId here? EbsBlockDevice has getKmsKeyId(), however Ebs does not so it would be just needed for this method

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/asg/AsgConfigHelper.java
jorgebee65: removed

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/asg/AsgConfigHelper.java
jorgebee65: Added Javadoc

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/asg/AsgConfigHelper.java
jorgebee65: Added

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy
jorgebee65: For now I just added a logic to assign the **blockDevices** if the **useAmiBlockDevice** is false then check if the **blockDevices** is null and then assign the default based on the instance type. 

and actually is there other method called **copySourceAttributes** where the same attribute is been assigned:
```
if (!description.copySourceCustomBlockDeviceMappings) {
      return description
    } else {
      description.blockDevices = buildBlockDeviceMappingsFromSourceAsg(sourceRegionScopedProvider, sourceAsg, description)
    }
```
is assigned in different places, probably should be convenient to centralize them in other PR.. :information_desk_person:

zachsmith1: sounds good!

zachsmith1 (approved):
LGTM

brantburnett:
This is affecting us and we're rolling back to 1.25 as a result, would love a backport to 1.26 so we can upgrade again. Thanks.

zachsmith1:
@brantburnett this should be backported this week cc: @link108

link108:

PR #5352, opened 2021-05-05
fix(kubernetes): Restore support for extensions/v1beta1 Ingress

kskewes:
Closes https://github.com/spinnaker/spinnaker/issues/6420

Largely a copy, paste, edit of the existing `networking.k8s.io/v1*` apiVersion handling.

Tested with `clouddriver-caching` no longer throwing exception:
```
com.netflix.spinnaker.clouddriver.kubernetes.op.handler.UnsupportedVersionException: No ingress is supported at api version extensions/v1beta1   
```

I'm not sure why the Groovy test for Ingress `extensions/v1beta1` is not working.


german-muzquiz (approved):
Thanks for fixing this!

PR #5354, opened 2021-05-06
feat(validators): add extensible description validator

KathrynLewis:
This PR aims to allows users to add custom description validators in their plugins and extensions.

We're adding the `GlobalDescriptionValidator` interface to the `clouddriver-api` package to make sure plugin creators can trust the method signature and functionality to be consistent.

`OperationsService` calls out to `AnnotationsBasedAtomicOperationsRegistry` to fetch the validator for each atomic operation description. We're hooking into this process by returning a `CompositeDescriptionValidator` instead of the single `DescriptionValidator`. The `CompositeDescriptionValidator` will call each validator from the autowired list of `GlobalDescriptionValidator`s as well as the standard validator for that atomic operation.


clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/deploy/ExtensibleDescriptionValidator.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.deploy;
+
+import com.netflix.spinnaker.kork.annotations.Beta;
+import java.util.List;
+
+@Beta
+public interface ExtensibleDescriptionValidator {
```
danielpeach, line 23: I would make this extend `SpinnakerExtensionPoint`.

danielpeach: I would consider changing the name. Most description validators work on a single description type. This kind of description validator works on all description types. You could call it `GlobalDescriptionValidator`. 

You could also consider whether it should be flexible and have another method like `boolean handles(...)` that determines if it should handle the description type. For your use-case, `handles` would just always return `true`.

clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/deploy/ExtensibleDescriptionValidator.java
```diff
@@ -0,0 +1,25 @@
...
+  void validate(List priorDescriptions, Object description, ValidationErrors errors);
```
danielpeach, line 24: I think the method signature should be this:

```
<T> void validate(List<T> priorDescriptions, T description, ValidationErrors errors);
```

clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/deploy/ExtensibleDescriptionValidator.java
danielpeach: Could you please also add docs comments to the interface and the method?

KathrynLewis: Sounds good, adding these changes 👍

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/orchestration/AnnotationsBasedAtomicOperationsRegistry.groovy
```diff
@@ -35,6 +36,9 @@ class AnnotationsBasedAtomicOperationsRegistry extends ApplicationContextAtomicO
   @Autowired
   List<CloudProvider> cloudProviders
 
+  @Autowired(required = false)
```
danielpeach, line 39: You might not need this, I think you'll just get an empty list if there aren't any `GlobalDescriptionValidator`s

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/orchestration/CompositeDescriptionValidator.java
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright 2021 Netflix, Inc.
```
danielpeach, line 2: s/Netflix/Armory

danielpeach (approved):

PR #5357, opened 2021-05-11
fix(cloudfoundry): remove null route mappings to avoid falsely thrown error

zachsmith1:
This fixes an issue when the CF Loadbalancer caching agent will succeed with failures because an old application was no longer in the route mappings. It takes one application caching agent cycle to hide this issue, but now that agents could be run in any order it occasionally happens. This doesn't actually create a problem as the caching agent will succeed the next time around but its better to avoid it than see errors in the logs. The fix will just catch any null apps and filter that out of the route mappings instead of throwing that error.


link108:

PR #5360, opened 2021-05-13
feat(kubernetes/job): Return the most recent pod name instead of the list of pod status

german-muzquiz:
Kubernetes jobs may be executed by many pods, and each pod status is saved to the pipeline execution context. If the number of pods is large enough, the pipeline execution context may grow big enough to make orca crash.

The only place where pod status is needed is in the UI to extract the logs of the most recently created pod. A new field `mostRecentPodName` was added to the response that replaces the list of pod status. The full list of pod status may still be returned by setting the new configuration flag `kubernetes.jobs.detailed-pod-status: true` in `clouddriver.yml`.

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesJobProvider.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesJobProvider.java
@@ -29,15 +29,15 @@ import com.netflix.spinnaker.clouddriver.model.JobProvider;
 import com.netflix.spinnaker.clouddriver.security.AccountCredentialsProvider;
 import io.kubernetes.client.openapi.models.V1Job;
 import io.kubernetes.client.openapi.models.V1Pod;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import lombok.Getter;
 import org.apache.commons.lang3.NotImplementedException;
+import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
 @Component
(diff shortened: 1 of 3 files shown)


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesJobProvider.java
```diff
@@ -46,12 +46,15 @@
   @Getter private final String platform = "kubernetes";
   private final AccountCredentialsProvider accountCredentialsProvider;
   private final KubernetesManifestProvider manifestProvider;
+  private final boolean detailedPodStatus;
 
   KubernetesJobProvider(
       AccountCredentialsProvider accountCredentialsProvider,
-      KubernetesManifestProvider manifestProvider) {
+      KubernetesManifestProvider manifestProvider,
+      @Value("${kubernetes.jobs.detailed-pod-status:false}") boolean detailedPodStatus) {
```
ncknt, line 54: Shouldn't it be true by default for backward compatibility?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesJobProvider.java
```diff
@@ -73,14 +76,32 @@ public KubernetesJobStatus collectJob(String account, String location, String id
...
                 })
-            .collect(Collectors.toList()));
+            .collect(Collectors.toList());
+
+    V1Pod mostRecentPod = typedPods.get(typedPods.size() - 1);
+    jobStatus.setMostRecentPodName(
+        mostRecentPod.getMetadata() != null ? mostRecentPod.getMetadata().getName() : "");
+
+    if (detailedPodStatus) {
+      jobStatus.setPods(
```
ncknt, line 102: Does `KubernetesJobStatus` serialize with `pods` if null?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesJobProvider.java
german-muzquiz: The new field `mostRecentPodName` in the response replaces the detailed pod statuses for what the UI needs, so it shouldn't have any difference in behavior unless there are pipelines grabbing statuses from the context with pipeline expressions (unlikely?).
Do you see this default may cause problems?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesJobProvider.java
german-muzquiz: `pods` are not included in the response

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesJobProvider.java
ncknt: I don't know for sure, but it's potentially used this way so I'd default to `true` with a note in the release notes that it may be switched to false in the future.

ncknt (approved):

ncknt:
LGTM!

PR #5361, opened 2021-05-14
fix(core): Unable to list project clusters

jcavanagh:
The affected code is attempting to merge image lists from the "same" builds across various clusters. In the case of Kubernetes, the list returned from KubernetesImageSummary is immutable.

So, always convert the (maybe) immutable list from the ImageSummary implementation to a regular mutable list.

Resolves #6149


dbyron-sf:
The change looks good to me.  Would you mind adding a test that demonstrates the bug/fix?

mattgogerly (approved):

mattgogerly:
@Mergifyio update

PR #5366, opened 2021-05-18
feat(aws): Adding support for modifying mixed instances policy proper…

pdk27:
## Issue
spinnaker/spinnaker#5989

## Summary
New features in modifyServerGroupLaunchTemplate operation:
1. convert a server group backed by launch template to use mixed instances policy 
1. added a new action `PrepareUpdateAutoScalingGroup` to modify operation
1. added unit tests for modify op actions
1. ability to skip modifying launch template (creating a new version) when not required

Changes to existing features: 
1. _new validation in modify op:_
    - **before:** modifyServerGroupLaunchTemplate op always modified a launch template, even when no launch template changes were included in the request. 
    - **now:** modify a launch template only if launch template config changes are requested
1. _`ModifyServerGroupLaunchTemplateDescription`_
    - **before:** extended `ModifyAsgLaunchConfigurationDescription`
    - **now:** has no association with launch configuration code path
1.  _refactors and typo corrections_ 
1. _API call for creating a new launch template version:_
    - **before:** New version was created with source version
    ```
    new CreateLaunchTemplateVersionRequest()
                .withSourceVersion(String.valueOf(sourceVersion.getVersionNumber()))
    ```
    - **now:**  New version is created without using source version. Instead, source launch template data is used to build new version's launch template data. This change was required the API call with param `withSourceVersion` does not support removing launch template parameters and we have a new use case for removing spot options from a launch template.
1. _Modify op without amiName_:
    - **before:** resulted in error `Failed to resolve image id for null`
    - **now:** no error, if amiName is not requested to be modified, existing LT's image ID is used
  
## Testing
- added / updated unit tests
- added integration tests for modify op - will be part of an upcoming PR - https://github.com/pdk27/clouddriver/commit/20a5a6bbe02af448ff35baf89347422e66c2164f
- comprehensive manual testing with various combinations of configuration parameters

### DIFF
--- a/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/UpdateAutoScalingGroup.java
+++ b/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/UpdateAutoScalingGroup.java
@@ -17,27 +17,40 @@
 
 package com.netflix.spinnaker.clouddriver.aws.deploy.ops.actions;
 
+import com.amazonaws.services.autoscaling.model.AutoScalingGroup;
+import com.amazonaws.services.autoscaling.model.InstancesDistribution;
+import com.amazonaws.services.autoscaling.model.LaunchTemplate;
+import com.amazonaws.services.autoscaling.model.LaunchTemplateOverrides;
 import com.amazonaws.services.autoscaling.model.LaunchTemplateSpecification;
+import com.amazonaws.services.autoscaling.model.MixedInstancesPolicy;
 import com.amazonaws.services.autoscaling.model.UpdateAutoScalingGroupRequest;
+import com.amazonaws.services.ec2.model.LaunchTemplateVersion;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 import com.netflix.spinnaker.clouddriver.aws.deploy.description.ModifyServerGroupLaunchTemplateDescription;
 import com.netflix.spinnaker.clouddriver.aws.deploy.ops.ModifyServerGroupLaunchTemplateAtomicOperation.LaunchTemplateException;
 import com.netflix.spinnaker.clouddriver.aws.security.NetflixAmazonCredentials;
+import com.netflix.spinnaker.clouddriver.aws.services.LaunchTemplateService;
 import com.netflix.spinnaker.clouddriver.aws.services.RegionScopedProviderFactory;
 import com.netflix.spinnaker.clouddriver.event.EventMetadata;
 import com.netflix.spinnaker.clouddriver.saga.SagaCommand;
 import com.netflix.spinnaker.clouddriver.saga.flow.SagaAction;
 import com.netflix.spinnaker.clouddriver.saga.models.Saga;
 import com.netflix.spinnaker.credentials.CredentialsRepository;
+import java.util.List;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import lombok.Builder;
 import lombok.Value;
 import lombok.experimental.NonFinal;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.springframework.stereotype.Component;
 
+/** Action to update an AWS EC2 Auto Scaling Group. */
+@Slf4j
 @Component
 public class UpdateAutoScalingGroup
     implements SagaAction<UpdateAutoScalingGroup.UpdateAutoScalingGroupCommand> {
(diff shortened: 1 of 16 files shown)


clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/UpdateAutoScalingGroup.java
```diff
@@ -60,19 +74,142 @@ public Result apply(
...
+          cleanUpOnFailure(
+              regionScopedProvider.getLaunchTemplateService(),
+              command.getLaunchTemplateVersion().getLaunchTemplateId(),
+              command.getNewLaunchTemplateVersionNumber());
+        }
+      } catch (Exception ex) {
+        exceptionMsg.append(
+            "Failed to clean up launch template version! Error: " + ex.getMessage());
+      }
+      throw new LaunchTemplateException(exceptionMsg.toString(), e);
```
pdk27, line 166: A sample error message in task API response for the scenario where:
- updateAutoScalingGroup fails 
- clean up of newly created launch template version is attempted but that also fails

Made the error message clearer to indicate 2 failures. 

```
{
   "key":"kato.tasks",
   "value":[
      {
         "exception":{
            "retryable":true,
            "cause":"Update Failed!!!",
            "details":[
               {
                  "message":"Update Failed!!!"
               },
               {
                  "retryable":true,
                  "userMessage":"null",
                  "message":"Failed to update server group hello-lt_pr-v004.Error: Update Failed!!!\nFailed to clean up launch template version! Error: Failed to delete launch template version 14 for launch template ID lt-077b0c6cd0a07309b because of launchTemplateVersionDoesNotExist",
                  "additionalAttributes":{
                     
                  }
               }
            ],
            "message":"Failed to update server group hello-lt_pr-v004.Error: Update Failed!!!\nFailed to clean up launch template version! Error: Failed to delete launch template version 14 for launch template ID lt-077b0c6cd0a07309b because of launchTemplateVersionDoesNotExist",
            "type":"EXCEPTION",
            "operation":"ModifyServerGroupLaunchTemplateAtomicOperation"
         },
         "resultObjects":[
            {
               "retryable":true,
               "cause":"Update Failed!!!",
               "details":[
                  {
                     "message":"Update Failed!!!"
                  },
                  {
                     "retryable":true,
                     "userMessage":"null",
                     "message":"Failed to update server group hello-lt_pr-v004.Error: Update Failed!!!\nFailed to clean up launch template version! Error: Failed to delete launch template version 14 for launch template ID lt-077b0c6cd0a07309b because of launchTemplateVersionDoesNotExist",
                     "additionalAttributes":{
                        
                     }
                  }
               ],
               "message":"Failed to update server group hello-lt_pr-v004.Error: Update Failed!!!\nFailed to clean up launch template version! Error: Failed to delete launch template version 14 for launch template ID lt-077b0c6cd0a07309b because of launchTemplateVersionDoesNotExist",
               "type":"EXCEPTION",
               "operation":"ModifyServerGroupLaunchTemplateAtomicOperation"
            }
         ],
         "id":"6f73a2ab-cabe-48ab-9d19-5e972aa8186d",
         "history":[
            {
               "phase":"ORCHESTRATION",
               "status":"Initializing Orchestration Task"
            },
            {
               "phase":"ORCHESTRATION",
               "status":"Processing op: ModifyServerGroupLaunchTemplateAtomicOperation"
            },
            {
               "phase":"ORCHESTRATION",
               "status":"Orchestration failed: ModifyServerGroupLaunchTemplateAtomicOperation | LaunchTemplateException: [Failed to update server group hello-lt_pr-v004.Error: Update Failed!!!\nFailed to clean up launch template version! Error: Failed to delete launch template version 14 for launch template ID lt-077b0c6cd0a07309b because of launchTemplateVersionDoesNotExist]"
            }
         ],
         "status":{
            "retryable":true,
            "completed":true,
            "failed":true
         }
      }
   ]
}
```

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/services/LaunchTemplateService.java
```diff
@@ -143,37 +130,155 @@ public LaunchTemplate createLaunchTemplate(
...
+
+    if (result.getUnsuccessfullyDeletedLaunchTemplateVersions() != null
+        && !result.getUnsuccessfullyDeletedLaunchTemplateVersions().isEmpty()) {
+      DeleteLaunchTemplateVersionsResponseErrorItem responseErrorItem =
+          result.getUnsuccessfullyDeletedLaunchTemplateVersions().get(0);
+      ResponseError failureResponseError = responseErrorItem.getResponseError();
+
+      // certain error codes can be considered success when they match the desired end state.
+      // this also acts as a safety net in retry scenarios.
+      List<String> codesConsideredSuccess =
```
pdk27, line 188: @jeyrschabu does this make sense or would you prefer to fail on every failure, irrespective of the error code?

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/UpdateAutoScalingGroup.java
jeyrschabu: Ref: https://github.com/spinnaker/clouddriver/pull/5367#discussion_r638110836
Thanks for running through the scenario. Sounds legit to me. I would like for us to revisit this section in a separate PR (to avoid big changes).  It would have been nice if the api supported updating an `asg` an inline `withTemplateVersion` 😄

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/services/LaunchTemplateService.java
jeyrschabu: Makes sense!

jeyrschabu (approved):
Looks good. As per our chat, my initial concern with mixed instance types is backward compatibility with folks who don't use them.

Also, would be nice to include the request format with `templates overrides`. Another thing that is a bit unclear is how overrides would affect our caching. You explained it to me but having it here for visibility.

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/UpdateAutoScalingGroup.java
pdk27: Sure, would you like to discuss concerns offline or suggest more changes to this Saga step? 
I would definitely like to get your concerns addressed, if any.

pdk27:
> Looks good. As per our chat, my initial concern with mixed instance types is backward compatibility with folks who don't use them.

I tried to maintain backwards compatibility and keep the existing use cases unchanged. With so many combinations of config parameters possible, I was mostly successful, except for the case when the request is to:
-  update launch template and the only non-metadata field requested to change is the spot price
-  the ASG is backed by launch template without mixed instances policy (MIP). 

With this PR, the outcome in this scenario is that ASG will be converted to use MIP and spot price will be set in MIP. 
Before this PR, the ASG launch template would have a new version created with the new spot price. 

The reasoning here was to simply the conditional logic [here](https://github.com/spinnaker/clouddriver/blob/626effd6769b6db0756ef8fe944f7db85abe5294/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/PrepareModifyServerGroupLaunchTemplate.java#L129) as the combination of conditions for special cases was getting out of hand. Secondly, using MIP is always more beneficial for Spot use cases - other features like spot allocation strategies are just a request parameter away. This also aligns with [Spot best practices](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-best-practices.html). 
Here is a unit test that covers this scenario but probably isn't obvious - [test#4](https://github.com/spinnaker/clouddriver/blob/626effd6769b6db0756ef8fe944f7db85abe5294/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/PrepareModifyServerGroupLaunchTemplateSpec.groovy#L201)

I guess this is worth more discussion offline? I just want to make sure there isn't an unexpected behavior or outcome that doesn't make sense for the common case. 


> Also, would be nice to include the request format with `templates overrides`. 

I am working on this. Such a sample request will be available in spinnaker.io docs. Here is a [peek](https://github.com/pdk27/spinnaker.github.io/blob/5ba299cf29e7ba71dcc61cb4326c5f3361aee82b/setup/features/server-group-launch-settings/aws-ec2/launch-templates.md#convert-a-server-group-with-launch-template-to-use-mixed-instances-policy-with-multiple-instance-types-and-capacity-weighting). 

> Another thing that is a bit unclear is how overrides would affect our caching. You explained it to me but having it here for visibility.

In [AWS API, Launch Template overrides ](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_LaunchTemplateOverrides.html)can be used to configure multiple instance types and / or multiple launch templates for overriding the launch template altogether. 

For now, the overrides in Spinnaker support [multiple instance types only](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy#L297).  The single launch template creation happens like before and nothing new is cached as no new launch templates are created. The overrides are [simply passed over to the AWS API ](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/asg/asgbuilders/AsgWithMixedInstancesPolicyBuilder.java#L84) as part of MixedInstancesPolicy which wraps the launch template and other diversification parameters like overrides, allocation strategies and AWS will override the instance type parameter in the launch template during ASG operations.

jeyrschabu:
> > Looks good. As per our chat, my initial concern with mixed instance types is backward compatibility with folks who don't use them.
> 
> I tried to maintain backwards compatibility and keep the existing use cases unchanged. With so many combinations of config parameters possible, I was mostly successful, except for the case when the request is to:
> 
> * update launch template and the only non-metadata field requested to change is the spot price
> * the ASG is backed by launch template without mixed instances policy (MIP).
> 
> With this PR, the outcome in this scenario is that ASG will be converted to use MIP and spot price will be set in MIP.
> Before this PR, the ASG launch template would have a new version created with the new spot price.
> 
> The reasoning here was to simply the conditional logic [here](https://github.com/spinnaker/clouddriver/blob/626effd6769b6db0756ef8fe944f7db85abe5294/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/PrepareModifyServerGroupLaunchTemplate.java#L129) as the combination of conditions for special cases was getting out of hand. Secondly, using MIP is always more beneficial for Spot use cases - other features like spot allocation strategies are just a request parameter away. This also aligns with [Spot best practices](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-best-practices.html).
> Here is a unit test that covers this scenario but probably isn't obvious - [test#4](https://github.com/spinnaker/clouddriver/blob/626effd6769b6db0756ef8fe944f7db85abe5294/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/PrepareModifyServerGroupLaunchTemplateSpec.groovy#L201)
> 
> I guess this is worth more discussion offline? I just want to make sure there isn't an unexpected behavior or outcome that doesn't make sense for the common case.
> 
> > Also, would be nice to include the request format with `templates overrides`.
> 
> I am working on this. Such a sample request will be available in spinnaker.io docs. Here is a [peek](https://github.com/pdk27/spinnaker.github.io/blob/5ba299cf29e7ba71dcc61cb4326c5f3361aee82b/setup/features/server-group-launch-settings/aws-ec2/launch-templates.md#convert-a-server-group-with-launch-template-to-use-mixed-instances-policy-with-multiple-instance-types-and-capacity-weighting).
> 
> > Another thing that is a bit unclear is how overrides would affect our caching. You explained it to me but having it here for visibility.
> 
> In [AWS API, Launch Template overrides ](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_LaunchTemplateOverrides.html)can be used to configure multiple instance types and / or multiple launch templates for overriding the launch template altogether.
> 
> For now, the overrides in Spinnaker support [multiple instance types only](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy#L297). The single launch template creation happens like before and nothing new is cached as no new launch templates are created. The overrides are [simply passed over to the AWS API ](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/asg/asgbuilders/AsgWithMixedInstancesPolicyBuilder.java#L84) as part of MixedInstancesPolicy which wraps the launch template and other diversification parameters like overrides, allocation strategies and AWS will override the instance type parameter in the launch template during ASG operations.

Thanks for this! very much appreciate the thoroughness. TBH this looks ready to be merged. The integration test should help catch regressions.

PR #5367, opened 2021-05-18
Adding integration tests for modify operation

pdk27:
Integration tests for `modifyServerGroupLaunchTemplate` / `updateLaunchTemplate` operation.

Follow-up PR for code changes merged as part of https://github.com/spinnaker/clouddriver/pull/5366


clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/UpdateAutoScalingGroup.java
```diff
@@ -60,19 +73,150 @@ public Result apply(
...
+                        .withSpotMaxPrice(description.getSpotPrice()));
+
+        updateReq.withMixedInstancesPolicy(mip);
+      } else {
+        updateReq.withLaunchTemplate(
+            new LaunchTemplateSpecification()
+                .withLaunchTemplateId(command.launchTemplateVersion.getLaunchTemplateId())
+                .withVersion(String.valueOf(command.launchTemplateVersion.getVersionNumber())));
+      }
+    }
```
jeyrschabu, line 173: Would like to see some of this lifted to other methods. Hard to follow with the nesting while setting values.

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/PrepareUpdateAutoScalingGroup.java
```diff
@@ -0,0 +1,102 @@
...
+                          .withInstanceType(o.getInstanceType())
+                          .withWeightedCapacity(o.getWeightedCapacity()))
+              .collect(Collectors.toList());
+    }
+
+    UpdateAutoScalingGroup.UpdateAutoScalingGroupCommand updateCommand =
+        UpdateAutoScalingGroup.UpdateAutoScalingGroupCommand.builder()
+            .description(description)
+            .launchTemplateVersion(command.launchTemplateVersion)
+            .launchTemplateOverrides(ltOverrides)
```
jeyrschabu, line 73: Does this take nullable values? `ltOverrides` would be `null` if not set in the condition.

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/UpdateAutoScalingGroup.java
```diff
@@ -60,19 +73,150 @@ public Result apply(
...
+
```
jeyrschabu, line 174: Would like to see this broken into smaller methods. Hard to follow with the nestings while checking and setting values.

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/UpdateAutoScalingGroup.java
```diff
@@ -60,19 +73,150 @@ public Result apply(
...
+        // launch template are cached.
+        // Not cleaning up will result in Internal Server Error for Clouddriver API requests and
+        // subsequent Deck errors.
+        if (command.getNewLaunchTemplateVersionNumber() != null) {
+          log.info("Cleaning up to keep the operation atomic.");
+          cleanUpOnFailure(
+              regionScopedProvider.getLaunchTemplateService(),
+              command.getLaunchTemplateVersion().getLaunchTemplateId(),
+              command.getNewLaunchTemplateVersionNumber());
+        }
```
jeyrschabu, line 197: This smells a bit. What happens if the delete fails? Preferably there would be a hint coming back from the call that this version is invalid or incomplete and instruct the caching agent to ignore those?

This could be put in a compensation logic but I forget if our saga implementation has semantics for that.

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/UpdateAutoScalingGroup.java
pdk27: Sure, will try to make it readable. 
The indentation changed after I compiled the code from `38` LOC to `83` :D

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/PrepareUpdateAutoScalingGroup.java
pdk27: Yes, in L#57, `ltOverrides` is null. If there are no overrides and `ltOverrides` remains null at this point, it will be simply passed over as `UpdateAutoScalingGroupCommand` specifies it as a nullable field in UpdateAutoScalingGroup.java L#231

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/UpdateAutoScalingGroup.java
pdk27: #### Reasoning for this clean up logic: 
Currently, the `ModifyServerGroupLaunchTemplate` operation is not atomic. E.g. If the launch template version was successfully created but the update ASG call failed, then the ASG would still refer to the `latest - 1` version which isn't cached. And this results in an  error response from the API for the server group.

So, as an attempt to solve this issue, added this clean up logic which is executed iff a new launch template version was created in previous steps. 

------
The delete logic in LaunchTemplateService.java L#174-192 verifies the responseError message in the result from the [delete API call.](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DeleteLaunchTemplateVersions.html) 

This call is trying to delete the launch template version that was successfully created in the previous steps. 
If the delete of the newly created launch template version fails, the error message should look something like this: 

```
"message": "Failed to update server group hello-lt_pr-v004. Error message(s): Udpate failed!!!The specified launch template ID 'lt' is malformed. (Service: AmazonEC2; Status Code: 400; Error Code: InvalidLaunchTemplateId.Malformed; Request ID: 26759f3d-2e73-4a67-a214-d96a1d439f43; Proxy: null)",
```

Now that I see this, the error message can be made clearer to separate the 2 separate failures (1) updateASG failure (2) deleteLaunchTemplateVersion failure `The specified launch template ID 'lt' is malformed`. Will work on that. 

-----
Can you elaborate on "this version is invalid or incomplete and instruct the caching agent to ignore those" ? Is there a use case you have in mind where the version is invalid? 
Note that we are trying to delete the LT version that was successfully created in the previous steps.

----
Not sure about the `compensation logic`. Would you mind pointing me to a sample snippet?

pdk27: Can you elaborate on "this version is invalid or incomplete and instruct the caching agent to ignore those" ? Is there a use case you have in mind where the version is invalid? 
Note that we are trying to delete the LT version that was successfully created in the previous steps.

----
Not sure about the `compensation logic`. Would you mind pointing me to a sample snippet?

pdk27: @jeyrschabu I addressed your comments in a commit linked to the[ modify op PR ](https://github.com/spinnaker/clouddriver/pull/5366) withOUT the integ tests. 
The current PR is a draft PR and I meant it to be a reference for the upcoming integration tests. Sorry about the confusion. 

Here is the commit that addresses your comments - https://github.com/spinnaker/clouddriver/pull/5366/commits/63123b24f4f604eaf78a0cf9a4818ab4994dea06

pdk27:
@jeyrschabu I addressed your comments in a commit linked to the[ modify op PR ](https://github.com/spinnaker/clouddriver/pull/5366) withOUT the integ tests. 
The current PR is a draft PR and I meant it to be a reference for the upcoming integration tests. Sorry about the confusion. 

Here is the commit that addresses your comments - https://github.com/spinnaker/clouddriver/pull/5366/commits/63123b24f4f604eaf78a0cf9a4818ab4994dea06

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/actions/UpdateAutoScalingGroup.java
jeyrschabu: Adressed in a different PR. This makes sense after a couple of back and forthwith. Some future work to make the overall operation more atomic or handle this failure mode may be needed but not required for this.

jeyrschabu (approved):
I had to do a double look since this change set looked familiar ha!

PR #5368, opened 2021-05-19
fix(manifest-replicas): more flexibility with get/set replicas value

th3morg:
Replicas value was not being set if it didn't already have a value, however users expect that they can omit the value when leveraging the use-source-capacity strategy. Fix is setting the value despite there not already being one. Additionally, if a third party integration creates a manifest via JSON that serializes the replica count as an integer, a casting exception occurs, so this includes an update to play nicely with integers.


nhtzr (approved):

th3morg:
After some further discussion with colleagues, it was pointed out that the kubernetes spec calls for an `int32`. Clouddriver is using doubles, which is technically incorrect. By using a `Number` instead, we can take the `intValue` and Clouddriver will still be happy.

german-muzquiz (approved):

PR #5373, opened 2021-05-29
refactor(cloudfoundry/process): Add the timeout to the application attributes and update the HealthCheckData

jorgebee65:
Add the timeout to the application attributes level and also in processes, modify the updateProcess method in order to populate the HealthCheckData with the proper info


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Processes.java
```diff
@@ -67,12 +67,17 @@ public void updateProcess(
       String guid,
       @Nullable String command,
       @Nullable String healthCheckType,
-      @Nullable String healthCheckEndpoint)
+      @Nullable String healthCheckEndpoint,
+      @Nullable Integer timeout)
       throws CloudFoundryApiException {
     final Process.HealthCheck healthCheck =
```
zachsmith1, line 73: Can we clean this up? I think we should just start by checking of any of them are != null. If so, just do setters for the !null items and then pass that object in. The legacy logic here is unreadable so better to clean it up a bit

zachsmith1 (approved):
One comment otherwise LGTM

zachsmith1 (approved):
LGTM

PR #5380, opened 2021-06-04
Fix null job name in status

dbyron-sf:
Before this PR, when using generateName in kubernetes manifests deployed via deploy manifest or run job (manifest), "null" appears for job names in some places in deck, e.g. Swapping out artifacts in job null from context..., and in pipeline execution source.  For example:
image
After this PR:
image (1)


german-muzquiz (approved):

dbyron-sf:
Thanks for the quick review @german-muzquiz

PR #5381, opened 2021-06-05
fix(aws): Relaxes health check for AWS 

rvazquezglez:
fix(aws): Relaxes health check for AWS using new `AccountHealthIndicator`.

Fixes https://github.com/spinnaker/spinnaker/issues/3923 for AWS.

A similar change was done for Docker Registry and Kubernetes here: https://github.com/spinnaker/clouddriver/pull/4354


ajordens:
The existing implementation should assert that all accounts are configured correctly before marking clouddriver up (and then never go out of service again).

This likely didn't exist when spinnaker/spinnaker#3923 was filed.

Curious why we'd want to move away from that? I'd worry about a service going up until it's downstream dependencies have been verified at least once.

rvazquezglez:
We have seen users with more than AWS 100 accounts, a permissions error with one of them takes clouddriver down. Currently, if you want to check what happened, you need to go to the logs and search through them for finding error, with this change,  it is easier to just check `/health` endpoint and see which is the account with failures.

The motivation is similar to the one that pushed https://github.com/spinnaker/clouddriver/pull/4354 forward.

> we want to report errors with an individual account, but don't want to bring down all of clouddriver because of failures with individual accounts.

ajordens:
I think one subtle difference (that's worked well for us at least) is a happily running clouddriver being taken down vs. a misconfigured clouddriver not starting up.

We've erred towards the latter. Although in the distant past, we did have the situation where a provider health check could take down a previously healthy instance and that's not great.

ajordens:
End of the day tho, I suppose this does make it consistent with some of the other providers. It's not going to affect our AWS usage so I can get behind it.

jasonmcintosh:
SO this is really more tied to dynamic accounts via things like:
https://github.com/spinnaker-plugins/aws-account-registration-plugin-spinnaker
Where all the accounts are loaded/setup, but someone say disables an AWS role temporarily (or longer).  The registration systems still is showing it, but Spinnaker breaks.  

The issue we've seen is CloudDriver continues running on an account failure but if you restart clouddriver to apply a change - clouddriver no longer starts.  We've hit this rather repeatedly unfortunately where we get into a degraded state :(

zachsmith1 (approved):
LGTM pending adding a new bean definition (`@Bean`) in AwsConfiguration or `@Component`on the class itself.

link108:
@mergifyio backport release-1.24.x release-1.25.x release-1.26.x

PR #5383, opened 2021-06-09
feat(cloudfoundry): add support for manifest random route

rvazquezglez:
This features supports random routes which can be defined in cloud foundry manifests. 

For more information about [random routes see here](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html#random-route).


zachsmith1:
Do you think we could just copy the list of words thats used by the cf cli and randomly pick those here as well?

zachsmith1 (approved):
LGTM

PR #5386, opened 2021-06-10
Adding priority to overrides for explicit ordering, adding validation to modify op and related refactor

pdk27:
## Summary
* Add priority to launch template overrides for explicit ordering and related changes
  Add validation to modify op
* Make MIP fields in desc take precedence in clone op
   Refactor in modify op and accommodate overrides priority

## Testing
* added / modified unit tests
* integ tests


jeyrschabu (approved):
👍

PR #5395, opened 2021-06-22
feat(aws): Adding support for capacity rebalance

pdk27:
## Summary
* Adding support for capacity rebalance
  * CreateServerGroup op
  * ModifyAsg op
  * Clone op 
* Refactor of Clone op unit tests

## Issue
https://github.com/spinnaker/spinnaker/issues/6457
Deck Impl: https://github.com/spinnaker/deck/pull/9369

## Testing
* all supported ops tested

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/CopyLastAsgAtomicOperation.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/CopyLastAsgAtomicOperation.groovy
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.clouddriver.aws.deploy.ops
 
 import com.amazonaws.services.autoscaling.model.AutoScalingGroup
 import com.amazonaws.services.autoscaling.model.DescribeAutoScalingGroupsRequest
+import com.amazonaws.services.autoscaling.model.DescribeLifecycleHooksRequest
 import com.amazonaws.services.autoscaling.model.LaunchTemplateSpecification
 import com.amazonaws.services.ec2.model.DescribeSubnetsRequest
 import com.amazonaws.services.ec2.model.LaunchTemplateVersion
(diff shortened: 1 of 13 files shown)


deverton (approved):
Looks good to me.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/CopyLastAsgAtomicOperation.groovy
```diff
@@ -275,10 +277,14 @@ class CopyLastAsgAtomicOperation implements AtomicOperation<DeploymentResult> {
         newDescription.ebsOptimized = description.ebsOptimized != null ? description.ebsOptimized : ebsOptimized
         newDescription.classicLinkVpcId = description.classicLinkVpcId != null ? description.classicLinkVpcId : classicLinkVPCId
         newDescription.classicLinkVpcSecurityGroups = description.classicLinkVpcSecurityGroups != null ? description.classicLinkVpcSecurityGroups : translateSecurityGroupIds(classicLinkVPCSecurityGroups)
-        newDescription.tags = description.tags != null ? description.tags : ancestorAsg.tags.collectEntries {
+        newDescription.tags = description.tags != null ? description.tags : ancestorAsg.tags?.collectEntries {
           [(it.getKey()): it.getValue()]
         }
-        newDescription.blockDevices
+//        newDescription.blockDevices  // todo(pdk27): FixMe. This seems like an incomplete LOC related to blockDevicesTags https://github.com/spinnaker/clouddriver/pull/5244/files.
```
jeyrschabu, line 283: Is this changing existing behavior or just a note to self? Followed the PR but couldn't quite make the connection.

jeyrschabu (approved):
One question but looks good!

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/CopyLastAsgAtomicOperation.groovy
pdk27: Existing behavior as I understand this LOC is that block devices are not cloned. The LOC seems incomplete from the PR linked. 
So, this shouldn't change existing behavior, AFAIK. 

Follow-up question: Do you think blockDevices should be cloned? Wondering if they were left out in the original impl intentionally?  
IMO, the cloning of block devices should follow the exact logic [here](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy#L587). Do you agree?

jeyrschabu: I don't have the historical context on cloning/not cloning block devices. Not sure if the change it was introduced with requires this so I will defer.

But as far as general approach, I would think letting the logic you linked handle that is probably a good idea since it deals with changing instance type ect...

zachsmith1:
Note: this PR adds new functionality and requires the `autoscaling:DescribeLifecycleHooks` permission in the IAM account.

PR #5396, opened 2021-06-23
feat(cloudfoundry): Adds last operation description 

rvazquezglez:
feat(cloudfoundry): Adds last operation description to have a better context of service instance status.

This info will be used in orca, related PR: https://github.com/spinnaker/orca/pull/4138


zachsmith1 (approved):
LGTM

PR #5401, opened 2021-06-24
feat(kubernetes): add retries to kubernetes job executor

apoorvmahajan:
We have experienced some pipelines failures from time to time due to transient kubectl errors. This PR attempts to address that by adding retrycapability for such kubectl calls. 

This capability can be enabled by setting `kubernetes.jobExecutor.retries.enabled: true`. It is turned off by default.

Any new kubectl errors on which retries will be performed can be configured using the `kuberentes.jobExecutor.retries.retryableErrorMessages` key, which accepts a list of strings. The code does a string.contains() check on this error message - so only certain keywords from the original message can be added here, for example, `TLS handshake timeout`.

Logs are added to indicate the k8s account, the action (such as get, delete etc), the k8s resource as well as retry attempts.
Example log:
```
Retry 'mock-account.topPod.test-pod', waiting PT10S until attempt #3. Last attempt failed with exception com.netflix.spinnaker.clouddriver.jobs.JobExecutionException: TLS handshake timeout
```


dbyron-sf (approved):
LGTM

german-muzquiz (approved):
Thanks for the contribution!, I liked the testing of different scenarios.

PR #5407, opened 2021-06-25
fix(aws): Fixing invalid creation of empty AsgConfiguration

pdk27:
## Summary
- Fixing a bug introduced in a previous PR. 
- Restricted usage of AsgConfiguration's default constructor as it doesn't makes sense to use it without config fields. 
- `blockDeviceTags` is not supported by `modifyServerGroupLaunchTemplateDescription`. So, replacing empty ASG Configuration with null to be explicit.


jeyrschabu (approved):
👍

PR #5411, opened 2021-06-30
chore(codeowners): Update the CODEOWNERS for the new googlecloud SIG.

plumpy:
Also update the kubernetes directory to the current kubernetes SIG
owners.


german-muzquiz (approved):

skandragon (approved):
LGTM

PR #5412, opened 2021-06-30
fix(aws): Fix copy of block devices in clone operation

pdk27:
## Summary
Fixed copy of block devices in clone op. 
Fixed LOC - https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/CopyLastAsgAtomicOperation.groovy#L283 
Seems like an incomplete LOC introduced in https://github.com/spinnaker/clouddriver/pull/5244/files

The "clone" behavior for block devices implemented in this PR, is the same behavior seen when creating a server group with `copySourceCustomBlockDeviceMappings` enabled - [code ref](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy#L442)

## Testing
Manually created server groups with block devices and cloned them.  Block devices were created / cloned as expected.


cfieber (approved):
lgtm thanks

PR #5413, opened 2021-07-01
fix(cf/process): check if healthcheck is null

jorgebee65:
Check if healthcheck type is configured on the manifest then proceed to assign the data otherwise skip the setter


zachsmith1 (approved):
LGTM

PR #5414, opened 2021-07-01
feat(cloudfoundry): add remove bindings option to destroy service

zachsmith1:
Adding a new flag to the destroy service stage that will attempt to unbind any applications prior to destroying the service. It will only allow unbinding server groups that belong to the spinnaker application running the pipeline.

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/description/DestroyCloudFoundryServiceDescription.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/description/DestroyCloudFoundryServiceDescription.java
@@ -16,11 +16,15 @@
 
 package com.netflix.spinnaker.clouddriver.cloudfoundry.deploy.description;
 
+import com.netflix.spinnaker.clouddriver.cloudfoundry.model.CloudFoundrySpace;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 
 @Data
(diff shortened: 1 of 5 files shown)


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/description/DestroyCloudFoundryServiceDescription.java
```diff
@@ -16,11 +16,23 @@
...
 
 @Data
 @EqualsAndHashCode(callSuper = true)
-public class DestroyCloudFoundryServiceDescription extends AbstractCloudFoundryServiceDescription {
+public class DestroyCloudFoundryServiceDescription
+    extends AbstractCloudFoundryServerGroupDescription {
   private String serviceInstanceName;
+  private CloudFoundrySpace space;
+  private boolean removeBindings;
+  private String application;
```
jorgebee65, line 32: We can use the *serverGroupName* (from AbstractCloudFoundryDescription) instead adding a new attribute here.

zachsmith1: Unfortunately this stage doesn't add that because it's not operating on a particular server group. However, we still want to validate that it has auth to do the work at the app level. This field is being added with the associated deck PR

jorgebee65 (approved):
LGTM

PR #5416, opened 2021-07-05
fix(cloudfoundry): include final asg when checking taken slots

mattgogerly:
The filter for getting existing ASGs didn't include the "final" (v999) ASG. At v998 any subsequent deployment tries to create v999, and fails if that already exists. This means at v999 the cluster name is unusable for further deployments.

To fix this I'm including the final ASG in the query filter, so that v999 is recognised as a "taken slot".


zachsmith1 (approved):
LGTM

PR #5421, opened 2021-07-08
fix(kube/test): Use kind instead of k3s

german-muzquiz:
The kubernetes version of k3s used in integration tests is EOL (https://github.com/rancher/k3d/issues/663). Also with newer docker engine versions the tests sometimes fail because of k3s container being killed or k3s losing network connectivity. Using k3d for managing the k3s cluster could work, however that adds yet another abstraction layer to an already complex test infrastructure.

This PR changes k3s for [kind](https://kind.sigs.k8s.io), which is another lightweight kubernetes cluster used for development. The tradeoff is that we need to download kind and kubectl as part of the tests.


clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeployManifestIT.java
```diff
@@ -1108,36 +1108,6 @@ public void shouldDeployRedBlackMultidoc() throws IOException, InterruptedExcept
     KubeTestUtils.disableManifest(baseUrl(), body, account1Ns, "replicaSet " + appName + "-v000");
 
     // ------------------------- then --------------------------
-    String port =
```
dbyron-sf, line 1111: I'm probably missing something...is there something different to do with kind to get the same behavior? Or, how does this "just work" now?

german-muzquiz: Unfortunately you cannot use services of type `NodePort` in kind and expect them to be reachable from `localhost`, (at least in non-linux hosts), because it's running inside a virtual docker network. I opted for removing these checks that query the exposed service and only rely on reading the pods labels to assert that only one pod has the selector label.

dbyron-sf: OK.  Maybe this is even a better test this way?  Like, if it fails then maybe there's some other place to wait for feedback?

dbyron-sf (approved):

nhtzr (approved):

link108:

link108:
@Mergifyio backport release-1.24.x release-1.25.x

PR #5424, opened 2021-07-12
feat(cloudfoundry): add versioning to cf services

zachsmith1:
This is a new flag that will allow cloudfoundry services to be versioned similar to how spinnaker handles applications. If this flag is checked, a version will be appended to the service name. I.e. servicename-v001. We are also returning the previous version name via a field called previousInstanceName to the stage so that a following stage can delete, etc. 
consolidates: https://github.com/spinnaker/clouddriver/pull/5422 and https://github.com/spinnaker/clouddriver/pull/5418


jorgebee65 (approved):
LGTM

PR #5427, opened 2021-07-14
feat(cloudfoundry): onlySpinnakerManaged flag for server groups

mattgogerly:
Adds a flag to Cloud Foundry accounts "onlySpinnakerManaged" that is disabled by default. When enabled any application in the foundation that does not end in "vXXX" (i.e. was not deployed by Spinnaker) will be ignored.

This is useful when foundations have historically been deployed to by other tooling. Since Spinnaker creates unconfigured applications for server groups it can see, these non-Spinnaker apps block the creation of the same app name in Spinnaker.

Whilst checking the app name ends in a version is a bit of a hack, unfortunately other avenues are unavailable:
* Metadata endpoints can't be supported until the minimum supported version for CF is bumped.
* Environment variable endpoints may be restricted for security reasons (so we can't check for the pipeline ID env variable)


mattgogerly:
Closing/reopening to force rebuild

zachsmith1 (approved):
LGTM

PR #5430, opened 2021-07-16
fix(ecs): handle service names without sequences

deverton:
The new naming logic for ECS services doesn't cope if the ECS cluster
only contains service names that have no sequence. This change fixes
that by ignoring services without a sequence for the purposes of
calculating the next free sequence number.

Reporting in Slack #ecs channel here https://spinnakerteam.slack.com/archives/C686H9WG7/p1626356252008700


igor-dmitriev (approved):
lgtm

allisaurus (approved):
LGTM, thanks for this fix + test!

PR #5433, opened 2021-07-20
fix(lambda): initialize evictions map with a mutable map

zachsmith1:
This bug would cause the lambda caching agent ondemand cache to fail because it would initialize the evictions map with an immutable Map and then try to put a key/value pair to it, resulting in an error and failing the caching agent. This fix changes the map to an empty HashMap.


rvazquezglez (approved):
lgtm

jorgebee65 (approved):
LGTM

PR #5437, opened 2021-07-20
fix(lambda):  refactor and fix handle and ondemand caching for lambda

zachsmith1:
The lambda caching agent wasn't handling evictions correctly so this adds the implementation. It's doing a very simple comparison of the keys in the db vs the new keys and adding eviction keys to the DefaultCacheResult. 

We also added a fix for loadData so that it properly stores data and reconciles ondemand data. 

We also added a fix for onDemand caching because it wasn't fully implemented. It now stores data correctly in the onDemand table and the regular interval will process it like most caching agents today.

We also implemented pendingOnDemandRequests so that a caller can see if their ondemand data has been processed.


rvazquezglez (approved):
LGTM

jorgebee65 (approved):

zachsmith1:
@Mergifyio backport release-1.26.x release-1.25.x release-1.24.x

PR #5442, opened 2021-07-23
chore(kubernetes): refactor BootstrapKubernetesConfigurationProvider class so that it can be reused elsewhere

apoorvmahajan:
## Purpose
Issue: https://github.com/spinnaker/spinnaker/issues/6492
this is PR 1 from the proposed fixes mentioned in the above issue

## Changes
 #5125 introduced a way to custom bind k8s config properties. This PR moves that logic to a common module so that it can used for non-k8s config properties as well


clouddriver-configserver/src/main/java/com/netflix/spinnaker/clouddriver/config/AbstractBootstrapCredentialsConfigurationProvider.java
```diff
@@ -0,0 +1,163 @@
+/*
+ * Copyright 2021 Netflix, Inc.
```
dbyron-sf, line 2: Salesforce?

dbyron-sf (approved):

clouddriver-kubernetes/clouddriver-kubernetes.gradle
```diff
@@ -52,6 +52,7 @@ dependencies {
   implementation project(":clouddriver-api")
   implementation project(":clouddriver-artifacts")
   implementation project(":clouddriver-core")
+  implementation project(":clouddriver-configserver")
   implementation project(":cats:cats-core")
   implementation project(":clouddriver-security")
```
dbyron-sf, line 57: I think you can remove json-flattener from here.

apoorvmahajan: I get a gradle build error if I remove that

dbyron-sf: Sorry, too many working directories...me too.

clouddriver-configserver/src/main/java/com/netflix/spinnaker/clouddriver/config/ConfigurationProvider.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * <p>This interface defines the necessary actions which would be required to do the same.
+ */
+public interface ConfigurationProvider {
+  /**
+   * Returns the desired configuration properties, bound to the target implementation class.
+   *
+   * @return a target implementation class for a property
+   */
+  Object getConfigurationProperties();
```
german-muzquiz, line 39: I think it will work better if you use generics instead of `Object`, so the interface definition would be something like
```
public interface ConfigurationProvider<T> {
   T getConfigurationProperties();
   ...
}
```

apoorvmahajan: fixed

german-muzquiz (approved):

PR #5444, opened 2021-07-26
refactor(kubernetes): move Kubernetes Accounts config to its own class

apoorvmahajan:
Purpose
Issue: spinnaker/spinnaker#6492
this is PR 2 from the proposed fixes mentioned in the above issue

Changes
- Move Kubernetes Accounts config to its own class.

This is required because of side effects from the custom binding logic introduced in the PR https://github.com/spinnaker/clouddriver/pull/5125. The custom property binding fix was specifically targeted to improve account loading times from the config file. This works well if the config class, to which the config file is being deserialized, only has accounts. But if it has any other properties, these would need to be explicitly handled by the custom binding logic. This becomes an issue if there are a lot of such properties. Moreover, these properties would need to be co-located in the same file where the accounts are defined.

By refactoring Accounts into its own class, we solve the above issues since only the accounts property is now being custom bound - all other properties are loaded by spring in the default manner. This means that accounts and these properties need not be co-located anymore.


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/config/KubernetesConfiguration.java
```diff
@@ -53,6 +56,19 @@ public KubernetesConfigurationProperties kubernetesConfigurationProperties() {
     return new KubernetesConfigurationProperties();
   }
 
+  @Bean
+  @RefreshScope
+  @ConfigurationProperties("kubernetes")
+  @ConditionalOnProperty(
```
german-muzquiz, line 62: I'm trying to understand this conditional, shouldn't we create this bean in all cases regardless of the value of `kubernetes.custom-property-binding-enabled`? What happens if this bean is not created?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/config/KubernetesConfiguration.java
```diff
@@ -53,6 +56,19 @@ public KubernetesConfigurationProperties kubernetesConfigurationProperties() {
...
+      prefix = "kubernetes",
+      name = "custom-property-binding-enabled",
+      havingValue = "false",
+      matchIfMissing = true)
+  @Primary
```
german-muzquiz, line 67: Is there a reason to mark it `@Primary`? Doing that will prevent any plugin to override it, which may be locking extensibility for kubernetes account management.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/config/KubernetesConfiguration.java
apoorvmahajan: So this property `kubernetes.custom-property-binding-enabled` is an opt-in property. By default it is set to false. So this bean is the one that should be selected when it is either set to false, or not defined. That would mean, the user is letting Springboot do the properties loading and binding.

If the bean is not created, then [this bean](https://github.com/spinnaker/clouddriver/blob/00d498ed9cad0f90a4e8a862ffae5ca49ae9fccb/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/config/KubernetesCustomBinderConfiguration.java#L43-L48) is used.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/config/KubernetesConfiguration.java
apoorvmahajan: Since there are two bean definitions, one here (the default bean) and the [one here](https://github.com/spinnaker/clouddriver/blob/00d498ed9cad0f90a4e8a862ffae5ca49ae9fccb/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/config/KubernetesCustomBinderConfiguration.java#L43-L48), I was getting the error where it says could not auto-wire, multiple bean definitions found. The way around it, at least for me, was to designate one of these as the primary. If there's a better way to do it, then I'm all ears.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/config/KubernetesConfiguration.java
german-muzquiz: I didn't notice the other bean definition, looks good.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/config/KubernetesConfiguration.java
german-muzquiz: According to the `@ConditionalOnProperty` annotations of both beans, they are mutually exclusive and only one should be created at any given time. What value of `kubernetes.custom-property-binding-enabled` did you have when you got the error? If it was `false`, then the beans from `KubernetesCustomBinderConfiguration` should not be created, maybe that annotation needs a syntax like this one?

apoorvmahajan: So I was getting this error with `kubernetes.custom-property-binding-enabled: true` which made me resort to using `@Primary`:
```
***************************
APPLICATION FAILED TO START
***************************

Description:

Parameter 1 of method kubernetesCredentialsLoader in com.netflix.spinnaker.config.KubernetesConfiguration required a single bean, but 2 were found:
	- kubernetesAccountProperties: defined in BeanDefinition defined in class path resource [com/netflix/spinnaker/config/KubernetesConfiguration.class]
	- kubernetesAccounts: defined in BeanDefinition defined in class path resource [com/netflix/spinnaker/config/KubernetesCustomBinderConfiguration.class]


Action:

Consider marking one of the beans as @Primary, updating the consumer to accept multiple beans, or using @Qualifier to identify the bean that should be consumed

```

It turns out I simply needed to name the bean methods the same (i.e. `kubernetesAccountProperties`) and that makes it work without the `@Primary` notation.

german-muzquiz (approved):

PR #5452, opened 2021-07-28
feat(lambda): run sdk calls concurrently

zachsmith1:
WIP


clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/service/LambdaService.java
```diff
@@ -0,0 +1,374 @@
...
+    this.amazonClientProvider = amazonClientProvider;
+    this.account = account;
+    this.region = region;
+    this.mapper = mapper;
+    this.TIMEOUT_MINUTES = lambdaServiceConfig.getRetry().getTimeout();
+    this.RETRIES = lambdaServiceConfig.getRetry().getRetries();
+    this.executorService =
+        Executors.newFixedThreadPool(
+            computeThreads(serviceLimitConfiguration, lambdaServiceConfig));
+  }
```
nimakaviani, line 68: great to see this happen. thanks for the implementation. 

I feel like we can actually split it to 2 classes for better readability. One being the `LambdaService` class that includes all the calls to the API endpoints, and another as the `LambdaDataSupplier` or something, that has `getAllFunctions()` as well as all the functions to add `attributes` and doing `retries`, `computeThreads` etc.

Basically one specifically to interface with the API and one to do the heavy lifting of reading the data and filling the cache.

clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/provider/agent/LambdaCachingAgent.java
```diff
@@ -128,52 +129,25 @@ public CacheResult loadData(ProviderCache providerCache) {
...
+      throw new SpinnakerException(
+          "Failed to populate the lambda cache for account '"
+              + account.getName()
+              + "' and region '"
+              + region
+              + "' because: "
+              + e.getMessage());
     }
 
+    buildCacheData(lambdaCacheData, appLambdaRelationships, allLambdas);
```
nimakaviani, line 149: Didnt dig enough to know if it is possible or not, but do we need to build the cache data in one batch, or can we do it incrementally? like when we have complete data for some functions collected, we update the cache for those functions and iterate. It will add even more performance benefits if possible

nimakaviani (commented):
thanks for implementing this. left some comments. will test this branch on my local spinnaker deployment too. exciting

clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/service/LambdaService.java
zachsmith1: I like this idea -- we can take this point away and implement further refactoring soon. I think at the moment, we want to try and get this in ASAP

clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/provider/agent/LambdaCachingAgent.java
zachsmith1: Same as above -- I think we can try to optimize a few areas. Will note this as well

rvazquezglez (approved):

zachsmith1:
@Mergifyio backport release-1.26.x release-1.25.x release-1.24.x

PR #5462, opened 2021-08-04
fix(kubernetes): don't try to manage traffic for manifests that can't be receive traffic

mattgogerly:
Traffic management is applied to all manifests in a Deploy Manifest stage. Once manifests have been sorted by priority, if the service does not already exist then the stage will fail when trying to manage traffic for manifests with higher priority than the service (e.g. namespace).


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/manifest/KubernetesDeployManifestOperation.java
```diff
@@ -124,11 +122,20 @@ public OperationResult operate(List<OperationResult> _unused) {
                     }
                   }
 
-                  setTrafficAnnotation(description.getServices(), manifest);
-                  if (description.isEnableTraffic()) {
-                    KubernetesManifestTraffic traffic =
-                        KubernetesManifestAnnotater.getTraffic(manifest);
-                    applyTraffic(traffic, manifest, inputManifests);
+                  KubernetesHandler handler =
```
german-muzquiz, line 125: I think this `handler` instance is the same as `deployer` above, so you can reuse that.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/ServerGroupHandler.java
```diff
@@ -20,7 +20,7 @@
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.view.model.KubernetesServerGroup;
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.view.provider.data.KubernetesServerGroupCacheData;
 
-public interface ServerGroupHandler {
+public interface ServerGroupHandler extends CanReceiveTraffic {
```
german-muzquiz, line 23: You need to also add this interface to `ServerGroupManagerHandler`, that's the handler for kubernetes `Deployment` objects. Probably it's not actually used for red/black deployments but to keep semantics correct, `Deployment` can also receive traffic.

german-muzquiz (commented):
Thanks for putting this together, I only have a couple of small comments. 

The problem of "load balancer does not exist" when the `Service` was deployed the first time as part of the same stage should have been fixed by https://github.com/spinnaker/clouddriver/pull/5248, so I see the net effect of this PR is just to not add traffic related labels/annotations to objects not receiving traffic like `ConfigMap`, more than avoiding "load balancer not found" errors.

mattgogerly:
> Thanks for putting this together, I only have a couple of small comments.
> 
> The problem of "load balancer does not exist" when the `Service` was deployed the first time as part of the same stage should have been fixed by #5248, so I see the net effect of this PR is just to not add traffic related labels/annotations to objects not receiving traffic like `ConfigMap`, more than avoiding "load balancer not found" errors.

We're still seeing the "load balancer x does not exist" error - it seems to be happening with ConfigMaps, which are deployed before the service itself. I haven't had a chance to dig into why yet, will have a closer look this afternoon.

Will also address your comments 👍

PR #5468, opened 2021-08-06
chore(lambda): Adds more tests for LambdaService.

rvazquezglez:
chore(lambda): Adds more tests for LambdaService.


zachsmith1 (approved):
LGTM

PR #5473, opened 2021-08-10
refactor(aws): refactor accounts

apoorvmahajan:
* move accounts out of the CredentialsConfig class so that it is easier to custom bind them when reading accounts from a local config file.
This is for the following issue: https://github.com/spinnaker/spinnaker/issues/6492


dbyron-sf (approved):
LGTM.  It's tempting to get lombok to add getters and setters in AccountConfiguration, but since that wasn't happening in CredentialsConfig, it's probably better to leave that for a subsequent PR.

PR #5475, opened 2021-08-11
fix(gce) : modified internal load balancer services to fix duplication and NPE

ravichander82:
This code change is to address the GCE issues [#6136,](https://github.com/spinnaker/spinnaker/issues/6136) [#6253](https://github.com/spinnaker/spinnaker/issues/6253)

With modification of of definition of 'ilbServices' both issues are resolved by removing the null values and duplication, 
this way deploy stage will be complete in case of both missing 'backend-service'  as parameter and when by manually adding valid 'region-backend-service' as parameter

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
@@ -320,9 +320,7 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
       List<String> existingRegionalLbs = instanceMetadata[REGIONAL_LOAD_BALANCER_NAMES]?.split(",") ?: []
       List<String> regionBackendServices = instanceMetadata[REGION_BACKEND_SERVICE_NAMES]?.split(",") as List ?: []
       def ilbServices = internalLoadBalancers.collect { it.backendService.name }
-      if (regionBackendServices.size() != 0) {
-        ilbServices.addAll(regionBackendServices)
-      }
+      ilbServices.addAll(regionBackendServices)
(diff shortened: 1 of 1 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -318,7 +318,12 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
     List<BackendService> regionBackendServicesToUpdate = []
     if (internalLoadBalancers || internalHttpLoadBalancers) {
       List<String> existingRegionalLbs = instanceMetadata[REGIONAL_LOAD_BALANCER_NAMES]?.split(",") ?: []
-      def ilbServices = internalLoadBalancers.collect { it.backendService.name } + (instanceMetadata[REGION_BACKEND_SERVICE_NAMES]?.split(",") as List) ?: []
+      List<String> regionBackendServices = instanceMetadata[REGION_BACKEND_SERVICE_NAMES]?.split(",") as List ?: []
+      def ilbServices = internalLoadBalancers.collect { it.backendService.name }
+      if (regionBackendServices.size() != 0) {
```
mattgogerly, line 323: Is there a reason for this if statement? If the list is empty then addAll will just do nothing anyway.

ravichander82: done those changes,removed the 'if' statement

skandragon:
@plumpy Any idea why the it-test and pull_request build are failing?  It seems to be due to things unrelated to this simple change.

ravichander82:
@plumpy / @skandragon  can you approve the build process I have rebased the branch with latest changes

ravichander82:
will raise a new PR

PR #5478, opened 2021-08-16
fix(sql): Cancel futures of sql scheduler executions

german-muzquiz:
The caching agent cleanup mechanism introduced in https://github.com/spinnaker/clouddriver/pull/5332 removes executions that have not finished in a configurable amount of time (1 hour default). Even though the execution object is removed from the running list, its `AgentExecutionAction-x` thread is still left running. These threads hold in memory the list of `KubernetesManifest` loaded by the caching agents, which may eventually lead to `OutOfMemory` errors.

This PR introduces these changes:
* Cancel the futures removed by the cleanup zombie logic, passing `interrupt=true` to interrupt the running thread.

### DIFF
--- a/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
+++ b/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
@@ -33,11 +33,6 @@ import com.netflix.spinnaker.config.ConnectionPools
 import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
 import com.netflix.spinnaker.kork.sql.routing.withPool
 import java.sql.SQLException
-import java.util.concurrent.ConcurrentHashMap
-import java.util.concurrent.ExecutorService
-import java.util.concurrent.Executors
-import java.util.concurrent.ScheduledExecutorService
-import java.util.concurrent.TimeUnit
 import java.util.regex.Pattern
 import java.util.regex.Pattern.CASE_INSENSITIVE
 import org.jooq.DSLContext
@@ -45,6 +40,7 @@ import org.jooq.impl.DSL.field
 import org.jooq.impl.DSL.table
 import org.slf4j.LoggerFactory
 import org.springframework.dao.DataIntegrityViolationException
+import java.util.concurrent.*
 
 /**
  * IMPORTANT: Using SQL for locking isn't a good idea. By enabling this scheduler, you'll be adding a fair amount of
@@ -76,6 +72,7 @@ class SqlClusteredAgentScheduler(
 
   private val agents: MutableMap<String, AgentExecutionAction> = ConcurrentHashMap()
   private val activeAgents: MutableMap<String, NextAttempt> = ConcurrentHashMap()
+  private val activeAgentsFutures: MutableMap<String, Future<*>> = ConcurrentHashMap()
   private val enabledAgents: Pattern
 
   private val referenceTable = "cats_agent_locks"
@@ -129,7 +126,7 @@ class SqlClusteredAgentScheduler(
     acquiredAgents.forEach { agentType, nextAttempt ->
       val exec = agents[agentType]
       if (exec != null) {
-        agentExecutionPool.submit(AgentJob(nextAttempt, exec, this::agentCompleted))
+        activeAgentsFutures[agentType] = agentExecutionPool.submit(AgentJob(nextAttempt, exec, this::agentCompleted))
       }
     }
   }
@@ -241,7 +238,16 @@ class SqlClusteredAgentScheduler(
       .forEach {
         log.warn("Found zombie agent {}, removing it", it.key)
         activeAgents.remove(it.key, it.value)
-    }
+        // Cancel zombie futures interrupting their AgentExecutionAction threads
+        val f = activeAgentsFutures.remove(it.key)
+        if (f == null) {
+          log.warn("Agent execution without future for cancelling it: {}", it.key)
+        } else {
+          if (!f.cancel(true) && !f.isCancelled) {
+            log.error("Unable to cancel execution for agent {} after {}ms (Future.cancel returned false). This may leak resources!", it.key, zombieAgentThreshold)
+          }
+        }
+      }
   }
 
   private fun tryAcquireSingle(agentType: String, now: Long, timeout: Long): Boolean {
@@ -300,6 +306,7 @@ class SqlClusteredAgentScheduler(
       releaseLock(agentType, nextExecutionTime)
     } finally {
       activeAgents.remove(agentType)
+      activeAgentsFutures.remove(agentType)
     }
   }
 


danielGz (approved):

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
```diff
@@ -241,7 +238,12 @@ class SqlClusteredAgentScheduler(
       .forEach {
         log.warn("Found zombie agent {}, removing it", it.key)
         activeAgents.remove(it.key, it.value)
-    }
+        // Cancel zombie futures interrupting their AgentExecutionAction threads
+        val f = activeAgentsFutures.remove(it.key)
+        if (f?.cancel(true) == false) {
+          log.warn("Unable to cancel agent execution future for {}", it.key)
```
dbyron-sf, line 244: Minor point, but is it worth calling [isCancelled](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Future.html#isCancelled()) and then only logging if that returns true?  And then logging an error instead of a warning?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutor.java
```diff
@@ -895,7 +895,7 @@ private Retry getRetry(RetryRegistry retryRegistry, String identifier) {
       // Hence, we need to translate that exception back into a JobResult object - but
       // we only need to do it for KubectlException (since that is explicitly thrown above) and not
       // for any other ones.
-      return finalResult.build();
+      return finalResult.error(e.getMessage()).build();
```
dbyron-sf, line 898: Doesn't finalResult.error get set above?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesCachingAgent.java
```diff
@@ -106,7 +106,12 @@ protected KubernetesCachingAgent(
   @Nonnull
   private ImmutableList<KubernetesManifest> loadNamespaceScopedResources(
       @Nonnull Iterable<KubernetesKind> kubernetesKinds) {
-    return getNamespaces().parallelStream()
+    return getNamespaces()
```
dbyron-sf, line 109: This feels like it belongs in a separate commit, or maybe even PR since it's not specific to the sql scheduler.

nhtzr (approved):

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
german-muzquiz: According to the javadoc of `cancel`:
```
After this method returns, subsequent calls to isDone will always return true. Subsequent calls to isCancelled will always return true if this method returned true.
```
So I think it's redundant to call `isCancelled` after `cancel`. I'll adjust the logging though to log in both cases.

dbyron-sf: When `cancel` returns false, I was thinking calling `isCancelled` could still make sense -- because we'd learn if the task completed normally.  I don't think we need to log (nor call `isCancelled`) if `cancel` returns true.

german-muzquiz: `isCancelled` doesn't indicate clearly if a task completed normally (only if it was cancelled before completing normally), in that case it's more useful to check `isDone`, so we know if the task is still left running after cleanupZombieAgents finish.

dbyron-sf: Good point.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutor.java
german-muzquiz: Ohh that's right. I misread the logs and thought the actual error was not being printed.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
```diff
@@ -241,7 +238,16 @@ class SqlClusteredAgentScheduler(
       .forEach {
         log.warn("Found zombie agent {}, removing it", it.key)
         activeAgents.remove(it.key, it.value)
-    }
+        // Cancel zombie futures interrupting their AgentExecutionAction threads
+        val f = activeAgentsFutures.remove(it.key)
+        if (f == null) {
+          log.warn("Agent execution without future for cancelling it: {}", it.key)
+        } else {
+          if (!f.cancel(true) && !f.isDone) {
```
dbyron-sf, line 246: Sorry for not catching this earlier. 
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Future.html#cancel(boolean) says:
```
After this method returns, subsequent calls to isDone() will always return true
```
Doesn't this mean the if condition will never be true?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesCachingAgent.java
german-muzquiz: Yeah, I'll have a separate PR for this one.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt
german-muzquiz: Haha that's right, so how about this
```
if (!f.cancel(true) && !f.isCancelled) {
            log.error("Unable to cancel execution for agent {} after {}ms (Future.cancel returned false). This may leak resources!", it.key, zombieAgentThreshold)
}
```
We need a manual to interpret the meaning of these booleans 😄

dbyron-sf: Trying to lay out the baby steps to see if I've got my head on straight.

I think we agree the thing that we care about is when we can't cancel the task.  And so the natural thing would be to log whenever the cancel method returns false.  Except for this little note in the javadocs:
```
false if the task could not be cancelled, typically because it has already completed normally
```
So if `cancel` returns false, and `isCancelled` is false, does that mean the task completed normally?  In that case we don't want to log.  And I guess, if `isCancelled` is true, we don't really need a log message either.

So that one case where `cancel` returns false, `isCancelled` is false, and the task didn't complete normally -- that's when we want to log...but I'm not sure we have enough information to determine that.  But maybe the log message you've got (the part about `may leak resources`) is sufficiently vague to be just right :)

dbyron-sf (approved):

link108:
@mergifyio backport release-1.25.x release-1.26.x release-1.26.x

link108:

PR #5479, opened 2021-08-17
fix(build): updating dependencies

link108:


gradle.properties
```diff
@@ -1,7 +1,7 @@
-fiatVersion=1.26.0
-korkVersion=7.92.0
+fiatVersion=version-1.14.4
+korkVersion=7.92.4
 org.gradle.parallel=true
-spinnakerGradleVersion=8.10.0
+spinnakerGradleVersion=8.11.0
```
dbyron-sf, line 4: I bet going to 8.14.0 would help this...and I'd probably go all the way to the most recent -- 8.15.0.

german-muzquiz: I confirmed that `8.14.0` works for not hitting bintray

link108: I'll update to 8.14.0, I did try 8.15.0 in another project for the release-1.24.x branch and I hit some weird gradle issues

PR #5480, opened 2021-08-17
fix(k8s/perf): Don't use ForkJoin commonPool in caching agents

german-muzquiz:
Replacing parallelStream by the sequential stream on KubernetesCachingAgent for processing namespace caching. 

This is because using parallelStream will use the ForkJoin common pool, which has very few threads (only one thread confirmed when running in EKS). This is causing contention because all the agents try to use this single thread from the common pool. Using the regular stream, caching time is improved because now the agents are using their respective AgentExecutionAction threads for caching.

Agent execution duration before (100 k8s accounts)
image

After
image


dbyron-sf (approved):

dbyron-sf:
@mergifyio update

link108:
@mergifyio backport release-1.24.x release-1.25.x release-1.26.x release-1.27.x

PR #5481, opened 2021-08-17
Sql caching duplicates

dbyron-sf:


cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -552,7 +553,12 @@ class SqlCache(
     items
       .filter { it.id != "_ALL_" && it.id.length <= sqlConstraints.maxIdLength }
       .forEach {
-        currentIds.add(it.id)
+        if (!currentIds.add(it.id)) {
+            log.error("agent: '${agent}': type: '$type': only one item with id '${it.id}' allowed")
```
deverton, line 557: Is this really an error? Maybe a warning would suffice. There's not a whole lot a user of clouddriver can do if this error pops up.

deverton (approved):
Seems reasonable to me.

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
dbyron-sf: Yeah, it's a tough call.  To me, it really is an error, but it requires a code change to fix.  I'm fine to make it a warning, as if to say...we're doing the best we can with partial data.

dbyron-sf:
@mergifyio update

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
dbyron-sf: https://github.com/spinnaker/clouddriver/pull/5481/commits/6248b7989e13550d7d64f75651740c4816225a12

link108:
@mergifyio backport release-1.26.x release-1.27.x

PR #5488, opened 2021-08-20
fix(provider/azure): Use health extension in instance status

kmccarp:
This fix fixes two things related to Azure instance health:

1. The health extension couldn't be applied to any instances because the health
   extension required int port (it was a String everywhere)
     - fixed by changing type to int
2. The health extension wasn't being used by the cluster view at all.
     - fixed by adding a health extension check after the default
       "is running" check

This has been tested in two ways:

1. Unit tests
2. Deployment of scale sets to my Azure account

### DIFF
--- a/clouddriver-azure/src/test/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroups/deploy/AzureServerGroupResourceTemplateSpec.groovy
+++ b/clouddriver-azure/src/test/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroups/deploy/AzureServerGroupResourceTemplateSpec.groovy
@@ -88,7 +88,7 @@ class AzureServerGroupResourceTemplateSpec extends Specification {
   }
 
   def 'should generate correct ServerGroup resource template health extension is blank'() {
-    description = createHealthDescription("", "", "")
+    description = createHealthDescription("", 0, "")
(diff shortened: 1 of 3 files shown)


jkschneider:
@kmccarp This looks to be a breaking change to existing pipeline stages:

image

clouddriver-azure/src/test/groovy/com/netflix/spinnaker/clouddriver/azure/resources/servergroups/deploy/AzureServerGroupResourceTemplateSpec.groovy
```diff
@@ -2814,7 +2814,7 @@ class AzureServerGroupResourceTemplateSpec extends Specification {
               "autoUpgradeMinorVersion" : true,
               "settings" : {
                 "protocol" : "https",
-                "port" : "7000",
+                "port" : 7000,
```
jkschneider, line 2817: This looks like a breaking change (string to int)

jkschneider: Sorry, I misinterpreted this. This is the request that will be shipped to Azure, which must be an int.

jkschneider (approved):

PR #5494, opened 2021-08-24
feat(provider/azure): Allow empty load balancer

kmccarp:
Allow server groups (Azure Virtual Machine Scale Sets) to be spun up without needing to be fronted by a load balancer (Azure Load Balancer or Azure Application Gateway).

**Reasons for this change:**

1. Azure instance health does not require a load balancer -- the health probe can exist on the health check
2. If you are using client-side load balancing (e.g. through Eureka), you do not need a load balancer.

**Gotchas with this change:**

1. The spin-deck UI does not support null load balancers, so changes have to be made through JSON / pipeline templates.
    1. This is consistent with how scale set health probes work; there is nowhere to specify this in deck. I am planning to add null load balancer functionality to deck, but this change can be considered separate, as it will work with JSON editing and pipeline templates.
1. There is no concept of "disabled" for server groups that are not fronted by a load balancer because "disabled" means "no traffic is being sent to these instances". Without a load balancer, nothing can control that. To mitigate that, I chose to mark the server group as "disabled" if its instance count is 0.
    1.  Additionally, deployment strategies do not appear to be implemented for Azure (at least, they do not work on 1.26.2). Therefore, I just destroy the old server group after deployment. I am planning to come back in and contribute a fix for deployment strategies, at which time I will remove the "destroy" logic from "create server group without load balancer" and replace it with logic based on the selected deployment strategy.
1. The "default" load balancer type is Azure Application Gateway, so there were a lot of `else` blocks that followed logic that was meant for Azure Application Gateways. I believe I caught all of these, added a `null` branch, and added an exception for 

**Testing done:**

1. Creating (via deployment), disabling, and destroying server groups in Azure
2. Unit testing via `AzureServerGroupResourceTemplateSpec`


kmccarp:
> We prefer that non-test backend code be written in Java or Kotlin, rather than Groovy. The following files have been added and written in Groovy:
> 
> * clouddriver-azure/src/main/java/com/netflix/spinnaker/clouddriver/azure/resources/servergroup/ops/CreateAzureServerGroupWithoutLoadBalancersAtomicOperation.groovy
> 
> See our server-side [commit conventions here](https://www.spinnaker.io/community/contributing/back-end-code/#choice-of-language).

That link gives me a 404, can a reviewer send an updated link?

PR #5500, opened 2021-08-31
feat(aws): enable custom property binding capability for aws accounts

apoorvmahajan:
This PR allows custom property binding support for loading aws accounts from a config file. This is turned off by default, which means that the accounts will be loaded via the default way, which is relying on spring boot to do the binding. 

This is enabled by setting `aws.custom-property-binding-enabled: true`.

This is for the following issue: https://github.com/spinnaker/spinnaker/issues/6492


dbyron-sf (approved):

deverton (approved):

dbyron-sf:
Thanks for the review @deverton !

PR #5504, opened 2021-09-02
fix(gce) : modified internal load balancer services to fix duplicatio…

ravichander82:
This code change is to address the GCE issues [#6136](https://github.com/spinnaker/spinnaker/issues/6136) [#6253](https://github.com/spinnaker/spinnaker/issues/6253)

With modification of of definition of 'ilbServices' both issues are resolved by removing the null values and duplication,
this way deploy stage will be complete in case of both missing 'backend-service' as parameter and when by manually adding valid 'region-backend-service' as parameter


skandragon (approved):
LGTM

kirangodishala:

CB-GuangyaoXie:
Can we backport this?

PR #5505, opened 2021-09-02
feat(artifacts): support tokenFile for GitRepoArtifactCredentials and reload auth headers

zachsmith1:
Added support for tokenAuth with gitRepoArtifactCredentials and made BasicHttpArtifactCredentials reload auth headers for every call instead of during object instantiation.

### DIFF
--- a/clouddriver-artifacts/src/test/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactAccountTest.java
+++ b/clouddriver-artifacts/src/test/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactAccountTest.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2021 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


clouddriver-artifacts/src/test/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactAccountTest.java
```diff
@@ -0,0 +1,52 @@
...
+        GitRepoArtifactAccount.builder()
+            .name("gitRepo-account")
+            .tokenFile(authFile.toAbsolutePath().toString())
+            .build();
+
+    assertThat(account.getTokenAsString().get()).isEqualTo("zzz");
+  }
+
+  @Test
+  void shouldGetTokenFromProperty() {
```
german-muzquiz, line 46: I think we need to add protection in case both `token` and `tokenFile` are specified, like throwing an exception at startup during beans creation to correct the config to only specify one property but not both.

german-muzquiz (commented):
Looks great, I just have a small suggestion

clouddriver-artifacts/src/test/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactAccountTest.java
zachsmith1: While I agree with you -- we would need to add that logic in a few different places because each of the 'account' definitions handles it individually. Maybe something to note for the future changes?

german-muzquiz (approved):

zachsmith1:
backport release-1.26.x release-1.27.x

zachsmith1:
@Mergifyio backport release-1.27.x release-1.26.x

PR #5508, opened 2021-09-02
fix(cloudfoundry): add configurable cache expiry for clients

mattgogerly:
Currently Clouddriver maintains an in-memory cache of GUIDs to Server Groups. If caching is handled by a specific pod (e.g. if cache sharding is enabled, or HA is enabled) and a different pod handles an operation that populates the cache, then this cache will never be updated.

If an application is deleted outside of Spinnaker and then redeployed with the same name, the app GUID will change. This results in Clouddriver trying to perform actions against the old GUID which no longer exists.

I've added a default expires-after-write for the Applications cache, but have made this configurable alongside an expires-after-access rule. Whilst I was at it I did the same for the Routes cache.

Also threw in some minor cleanup in files whilst I was there, making things final, using lambdas etc.


mattgogerly:
@zachsmith1 Not sure how best to demonstrate the issue as it's more a logical issue than a code one. Any ideas let me know :)

link108:

PR #5509, opened 2021-09-02
fix(provider/azure): Fix isServerGroupDisabled for app gateway

kmccarp:
Caused by this PR: https://github.com/spinnaker/clouddriver/pull/5494

This usage of `isServerGroupDisabled` (which was renamed) was missed.


jkschneider (approved):
LGTM

PR #5510, opened 2021-09-02
feat(provider/azure): Use v2-sku for Application Gateways

kmccarp:
In order to be able to store TLS certificates in key-vault and use them on an application gateway, you are required to use a Standard_V2 Application Gateway.

This change, coupled with a change in deck (forthcoming), will allow users to select between Standard_Small (v1) and Standard_V2 SKUs. This change is safe on its own, but not beneficial by itself.

**Summary of changes**

1. Use input `sku` for App Gateway sku
    1. When updating existing App Gateway, keep the same sku
2. Select tier based on input sku (`Standard` for `Standard_Small` sku and `Standard_v2` for `Standard_v2` sku)
3. Select sku of public IP based on SKU of App Gateway
4. Update `apiVersion` to `2018-04-01` to allow v2 App Gateway sku

**Testing Done**

1. Unit tests
2. Tested (together with changes in deck) the following:
    1. Modify existing v1-sku App Gateways
    2. Create new v1-sku App Gateways
    3. Create new v2-sku App Gateways
    4. Delete v1-sku App Gateways
    5. Delete v2-sku App Gateways
    6. Deploy to v1-sku App Gateway
    6. Deploy to v2-sku App Gateway


jkschneider (approved):
LGTM

PR #5513, opened 2021-09-08
fix(kubernetes/artifacts): Support for legacy Clouddriver's artifact replacement logic

cristhian-castaneda:
This change adds support for this breaking change https://github.com/spinnaker/clouddriver/pull/4874.

We added an optional flag to enable the old behavior

```
kubernetes:
  artifact-binding:
    docker-image: match-name-only
```

This is mainly to help current users that rely on the old behavior.


german-muzquiz (commented):
I see a potential race condition where this new flag may not work depending on the order in which spring creates the beans: if any spring bean that imports `Replacer` is created before `ReplacerFactory` bean is created, when the runtime processes the imports it will invoke the static method of `ReplacerFactory`, which hasn't yet read the value of the config file and uses the default of `false`.

This problem is because `Replacer` is a bunch of static objects and the config flag can only be read from instance fields.

Another approach would be to move using the config flag until the filter is applied rather than when creating the replacers. This means that `Replacer` would need to have two instances of `replacePathSupplier`: one standard and one legacy, but which one is used is determined until the replace call. There's a single point of usage in `ArtifactReplacer`, so we can add the boolean flag to the `ArtifactReplacer` constructor and populate it from the nearest spring bean that can load it.

Another comment is to change the config setting from `artifacts.legacy-binding.enabled` to `kubernetes.artifact-binding.docker-image`, with possible values `match-name-and-tag` (default) and `match-name-only` (which is the legacy behavior). In this way we're prepared if we need to add more settings related with how artifact binding works, for example if `kubernetes.artifact-binding.enabled` is ever implemented globally.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/KubernetesHandler.java
```diff
@@ -45,10 +45,20 @@
 
   private final ArtifactReplacer artifactReplacer;
 
+  protected String dockerImageBinding;
+
   protected KubernetesHandler() {
     this.artifactReplacer = new ArtifactReplacer(artifactReplacers());
   }
 
+  protected void setDockerImageBinding(String dockerImageBinding) {
```
cristhian-castaneda, line 54: This method was defined so the concrete class can assign the value from config properties.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/KubernetesHandler.java
```diff
@@ -45,10 +45,20 @@
 
   private final ArtifactReplacer artifactReplacer;
 
+  protected String dockerImageBinding;
```
german-muzquiz, line 48: Have you tried to inject the value here with `@Value` instead of injecting in every subclass? In theory it should work

cristhian-castaneda: Yeah, you are right, it works, I did some test with this in my first approach but didn't work, never tried that in this approach.

german-muzquiz (approved):

cristhian-castaneda:

link108:
@Mergifyio backport release-1.26.x release-1.27.x

PR #5515, opened 2021-09-08
feat(k8s): provide an option to bypass getting namespaces when adding new k8s accounts

apoorvmahajan:
## Purpose
- This PR implements the first fix from this issue: https://github.com/spinnaker/spinnaker/issues/6528. 
- We have more than 2000+ k8s accounts in our setup. When clouddriver starts up, it attempts to load all the declared namespaces for each of these accounts by making a live API call to the cluster. In our setup, we have observed this action to take 35-40 mins, which is how much of a benefit it is for us to disable it.
- Since users may be relying on this capability, and would not like this to be disabled by default, we have made this an opt-in capability. By default, these namespaces will be loaded when credentials are added. If someone wants to opt out of it, they can set 
`kubernetes.loadNamespacesInAccount: false`

Note: we only provide this option when a credential/account is added. If that account needs to be updated, the calls to load namespaces will be made - there is no option provided in this PR to bypass those. The reason why we are not changing it there is because we are only concerned with making clouddriver start up faster, and adding that in the credentialsAdded() works well in that regard.

## Changes:
- provides an option to skip getting declared namespaces when adding k8s accounts

### DIFF
--- a/clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesCredentialsLifecycleHandlerTest.java
+++ b/clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesCredentialsLifecycleHandlerTest.java
@@ -19,49 +19,79 @@ package com.netflix.spinnaker.clouddriver.kubernetes.security;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import com.google.common.collect.ImmutableList;
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.KubernetesProvider;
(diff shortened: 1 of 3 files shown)


clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesCredentialsLifecycleHandlerTest.java
```diff
@@ -19,49 +19,79 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.junit.platform.runner.JUnitPlatform;
 import org.junit.runner.RunWith;
 import org.mockito.ArgumentMatchers;
-import org.mockito.Mockito;
 
 @RunWith(JUnitPlatform.class)
 public class KubernetesCredentialsLifecycleHandlerTest {
+  KubernetesProvider provider = new KubernetesProvider();
```
dbyron-sf, line 45: If we're doing this in setup below, do we need it here too?

apoorvmahajan: no, I'll fix this

dbyron-sf (approved):

PR #5518, opened 2021-09-09
fix(clouddriver-lambda): flakey test

xibz:
buildCacheData uses parallel stream, so passing in anything other than a
concurrent hashmap results in race conditions. In this case, the test
was passing in a plain hashmap when it should have been a concurrent
hashmap

Signed-off-by: benjamin_j_powell <benjamin_j_powell@apple.com>


zachsmith1 (approved):

xibz:
@zachsmith1 - Is there anything else you need from me? Or is this good to be merged?

xibz:
Awesome thank you!

PR #5542, opened 2021-09-27
test(kubernetes/delete): add test coverage for deleting crd and custo…

dbyron-sf:
…m resources

Most of this is for functionality added by
https://github.com/spinnaker/clouddriver/pull/5334.  It was possible to delete
CustomResourceDefinitions before that, though not custom resources.


apoorvmahajan (commented):
lgtm. Thanks for adding these tests!

german-muzquiz (approved):

PR #5547, opened 2021-09-30
feature(gce):to extend memory feature , a boolean flag is introduced …

ravichander82:
The code change is to implement the 'extendMemory' feature addressing [4773](https://github.com/spinnaker/spinnaker/issues/4773) 

It requires both frontend and backend changes 

This PR is to address backend changes 
Another PR raised is for frontend changes

With this feature change users can give memory range as per their wish and the validation is done by GCP console


skandragon:
LGTM

PR #5553, opened 2021-10-07
feat(aws/acm): Add caching agent for AWS Certificate Manager

jaydee864:
Spinnaker Managed Delivery for EC2 depends on Clouddriver to list the available SSL certificates in each AWS account. If a certificate is not in the Clouddriver cache, then Managed Delivery cannot use that certificate.

Currently, Clouddriver only caches IAM certificates, using [AmazonCerificateCachingAgent](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonCertificateCachingAgent.groovy). This pull request seeks to enable Clouddriver to cache certificates stored in AWS Certificate Manager as well.

I have tested this change on our development Spinnaker environment, and after deploying I have verified that Clouddriver's `/certificates/aws` API returned both ACM and IAM certificates. I was also able to use an ACM certificate in my managed delivery configuration; I set the `certificate:` value under the listener config to the domain name of the certificate I wanted to use.

Spinnaker's [contributing guidelines](https://spinnaker.io/docs/community/contributing/code/back-end-code/) recommend avoiding creating new classes in Groovy, but in this case I don't think it's really feasible to do so. The AmazonCertificate class in the Groovy code uses the Canonical annotation to generate its constructors, and there is [an issue](https://stackoverflow.com/a/23940160/13203194) at compile-time when referencing such a constructor from Java code in the same module. There is a proposed fix for a Maven project in the linked StackOverflow question, but I don't know if that's applicable to Gradle and I'm also not confident in making that change to Clouddriver's configuration without breaking other things.

### DIFF
--- a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AwsCertificateManagerCachingAgent.groovy
+++ b/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AwsCertificateManagerCachingAgent.groovy
@@ -1,0 +1,184 @@
+/*
+ * Copyright 2021 Expedia, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License
+ * for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.aws.provider.agent
+
+import com.amazonaws.services.certificatemanager.AWSCertificateManager
+import com.amazonaws.services.certificatemanager.model.AWSCertificateManagerException
+import com.amazonaws.services.certificatemanager.model.CertificateDetail
+import com.amazonaws.services.certificatemanager.model.CertificateSummary
+import com.amazonaws.services.certificatemanager.model.DescribeCertificateRequest
+import com.amazonaws.services.certificatemanager.model.DescribeCertificateResult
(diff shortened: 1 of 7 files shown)


JonGilmore:
@luispollo wondering if you might be able to get some eyes on this PR? We're facing the same issue in our environment as well.

luispollo:
I'll have to defer to @robfletcher and/or @emjburns on this one. My knowledge of clouddriver is pretty limited.

dmart:
Any chance we can get some eyes on this?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AwsCertificateManagerCachingAgent.groovy
```diff
@@ -0,0 +1,184 @@
...
+  final NetflixAmazonCredentials account
+  final String region
+  final ObjectMapper objectMapper
+  final Registry registry
+  final Id securityTokenExceptionGauge
+
+  protected Instant lastFailure
+
+  static final Set<AgentDataType> types = Collections.unmodifiableSet([
+    AUTHORITATIVE.forType(CERTIFICATES.ns)
```
deverton, line 63: So looking at https://github.com/spinnaker/clouddriver/blob/f3b393402c603f5b19f7cc10ac88d137269b702c/clouddriver-api/src/main/java/com/netflix/spinnaker/cats/agent/AgentDataType.java#L24 it seems like if this is listing itself as `AUTHORITATIVE` then it's going to fight with https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonCertificateCachingAgent.groovy over which agent has the complete list.

I think it might be better to add the reading from Certificate Manager to the `AmazonCertificateCachingAgent` so that there's only one agent.

deverton (changes requested):

PR #5554, opened 2021-10-08
fix(provider/kubernetes): fix cron job image replacer (#6132)

pemmasanikrishna:
Fixes the issue https://github.com/spinnaker/spinnaker/issues/6132


pemmasanikrishna:
@german-muzquiz , I have added the Integration test case, couldn't verify it locally. Looks like the test cases can only be run on a non-Windows machine.

pemmasanikrishna:
@german-muzquiz  looks like this is an intermittent issue. can we retrigger the build

PR #5557, opened 2021-10-18
fix(namespaceValidation): revert #5359 which caused a new namespace deploy to fail

zachsmith1:
Reverting this because it causes a new namespace deployment to fail. Ideally, we address both use-cases in a future PR but reverting for now.


zachsmith1:

PR #5563, opened 2021-10-28
fix(gce) : duplication of firewall rules when using shared vpc network

ravichander82:
The code change is to address the issue of duplication of securityGroups  when using a SharedVPC network [#4373](https://github.com/spinnaker/spinnaker/issues/4373)

In case of using a Shared VPC Network the firewall rules of service project are adding the firewalls of host project because of which there is duplication of securityGroups

With this modification only the firewalls of existing project will be populated and resolves the issue

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleSecurityGroupCachingAgent.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/provider/agent/GoogleSecurityGroupCachingAgent.groovy
@@ -206,12 +206,15 @@ class GoogleSecurityGroupCachingAgent extends AbstractGoogleCachingAgent impleme
       List<Firewall> firewalls = timeExecute(compute.firewalls().list(project),
                                              "compute.firewalls.list", TAG_SCOPE, SCOPE_GLOBAL).items as List
 
-      if (xpnHostProject) {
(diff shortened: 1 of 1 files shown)


ravichander82:
@skandragon , @plumpy can you please review the code changes

mattgogerly:
Are there any tests that could be written for this? It's not clear what the side effects could be.

ravichander82:
Hi @mattgogerly, 

The definition of firewalls is modified so that only the existing project (service project) firewalls will be returned and the firewall rules of host project will not be added, in this way duplication is resolved

If you ask about side effects or any risks involved, apart from build checks and existing test cases nothing else I could find.

In GCP console when firewalls are viewed for a project ,only which are present in the project are viewed so I took it as benchmark and modified the definition  so that both are consistent.

mattgogerly:
That sounds ok to me, I'm just not very familiar with GCloud. Is there a possibility that rules in the "host account" are not present in the child? Would these rules affect the child? (I.e. would there be rules affecting the app that wouldn't be surfaced to Spinnaker?)

ravichander82:
> That sounds ok to me, I'm just not very familiar with GCloud. Is there a possibility that rules in the host project are not present in the child? Would these rules affect the child? (I.e. would there be rules affecting the app that wouldn't be surfaced to Spinnaker?)


Firewall rules being present in host project and not present in service project is possible 

It depends on the definition of firewalls

skandragon (approved):
LGTM

PR #5564, opened 2021-10-29
fix(cloudfoundry): Avoid NPE when no instances are present.

cristhian-castaneda:
Avoid an NPE in orca when no instances are present

Image 2021-10-28 at 7 28 55 PM


cristhian-castaneda:
I just added a  test for a ServerGroup without instances, the issue itself it's produced in orca, in this line https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/servergroup/WaitForRequiredInstancesDownTask.groovy#L46
The test is intended to ensure that at lest an empty list of instances is returned.

zachsmith1 (approved):
LGTM

zachsmith1:

PR #5567, opened 2021-11-02
fix(cloudfoundry): add timeout to converter

zachsmith1:
Manifest attributes are unfortunately set/transferred between two different manifest attributes classes. Timeout was missing from the converter class. This code should ultimately be changed to leverage a single class for representation.


cristhian-castaneda (approved):
LGTM

dogonthehorizon (approved):

zachsmith1:
@Mergifyio backport release-1.27.x release-1.26.x

PR #5570, opened 2021-11-03
fix(cloudfoundry): ensure process health is built

zachsmith1:
The process health object should be built with any of the parameters submitted. Previously it would only be built if the health check type was not null. This would ignore any further fields. Now, it will try to create an object if any of the fields related to health are present. This could result in a runtime issue with cloudfoundry, but that issue should boil up to the user at that point.


dogonthehorizon (approved):

zachsmith1:

PR #5572, opened 2021-11-03
fix(cloudfoundry): add test for non-empty fields

cristhian-castaneda:


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/model/v3/Process.java
```diff
@@ -40,9 +41,7 @@ private HealthCheck() {}
 
     @Nullable private String type;
 
-    @JsonInclude(value = JsonInclude.Include.CUSTOM, valueFilter = HealthCheckData.class)
```
zachsmith1, line 43: wait why did we remove this?

zachsmith1 (approved):

PR #5576, opened 2021-11-05
fix(artifacts): close leaking input stream when downloading artifacts

dbyron-sf:
At least for s3, this can lead to connection pool exhaustion.  See https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/S3Object.html#getObjectContent-- which says:

> If you retrieve an S3Object, you should close this input stream as soon as possible, because the object contents aren't buffered in memory and stream directly from Amazon S3. Further, failure to close this stream can cause the request pool to become blocked.


mattgogerly (approved):
LGTM

german-muzquiz (approved):
Nice catch

PR #5577, opened 2021-11-07
fix(cloudfoundry): skip API calls for apps that won't be cached

mattgogerly:
Moving the "should cache" checks prior to the call to map so that unnecessary API calls aren't made for apps that won't be cached anyway.


zachsmith1:
@mattgogerly this LGTM but ideally we get some kind of test. Would this be easily testable?

mattgogerly:
> @mattgogerly this LGTM but ideally we get some kind of test. Would this be easily testable?

Will give it a go. The tests for this method are quite sparse so might have to come up with something.

mattgogerly:
@zachsmith1 added some verify() calls to the test from #5427

zachsmith1 (approved):

zachsmith1:

PR #5589, opened 2021-12-03
fix(clouddriver-sql): add missing com.fasterxml.jackson.core:jackson-annotations and com.fasterxml.jackson.core:jackson-databind dependencies

madhubkk:
When we are using spring boot 2.2.x (existing kork) we did not face below error, and as we move for spring boot 2.3.x we face error as given below:
```
> Task :clouddriver-sql:compileKotlin FAILED
e: /clouddriver/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/SqlTaskRepository.kt: (18, 30): Unresolved reference: databind
e: /clouddriver/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/SqlTaskRepository.kt: (40, 23): Unresolved reference: ObjectMapper
e: /clouddriver/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/TaskMapper.kt: (18, 30): Unresolved reference: core
e: /clouddriver/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/TaskMapper.kt: (19, 30): Unresolved reference: databind
e: /clouddriver/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/TaskMapper.kt: (32, 23): Unresolved reference: ObjectMapper
e: /clouddriver/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/TaskMapper.kt: (38, 42): Unresolved reference: TypeReference
```
After introducing explicit implementation of **com.fasterxml.jackson.core:jackson-annotations** and **com.fasterxml.jackson.core:jackson-databind** in clouddriver-sql.gradle, build got success.

### DIFF
--- a/clouddriver-sql/clouddriver-sql.gradle
+++ b/clouddriver-sql/clouddriver-sql.gradle
@@ -29,6 +29,8 @@ dependencies {
   implementation "de.huxhorn.sulky:de.huxhorn.sulky.ulid"
   implementation "org.jooq:jooq"
   implementation "org.hibernate.validator:hibernate-validator"
+  implementation "com.fasterxml.jackson.core:jackson-annotations"
+  implementation "com.fasterxml.jackson.core:jackson-databind"
 
   testImplementation project(":clouddriver-core-tck")
 


mattgogerly:
Is this error going to be present in other services that use SQL (Orca, Echo, Front50, Fiat)?

If so it might make sense to have these dependencies in Kork rather than adding them to each repo.

madhubkk:
> Is this error going to be present in other services that use SQL (Orca, Echo, Front50, Fiat)?
> 
> If so it might make sense to have these dependencies in Kork rather than adding them to each repo.

This error is present in only clouddriver and not in any component so I did not  do it in kork.

j-sandy (approved):
This PR is raised as part of identifying potential issues, expected to face while upgrading spring boot from 2.2.13 (existing) to 2.3.12 and to fix them. This PR will have no impact on current springboot version however will rectify an upcoming issue with springboot 2.3.12. It is raised after verifying the effect of upgrading Kork with Springboot version [here](https://github.com/j-sandy/kork/commit/3ad602de5b346cce4abe51df43f9d1ecf51b552a) on all the components.

dbyron-sf (approved):

PR #5592, opened 2021-12-07
fix(build): publish to maven central, switch groupId to io.spinnaker.…

edgarulg:
There is an issue in the Plugins Framework that was solved in this PR: spinnaker/kork#866

but this fix was released in kork 7.110.0

I bumped the version of kork to include the fix in Gate but this involved changing the groupId of kork because now it's been published under "io.spinnaker" and this release branch is pretty old.

I had to bump the version of Fiat as well because it has issues when trying to download the kork version in fiat due to the groupId change.


dbyron-sf (commented):
LGTM but would love @link108 to have a look as well.

link108:
LGTM, thanks for making these updates!

PR #5586, opened 2021-11-30
fix(elasticsearch): Unpinning elasticsearch package and removing deprecated Elasticsearch transport and Jest library

j-sandy:
In Spring Boot 2.3, native elasticsearch transport has been removed as both elasticsearch and spring data won’t support in upcoming releases. Support for the Jest library has also been removed. Please refer https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.3-Release-Notes

This fix is required for upgrade activity of spring boot from 2.2.x to 2.3.x

PR #5594, opened 2021-12-08
feat(core): Add experimental account storage API

jvz:
This adds an experimental API for storing and loading account
credentials definitions from an external durable store such as a SQL
database. Secrets can be referenced through the existing Kork
SecretEngine API which will be fetched on demand. Initial support is for
Kubernetes accounts given the lack of existing Kubernetes cluster
federation standards compared to other cloud providers, though this API
is made generic to allow for other cloud provider APIs to participate in
this system.

https://github.com/spinnaker/spinnaker/issues/6525


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy
```diff
@@ -54,8 +53,10 @@ import java.security.Security
   GroovyTemplateAutoConfiguration,
   GsonAutoConfiguration,
   DataSourceAutoConfiguration,
-  ElasticsearchDataAutoConfiguration,
-  ElasticSearchRestHealthContributorAutoConfiguration
+  ElasticSearchJestHealthContributorAutoConfiguration,
```
dbyron-sf, line 56: Curious why there are any elasticsearch-related changes here...but this line in particular appears to be causing a compile error [here](https://github.com/spinnaker/clouddriver/runs/4514289644?check_suite_focus=true):
```
> Task :clouddriver-web:compileGroovy FAILED
/home/runner/work/clouddriver/clouddriver/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy: 56: Only classes and closures can be used for attribute 'exclude' in @org.springframework.boot.autoconfigure.EnableAutoConfiguration
 @ line 56, column 3.
     ElasticSearchJestHealthContributorAutoConfiguration,
     ^
```
In case it's relevant, https://github.com/spinnaker/clouddriver/pull/5586 went in recently with some nearby changes.

jvz: This was some sort of change to make things compilable back when I first updated this. That class might also need to be moved to a string? Or a rebase here might be relevant, yeah.

clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionRepository.java
```diff
@@ -0,0 +1,53 @@
...
+import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/** Provides CRUD persistence operations for account {@link CredentialsDefinition} instances. */
+@Alpha
+@NonnullByDefault
+public interface AccountDefinitionRepository {
+  @Nullable
```
dbyron-sf, line 30: javadoc comments please

dbyron-sf:
Hey @jvz I'm probably already supposed to know this, but does this PR enable dynamic refresh of the accounts, or are they still loaded only at startup?

jvz:
Refreshing is done via credentials polling currently. If you enable that, then this gets refreshed along with any other credentials definition sources you have.

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionSource.java
```diff
@@ -0,0 +1,71 @@
...
+        () ->
+            repository.listByType(typeName).stream().map(type::cast).collect(Collectors.toList()));
+    sources.addAll(additionalSources);
+    this.sources = List.copyOf(sources);
+  }
+
+  @Override
+  public List<T> getCredentialsDefinitions() {
+    return sources.stream()
+        .flatMap(source -> source.getCredentialsDefinitions().stream())
```
german-muzquiz, line 68: What is taking precedence if there are two accounts with the same name in different sources?

jvz: Right now, it would be based on the ordering of bean definitions which isn't that ideal. Any suggestions on conflict strategies to apply here? Like throwing an exception or logging a warning or what?

dbyron-sf: Part of me wants to allow only one source, but then I suspect it will be helpful in some situations to have some "bootstrap" accounts in e.g. clouddriver-local.yaml and then have more this way....and there isn't really a precedence issue unless more than one source defines an account with the same name.  And I guess it makes sense for clouddriver-local.yaml to take precedence...but with dynamic refresh also possible via spring cloud config I'm not sure that's possible/realistic/a good idea....

Does it make sense to add "if account name already exists, then log a warning, skip this account, but continue processing other accounts" logic to "all the places" ?  Or at least here?

jvz: I think logging a warning that the account definition will be skipped as a duplicate.

dbyron-sf: Great.  Is putting it here enough, or are there other places that are now subject to this issue?

jvz: This is where the bean definitions would all be merged for a specific account type. It's reusable for any other credentials classes that opt in to the system.

jvz: For context, I initially developed this to merge in existing sources because I have another custom source for Kubernetes credentials used on some clusters involving a CRD to configure these accounts. Combined with a naming scheme, I didn't expect to see any conflicts, though it's definitely a good idea to warn about them.

jvz:
Alright, I've updated based on most of the feedback so far. I need to dig a little deeper into the SQL implementation to update the deletion behavior documented in https://github.com/spinnaker/gate/pull/1494

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionSource.java
jvz: Updated with logs and filtering dupes.

clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/security/SqlAccountDefinitionRepository.kt
```diff
@@ -0,0 +1,216 @@
...
+          .set(versionColumn, version)
+          .execute()
+      }
+    }
+  }
+
+  override fun delete(name: String) {
+    withPool(poolName) {
+      jooq.transactional { ctx ->
+        ctx.update(accountsTable)
```
jvz, line 184: - [ ] Change deletion behavior to use `DELETE FROM` rather than updating a deletion flag.

clouddriver-sql/src/main/resources/db/changelog/20210927-accounts.yml
```diff
@@ -0,0 +1,146 @@
...
+            type: bigint
+            constraints:
+              nullable: false
+        - column:
+            name: last_modified_by
+            type: varchar(255)
+            constraints:
+              nullable: false
+        - column:
+            name: is_deleted
```
jvz, line 44: - [ ] Remove or move this column to history table instead.

clouddriver-sql/src/main/resources/db/changelog/20210927-accounts.yml
```diff
@@ -0,0 +1,146 @@
+databaseChangeLog:
+- changeSet:
+    id: create-accounts-table
+    author: msicker
+    validCheckSum:
+      # Original changeset checksum
+      - 8:af38865f77cc07c876e92b843be60fd2
```
jvz, line 7: - [ ] Remove changeset checksum; I can nuke the tables in my dev environment with updated version

jvz:
I've updated the deletion behavior and made the revision history a little better. I've also removed some unused endpoints in the API.

jvz:
That's what I expected; just wanted to make sure I finished the PR updates before the same new year before I forgot about them!

jvz:
Oh, also helps to push this to my fork and not just a private repo. Pushed!

dbyron-sf:
@jvz I'm still struggling to catch up from the flood of messages after being off for awhile...this is still very much on my radar though...

jvz:
No worries! I just got back from holidays, so I made sure to leave pings where applicable.

german-muzquiz (commented):
Is it correct to assume that for this to work, `clouddriver.yml` needs to enable the poller?
```
credentials:
  poller:
    enabled: true
    types:
      kubernetes:
        reloadFrequencyMs: 60000

File: clouddriver-sql/clouddriver-sql.gradle [hunk 7]
```diff
@@ -24,13 +24,15 @@ dependencies {
...
   implementation "io.spinnaker.kork:kork-core"
+  implementation "io.spinnaker.kork:kork-secrets"
   implementation "io.spinnaker.kork:kork-sql"
   implementation "io.spinnaker.kork:kork-telemetry"
   implementation "de.huxhorn.sulky:de.huxhorn.sulky.ulid"
   implementation "org.jooq:jooq"
   implementation "org.hibernate.validator:hibernate-validator"
   implementation "com.fasterxml.jackson.core:jackson-annotations"
   implementation "com.fasterxml.jackson.core:jackson-databind"
+  implementation "org.springframework.boot:spring-boot-starter-web"
```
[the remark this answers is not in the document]
jvz: Good catch; I think this is a relic from when I was configuring the Jackson type mapping in this module. That's since been moved.

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/config/AccountDefinitionConfiguration.java
```diff
@@ -0,0 +1,138 @@
...
+        .filter(Objects::nonNull)
+        .map(AccountDefinitionConfiguration::loadCredentialsDefinitionType)
+        .filter(type -> type.isAnnotationPresent(JsonTypeName.class))
+        .toArray(Class[]::new);
+  }
+
+  private static Class<? extends CredentialsDefinition> loadCredentialsDefinitionType(
+      String className) {
+    try {
+      return ClassUtils.forName(className, null).asSubclass(CredentialsDefinition.class);
```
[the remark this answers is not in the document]
jvz: The `ClassPathScanningCandidateComponentProvider` already has a filter to ensure classes implement `CredentialsDefinition`.

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/config/AccountDefinitionConfiguration.java
```diff
@@ -0,0 +1,138 @@
...
+    return new AccountDefinitionModule(findAccountDefinitionTypes(loader, properties));
+  }
+
+  private static StringDeserializer createSecretDecryptingDeserializer(
+      SecretManager secretManager) {
+    return new StringDeserializer() {
+      @Override
+      public String deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
+        var string = super.deserialize(p, ctxt);
+        return EncryptedSecret.isEncryptedSecret(string) ? secretManager.decrypt(string) : string;
```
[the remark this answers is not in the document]
jvz: Updated to support the `EncryptedSecret.isEncryptedFile()` use case as well.

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/config/AccountDefinitionConfiguration.java
```diff
@@ -0,0 +1,138 @@
...
+import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder;
+import org.springframework.util.ClassUtils;
+
+/**
+ * Provides configuration settings related to managing account credential definitions at runtime.
+ *
+ * @see Properties
+ */
+@Configuration
+@EnableConfigurationProperties(AccountDefinitionConfiguration.Properties.class)
```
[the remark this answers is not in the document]
jvz: Updated.

clouddriver-sql/clouddriver-sql.gradle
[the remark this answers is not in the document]
jvz: Fixed.

clouddriver-web/src/main/java/com/netflix/spinnaker/clouddriver/controllers/AccountController.java
```diff
@@ -0,0 +1,88 @@
...
+@RestController
+@ConditionalOnProperty("account.storage.enabled")
+public class AccountController {
+  private final AccountDefinitionRepository repository;
+
+  public AccountController(AccountDefinitionRepository repository) {
+    this.repository = repository;
+  }
+
+  @GetMapping("/account/{accountName}")
```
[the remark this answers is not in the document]
jvz: Done.

clouddriver-web/src/main/java/com/netflix/spinnaker/clouddriver/controllers/AccountController.java
```diff
@@ -0,0 +1,88 @@
...
+  @PreAuthorize("hasPermission(#accountName, 'ACCOUNT', 'READ')")
+  public CredentialsDefinition getAccountByName(@PathVariable String accountName) {
+    var definition = repository.getByName(accountName);
+    if (definition == null) {
+      throw new NotFoundException(String.format("No account found with name `%s`", accountName));
+    }
+    return definition;
+  }
+
+  @GetMapping("/accounts/{accountType}")
```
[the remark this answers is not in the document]
jvz: Done.

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/config/AccountDefinitionConfiguration.java
```diff
@@ -0,0 +1,148 @@
...
+
+  private static StringDeserializer createSecretDecryptingDeserializer(
+      SecretManager secretManager) {
+    return new StringDeserializer() {
+      @Override
+      public String deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
+        var string = super.deserialize(p, ctxt);
+        if (EncryptedSecret.isEncryptedSecret(string)) {
+          return EncryptedSecret.isEncryptedFile(string)
+              ? secretManager.decryptAsFile(string).toString()
```
jvz, line 92: Should this add a `file:` prefix, too?

clouddriver-web/src/main/java/com/netflix/spinnaker/clouddriver/controllers/AccountController.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  @DeleteMapping("/{accountName}")
+  @PreAuthorize("hasPermission(#accountName, 'ACCOUNT', 'WRITE')")
+  public void deleteAccount(@PathVariable String accountName) {
+    repository.delete(accountName);
+  }
+
+  @GetMapping("/{accountName}/history")
+  @PreAuthorize("hasPermission(#accountName, 'ACCOUNT', 'READ')")
```
jvz, line 75: This should probably be changed to WRITE permissions before being exposed in Gate.

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/config/AccountDefinitionConfiguration.java
german-muzquiz: Apparently that's only used for SAML related files that may be used in gate or fiat, so I'd say this is fine without the `file:` prefix.

german-muzquiz (approved):

clouddriver-web/src/main/java/com/netflix/spinnaker/clouddriver/controllers/CredentialsController.java
```diff
@@ -0,0 +1,161 @@
...
+            .contains(credentials.getEnvironment()));
+    cred.put(
+        "primaryAccount",
+        credentialsConfiguration.getPrimaryAccountTypes().contains(credentials.getAccountType()));
+    return cred;
+  }
+
+  @GetMapping("/type/{accountType}")
+  @PostFilter("hasPermission(filterObject.name, 'ACCOUNT', 'READ')")
+  public List<? extends CredentialsDefinition> listAccountsByType(
```
dbyron-sf, line 116: Possible to mark these new endpoints as Alpha?

dbyron-sf (approved):

PR #5604, opened 2021-12-15
feat(artifacts/bitbuket): add token auth credentials support

juanpabloprado:
Add token auth credentials support

* Implements TokenAuth interface in BitBucketArtifactAccount
* Update on BitbucketArtifactCredentials (similar to GitHubArtifactCredentials)

When using a valid token, after invalid token was used
Screen Shot 2022-01-10 at 11 09 14

Swap token inside tokenFile
swap-token

### DIFF
--- a/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactAccount.java
+++ b/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactAccount.java
@@ -20,6 +20,7 @@ package com.netflix.spinnaker.clouddriver.artifacts.bitbucket;
 import com.google.common.base.Strings;
 import com.netflix.spinnaker.clouddriver.artifacts.config.ArtifactAccount;
 import com.netflix.spinnaker.clouddriver.artifacts.config.BasicAuth;
+import com.netflix.spinnaker.clouddriver.artifacts.config.TokenAuth;
 import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
 import java.util.Optional;
 import javax.annotation.ParametersAreNullableByDefault;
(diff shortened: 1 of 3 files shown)


clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactAccount.java
```diff
@@ -29,20 +30,29 @@
 
 @NonnullByDefault
 @Value
-public class BitbucketArtifactAccount implements ArtifactAccount, BasicAuth {
-  private final String name;
```
juanpabloprado, line 33: Screen Shot 2021-12-15 at 9 51 50
Screen Shot 2021-12-15 at 9 51 58

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactAccount.java
```diff
@@ -29,20 +30,29 @@
...
   @Builder
   @ConstructorBinding
   @ParametersAreNullableByDefault
   BitbucketArtifactAccount(
-      String name, String username, String password, String usernamePasswordFile) {
+      String name,
+      String username,
+      String password,
+      String usernamePasswordFile,
+      String token,
```
jasonmcintosh, line 49: A minor concern - we're changing the initialization to include parameters.  Anyone extending this would get broken method signatures.  I'd use constructor overloading to reduce any risk vs. completely changing constructor signatures... (may not really be an issue, just reduces risk)

juanpabloprado: These are configuration properties of the possible values that can be set, new ones being token and tokenFile, I think these kind of clases shouldn't be extended, in case there is an example of it being extended, I could take a look to it.
In test code it uses the Builder pattern to build the objects

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactCredentials.java
```diff
@@ -38,6 +42,18 @@
     this.name = account.getName();
   }
 
+  @Override
+  protected Headers getHeaders(BitbucketArtifactAccount account) {
+    Headers.Builder headers = new Headers.Builder();
+    Optional<String> token = account.getTokenAsString();
+    if (token.isPresent()) {
+      headers.set(AUTHORIZATION, "Bearer " + token.get());
```
zachsmith1, line 50: Does Authorization need to be capitalized?

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactAccount.java
```diff
@@ -29,20 +30,29 @@
 
 @NonnullByDefault
 @Value
-public class BitbucketArtifactAccount implements ArtifactAccount, BasicAuth {
-  private final String name;
-  private final Optional<String> username;
-  private final Optional<String> password;
-  private final Optional<String> usernamePasswordFile;
+public class BitbucketArtifactAccount implements ArtifactAccount, BasicAuth, TokenAuth {
+  String name;
```
zachsmith1, line 34: I believe these can still be private

zachsmith1 (commented):
Just two items -- otherwise LGTM

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactAccount.java
juanpabloprado: @Value already marks non-static fields final.
@Value already marks non-static, package-local fields private.

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactCredentials.java
juanpabloprado: I used a constant i.e.
`public static final String AUTHORIZATION = "Authorization";`
by importing
`import static org.springframework.http.HttpHeaders.AUTHORIZATION;`

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactAccount.java
juanpabloprado: Overloaded the constructors

clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/bitbucket/BitbucketArtifactAccount.java
juanpabloprado: Screen Shot 2022-01-12 at 14 08 22

PR #5605, opened 2021-12-15
feat(kubernetes): improve messaging in the delete (manifest) stage

dbyron-sf:
specifically when there's an attempt to delete an unknown kind

Also improve the "delete operation completed successfully for" message for coordinates
that have no name (which is the case when pipelines specify kinds and not manifestName).

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/manifest/KubernetesDeleteManifestOperation.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/manifest/KubernetesDeleteManifestOperation.java
@@ -22,6 +22,7 @@ import com.netflix.spinnaker.clouddriver.data.task.Task;
 import com.netflix.spinnaker.clouddriver.data.task.TaskRepository;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.KubernetesCoordinates;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesDeleteManifestDescription;
+import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesKind;
 import com.netflix.spinnaker.clouddriver.kubernetes.op.OperationResult;
 import com.netflix.spinnaker.clouddriver.kubernetes.op.handler.KubernetesHandler;
 import com.netflix.spinnaker.clouddriver.kubernetes.security.KubernetesCredentials;
@@ -76,7 +77,15 @@ public class KubernetesDeleteManifestOperation implements AtomicOperation<Operat
               .updateStatus(OP_NAME, "Looking up resource properties for " + c.getKind() + "...");
           KubernetesHandler deployer =
               credentials.getResourcePropertyRegistry().get(c.getKind()).getHandler();
-          getTask().updateStatus(OP_NAME, "Calling delete operation...");
+          if (deployer.kind().equals(KubernetesKind.NONE)) {
+            // No point in invoking kubectl with a resource type of none, as it fails with
+            //
+            // error: the server doesn't have a resource type "none"
+            //
+            // Throw an exception to fail the stage, similar to what happens when kubectl fails.
+            throw new IllegalStateException("Unable to delete unknown kind '" + c.getKind() + "'");
+          }
+          getTask().updateStatus(OP_NAME, "Calling delete operation for resource" + c + "...");
           result.merge(
               deployer.delete(
                   credentials,
@@ -84,10 +93,12 @@ public class KubernetesDeleteManifestOperation implements AtomicOperation<Operat
                   c.getName(),
                   description.getLabelSelectors(),
                   deleteOptions));
-          getTask()
-              .updateStatus(OP_NAME, " delete operation completed successfully for " + c.getName());
+          getTask().updateStatus(OP_NAME, "Delete operation completed successfully for " + c);
         });
 
+    getTask()
+        .updateStatus(
+            OP_NAME, "Delete operation completed successfully for all applicable resources");
     return result;
   }
 }
--- a/clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/op/KubernetesDeleteManifestOperationTest.java
+++ b/clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/op/KubernetesDeleteManifestOperationTest.java
@@ -17,13 +17,16 @@
 
 package com.netflix.spinnaker.clouddriver.kubernetes.op;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
@@ -40,6 +43,7 @@ import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.Kuberne
 import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesKind;
 import com.netflix.spinnaker.clouddriver.kubernetes.op.handler.KubernetesCustomResourceDefinitionHandler;
 import com.netflix.spinnaker.clouddriver.kubernetes.op.handler.KubernetesCustomResourceHandler;
+import com.netflix.spinnaker.clouddriver.kubernetes.op.handler.KubernetesDeploymentHandler;
 import com.netflix.spinnaker.clouddriver.kubernetes.op.handler.KubernetesHandler;
 import com.netflix.spinnaker.clouddriver.kubernetes.op.handler.KubernetesReplicaSetHandler;
 import com.netflix.spinnaker.clouddriver.kubernetes.op.handler.KubernetesServiceHandler;
@@ -75,6 +79,7 @@ public class KubernetesDeleteManifestOperationTest {
   private static final GlobalResourcePropertyRegistry resourcePropertyRegistry =
       new GlobalResourcePropertyRegistry(
           ImmutableList.of(
+              new KubernetesDeploymentHandler(),
               new KubernetesReplicaSetHandler(),
               new KubernetesServiceHandler(),
               new KubernetesCustomResourceDefinitionHandler()),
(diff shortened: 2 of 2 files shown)


mattgogerly (approved):
Awesome!

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/manifest/KubernetesDeleteManifestOperation.java
```diff
@@ -76,18 +77,28 @@ public OperationResult operate(List<OperationResult> priorOutputs) {
               .updateStatus(OP_NAME, "Looking up resource properties for " + c.getKind() + "...");
           KubernetesHandler deployer =
               credentials.getResourcePropertyRegistry().get(c.getKind()).getHandler();
-          getTask().updateStatus(OP_NAME, "Calling delete operation...");
+          if (deployer.kind().equals(KubernetesKind.NONE)) {
```
german-muzquiz, line 80: Does this work for dynamically registered CRDs (those that are not specified in the kubernetes account configuration)? I think it should work based on previous PRs to fix delete and patch manifest stages for CRDs, but I'd like to double check.

There's a risk of breaking backwards compatibility for an obscure use case: native kubernetes kinds are [hardcoded in this file](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/manifest/KubernetesKind.java#L34), and we're not keeping it up to date with changes in kubernetes API versions, which is ok, because the way clouddriver works is that if it finds a kind that is not a CRD and is not in that list, it assigns the `NONE` kind. Then the default implementation of [delete](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/CanDelete.java#L35) uses the name of the manifest for the kubectl call, which becomes something like `kubectl delete corekind`, which works because it doesn't need the group name to be specified, because it's a core kind.

In the end, this `if` condition breaks the case of deleting core kinds which are not in the hardcoded list in clouddriver. For example, try deleting a ReplicationController or ResourceQuota.

dbyron-sf: > Does this work for dynamically registered CRDs (those that are not specified in the kubernetes account configuration)? I think it should work based on previous PRs to fix delete and patch manifest stages for CRDs, but I'd like to double check.

Yes it does. [This test](https://github.com/spinnaker/clouddriver/blob/33c6b7c4b94696b6353867ba2912e7eaeb8883f9/clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/op/KubernetesDeleteManifestOperationTest.java#L145) still passes with this PR.

> There's a risk of breaking backwards compatibility for an obscure use case: native kubernetes kinds are [hardcoded in this file](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/manifest/KubernetesKind.java#L34), and we're not keeping it up to date with changes in kubernetes API versions, which is ok, because the way clouddriver works is that if it finds a kind that is not a CRD and is not in that list, it assigns the `NONE` kind. Then the default implementation of [delete](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/CanDelete.java#L35) uses the name of the manifest for the kubectl call, which becomes something like `kubectl delete corekind`, which works because it doesn't need the group name to be specified, because it's a core kind.
> 
> In the end, this `if` condition breaks the case of deleting core kinds which are not in the hardcoded list in clouddriver. For example, try deleting a ReplicationController or ResourceQuota.

Hmmm...any thoughts on a good way forward that doesn't break this case?

german-muzquiz: I can think on these options:
1. Add missing core kubernetes kinds to the hardcoded list (so anything coming from `kubectl api-resources` in a newer version cluster). One problem is that this is an ongoing effort, if kubernetes adds another core kind in the future and clouddriver is not updated, spinnaker will not be able to delete those kinds.
2. Catch the `KubectlException` here when running `deployer.delete` and parse the error message to know if it is because of a truly missing kind in the cluster. I don't know if the error message is clear enough to be parsed.
3. Add a new method like `isValidKind` to `KubectlJobExecutor` that runs a `kubectl api-resources`, and call it before a delete. If the kind is valid, register it dynamically like CRDs, so subsequent deletes have the right KubernetesKind instance:
```
if kind == NONE
    if kubectlJobExcutor.isValidKind
        // register kind, same logic as dynamic CRDs
    else
        throw unknown kind exception
```

dbyron-sf: Thanks for these suggestions.  Of these, I think the third one has the best chance of success.  If we were using a java client library instead of invoking kubectl, perhaps option 2 would be more feasible, but I think it'll be tough to get right as it is.  And I agree that the maintenance burden of option 1 is too big.

I started pondering how to implement this.  Probably makes sense to discuss in the sig, but using https://github.com/spinnaker/clouddriver/pull/5334 as a model, it feels like adding code to KubernetesCredentials to periodically call a new KubectlJobExecutor method that runs `kubectl api-resources` makes sense.  Not sure what subclass of KubernetesHandler will make sense...maybe one that already exists, or maybe a new one.  At least by name, `KubernetesCustomResourceHandler` doesn't feel right, but it may do the right thing.

From there, adding a new member and mutator method to GlobalResourcePropertyRegistry and adjusting the get method to use it...and then I'm not sure whether KubernetesKindRegistry needs to know about these new kinds or not.

german-muzquiz: Sounds right, the only thing is that probably we can get away without another memoizer periodically calling `kubectl api-resources`, if we call it on demand when trying to process a kind `NONE`, or better yet, when creating the instance of `KubernetesKind` that assigns `NONE` when there's no known kind in the hardcoded list or in the CRD dynamic list.

dbyron-sf: OK, I'll look at doing it that way.  Any thoughts on if there's an existing subclass of KubernetesHandler that makes sense to use?

german-muzquiz: I think we can create a new class like `KubernetesGenericCoreKindHandler`, which can be a copy of `KubernetesCustomResourceHandler`, but returning the `KubernetesCoreCachingAgent` from the cachingAgentFactory method.

dbyron-sf: One more link that I think is helpful is to [KubectlJobExecutor::kubectlLookupInfo](https://github.com/spinnaker/clouddriver/blob/7e5d5f53c5e738764e98da529ddd6c29d8ebe8d5/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutor.java#L603), which [KubectlJobExecutor::delete](https://github.com/spinnaker/clouddriver/blob/7e5d5f53c5e738764e98da529ddd6c29d8ebe8d5/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutor.java#L137) uses to build the kubectl command.

@german-muzquiz I'm questioning this:
> Then the default implementation of [delete](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/CanDelete.java#L35) uses the name of the manifest for the kubectl call, which becomes something like kubectl delete corekind, which works because it doesn't need the group name to be specified, because it's a core kind.
because of [this code in kubectlLookupInfo](https://github.com/spinnaker/clouddriver/blob/7e5d5f53c5e738764e98da529ddd6c29d8ebe8d5/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutor.java#L608-L612):
```
    if (!Strings.isNullOrEmpty(name)) {
      command.add(kind + "/" + name);
    } else {
      command.add(kind.toString());
    }
```
Doesn't this mean the kind is important / shows up in the resulting command in addition to the name?  Instead of `kubectl delete corekind` we'd end up with `kubectl delete none/corekind`?  Or maybe I'm missing something.  Adding a test is the likely way forward, but still wanted to raise the question here.

german-muzquiz: The kind needs to be included in the command, so if the manifest name is not empty the command in the current implementation is `kubectl delete corekind/name` rather than `kubectl delete none/corekind`.
This is because the `KubernetesKind` is a dynamically created kind with the right name but with group `NONE`, so `kind.toString()` is `corekind`.

dbyron-sf: Aaah, ok.  I seem to have lost the place where that dynamically created kind comes from.

dbyron-sf: We discussed this in the kubernetes sig.  See [the minutes from 2022-03-08](https://docs.google.com/document/d/1db_yw1uru99Byvin4lQgm7aUZ9xMmvsARtEiiNUrmBo/edit#heading=h.nerwc578kmzc).  Because `deployer.delete` gets the kind from the credentials (as opposed to the `c` object or `c.getKind()`, it's not clear how deleting core kinds not mentioned in [KubernetesKindProperties::getGlobalKindProperties](https://github.com/spinnaker/clouddriver/blob/272fb8f8e54a406236bbf7f42cabafe568474f46/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/manifest/KubernetesKindProperties.java#L31) could ever work...but we generally think it does. Time to add some tests....

dbyron-sf:
@mergifyio update

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/manifest/KubernetesDeleteManifestOperation.java
mattgogerly: I've actually just run into a very similar sounding issue. If a custom `Kind` is not explicitly registered in the credentials definition, and there are no instances of that resource in the account, then the `KubernetesUnregisteredCustomResourceHandler` handler is used and returns a `none` kind, and the `Delete (Manifest)` stage fails.

If there are one or more the resources in the account, the `KubernetesCustomResourceHandler` is used and this returns the correct kind. This behaviour seems to have been introduced by https://github.com/spinnaker/clouddriver/pull/5334 and I'm really not sure how best to tackle it.

PR #5609, opened 2021-12-21
feat(aws): Improving AWS EC2 instance types API integration and caching, feat(aws): Adding archi type to images API and caching

pdk27:
## Changes:
- Improving AWS EC2 instance types API integration and caching:
   - Replaced AWS EC2 pricing docs integration with [AWS EC2 describe-instance-types API](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instance-types.html). Pricing docs precedes the API. Instance types retrieved via API are up-to-date and include other associated metadata.
   - Added instance type information to the cache.
- Adding architecture type to images API and caching

### Clouddriver API response before-after:
Before: 

After: 

## Use cases / reasoning for the changes:
- Deck - filtering out incompatible instance types
- Deck - display instance type info in instance type selector drop down and support filtering by metadata (PR to follow) 
See draft PR and demo in https://github.com/spinnaker/deck/pull/9793 

## Instructions to be included in release (preview) notes:
- ~As part of upgrading to a Spinnaker version that includes these changes, repopulate caches by running `hal deploy apply --flush-infrastructure-caches`.~
See https://github.com/spinnaker/clouddriver/pull/5609#issuecomment-1222860970


pdk27:
Reviewers, please also review the `Instructions to be included in release (preview) notes` in PR overview.

mattgogerly:
Haven't had a chance to look at the code yet, but for the release note - not everybody deploys Spinnaker using Halyard. Are there instructions for other installation methods (i.e. what is that flag actually doing?)

pdk27:
> Haven't had a chance to look at the code yet, but for the release note - not everybody deploys Spinnaker using Halyard. Are there instructions for other installation methods (i.e. what is that flag actually doing?)

Thanks for the callout @mattgogerly. That is where I need help :) 
I believe `hal deploy apply --flush-infrastructure-caches` [refreshes the caches](https://spinnaker.io/docs/setup/install/faq/#im-seeing-duplicatebad-infrastructure-entries-in-the-ui-after-deploying-config-changes). Someone familiar with it can add more details. I will post about this PR and ask for help in the Slack channel too.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ImageCachingAgent.groovy
```diff
@@ -159,6 +159,7 @@ class ImageCachingAgent implements CachingAgent, AccountAware, DriftMetric, Cust
         namedImageCacheDataMap.put(namedImageId, new DefaultCacheData(namedImageId, [
           name              : image.name,
           virtualizationType: image.virtualizationType,
+          architecture      : image.architecture,
```
dbyron-sf, line 162: Is there anything that uses this field?  Without some change to [AmazonNamedImageLookupController](https://github.com/spinnaker/clouddriver/blob/8147e9b9244dd6fd2044737acc63545016f4a2ce/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/controllers/AmazonNamedImageLookupController.groovy), what does this accomplish?

Also, is this (NAMED_IMAGES) a good place for this information?  It already feels unsavory having virtualizationType and creationDate here, since it's possible for two different images with the same name to have different values...The same is true for architecture as well though, right?  Is it enough to have it in IMAGES, which I think 
```
Map<String, Object> attributes = objectMapper.convertValue(image, ATTRIBUTES)
```
and
```
imageCacheData.add(new DefaultCacheData(imageId, attributes, [(NAMED_IMAGES.ns): [namedImageId]]))
```
already does.

mattgogerly:
> > Haven't had a chance to look at the code yet, but for the release note - not everybody deploys Spinnaker using Halyard. Are there instructions for other installation methods (i.e. what is that flag actually doing?)
> 
> Thanks for the callout @mattgogerly. That is where I need help :) I believe `hal deploy apply --flush-infrastructure-caches` [refreshes the caches](https://spinnaker.io/docs/setup/install/faq/#im-seeing-duplicatebad-infrastructure-entries-in-the-ui-after-deploying-config-changes). Someone familiar with it can add more details. I will post about this PR and ask for help in the Slack channel too.

All I can find is that it wipes the Redis instance, but that would only work if you're using the built in Redis. Why is it necessary to do that?

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
```diff
@@ -103,155 +80,148 @@ public AmazonInstanceTypeCachingAgent(
...
+    if (instanceTypesInfo != null && !instanceTypesInfo.isEmpty()) {
+      List<CacheData> instanceTypeData =
+          instanceTypesInfo.stream()
+              .map(
+                  i -> {
+                    Map<String, Object> attributes = objectMapper.convertValue(i, ATTRIBUTES);
+                    attributes.put("account", account.getName());
+                    attributes.put("region", region);
+                    attributes.put("name", i.getInstanceType());
+                    attributes.put("defaultVCpus", i.getVCpuInfo().getDefaultVCpus());
```
deverton, line 106: Can `.getVCpuInfo()` or `getMemoryInfo()` ever return `null`? There's a `null` check for `getInstanceStorageInfo()` below so wanted to be sure these were different.

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
```diff
@@ -103,155 +80,148 @@ public AmazonInstanceTypeCachingAgent(
...
+                                          "gpuSizeInMiB", g.getMemoryInfo().getSizeInMiB());
+                                      return gpuDeviceInfo;
+                                    })
+                                .collect(Collectors.toList()));
+                      }
+                      attributes.put("gpuInfo", gpuInfoAttributes);
+                    }
+
+                    if (i.getNetworkInfo() != null) {
+                      attributes.put("ipv6Supported", i.getNetworkInfo().getIpv6Supported());
```
deverton, line 156: Would it be better to have this attribute not be optional? It's always a bit clunky consuming these values from the map when you have to do something like `if (attributes.get("key") != null && attributes.get("key") == true)` or equivalent.

link108:
Hi @pdk27, just wanted to follow up on this PR, I saw the clouddriver feature was merged (https://github.com/spinnaker/clouddriver/pull/5610) and want to ensure this gets in as well 👍

pdk27:
> Hi @pdk27, just wanted to follow up on this PR, I saw the clouddriver feature was merged (#5610) and want to ensure this gets in as well 👍

Thanks for the heads-up @link108. Will address the comments here soon.

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
pdk27: `getVCpuInfo()` and `getMemoryInfo()` can never be null. `getInstanceStorageInfo()` can return null for instance types that don't support instance storage, and support EBS-only for example. Here are some examples - https://aws.amazon.com/ec2/instance-types/

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
pdk27: The `i` - [InstanceTypeInfo](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_InstanceTypeInfo.html) here is a result object returned from the AWS API `DescribeInstanceTypes` and `networkInfo` is an optional parameter. Also, not all instance types support IPv6, hence, the null check. Hope this adds clarification.

pdk27:
> > > Haven't had a chance to look at the code yet, but for the release note - not everybody deploys Spinnaker using Halyard. Are there instructions for other installation methods (i.e. what is that flag actually doing?)
> > 
> > 
> > Thanks for the callout @mattgogerly. That is where I need help :) I believe `hal deploy apply --flush-infrastructure-caches` [refreshes the caches](https://spinnaker.io/docs/setup/install/faq/#im-seeing-duplicatebad-infrastructure-entries-in-the-ui-after-deploying-config-changes). Someone familiar with it can add more details. I will post about this PR and ask for help in the Slack channel too.
> 
> All I can find is that it wipes the Redis instance, but that would only work if you're using the built in Redis. Why is it necessary to do that?

@mattgogerly 
Its needed to refresh the cache, and to be able to use the features in this PR.

pdk27:
> 

Gotcha! That doesn't sound ideal. As part of testing these changes, I couldn't see the new cache fields without flushing caches in my setup. Any ideas on how to confirm if flushing caches can be avoided? I will try a few things on my setup and update here.

pdk27:
> Gotcha! That doesn't sound ideal. As part of testing these changes, I couldn't see the new cache fields without flushing caches in my setup. Any ideas on how to confirm if flushing caches can be avoided? I will try a few things on my setup and update here.

Gotcha! That doesn't sound ideal. As part of testing these changes, I couldn't see the new cache fields without flushing caches in my setup. Any ideas on how to confirm if flushing caches can be avoided? I will try a few things on my setup and update here.

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ImageCachingAgent.groovy
pdk27: Great callouts! @dbyron-sf 

As far as the refactoring of images and named images go, it might make sense to tackle that separately. I am not very well versed with the images API and such a refactor will probably need a deep dive into its impact. 

I verified that adding architecture to attributes is not enough and won't reflect in the API response. 

----

The architecture addition in named images, `ImageCachingAgent` is rendered by [AmazonNamedImageLookupController](https://github.com/spinnaker/clouddriver/blob/8147e9b9244dd6fd2044737acc63545016f4a2ce/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/controllers/AmazonNamedImageLookupController.groovy) as is. So, adding `architecture` here reflected in the response of the find images API, when the parameter is the name of an image (e.g. http://localhost:8084/images/find?q=hello*). 

However, a change to [AmazonNamedImageLookupController](https://github.com/spinnaker/clouddriver/blob/8147e9b9244dd6fd2044737acc63545016f4a2ce/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/controllers/AmazonNamedImageLookupController.groovy) is required for responses to find images API with image ID parameter (e.g. http://localhost:8084/images/find?q=ami-23751835) for architecture to be included in its response. Thanks for bringing my attention to this. I have updated the PR. Here is a screenshot of a sample: 
image

pdk27:
> > > > > > Haven't had a chance to look at the code yet, but for the release note - not everybody deploys Spinnaker using Halyard. Are there instructions for other installation methods (i.e. what is that flag actually doing?)
> > > > > 
> > > > > 
> > > > > Thanks for the callout @mattgogerly. That is where I need help :) I believe `hal deploy apply --flush-infrastructure-caches` [refreshes the caches](https://spinnaker.io/docs/setup/install/faq/#im-seeing-duplicatebad-infrastructure-entries-in-the-ui-after-deploying-config-changes). Someone familiar with it can add more details. I will post about this PR and ask for help in the Slack channel too.
> > > > 
> > > > 
> > > > All I can find is that it wipes the Redis instance, but that would only work if you're using the built in Redis. Why is it necessary to do that?
> > > 
> > > 
> > > @mattgogerly Its needed to refresh the cache, and to be able to use the features in this PR.
> > 
> > 
> > Does the cache not get updated by normal clouddriver caching? The equivalent to the hal command for SQL/non-Halyard would be to wipe the database, at least of AWS data, which isn't ideal operationally.
> 
> Gotcha! That doesn't sound ideal. As part of testing these changes, I couldn't see the new cache fields without flushing caches in my setup. Any ideas on how to confirm if flushing caches can be avoided? I will try a few things on my setup and update here.

I was able to verify in my setup that flushing caches is required in order to see the changes to the cached data. May be that is too extreme? Do you know if there is a softer option like a refresh? Will reach on Slack again to see if anyone has a solution. 

Without running `hal deploy apply --flush-infrastructure-caches`, the images API response just didn't include the new `architecture` field and the response looked identical to one from before the change. I think the question here is related to changes to Spinnaker cache setup for those users that are interested in the changes in this PR and it doesn't need to block this PR from getting merged. I can followup and a note to release notes. Do you agree?

german-muzquiz:
I'm not too familiar with the AWS provider, but as a general principle the cache in clouddriver should be able to automatically rebuild itself if the underlying datastore is deleted (sql or redis), which according to the comments is working fine.

Then what I'm hearing is that the cache for instance types only gets populated once and then never gets updated. Probably that was expected in the beginning since instance types is not something that changes regularly, but then I think ideally this PR needs to automatically refresh that cache. If there are no caching agents for instance types, maybe refresh it upon startup or use something like `last_modified` fields to know that it should be refreshed?

pdk27:
> I'm not too familiar with the AWS provider, but as a general principle the cache in clouddriver should be able to automatically rebuild itself if the underlying datastore is deleted (sql or redis), which according to the comments is working fine.
> 
> Then what I'm hearing is that the cache for instance types only gets populated once and then never gets updated. Probably that was expected in the beginning since instance types is not something that changes regularly, but then I think ideally this PR needs to automatically refresh that cache. If there are no caching agents for instance types, maybe refresh it upon startup or use something like `last_modified` fields to know that it should be refreshed?

Thanks for the pointer @german-muzquiz. [This](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/config/ProviderHelpers.java#L74) snippet indicates that the instanceType Caching agent (existed before this PR) is conditionally used. I will dig into this to verify the same and get back to you.

pdk27:
> > I'm not too familiar with the AWS provider, but as a general principle the cache in clouddriver should be able to automatically rebuild itself if the underlying datastore is deleted (sql or redis), which according to the comments is working fine.
> > Then what I'm hearing is that the cache for instance types only gets populated once and then never gets updated. Probably that was expected in the beginning since instance types is not something that changes regularly, but then I think ideally this PR needs to automatically refresh that cache. If there are no caching agents for instance types, maybe refresh it upon startup or use something like `last_modified` fields to know that it should be refreshed?
> 
> Thanks for the pointer @german-muzquiz. I will dig into this further and get back to you.


@german-muzquiz 
Just to give you a brief context, there are 2 API related changes in this PR: instance type API, and images API

Within images API, there are 2 kinds: 
(1) find by name like http://localhost:8084/images/find?q=hello* 
(2) find by ID like http://localhost:8084/images/find?q=ami-23751835 
The change in this PR includes adding a new attribute `architecture` to both the images APIs. I can see the new field in the API response for (2) without flushing the caches, but not in the API response for (1) which comes from `namedImages`, built in [ImageCachingAgent](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ImageCachingAgent.groovy#L159). Flushing caches is non-ideal for certain types of cache setups as @mattgogerly pointed out that flushing caches means wiping the SQL/ non-Halyard database.

My observations:
- Flushing infrastructure caches  via `hal deploy apply --service-names clouddriver --flush-infrastructure-caches` is the only way I could get the public images in the caches to reload with the new `architecture` field. Re-deploying clouddriver didn't help.

Some questions we need help with:
- Is this a problem in all types of cache setups? (I could verify the behavior only for Halyard + Redis cache)
   May be there are other ways to reload images cache in other setups, without the need for wiping the whole cache? 
- How are new attributes added to cached data in general?
- With the given findings, how do we proceed?

german-muzquiz:
@pdk27 I don't see anything wrong in the code leading to the issue of the architecture not showing up in `http://localhost:8084/images/find?q=hello` without flushing caches, the ImageCachingAgent should be rebuilding the cache each time for both named images and ami images.

My suggestion is to investigate as follows:
* Start with an empty redis
* Run clouddriver without your changes, so that images are initially cached _without_ the architecture field.
* Verify in redis that images don't have the architecture field. This can be done through `redis-cli` with a command like this:
```
mget "com.netflix.spinnaker.clouddriver.aws.provider.AwsProvider:namedImages:attributes:aws:namedImages:{account}:{image name}"
```
Where you replace {account} and {image name} for their respective values.
* Verify that the request to clouddriver `http://localhost:7002/aws/images/find?q={image name}` doesn't include the architecture field
* Stop clouddriver and then run it with your changes, _without_ flushing the caches
* Check again redis, this time the architecture field should have been stored
* Check again the response of clouddriver REST API, this time the architecture field should be included 

The best approach is to not having to delete the cache (redis or sql) in order for this change to work, because the images cache is rebuilt every 30 seconds anyway.

pdk27:
Apologies for the delay following-up on this. 

@mattgogerly @dbyron-sf I tried the instructions provided by @german-muzquiz and it worked. i.e. I didn't need to flush infrastructure caches to see the architecture type in response. 

There is no other pending action from my side. Can you please help me identify the right reviewers for this PR?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ImageCachingAgent.groovy
dbyron-sf: Thanks.  https://github.com/spinnaker/clouddriver/pull/5609/commits/1d3b82efd68d9c6de1d5c84144eb308ef8ba5c8e looks right to me.

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
```diff
@@ -103,155 +80,148 @@ public Collection<AgentDataType> getProvidedDataTypes() {
...
-        .map(Header::getElements)
-        .filter(e -> e.length > 0)
-        .map(e -> e[0].getName());
+    return instanceTypeInfoList;
   }
 
   @Override
   public String getAgentType() {
-    return getClass().getSimpleName() + "/" + region;
+    return String.format("%s/%s/%s", getClass().getSimpleName(), account.getName(), region);
```
dbyron-sf, line 211: With info cached with the previous agent type get evicted?  Or does this change make it so that old info will stick around?

pdk27: You suspected right, I see both the old and new keys in my redis cache. 
image

Old key: `com.netflix.spinnaker.clouddriver.aws.provider.AwsInfrastructureProvider:AmazonInstanceTypeCachingAgent/eu-central-1:relationships`
New key: `com.netflix.spinnaker.clouddriver.aws.provider.AwsInfrastructureProvider:AmazonInstanceTypeCachingAgent/my-aws-devel-acct/eu-central-1:relationships`

I don't see a default TTL in Redis or Clouddriver config. As a result the old keys live forever until the redis database is flushed. This is probably not a common occurence, but would you know of a solution to cleanup the old keys? 

The only simple option I can think of is to keep the old agent type without the account name for the sake of backwards compatibility.

If we want to maintain the correctness of agent type and if a manual step before using these changes is okay, the release notes could include some commands to clean up the old keys?

dbyron-sf: I think it's important to include the account name, since the info is now account specific.  I suppose we're ending up with a bunch of duplicate information since I can't imagine instance types vary much across accounts.  Not sure what to do about that though.  Probably nothing.

One reason I think including the account name is important is to support dynamic accounts.  When an account disappears, for code to clean up information associated with the account, I suspect the name needs to be there.

I think a release note to remove the old info is ok.

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
```diff
@@ -103,155 +80,148 @@ public Collection<AgentDataType> getProvidedDataTypes() {
...
+    // cache instance types for key "metadata" for backwards compatibility
+    Set<String> instanceTypes =
+        instanceTypesInfo.stream()
+            .map(InstanceTypeInfo::getInstanceType)
+            .collect(Collectors.toSet());
+    DefaultCacheData metadata = buildCacheDataForMetadataKey(providerCache, instanceTypes);
+    cacheResults.put(getAgentType(), Collections.singleton(metadata));
+
+    // cache instance types info
+    if (instanceTypesInfo != null && !instanceTypesInfo.isEmpty()) {
```
nemesisOsorio, line 97: I see some nested `if`s  do you think we can refactor and use early return pattern to make it more readable?
for example
```
    if (instanceTypesInfo == null || instanceTypesInfo.isEmpty()) {
      return new DefaultCacheResult(cacheResults);
    }
```

pdk27:
> Apologies for the delay following-up on this.
> 
> @mattgogerly @dbyron-sf I tried the instructions provided by @german-muzquiz and it worked. i.e. I didn't need to flush infrastructure caches to see the architecture type in response.
> 
> There is no other pending action from my side. Can you please help me identify the right reviewers for this PR?

Here is a demo of the testing done: 

https://user-images.githubusercontent.com/3614196/186512114-f287d071-e397-4c83-acab-ec0501be70e9.mov

pdk27:
@dbyron-sf Here is the [PR](https://github.com/spinnaker/spinnaker.io/pull/230) for release notes.

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
dbyron-sf: Thanks for https://github.com/spinnaker/spinnaker.io/pull/230.

dbyron-sf (approved):

PR #5611, opened 2022-01-01
chore(build): setup-java v2

sullis:
GitHub Actions:  setup-java v1 to v2

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,14 +16,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-java@v1
+      - uses: actions/setup-java@v2
         with:
           java-version: 11
-      - uses: actions/cache@v1
-        with:
-          path: ~/.gradle
-          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
-          restore-keys: |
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -16,14 +16,10 @@ jobs:
...
+      - uses: actions/setup-java@v2
         with:
           java-version: 11
-      - uses: actions/cache@v1
-        with:
-          path: ~/.gradle
-          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
-          restore-keys: |
-            ${{ runner.os }}-gradle-
+          distribution: 'temurin'
```
dbyron-sf, line 22: Thanks for doing this!

According to https://github.com/actions/setup-java#v2-vs-v1 and https://github.com/actions/setup-java/blob/main/docs/switching-to-v2.md#java-distribution we were getting zulu before, so this change does more than switch to setup-java v2...it also changes to temurin.  I could see us wanting to do that, but I'd like the commit message to reflect that change, and also explain the reasoning behind it...or maybe it's simpler to change to temurin in a separate PR?

sullis: Let's go with 'zulu'

dbyron-sf: Thanks!

dbyron-sf (approved):

PR #5614, opened 2022-01-13
fix(gce) : retry logic on GCP APIs

ravichander82:
The code change is to address the [#6544](https://github.com/spinnaker/spinnaker/issues/6544) issue, where the retry logic is not performed for 503 , 429 error codes 

In the [GCP document](https://cloud.google.com/apis/design/errors#retrying_errors) those are suggested for retry operation 

The RETRY_ERROR_CODES list is modified to include 429 , 503 error codes


mattgogerly:
Are the back offs implemented as required in that doc?

skandragon:
> 

It seems to use the same `SafeRetry` code for any of the responses, so this looks like it will have some form of retry with backoff logic.

skandragon (approved):
LGTM

PR #5615, opened 2022-01-14
fix(cloudfoundry): Remove lombok generated getter

sergio-quintero:
The `/credentials` endpoint was failing when clouddriver had an invalid cloudfoundry account, returning a `400 - Bad Request`.  The root cause was that Lombok was generating an additional unignored getter for the cloudfoundry client field (the class already has 2 Json Ignored getters for that field), so whenever a call for expanded credentials `/credentials?expand=true` was made, jackson tried to serialize it and made expensive (and unnecessary) calls to the cloudfoundry api and caused the 400 errors if the account is invalid.

This small fix improves latency for valid accounts and removes the 400 errors for invalid accounts.

This should be backported to the latest `1.27.x` release.

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/security/CloudFoundryCredentials.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/security/CloudFoundryCredentials.java
@@ -37,6 +37,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
+import lombok.AccessLevel;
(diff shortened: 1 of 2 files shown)


zachsmith1:
@sergio-quintero would you be able to provide a quick test to cover this in order to backport?

sergio-quintero:
> @sergio-quintero would you be able to provide a quick test to cover this in order to backport?

@zachsmith1 I added a test that validates the ignored fields via annotation are working properly. As long as these fields are not serialized when calling `/credentials` we should not see a Bad Request error again. The test also helps to keep the ignored fields list up to date when making changes to the `CloudFoundryCredentials` class. Let me know if that covers the change.

zachsmith1 (approved):

link108:
@mergifyio backport release-2.27.x

link108:

PR #5617, opened 2022-01-18
chore(build): update Mergify config

mattgogerly:


dbyron-sf:
Similar to https://github.com/spinnaker/deck/pull/9785 + https://github.com/spinnaker/deck/pull/9786, right?  Any differences worth calling out?

PS Thanks for doing this.

mattgogerly:
> Similar to [spinnaker/deck#9785](https://github.com/spinnaker/deck/pull/9785) + [spinnaker/deck#9786](https://github.com/spinnaker/deck/pull/9786), right? Any differences worth calling out?
> 
> PS Thanks for doing this.

Yup, looks like we're using strict across all repos and it was removed on the 6th.

Only difference is Deck's CI is slightly more complicated so the queue condition is different, but otherwise identical.

dbyron-sf (approved):

link108:
@mergifyio backport release-1.27.x release-1.26.x

link108:
@Mergifyio backport release-1.25.x

PR #5626, opened 2022-02-04
fix(cf): handle potential nulls in deploy stage

mattgogerly:
We've seen a couple of places where the operation description converter would throw a NullPointerException, resulting in an error message in Deck that just says "null". Hopefully catching them here so a useful error message can be displayed.

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/converters/DeployCloudFoundryServerGroupAtomicOperationConverter.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/converters/DeployCloudFoundryServerGroupAtomicOperationConverter.java
@@ -23,7 +23,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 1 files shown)


clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/deploy/converters/DeployCloudFoundryServerGroupAtomicOperationConverter.java
```diff
@@ -176,11 +180,9 @@ private ArtifactCredentials getArtifactCredentials(
       Map<Object, Object> manifestMap) {
     List<CloudFoundryManifest> manifestApps =
         new ObjectMapper()
-            .setPropertyNamingStrategy(PropertyNamingStrategy.KEBAB_CASE)
+            .setPropertyNamingStrategy(PropertyNamingStrategies.KEBAB_CASE)
```
dbyron-sf, line 183: This looks different than adding a null check. Can you explain what this and the TypeReference change below does?

mattgogerly: Yup that was just general cleanup while I was in the file - `PropertyNamingStrategy` [was deprecated](https://fasterxml.github.io/jackson-databind/javadoc/2.12/com/fasterxml/jackson/databind/PropertyNamingStrategy.KebabCaseStrategy.html) in favour of `PropertyNamingStrategies`, and the type reference parameter isn't necessary.

dbyron-sf (approved):

PR #5627, opened 2022-02-04
fix(core): improve robustness of account API

jvz:
This smooths out some rough edges discovered in the alpha version of the
Account Management API. In particular, this addresses a few concerns:
- Ignore class loading errors during classpath scan for supported
  CredentialsDefinition types. While these exceptions are purely
  theoretical using the existing Spring-provided class, this isn't
  always guaranteed to work via reflection.
- Ensure created and updated credentials contain appropriate permissions
  to prevent locking out the user making the API call.
- Catch exceptions thrown by secrets decryption when loading accounts.
  This prevents Clouddriver from crashing due to user error or service
  outages in the underlying secrets manager.
- Fix version used for newly created accounts when that account name had
  a previous history (e.g., deleting an account and recreating it).
- Extracts credentials controller logic into a service class.

Related to https://github.com/spinnaker/spinnaker/issues/6525 and https://github.com/spinnaker/gate/pull/1514

### DIFF
--- a/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/config/AccountDefinitionConfiguration.java
+++ b/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/config/AccountDefinitionConfiguration.java
@@ -19,19 +19,26 @@ package com.netflix.spinnaker.clouddriver.config;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.deser.std.StringDeserializer;
 import com.netflix.spinnaker.clouddriver.jackson.AccountDefinitionModule;
+import com.netflix.spinnaker.clouddriver.security.AccountCredentialsProvider;
 import com.netflix.spinnaker.clouddriver.security.AccountDefinitionMapper;
+import com.netflix.spinnaker.clouddriver.security.AccountDefinitionRepository;
+import com.netflix.spinnaker.clouddriver.security.AccountDefinitionService;
 import com.netflix.spinnaker.credentials.definition.CredentialsDefinition;
+import com.netflix.spinnaker.fiat.shared.FiatPermissionEvaluator;
 import com.netflix.spinnaker.kork.secrets.EncryptedSecret;
 import com.netflix.spinnaker.kork.secrets.SecretManager;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import javax.annotation.Nullable;
 import lombok.Data;
 import lombok.extern.log4j.Log4j2;
 import org.springframework.beans.factory.config.BeanDefinition;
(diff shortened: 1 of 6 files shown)


dbyron-sf:
Sorry to be a broken record...gotta ask for tests here too please.

jvz:
Tests around which point in particular? Some of the addressed items require non-trivial test setup to emulate. :/

jvz:
Oh right, I remember now. I _did_ try to set up a test for `AccountDefinitionService`, but setting up a proper spring test would require an immense refactor of existing spring config classes throughout Clouddriver (the only spring tests we have in Clouddriver currently load everything from `Main` which is in clouddriver-web, not core) or they'd be simply mocking all the important parts and end up testing how good I am at testing mocks. The SQL stuff could potentially use an integration test, though these are somewhat a pain given how long the existing SQL-related tests take to execute.

dbyron-sf:
Thanks much.  I realize breaking new ground on tests like this is a pile of work.  Let me know if I can help.

jvz:
So, note that it's been like a couple weeks since I originally wrote these diffs (just finally got around to cherry-picking them here), so I'm a little out of context now, but from what I remember, the main issue I had with trying to write an automated test here was that one of the things I'm trying to test is how Fiat permissions behave, not necessarily Spring Security in general. While I can always set up a mock security user with arbitrary roles, that doesn't exactly map to Spinnaker auth. Setting up a fuller test with Fiat enabled (even if mostly done via mocking the relevant Fiat classes) is what seems to require a bit more bean untangling. Maybe this sort of thing makes sense as a sort of TCK test? We could have an in-memory account repository for that test.

As for the issues around catching exceptions where needed, that seems a bit easier to test (potentially with some refactoring involved). The related fix in https://github.com/spinnaker/gate/pull/1514 is somewhat trivial to test for as it was related to mixing up how Jackson and Spring handle dynamic properties (I had only relied on the Jackson side of things before, but Spring Security won't check that in the SpEL expressions apparently).

dbyron-sf:
Since this is clouddriver, it feels OK/good to mock fiat for tests.  And having an in-memory account repository also makes sense, for the stuff that's not/above sql.

jvz:
I've added some basic smoke tests. I tried to test out the SecretEngine stuff, too, but that doesn't seem to want to get configured properly in this test. I could mock it, but then it'd be an overly mocked test. ¯\_(ツ)_/¯

jvz:
Oh no, I think I forgot to add a negative test case as I was fighting with spring-boot-test for a while to get the positive test case to pass (which is also why I noticed there wasn't an admin check in my added validation until just this last commit). As for testing the other methods, it's a little less straightforward as the usual mechanism for synchronizing between the repository and the provider is via credentials polling; I'll need to set up a test class for the corresponding `AccountCredentials` that would be generated from the `TestAccount` along with something to force them to reload right away. Otherwise, we can verify data by looking through the repository directly, though that's somewhat trivial. Though perhaps if we can figure out a way to test that without credentials polling, then we might be able to reduce or remove the need for the polling in general here.

dbyron-sf:
Maybe worth a chat, but is there a way to make a test class where polling is a no-op and reading the credentials gets the latest from the in-memory repository?

jvz:
That probably works. I'll try it out later (got a few things to do today unfortunately).

jvz:
Going to recreate this with other updates.

PR #5628, opened 2022-02-08
fix(cf): service binding name validation

mattgogerly:
#5420 began setting the service binding name when Spinnaker bound a service to an application. Unfortunately the set of characters that are valid for a service binding name is only a subset of the characters valid for a service instance name. Therefore if one of those characters was included in the service instance name Spinnaker would (silently) fail to bind the service to the application.


mattgogerly:
> I can’t remember if there is a delete service binding stage but if a user inputs two incorrectly named bindings(service bind stage and service unbind) will it automatically covert to a valid one for both?

I think so? The create binding stage uses the `CreateServiceBinding` model, as does the server group deploy stage. So whenever Spinnaker creates a binding it should go through this conversion. When you run a delete service binding stage it asks for the service instance name, and this change will run that through the conversion too so that it can find it.

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryClientUtils.java
```diff
@@ -34,6 +34,8 @@
 import retrofit2.Response;
 
 public final class CloudFoundryClientUtils {
+  private static final String VALID_BINDING_NAME_PATTERN = "[^A-Za-z0-9-]+";
```
dogonthehorizon, line 37: Is there documentation we can point to that outlines the binding name pattern?

dogonthehorizon (approved):

clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/CloudFoundryClientUtils.java
mattgogerly: It's not published as far as I can see, have to go digging in the CAPI code to find it.

As it turns out I missed that _ is allowed too, I'll open a PR to include that and link the relevant code tomorrow.

PR #5631, opened 2022-02-17
feat(kube/ext): Make KubernetesCredentials and KubectlJobExecutor extensible

german-muzquiz:
These classes (specially `KubernetesCredentials`) cannot be extended in plugins and contain a lot of logic. Allowing them to be extended will make it possible for users to change the behavior of the kubernetes provider using plugins.


dbyron-sf (approved):
LGTM

PR #5638, opened 2022-02-22
feat(kube/cache): Only cache spinnaker infra kube objects

german-muzquiz:
BREAKING: The changes in this PR modify the default behavior of the kubernetes provider from caching all infrastructure, to caching only objects shown in the classic spinnaker infrastructure screens (clusters, load balancers and firewalls). Running pipelines no longer rely on the cache anymore since Spinnaker 1.23, so it's safe to assume that the cache is only used for populating the UI.

The config options introduced allow to continue caching all the kinds like before, useful when using the "raw resources" UI screen, which displays every kubernetes kind.

Configuration
```
kubernetes:
  cache:
    enabled: (true|false)   # (Default: true). Whether caching is enabled at the provider level.
    cacheAll: (true|false)  # (Default: false). Whether to cache all kubernetes kinds or not. If this value is "true", the setting "cacheKinds" is ignored.
    cacheKinds: []          # (Default: list with kinds populating spinnaker infrastructure scren). Only cache the kubernetes kinds in this list. Names are in {kind.group} format, where the group is optional for core kinds. If the setting cacheAll is true, this setting is ignored.
    cacheOmitKinds: []      # (Default: empty). Do not cache the kinds in this list. The format of the list is the same as cacheKinds.
```

The default list of kinds to cache is:
* Deployment
* ReplicaSet
* CronJob
* DaemonSet
* Job
* StatefulSet
* Pod
* Ingress
* Service
* NetworkPolicy
* _Any CRD configured in `kubernetes.accounts[].customResourceDefinitions` having one of the supported spinnaker kinds (server group managers, server groups, instances, load balancers or security groups)_

With the default settings, no action is needed when upgrading to a new version of Spinnaker. Pipelines will continue to work, as well as the UI screens for clusters, load balancers and firewalls. Only for users working with the raw resources screen, they would need to set `kubernetes.cache.cacheAll: true` to continue using it.

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesCachingAgent.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesCachingAgent.java
@@ -31,7 +31,10 @@ import com.netflix.spinnaker.cats.cache.CacheData;
 import com.netflix.spinnaker.cats.provider.ProviderCache;
 import com.netflix.spinnaker.clouddriver.kubernetes.KubernetesCloudProvider;
 import com.netflix.spinnaker.clouddriver.kubernetes.config.KubernetesCachingPolicy;
+import com.netflix.spinnaker.clouddriver.kubernetes.config.KubernetesConfigurationProperties;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.KubernetesCoordinates;
+import com.netflix.spinnaker.clouddriver.kubernetes.description.KubernetesSpinnakerKindMap;
+import com.netflix.spinnaker.clouddriver.kubernetes.description.SpinnakerKind;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesCachingProperties;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesKind;
 import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesKindProperties;
@@ -41,11 +44,7 @@ import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.Kuberne
 import com.netflix.spinnaker.clouddriver.kubernetes.op.job.KubectlJobExecutor;
 import com.netflix.spinnaker.clouddriver.kubernetes.security.KubernetesCredentials;
 import com.netflix.spinnaker.clouddriver.kubernetes.security.KubernetesNamedAccountCredentials;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -58,6 +57,14 @@ import org.slf4j.LoggerFactory;
 public abstract class KubernetesCachingAgent
     implements AgentIntervalAware, CachingAgent, AccountAware {
   private static final Logger log = LoggerFactory.getLogger(KubernetesCachingAgent.class);
+
+  public static final List<SpinnakerKind> SPINNAKER_UI_KINDS =
+      Arrays.asList(
+          SpinnakerKind.SERVER_GROUP_MANAGERS,
+          SpinnakerKind.SERVER_GROUPS,
+          SpinnakerKind.INSTANCES,
+          SpinnakerKind.LOAD_BALANCERS,
+          SpinnakerKind.SECURITY_GROUPS);
 
   @Getter @Nonnull protected final String accountName;
   protected final Registry registry;
@@ -72,13 +79,19 @@ public abstract class KubernetesCachingAgent
 
   @Getter protected final Long agentInterval;
 
+  protected final KubernetesConfigurationProperties configurationProperties;
+
+  protected final KubernetesSpinnakerKindMap kubernetesSpinnakerKindMap;
+
   protected KubernetesCachingAgent(
       KubernetesNamedAccountCredentials namedAccountCredentials,
       ObjectMapper objectMapper,
       Registry registry,
       int agentIndex,
       int agentCount,
-      Long agentInterval) {
+      Long agentInterval,
+      KubernetesConfigurationProperties configurationProperties,
+      KubernetesSpinnakerKindMap kubernetesSpinnakerKindMap) {
     this.accountName = namedAccountCredentials.getName();
     this.credentials = namedAccountCredentials.getCredentials();
     this.objectMapper = objectMapper;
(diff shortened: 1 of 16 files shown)


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesUnregisteredCustomResourceCachingAgent.java
```diff
@@ -50,4 +62,9 @@ public KubernetesUnregisteredCustomResourceCachingAgent(
   protected ImmutableList<KubernetesKind> primaryKinds() {
     return credentials.getCrds();
   }
+
+  @Override
+  protected boolean cachesKind(KubernetesKind kind) {
+    return !kind.getApiGroup().isNativeGroup();
```
dbyron-sf, line 68: I guess I don't have the context in my head about what KubernetesUnregisteredCustomResourceCachingAgent does differently from KubernetesCoreCachingAgent...but I could use some comments here about why this method does what it does.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesCoreCachingAgent.java
```diff
@@ -60,4 +73,9 @@ public KubernetesCoreCachingAgent(
   protected List<KubernetesKind> primaryKinds() {
     return credentials.getGlobalKinds();
   }
+
+  @Override
+  protected boolean cachesKind(KubernetesKind kind) {
+    return kind.getApiGroup().isNativeGroup();
```
dbyron-sf, line 79: Similar request for comments as https://github.com/spinnaker/clouddriver/pull/5638/files#r812228254.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesCachingAgent.java
```diff
@@ -72,31 +79,87 @@
...
     Map<String, Object> result = new HashMap<>();
     result.put("namespaces", getNamespaces());
-    result.put("kinds", primaryKinds());
+    result.put("kinds", kindsToCache());
     return result;
   }
 
   protected abstract List<KubernetesKind> primaryKinds();
 
+  protected abstract boolean cachesKind(KubernetesKind kind);
```
dbyron-sf, line 115: Could use some javadoc here explaining why we need this method and what it's supposed to do.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesUnregisteredCustomResourceCachingAgent.java
dbyron-sf: Maybe you could do us all a big favor and add some comments to the class about what it's for, etc. ?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesCachingAgent.java
```diff
@@ -72,31 +79,87 @@
...
+              .map(KubernetesKind::fromString)
+              .filter(this::cachesKind)
+              .filter(credentials::isValidKind)
+              .collect(Collectors.toList());
+
+    } else {
+      // Only cache kinds used in Spinnaker's classic infrastructure screens, which are the kinds
+      // mapped to Spinnaker kinds like ServerGroups, Instances, etc.
+      cacheKinds =
+          kubernetesSpinnakerKindMap.allKubernetesKinds().stream()
```
dbyron-sf, line 137: Is it important to start with `kubernetesSpinnakerKindMap.allKubernetesKinds()` as opposed to `primaryKinds()` here?

dbyron-sf: Or maybe even more directly, start with SPINNAKER_UI_KINDS, and call `translateSpinnakerKind` ... I guess we still need
```
              .filter(this::cachesKind)
              .filter(credentials::isValidKind)
```
but it seems simpler.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesCachingAgent.java
```diff
@@ -72,31 +79,87 @@
...
+  protected abstract boolean cachesKind(KubernetesKind kind);
+
+  protected List<KubernetesKind> kindsToCache() {
+    List<KubernetesKind> cacheKinds;
+
+    if (configurationProperties.getCache().isCacheAll()) {
+      cacheKinds = primaryKinds();
+
+    } else if (configurationProperties.getCache().getCacheKinds() != null
+        && configurationProperties.getCache().getCacheKinds().size() > 0) {
```
dbyron-sf, line 124: If someone wants to cache nothing, how do they do it?  Seems like an empty list ends up caching SPINNAKER_UI_KINDS at the moment.

dbyron-sf: There may be a flag to cache nothing for all cloud providers, but I could see wanting to cache ec2 accounts, but not kubernetes accounts (i.e. if the infrastructure tabs are disabled in the UI).

dbyron-sf:
> @dbyron-sf I updated the code to make it simpler by always grabbing the list returned from `primaryKinds` and applying the filters to that list. Also I added the `kubernetes.cache.enabled` setting to be able to completely disable caching only for the kubernetes provider.

Thanks @german-muzquiz !  I'll take a look.  Do you mind updating the PR description with the new config flag?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/config/KubernetesConfigurationProperties.java
```diff
@@ -62,4 +64,45 @@
       long exponentialBackOffIntervalMs = 10000;
     }
   }
+
+  @Data
+  public static class Cache {
+
+    /** Whether caching is enabled in the kubernetes provider. */
```
dbyron-sf, line 71: If it's easy to add a test for this, please do.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesCachingAgent.java
dbyron-sf: Thanks for adding `kubernetes.cache.enabled`.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/agent/KubernetesUnregisteredCustomResourceCachingAgent.java
dbyron-sf: Thank you!

PR #5642, opened 2022-02-28
chore(it): Add delete manifest integration tests

AdanUrbanReyesArmory:
Add the test cases : 

- Delete static target
- Delete dynamic target
- Delete without cascading
- No target found

### DIFF
--- a/clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeleteManifestIT.java
+++ b/clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeleteManifestIT.java
@@ -17,20 +17,16 @@
 package com.netflix.spinnaker.clouddriver.kubernetes.it;
 
 import static io.restassured.RestAssured.given;
-import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.*;
 
 import com.netflix.spinnaker.clouddriver.kubernetes.it.utils.KubeTestUtils;
-import io.restassured.path.json.JsonPath;
 import io.restassured.response.Response;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.TimeUnit;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
-import org.opentest4j.AssertionFailedError;
 import org.springframework.http.HttpStatus;
 
 public class DeleteManifestIT extends BaseTest {
@@ -49,20 +45,27 @@ public class DeleteManifestIT extends BaseTest {
           + "Then the secret is deleted\n===")
   @Test
   public void shouldDeleteByStaticTarget() throws IOException, InterruptedException {
-    final String s = "secret";
-    final String sn = "mysecret";
-    Map<String, Object> sm =
+    // ------------------------- given --------------------------
+    final String kind = "secret";
+    final String name = "mysecret";
+    Map<String, Object> manifest =
         KubeTestUtils.loadYaml("classpath:manifests/secret.yml")
             .withValue("metadata.namespace", account1Ns)
-            .withValue("metadata.name", sn)
+            .withValue("metadata.name", name)
             .asMap();
-    kubeCluster.execKubectl("-n " + account1Ns + " apply -f -", sm);
-    List<Map<String, Object>> rb = buildStaticRequestBody(String.format("%s %s", s, sn), true);
-    List<String> on = KubeTestUtils.sendOperation(baseUrl(), rb, account1Ns);
-    String ss =
+    kubeCluster.execKubectl("-n " + account1Ns + " apply -f -", manifest);
+    // ------------------------- when ---------------------------
+    List<Map<String, Object>> request =
+        buildStaticRequestBody(String.format("%s %s", kind, name), true);
+    List<String> deletions = KubeTestUtils.sendOperation(baseUrl(), request, account1Ns);
+    // ------------------------- then ---------------------------
+    String exist =
         kubeCluster.execKubectl(
-            String.format("-n %s get %s %s --ignore-not-found", account1Ns, s, sn));
-    assertTrue(ss.isBlank() && on.size() == 1 && on.get(0).equals(String.format("%s %s", s, sn)));
+            String.format("-n %s get %s %s --ignore-not-found", account1Ns, kind, name));
+    assertTrue(
+        exist.isBlank()
+            && deletions.size() == 1
+            && deletions.get(0).equals(String.format("%s %s", kind, name)));
   }
 
   @DisplayName(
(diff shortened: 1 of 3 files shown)


clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeleteManifestIT.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  @DisplayName(
+      ".\n===\n"
+          + "Given an existing static target\n"
+          + "When sending delete manifest request\n"
+          + "It should wait until the target is deleted\n===")
+  @Test
+  public void staticTarget() throws IOException, InterruptedException {
+    final String sn = "mysecret";
```
german-muzquiz, line 93: There are missing comments like 
```
// ------------------------- given --------------------------
// ------------------------- when --------------------------
// ------------------------- then --------------------------
```
These are used to understand what is part of the actual tests and what is part of preparations or verification steps. These comments are also useful to correlate to the test case `DisplayName`.

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeleteManifestIT.java
```diff
@@ -0,0 +1,157 @@
...
+    Map<String, Object> sm =
+        KubeTestUtils.loadYaml("classpath:manifests/secret.yml")
+            .withValue("metadata.namespace", account1Ns)
+            .withValue("metadata.name", sn)
+            .asMap();
+    kubeCluster.execKubectl("-n " + account1Ns + " apply -f -", sm);
+    List<Map<String, Object>> rb = buildStaticRequestBody(String.format("secret %s", sn), true);
+    List<String> on = KubeTestUtils.delete(baseUrl(), rb, account1Ns);
+    assertEquals(1, on.size());
```
german-muzquiz, line 102: You still need to verify with `kubectl` that the secret was really deleted. Also, what's the actual value that the `sendOperation` method should return (not just the size of the list)?

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeleteManifestIT.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  @DisplayName(
+      ".\n===\n"
+          + "Given a existing dynamic target\n"
+          + "When sending delete manifest request\n"
+          + "It should wait until the target is deleted\n===")
+  @Test
+  public void dynamicTarget() throws InterruptedException, IOException {
+    deployReplicaSets((byte) 2);
```
german-muzquiz, line 112: It's preferable to use here the `KubeTestUtils.deployAndWaitStable` method instead of helper methods, even if that leads to duplicate code in test cases. The reason is that it's easier to understand and fix a failed test case if we know exactly what's doing without digging in too much nested helper methods.

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeleteManifestIT.java
```diff
@@ -0,0 +1,157 @@
...
+    List<String> on = KubeTestUtils.delete(baseUrl(), rb, account1Ns);
+    assertEquals(1, on.size());
+  }
+
+  @DisplayName(
+      ".\n===\n"
+          + "Given a existing deployment manifest\n"
+          + "When sending delete manifest request\n"
+          + "without the cascading option enable\n"
+          + "It should delete the deployment and on the background the pods\n===")
```
german-muzquiz, line 125: This description is misleading, the pods should NOT be deleted.

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeleteManifestIT.java
```diff
@@ -0,0 +1,157 @@
+/*
+ * Copyright 2022 Netflix, Inc.
```
german-muzquiz, line 2: Copyright

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeleteManifestIT.java
```diff
@@ -0,0 +1,157 @@
...
+          + "When sending delete manifest request\n"
+          + "It should wait until the target is deleted\n===")
+  @Test
+  public void dynamicTarget() throws InterruptedException, IOException {
+    deployReplicaSets((byte) 2);
+    List<Map<String, Object>> rb =
+        buildDynamicRequestBody(
+            "replicaSet nginx-v000", true, "replicaSet nginx", "newest", "replicaSet");
+    List<String> on = KubeTestUtils.delete(baseUrl(), rb, account1Ns);
+    assertEquals(1, on.size());
```
german-muzquiz, line 117: This is not verifying that actually the `newest` replicaset was deleted, which could be verified with `kubectl` to see which replicasets are left.
Also, can you add test cases for deleting the `Second Newest`, `Oldest`, `Largest` and `Smallest`?

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/utils/KubeTestUtils.java
```diff
@@ -398,4 +398,10 @@ public static void deployIfMissing(
 
     deployAndWaitStable(baseUrl, account, namespace, kind, name, app, image);
   }
+
+  public static List<String> delete(
```
german-muzquiz, line 402: This method could be deleted and directly use `sendOperation` instead.

german-muzquiz (commented):
Thanks for improving the test coverage of Spinnaker!, I have some comments to keep the tests with the same convention as others.

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeleteManifestIT.java
```diff
@@ -0,0 +1,347 @@
...
+          + "When sending a delete manifest operation using smallest dynamic target criteria\n"
+          + "Then it gets a 404 while fetching the manifest\n===")
+  @Test
+  public void shouldNotFoundDynamicTarget() throws InterruptedException {
+    String rs = "replicaSet";
+    String c = "smallest";
+    String rsn = String.format("not-exists-%s-test", c);
+    String rdn = String.format("%s-v000", rsn);
+    assertThrows(
+        AssertionFailedError.class,
```
german-muzquiz, line 301: This test looks looks like it can be simplified to just sending the `GET /manifests` request and expecting a 404. Asserting that it throws an assert exception is like having a double negated `if` condition.

german-muzquiz (commented):
Tests look better, a couple of general comments:

1. Can you add the `given`, `when`, `then` comments to made very clear what step of the test is doing each block of code?
```
 // ------------------------  given -------------------------
 // ------------------------  when -------------------------
 // ------------------------  then -------------------------
```
2. Use more descriptive variable names, for example instead of `rs` use `kind`. It helps with readability because when reading this code I had to go back to the variable assignment to know what's the difference between `rs`, `rsn` and `rdn`.

german-muzquiz (approved):

german-muzquiz:
@AdanUrbanReyesArmory looks like you need to manually update the branch with the latest changes from `master`

PR #5644, opened 2022-03-01
chore: bump google cloud sdk to v374.0.0

perek:
The new version of the Appengine SDK allows for specifying a service_account other than the default service account. This is broken on the version currently in CD. This PR bumps the version, I am not sure how to confirm backward compact, is there an integration test to validate?


dbyron-sf:
> The new version of the Appengine SDK allows for specifying a service_account other than the default service account. This is broken on the version currently in CD. This PR bumps the version, I am not sure how to confirm backward compact, is there an integration test to validate?

I'm not sure how much test coverage we have for this.  Is

> Added support for specifying a service account for each App Engine version during deployment. This can be done by either running gcloud app deploy --service-account=<SERVICE_ACCOUNT> or by specifying 'service_account: ' in app.yaml file. If both are specified, the CLI parameter will override the app.yaml spec.

from [the release notes for 342.0.0](https://cloud.google.com/sdk/docs/release-notes#34200_2021-05-25) the fix you're looking for?  I'm all for bringing things all the way up to date, but if we're afraid of breaking stuff, maybe only going as far as we need is better?

Maybe @skandragon or one of the other google cloud sig folks can say more?

perek:
I am open to decreasing the release, just figured we might as well take the pill and upgrade it all the way.

perek:
I went through all the breaking changes and the only two I saw that looked suspicious is
I dont think we use --client-id-file, but perhaps a kube config file could? I also dont think we create managed IG ?
0.373.0 - (Google Cloud CLI) Updated gcloud auth application-default login to use --no-browser in place of --no-launch-browser when using --client-id-file.
0.353.0 - (Compute Engine) Update gcloud compute instance-groups managed create to throw an error when creating a stateful regional managed instance group, if --instance-redistribution-type=NONE is not specified.

dbyron-sf:
Great that you've done that work.  I did some grepping and don't *think* those changes will affect us.  Would still love to hear from someone who knows more than me.

perek:
@dbyron-sf we validated in our prod env, this fix works to resolve the issue, no other issues noticed

dbyron-sf (approved):

PR #5645, opened 2022-03-01
fix(clouddriver): NPE applying empty artifact manifest

sepulvedablanco:
When a `Deploy (Manifest)` step is executed configuring an artifact without content, a `java.lang.NullPointerException` is thrown, which is displayed in Deck and the step fails as can be seen in the following screenshot:

image

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/manifest/KubernetesDeployManifestOperation.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/manifest/KubernetesDeployManifestOperation.java
@@ -43,6 +43,7 @@ import lombok.RequiredArgsConstructor;
 import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.util.CollectionUtils;
 
 public class KubernetesDeployManifestOperation implements AtomicOperation<OperationResult> {
   private static final Logger log =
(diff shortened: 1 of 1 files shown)


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/manifest/KubernetesDeployManifestOperation.java
```diff
@@ -167,18 +168,21 @@ public OperationResult operate(List<OperationResult> _unused) {
...
-      // check `manifest` for backwards compatibility until all existing stages have been updated.
-      @SuppressWarnings("deprecation")
-      KubernetesManifest manifest = description.getManifest();
+
+    // The stage currently only supports using the `manifests` field but we need to continue to
+    // check `manifest` for backwards compatibility until all existing stages have been updated.
+    @SuppressWarnings("deprecation")
+    KubernetesManifest manifest = description.getManifest();
+
+    if (CollectionUtils.isEmpty(inputManifests) && manifest != null) {
```
german-muzquiz, line 177: If both `description.getManifests()` and `description.getManifest()` return `null`, there will be a new `NullPointerException` in line 186, so we need to make sure that `inputManifests` is never `null`.

german-muzquiz (commented):
Thanks for contributing!
Can you add a unit test validating that the issue that you were seeing is fixed? I assume that's by supplying a description with null values for both `getManifests` and `getManifest`. Also we need a test that uses the singular version because now all of them use only the plural version.

mattgogerly:
@dbyron-sf think this is a duplicate of what you just fixed in #6337?

dbyron-sf:
> @dbyron-sf think this is a duplicate of what you just fixed in #6337?

Aaah yes, I believe it is.

PR #5661, opened 2022-03-16
chore(ci): GHA - simplify build versioning

kskewes-sf:
- collapse version info gathering steps into single `build_variables` step
- collapse version info parts into single string and use everywhere.
- use short git sha: `(git rev-parse --short HEAD)`

Note `build.yml` versioning is not compatible with Debian package
building as gradle plugin enforces `^[0-9]+`.
We don't publish master branch or release-* branches to GAR apt
repository though.
Prefixing the version with `<tag>-dev-` or something and publishing Debian
packages is possible but may pollute `apt-cache policy spinnaker-rosco` output
and overall be unnecessary with regular releases.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,14 +22,15 @@ jobs:
           java-version: 11
           distribution: 'zulu'
           cache: 'gradle'
-      - name: Extract repository name
-        id: extract_repo_name
-        run: echo ::set-output name=REPO::${GITHUB_REPOSITORY##*/}
+      - name: Prepare build variables
+        id: build_variables
+        run: |
+          echo ::set-output name=REPO::${GITHUB_REPOSITORY##*/}
+          echo ::set-output name=VERSION::"${GITHUB_REF_NAME}-$(git rev-parse --short HEAD)-$(date --utc +'%Y%m%d%H%M')"
       - name: Build
-        run: ./gradlew build --stacktrace ${{ steps.extract_repo_name.outputs.REPO }}-web:installDist
-      - name: Get date
-        id: get_date
-        run: echo ::set-output name=DATETIME::$(date --utc +'%Y%m%d%H%M')
+        env:
+          ORG_GRADLE_PROJECT_version: ${{ steps.build_variables.outputs.VERSION }}
+        run: ./gradlew build --stacktrace ${{ steps.build_variables.outputs.REPO }}-web:installDist
       - name: Login to GAR
         # Only run this on repositories in the 'spinnaker' org, not on forks.
         if: startsWith(github.repository, 'spinnaker/')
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -48,10 +49,10 @@ jobs:
           file: Dockerfile.slim
           push: true
           tags: |
-            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:${{ github.ref_name }}-latest-unvalidated"
-            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:${{ github.ref_name }}-${{ github.sha }}-${{ steps.get_date.outputs.DATETIME }}-unvalidated"
-            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:${{ github.ref_name }}-latest-unvalidated-slim"
-            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:${{ github.ref_name }}-${{ github.sha }}-${{ steps.get_date.outputs.DATETIME }}-unvalidated-slim"
+            "${{ env.CONTAINER_REGISTRY }}/${{ steps.build_variables.outputs.REPO }}:${{ github.ref_name }}-latest-unvalidated"
```
mattgogerly, line 52: Potentially stupid question - why is the slim Dockerfile tagged as slim and normal? Is the other Dockerfile not the "unslim" version?

mattgogerly: Also unrelated to this PR, I noticed PR builds are building containers - is that intentional?

kskewes-sf: Cheers Matt.

> why is the slim Dockerfile tagged as slim and normal?
This is the way it has always been and I guess it makes it explicit.
The same container image SHA will have both "none" and slim tags - so one can see that there aren't 3 versions (none, slim, ubuntu)...?

See: https://console.cloud.google.com/artifacts/docker/spinnaker-community/us/docker/orca

> PR builds are building containers - is that intentional?
Yes, this exercises Dockerfile related changes which is helpful for services doing a bit of extra work like downloading binaries, running scripts etc. The images aren't pushed.
It would be ideal to enable docker layer caching in the action and we probably can now that images are seeded in GAR. 
Maybe someone else can pick this up though as it's not blocking a 1.27.0 release?

mattgogerly: Cool, I just know there used to be Dockerfile and Dockerfile.slim, wasn't sure if there was an important distinction there.

And sounds good, just checking we're not burning resources/money accidentally ;)

dbyron-sf (approved):

PR #5666, opened 2022-03-17
feat(core): Update account management API

jvz:
This refactors some things in preparation for user secrets support along with updating the type discriminator handling for account definitions.

These changes are independent of https://github.com/spinnaker/fiat/pull/928 and https://github.com/spinnaker/kork/pull/942, and I've noted TODO comments on where relevant updates related to those PRs will be added (with additional tests).

### DIFF
--- a/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/CloudDriverConfig.groovy
+++ b/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/CloudDriverConfig.groovy
@@ -88,6 +88,7 @@ import com.netflix.spinnaker.clouddriver.search.executor.SearchExecutorConfig
 import com.netflix.spinnaker.clouddriver.security.AccountCredentials
 import com.netflix.spinnaker.clouddriver.security.AccountCredentialsProvider
 import com.netflix.spinnaker.clouddriver.security.AccountCredentialsRepository
+import com.netflix.spinnaker.clouddriver.security.AccountDefinitionSecretManager
 import com.netflix.spinnaker.clouddriver.security.DefaultAccountCredentialsProvider
 import com.netflix.spinnaker.clouddriver.security.MapBackedAccountCredentialsRepository
 import com.netflix.spinnaker.clouddriver.security.config.SecurityConfig
@@ -381,11 +382,13 @@ class CloudDriverConfig {
   @Bean
   DescriptionAuthorizerService descriptionAuthorizerService(Registry registry,
                                                             Optional<FiatPermissionEvaluator> fiatPermissionEvaluator,
-                                                            SecurityConfig.OperationsSecurityConfigurationProperties opsSecurityConfigProps) {
+                                                            SecurityConfig.OperationsSecurityConfigurationProperties opsSecurityConfigProps,
+                                                            AccountDefinitionSecretManager secretManager) {
     return new DescriptionAuthorizerService(
-      registry,
-      fiatPermissionEvaluator,
-      opsSecurityConfigProps
+            registry,
+            fiatPermissionEvaluator,
+            opsSecurityConfigProps,
+            secretManager
     )
   }
 
--- a/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/config/AccountDefinitionConfiguration.java
+++ b/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/config/AccountDefinitionConfiguration.java
@@ -19,19 +19,26 @@ package com.netflix.spinnaker.clouddriver.config;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.deser.std.StringDeserializer;
 import com.netflix.spinnaker.clouddriver.jackson.AccountDefinitionModule;
+import com.netflix.spinnaker.clouddriver.security.AccountCredentialsProvider;
 import com.netflix.spinnaker.clouddriver.security.AccountDefinitionMapper;
+import com.netflix.spinnaker.clouddriver.security.AccountDefinitionRepository;
+import com.netflix.spinnaker.clouddriver.security.AccountDefinitionService;
 import com.netflix.spinnaker.credentials.definition.CredentialsDefinition;
+import com.netflix.spinnaker.fiat.shared.FiatPermissionEvaluator;
 import com.netflix.spinnaker.kork.secrets.EncryptedSecret;
 import com.netflix.spinnaker.kork.secrets.SecretManager;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import javax.annotation.Nullable;
 import lombok.Data;
 import lombok.extern.log4j.Log4j2;
 import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.ConfigurationProperties;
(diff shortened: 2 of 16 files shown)


jvz:
Yeah yeah yeah, bot, I'll be squashing this PR into one commit. As for the test failure, this is strange as I didn't get that locally. I'm not sure how that happened :/

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/CloudDriverConfig.groovy
```diff
@@ -381,11 +382,13 @@ class CloudDriverConfig {
   @Bean
   DescriptionAuthorizerService descriptionAuthorizerService(Registry registry,
                                                             Optional<FiatPermissionEvaluator> fiatPermissionEvaluator,
-                                                            SecurityConfig.OperationsSecurityConfigurationProperties opsSecurityConfigProps) {
+                                                            SecurityConfig.OperationsSecurityConfigurationProperties opsSecurityConfigProps,
```
mattgogerly, line 385: Has spotless run on this? Seems much further indented than normal.

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionSecretManager.java
```diff
@@ -0,0 +1,49 @@
...
+  public boolean canAccessAccountWithSecrets(@Nonnull String accountName) {
+    var username = SecurityContextHolder.getContext().getAuthentication().getName();
+    if (authorizer.isAdmin(username)) {
+      return true;
+    }
+    var userRoles = authorizer.getRoles(username);
+    if (userRoles.isEmpty()) {
+      return false;
+    }
+    // TODO(jvz): update with https://github.com/spinnaker/kork/pull/942
```
mattgogerly, line 45: Is the intent here to check whether the account contains secrets in its definition, and if so whether the user can access the given secret?

If so what would happen if say a K8s account was defined with the Kubeconfig file as a secret but the user trying to deploy to that account didn't have access to the Kubeconfig secret? Does it matter they don't have access to the secret itself if they're just deploying?

clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/config/CloudDriverConfig.groovy
jvz: Not sure that spotless runs on Groovy files.

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionSecretManager.java
jvz: The idea here would be that if an account is using a user secret (as being proposed in the Kork PR), then the user using the account must also be allowed to read the user secret (i.e., the user will have to be in one of the listed roles for the secret in addition to having write access to the account itself).

german-muzquiz (commented):
Integration tests are failing because of a dependency version mismatch for `io.zipkin.brave:brace-instrumentation-http`: some part of the code is using version `5.12.3` and other part is using `5.10.1`. Not sure if this is a result of the recent version bump (and then revert) of spring boot: https://github.com/spinnaker/kork/pull/941
cc @dbyron-sf

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionSecretManager.java
```diff
@@ -0,0 +1,49 @@
...
+        : secretManager.decrypt(encryptedUri);
+  }
+
+  public boolean canAccessAccountWithSecrets(@Nonnull String accountName) {
+    var username = SecurityContextHolder.getContext().getAuthentication().getName();
+    if (authorizer.isAdmin(username)) {
+      return true;
+    }
+    var userRoles = authorizer.getRoles(username);
+    if (userRoles.isEmpty()) {
```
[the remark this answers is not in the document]
jvz: If Fiat is disabled, `isAdmin` returns `true`, so this code doesn't get executed. Good catch, though!

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionSecretManager.java
mattgogerly: Ok, not sure if I'm overthinking this but is there potential for secrets that should be restricted that would prevent somebody from deploying?

Thinking along the lines of a Spinnaker operator using the new secrets API to onboard a cluster role Kubeconfig, and then creating accounts using it. Without READ permission on that secret the accounts couldn't be deployed to, but presumably the operator wouldn't want users having read either?

Perhaps I'm misunderstanding the use case behind this!

jvz: If a user isn't in one of the listed roles in the user secret, then they won't be able to use the account even if they have write access to the account. That's why I have this abstraction here despite it not doing much yet as I need to do two authZ checks: when storing the account and when using the account.

mattgogerly: Yeah I get that - my question is around whether that should be the intention. Is there a use case where a user wouldn't have permissions on a secret but would still want/need to deploy to an account?

It feels a bit weird to have to get permissions for two different entities when the end user (read: joe bloggs deploying their app) might not know anything about the secret configured on an account, just what it's called - not to mention it's extra load on Fiat which already struggles at scale.

Again I'm not sure if I'm just overthinking this or if what I'm asking makes sense 👍

jvz: My issue with that is deploying with an account means you have arbitrary access to the secret (i.e., once you have it, there's always a way to sneak it out from there), so in an externalized user secrets model, that should not be allowed. Maybe there can be a relaxed mode (though that sort of already exists by just configuring `EncryptedSecret` URIs instead of the new `UserSecretReference` URIs).

mattgogerly: Yup that makes sense - falling back to EncryptedSecret.

Cheers!

german-muzquiz (approved):
Looks good to me

dbyron-sf:
@mergifyio update

dbyron-sf:
Hey @jvz did you mention some deck changes that go along with changing the type discriminator?  Or is it that deck works better but there aren't any changes?

jvz:
I said I changed the type discriminator in this PR which goes with the Gate PR, and it now matches how the existing type discriminator works in Deck (no changes to be made there).

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/config/AccountDefinitionConfiguration.java
```diff
@@ -53,6 +65,21 @@
 @Log4j2
 public class AccountDefinitionConfiguration {
 
+  @Bean
+  @ConditionalOnMissingBean
+  public AccountDefinitionAuthorizer accountDefinitionAuthorizer(
+      Optional<FiatPermissionEvaluator> maybePermissionEvaluator,
+      @Value("${services.fiat.enabled:false}") boolean isFiatEnabled) {
```
dbyron-sf, line 72: I'd love to pass a configuration object here instead of `@Value` but I'm not sure such an object exists today...

dbyron-sf (approved):

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/config/AccountDefinitionConfiguration.java
jvz: Well, there's the object where the evaluator comes from, but that wasn't enough of a hint here.

PR #5668, opened 2022-03-22
chore(cats): make relationship log info instead of warn

dogonthehorizon:
As the number of ECS providers increases the chattiness of this
message can creates upwards of 400GB/day in data. We do not believe
there is an adverse effect in changing the log level here. Info messages
can be safely turned off for those concerned with log storage costs.


jasonmcintosh (approved):

mattgogerly:
Is there any merit in making it a debug message? Not clued up on what the message actually means, but if it's logged constantly it seems like it's not that valuable from an operations perspective?

PR #5680, opened 2022-04-06
fix(sql): active tasks listing for graceful shutdown

st8f:
Fix active tasks listing which is used for graceful shutdown.

First problem is that we're comparing a.task_id with constant literal 't.id' instead of read field value. So results are always empty.

Second is that we're fetching results by field name with table alias (t.task_id) which is not working, at least on standard mysql 5.7 (Field (null) is not contained in Row). Fixed this by explicitly selecting required field (a.task_id) instead of all fields.

### DIFF
--- a/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/SqlTaskRepository.kt
+++ b/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/SqlTaskRepository.kt
@@ -284,7 +284,7 @@ class SqlTaskRepository(
    */
   private fun runningTaskIds(ctx: DSLContext, thisInstance: Boolean): Array<String> {
     return withPool(poolName) {
-      val baseQuery = ctx.select()
+      val baseQuery = ctx.select(field("a.task_id"))
         .from(taskStatesTable.`as`("a"))
         .innerJoin(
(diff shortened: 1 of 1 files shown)


clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/SqlTaskRepository.kt
```diff
@@ -295,7 +295,7 @@ class SqlTaskRepository(
 
       val select = if (thisInstance) {
         baseQuery
-          .innerJoin(tasksTable.`as`("t")).on(field("a.task_id").eq("t.id"))
+          .innerJoin(tasksTable.`as`("t")).on(sql("a.task_id = t.id"))
```
mattgogerly, line 298: `.on(field("a.task_id").eq(field("t.id"))`?

st8f: Yes, that was the first thing i wrote. But it has:
`Overload resolution ambiguity: 
public abstract fun eq(p0: Any!): Condition! defined in org.jooq.Field
public abstract fun eq(p0: Field<Any!>!): Condition! defined in org.jooq.Field`
And I don't have a good idea how to fix it. 

Also we already have `on(sql("a.task_id = b.task_id and a.created_at = b.created"))` in this query and maybe for the same reason.

PR #5688, opened 2022-04-13
fix(dependency): Issue of missing javax.validation:validation-api dependency while upgrading the spring cloud to Hoxton.SR12 in kork

j-sandy:
While upgrading spring cloud to Hoxton.SR12, we encountered below issues :

```
/clouddriver/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/AbstractAtomicOperationsCredentialsConverter.java:27: error: package javax.validation.constraints does not exist
import javax.validation.constraints.NotNull;
                                   ^
/clouddriver/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/AbstractAtomicOperationsCredentialsConverter.java:44: error: cannot find symbol
  public T getCredentialsObject(@NotNull final String name) {
                                 ^
  symbol:   class NotNull
  location: class AbstractAtomicOperationsCredentialsConverter<T>
  where T is a type-variable:
    T extends AccountCredentials<?> declared in class AbstractAtomicOperationsCredentialsConverter
/clouddriver/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/AbstractAtomicOperationsCredentialsConverter.java:43: error: cannot find symbol
  @NotNull
   ^
  symbol:   class NotNull
  location: class AbstractAtomicOperationsCredentialsConverter<T>
  where T is a type-variable:
    T extends AccountCredentials<?> declared in class AbstractAtomicOperationsCredentialsConverter
3 errors
```
The root cause is missing javax.validation:validation-api dependency, the reason being upgrade of transitive dependency resilience4j from [1.0.0](https://mvnrepository.com/artifact/io.github.resilience4j/resilience4j-framework-common/1.0.0) to [1.7.0](https://mvnrepository.com/artifact/io.github.resilience4j/resilience4j-framework-common/1.7.0) while upgrade to Hoxton.SR12.

To fix this issue we require to explicitly implement the dependency in clouddriver-core.gradle.

### DIFF
--- a/clouddriver-core/clouddriver-core.gradle
+++ b/clouddriver-core/clouddriver-core.gradle
@@ -17,6 +17,7 @@ dependencies {
   api "org.codehaus.groovy:groovy"
 
   implementation "javax.inject:javax.inject:1"
+  implementation "javax.validation:validation-api"
   implementation "net.logstash.logback:logstash-logback-encoder"
   implementation "com.fasterxml.jackson.module:jackson-module-kotlin"
   implementation "com.fasterxml.jackson.datatype:jackson-datatype-joda"
--- a/clouddriver-sql/clouddriver-sql.gradle
+++ b/clouddriver-sql/clouddriver-sql.gradle
@@ -32,6 +32,7 @@ dependencies {
   implementation "org.hibernate.validator:hibernate-validator"
   implementation "com.fasterxml.jackson.core:jackson-annotations"
   implementation "com.fasterxml.jackson.core:jackson-databind"
+  implementation "javax.validation:validation-api"
 
   testImplementation project(":clouddriver-core-tck")
 


dbyron-sf:
I also see javax.validation used in clouddriver-sql [here](https://github.com/spinnaker/clouddriver/blob/7845c38802a156e516fd31968a1c524a17e5cd0a/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/config/SqlEventCleanupAgentConfigProperties.kt#L19), so please add 
```
implementation "javax.validation:validation-api"
```
to [clouddriver-sql.gradle](https://github.com/spinnaker/clouddriver/blob/7845c38802a156e516fd31968a1c524a17e5cd0a/clouddriver-sql/clouddriver-sql.gradle) as well.

j-sandy:
org.hibernate.validator:hibernate-validator is already part of [clouddriver-sql.gradle](https://github.com/spinnaker/clouddriver/blob/7845c38802a156e516fd31968a1c524a17e5cd0a/clouddriver-sql/clouddriver-sql.gradle#L32). So, it seems we may not require javax.validation:validation-api.

dbyron-sf:
> org.hibernate.validator:hibernate-validator is already part of [clouddriver-sql.gradle](https://github.com/spinnaker/clouddriver/blob/7845c38802a156e516fd31968a1c524a17e5cd0a/clouddriver-sql/clouddriver-sql.gradle#L32). So, it seems we may not require javax.validation:validation-api.

We may not require it, but in the spirit of being explicit about dependencies, I'd do it.

dbyron-sf (approved):

PR #5692, opened 2022-04-22
fix(cf): scale app before building droplet

mattgogerly:
Currently we set process config and scale the app/processes after building the droplet. This causes issues if the droplet requires more memory to build (e.g. lots of Python dependencies to install).

`cf push` uses a v2 endpoint to create applications which allows specifying scale when creating the app, whereas we are using a v3 endpoint which does not have those fields. Scaling is therefore a separate call.

### DIFF
--- a/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
+++ b/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/client/Applications.java
@@ -656,8 +656,11 @@ public class Applications {
                     "Cloud Foundry signaled that package upload succeeded but failed to provide a response."));
   }
 
-  public String createBuild(String packageGuid) throws CloudFoundryApiException {
-    return safelyCall(() -> api.createBuild(new CreateBuild(packageGuid)))
+  public String createBuild(
+      String packageGuid, @Nullable Integer memoryAmount, @Nullable Integer diskSizeAmount)
+      throws CloudFoundryApiException {
+    return safelyCall(
+            () -> api.createBuild(new CreateBuild(packageGuid, memoryAmount, diskSizeAmount)))
         .map(Build::getGuid)
         .orElseThrow(
             () ->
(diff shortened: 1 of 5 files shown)


zachsmith1:
I'd recommend doing a push with the same Python stuff via the CLI and with the verbose flag. Also check the sequence of api calls. I vaguely remember we had to do it this way but we should prob do whatever the CLI does

mattgogerly:
> I'd recommend doing a push with the same Python stuff via the CLI and with the verbose flag. Also check the sequence of api calls. I vaguely remember we had to do it this way but we should prob do whatever the CLI does

The CF CLI is [calling](https://github.com/cloudfoundry/cli/blob/d3bb1ecfc94ee12432f8defb30ab9459e83cc727/cf/api/applications/applications.go#L44) POST /v2/apps [which accepts the full app definition (memory, disk, build packs etc).](https://apidocs.cloudfoundry.org/16.22.0/apps/creating_an_app.html) rather than v3, which only accepts the app name and lifecycle.

We've built an image with this change internally and we're able to deploy as normal with no issues, so if you have any idea what the specific problem was that we can test that'd be helpful!

nemesisOsorio:
@mattgogerly we checked the cli calls using version `8.3.0` this version is using API `v3`
we noticed the cli is calling `POST /v3/packages` follow by `POST /v3/builds`(droplet)
we used the command `cf push nginx0001 -v --docker-image nginx:alpine`
and then we scaled with `cf scale nginx0001 -i 3`
we are worried about the order of the api calls
just to clarify can you confirm that you deployed a docker image using these changes?
as well can you provide us an example of your pipeline?

mattgogerly:
I don't have any way to test Docker stuff, which I assume means the issue @zachsmith1 pointed out is limited to that. Nonetheless, thanks for checking - I double checked and the /v3/builds endpoint has parameters for setting build sizing.

I've updated the PR to use those and reverted the ordering 👍

nemesisOsorio (approved):
LGTM

link108:
@mergifyio backport release-1.28.x

PR #5702, opened 2022-04-28
feat(cats/sql): add ability to specify an account name regex that an …

mattgogerly:
…instance can schedule caching agents for

Adds an optional `cache-sharding.account-name-regex` config that will allow us to specify a regular expression for account names, for which agents can be scheduled on this replica. Part of https://github.com/spinnaker/spinnaker/issues/6651.

Idea being to have a Clouddriver cluster in one location caching certain accounts, and a cluster in another location caching others, but obviously this is flexible.


apoorvmahajan:
How would this be different than `sql.agent.enabled-pattern` that already exists today? See [this for more details](https://docs.armory.io/armory-enterprise/armory-admin/caching-agents-config/#sql-global-caching-agents-configuration) and [usages of this property](https://github.com/spinnaker/clouddriver/blob/master/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cluster/SqlClusteredAgentScheduler.kt#L176)

PR #5705, opened 2022-05-11
fix(core): remove duplicate LoggingInstrumentation, MetricInstrumenta…

dbyron-sf:
…tion and CatsOnDemandCacheUpdater beans

CacheConfig has @Bean methods that construct these, so annotating each class with @Component creates duplicates.

clouddriver's Main class has
```
    System.setProperty("spring.main.allow-bean-definition-overriding", "true")
```
which is likely what keeps things working.


link108 (approved):
LGTM 👍

PR #5703, opened 2022-05-09
refactor(groovy): migrate things to Java

mattgogerly:
Groovy 👎 
Java 👍

PR #5716, opened 2022-05-20
fix(web): copy the MDC to threads that service async endpoints

dbyron-sf:
to facilitate troubleshooting.  For example, logging that's configured to include X-SPINNAKER* request headers via the MDC and AuthenticatedRequestFilter now has something to log.

Depends on https://github.com/spinnaker/kork/pull/959.


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/WebConfig.java
```diff
@@ -0,0 +1,110 @@
+/*
```
dbyron-sf, line 1: @mattgogerly I saw https://github.com/spinnaker/clouddriver/pull/5703 with some other groovy --> java, but I didn't see this file so I think there's no conflict/race.

dbyron-sf:
@mergifyio update

jvz (approved):

PR #5718, opened 2022-05-25
fix(google): Added scope to the credentials

kirangodishala:
This is to address the issue [#6661](https://github.com/spinnaker/spinnaker/issues/6661)

Without scope, application default credentials results in authenticate failure. This PR fixes the issue by adding the scope.


sanopsmx (approved):
lgtm

PR #5719, opened 2022-05-25
feat(core): User secrets and account upsert

jvz:
- Adds the AccountDefinitionRepository::save API for upserting accounts along with changing the default PUT behavior on credentials to use this (while introducing a dedicated PUT endpoint for updating an existing account).
- Adds the AccessControlledAccountDefinition interface to provide a common interface for account definitions that have permissions defined.
- Adds AccountDefinitionTypeProvider to allow plugins to introduce credentials types.
- Renames AccountDefinitionAuthorizer to AccountSecurityPolicy.
- Adds AuthorizedRolesExtractor bean.
- Prepares for account pubsub API support and beta availability of accounts storage feature.


clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionSecretManager.java
```diff
@@ -16,34 +16,72 @@
...
     }
-    var userRoles = authorizer.getRoles(username);
+    var userRoles = policy.getRoles(username);
     if (userRoles.isEmpty()) {
       return false;
     }
-    // TODO(jvz): update with https://github.com/spinnaker/kork/pull/942
-    //  to add user secrets usage tracking for time of use authz checks
-    return authorizer.canAccessAccount(username, accountName);
+    if (refsByAccountName.getOrDefault(accountName, Set.of()).stream()
```
dbyron-sf, line 80: I could use some comments here...mostly a reminder that Collections.disjoint returns true if there are no common elements...meaning that none of the userRoles provide access to the secret.

clouddriver-core/src/test/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionMapperTest.java
```diff
@@ -53,9 +54,15 @@ void canConvertAdditionalAccountTypes() throws JsonProcessingException {
...
 
+  @Test
+  void canConvertJacksonizedAccountTypes() throws JsonProcessingException {
+    var account = ValueAccount.builder().name("james").value("meowth").build();
+    assertEquals(account, mapper.deserialize(mapper.serialize(account)));
+  }
+
   @Test
   void canDecryptSecretUris() {
-    var data = "{\"type\":\"test\",\"name\":\"bar\",\"password\":\"encrypted:noop!v:hunter2\"}";
```
dbyron-sf, line 58: It's important (or at least makes backwards-compatibility/upgrading easier) for "the old way" to still work if at all possible.

clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionRepository.java
```diff
@@ -62,18 +62,26 @@ List<? extends CredentialsDefinition> listByType(
 
   /**
    * Creates a new account definition using the provided data. Secrets should use {@code
-   * SecretEngine} encrypted URIs (e.g., {@code
-   * encrypted:secrets-manager!r:us-west-2!s:my-account-credentials}) when the underlying storage
+   * UserSecretReference} encrypted URIs (e.g., {@code
+   * secret://secrets-manager?r=us-west-2&s=my-account-credentials}) when the underlying storage
```
dbyron-sf, line 66: At least for now, describe both `encrypted:` and `secret://` and perhaps describe a migration path?

dbyron-sf: ...and an end-of-life plan for `encrypted:` if possible.

clouddriver-core/src/test/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionMapperTest.java
```diff
@@ -53,9 +54,15 @@ void canConvertAdditionalAccountTypes() throws JsonProcessingException {
...
+    var data = "{\"type\":\"test\",\"name\":\"bar\",\"password\":\"secret://noop?v=hunter2&k=v\"}";
```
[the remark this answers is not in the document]
jvz: I realized why this is here: the noop engine parses this as a map with `v: hunter2` and `k: v` where the `k` parameter tells us which key to look up from the map (which is `v` in this case which is `hunter2`).

clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionRepository.java
jvz: Re-added support for this. Still need an EOL plan, though not immediately urgent. Or we could keep this as an optional feature since it sort of matches the existing behavior with Spring Cloud Config encrypted properties.

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/DefaultAccountSecurityPolicy.java
```diff
@@ -0,0 +1,84 @@
...
+  @Override
+  public Set<String> getRoles(String username) {
+    return Optional.ofNullable(permissionEvaluator.getPermission(username)).stream()
+        .flatMap(permission -> permission.getRoles().stream().map(Role.View::getName))
+        .collect(Collectors.toSet());
+  }
+
+  @Override
+  public boolean canAccessAccount(@Nonnull String username, @Nonnull String account) {
+    return Optional.ofNullable(permissionEvaluator.getPermission(username))
```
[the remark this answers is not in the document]
jvz: Renamed.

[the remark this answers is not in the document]
dbyron-sf: Thanks.  I think you explained this to me offline, but I've forgotten the details.  Why isn't Authorization.READ sufficient here?

clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionRepository.java
dbyron-sf: Thanks.  Does it make sense to include both `encrypted:` and `secret://` in the javadoc?

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/config/AccountDefinitionConfiguration.java
```diff
@@ -63,110 +68,112 @@
...
   @ConditionalOnMissingBean
-  public AccountDefinitionAuthorizer accountDefinitionAuthorizer(
-      Optional<FiatPermissionEvaluator> maybePermissionEvaluator,
-      @Value("${services.fiat.enabled:false}") boolean isFiatEnabled) {
-    return new AccountDefinitionAuthorizer(
-        maybePermissionEvaluator.filter(ignored -> isFiatEnabled).orElse(null));
+  public AccountSecurityPolicy accountSecurity(
+      @Nullable FiatPermissionEvaluator permissionEvaluator,
+      @Value("${services.fiat.enabled:false}") boolean fiatEnabled) {
+    return fiatEnabled && permissionEvaluator != null
```
dbyron-sf, line 81: Trying to wrap my head around when FiatPermissionEvaluator could be null...[CacheConfig](https://github.com/spinnaker/clouddriver/blob/0f1c3b83d2e8673bf4d545ab6f72445dfea988cb/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CacheConfig.groovy#L111) has it Optional which seems like a good precedent, though now that I've dug I don't think that's necessary.  It's a long and winding path, but..

* FiatPermissionEvaluator is annotated with [`@Component`](https://github.com/spinnaker/fiat/blob/8b6d86cea9ca40df3476f60a71dc1988c93b8ff6/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java#L58) and in a package (`com.netflix.spinnaker.fiat.shared`) that's `@ComponentScan`ned by [FiatAuthenticationConfig](https://github.com/spinnaker/fiat/blob/8b6d86cea9ca40df3476f60a71dc1988c93b8ff6/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java#L55).
* FiatAuthenticationConfig comes in via [EnableFiatAutoConfig](https://github.com/spinnaker/fiat/blob/8b6d86cea9ca40df3476f60a71dc1988c93b8ff6/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/EnableFiatAutoConfig.java#L43).
* clouddriver's SecurityConfig has [`@EnableFiatAutoConfig`](https://github.com/spinnaker/clouddriver/blob/0f1c3b83d2e8673bf4d545ab6f72445dfea988cb/clouddriver-security/src/main/groovy/com/netflix/spinnaker/clouddriver/security/config/SecurityConfig.groovy#L31).
* clouddriver's Main imports [SecurityConfig](https://github.com/spinnaker/clouddriver/blob/0f1c3b83d2e8673bf4d545ab6f72445dfea988cb/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy#L46).

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/DefaultAccountSecurityPolicy.java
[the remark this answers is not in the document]
jvz: The READ permission is only used with accounts for linked resources like server groups, firewalls, etc. The READ permission allows you to view those related resources in the UI, but you can't deploy to them or modify them or anything.

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/config/AccountDefinitionConfiguration.java
jvz: IntelliJ complains about this a bit, though yeah, this is typically available at least even if Fiat is disabled since that component short circuits its APIs when Fiat is disabled. I think I used `@Nullable` here instead of `Optional` as, again, IntelliJ static code analysis complains about using `Optional` in parameters and fields, and given our annotation processing, you get the same benefits from using `@Nullable` as `Optional` in such use cases (and Spring supports both).

I think this could be refactored a little to work more like `FiatPermissionEvaluator` where it checks `FiatStatus` for whether or not Fiat is enabled rather than a startup configuration bean.

dbyron-sf: Optional vs. Nullable isn't so interesting for me.  I did try taking it out locally and ran into a bunch of struggles providing the right beans in some tests.  You may have an easier time with it than me.  It's also OK to punt on this for this PR if it's too painful.

clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/DefaultAccountSecurityPolicy.java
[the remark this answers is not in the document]
dbyron-sf: Thanks: https://github.com/spinnaker/clouddriver/pull/5719/files/9879aadf0a33a02e0742c266f3256fc1d89d1ecd..27eb955f9c90fe46de2ffa427b54a07b926f0e90

clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionRepository.java
dbyron-sf: Done in https://github.com/spinnaker/clouddriver/pull/5719/files/9879aadf0a33a02e0742c266f3256fc1d89d1ecd..27eb955f9c90fe46de2ffa427b54a07b926f0e90

PR #5726, opened 2022-06-03
fix(core): ignore roles for anonymous user to access account

nemesisOsorio:
If you enable Fiat and if you do not set permissions to accounts in Clouddriver the automatic triggers fail with permission denied.
image

before this change, https://github.com/spinnaker/clouddriver/pull/5666 anonymous user had permissions to access Clouddriver accounts

### DIFF
--- a/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionSecretManager.java
+++ b/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionSecretManager.java
@@ -38,10 +38,11 @@ public class AccountDefinitionSecretManager {
     if (authorizer.isAdmin(username)) {
       return true;
     }
-    var userRoles = authorizer.getRoles(username);
-    if (userRoles.isEmpty()) {
+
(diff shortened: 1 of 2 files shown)


edgarulg (approved):
LGTM

link108 (approved):
Thank you @nemesisOsorio !

link108:
@mergifyio backport release-1.28.x

jvz:
Doesn't this defeat the purpose of authorization enforcement?

xibz:
Uhh, guys? This just gave so much power to anonymous users lol

jvz:
Is the idea here that if both the user and account have no permissions defined, then the action should be allowed? Because if the idea here is that `anonymous` should have admin permissions, then that should be handled differently (and not enabled by default).

xibz:
Nvm, I see what's going on now. I think the mix up was I assumed this gave anonymous user permissions to run anything.

jvz:
I've got a more thorough fix for this in my upcoming PR that will make this work in the scenario where both the user and the account have no permissions defined (which includes `anonymous` as one such user of course).

link108:
@jvz awesome, thank you! The previous set of changes broke some of our test pipelines, hence this PR

PR #5729, opened 2022-06-06
fix(provider/azure): Fix scale health

kmccarp:
Before this change, if a scale set was unhealthy at the outset and became healthy after a period of time, the instance state never got updated, so Spinnaker always saw it was unhealthy.


zieka (approved):

jkschneider (approved):
LGTM

PR #5733, opened 2022-06-14
fix(gitrepo): Support to authenticate with user and token

nemesisOsorio:
Cloning Bitbucket is failing for git repo artifact using username and token
image

After this change we tested different scenarios including Github
image


ovidiupopa07:
LGTM

link108 (approved):

PR #5740, opened 2022-06-17
fix(azure): increase health interval

zieka:
#### Problem
- Azure healthy instance checks create many calls to the resource pages on Azure ultimately resulting in "Too many requests" errors on the portal pages.

#### Solution
- increase the timeout to 5 seconds


kmccarp (approved):

jkschneider (approved):

jkschneider:
This is understood to be a temporary fix that prevents Azure API rate limiting from being triggered, and a follow up PR is expected.

PR #5741, opened 2022-06-17
config(web): explicit qualifier bean for AsyncTaskExecutor in WebConfig

nemesisOsorio:
we could have more beans of type `AsyncTaskExecutor`
for example, if you enable Config server using [Vault](https://github.com/spring-projects/spring-vault/blob/dd9c3cfab8aca6ee4572f9c81b9a02000b2db480/spring-vault-core/src/main/java/org/springframework/vault/config/AbstractVaultConfiguration.java#L186)


armory-abedonik (approved):

mrampton:
@Mergifyio refresh

nemesisOsorio:
seems like there was an attempt to fix the test that is failing
https://github.com/spinnaker/clouddriver/pull/5621

PR #5742, opened 2022-06-21
fix(caching): K8s cache affected when not enough permissions in service account

danielGz:
k8s-kind listing is able to return partial results when not enough permissions in the service account exist but because we are throwing an error even partial entries are not being cached.
This fix will not throw an error when it is permission related only for k8s caching.

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutor.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutor.java
@@ -453,8 +453,14 @@ public class KubectlJobExecutor {
             parseManifestList());
 
     if (status.getResult() != JobResult.Result.SUCCESS) {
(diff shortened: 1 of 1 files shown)


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutor.java
```diff
@@ -453,8 +453,14 @@ public ImmutableList<KubernetesManifest> list(
             parseManifestList());
 
     if (status.getResult() != JobResult.Result.SUCCESS) {
-      throw new KubectlException(
-          "Failed to read " + kinds + " from " + namespace + ": " + status.getError());
+      boolean permissionError =
```
mattgogerly, line 456: Is using Apache Commons necessary? The next if block below just does a simple contains

link108 (approved):

danielGz:
@mergifyio backport release-1.28.x release-1.27.x release-1.26.x

PR #5769, opened 2022-08-18
feat(provider/oracle): Add user agent header to OCI clients.

shivalish:
This change adds the user agent header to OCI clients so that Spinnaker calls to OCI can be tracked.


mattgogerly:
Thanks for this @shivalish! Mind fixing the commit message so the bot is happy?

PR #5778, opened 2022-08-30
feat(kubernetes): add support for retrieving run job logs from a specific pod

apoorvmahajan:
Spinnaker doesn't seem to handle retrieving property file contents properly when a job may have multiple pods. Under the covers, it runs the `kubectl logs jobs` command, which doesn't give the right answer if there are two pods running at the same time for the same job. In such cases, kubernetes always defaults to the first pod, but in certain scenarios, we would want to query the second pod for logs as the first one may have failed but the second one may have succeeded.

This PR enables adding a new endpoint to directly query property file contents from a pod, i.e. do `kubectl logs pod/<pod-name>` under the covers.

it is the responsibility of the client (orca in this case) to supply the correct pod name.

### DIFF
--- a/clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesJobProviderSpec.groovy
+++ b/clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesJobProviderSpec.groovy
@@ -20,6 +20,7 @@ package com.netflix.spinnaker.clouddriver.kubernetes.provider.view
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.view.model.KubernetesManifestContainer
 import com.netflix.spinnaker.clouddriver.kubernetes.caching.view.provider.KubernetesManifestProvider
 import com.netflix.spinnaker.clouddriver.kubernetes.description.manifest.KubernetesManifest
+import com.netflix.spinnaker.clouddriver.kubernetes.op.job.KubectlJobExecutor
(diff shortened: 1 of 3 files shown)


clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesJobProviderSpec.groovy
```diff
@@ -108,10 +108,65 @@ class KubernetesJobProviderSpec extends Specification {
     }
 
     when:
-    def provider = new KubernetesJobProvider(mockAccountCredentialsProvider, mockManifestProvider)
+    def provider = new KubernetesJobProvider(mockAccountCredentialsProvider, mockManifestProvider, true)
```
dbyron-sf, line 111: What's the extra `true` for?

apoorvmahajan: The [KubernetesJobProvider constructor](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/KubernetesJobProvider.java#L52-L55) has three parameters - `true` is the default value for the `detailedPodStatus` parameter.

dbyron-sf: Aaah, thanks.

dbyron-sf (approved):

PR #5780, opened 2022-09-09
chore(aws/integration): make it easier to diagnose failures in CreateServerGroupSpec

dbyron-sf:


clouddriver-aws/src/integration/java/com/netflix/spinnaker/clouddriver/aws/test/CreateServerGroupSpec.java
```diff
@@ -87,7 +89,9 @@ public class CreateServerGroupSpec extends AwsBaseSpec {
   private AmazonEC2 mockEc2 = mock(AmazonEC2.class);
 
   @BeforeEach
-  public void setup() {
+  private void init(TestInfo testInfo) {
+    System.out.println("--------------- Test " + testInfo.getDisplayName());
```
jasonmcintosh, line 93: System.out?  Would... ideally be a logger log I'd think...

jasonmcintosh (approved):
Would welcome use of log vs. System.out.println but ... can go for it and be ok for a test class.

clouddriver-aws/src/integration/java/com/netflix/spinnaker/clouddriver/aws/test/CreateServerGroupSpec.java
dbyron-sf: Perhaps, but this way we still see the messages in the face of logging troubles.

PR #5797, opened 2022-10-13
chore(kubernetes): stop specifying the version of io.kubernetes:client-java

dbyron-sf:
since spring 2.4.13 brings it in (transitively).  The version from spring boot takes precedence, so the value we specify here is ignored.

Another approach would be to make the version here match what spring boot brings in (currently 11.0.3), but then future changes could again invalidate what's here.  If some future version of spring boot stops bringing in io.kubernetes:client-java, builds will fail with unspecified version, at which point we can restore kubernetesClientVersion.

Snippets of ./gradlew clouddriver-web:dependencies

before:
```
+--- project :clouddriver-kubernetes
|    +--- io.kubernetes:client-java:11.0.1 -> 11.0.3
```
after:
```
+--- project :clouddriver-kubernetes
|    +--- io.kubernetes:client-java -> 11.0.3
```
Related orca PR: https://github.com/spinnaker/orca/pull/4310

### DIFF
--- a/clouddriver-kubernetes/clouddriver-kubernetes.gradle
+++ b/clouddriver-kubernetes/clouddriver-kubernetes.gradle
@@ -80,7 +80,7 @@ dependencies {
   implementation "io.spinnaker.kork:kork-moniker"
   implementation "io.spinnaker.kork:kork-secrets"
   implementation "io.spinnaker.kork:kork-security"
-  implementation "io.kubernetes:client-java:$kubernetesClientVersion"
+  implementation "io.kubernetes:client-java"
   implementation "org.apache.commons:commons-lang3"
   implementation "org.springframework.boot:spring-boot-actuator"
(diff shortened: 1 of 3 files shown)


mattgogerly (approved):

j-sandy:
Adding transitive path detail of `io.kubernetes:client-java` for future reference:
[org.springframework.cloud:spring-cloud-dependencies:2020.0.6 ](https://repo1.maven.org/maven2/org/springframework/cloud/spring-cloud-dependencies/2020.0.6/spring-cloud-dependencies-2020.0.6.pom)---> [org.springframework.cloud:spring-cloud-kubernetes-dependencies:2.0.6](https://repo1.maven.org/maven2/org/springframework/cloud/spring-cloud-kubernetes-dependencies/2.0.6/spring-cloud-kubernetes-dependencies-2.0.6.pom) ---> io.kubernetes:client-java:11.0.4

PR #5803, opened 2022-10-28
fix(provider/azure): Fix CreateAzureServerGroupWithAzureLoadBalancerAtomicOperation and DestroyAzureServerGroupAtomicOperation after migrating to latest Azure SDK

ovidiupopa07:
After the migration of the SDK to com.azure.resourcemanager:azure-resourcemanager:2.19.0 `CreateAzureServerGroupWithAzureLoadBalancerAtomicOperation` started to fail because the `com.netflix.spinnaker.clouddriver.azure.client.AzureNetworkClient#enableServerGroupWithLoadBalancer `function is not able to update the load-balancing rule

The reason is that the old SDK, defines only one `backendAddressPool` property in `LoadBalancingRuleInner`. But with the new SDK, two properties are defined `backendAddressPool & backendAddressPools`.  The atomic operation updates the `backendAddressPool` to point to the new server group, while the list contains the default backend pool.  Because of this, Azure API is throwing `LoadBalancingRuleBackendAdressPoolAndBackendAddressPoolsCannotBeSetAtTheSameTimeWithDifferentValue` exception. The fix is to set both properties to use the same value.

Closes https://github.com/spinnaker/spinnaker/issues/6736

### DIFF
--- a/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureNetworkClient.groovy
+++ b/clouddriver-azure/src/main/groovy/com/netflix/spinnaker/clouddriver/azure/client/AzureNetworkClient.groovy
@@ -705,6 +705,8 @@ class AzureNetworkClient extends AzureBaseClient {
           Method setRangeMethod = o.getClass().getMethod("toBackend", String.class)
           setRangeMethod.setAccessible(true)
           setRangeMethod.invoke(o, serverGroupName)
+          LoadBalancingRule lbrule = o as LoadBalancingRule;
+          lbrule.innerModel().withBackendAddressPools(List.of(lbrule.innerModel().backendAddressPool()))
         } catch (NoSuchMethodException e) {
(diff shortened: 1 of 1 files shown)


dbyron-sf:
Is it possible to add a test for this?

ovidiupopa07:
> Is it possible to add a test for this?

I don't see an easy way to unit test the Azure Clients (`AzureComputeClient`, `AzureNetworkClient`, `AzureResourceManagerClient`) without refactoring how the clients are initialized right now.  Currently, all clients are extending `AzureBaseClient`.  The `AzureBaseClient` is a wrapper for the `AzureResourceManager`. 

I plan to do a refactoring: change the relationship from inheritance to composition.  It should make the code more test friendly than it is right now.

dbyron-sf:
Fair enough, thanks for the explanation.

sergio-quintero (approved):

mattgogerly (approved):

dogonthehorizon (approved):

PR #5814, opened 2022-11-17
fix(kubernetes): teach KubernetesManifest to support kubernetes resources where the spec is not a map

dbyron-sf:
Before this change, attempts to deploy a manifest where the spec was a list, e.g.
```
apiVersion: test.example/v1alpha1
kind: Custom1
metadata:
  name: default-custom1
  labels:
    app: app1
spec:
  - id: my-id
    description: my-description
```
would result in
image

With these changes, spinnaker is able to deploy it.


jvz (approved):

dbyron-sf:
@mergifyio backport release-1.27.x release-1.28.x release-1.29.x

PR #5840, opened 2022-12-09
chore(dependencies): remove dependency on groovy-all where straightforward

dbyron-sf:
with a specific goal to get org.testng:testng:7.4.0 out of shipping code, since it's vulnerable to CVE-2022-4065.

Note: groovy-all and testng remain in clouddriver-yandex, but they're gone (besides test configurations) elsewhere.


cats/cats-core/cats-core.gradle
```diff
@@ -3,7 +3,7 @@ dependencies {
 
   implementation "org.slf4j:slf4j-api"
   implementation "com.fasterxml.jackson.core:jackson-annotations"
-  implementation "org.codehaus.groovy:groovy-all"
+  implementation "org.codehaus.groovy:groovy"
```
jasonmcintosh, line 6: IF we change this to API, it should expose it downstream (aka cats-core being a java librayr, no need for implementation downstream)

jasonmcintosh (approved):
Going ahead and approving - can discuss dep locations/transitive deps/etc. in future, as it isn't wrong to be explicit on the dependency on groovy.

PR #5841, opened 2022-12-14
chore(dependency): Introducing kork-cloud-config-server dependency in clouddriver-web due to expected kork refactoring

j-sandy:
`com.netflix.spinnaker.kork.configserver.ConfigServerBootstrap` contains all the relevant cloud config server properties to bootstrap and manage the embedded config server. This class is part of kork-config module along with com.netflix.spinnaker.kork.boot.DefaultPropertiesBuilder. However, ConfigServerBootstrap class belongs to kork-cloud-config-server. This refactor will break clouddriver-web and require to introduce kork-cloud-config-server as dependency.


dbyron-sf:
Included in https://github.com/spinnaker/clouddriver/pull/5842

PR #5846, opened 2023-01-04
feat(kubernetes): add visibility to KubectlJobExecutor by adding logs…

dbyron-sf:
… and extending the task object to include stdout and stderr info

This adds a new database table named task_outputs to store stdout and stderr from kubectl when this functionality is enabled.  The existing SqlTaskCleanupAgent cleans it up, as it does the other related tables (i.e. tasks, task_states, task_results).

This adds two new configuration parameters that default to false:

- kubernetes.jobExecutor.persistTaskOutput
- kubernetes.jobExecutor.enableTaskOutputForAllAccounts

When kubernetes.jobExecutor.persistTaskOutput is true, KubectlJobExecutor persists stderr and stdout of kubectl for mutating kubectl operations: create, delete, deploy, patch, replace, rollout resume, rollout and scale when kubernetes.jobExecutor.enableTaskOutputForAllAccounts is true, or for accounts that have their debug property set to true.  Since this information is part of the task object, it's available via e.g. the GET /task/{id} endpoint.

Consumed by https://github.com/spinnaker/orca/pull/4374.


jvz (approved):
Looks useful!

PR #5849, opened 2023-01-06
fix(google): added null check for autoscaler custom metric scaling policies

SusmithaGundu:
To address the validation issue that is occurring while using custom metric utilisations in Autoscaler feature of creation of server group in google cloud provider.
As part of addressing the issue mentioned in [6784](https://github.com/spinnaker/spinnaker/issues/6784)

While selecting one of two scaling policies available(SingleInstanceAssignment and Utilization Target) the other one is null.Hence added validation to check for null.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/validators/StandardGceAttributeValidator.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/validators/StandardGceAttributeValidator.groovy
@@ -474,14 +474,16 @@ class StandardGceAttributeValidator {
           utilization.with {
             validateNotEmpty(metric, "${path}.metric")
 
-            if (utilizationTarget <= 0) {
+            if (utilizationTarget!=null && utilizationTarget <= 0) {
               errors.rejectValue("${context}.${path}.utilizationTarget",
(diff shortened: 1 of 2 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/validators/StandardGceAttributeValidator.groovy
```diff
@@ -474,14 +474,14 @@ class StandardGceAttributeValidator {
...
             validateNotEmpty(metric, "${path}.metric")
 
-            if (utilizationTarget <= 0) {
+            if (utilizationTarget!=null && utilizationTarget <= 0) {
               errors.rejectValue("${context}.${path}.utilizationTarget",
                 "${context}.${path}.utilizationTarget must be greater than zero.")
-            }
 
-            validateNotEmpty(utilizationTargetType, "${path}.utilizationTargetType")
+              validateNotEmpty(utilizationTargetType, "${path}.utilizationTargetType")
```
jasonmcintosh, line 481: You moved the validate not empty under the if check... I'm not sure if that was intentional?  That'd change I think the behavior on this.... Example:
* If utilization target is 5... the validate not empty check would no longer fire.
* That MAY be... irrelevant, but maybe not, not as familiar with the impact on this...

mattgogerly (approved):

PR #5854, opened 2023-01-13
fix(kubectl): add metrics to retries, make log messages more descriptive and support retries for all kubectl actions

dbyron-sf:
Turns out the changes from https://github.com/spinnaker/clouddriver/pull/5401 needed some adjustment.

## Purpose

* add tests to show duplicate logs aren't being printed when multiple threads are running the same command
* add metrics support for capturing retries.
* refactor criteria that decides if an error should be retried. This is needed so that metrics are reported correctly
* refactor tests
* add short custom log messages for retries instead of the default verbose ones

Here are some log messages showing that retries help in fixing the problem in subsequent attempts:
```
{"@timestamp":"2021-11-12T16:55:15.987+00:00","@version":1,"message":"Retrying kubectl command for my-account Attempt #1 failed with exception: com.netflix.spinnaker.clouddriver.kubernetes.op.job.KubectlJobExecutor$KubectlException: command: 'kubectl --namespace=test -o json get job my-job' in account: my-account failed. Error: Unable to connect to the server: net/http: TLS handshake timeout\n","logger_name":"com.netflix.spinnaker.clouddriver.kubernetes.op.job.KubectlJobExecutor","thread_name":"http-nio-7002-exec-13","level":"INFO","level_value":20000}
{"@timestamp":"2021-11-12T16:55:32.037+00:00","@version":1,"message":"Kubectl command for my-account is now successful in attempt #2. Last attempt had failed with exception: com.netflix.spinnaker.clouddriver.kubernetes.op.job.KubectlJobExecutor$KubectlException: command: 'kubectl --namespace=test -o json get job my-job' in account: my-account failed. Error: Unable to connect to the server: net/http: TLS handshake timeout\n","logger_name":"com.netflix.spinnaker.clouddriver.kubernetes.op.job.KubectlJobExecutor","thread_name":"http-nio-7002-exec-13","level":"INFO","level_value":20000}
```

### DIFF
--- a/clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutorTest.java
+++ b/clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutorTest.java
@@ -19,17 +19,22 @@ package com.netflix.spinnaker.clouddriver.kubernetes.op.job;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertThrows;
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.LoggerContext;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.read.ListAppender;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSetMultimap;
 import com.google.common.io.Resources;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.netflix.spinnaker.clouddriver.jobs.JobExecutionException;
 import com.netflix.spinnaker.clouddriver.jobs.JobExecutor;
 import com.netflix.spinnaker.clouddriver.jobs.JobRequest;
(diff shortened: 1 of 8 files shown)


clouddriver-kubernetes/src/test/resources/com/netflix/spinnaker/clouddriver/kubernetes/op/job/mock-kubectl-stdin-command.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+#
+# Copyright 2022 Salesforce, Inc.
```
mattgogerly, line 3: 2023? 😄 

LGTM though!

dbyron-sf: Yeah, we've been using this for awhile...

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutorTest.java
```diff
@@ -135,44 +181,251 @@ void topPodMultipleContainers() throws Exception {
...
+      int numberOfThreads) {
+    // setup
+    kubernetesConfigurationProperties.getJobExecutor().getRetries().setEnabled(true);
+
+    // to test log messages
+    Logger logger = (Logger) LoggerFactory.getLogger(KubectlJobExecutor.class);
+    ListAppender<ILoggingEvent> listAppender = new ListAppender<>();
+    listAppender.setContext((LoggerContext) LoggerFactory.getILoggerFactory());
+    logger.addAppender(listAppender);
+    listAppender.start();
```
jvz, line 227: I assume you're already following an established testing pattern in this file, but in the future, note that Log4j2 (log4j-core-test or tests, I forget the full name) provides a `ListAppender` in its test code which has some additional methods useful for waiting on async things to finish logging, etc., along with the whole `@LoggerContextSource` JUnit 5 extension for using test-specific logging configurations.

jvz (approved):
lgtm

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutorTest.java
dbyron-sf: https://github.com/spinnaker/clouddriver/pull/5854/commits/4d9e4877a6f3190561c93c9dc57ee7c5a10ef345 uses [MemoryAppender](https://github.com/spinnaker/kork/blob/2244cc04c5d6cb09462a412168bc2a47d2eb5636/kork-test/src/main/java/com/netflix/spinnaker/kork/test/log/MemoryAppender.java) from kork, which seems like an improvement.  I need to learn about these other mechanisms since I bet they're even better.

In an even less exciting discovery, none of these junit tests in clouddriver-kubernetes are actually running.  https://github.com/spinnaker/clouddriver/pull/5855 takes care of that.

PR #5856, opened 2023-01-17
fix(permissions): Added evaluator case when fiat is enabled

OscarMichelH:
When fiat is enabled and validation for accounts without roles is requested, the access is denied by emptyRoles condition, so when a dummy role is set this condition is skipped. Adding extra old fiat validator we preserved a consistent behaviour on accounts with roles vs without roles

Steps to reproduce issue:
With fiat enabled create 2 serviceAccounts, one with role, and other without role
image

Try apply ops operations like scale manifest deployment with different accounts, so we gonna have the inconsistent scenario where dummy roled account is succeeded  and empty role account received a denied access error

Account without roles:
image
image

Account with dummy roles:
image
image

Use case for empty role accounts was introduced in https://github.com/spinnaker/clouddriver/commit/da657157d6f8c7050ee577e147db99c9936b0b72
previously accounts without roles were working

### DIFF
--- a/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/deploy/DescriptionAuthorizerService.java
+++ b/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/deploy/DescriptionAuthorizerService.java
@@ -114,9 +114,7 @@ public class DescriptionAuthorizerService {
       && (
       (resourceTypes.contains(ResourceType.ACCOUNT)
         && !secretManager.canAccessAccountWithSecrets(account))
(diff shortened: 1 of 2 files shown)


OscarMichelH:
Note: Full solution for master was applied in https://github.com/spinnaker/clouddriver/commit/9f0a2ee58fc219f192546e78abcbe081e10570b8 with a mix of different features so it can not be backported

OscarMichelH:
@spinnakerbot update

OscarMichelH:
@spinnakerbot2 update

PR #5876, opened 2023-02-14
fix(kubernetes): Revert to using the dockerImage Artifact Replacer for cronjobs

apoorvmahajan:
PR #5554 updated the docker image artifact replacer for K8s Cronjobs. While it works if there is no tag defined in the manifest, it fails if there is a tag defined in the manifest and the artifact binding strategy is set to `match-name-and-tag`. 
Commit: https://github.com/spinnaker/clouddriver/pull/5876/commits/e55cbcb1769cbb6ec70bf0087697fcac9e9a0f37 adds a test that demonstrates that failure. With the artifact replacer set to `dockerImage()`, we don't see that failure.

### DIFF
--- a/clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/ReplacerTest.java
+++ b/clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/ReplacerTest.java
@@ -1273,7 +1273,7 @@ final class ReplacerTest {
   @Test
   void findCronJobDockerImages() {
     ArtifactReplacer artifactReplacer =
-        new ArtifactReplacer(ImmutableList.of(Replacer.cronJobDockerImage()));
+        new ArtifactReplacer(ImmutableList.of(Replacer.dockerImage()));
     KubernetesManifest cronJob = getCronJob();
 
     Set<Artifact> artifacts = artifactReplacer.findAll(cronJob);
@@ -1287,7 +1287,7 @@ final class ReplacerTest {
             artifact -> {
               assertThat(artifact).isNotNull();
               assertThat(artifact.getType()).isEqualTo("docker/image");
-              assertThat(artifact.getName()).isEqualTo("gcr.io/my-repository/my-image:my-tag");
+              assertThat(artifact.getName()).isEqualTo("gcr.io/my-repository/my-image");
               assertThat(artifact.getReference()).isEqualTo("gcr.io/my-repository/my-image:my-tag");
             });
 
@@ -1305,7 +1305,7 @@ final class ReplacerTest {
   @Test
   void replaceCronJobDockerImages() {
     ArtifactReplacer artifactReplacer =
-        new ArtifactReplacer(ImmutableList.of(Replacer.cronJobDockerImage()));
+        new ArtifactReplacer(ImmutableList.of(Replacer.dockerImage()));
     KubernetesManifest cronJob = getCronJob();
 
     Artifact inputArtifact =
@@ -1341,50 +1341,6 @@ final class ReplacerTest {
     Set<Artifact> artifacts = replaceResult.getBoundArtifacts();
     assertThat(artifacts).hasSize(1);
     assertThat(Iterables.getOnlyElement(artifacts)).isEqualTo(inputArtifact);
-  }
-
-  @ParameterizedTest(name = "{index} ==> with docker image artifact binding strategy = {0}")
-  @ValueSource(strings = {DEFAULT_DOCKER_IMAGE_BINDING, MATCH_NAME_ONLY_ARTIFACT_BINDING_STRATEGY})
-  void replaceCronJobDockerImageWithArtifactBindingStrategyUsingCronJobDockerImageReplacer(
-      String artifactBindingStrategy) {
-    ArtifactReplacer artifactReplacer =
-        new ArtifactReplacer(ImmutableList.of(Replacer.cronJobDockerImage()));
-    KubernetesManifest cronJob = getCronJobWithOneContainerWithImageTag();
-    Artifact inputArtifact =
-        Artifact.builder()
-            .type("docker/image")
-            .name("gcr.io/my-repository/my-image")
-            .reference("gcr.io/my-repository/my-image:expected-tag")
-            .build();
-    ReplaceResult replaceResult =
-        artifactReplacer.replaceAll(
-            artifactBindingStrategy, cronJob, ImmutableList.of(inputArtifact), NAMESPACE, ACCOUNT);
-
-    int expectedBoundArtifacts = 0;
-    String expectedImageAndTag = "gcr.io/my-repository/my-image:original-tag";
-    if (artifactBindingStrategy.equals(DEFAULT_DOCKER_IMAGE_BINDING)) {
-      expectedBoundArtifacts = 1;
-      expectedImageAndTag = "gcr.io/my-repository/my-image:expected-tag";
-    }
-
-    V1beta1CronJob replacedCronJob =
-        KubernetesCacheDataConverter.getResource(replaceResult.getManifest(), V1beta1CronJob.class);
-    assertThat(
-            replacedCronJob
-                .getSpec()
-                .getJobTemplate()
-                .getSpec()
-                .getTemplate()
-                .getSpec()
-                .getContainers())
-        .extracting(V1Container::getImage)
-        .containsExactly(expectedImageAndTag);
-
-    Set<Artifact> artifacts = replaceResult.getBoundArtifacts();
-    assertThat(artifacts).hasSize(expectedBoundArtifacts);
-    if (expectedBoundArtifacts > 0) {
-      assertThat(Iterables.getOnlyElement(artifacts)).isEqualTo(inputArtifact);
-    }
   }
 
   @ParameterizedTest(name = "{index} ==> with docker image artifact binding strategy = {0}")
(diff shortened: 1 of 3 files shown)


mattgogerly:
I'm confused by this - the `dockerImage()` replacer doesn't handle jobTemplates within the manifest, only `spec.template.spec['containers', 'initContainers'] - so how does using it fix the issue?

apoorvmahajan:
> I'm confused by this - the `dockerImage()` replacer doesn't handle jobTemplates within the manifest, only `spec.template.spec['containers', 'initContainers']` - so how does using it fix the issue?

Hi Matt, thanks for taking a look at this. I tried recreating the cronjob manifest i[n this test](https://github.com/spinnaker/clouddriver/pull/5876/files#diff-14d5233feb69939cacaa122fe20aca2dcb86714dd986f7451810ec13f5b30cf9R1346-R1392) which contains JobTemplates in its spec. The dockerImage() has a path of `$..spec.template.spec['containers', 'initContainers'].[?].image"` so it is able to recursively get to the image because of the `..` in the path iiuc.

dbyron-sf:
FWIW https://github.com/spinnaker/clouddriver/pull/5554 was first released in 1.28, so I'm planning to backport this to 1.28 and 1.29.

dbyron-sf (approved):

dbyron-sf (changes requested):
gotta fix a merge conflict typo.

clouddriver-kubernetes/src/test/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/ReplacerTest.java
```diff
@@ -1284,7 +1287,7 @@ void findCronJobDockerImages() {
             artifact -> {
               assertThat(artifact).isNotNull();
               assertThat(artifact.getType()).isEqualTo("docker/image");
-              assertThat(artifact.getName()).isEqualTo("gcr.io/my-repository/my-image:my-tag");
+              assertThat(artifact.getName()).isEqualTo("gcr.io/my-repository/my-image");
```
apoorvmahajan, line 1290: this is the side effect of using dockerImages in this test whose repercussions I'm not sure of.  The reference still contains the tag which seems like how it should be to me atleast.

dbyron-sf: I need to stare at it some more...like, maybe this change ends up in an initial commit with a good explanation...and then we proceed with the other changes...

dbyron-sf: Well, my digging hasn't been altogether satisfying.  The relevant difference between `dockerImage()` and `cronJobDockerImage()` (and really DOCKER_IMAGE vs. CRON_JOB_DOCKER_IMAGE) for this test is the `nameFromReference` member since path/findPath are I believe equivalent between the two.  Replacer.getArtifacts transforms the artifact name by using `nameFromReference`....only DOCKER_IMAGE specifies it to some code that removes the tag, and since CRON_JOB_DOCKER_IMAGE doesn't specify it, the name is the entire reference (which in this case includes the tag).

There's one other existing Replacer that operates on `KubernetesArtifactType.DockerImage`: POD_DOCKER_IMAGE.  It (like CRON_JOB_DOCKER_IMAGE) doesn't specify `nameFromReference`.

Before https://github.com/spinnaker/clouddriver/pull/5554, images in cron jobs were handled by DOCKER_IMAGE so I figure it's OK to go back to that behavior.  It does make me wonder whether POD_DOCKER_IMAGE is correct though, or whether we'd be better off if it has the same `nameFromReference` as DOCKER_IMAGE.

It seems the only place this ends up getting used outside of tests is via KubernetesHandler.listArtifacts, which is only called by KubernetesManifestContainerBuilder.buildManifest...and on up the chain to KubernetesJobProvider.getKubernetesJob (which doesn't seem relevant here since it's specific to jobs, not cron jobs) and ManifestController.getForAccountLocationAndName which orca calls, and the resulting artifacts make it into the pipeline execution context.....so changing this could potentially break SpEL expressions....

I suppose this is all something to ponder for the future.  This PR seems like progress.  It fixes artifact binding for cron jobs and I guess as a side effect, restores the naming of cron jobs back to what 1.27 did.

mattgogerly:
> > I'm confused by this - the `dockerImage()` replacer doesn't handle jobTemplates within the manifest, only `spec.template.spec['containers', 'initContainers']` - so how does using it fix the issue?
> 
> Hi Matt, thanks for taking a look at this. I tried recreating the cronjob manifest i[n this test](https://github.com/spinnaker/clouddriver/pull/5876/files#diff-14d5233feb69939cacaa122fe20aca2dcb86714dd986f7451810ec13f5b30cf9R1346-R1392) which contains JobTemplates in its spec. The dockerImage() has a path of `$..spec.template.spec['containers', 'initContainers'].[?].image"` so it is able to recursively get to the image because of the `..` in the path iiuc.
> 
> If you'd like, you can try [this manifest checked in this repo](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/integration/resources/manifests/cronJob.yml) at [jsonpath.com](https://jsonpath.com/):
> 
> ```
> {
>   "apiVersion": "batch/v1",
>   "kind": "CronJob",
>   "metadata": {
>     "name": "myapp"
>   },
>   "spec": {
>     "schedule": "*/1 * * * *",
>     "jobTemplate": {
>       "spec": {
>         "template": {
>           "spec": {
>             "containers": [
>               {
>                 "name": "myapp",
>                 "image": "index.docker.io/library/alpine:3.12",
>                 "imagePullPolicy": "IfNotPresent",
>                 "command": [
>                   "tail",
>                   "-f",
>                   "/dev/null"
>                 ]
>               }
>             ],
>             "restartPolicy": "Never"
>           }
>         }
>       }
>     }
>   }
> }
> ```
> 
> with the jsonPath as `$..spec.template.spec[containers,initContainers]` a.k.a similar to what the dockerImage() has and it shows the contents of the `containers` field here. I doubt we can use the filters as is since it requires some filtering programatically, but it should also prove, besides the test, that it can get to the image properly

Thanks for the explanation! Any change you could comment this so anybody who touches this in the future knows it recursively affects Jobs too?

dbyron-sf:
> > Thanks for the explanation! Any change you could comment this so anybody who touches this in the future knows it recursively affects Jobs too?
> 
> It doesn't affect K8s Jobs. That already uses the [dockerImage() replacer](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/KubernetesJobHandler.java#L46)

I think he's asking for a comment in the code next to
```
          .path("$..spec.template.spec['containers', 'initContainers'].[?].image")
```
to call out that the `$..` makes this match cron jobs, deployments, and yes I believe jobs too.  I found some info at https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-01.html#section-3.5.7.

dbyron-sf:
FWIW, I made this pipeline to exercise this functionality:
```
{
  "appConfig": {},
  "keepWaitingPipelines": false,
  "lastModifiedBy": "dbyron@salesforce.com",
  "limitConcurrent": true,
  "spelEvaluator": "v4",
  "stages": [
    {
      "account": "my-account",
      "app": "dbyrontest",
      "cloudProvider": "kubernetes",
      "location": "spinnaker",
      "manifestName": "deployment dbyron-test-deployment",
      "mode": "dynamic",
      "name": "Find Artifacts From Deployment",
      "refId": "1",
      "requisiteStageRefIds": [],
      "type": "findArtifactsFromResource"
    },
    {
      "account": "my-account",
      "app": "dbyrontest",
      "cloudProvider": "kubernetes",
      "location": "spinnaker",
      "manifestName": "cronjob dbyron-test-cronjob",
      "mode": "dynamic",
      "name": "Find Artifacts From Cronjob",
      "refId": "2",
      "requisiteStageRefIds": [],
      "type": "findArtifactsFromResource"
    },
    {
      "failOnFailedExpressions": true,
      "name": "Evaluate Variables",
      "refId": "3",
      "requisiteStageRefIds": [
        "1",
        "2"
      ],
      "type": "evaluateVariables",
      "variables": [
        {
          "key": "deploymentArtifactName",
          "value": "${#stage(\"Find Artifacts From Deployment\")[\"outputs\"][\"artifacts\"][0][\"name\"]}"
        },
        {
          "key": "cronjobArtifactName",
          "value": "${#stage(\"Find Artifacts From Cronjob\")[\"outputs\"][\"artifacts\"][0][\"name\"]}"
        }
      ]
    }
  ],
  "triggers": [],
  "updateTs": "1676490389933"
}
```
With code that has neither this PR, nor https://github.com/spinnaker/clouddriver/pull/5554 (so, the 1.27-and-before functionality), the variables were docker image names with no tags:
```
"deploymentArtifactName": "my.docker.registry/path/to/my-docker-image",
"cronjobArtifactName": "my.docker.registry/path/to/my-other-docker-image"
```

dbyron-sf:
And when I added a similar stage for working with a pod, I get:
```
"podArtifactName": "my.docker.registry/path/to/my-third-docker-image@sha256:<some_sha>"
```

dbyron-sf (approved):
Added a comment about the jsonpath descendant selector stuff [here](https://github.com/spinnaker/clouddriver/pull/5876/commits/62624e2988795a693d5abca43041850cafed6977#diff-096fe232dcf949acde36cdb2e0782ecc8245aa73555cf349f38925a071fe8d7bR168-R173).

mattgogerly (approved):

dbyron-sf:
@mergify backport release-1.28.x release-1.29.x

PR #5882, opened 2023-02-20
fix(gce): fixed the error caused by account removal

SusmithaGundu:
As mentioned in issue [https://github.com/spinnaker/spinnaker/issues/6800](url) this PR is a fix for the error which is being caused every time the /securityGroups and subnets are queried. As part of the fix, the project name is added as an attribute by GoogleSecurityGroupCachingAgent and GoogleSubnetCachingAgent. So that when the securityGroups and subnets are queried there is no need to derive the project Id from the GoogleNamedAccountCredentials because when an account is removed the accountCredentials become null and the project id cannot be found in that case and hence the exception was being thrown.


kirangodishala:
lgtm. This needs to be back ported to 1.29.x as well.

j-sandy:
@Mergifyio backport release-1.29.x

mattgogerly:
closes https://github.com/spinnaker/spinnaker/issues/6800

PR #5885, opened 2023-02-21
fix(ecr): Two credentials with same accountId confuses EcrImageProvider with different regions

wattdave:
When Autodesk set up our Spinnaker instances, we created separate Credentials for each (accountId, region) duple.

So for example, ('123456789012', 'us-west-1') is one credential, and later on, ('123456789012', 'us-east-1') is a second one.

We discovered that if we attempt to validate a container present in the account's second region (ordering determined by the getCredentials API), then it would fail to validate, because only the account id is matched by the credential search, not the regions within it.

This PR contributes a test that fails on the existing code, and a proposed change to address it.

I'm not 100% happy about including the test for no-regions-specified in the code change I'm offering, but it reduced the number of tests that needed to change; I wasn't sure if the semantics require the regions or not.

### DIFF
--- a/clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcrImageProviderSpec.groovy
+++ b/clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcrImageProviderSpec.groovy
@@ -155,7 +155,7 @@ class EcrImageProviderSpec extends Specification {
 
     Map<String, Object> region1 = Map.of(
       'name', 'eu-west-1',
-      'availabilityZones', Arrays.asList('us-west-1a', 'us-west-1b', 'us-west-1c')
+      'availabilityZones', Arrays.asList('eu-west-1a', 'eu-west-1b', 'eu-west-1c')
     )
     Map<String, Object> overrides1 = Map.of(
       'accountId', accountId,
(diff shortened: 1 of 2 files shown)


clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcrImageProviderSpec.groovy
```diff
@@ -136,6 +136,65 @@ class EcrImageProviderSpec extends Specification {
...
+      repositoryName: repoName,
+      registryId: accountId,
+      imageDigest: digest,
+      imageTags: List.of(tag),
+      imagePushedAt: creationDate
+    )
+
+    Map<String, Object> region1 = Map.of(
+      'name', 'eu-west-1',
+      'availabilityZones', Arrays.asList('us-west-1a', 'us-west-1b', 'us-west-1c')
```
wattdave, line 158:
```suggestion
      'availabilityZones', Arrays.asList('eu-west-1a', 'eu-west-1b', 'eu-west-1c')
```

clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcrImageProviderSpec.groovy
```diff
@@ -136,6 +136,65 @@ class EcrImageProviderSpec extends Specification {
...
+      'availabilityZones', Arrays.asList('eu-west-1a', 'eu-west-1b', 'eu-west-1c')
+    )
+    Map<String, Object> overrides1 = Map.of(
+      'accountId', accountId,
+      'regions', Arrays.asList(region1)
+    )
+
+    Map<String, Object> region2 = Map.of(
+      'name', region,
+      'availabilityZones', Arrays.asList('us-west-1a', 'us-west-1b', 'us-west-1c')
```
wattdave, line 167:
```suggestion
      'availabilityZones', Arrays.asList('us-east-1a', 'us-east-1b', 'us-east-1c')
```

jasonmcintosh (approved):

jasonmcintosh:
@mergify backport release-1.27.x release-1.28.x release-1.29.x

PR #5893, opened 2023-02-27
fix(helm): propagate throwable on helm artifact download error

webmakersteve:
Simply adds the cause of the `FailedDownloadException` to the exception so when logged more context is provided.

We are running into an issue personally likely due to certificate issues where our internal helm registry cannot be accessed and debugging the root cause is quite difficult.

This, I believe, would be an easy way to improve things. IOException is such a broad exception that just the simple error message doesn't really cut it. Many of the IOExceptions thrown by the base class include things like status codes and more that are all swallowed.


dbyron-sf:
Any chance for before/after screenshots of a bake manifest stage, or differences in log messages/http responses?

dbyron-sf:
And potentially an automated test that demonstrates the difference please.

webmakersteve:
I'll adjust the commit message.

As for automated tests: what do you have in mind? This is simply passing in the underlying throwable to the constructor for the custom exception.

Do you want me to ensure that it gets propagated (thus testing the java base throwable class behavior?) or are you looking to make a bulwark against regression which might remove the parameter?

@dbyron-sf

dbyron-sf:
I'm mostly concerned that some test exercises the line of code in question.  If there's something verifying that there's an exception thrown, I think it'd be straightforward to verify that the cause is non-null + the expected IOException.

webmakersteve:
@dbyron-sf is a test like that what you're looking for? Just updated the commit.

Will get screenshots when I can. Need to deploy my custom build to get it easiest since this only happens in our cluster.

dbyron-sf:
This looks great!  Thank you @webmakersteve .

webmakersteve:
@dbyron-sf can't really get a screenshot. Wanted to deploy it to our cluster but it isn't working - we're on a really old outdated version of clouddriver. 

Will need to wait until we upgrade in the next few weeks. We were able to work around the problem for the time being so we're not blocked. It just would have helped to have the error messaging.

dbyron-sf:
> @dbyron-sf can't really get a screenshot. Wanted to deploy it to our cluster but it isn't working - we're on a really old outdated version of clouddriver.
> 
> Will need to wait until we upgrade in the next few weeks. We were able to work around the problem for the time being so we're not blocked. It just would have helped to have the error messaging.

Maybe you can get a "before" screenshot now, but no "after"?

I'd say let's merge this, and when you upgrade perhaps you can come back and fill in the "after"....

mattgogerly (approved):

dbyron-sf (approved):

PR #5903, opened 2023-03-15
fix(cats-redis): release Semaphore to avoid leaking

Young-Zen:
### Preconditions to reproduce
The settings in [clouddriver.yaml](https://github.com/spinnaker/clouddriver/blob/909b6ff11dad82268a0b4401f70cabc9b4a3f3ed/clouddriver-web/config/clouddriver.yml#L11) do apply: Set AgentScheduler type to ClusteredSortAgentScheduler and redis.parallelism greater than 0.
```
redis:
  scheduler: sort
  parallelism: 10 # greater than 0
```

### Additional context
In this [Code](https://github.com/spinnaker/clouddriver/blob/909b6ff11dad82268a0b4401f70cabc9b4a3f3ed/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java#L416), the runningAgents has no release a permit to the Semaphore:
1. When worker is null or score is null;
2. When acquireAgent method throw an Exception.

### DIFF
--- a/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
+++ b/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.cats.redis.cluster;
 
 import static com.netflix.spinnaker.cats.agent.ExecutionInstrumentation.elapsedTimeMs;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.netflix.spinnaker.cats.agent.Agent;
 import com.netflix.spinnaker.cats.agent.AgentExecution;
@@ -79,8 +80,8 @@ public class ClusteredSortAgentScheduler extends CatsModuleAware
   private Optional<Semaphore> runningAgents;
 
   // This code assumes that every agent being run is in exactly either the WAITING or WORKING set.
-  private static final String WAITING_SET = "WAITZ";
-  private static final String WORKING_SET = "WORKZ";
+  @VisibleForTesting static final String WAITING_SET = "WAITZ";
+  @VisibleForTesting static final String WORKING_SET = "WORKZ";
   private static final String ADD_AGENT_SCRIPT = "addAgentScript";
   private static final String VALID_SCORE_SCRIPT = "validScoreScript";
   private static final String SWAP_SET_SCRIPT = "swapSetScript";
@@ -374,7 +375,8 @@ public class ClusteredSortAgentScheduler extends CatsModuleAware
     }
   }
 
-  private void saturatePool() {
+  @VisibleForTesting
+  void saturatePool() {
     try (Jedis jedis = jedisPool.getResource()) {
       // Occasionally repopulate the agents in case redis went down. If they already exist, this is
       // a NOOP
@@ -409,22 +411,31 @@ public class ClusteredSortAgentScheduler extends CatsModuleAware
 
       // Loop until we either run out of threads to use, or agents (which are keys) to run.
       while (!keys.isEmpty() && runningAgents.map(Semaphore::tryAcquire).orElse(true)) {
-        String agent = keys.remove(0);
+        try {
+          String agent = keys.remove(0);
 
-        AgentWorker worker = agents.get(agent);
-        ScoreTuple score;
-        if (worker != null && (score = acquireAgent(worker.agent)) != null) {
-          // This score is used to determine if the worker thread running the agent is allowed to
-          // store its results.
-          // If on release of this agent, the scores don't match, this agent was rescheduled by a
-          // separate thread.
-          worker.setScore(score.acquireScore);
-          workers.add(worker);
+          AgentWorker worker = agents.get(agent);
+          ScoreTuple score;
+          if (worker != null && (score = acquireAgent(worker.agent)) != null) {
+            // This score is used to determine if the worker thread running the agent is allowed to
+            // store its results.
+            // If on release of this agent, the scores don't match, this agent was rescheduled by a
+            // separate thread.
+            worker.setScore(score.acquireScore);
+            if (workers.add(worker)) {
+              agentWorkPool.submit(worker);
+              continue;
+            }
+          }
+          // This agent worker has not been submitted to agentWorkPool, and the acquired permit must
+          // be released to Semaphore.
+          runningAgents.ifPresent(Semaphore::release);
+        } catch (Throwable t) {
+          log.error("Failed to submit AgentWorker to agentWorkPool", t);
+          runningAgents.ifPresent(Semaphore::release);
+          // Better to ignore(not re-throw), so that the agentWorker can be submitted to the
+          // agentWorkPool as much as possible.
         }
-      }
-
-      for (AgentWorker worker : workers) {
-        agentWorkPool.submit(worker);
       }
     }
   }
--- a/cats/cats-redis/src/test/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentSchedulerTest.java
+++ b/cats/cats-redis/src/test/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentSchedulerTest.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2023 THL A29 Limited, a Tencent company.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.cats.redis.cluster;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyList;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import com.netflix.spinnaker.cats.agent.CachingAgent;
+import com.netflix.spinnaker.cats.agent.DefaultCacheResult;
+import com.netflix.spinnaker.cats.agent.ExecutionInstrumentation;
+import com.netflix.spinnaker.cats.cluster.DefaultAgentIntervalProvider;
+import com.netflix.spinnaker.cats.test.TestAgent;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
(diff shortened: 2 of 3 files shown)


cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -409,17 +409,24 @@ private void saturatePool() {
...
+            // If on release of this agent, the scores don't match, this agent was rescheduled by a
+            // separate thread.
+            worker.setScore(score.acquireScore);
+            workers.add(worker);
+          } else {
+            runningAgents.ifPresent(Semaphore::release);
+          }
+        } catch (Throwable t) {
+          log.error("Failed to add AgentWorker", t);
+          runningAgents.ifPresent(Semaphore::release);
```
dbyron-sf, line 429: Bettter to re-throw t to retain the current behavior?  I don't know enough about this code to know which way is better.

cats/cats-redis/src/test/groovy/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentSchedulerSpec.groovy
```diff
@@ -0,0 +1,81 @@
...
+import com.netflix.spinnaker.cats.agent.ExecutionInstrumentation
+import com.netflix.spinnaker.cats.cluster.DefaultAgentIntervalProvider
+import com.netflix.spinnaker.cats.cluster.NodeStatusProvider
+import com.netflix.spinnaker.cats.test.TestAgent
+import redis.clients.jedis.Jedis
+import redis.clients.jedis.JedisPool
+import spock.lang.Specification
+import spock.lang.Subject
+
+class ClusteredSortAgentSchedulerSpec extends Specification {
```
dbyron-sf, line 30: It's perhaps too late for this PR, but we'd much prefer new code -- even test code -- in any language but groovy :)  Java or kotlin please.

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
Young-Zen: Re-throw the Throwable will also cause runningAgents to not release permission to the Semaphore.
image

cats/cats-redis/src/test/groovy/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentSchedulerSpec.groovy
Young-Zen: Ok, I will try.

Young-Zen: Done.

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
dbyron-sf: I figured the additional throw would come after the release you added here, but maybe that still doesn't work.

cats/cats-redis/src/test/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentSchedulerTest.java
```diff
@@ -0,0 +1,137 @@
...
+  @Before
+  public void setUp() throws IllegalAccessException {
+    when(jedisPool.getResource()).thenReturn(jedis);
+    when(jedis.scriptLoad(anyString())).thenReturn("testScriptSha");
+    when(jedis.time()).thenReturn(List.of("1678784468", "374338"));
+    DefaultAgentIntervalProvider intervalProvider = new DefaultAgentIntervalProvider(6000000);
+    clusteredSortAgentScheduler =
+        new ClusteredSortAgentScheduler(jedisPool, () -> false, intervalProvider, parallelism);
+
+    WAITING_SET =
```
dbyron-sf, line 74: Easier to make these fields package private in ClusteredSortAgentScheduler?  Marking them as `@VisibleForTesting` makes it pretty clear why...then we don't need a copy here.

cats/cats-redis/src/test/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentSchedulerTest.java
```diff
@@ -0,0 +1,137 @@
...
+            FieldUtils.getDeclaredField(ClusteredSortAgentScheduler.class, "WAITING_SET", true)
+                .get(clusteredSortAgentScheduler);
+    WORKING_SET =
+        (String)
+            FieldUtils.getDeclaredField(ClusteredSortAgentScheduler.class, "WORKING_SET", true)
+                .get(clusteredSortAgentScheduler);
+    saturatePoolMethod =
+        MethodUtils.getMatchingMethod(ClusteredSortAgentScheduler.class, "saturatePool");
+    saturatePoolMethod.setAccessible(true);
+    runningAgents =
```
dbyron-sf, line 85: I'd say it's worth adding a getter for testing, but it's also nice to leave it hidden.

cats/cats-redis/src/test/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentSchedulerTest.java
```diff
@@ -0,0 +1,137 @@
...
+              latch.countDown();
+              return null;
+            })
+        .when(agentExecution)
+        .storeAgentResult(any(), any());
+
+    saturatePoolMethod.invoke(clusteredSortAgentScheduler);
+    latch.await(10, TimeUnit.SECONDS);
+
+    assertEquals(runningAgents.map(Semaphore::availablePermits).orElse(-1), parallelism);
```
dbyron-sf, line 135: I could use a comment here, specifically about the `orElse`.  Under what conditions does the orElse actually happen?

cats/cats-redis/src/test/groovy/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentSchedulerSpec.groovy
dbyron-sf: Thank you!

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
Young-Zen: Yes, you're right. Before this PR, it released the acquired permit to the Semaphore [here](https://github.com/spinnaker/clouddriver/blob/909b6ff11dad82268a0b4401f70cabc9b4a3f3ed/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java#L471), which will run in the worker thread managed by [agentWorkPool](https://github.com/spinnaker/clouddriver/blob/909b6ff11dad82268a0b4401f70cabc9b4a3f3ed/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java#L427). In this way, before the agentWorker is submitted to the agentWorkPool, the acquired permit will not be released due to an exception throw. Regarding the question of whether to re-throw, I think it is better to ignore, so that the agentWorker can be submitted to the agentWorkPool as much as possible.

cats/cats-redis/src/test/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentSchedulerTest.java
Young-Zen: Ok, I get it, thanks!

cats/cats-redis/src/test/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentSchedulerTest.java
Young-Zen: In this [code](https://github.com/spinnaker/clouddriver/blob/909b6ff11dad82268a0b4401f70cabc9b4a3f3ed/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java#L109), when parallelism is -1, the runningAgents is Optional.empty().

cats/cats-redis/src/test/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentSchedulerTest.java
Young-Zen: Yeah, i prefer to leave it hidden.

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
```diff
@@ -59,7 +60,7 @@
  */
 public class ClusteredSortAgentScheduler extends CatsModuleAware
     implements AgentScheduler<ClusteredSortAgentLock>, Runnable {
-  private static enum Status {
+  private enum Status {
```
dbyron-sf, line 63: Is this change necessary in this PR?

cats/cats-redis/src/test/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentSchedulerTest.java
```diff
@@ -0,0 +1,119 @@
...
+  @Test
+  public void testRunningAgentsSemaphore() {
+    when(jedis.zrangeByScore(eq(ClusteredSortAgentScheduler.WORKING_SET), anyString(), anyString()))
+        .thenReturn(new HashSet<>());
+    when(jedis.zrangeByScore(eq(ClusteredSortAgentScheduler.WAITING_SET), anyString(), anyString()))
+        .thenReturn(Set.of("testAgentType"));
+
+    clusteredSortAgentScheduler.saturatePool();
+
+    // When parallelism is -1, the runningAgents is Optional.empty().
```
dbyron-sf, line 83: But isn't parallelism initialized to 2 above?

cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentScheduler.java
Young-Zen: Unnecessary.

cats/cats-redis/src/test/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentSchedulerTest.java
Young-Zen: No need.

cats/cats-redis/src/test/java/com/netflix/spinnaker/cats/redis/cluster/ClusteredSortAgentSchedulerTest.java
Young-Zen: At first I thought that if parallelism is -1, both test cases will also pass. But think about it, these two test cases are used to test Semaphore's permit leaking, if parallelism is -1, there is no Semaphore.

dbyron-sf (approved):

PR #5904, opened 2023-03-17
chore(aws): Update AWS IAM Authenticator version

austinthao5:
Updating to v0.5.9 supports both v1alpha1 and v1beta1 apiVersions.


jasonmcintosh (approved):

jasonmcintosh:
@mergify backport release-1.28.x release-1.29.x

jasonmcintosh:
Backported this as it's really a fix to a regression in aws-iam-authenticator behavior.  In 0.5.5, a change was forced on the format of kubeconfig files to handle deprecated exec auth APIs in kubectl.  In 0.5.7 or so instead of requiring a change in kubeconfigs and breaking existing configuration, the aws-iam-authenticator would read a kubectl parameter that was passed to exec system and support both old and new exec APIs.  This lets you safely migrate off of the older kubeconfig exec format without a breaking any existing installations.

PR #5906, opened 2023-03-21
fix(core): Renamed a query parameter for template tags

kbatalin:
The parameter in the Front50 API is called `tag`, but in the Retrofit client it's called `version`. Because of this, Clouddriver ignores the tags during template lookup. 
The parameter in the API: https://github.com/spinnaker/front50/blob/a2f188767a1f9584452e6ab61c555dd1a630c468/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/V2PipelineTemplateController.java#L151

This PR fixes the issue: https://github.com/spinnaker/spinnaker/issues/6583


dbyron-sf (approved):

dbyron-sf:
@mergify backport release-1.27.x release-1.28.x release-1.29.x

PR #5940, opened 2023-04-05
bumped aws cli to 1.22 for FIPS complience

cs-simental:
Version 1.22 is the lowest version that enables FIPS endpoints for GovCloud (US) regions in SDK


mattgogerly:
This is a pretty huge version bump.. 4 minor versions and about 400 releases. Have you used this change in production?

jasonmcintosh:
AWS CLI isn't used that much.  It's PRIMARILY from what i'd seen used for external CLI auth commands.  AKA kubeconfig files that use exec auth.  SDK change would be more... interesting and I've seen THAT break things badly.  We've been using 1.21.2 of the AWS CLi for quite a while.  aws-iam-authenticator binaries are a TOUCH more annoying.

OTHER place this could impact is /home/spinnaker/.aws/config mount options - e.g. using aws auth for role assumption but that's still MOSTLY impacting SDK vs. CLI operations.   Good to callout the bump as it can impact things like that, but IN THEORY... CLI calls should have minimal impacts on most spinnaker operations.

cs-simental:
Removed the upgrade from the ubuntu Dockerfile since ubuntu:bionic doesn't have the 1.22 version yet.

jasonmcintosh (approved):

jasonmcintosh:
Close since other PRs handle this

PR #5949, opened 2023-05-05
feat(kubernetes): allow operations on all kinds

mattgogerly:
See https://github.com/spinnaker/spinnaker/issues/6833


jasonmcintosh:
Hrmm, I'm wondering about this and whether we should leave the API calls, but add a new flag to just skip validation when flag is changed.

mattgogerly:
> Hrmm, I'm wondering about this and whether we should leave the API calls, but add a new flag to just skip validation when flag is changed.

Discussed it a bit on Slack. It's a pain to get feature flags down to `KubernetesValidationUtil` because it's created for every op. Would have to pass it down the chain a fair way.

dbyron-sf (approved):
This needs a release note, but LGTM.

PR #5954, opened 2023-05-27
fix(aws): ECS Service Tagging broken

HaroonSaid:
The `isTaggingEnabled` method doesn't check `listAccountSettings` is returning a token for a paged response. 
This caused the deployment to fail even if the AWS Account has been enabled for the tagging feature.


scadu:
Related discussion on [Spinnaker's Slack](https://spinnakerteam.slack.com/archives/C091CCWRJ/p1685060444050859).
Without the fix introduced in this PR, many folks cannot deploy to ECS.

goatherder:
I have tested this patch and can confirm it resolves issues we were seeing.  ie:
With the patch - we now see 2 calls to `ListAccountSettingsRequest`.  `isServiceLongArnFormatEnabled` was on the first page in the response, `isTaskLongArnFormatEnabled` on the second.

link108:
@mergifyio backport release-1.27.x release-1.28.x release-1.29.x release-1.30.x

link108:
@mergifyio backport release-1.30.x

PR #5982, opened 2023-07-17
perf(cache): Optimise heap usage in SqlCache

kbatalin:
This pull request addresses a performance issue discovered in the Clouddriver, specifically within the `SqlCache.getHash method`.

I used async-profiler to build an Allocation profile and a CPU profile. From the Allocation profile, it was found that the `SqlCache.getHash` method was responsible for 80% of the memory allocation, mostly due to the heavy use of `String.format` within this method. Furthermore, the CPU profile indicated that 34% of CPU usage was also attributed to this method.

So I'm making a conclusion that `SqlCache.getHash` is generating a significant amount of garbage, contributing to high heap and CPU usage. This could potentially be a root cause for Out Of Memory errors if the Garbage Collector is unable to clean up the generated garbage efficiently. It also may add to high CPU usage. Perhaps, relevant issues: [#6835](https://github.com/spinnaker/spinnaker/issues/6835), [#6797](https://github.com/spinnaker/spinnaker/issues/6797), and [#6810](https://github.com/spinnaker/spinnaker/issues/6810).

The proposed optimisation in this pull request aims to reduce the amount of garbage generated by this method by pre-generating string representations for all possible byte values and using `StringBuilder` with a predefined size.

This optimisation has been deployed in our production environment almost for a month and has shown promising results, reducing resource consumption by 2-5 times. I attached the screenshots of the graphs to this pull request for illustration.

image
image
image
image

### DIFF
--- a/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
+++ b/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
@@ -79,8 +79,16 @@ class SqlCache(
 
   private var createdTables = ConcurrentSkipListSet<String>()
 
+  private val hexStrings: List<String>
+
   init {
     log.info("Configured for $name")
+
+    this.hexStrings = arrayListOf()
+    for (byte in Byte.MIN_VALUE..Byte.MAX_VALUE) {
+      val str = "%02x".format(byte.toByte())
+      this.hexStrings.add(str)
+    }
   }
 
   /**
@@ -927,9 +935,11 @@ class SqlCache(
     return try {
       val digest = MessageDigest.getInstance("SHA-256")
         .digest(body.toByteArray())
-      digest.fold("") { str, it ->
-        str + "%02x".format(it)
+      val builder = StringBuilder(64)
+      for (byte in digest) {
+        builder.append(this.hexStrings[byte - Byte.MIN_VALUE])
       }
+      builder.toString()
     } catch (e: Exception) {
       log.error("error calculating hash for body: $body", e)
       null
--- a/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
+++ b/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
@@ -84,6 +84,9 @@ class SqlCache(
   init {
     log.info("Configured for $name")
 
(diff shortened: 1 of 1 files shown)


cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -79,8 +79,16 @@ class SqlCache(
...
+  private val hexStrings: List<String>
+
   init {
     log.info("Configured for $name")
+
+    this.hexStrings = arrayListOf()
+    for (byte in Byte.MIN_VALUE..Byte.MAX_VALUE) {
+      val str = "%02x".format(byte.toByte())
+      this.hexStrings.add(str)
+    }
```
mattgogerly, line 91: Would you mind adding a comment here explaining what this is doing? It will look like voodo for anyone new to the codebase 😄

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
```diff
@@ -927,9 +935,11 @@ class SqlCache(
     return try {
       val digest = MessageDigest.getInstance("SHA-256")
         .digest(body.toByteArray())
-      digest.fold("") { str, it ->
-        str + "%02x".format(it)
+      val builder = StringBuilder(64)
+      for (byte in digest) {
+        builder.append(this.hexStrings[byte - Byte.MIN_VALUE])
       }
+      builder.toString()
```
mattgogerly, line 942: Same as above - mind commenting this?

mattgogerly (commented):
This looks incredibly promising - good stuff!

cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
kbatalin: 👍 Fixed. Added a comment

mattgogerly (approved):

jvz (commented):
https://www.baeldung.com/kotlin/byte-arrays-to-hex-strings has some suggestions on efficient implementations for encoding to hex. Once we're on Java 17, we can use the built-in `HexFormat` class. Otherwise, you can use [`Hex`](https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Hex.html) from `commons-codec` to avoid reimplementing anything.

PR #5989, opened 2023-08-04
feat(provider/kubernetes): support for kubectl server-side-apply strategy

a7i:
kubernetes server-side apply (SSA) was released back in 1.14 and became GA In 1.22. This new strategy will use the new merging algorithm, as well as tracking field ownership at the kubernetes api-server

fixes: https://github.com/spinnaker/spinnaker/issues/6865

### DIFF
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/manifest/KubernetesManifestStrategy.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/manifest/KubernetesManifestStrategy.java
@@ -107,7 +107,8 @@ public final class KubernetesManifestStrategy {
   public enum DeployStrategy {
     APPLY(null),
     RECREATE(STRATEGY_ANNOTATION_PREFIX + "/recreate"),
-    REPLACE(STRATEGY_ANNOTATION_PREFIX + "/replace");
+    REPLACE(STRATEGY_ANNOTATION_PREFIX + "/replace"),
+    SERVER_SIDE_APPLY(STRATEGY_ANNOTATION_PREFIX + "/server-side-apply");
 
     @Nullable private final String annotation;
 
@@ -121,6 +122,9 @@ public final class KubernetesManifestStrategy {
       }
       if (Boolean.parseBoolean(annotations.get(REPLACE.annotation))) {
         return REPLACE;
+      }
+      if (Boolean.parseBoolean(annotations.get(SERVER_SIDE_APPLY.annotation))) {
+        return SERVER_SIDE_APPLY;
       }
       return APPLY;
     }
--- a/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/manifest/KubernetesManifestStrategy.java
+++ b/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/manifest/KubernetesManifestStrategy.java
@@ -40,10 +40,15 @@ public final class KubernetesManifestStrategy {
   private static final String USE_SOURCE_CAPACITY =
       STRATEGY_ANNOTATION_PREFIX + "/use-source-capacity";
 
+  private static final String SERVER_SIDE_APPLY_STRATEGY =
+      STRATEGY_ANNOTATION_PREFIX + "/server-side-apply";
+  private static final String SERVER_SIDE_APPLY_FORCE_CONFLICTS = "force-conflicts";
+
   private final DeployStrategy deployStrategy;
   private final Versioned versioned;
   private final OptionalInt maxVersionHistory;
   private final boolean useSourceCapacity;
+  private final ServerSideApplyStrategy serverSideApplyStrategy;
 
   @Builder
   @ParametersAreNullableByDefault
@@ -51,18 +56,22 @@ public final class KubernetesManifestStrategy {
       DeployStrategy deployStrategy,
       Versioned versioned,
       Integer maxVersionHistory,
-      boolean useSourceCapacity) {
+      boolean useSourceCapacity,
+      ServerSideApplyStrategy serverSideApplyStrategy) {
     this.deployStrategy = Optional.ofNullable(deployStrategy).orElse(DeployStrategy.APPLY);
     this.versioned = Optional.ofNullable(versioned).orElse(Versioned.DEFAULT);
     this.maxVersionHistory =
         maxVersionHistory == null ? OptionalInt.empty() : OptionalInt.of(maxVersionHistory);
     this.useSourceCapacity = useSourceCapacity;
+    this.serverSideApplyStrategy =
+        Optional.ofNullable(serverSideApplyStrategy).orElse(ServerSideApplyStrategy.DEFAULT);
   }
 
   static KubernetesManifestStrategy fromAnnotations(Map<String, String> annotations) {
     return KubernetesManifestStrategy.builder()
         .versioned(Versioned.fromAnnotations(annotations))
         .deployStrategy(DeployStrategy.fromAnnotations(annotations))
+        .serverSideApplyStrategy(ServerSideApplyStrategy.fromAnnotations(annotations))
         .useSourceCapacity(Boolean.parseBoolean(annotations.get(USE_SOURCE_CAPACITY)))
         .maxVersionHistory(Ints.tryParse(annotations.getOrDefault(MAX_VERSION_HISTORY, "")))
         .build();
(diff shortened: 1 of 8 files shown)


clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/CanDeploy.java
```diff
@@ -59,6 +59,9 @@ default OperationResult deploy(
       case REPLACE:
         deployedManifest = credentials.createOrReplace(manifest, task, opName);
         break;
+      case SERVER_SIDE_APPLY:
+        deployedManifest = credentials.deploy(manifest, task, opName, "--server-side");
```
dbyron-sf, line 63: Doesn't this need to be `--server-side=true` ?

dbyron-sf:
Still wrapping my head around the approach, but it very well could be OK.  Would you please add some [integration tests](https://github.com/spinnaker/clouddriver/tree/3cbd4788c564dfdb4c8ab252a3d5f5a84d673305/clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it) to get a bit more verification that it works properly?

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/manifest/KubernetesManifestStrategy.java
```diff
@@ -107,7 +107,8 @@ ImmutableMap<String, String> toAnnotations() {
   public enum DeployStrategy {
     APPLY(null),
     RECREATE(STRATEGY_ANNOTATION_PREFIX + "/recreate"),
-    REPLACE(STRATEGY_ANNOTATION_PREFIX + "/replace");
+    REPLACE(STRATEGY_ANNOTATION_PREFIX + "/replace"),
+    SERVER_SIDE_APPLY(STRATEGY_ANNOTATION_PREFIX + "/server-side-apply");
```
dbyron-sf, line 111: From https://kubernetes.io/docs/reference/using-api/server-side-apply/#conflicts, it seems worth trying to support `--force-conflicts`.

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/handler/CanDeploy.java
a7i: Specifying the flag will default it to true

a7i:
> Still wrapping my head around the approach, but it very well could be OK. Would you please add some [integration tests](https://github.com/spinnaker/clouddriver/tree/3cbd4788c564dfdb4c8ab252a3d5f5a84d673305/clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it) to get a bit more verification that it works properly?

Thanks for the suggestion @dbyron-sf I have implemented two integration tests per your suggestion (client-side vs server-side apply)

clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/manifest/KubernetesManifestStrategy.java
a7i: what is your recommendation? as another strategy annotation?

this can get tricky with fields such as `replicas` where HPA can become the manager, but force could take over.

dbyron-sf: I was thinking of using a value for this annotation, something like `strategy.spinnaker.io/server-side-apply: force` to indicate `--force-conflicts`.

a7i: took a quick look and it may require some refactoring since some code paths assume the value to be a boolean.

would `strategy.spinnaker.io/server-side-apply-force-conflicts` be a good option here?

a7i: @dbyron-sf anything I can do to get this PR moving?

Looking at Armory scale agent, it seems to only support `serverside-apply` flag but it seems to add force-conflicts by default. Thoughts?

dbyron-sf: Sorry, I've been swamped lately.  I'm pretty sure
```
strategy.spinnaker.io/server-side-apply: force
```
for force, and...I dunno, for the no-force case, maybe
```
strategy.spinnaker.io/server-side-apply: no-force
```
I believe adding something like this (untested):
```
  public enum ServerSideApplyStrategy {
    FORCE(ImmutableMap.of(SERVER_SIDE_APPLY_STRATEGY, "force")),
    NO_FORCE(ImmutableMap.of(SERVER_SIDE_APPLY_STRATEGY, "no-force")),
    DEFAULT(ImmutableMap.of());
    private final ImmutableMap<String, String> annotations;

    ServerSideApplyStrategy(ImmutableMap<String, String> annotations) {
      this.annotations = annotations;
    }

    static ServerSideApplyStrategy fromAnnotations(Map<String, String> annotations) {
      if (annotations.containsKey(SERVER_SIDE_APPLY_STRATEGY)) {
        switch (annotations.get(SERVER_SIDE_APPLY_STRATEGY)) {
          case "force":
            return FORCE;
          case "no-force":
            return NO_FORCE;
          default:
            return DEFAULT;
        }
      }
      return DEFAULT;
    }

    ImmutableMap<String, String> toAnnotations() {
      return annotations;
    }
  }
```
and a
```
private final ServerSideApplyStrategy serverSideApplyStrategy;
```
member to KubernetesManifestStrategy, initialized with:
```
    this.serverSideApplyStrategy =
        Optional.ofNullable(serverSideApplyStrategy).orElse(ServerSideApplyStrategy.DEFAULT);
```
 with a call to
```
    builder.putAll(serverSideApplyStrategy.toAnnotations());
```
in toAnnotations and
```
        .serverSideApplyStrategy(ServerSideApplyStrategy.fromAnnotations(annotations))
```
works.

mattgogerly (approved):

a7i:
@jasonmcintosh thanks for following up. I took a stab at it, and it just seemed a bit messy to have yet another annotation. Also, one more arg that I had to pass around for deployment, etc. I also really dislike the idea of annotations as an API but it's the existing pattern so not much I can do about that.
I'll take another look.

jasonmcintosh:
Yeah the goal is ONE annotation just with values for it.  This would match how ALB ingress controllers work.  It's not... "ideal" but not sure there's a great answer without that to pass this kinda thing around without changing it on deploy stages.

a7i:
> Yeah the goal is ONE annotation just with values for it. This would match how ALB ingress controllers work. It's not... "ideal" but not sure there's a great answer without that to pass this kinda thing around without changing it on deploy stages.
Let me know how you feel about the new commit I pushed

| Annotation | Behavior|
| ----------- | -------------- |
| `strategy.spinnaker.io/server-side-apply: 'false'` | Defaults to client-side Apply |
| `strategy.spinnaker.io/server-side-apply: 'true'` | Defaults to Server-Side Apply (i.e. `kubectl apply --server-side=true` |
| `strategy.spinnaker.io/server-side-apply: 'force-conflicts'` | Defaults to Server-Side Apply (i.e. `kubectl apply --server-side=true --force-conflicts=true` |

jasonmcintosh:
Think this should work well, allowing all the options.  LONG TERM, as said, would love it if "Server side apply w/force" was just the default, another release or two for that default ;)

jasonmcintosh (approved):

a7i:
> Think this should work well, allowing all the options. LONG TERM, as said, would love it if "Server side apply w/force" was just the default, another release or two for that default ;)

How do you feel about field-managers? should that be another annotation?

for example, apply a Deployment but let HPA handle the replicas.
```
kubectl apply -f nginx-deployment-replicas-only.yaml \
  --server-side --field-manager=handover-to-hpa
```

jasonmcintosh:
For. the MOMENT probably an annotation.  There's two sorta strategies that come into play
* Do it as a pipeline argument, aka "overrideNamespace"
* Annotation driven - for this particular resource, do X (eg disabling versioning)
ARGUABLY (to my mind)... some of this would be better off NOT as an annotation but as an option on the stage itself similar to overrideNamespace.  It's just... more work/trickier/etc. on the stage AND you lose a lot of granularity of control with the advantage that you can apply these without changing manifests (though I'd argue defining it in the manifest is a bit more DRY). As much as I like stage overrides, annotations offer a LOT of capability as well that can be done PER resource.  E.g. "For a configmap do a server side apply but for a deployment don't".  That CAN be critical as we've seen VERY different behavior on server side applies vs. kubectl applies.

It'd be interesting to have more conversations with the Cloud SIG on direction on this long term and impacts.

jasonmcintosh:
Merging this - but would love a docs update on the force changes ;) Thanks!!

a7i:
@jasonmcintosh https://github.com/spinnaker/spinnaker.io/pull/363

PR #5995, opened 2023-08-16
chore(awscli): Bump AWS CLI and fix install of AWS CLI

jasonmcintosh:
See: https://github.com/aws/aws-cli/issues/8036#issuecomment-1638544754
This also bumps to a later release of the CLI (used MOSTLY for kubeconfig exec auth & ECR token fetch, so impact should be minimal).  Note armory has been using 1.22 of the CLI for quite a while (since 1.28).  This PR bumps BOTH ubuntu & slim versions and fixes the ubuntu version to match.

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -3,7 +3,7 @@ LABEL maintainer="sig-platform@spinnaker.io"
 ARG TARGETARCH
 
 # KUBECTL_RELEASE kept one minor version behind latest to maximise compatibility overlap
-ENV KUBECTL_RELEASE=1.20.6
+ENV KUBECTL_RELEASE=1.22.17
 ENV AWS_CLI_VERSION=1.22
 ENV AWS_CLI_S3_CMD=2.0.2
 ENV AWS_AIM_AUTHENTICATOR_VERSION=0.5.9
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -4,7 +4,7 @@ ARG TARGETARCH
 
 # KUBECTL_RELEASE kept one minor version behind latest to maximise compatibility overlap
 ENV KUBECTL_RELEASE=1.20.6
-ENV AWS_CLI_VERSION=1.18.152
+ENV AWS_CLI_VERSION=1.22
 ENV AWS_CLI_S3_CMD=2.0.2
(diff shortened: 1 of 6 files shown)


Dockerfile.slim
```diff
@@ -23,7 +23,7 @@ RUN apk update \
     openssh-client
 
 # AWS CLI
-RUN pip install --upgrade awscli==${AWS_CLI_VERSION} s3cmd==${AWS_CLI_S3_CMD} python-magic \
+RUN pip install --upgrade  --no-build-isolation awscli==${AWS_CLI_VERSION} s3cmd==${AWS_CLI_S3_CMD} python-magic \
```
dbyron-sf, line 26: What does --no-build-isolation do?

jasonmcintosh: From what I understand, it prevents some automatic install of some dependencies:  
https://stackoverflow.com/questions/62889093/what-does-no-build-isolation-do/62889267#62889267

This was the recommended flag from AWS to get the CLI install working.  Whether it's "right" is a different question. 
 I'll bet testing more shortly BUT getting this out here now while I do that testing.

jasonmcintosh: Additional notes:  this seems to be dependent upon which version of python & libs.  AKA builds failed on ubuntu.  Ubuntu  bionic is out of standard support so this MAY not be required UNTIl we start other upgrades.

jasonmcintosh: What breaks is the pip install.  Will NOT impact any EXISTING built images and It's PRIMARILY impacting Alpine builds - started in July? due to a dependency of the aws cli pip install.   Bionic may not have the issue but bionic is WAY out of date, testing that as well in this.  Will LIKELY revert the bionic and do a separate PR with JUST the ubuntu changes.

Dockerfile.ubuntu
```diff
@@ -1,8 +1,10 @@
-FROM ubuntu:bionic
+FROM ubuntu:jammy
```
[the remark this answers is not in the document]
jasonmcintosh: Yes :) Planning on it - using this to test this the CI workflow but will break it into a separate PR :)

mattgogerly (approved):

PR #6004, opened 2023-08-24
chore(dependencies): bump kork to 7.187.0

dbyron-sf:
The last autobump to 7.186.0 (https://github.com/spinnaker/clouddriver/pull/5994) hadn't gone in before 7.187.0 arrived, so we're manually bumping.


dbyron-sf:
7.187.0 fails here.  https://github.com/spinnaker/kork/pull/1094 has the fix, so closing and letting the autobump PR to 7.188.0 move us forward.

PR #6048, opened 2023-09-20
fix(lambda): Lambda is leaking threads on agent refreshes.  

jasonmcintosh:
The custom executor service thing which seems like a perf improvement is causing thread leak and eventual failures.  Remove it entirely.


clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/asg/AsgConfigHelperSpec.groovy
```diff
@@ -53,8 +53,9 @@ class AsgConfigHelperSpec extends Specification {
     when:
     def actualName = AsgConfigHelper.createName(baseName, suffix)
 
+    //ignore the end precision for tests.
     then:
-    actualName.contains(expectedName)
+    actualName.contains(expectedName.substring(0, expectedName.length() - 3))
```
jasonmcintosh, line 58: removed a bit of precision here.  This was causing regular test failures.

PR #6066, opened 2023-10-02
feat: Replace the default properties handler setter, which is currently employed for plugins purposes, with the CRD properties resolver.

armory-abedonik:


dbyron-sf:
> Note: Create backports for 1.28.x, 1.29.x, 1.30.x, 1.31.x, 1.32.x

We don't usually backport features...

dbyron-sf:
This could use tests, and some more background on what problem this solves, and why this is a good solution.

armory-abedonik:
> This could use tests, and some more background on what problem this solves, and why this is a good solution.

I will add unit tests to encompass the changes in question.

Regarding the purpose of these changes, our objective is to create a plugin that will assign a specific Kubernetes handler for Custom Resource Definition (CRD) resources based on the kind. Unfortunately, PF4J does not provide the capability to override an entire class. Furthermore, we are constrained from introducing another bean that either implements ResourcePropertyRegistry or extends GlobalResourcePropertyRegistry, primarily due to the presence of the @Component annotation. It's worth noting that employing the @Primary annotation is not a viable solution, as PF4J imposes limitations. Specifically, we encounter issues with the initialization process of both beans within the Spring context, rendering them nonfunctional.

armory-abedonik:
This pull request is no longer relevant. Closed.

PR #6077, opened 2023-10-06
feat(lambda): Allow timeouts to be set for operations

jasonmcintosh:
This includes commenting out some of the dead paths.  I've ONLY commented them out vs. removing them entirely at this point.  Some key things
* This enables OTHER services to potentially increase their SDK timeout/configs.  It also caches those timeout configs since that's PER "Client" operation for ALL such client operations/account/region configs
* I've completely removed the custom retry in favor of the SDK retry which supports things LIKE exponential backoffs and similar strategies.  
* I'm also trying to collapse/cleanup some of the lambda configuration options as well.


clouddriver-lambda/src/main/java/com/netflix/spinnaker/config/LambdaServiceConfig.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright 2021 Armory, Inc.
```
ovidiupopa07, line 2: 2023?

clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/deploy/ops/AbstractLambdaAtomicOperation.java
```diff
@@ -16,24 +16,15 @@
...
 import com.netflix.spinnaker.clouddriver.orchestration.AtomicOperation;
-import com.netflix.spinnaker.clouddriver.security.AccountCredentialsProvider;
 import org.springframework.beans.factory.annotation.Autowired;
 
 public abstract class AbstractLambdaAtomicOperation<T extends AbstractLambdaFunctionDescription, K>
-    implements AtomicOperation<K> {
-  @Autowired AmazonClientProvider amazonClientProvider;
-
-  @Autowired AccountCredentialsProvider accountCredentialsProvider;
+    extends AbstractLambdaProvider implements AtomicOperation<K> {
```
ovidiupopa07, line 27: Would it make sense to do composition instead of inheritance here?

clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/deploy/ops/InvokeLambdaAtomicOperation.java
```diff
@@ -47,7 +48,7 @@ public class InvokeLambdaAtomicOperation
 
   @Autowired private ArtifactDownloader artifactDownloader;
 
-  @Autowired LambdaOperationsConfig operationsConfig;
+  @Autowired LambdaServiceConfig operationsConfig;
```
ovidiupopa07, line 51: this looks like it is unused. is it on purpose?

clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/service/LambdaService.java
```diff
@@ -273,33 +239,26 @@ private Void addTargetGroupAttributes(
...
 
+  private static final Predicate<Statement> isLambdaInvokeAction =
+      statement ->
+          statement.getActions().stream()
+              .anyMatch(action -> action.getActionName().equals("lambda:InvokeFunction"));
+  private static final Predicate<Statement> isElbPrincipal =
+      statement ->
+          statement.getPrincipals().stream()
+              .anyMatch(
+                  principal -> principal.getId().equals("elasticloadbalancing.amazonaws.com"));
```
ovidiupopa07, line 250: does it make sense to inverse the equals to avoid a possible NPE `"elasticloadbalancing.amazonaws.com".equals(principal.getId())` ?

clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/service/LambdaService.java
```diff
@@ -315,35 +274,10 @@ private List<String> getTargetGroupNames(String functionName) {
 
     } catch (ResourceNotFoundException e) {
       // ignore the exception.
+    } catch (NullPointerException e) {
```
ovidiupopa07, line 277: does it make sense to log anything here?

clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/deploy/ops/InvokeLambdaAtomicOperation.java
```diff
@@ -34,17 +34,21 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
+import lombok.extern.log4j.Log4j2;
 import org.apache.commons.compress.utils.IOUtils;
 import org.apache.commons.io.FileUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 
+@Log4j2
```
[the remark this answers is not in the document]
jasonmcintosh: This matches a general move to the Log4j2 vs. slf4j for APi for logs.

clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/deploy/ops/InvokeLambdaAtomicOperation.java
```diff
@@ -71,12 +75,14 @@ private InvokeLambdaFunctionOutputDescription invokeFunction(
...
             .withLogType(LogType.Tail)
-            .withPayload(payload);
+            .withPayload(payload)
+            .withSdkRequestTimeout(operationsConfig.getInvokeTimeoutMs());
 
     String qualifierRegex = "|[a-zA-Z0-9$_-]+";
     if (description.getQualifier().matches(qualifierRegex)) {
       req.setQualifier(description.getQualifier());
     }
+    log.info("Invoking Lmabda function " + functionName + " and waiting for it to complete");
```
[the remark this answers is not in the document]
jasonmcintosh: Yeah arguable should be a debug but ;) Good change

clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/service/LambdaService.java
jasonmcintosh: It's not been logged in the past... but I"ll add an info level log.  More logs are useful :)

clouddriver-lambda/src/main/java/com/netflix/spinnaker/config/LambdaServiceConfig.java
jasonmcintosh: Since this was a move... kept the original date.

clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/service/LambdaService.java
jasonmcintosh: Probably wasn't super critical - I think ID was always there, but done ;)

clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/deploy/ops/AbstractLambdaAtomicOperation.java
jasonmcintosh: I'd LEAN towards... at this point.. leaving it where it is.  Renamed the class, but all of this could use a more significant refactor in the future.

ovidiupopa07 (approved):

jasonmcintosh:
To note:
* Since "retries" are actually an SDK configuration option at lambda client aquisition point, setting retries is no longer viable.  Instead the SDK decides when to retry an operation.  Spinnaker doesn't expose retry support in a currently viable way that's unique to each operation.
* Prior to this PR, there were conflicting retries on BOTH The SDK & custom retry logic, leading to duplicate retry efforts
* Long term, can see if we can better expose raw client configuration options to things like the lambda stages, so they can override PER operation things like retry behavior.  This would take more complicated changes to the way the current AWS Client objects are exposed to downstream operations (aka they're currently shared/cached by region & account vs. created at execution time).  An exposure which would allow direct, non-cached client operations could be useful for "ORCHESTRATION" type operations that are more.. real time queries.

ovidiupopa07 (approved):

link108 (approved):

PR #6089, opened 2023-11-10
fix(aws/test): Fix requests for time in "where" not fixed

E-Paine:
Previously, calls to `AsgConfigHelper.createDefaultSuffix()` in the `AsgConfigHelperSpec` test ["should return name correctly"](https://github.com/spinnaker/clouddriver/blob/8825c133897ae962c9de674ef5cf1b6f9fe8a3c6/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/asg/AsgConfigHelperSpec.groovy#L63) used the actual local time, since the `setup` function runs after a `where` block. This would very occasionally result in a test failure.

This PR resolves the issue by using `setupSpec` instead (correspondingly changing `cleanup` to `cleanupSpec`).


dbyron-sf:
Thanks for this change @E-Paine !  The `test` prefix is for commits that add test coverage.  In this case, I think `fix(aws/test)` is more correct.  Would you mind adjusting the commit message and PR description?

dbyron-sf:
I took care of it.  Thanks again @E-Paine !

dbyron-sf (approved):

PR #5976, opened 2023-06-29
feat(artifacts): Adds ArtifactStore logic to clouddriver

xibz:
This commit adds artifact storage to clouddriver. The idea in this
commit is artifact will automatically be expanded when the server
receives a request since clouddriver is more than likely interested in
the artifact.

PR #6114, opened 2023-11-25
chore(artifacts): remove reference to artifact-store.enabled

dbyron-sf:
and always construct artifact store-related beans

since https://github.com/spinnaker/kork/pull/1120 removes the artifact-store.enabled property such that ArtifactStore beans are always present.


clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy
```diff
@@ -88,20 +86,12 @@ class Main extends SpringBootServletInitializer {
 
   @Bean
   @Primary
```
dbyron-sf, line 88: Not a fan of `@Primary`, but it was here before https://github.com/spinnaker/clouddriver/pull/5976/files#diff-696765285cfe1add5c28a1a59f9d7ad2249e9bc9e9383104767c076af686bc64R90 added the artifact store stuff.

xibz: Agreed! I think we can figure out a better pattern at some point, but clouddriver relies on it too heavily at the moment. So we can take some time to refactor this at some point

PR #6134, opened 2024-01-08
feat(kubernetes): Support multiple kubectl versions

jervi:
Currently it is hard to support clusters that runs different Kubernetes versions. Because kubectl executables can be defined per cluster, we can overcome this limitation by installing multiple kubectl versions, and then define which binary to use in the cluster config. The kubectl versions are stored without patch version, so that we can upgrade to newer versions without having to change the configuration. We also symlink `kubectl` to the current default version to not break existing configuration.

To opt in to a newer version, define it like this in the cluster config: `kubectlExecutable: /usr/local/bin/kubectl-1.28` E.g.:
```
kubernetes:
  enabled: true
  primaryAccount: my-cluster
  accounts:
    - name: my-cluster
      kubeconfigFile: configserver:k8s/my-cluster.yml
      cacheThreads: 1
      cachingPolicies: []
      checkPermissionsOnStartup: true
      customResources: []
      dockerRegistries: []
      kinds: []
      liveManifestCalls: false
      namespaces: []
      oAuthScopes: []
      omitKinds: []
      omitNamespaces: []
      onlySpinnakerManaged: false
      providerVersion: V2
      kubectlExecutable: /usr/local/bin/kubectl-1.28
```


dbyron-sf:
This looks great, thank you!  Can I convince you to take a look at adding some mechanism for testing these additional kubectl versions, perhaps by teaching [the integrations tests](https://github.com/spinnaker/clouddriver/blob/f335fe6420bb89b3a199a788ac3eb4c2d81ba175/.github/workflows/integration-tests-kubernetes.yml) to have yet another dimension for kubectl version. [KubernetesCluster](https://github.com/spinnaker/clouddriver/blob/f335fe6420bb89b3a199a788ac3eb4c2d81ba175/clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/containers/KubernetesCluster.java) is also likely relevant.

jervi:
Seems like the branch protections needs to be updated as well, once this is ready to be merged.

jervi:
@dbyron-sf Any chance you can have another look at this? 😄

dbyron-sf:
Sorry, I've been slammed with other things.  Definitely have this on my radar though.

.github/workflows/integration-tests-kubernetes.yml
```diff
@@ -10,16 +10,15 @@ jobs:
...
-          - "kindest/node:v1.27.3@sha256:9dd3392d79af1b084671b05bcf65b21de476256ad1dcc853d9f3b10b4ac52dde"
-          - "kindest/node:v1.26.6@sha256:5e5d789e90c1512c8c480844e0985bc3b4da4ba66179cc5b540fe5b785ca97b5"
-          - "kindest/node:v1.25.11@sha256:ca7371d5d1f680778e40e9bcbbea06c3c61ac6f0e1a29cba5c58ff2dae1ba7f5"
-          - "kindest/node:v1.24.15@sha256:24473777a1eef985dc405c23ab9f4daddb1352ca23db60b75de9e7c408096491"
-          - "kindest/node:v1.23.0@sha256:49824ab1727c04e56a21a5d8372a402fcd32ea51ac96a2706a12af38934f81ac"
-          - "kindest/node:v1.22.0@sha256:b8bda84bb3a190e6e028b1760d277454a72267a5454b57db34437c34a588d047"
-          - "kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6"
-          - "kindest/node:v1.20.7@sha256:cbeaf907fc78ac97ce7b625e4bf0de16e3ea725daf6b04f930bd14c67c671ff9"
-          - "kindest/node:v1.19.11@sha256:07db187ae84b4b7de440a73886f008cf903fcf5764ba8106a9fd5243d6f32729"
+          - "kindest/node:v1.29.0@sha256:eaa1450915475849a73a9227b8f201df25e55e268e5d619312131292e324d570"
```
dbyron-sf, line 14: Possible to add a link to the source of this info?  Plus also I see at [the current HEAD of master](https://github.com/spinnaker/clouddriver/blob/069140bae4cc171f9d0e85b7ec40141affd4b5b9/clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/containers/KubernetesCluster.java#L39) we're already specifying kind 0.20.0, so maybe the commit message could use a tweak?

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeployManifestIT.java
```diff
@@ -16,27 +16,28 @@
...
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
-import org.apache.logging.log4j.util.Strings;
-import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.DisplayName;
-import org.junit.jupiter.api.Test;
+
+import static io.restassured.RestAssured.given;
```
dbyron-sf: Can we keep the static imports at the top please?

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/DeployManifestIT.java
```diff
@@ -16,27 +16,28 @@
...
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.netflix.spinnaker.clouddriver.kubernetes.it.utils.KubeTestUtils;
 import io.restassured.response.Response;
+import org.apache.logging.log4j.util.Strings;
```
dbyron-sf: I think the convention is straight alphabetical order as opposed to putting the java imports at the end.  Did spotless do this?

.github/workflows/integration-tests-kubernetes.yml
```diff
@@ -10,16 +10,16 @@ jobs:
...
-          - "kindest/node:v1.27.3@sha256:9dd3392d79af1b084671b05bcf65b21de476256ad1dcc853d9f3b10b4ac52dde"
-          - "kindest/node:v1.26.6@sha256:5e5d789e90c1512c8c480844e0985bc3b4da4ba66179cc5b540fe5b785ca97b5"
-          - "kindest/node:v1.25.11@sha256:ca7371d5d1f680778e40e9bcbbea06c3c61ac6f0e1a29cba5c58ff2dae1ba7f5"
-          - "kindest/node:v1.24.15@sha256:24473777a1eef985dc405c23ab9f4daddb1352ca23db60b75de9e7c408096491"
-          - "kindest/node:v1.23.0@sha256:49824ab1727c04e56a21a5d8372a402fcd32ea51ac96a2706a12af38934f81ac"
-          - "kindest/node:v1.22.0@sha256:b8bda84bb3a190e6e028b1760d277454a72267a5454b57db34437c34a588d047"
-          - "kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6"
-          - "kindest/node:v1.20.7@sha256:cbeaf907fc78ac97ce7b625e4bf0de16e3ea725daf6b04f930bd14c67c671ff9"
-          - "kindest/node:v1.19.11@sha256:07db187ae84b4b7de440a73886f008cf903fcf5764ba8106a9fd5243d6f32729"
+          # Consider updating the kubectl version mapping in BaseTest.java when adding new kubernetes versions
```
dbyron-sf, line 13: The mapping is in KubernetesCluster.java, right?  And doesn't this comment more belong in the Dockerfiles?

.github/workflows/integration-tests-kubernetes.yml
jervi: You are correct that the Kind version being installed is already 0.20.0. However, if you [compare the hashes here](https://github.com/kubernetes-sigs/kind/releases/tag/v0.11.1), you'll see that the kindest/node images referenced were using Kind 0.11.1

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/containers/KubernetesCluster.java
```diff
@@ -197,4 +209,49 @@ private String runKindCmd(String args) throws IOException, InterruptedException
...
     return output;
   }
+
+  /**
+   * Returns the kubectl version to use for the given kubernetes version. This is used to download
+   * the correct kubectl binary for the cluster.
+   *
+   * @return The kubectl version to use.
+   */
+  private static String getKubectlVersion() {
```
dbyron-sf, line 219: I'm torn about this.  Part of me really likes it.  Part of me thinks we end up testing something farther away than what we ship, because the default version of kubectl is the only one that's used by default.

Does it make sense to change the logic e.g. [here](https://github.com/spinnaker/clouddriver/blob/069140bae4cc171f9d0e85b7ec40141affd4b5b9/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutor.java#L698-L702) to dynamically choose the kubectl binary that matches the target cluster version?  Which I suppose would then mean on startup / periodically we'd have to query target clusters for their version?

.github/workflows/integration-tests-kubernetes.yml
jervi: You are correct, I'll update the comment to reference `KubernetesCluster.java` instead. But no, the mapping is only used for integration tests (and lives within the integration test sources). So the mapping is independent of what is defined in the Dockerfile (unless we implement the changes you [mention here](https://github.com/spinnaker/clouddriver/pull/6134#pullrequestreview-1844806402), of course).

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/containers/KubernetesCluster.java
```diff
@@ -197,4 +209,49 @@ private String runKindCmd(String args) throws IOException, InterruptedException
...
+    Pattern pattern = Pattern.compile("v([0-9]*\\.[0-9]*)");
+    Matcher matcher = pattern.matcher(IMAGE);
+    matcher.find();
+    String kubectlVersion;
+    switch (matcher.group(1)) {
```
dbyron-sf, line 224: Replace this with a map?

.github/workflows/integration-tests-kubernetes.yml
dbyron-sf: Aaah right, these tests download kubectl, they don't depend on what's in the docker image.

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/containers/KubernetesCluster.java
jervi: I'd love that feature! It would mean that the scope of this PR grows substantially, however... Do you want me to give it a go? What k8s versions would we support then? Also I haven't found a good way to automatically download the latest patch version of a given `kubectl` release. It would be awesome to not have to manually maintain that kubectl list in all of the dockerfiles and the deb packaging code.

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/containers/KubernetesCluster.java
jervi: Done

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/containers/KubernetesCluster.java
dbyron-sf: I wasn't thinking about dynamically downloading kubectl binaries based on clusters we discover.  I suppose that's a possibility, but could be tough to keep compliance folks happy.

I was more thinking about choosing from among the ones we've got.  It probably does belong in another PR.

For this PR, what do you think of continuing to test the default kubectl version against all the k8s cluster versions, and then one more round of testing using the mapping here, and if we're very lucky, removing dups?  That's way less than every client x every server, but also doesn't reduce the coverage we're getting today.

Then if we ever do dynamically choose the k8s version, we could drop the default round of testing.

One other thought that's been rolling around.  I'd love to have another step in the PR build github action that invokes a new gradle Test task to exercise the just-built docker image with testcontainers.  Not sure what the magic is to learn the name and tag, or perhaps we get to specify the tag and then know what to tell gradle.

With this, we could potentially adjust these tests to use the kubectl binaries in the docker image and not have to download them.

.github/workflows/integration-tests-kubernetes.yml
```diff
@@ -9,17 +9,35 @@ jobs:
...
+          - kubectl-version: 1.26.13
+            kubernetes-image: "kindest/node:v1.26.6@sha256:6e2d8b28a5b601defe327b98bd1c2d1930b49e5d8c512e1895099e4504007adb"
+          - kubectl-version: 1.25.16
+            kubernetes-image: "kindest/node:v1.25.11@sha256:227fa11ce74ea76a0474eeefb84cb75d8dad1b08638371ecf0e86259b35be0c8"
+          - kubectl-version: 1.24.17
+            kubernetes-image: "kindest/node:v1.24.15@sha256:7db4f8bea3e14b82d12e044e25e34bd53754b7f2b0e9d56df21774e6f66a70ab"
+          - kubectl-version: 1.23.17
+            kubernetes-image: "kindest/node:v1.23.17@sha256:59c989ff8a517a93127d4a536e7014d28e235fb3529d9fba91b3951d461edfdb"
+          - kubectl-version: 1.21.14
+            kubernetes-image: "kindest/node:v1.21.14@sha256:8a4e9bb3f415d2bb81629ce33ef9c76ba514c14d707f9797a01e3216376ba093"
```
jervi, line 40: A bit ugly to repeat all the kindest images here, but that was the best way I could find to conditionally add to an existing matrix. If anyone have some better solutions, I'm listening. [Relevant documentation](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations).

clouddriver-kubernetes/src/integration/java/com/netflix/spinnaker/clouddriver/kubernetes/it/containers/KubernetesCluster.java
jervi: > For this PR, what do you think of continuing to test the default kubectl version against all the k8s cluster versions, and then one more round of testing using the mapping here, and if we're very lucky, removing dups? That's way less than every client x every server, but also doesn't reduce the coverage we're getting today. 

I've removed the in-code kubectl selection stuff and reverted to just using GitHub Actions. It's a bit cleaner (albeit with a bit of duplication in the action matrix), and it achieves what you describe above: one round of testing with the default kubectl version and another round using the version that corresponds to the k8s version.

.github/workflows/integration-tests-kubernetes.yml
dbyron-sf: Yeah, it's clunky, but if it works, it's ok with me.  And from the looks of the PR checks that ran, it does:
image

dbyron-sf (approved):

PR #6136, opened 2024-01-08
remove failures from ./gradlew dependencies

dbyron-sf:
Various gradle cleanups with the goal of removing failures from `$ ./gradlew dependencies` output from all clouddriver modules.  Happily, all the failures are gone.

FWIW, I often use this to poke around for dependency changes
```
alias gradle-all-deps='./gradlew --no-parallel --stacktrace dependencies $(./gradlew -q projects \
    | grep -Fe ---\ Project \
    | sed -Ee "s/^.+--- Project '"'([^']+)'/\1:dependencies/"'" | sort)'
```
inspired by https://stackoverflow.com/a/65945289

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,6 +54,7 @@ subprojects {
     sourceSets.main.groovy.srcDirs += ["src/main/java"]
 
     dependencies {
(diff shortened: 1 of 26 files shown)


jvz (approved):
It's repetitive stuff like this where Maven really shines compared to Gradle. :D

PR #6137, opened 2024-01-10
fix(docker): Parse date into Instant when using sortTagsByDate

wwalters12:
Potential fix for https://github.com/spinnaker/spinnaker/issues/6921

I know this feature isn't recommend (https://github.com/spinnaker/spinnaker/issues/5538), but we have a small repository we are caching and sorting the tags for it would be nice.

We are using Redis and the date value is cached as a string:
```
bash-4.4# redis-cli --scan --pattern dockerRegistry:taggedImage:attributes:dockerRegistry:taggedImage:**** | xargs redis-cli mget
1) "{\"account\":\"***\",\"date\":\"2024-01-08T15:23:15.280797435Z\",\"name\":\"***\"}"
```

With this change, the date value will be parsed to an `Instant` if the date is cached as a `String`.

### DIFF
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/controllers/DockerRegistryImageLookupController.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/controllers/DockerRegistryImageLookupController.groovy
@@ -51,7 +51,21 @@ class DockerRegistryImageLookupController {
       Keys.getTaggedImageKey(account, repository, "*")
     ).sort { a, b ->
       if (credentials.sortTagsByDate) {
-        b.attributes.date.epochSecond <=> a.attributes.date.epochSecond
+        Instant dateA
+        if (a.attributes.date instanceof String) {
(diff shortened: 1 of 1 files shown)


dbyron-sf:
Looks like there are some build failures to resolve.  Could you look into adding an automated test for this please, and ideally explaining how info shows up as instants vs. strings?

wwalters12:
> Looks like there are some build failures to resolve. Could you look into adding an automated test for this please, and ideally explaining how info shows up as instants vs. strings?

@dbyron-sf Added a missing import, that should fix the build. And yes, I'll look into adding a test to verify the tags can be sorted if the date is stored as a String or Instant.

I think the previous change to this fixed the functionality for using a SQL cache (https://github.com/spinnaker/clouddriver/pull/5098) where the image date is stored as a map with the key `epochSecond`, but when using a Redis cache the date is stored as a string.

I'm also realizing this means using the `epochSecond` key is incorrect as this will still break if the date is stored as a string instead of a map. I'll fix that and get a test added in the next commit.

jasonmcintosh:
@wwalters12 did you get a chance to work on any of this? :)

PR #6139, opened 2024-01-22
fix(provider/docker): Support multiple challenges from `WWW-Authenticate` header

Badbond:
Fixes https://github.com/spinnaker/spinnaker/issues/6922.

Ultimately, considers the `WWW-Authenticate` header value as a `List<String>` as opposed to casting it directly to a `String`.

According to [Mozilla's `WWW-Authenticate` header documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate), it is possible for the server to respond with multiple challenges through multiple header values. As such, I decided not to simply fetch the first header value, but now loop over the values and handle the first to match the `bearer` or `basic` prefix. It might be time to extract this to a separate method, but I kept the diff as small as possible. 

I am not too familiar with Groovy nor Spock, so feel free to apply improvements where you deem fit.

The first commit serves as a regression test that the second commit fixes.

### DIFF
--- a/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
+++ b/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
@@ -497,7 +497,7 @@ class DockerRegistryClient {
         // note, this is a workaround for registries that should be returning
         // 401 when a token expires
         if ([400, 401].contains(status)) {
-          String authenticateHeader = null
+          List<String> authenticateHeader = null
 
           error.headers.entrySet().forEach { header ->
             if (header.key.equalsIgnoreCase("www-authenticate")) {
@@ -505,7 +505,7 @@ class DockerRegistryClient {
             }
           }
 
-          if (!authenticateHeader) {
+          if (!authenticateHeader || authenticateHeader.isEmpty()) {
             log.warn "Registry $address returned status $status for request '$target' without a WWW-Authenticate header"
             tokenService.clearToken(target)
             throw error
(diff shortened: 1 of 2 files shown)


dbyron-sf:
Can you look into the failures please?

Badbond:
> Can you look into the failures please?

Based on [this line](https://github.com/spinnaker/clouddriver/actions/runs/7612721025/job/20735275406?pr=6139#step:6:307), it looks like a transient issue while initializing (a) Testcontainer(s). Locally the tests pass for me, and debugging them shows that they do not execute any code path in the changed classes. Could you please retrigger the build? If it still fails, I'd be happy to take a further look. It would be nice to have some more logs of the failing tests, if you have a way of retrieving those.

dbyron-sf (approved):

aRobinson-R7:
@Badbond @dbyron-sf can we get this back-ported to v1.33 release branch please?

dbyron-sf:
@mergifyio backport release-1.33.x

aRobinson-R7:
thank you @dbyron-sf

snorlaX-sleeps:
also @dbyron-sf - will this end up in the next Halyard release? Error is also occurring when editing existing dockerHub accounts added

snorlaX-sleeps:
@dbyron-sf I meant specifically that the logic from this is imported into Halyard (Afaik from reading other issues)
I was just asking if these sort of patches make it into the next release of it

dbyron-sf:
Halyard has its own release cycle that's separate from the rest of spinnaker.  As far as I know this logic isn't relevant to halyard.  Please correct me if that's not true.

Badbond:
As mentioned in the ticket, it also affects Halyard. This is because it uses Clouddriver's Docker client [here](https://github.com/spinnaker/halyard/blob/20dc107751281c462259c8df275475cb5dd126de/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/dockerRegistry/DockerRegistryAccountValidator.java#L193) for performing validations. IIUC this requires Halyard to use the new Clouddriver release and issue a release as well. 

However, it _may_ be that older versions of Halyard (< 1.63.0) are not affected as they rely on an older version of Clouddriver predating this bug. I did not backtrack this all the way (yet).

dbyron-sf:
OK, so as a start, let me publish new clouddriver jars from the HEAD of master that include this fix.  Then the next step will be to release a new version of halyard.

dbyron-sf:
OK, I've published those clouddriver jars and released version 1.64 of halyard.  Can you try it out please?

Badbond:
As for this particular fix in Clouddriver, I was able to upgrade Spinnaker 1.31 -> 1.33 using Halyard 1.62.0 while specifying the `5.84.0-unvalidated` tag as `artifactId` override for Clouddriver. The issue is not apparent anymore. :tada:

dbyron-sf:
@Badbond I have a feeling https://github.com/spinnaker/halyard/pull/2144 fixes the halyard startup issues.

Badbond:
Thank you @dbyron-sf, it is working! :+1: We were able to upgrade to Spinnaker 1.33.0 and Halyard 1.65.0 now. Unfortunately we did stumble upon some new unrelated issues while testing and upgrading, which I have filed for completion here:

- https://github.com/spinnaker/spinnaker/issues/6941
- https://github.com/spinnaker/spinnaker/issues/6942

snorlaX-sleeps:
Thanks @dbyron-sf ! (Unfortunately I cannot test as our fix was just to remove the ability to access Dockerhub at all 😂)

PR #6153, opened 2024-02-19
fix: Change the agent type name to not include the account name since this would generate LOTS of tables and cause problems long term

ovidiupopa07:


jasonmcintosh (approved):
Wish we had better tests around this and the impacts... but given the fix... merging

jasonmcintosh:
Need to get this into master release too

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
```diff
@@ -209,7 +209,7 @@ private List<InstanceTypeInfo> getInstanceTypes(AmazonEC2 ec2) {
 
   @Override
   public String getAgentType() {
-    return String.format("%s/%s/%s", account.getName(), region, getClass().getSimpleName());
+    return getClass().getSimpleName() + "/" + region;
```
dbyron-sf, line 212: With this change, does the class still need to implement AccountAware?

jasonmcintosh: Long term yes - it's mostly impacting the storage layer at this time (table per account) though think there are other changes we COULD make to change how the cache data is fed to the storage and query layers.  EVENTUALLY be interesting if the AWS API's would include account filters so instance types could be picked BY account vs "Global"

jasonmcintosh:
Think we'll close this PR, do one to master FYI

clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java
jasonmcintosh: No, instance types are set by account.  Some accounts restrict access to GPU instance types for example.  However the APIs that request instance types are NOT account aware APIs (currently). 

Instead of changing agentType we COULD have changed the cache data that's fed back.  Example for image caches: 
https://github.com/spinnaker/clouddriver/blob/3f31190d40419218fd1d31ca78bb1661f1ef881a/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/agent/ImageCachingAgent.groovy#L107
but the instance type uses:
https://github.com/spinnaker/clouddriver/blob/6759d564185872f35f65d957383231f4745c40ef/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java#L94

The OTHER impact to this is if I recall the cats locks on which can run these concurrently.

jasonmcintosh: So image caching agent operates as this does, but image caching agent does NOT use agentType as part of the cache data base name.

jasonmcintosh: This PR basically reverts back to the 1.28 behavior on agent type:
https://github.com/spinnaker/clouddriver/blob/release-1.28.x/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonInstanceTypeCachingAgent.java#L214
so this fixes the "table per account" that way, but... the instance type caching agent is just... not ideal in general.

ovidiupopa07:
Closign in favor of https://github.com/spinnaker/clouddriver/pull/6158

PR #6156, opened 2024-02-22
feat(aws): Install AWS CLI v2 and upgrade aws-iam-authenticator

jervi:
AWS CLI v1 is becoming harder to use with newer Kubernetes versions. If you use `aws eks get-token` for EKS authentication instead of `aws-iam-authenticator`, you are stuck with `apiVersion: client.authentication.k8s.io/v1alpha1`, because `v1beta1` isn't supported by AWS CLI v2. Upgrading to v2 fixes this.

### DIFF
--- a/clouddriver-web/pkg_scripts/postInstall.sh
+++ b/clouddriver-web/pkg_scripts/postInstall.sh
@@ -3,6 +3,8 @@
 # Remember to also update Dockerfile.*
 KUBECTL_DEFAULT_RELEASE=1.22.17
 KUBECTL_RELEASES="${KUBECTL_DEFAULT_RELEASE} 1.26.12 1.27.9 1.28.5 1.29.0"
+AWS_CLI_VERSION=2.15.22
+AWS_AIM_AUTHENTICATOR_VERSION=0.6.14
 
 # ubuntu
 # check that owner group exists
@@ -26,6 +28,18 @@ install_kubectl() {
   fi
 }
 
+install_awscli2() {
+  curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-${AWS_CLI_VERSION}.zip" -o "awscliv2.zip"
+  unzip awscliv2.zip
+  ./aws/install
+  rm -rf ./awscliv2.zip ./aws
+
+  curl "https://github.com/kubernetes-sigs/aws-iam-authenticator/releases/download/v${AWS_AIM_AUTHENTICATOR_VERSION}/aws-iam-authenticator_${AWS_AIM_AUTHENTICATOR_VERSION}_linux_amd64" -O aws-iam-authenticator
+  chmod +x ./aws-iam-authenticator \
+  mv ./aws-iam-authenticator /usr/local/bin/aws-iam-authenticator
+}
+
 install_kubectl
+install_awscli2
 
 install --mode=755 --owner=spinnaker --group=spinnaker --directory /var/log/spinnaker/clouddriver
--- a/Dockerfile.java11.slim
+++ b/Dockerfile.java11.slim
@@ -4,9 +4,8 @@ ARG TARGETARCH
 
 ENV KUBECTL_DEFAULT_RELEASE=1.22.17
 ENV KUBECTL_RELEASES="${KUBECTL_DEFAULT_RELEASE} 1.26.12 1.27.9 1.28.5 1.29.0"
-ENV AWS_CLI_VERSION=1.22
-ENV AWS_CLI_S3_CMD=2.0.2
-ENV AWS_AIM_AUTHENTICATOR_VERSION=0.5.9
+ENV AWS_CLI_VERSION=2.15.22
+ENV AWS_AIM_AUTHENTICATOR_VERSION=0.6.14
 ENV GOOGLE_CLOUD_SDK_VERSION=412.0.0
 ENV ECR_TOKEN_VERSION=v1.0.2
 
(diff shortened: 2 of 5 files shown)


dbyron-sf (approved):

jasonmcintosh:
Yes though I'd prefer to NOT have it in 1.34 as it'd break kubeconfig with out a good migration path currently documented.  The versions currently included SHOULD work with v1beta (aws cli and aws-iam-authenticator) allowing a migration path to v1beta1.  Note I believe v1beta1 is also deprecated at this point in time and slated for removal.  aws-iam-authenticator in 0.6 removed v1alpha entirely but in older versions supported both exec APIs and aws cli 1.22 SHOULD work with both APIs as well.  I'd get 1.34 out first with a docs PR saying "Caution! You need to upgrade your kubeconfig files" then in 1.35 upgrade these to remove the v1alpha1 support.  

Kube exec API calls pass an env variables exec's can use to determine which API spec to support.  
https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/439
This PR and
https://github.com/aws/aws-cli/pull/6476/files
for more information.

Docs PR for reference (which could use some language tweaks) calling out that this will break users:
https://github.com/spinnaker/spinnaker.io/pull/371

clouddriver-web/pkg_scripts/postInstall.sh
```diff
@@ -26,6 +28,18 @@ install_kubectl() {
   fi
 }
 
+install_awscli2() {
+  curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-${AWS_CLI_VERSION}.zip" -o "awscliv2.zip"
```
jasonmcintosh, line 32: Note arch here is wrong - needs to support arch argument to work on ARM instances.

jasonmcintosh (changes requested):
Arch missing on the one CLI installer

jasonmcintosh:
Note looks like v1beta1 is still supported/not deprecated if I've got the right spot https://kubernetes.io/docs/reference/config-api/client-authentication.v1beta1/ though there's a plain v1.  I'd have to check on state of v1 support on the cli tools.

clouddriver-web/pkg_scripts/postInstall.sh
```diff
@@ -26,6 +28,18 @@ install_kubectl() {
...
+  unzip awscliv2.zip
+  ./aws/install
+  rm -rf ./awscliv2.zip ./aws
+
+  curl "https://github.com/kubernetes-sigs/aws-iam-authenticator/releases/download/v${AWS_AIM_AUTHENTICATOR_VERSION}/aws-iam-authenticator_${AWS_AIM_AUTHENTICATOR_VERSION}_linux_amd64" -O aws-iam-authenticator
```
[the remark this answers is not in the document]
jervi: Same

jervi:
> The versions currently included SHOULD work with v1beta (aws cli and aws-iam-authenticator) allowing a migration path to v1beta1.

I'm not entirely sure why it fails for me, but running Clouddriver nightly builds (with AWS CLI v1), it fails when using `kubectl` v1.28 and `v1beta1`. It may work with older kubectl versions though. The same combination works with AWS CLI v2 (running a version with this PR applied). I'm pretty sure it should work with `v1` as well given that you're running a new enough Kubernetes version.

The error I'm getting is:
```
no kind "ExecCredential" is registered for version "client.authentication.k8s.io/v1alpha1" in scheme "pkg/client/auth/exec/exec.go:62
```
Not sure where it gets `v1alpha1` from, because the kubeconfig files are using `v1beta1`.

That being said, I'm not arguing for this to be in v1.34. I'm fine with it going in 1.35.

jasonmcintosh:
Hrmm, I'm thinking I may have missed this - 1.22 is older - Harness (aka armory) has a different build that's running 1.29.  Let me do some testing, but it's possible for 1.34 we should bump to 1.29 then note that "alpha1" is deprecated, then 1.35 release with this.

clouddriver-web/pkg_scripts/postInstall.sh
jasonmcintosh: Hrm, that's... potentially problematic.  I need to look at this more then...

jasonmcintosh:
Pointing out here:
* AWS CLI 1.24 supports both v1alpha1 & v1beta1 - so look to upgrade that for 1.34 then this PR for 1.35.  Hope to come back look at the post install scripts for support for ARM stuff.

clouddriver-web/pkg_scripts/postInstall.sh
jervi: I can fix it if you want. But I don't think anyone has ever installed the package on an ARM instance. We are probably one of the last companies to actually use these packages, and now we're also migrating to k8s. Looking at the script it is also pretty lacking, without support for anything but AWS. But just let me know if you want me to add support for ARM before we merge this.

jasonmcintosh: *wince* I'd love ARM support, but you're probably right on lack of usage. That said I've missed stuff like this before and it's bitten me :(

jasonmcintosh (approved):

chris-h-phillips:
I just tried to use the latest 1.35.2 docker image. I use the `aws eks get-token` method. Stages fail with: `/usr/local/bin/aws: No such file or directory`

Indeed if you go looking in the container:

```bash
bash: /usr/local/bin/aws: No such file or directory
bash-5.1$ ls -al /usr/local/bin/aws
lrwxrwxrwx    1 root     root            37 Aug 20 17:48 /usr/local/bin/aws -> /usr/local/aws-cli/v2/current/bin/aws
bash-5.1$ ls -al /usr/local/aws-cli/v2/current/bin
total 0
drwxr-xr-x    2 root     root            38 Aug 20 17:48 .
drwxr-xr-x    4 root     root            44 Aug 20 17:48 ..
lrwxrwxrwx    1 root     root            11 Aug 20 17:48 aws -> ../dist/aws
lrwxrwxrwx    1 root     root            21 Aug 20 17:48 aws_completer -> ../dist/aws_completer
bash-5.1$ ls -al ../dist/aws
ls: ../dist/aws: No such file or directory
```

@jervi I'm wondering if you might have any insight here?

jervi:
> @jervi I'm wondering if you might have any insight here?

You're right. Seems like an Alpine Linux problem. An easy workaround is to use `spinnaker-1.35.2-ubuntu` for Clouddriver instead.
```
kubectl run -i --namespace default --tty --rm debug --image=us-docker.pkg.dev/spinnaker-community/docker/clouddriver:spinnaker-1.35.2-ubuntu --restart=Never -- bash
If you don't see a command prompt, try pressing enter.
spinnaker@debug:/$ aws --version
aws-cli/2.15.22 Python/3.11.6 Linux/5.10.219-208.866.amzn2.x86_64 exe/x86_64.ubuntu.22 prompt/off
```

jervi:
Seems like there's pre-built binaries available for Alpine now: https://pkgs.org/download/aws-cli
The docker image should probably be changed to just install AWS CLI v2 using `apk`.

dbyron-sf:
> Seems like there are pre-built binaries available for Alpine now: https://pkgs.org/download/aws-cli The docker image should probably be changed to just install AWS CLI v2 using `apk`.

Mind taking a crack at that @jervi ?

jervi:
https://github.com/spinnaker/clouddriver/pull/6279

PR #6158, opened 2024-02-22
fix: Change the agent type name to not include the account name since this would generate LOTS of tables and cause problems long term

ovidiupopa07:
The standard in spinnaker is to use a single table for a cache agent TYPE without the account name,


jasonmcintosh:
Full conversation on some of the change
https://github.com/spinnaker/clouddriver/pull/6153#pullrequestreview-1896368173

jasonmcintosh (approved):
I don't know that I LIKE this but this is an area long term could do differently/better.

ovidiupopa07:
@Mergifyio backport release-1.31.x release-1.32.x release-1.33.x

PR #6161, opened 2024-02-26
fix(builds): Fixes JVM memory for builds to be set and lowers compilation

jasonmcintosh:


dbyron-sf:
Looks good.  Does it make sense to reduce Xmx and Xms in https://github.com/spinnaker/clouddriver/blob/6d6afde4b350710dffc8e73b1831331b87438daa/.github/workflows/integration_tests.yml#L10 too?

jasonmcintosh:
Gradle ops should override the local and ... may not even be necessary depending on the phase.  The stack memory is ALSO really high to me...

dbyron-sf (approved):

PR #6186, opened 2024-04-11
fix(account): Make docker account properties comparable, thus reduce load on DB and docker registry

kbatalin:
With enabled dynamic account management for docker accounts, `BasicCredentialsLoader` regularly loads accounts from a storage, filters changed accounts and updates them in `MapBackedCredentialsRepository`. Then, for all updated accounts `CredentialsLifecycleHandler` reschedules caching agents (basically, just deletes the agents and adds them again).

Docker accounts settings class doesn't implement `equals` and `hashCode` so even identical settings are considered as different, and Clouddriver reschedules caching agents each polling. This leads to increased polling frequency and polling interval isn't respected, because recreated caching agents are ready to run right after scheduling.

The final load on the DB depends on a number of Clouddriver pods, docker accounts and dynamic accounts polling interval. In our case the docker images table is top by reads, with this fix we decreased the reads by 20% and 50% (`cats_v1_taggedImage` and `cats_v1_imageId` tables respectively).

I didn't implement `equals` and `hashCode` for other `CredentialsDefinition`'s child classes because I couldn't test other classes in a real environment. But I'm happy to implement the methods for all of them if you're okay with this.


clouddriver-docker/src/test/groovy/com/netflix/spinnaker/clouddriver/docker/registry/config/DockerRegistryConfigurationPropertiesSpec.groovy
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2024 Netflix, Inc.
```
dbyron-sf, line 2: Netflix?

clouddriver-docker/src/test/groovy/com/netflix/spinnaker/clouddriver/docker/registry/config/DockerRegistryConfigurationPropertiesSpec.groovy
```diff
@@ -0,0 +1,47 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver.docker.registry.config
+
+import spock.lang.Specification
+
+class DockerRegistryConfigurationPropertiesSpec extends Specification {
```
dbyron-sf, line 21: Ideally new code (even test code) would not be in groovy.  Any chance you can convert this to java?  Thanks!

clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/config/DockerRegistryConfigurationProperties.groovy
```diff
@@ -18,13 +18,16 @@ package com.netflix.spinnaker.clouddriver.docker.registry.config
 
 import com.fasterxml.jackson.annotation.JsonTypeName
 import com.netflix.spinnaker.credentials.definition.CredentialsDefinition
+import groovy.transform.EqualsAndHashCode
 import groovy.transform.ToString
 
 @ToString(includeNames = true)
+@EqualsAndHashCode
```
dbyron-sf, line 25: The new test still passes without this.  Did you try without it and see the same reduced reads?  I suppose it doesn't hurt to keep it, but if we don't need it, I'd rather leave it out.

kbatalin: It's unused, deleted 👍

dbyron-sf (approved):

PR #6187, opened 2024-04-12
feat(docker): Add RBAC support to Docker accounts

jvz:
This is a minimal update to the Docker Registry account classes to support RBAC. The read permission is used for viewing things related to the account (tags and images) while the write permission is used for gating access to the account in atomic operations. As the existing code technically has a "required group membership" list, I've included migration code in case that was somehow set by anyone in existing accounts (the lack of an equivalent value in the class that's bound to Spring properties makes me think this isn't the case for anyone). As a bonus, I noticed an unused variable (`dockerconfigFile`) which I've passed along to the appropriate builder method that it was forgotten by.

Due to the use of annotations for security, there are no additional tests defined here as it would require either an extremely sophisticated integration test setup or use of something completely unrelated to Spring (and possibly Java).


dbyron-sf:
Couldn't we get at least some test coverage by mocking FiatService?

jvz:
No because that's unrelated to how this all works under the hood. All existing test code related to permission checks mocks high level APIs like hasPermission which defeats the purpose of testing anything there. Search existing code for these annotations and you'll see nothing is explicitly tested around them. The closest you can come to it is MockMVC where you end up mocking all the authorization logic and testing nothing practical in result. This all stems from the choice of Spring as the framework and Java annotations as the implementation detail behind authorization (which has no standard testing story).

dbyron-sf:
At least you can test that the right things happen when hasPermission returns various things....by mocking a PermissionEvaluator, or using FiatPermissionEvaluator and mocking FiatService, or something.  Seems worth it to me.

jvz:
Yeah, I'll try to figure something out. Hopefully something useful that can be copied in future tests.

clouddriver-docker/src/test/java/com/netflix/spinnaker/clouddriver/docker/registry/controllers/DockerRegistryImageLookupControllerTest.java
```diff
@@ -0,0 +1,211 @@
...
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureWebMvc;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Import;
+import org.springframework.security.test.context.support.WithMockUser;
+import org.springframework.test.web.servlet.MockMvc;
+
+@SpringBootTest(classes = TestConfig.class, properties = "services.fiat.cache.max-entries=0")
```
jvz, line 65: I'll note that the cache setting here makes it so we can use `@WithMockUser` on all the tests. Without this, each test needs a unique username since the permissions get cached long enough to otherwise fail subsequent tests.

dbyron-sf (approved):
Thank you!

PR #6194, opened 2024-04-19
fix(databaseChangeLog): fix for 'addAfterColumn is not allowed on postgresql'

kirangodishala:
This PR addresses https://github.com/spinnaker/spinnaker/issues/6942

The following PR provides more details on the issue and similar fix provided for orca: https://github.com/spinnaker/orca/pull/4594


dbyron-sf (approved):

xibz:
How difficult would it be to have a containerized test for this? Similar to what we do for redis?

PR #6093, opened 2023-11-13
feat(gcp): provide a configurable option to bypass gcp account health check.

armory-abedonik:

PR #6200, opened 2024-05-01
fix(gcp): Relaxed health check for GCP accounts

christosarvanitis:
The majority of cloud providers (AWS, K8s) and dockerRegistry have already implementations for relaxed health checks via a configuration property in the respective provider block.

On GCP provider though it was attempted to be addressed by https://github.com/spinnaker/clouddriver/pull/6093 which however was not the correct implementation since the Health check is registered in initialisation and it is a scheduled task. 

This PR fixes the issue https://github.com/spinnaker/spinnaker/issues/3923 for the GCP provider


ovidiupopa07 (approved):

jasonmcintosh:
Ideally this gets changed a bit in the future.  With the Bean, it would disable the thread & scheduled job which would output a log message every 30 seconds.  Since it was also a component that Bean disable didn't help this.  Doing this method, we still get the log message, the scheduled run that outputs the log message every 30 seconds, plus the whole if check thing.  It's MINOR but... we shouldn't really load these health checks this way.  Be good to fix all of them to disable them vs. the current implementation.

jasonmcintosh:
@Mergifyio backport release-1.28.x

jasonmcintosh:
@Mergifyio backport release-1.34.x release-1.33.x release-1.32.x

PR #6206, opened 2024-05-01
feat(build): add fiat-integration module to exercise the just-built docker imageTest docker image

dbyron-sf:
Also, add a healthcheck to the docker images.


Dockerfile.java11.slim
```diff
@@ -64,5 +65,5 @@ COPY clouddriver-web/build/install/clouddriver /opt/clouddriver
 RUN mkdir -p /opt/clouddriver/plugins && chown -R spinnaker:nogroup /opt/clouddriver/plugins
 
 USER spinnaker
-
+HEALTHCHECK CMD curl http://localhost:7002/health | grep UP || exit 1
```
xibz, line 68: Hmm, this wouldn't work right? Because it's a complete object.

```
{"status":"DOWN","components":{"clientConfigServer":{"status":"UNKNOWN","details":{"error":"no property sources located"}},"configServer":{"status":"UP","details":{"repositories":[{"name":"app","profiles":["default"],"label":null}]}},"db":{"status":"UP","details":{"database":"POSTGRES"}},"discoveryComposite":{"description":"Discovery Client not initialized","status":"UNKNOWN","components":{"discoveryClient":{"description":"Discovery Client not initialized","status":"UNKNOWN"}}},"diskSpace":{"status":"UP","details":{"total":1000240963584,"free":361212166144,"threshold":10485760,"exists":true}},"dockerRegistry":{"status":"UP"},"kubernetes":{"status":"UP"},"ping":{"status":"UP"},"reactiveDiscoveryClients":{"description":"Discovery Client not initialized","status":"UNKNOWN","components":{"Simple Reactive Discovery Client":{"description":"Discovery Client not initialized","status":"UNKNOWN"}}},"redisHealth":{"status":"UP","details":{"maxIdle":100,"minIdle":25,"numActive":0,"numIdle":3,"numWaiters":0}},"refreshScope":{"status":"DOWN"}}}
````

We need to check if the root status is `UP` and not a wide search, otherwise it could be a false positive.

Comment (jasonmcintosh) [reply Dockerfile.java11.slim:68]: MOST of the time it's only a complete object IF the "management" exposes ALL data.  USUALLY it's a single line check.

Comment (xibz) [reply Dockerfile.java11.slim:68]: Understood, but better safe to assume that things may be configured differently, than have a pleasant surprise :p

Comment (jasonmcintosh) [reply Dockerfile.java11.slim:68]:
```
{"status":"DOWN"}
```
AKA this data is NORMALLY what's returned unless you use:  
``` 
       management:
          endpoints.web:
            exposure.include: "*"
```

jasonmcintosh: (or similar)

jasonmcintosh: BUT... jq adds more deps... and this isn't used in k8s - only in say testiing via docker CLI.  I'm not AGAINST JQ - particularly since can be useful... just a "which is the less evil"

xibz: It's a small dep for preventing bugs. So I'd argue that is probably more satisfactory. We could also use something like perl, awk, etc to achieve this, but the command is going to be much more unreadable if dependencies are that much of a concern. If we build our docker images intelligently, it should not be too bad to add a single binary

xibz: And we are already adding curl, so adding jq doesn't seem to be that bad since curl isn't used in k8s either

dbyron-sf: How's https://github.com/spinnaker/clouddriver/pull/6206/commits/ad6cf9a8c1336369b0fb89ea0e9412c812ee7c9f ?

xibz (approved):
Approved! I left a comment, but it seemed to have vanished lol. But good approach! Forgot about `--fail` and no extra deps!

dbyron-sf:
Let's see if mergify will let me backport this even though it's marked as a feature....There's a bug migrating from release 1.32.x to the current (5-jun-24) HEAD of master when running postgres, and there's a way to demonstrate this with an extension of these tests...but to do that we need a docker image with a HEALTHCHECK in the release-1.32.x branch...

dbyron-sf:
@mergifyio backport release-1.32.x

dbyron-sf:
And similarly, let's backport to 1.33.x and 1.34.x so we can verify fixes there too.

dbyron-sf:
@mergifyio backport release-1.33.x release-1.34.x

PR #6210, opened 2024-05-15
fix(ClusterController): Fix GetClusters to return all providers clusterNames of an application

christosarvanitis:
When an application includes clusters for more than 2 cloud-providers then the GET /clusters endpoint was returning only the last 2 providers clusterNames instead of ALL the clusters deployed by the application. 

We didnt notice this bug existed because Deck is rendering based on the clusters and serverGroups and when not filtered the serverGroups was including the missing clusters.

It is noticeable though for the end-user when the SETTINGS.onDemandClusterThreshold  is used to make a searchable rendering of clusters/serverGroups in Deck.


jasonmcintosh:
@mergifyio backport release-1.32.x release-1.33.x release-1.34.x

PR #6217, opened 2024-05-22
fix(liquibase): fix checkSum errors occurring with spinnaker upgrade

kirangodishala:
This PR fixes checkSum errors caused after liquibase upgrade(to 4.24.0) as the checksum calculation underwent some changes causing this issue.

This PR fixes the checksum errors by adding validChecksum tags to the failing changesets.

Added an integration test to verify the migration scenario with postgres. This will be useful to identify and avoid issues when liquibase is upgraded or if any changesets are modified.

Similar orca PR for reference: https://github.com/spinnaker/orca/pull/4727

### DIFF
--- a/clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/BaseContainerTest.java
+++ b/clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/BaseContainerTest.java
@@ -40,7 +40,7 @@ public class BaseContainerTest {
 
   protected final Network network = Network.newNetwork();
 
-  private static final int CLOUDDRIVER_PORT = 7002;
+  protected static final int CLOUDDRIVER_PORT = 7002;
 
   protected GenericContainer<?> clouddriverContainer;
 
--- a/clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/PostgresMigrationContainerTest.java
+++ b/clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/PostgresMigrationContainerTest.java
@@ -16,14 +16,11 @@
 
 package com.netflix.spinnaker.clouddriver;
 
-import static org.junit.jupiter.api.Assumptions.assumeTrue;
-
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.time.Duration;
 import java.util.Map;
 import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
@@ -51,17 +48,14 @@ public class PostgresMigrationContainerTest extends BaseContainerTest {
 
   private GenericContainer<?> clouddriverInitialContainer;
 
-  private static DockerImageName previousDockerImageName;
+  // this is the latest image that is still running on liquibase 3.10.3 which create the conditions
+  // similar to real scenario so that test identifies when validChecksums are not added in the later
+  // version of clouddriver where higher liquibase versions are used
+  private static final DockerImageName previousDockerImageName =
+      DockerImageName.parse(
+          "us-docker.pkg.dev/spinnaker-community/docker/clouddriver:5.82.2-dev-release-1.32.x-7a8e6e8b3-202406051721-unvalidated");
 
   private String jdbcUrl = "";
-
-  @BeforeAll
-  static void setupInitial() {
-    String fullDockerImageName = System.getenv("PREVIOUS_FULL_DOCKER_IMAGE_NAME");
-    // Skip the tests if there's no docker image.  This allows gradlew build to work.
-    assumeTrue(fullDockerImageName != null);
-    previousDockerImageName = DockerImageName.parse(fullDockerImageName);
-  }
 
   @BeforeEach
   void setup() throws Exception {
@@ -83,7 +77,7 @@ public class PostgresMigrationContainerTest extends BaseContainerTest {
     clouddriverInitialContainer =
         new GenericContainer(previousDockerImageName)
             .withNetwork(network)
-            .withExposedPorts(7002)
+            .withExposedPorts(CLOUDDRIVER_PORT)
             .waitingFor(Wait.forHealthcheck().withStartupTimeout(Duration.ofSeconds(120)))
             .dependsOn(postgres)
             .withEnv("SPRING_APPLICATION_JSON", getSpringApplicationJson());
--- a/clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/PostgresMigrationContainerTest.java
+++ b/clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/PostgresMigrationContainerTest.java
@@ -1,0 +1,142 @@
+/*
+ * Copyright 2024 Salesforce, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.clouddriver;
+
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.time.Duration;
+import java.util.Map;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.PostgreSQLContainer;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.junit.jupiter.Testcontainers;
+import org.testcontainers.utility.DockerImageName;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@Testcontainers
+public class PostgresMigrationContainerTest extends BaseContainerTest {
+
+  private static final Logger logger =
+      LoggerFactory.getLogger(PostgresMigrationContainerTest.class);
+
+  private static final String POSTGRES_NETWORK_ALIAS = "postgresHost";
+
+  private static final int POSTGRES_PORT = 5432;
+
+  private PostgreSQLContainer<?> postgres;
+
+  private GenericContainer<?> clouddriverInitialContainer;
+
+  private static DockerImageName previousDockerImageName;
+
+  private String jdbcUrl = "";
+
+  @BeforeAll
+  static void setupInitial() {
+    String fullDockerImageName = System.getenv("PREVIOUS_FULL_DOCKER_IMAGE_NAME");
+    // Skip the tests if there's no docker image.  This allows gradlew build to work.
+    assumeTrue(fullDockerImageName != null);
+    previousDockerImageName = DockerImageName.parse(fullDockerImageName);
+  }
+
+  @BeforeEach
+  void setup() throws Exception {
+    postgres =
+        new PostgreSQLContainer<>("postgres:15")
+            .withDatabaseName("clouddriver")
+            .withUsername("postgres")
+            .withPassword("postgres")
+            .withNetwork(network)
+            .withNetworkAliases(POSTGRES_NETWORK_ALIAS)
+            .withInitScript("postgres_init.sql")
+            .withReuse(true);
+    postgres.start();
+    jdbcUrl =
+        String.format("jdbc:postgresql://%s:%d/clouddriver", POSTGRES_NETWORK_ALIAS, POSTGRES_PORT);
+
+    // Start the first clouddriver(from previous release) container so that all the db changelog
+    // sets are executed
+    clouddriverInitialContainer =
+        new GenericContainer(previousDockerImageName)
+            .withNetwork(network)
+            .withExposedPorts(7002)
+            .waitingFor(Wait.forHealthcheck().withStartupTimeout(Duration.ofSeconds(120)))
+            .dependsOn(postgres)
+            .withEnv("SPRING_APPLICATION_JSON", getSpringApplicationJson());
+    clouddriverInitialContainer.start();
+    Slf4jLogConsumer logConsumer = new Slf4jLogConsumer(logger);
+    clouddriverInitialContainer.followOutput(logConsumer);
+    clouddriverInitialContainer.stop();
+
+    // Start the second clouddriver(latest) container to validate migration
+    clouddriverContainer
+        .dependsOn(postgres)
+        .withEnv("SPRING_APPLICATION_JSON", getSpringApplicationJson())
+        .start();
+
+    clouddriverContainer.followOutput(logConsumer);
+  }
+
+  private String getSpringApplicationJson() throws JsonProcessingException {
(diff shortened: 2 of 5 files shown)


clouddriver-sql/src/main/resources/db/changelog/20181120-cats.yml
```diff
@@ -216,6 +216,7 @@ databaseChangeLog:
         newDataType: varchar(255)
 
 - changeSet:
+    validCheckSum: 8:e252377f81b5153444f82dd257b1db7a
```
dbyron-sf, line 219: Can you explain where this value (and the one below) come from?

kirangodishala: This checksum value is corresponding to the changeSet - `20190913-task-sagaids.yml::add-task-sagaids-column` and liquibase 4.27.0 combination. i.e, for the same changeSet and liquibase 4.27.0, we get the same checksum - 8:e252377f81b5153444f82dd257b1db7a in any machine.

So when migrating from liquibase 3.10.0 to 4.27.0 (spinnaker <1.33.0 to Spinnaker >=1.33.0) I received this error:
 `...db/changelog/20181120-cats.yml::20181120-cats.yml::application-index::afeldman was: 8:03b00d0af09f2e7081d187f246ea4d26 but is now: 8:e252377f81b5153444f82dd257b1db7a`

 Similarly for the other changeset I got this - `db/changelog/20190913-task-sagaids.yml::add-task-sagaids-column::robzienert was: 8:91cfabe4a8fa0517124436ef9675708e but is now: 8:dac7716de0de84f4fd508d906d13bcaa`

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/BaseContainerTest.java
```diff
@@ -5,110 +5,64 @@
...
-  @AfterAll
-  static void cleanupOnce() {
-    if (clouddriverContainer != null) {
-      clouddriverContainer.stop();
-    }
-
-    if (redis != null) {
-      redis.stop();
-    }
+            .waitingFor(Wait.forHealthcheck().withStartupTimeout(Duration.ofSeconds(90)));
```
dbyron-sf, line 58: This fails for me locally with a 90 second timeout, but passes with 120.  Mind changing it to:
```suggestion
            .waitingFor(Wait.forHealthcheck().withStartupTimeout(Duration.ofSeconds(120)));
```
I suppose at some point different tests could use different timeouts, but maybe we're not quite at that point yet.

kirangodishala: Done.

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/BaseContainerTest.java
```diff
@@ -5,110 +5,64 @@
...
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.time.Duration;
-import java.util.Map;
-import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.TestInfo;
+import org.junit.jupiter.api.*;
```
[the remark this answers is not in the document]
kirangodishala: Done!

clouddriver-sql/src/main/resources/db/changelog/20190913-task-sagaids.yml
```diff
@@ -4,6 +4,7 @@ databaseChangeLog:
       dbms: postgresql
       remove: afterColumn
   - changeSet:
+      validCheckSum: 8:dac7716de0de84f4fd508d906d13bcaa
```
dbyron-sf, line 7: I removed this line and the validCheckSum from 20181120-cats.yml and PostgresContainerTest still passed.  Is that expected?

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/BaseContainerTest.java
```diff
@@ -5,110 +5,66 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.Network;
-import org.testcontainers.containers.output.Slf4jLogConsumer;
 import org.testcontainers.containers.wait.strategy.Wait;
 import org.testcontainers.junit.jupiter.Testcontainers;
 import org.testcontainers.utility.DockerImageName;
 
 @Testcontainers
```
dbyron-sf, line 38: Not sure whether it's more correct to have this annotation here or in the child classes, but I suspect either one is more correct than having it in both.

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/BaseContainerTest.java
```diff
@@ -5,110 +5,66 @@
...
+  protected static final Network network = Network.newNetwork();
 
-  private static final GenericContainer redis =
-      new GenericContainer(DockerImageName.parse("library/redis:5-alpine"))
-          .withNetwork(network)
-          .withNetworkAliases(REDIS_NETWORK_ALIAS)
-          .withExposedPorts(REDIS_PORT);
-
-  private static GenericContainer clouddriverContainer;
+  protected static GenericContainer clouddriverContainer;
```
dbyron-sf, line 45: Is it really correct to have this be static?  It does seem to work, but I can't tell if we're getting lucky.  Don't we want separate clouddriver containers for each test class?  And probably the network too.

clouddriver-sql/src/main/resources/db/changelog/20190913-task-sagaids.yml
kirangodishala: With liquibase 4.27.0 it passes. If you use 4.24.0, it fails.

dbyron-sf: https://github.com/spinnaker/kork/pull/1184 hasn't been merged yet, so this PR still uses liquibase 4.24.0.

dbyron-sf: When I hand-force the dependency to liquibase 4.27.0, PostgresContainerTest still passes for me with the validChecksum lines removed.

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/BaseContainerTest.java
kirangodishala: The way I separated the configs between base and sub classes will make the tests always work but I see it doesn't seem right to code it this way. I will refactor this.

clouddriver-sql/src/main/resources/db/changelog/20190913-task-sagaids.yml
kirangodishala: My bad. It doesn't matter whether it's 4.24.0 or 4.27.0, the failure for new installation is not due to missing validchecksum (which causes failure to migrating spinnaker) but due to missing `removeChangeSetProperty` tag which was fixed through this commit - https://github.com/spinnaker/clouddriver/commit/db8897c38b5e676574cd6101df6d84514aa84a5d.

dbyron-sf: Aaah ok.  I guess we haven't reached the platonic ideal of having a test that demonstrates the struggle with missing validCheckSum yet.  Testing a migration like that does sound fiddly, but might not even need a standalone clouddriver test like this.  It could be a "regular junit test" in clouddriver-sql.

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/BaseContainerTest.java
kirangodishala: Removed the annotation from base class.

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/BaseContainerTest.java
kirangodishala: Updated all the classes now. The test setup and containers are fully isolated now.

clouddriver-sql/src/main/resources/db/changelog/20190913-task-sagaids.yml
kirangodishala: One way I can think of using test containers for migration testing is to add a script to pre-fill DatabaseChangeLog table so that when the test is run it mimics migration. But the script needs to be updated soon after every liquibase upgrade to keep it up-to date. 

I  am not sure of your proposal on regular junit test.

dbyron-sf: I was thinking of adding an extra migration in clouddriver-sql/src/test/resources, but I'm not sure what it would take to be able to control when it's available.  As in, run a test that uses only the "real" migrations first and then adds the additional ones in a second step.

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/BaseContainerTest.java
```diff
@@ -32,37 +32,36 @@
...
 import org.testcontainers.containers.Network;
 import org.testcontainers.containers.wait.strategy.Wait;
-import org.testcontainers.junit.jupiter.Testcontainers;
 import org.testcontainers.utility.DockerImageName;
 
-@Testcontainers
 public class BaseContainerTest {
 
-  private static final Logger logger = LoggerFactory.getLogger(BaseContainerTest.class);
+  protected final Logger logger = LoggerFactory.getLogger(BaseContainerTest.class);
```
dbyron-sf, line 39: This could still be private/static, yes?

kirangodishala: Yes. changing it.

clouddriver-sql/src/main/resources/db/changelog/20190913-task-sagaids.yml
kirangodishala: But then both of them uses the same liquibase version. It won't be a true test, right?

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/MySqlContainerTest.java
```diff
@@ -58,12 +61,12 @@ static void setup() throws Exception {
         .start();
 
     Slf4jLogConsumer logConsumer = new Slf4jLogConsumer(logger);
-    clouddriverContainer.start();
     clouddriverContainer.followOutput(logConsumer);
+    clouddriverContainer.start();
```
dbyron-sf, line 65: Did you run into a scenario where having followOutput before start makes a difference?

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/BaseContainerTest.java
dbyron-sf: Thanks!

clouddriver-sql/src/main/resources/db/changelog/20190913-task-sagaids.yml
dbyron-sf: OK, that's another dimension then...and I guess a required one.  Pre-filling the database changelog table does seem like the way forward.

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/MySqlContainerTest.java
kirangodishala: No, but it's a configuration and I found all examples on google using this way, so changed the order.

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/BaseContainerTest.java
kirangodishala: Done

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/MySqlContainerTest.java
```diff
@@ -20,36 +20,39 @@
...
 
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
 @Testcontainers
 public class MySqlContainerTest extends BaseContainerTest {
 
-  private static final String MYSQL_NETWORK_ALIAS = "mysqlHost";
+  private static final Logger logger = LoggerFactory.getLogger(MySqlContainerTest.class);
 
-  private static final int MYSQL_PORT = 3306;
+  private final String MYSQL_NETWORK_ALIAS = "mysqlHost";
```
dbyron-sf, line 38: this one and MYSQL_PORT could still be static, yes?

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/PostgresContainerTest.java
```diff
@@ -20,36 +20,39 @@
...
 
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
 @Testcontainers
 public class PostgresContainerTest extends BaseContainerTest {
 
-  private static final String POSTGRES_NETWORK_ALIAS = "postgresHost";
+  private static final Logger logger = LoggerFactory.getLogger(PostgresContainerTest.class);
 
-  private static final int POSTGRES_PORT = 5432;
+  private final String POSTGRES_NETWORK_ALIAS = "postgresHost";
```
dbyron-sf, line 38: static here and for POSTGRES_PORT?

kirangodishala:
Moved integration testes to a  separate PR - https://github.com/spinnaker/clouddriver/pull/6221

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/PostgresMigrationContainerTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+  private String jdbcUrl = "";
+
+  @BeforeAll
+  static void setupInitial() {
+    // using the latest image that is still running on liquibase 3.10.3 to create the conditions
+    // similar to real scenario
+    String fullDockerImageName =
+        "us-docker.pkg.dev/spinnaker-community/docker/clouddriver:5.82.2-dev-release-1.32.x-7a8e6e8b3-202406051721-unvalidated";
+    // Skip the tests if there's no docker image.  This allows gradlew build to work.
```
dbyron-sf, line 64: No need for this comment or the `assumeTrue` anymore.  Truth is, this whole method could disappear and the above could become 
```
private static DockerImageName previousDockerImageName = DockerImageName.parse("us-docker.pkg.dev/spinnaker-community/docker/clouddriver:5.82.2-dev-release-1.32.x-7a8e6e8b3-202406051721-unvalidated");
```
with some javadoc.  Would be nice if the comment was more explicit about why we're picking this image. It's not only that it used liquibase 3.10.3, but also because migrations fail without the additional validCheckSum bits.

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/PostgresMigrationContainerTest.java
```diff
@@ -0,0 +1,145 @@
...
+    postgres.start();
+    jdbcUrl =
+        String.format("jdbc:postgresql://%s:%d/clouddriver", POSTGRES_NETWORK_ALIAS, POSTGRES_PORT);
+
+    // Start the first clouddriver(from previous release) container so that all the db changelog
+    // sets are executed
+    clouddriverInitialContainer =
+        new GenericContainer(previousDockerImageName)
+            .withNetwork(network)
+            .withExposedPorts(7002)
```
dbyron-sf, line 89: change CLOUDDRIVER_PORT to protected in BaseContainerTest and use it here?

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/PostgresMigrationContainerTest.java
kirangodishala: ok, will make these changes

clouddriver-integration/src/test/java/com/netflix/spinnaker/clouddriver/PostgresMigrationContainerTest.java
kirangodishala: will do

dbyron-sf:
@mergifyio update

dbyron-sf (approved):

dbyron-sf:
@mergifyio backport release-1.33.x release-1.34.x

dbyron-sf:
@mergifyio backport release-1.33.x

PR #6225, opened 2024-06-07
fix(Google-api-services-compute-beta): force rev20201102 1.30.10

shlomodaari:
In this issue https://github.com/spinnaker/spinnaker/issues/6958 
You can see that rev20201102 1.30.10 isn't forced successfully. 
The reason is the version com.google.apis:google-api-services-compute:alpha-rev20200526-1.30.9 is leaked from other places. 
I could see the correct version when I forced it at other places in Clouddriver. 

Screenshot 2024-06-07 at 11 02 23 AM
Screenshot 2024-06-07 at 11 05 47 AM
Screenshot 2024-06-07 at 11 09 12 AM
Screenshot 2024-06-07 at 11 12 14 AM
Screenshot 2024-06-07 at 11 13 00 AM
Screenshot 2024-06-07 at 12 02 04 PM


shlomodaari:
It is getting leaked from one of the others, so I changed it all over. @dbyron0 Let me know if you think I should try to change it only in clouddriver-web.

shlomodaari:
It is generating some errors while changing it all over. Let me look into it and confirm

dbyron-sf:
I suspect the place to do this is in kork, but I haven't worked much with the google dependencies.  @j-sandy do you know?

shlomodaari:
@dbyron-sf I didn't do it in Kork because it might affect and break other services. 
If you look at the issue, I shared there the PR they used force in the Clouddriver service only as I believe the reason is not to break other services. 

The problem was that the force didn't work since it got leaked from other parts of Clouddriver. 

In any case. @j-sandy  If you prefer me to create a PR for that in the Kork repo, let me know, and I will do so shortly. :)

shlomodaari:
FYI @dbyron-sf this is the [PR](https://github.com/spinnaker/clouddriver/commit/6e3cd55d7b7df569c2c6bd6dd7b8b0c1e3488c51)

jasonmcintosh:
I'd say bump kork & the fix anything.  Frankly we're on such an old version of this... it kinda scares me.  DEFINITELY need to bump some of these in several areaes, but at least getting to beta in kork IMHO.  I don't know that anything outside of clouddriver SHOULD be using these constraints so I suspect it should be low impact to bump in kork.

jasonmcintosh:
Closing in favor of kork based changes https://github.com/spinnaker/kork/pull/1188

PR #6239, opened 2024-06-25
fix(deps): Fixes gradle deps to use strictly

jasonmcintosh:
See: https://github.com/gradle/gradle/issues/29506#issuecomment-2162486040 - "strictly" is apparently the way to go for transitive deps (aka clouddriver-web) that would tell web I MUST have this version.  The configuration strategy resolution doesn't work for downstream projects.  This WILL Cause breakage until Kork is updated as well OR clouddriver-web does a force beyond what was set.  It's non ideal IMHO but an issue more with gradle and how it's handling transitive dependencies & force overrides.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,7 +52,10 @@ subprojects {
 
     sourceSets.main.java.srcDirs = []
     sourceSets.main.groovy.srcDirs += ["src/main/java"]
-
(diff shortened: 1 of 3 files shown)


jasonmcintosh:
Probably going to close this PR in favor of another removing the dep constraint entirely in favor of kork... we'll watch for transitive dep overrides like this in the future.

PR #6254, opened 2024-07-25
feat(kubernetes): Skip applying labels to spec.template.metadata if skipSpecTemplateLabels is true

kunalsharma05:
Moved the OperationDescription from clouddriver-api to kork-moniker and updated all the relevant references. This will allow us to pass Manifest Description to the Namer class.

Setting skipSpecTemplateLabels to true, clouddriver will not apply the app.kubernetes.io/* labels to the manifests' spec.template.metadata.labels.

Kork PR that needs to be included prior to this: https://github.com/spinnaker/kork/pull/1199


dbyron-sf:
@mergifyio update

dbyron-sf (approved):

PR #6255, opened 2024-07-31
perf(cache): Optimise heap usage in KubernetesCachingAgent

kbatalin:
This pull request addresses a performance issue discovered in the Clouddriver, specifically within the `KubernetesServiceHandler.addAllReplicaSetLabels` method.

With allocation profiling I found that 40% of allocations happen for converting `replicaSet` object to `KubernetesManifest` class. The thing is that this conversion is redundant because `replicaSet` is already `KubernetesManifest`.

The objects aren't changed later, so it should be safe to stop parsing and making a copy. After this optimisation we reduced memory usage by half and also got rid of regular GC pauses.

Screenshots

Allocation profile:
image

Memory usage:
image

GC pauses:
image

Memory allocation rate:
image

GC promotion rate:
image


kbatalin:
Link a relevant issue: https://github.com/spinnaker/spinnaker/issues/6835

PR #6256, opened 2024-08-06
perf(ecs): Narrowing the cache search for the ECS provider on views

christosarvanitis:
Attempt to address some of the issues described in https://github.com/spinnaker/spinnaker/issues/6084

Improving the response times on:
- /clusters
- /applications
- /serverGroups
Endpoints when ECS is enabled and a substantial amount of accounts/services exist in cache.

The perf issue with the Alarms still exists and will be addressed in a future PR

Adding some results from a performance test clouddriver response times:
image
image
image

- GET {CLOUDDRIVER_URL}/applications
    - Average: 104ms → 92.1ms (11% improvement)
    - 95th Percentile: 130ms → 126ms
- GET {CLOUDDRIVER_URL}/applications/{application_name}
  - Average: 7.48s → 4.2s (43% improvement)
  - 95th Percentile: 8.55s → 5.86s
- GET {CLOUDDRIVER_URL}/applications/{application_name}/serverGroups
  - Average: 2.72s → 2.16s (20% improvement)
  - 95th Percentile: 3.17s → 3.11s
- GET {CLOUDDRIVER_URL}/applications/{application_name}/clusters
  - Average: 107ms → 43.3ms (59% improvement)
  - 95th Percentile: 135ms → 88.4ms

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/AbstractCacheClient.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/AbstractCacheClient.java
@@ -65,6 +65,11 @@ abstract class AbstractCacheClient<T> {
     return convertAll(data);
   }
 
+  public Collection<T> getAll(Collection<String> identifiers) {
+    Collection<CacheData> allData = cacheView.getAll(keyNamespace, identifiers);
+    return convertAll(allData);
+  }
+
   /**
    * @param key A key within the key namespace that will be used to retrieve the object.
    * @return An object of the generic type that is associated to the key.
@@ -109,5 +114,9 @@ abstract class AbstractCacheClient<T> {
     }
 
     return allData;
+  }
+
+  public Collection<String> filterIdentifiers(String glob) {
+    return cacheView.filterIdentifiers(keyNamespace, glob);
   }
 }
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/AbstractCacheClient.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/AbstractCacheClient.java
@@ -67,6 +67,9 @@ abstract class AbstractCacheClient<T> {
 
   public Collection<T> getAll(Collection<String> identifiers) {
     Collection<CacheData> allData = cacheView.getAll(keyNamespace, identifiers);
+    if (allData == null) {
+      return Collections.emptyList();
+    }
     return convertAll(allData);
   }
 
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/EcsCloudWatchAlarmCacheClient.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/EcsCloudWatchAlarmCacheClient.java
@@ -73,6 +73,8 @@ public class EcsCloudWatchAlarmCacheClient extends AbstractCacheClient<EcsMetric
   public List<EcsMetricAlarm> getMetricAlarms(
       String serviceName, String accountName, String region) {
     List<EcsMetricAlarm> metricAlarms = new LinkedList<>();
+    // we can filter more here.
+
     Collection<EcsMetricAlarm> allMetricAlarms = getAll(accountName, region);
 
     outLoop:
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/EcsCloudWatchAlarmCacheClient.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/EcsCloudWatchAlarmCacheClient.java
@@ -20,6 +20,7 @@ import static com.netflix.spinnaker.clouddriver.ecs.cache.Keys.Namespace.ALARMS;
 
 import com.netflix.spinnaker.cats.cache.Cache;
 import com.netflix.spinnaker.cats.cache.CacheData;
+import com.netflix.spinnaker.clouddriver.ecs.cache.Keys;
 import com.netflix.spinnaker.clouddriver.ecs.cache.model.EcsMetricAlarm;
 import java.util.Collection;
 import java.util.Collections;
@@ -71,33 +72,19 @@ public class EcsCloudWatchAlarmCacheClient extends AbstractCacheClient<EcsMetric
   }
 
   public List<EcsMetricAlarm> getMetricAlarms(
-      String serviceName, String accountName, String region) {
+      String serviceName, String accountName, String region, String ecsClusterName) {
     List<EcsMetricAlarm> metricAlarms = new LinkedList<>();
-    // we can filter more here.
 
-    Collection<EcsMetricAlarm> allMetricAlarms = getAll(accountName, region);
+    String glob = Keys.getAlarmKey(accountName, region, "*", ecsClusterName);
+    Collection<String> metricAlarmsIds = filterIdentifiers(glob);
+    Collection<EcsMetricAlarm> allMetricAlarms = getAll(metricAlarmsIds);
 
-    outLoop:
     for (EcsMetricAlarm metricAlarm : allMetricAlarms) {
-      for (String action : metricAlarm.getAlarmActions()) {
-        if (action.contains(serviceName)) {
-          metricAlarms.add(metricAlarm);
-          continue outLoop;
-        }
-      }
-
-      for (String action : metricAlarm.getOKActions()) {
-        if (action.contains(serviceName)) {
-          metricAlarms.add(metricAlarm);
-          continue outLoop;
-        }
-      }
-
-      for (String action : metricAlarm.getInsufficientDataActions()) {
-        if (action.contains(serviceName)) {
-          metricAlarms.add(metricAlarm);
-          continue outLoop;
-        }
+      if (metricAlarm.getAlarmActions().stream().anyMatch(action -> action.contains(serviceName))
+          || metricAlarm.getOKActions().stream().anyMatch(action -> action.contains(serviceName))
+          || metricAlarm.getInsufficientDataActions().stream()
+              .anyMatch(action -> action.contains(serviceName))) {
+        metricAlarms.add(metricAlarm);
       }
     }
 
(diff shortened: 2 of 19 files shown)


christosarvanitis:
@dbyron0 @deverton would appreciate your feedback on this change. There are still improvements to be made as the current implementation of ECS goes through every region per account to retrieve the necessary data from cache which is far from ideal when there are hundreds of accounts.
The main idea here is to limit the retrieval with an application name when we can.

The perf of alarms is still a problem as right now it goes through all the alarms and tries to match with a service but this will be addressed in a future PR.

christosarvanitis:
@dbyron-sf @jasonmcintosh Added some results from an internal testing related to this change. Would appreciate any feedback!

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/AbstractCacheClient.java
```diff
@@ -65,6 +65,11 @@ public Collection<T> getAll(String account, String region) {
     return convertAll(data);
   }
 
+  public Collection<T> getAll(Collection<String> identifiers) {
+    Collection<CacheData> allData = cacheView.getAll(keyNamespace, identifiers);
```
jasonmcintosh, line 69: NIT: merge the variable dont' need it.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsServerClusterProvider.java
```diff
@@ -480,6 +485,15 @@ public Map<String, Set<EcsServerCluster>> getClusters() {
     return clusterMap;
   }
 
+  public Map<String, Set<EcsServerCluster>> getClusters0(String application) {
```
jasonmcintosh, line 488: getClusters0 seems like a bad function name.  I'm... wondering why this method vs. the getClusterSummaries up above.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/view/EcsApplicationProvider.java
```diff
@@ -50,31 +51,32 @@ public EcsApplicationProvider(
 
   @Override
   public Application getApplication(String name) {
-
-    for (Application application : getApplications(true)) {
+    name = name.toLowerCase();
```
jasonmcintosh, line 54: This can have an impact below on casing, since it's lower case & application (though app should be lower case... just a caution of COULD be a bug someplace here).  IF going to lower case, should do name equals ignore case or similar.  KINDA concerned about the filtering as well on casing but not dug into that.

jasonmcintosh:
Few minor things but overall looks good.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsServerClusterProvider.java
christosarvanitis: Im following the same pattern/naming convention established here for the AWS provider https://github.com/spinnaker/clouddriver/blob/0226100a95b16e1176a296198b1f892c4514e4d6/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonClusterProvider.groovy#L70:L78 

But for the ECS provider the Cluster view never implemented the clusterSummaries. It was always returning the clusterDetails. 

Which is another area of improvement. getClusterSummaries should return only the names of the serverGroups and loadbalancers. getClusterDetails should return the full details of the serverGroups and loadbalancers. 

Will actually address this in this PR which makes sense to improve here.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/view/EcsApplicationProvider.java
christosarvanitis: Yeah it shouldnt be a problem. It is always coming as lower case as the caching agents are inserting the keys in lowercase. Similar to here in the AWS provider https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonApplicationProvider.groovy#L52

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsServerClusterProvider.java
jasonmcintosh: Key differences:
* getClusters0 in AWS is a parameterized private method which takes an argument to control whether details are added. In this case, there's no difference between the getDetails & getSummaries calls (which is a question - WHY there isn't a difference, but not looked at the data returned).  
* As such, it's extra code that doesn't help code deduplication.  AN alternative is making getClusters0 private but I'd prefer merge the logic and just have one call the other vs. a separate method that's private that both call increasing the call stack.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/view/EcsApplicationProvider.java
jasonmcintosh: Maybe both are setting it, but the caution flag is:  Don't trust that, as I've seen cases where SOMEHOW the data internally gets into mixed cases depending on the location.  I'd rather avoid any POTENTIAL bug in the future here.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/AbstractCacheClient.java
christosarvanitis: Done!

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/view/EcsApplicationProvider.java
christosarvanitis: Reverting the change to not forcing a lowercase as the search previously was. I see what you are concerned about - we dont know how the ECS clusters/services are deployed (lower/uppercase etc) and how the application/moniker are configured. 
Lets keep the logic the same which so far didnt create any issues.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/view/EcsServerClusterProvider.java
christosarvanitis: Sounds good! No objection in implementing the getDetails and getSummaries better :) Just pushed the changes + i verified that the API responses before and after the change are the same.

With the difference that previously getClusterSummaries was returning All the ECS clusters/services for all the accounts (no application filtered). After the change it will return only the application clusters/services. 

getSummaries is called via the applicationController -> GET /applications/<applicationName>

getDetails is called via the serverGroup controller -> GET /applications/<applicationName>/serverGroups

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/EcsCloudWatchAlarmCacheClient.java
```diff
@@ -71,33 +72,19 @@ protected EcsMetricAlarm convert(CacheData cacheData) {
   }
 
   public List<EcsMetricAlarm> getMetricAlarms(
-      String serviceName, String accountName, String region) {
+      String serviceName, String accountName, String region, String ecsClusterName) {
     List<EcsMetricAlarm> metricAlarms = new LinkedList<>();
-    // we can filter more here.
 
-    Collection<EcsMetricAlarm> allMetricAlarms = getAll(accountName, region);
```
christosarvanitis, line 76: Before All the alarms for an ECS account/region where fetched and iterated through to match the service. This is extremely costly. 

After the change the ECSCluster is added during the caching cycles to the cache key id for the ECS provider in the alarms. We retrieve the IDs with ECS account/region/EcsClusterName and then try to match the service.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/EcsCloudWatchAlarmCacheClient.java
```diff
@@ -71,33 +72,19 @@ protected EcsMetricAlarm convert(CacheData cacheData) {
...
-          continue outLoop;
-        }
-      }
-
-      for (String action : metricAlarm.getInsufficientDataActions()) {
-        if (action.contains(serviceName)) {
-          metricAlarms.add(metricAlarm);
-          continue outLoop;
-        }
+      if (metricAlarm.getAlarmActions().stream().anyMatch(action -> action.contains(serviceName))
```
christosarvanitis, line 83: Small refactoring here to make it more readable

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/EcsCloudMetricAlarmCachingAgent.java
```diff
@@ -118,7 +118,13 @@ Map<String, Collection<CacheData>> generateFreshData(Set<MetricAlarm> cacheableM
     Map<String, Collection<CacheData>> newDataMap = new HashMap<>();
 
     for (MetricAlarm metricAlarm : cacheableMetricAlarm) {
-      String key = Keys.getAlarmKey(accountName, region, metricAlarm.getAlarmArn());
+      String cluster =
+          metricAlarm.getDimensions().stream()
```
christosarvanitis, line 122: Based on the AWS SDK a cloudwatch alarm for the ECS contains 2 dimensions depending for the type:
- Service alarm contains the dimension ECSCluster and ServiceName
- Autoscaling group alarm of an ECS cluster contains the ECSCluster and the Capacity provider.

This change includes the ECSClusterName in the cached key id to make the search less costly

christosarvanitis:
Following up with the Alarm changes the perf improvements:
image

The GET ServerGroups call reduced from 7+ secs on an application to 3.5secs.

And timings on a single request before the change:
image
And after the change:
image

jasonmcintosh:
Overall I think this looks good ;) Would like one more set of eyes given the changes on the way ECS operates.  One concern is how the change on the cache id's will be cleaned up since this changes the storage ids - but that MAY get taken care of by one of the cleanup jobs (need to confirm).  Talked in slack - the ID's should still fit the max column length so adding that shouldn't impact any database stuff.

jasonmcintosh (approved):

jasonmcintosh:
OK will merge shortly.  I'd LIKE to get a release notes updated for this please!

PR #6268, opened 2024-08-17
fix(gha): only bump halyard on master

dbyron-sf:
Since halyard uses a different versioning scheme, it doesn't make sense for halyard to consume non-master (e.g. release-1.34.x) versions of clouddriver.


dbyron-sf:
@mergifyio backport release-1.34.x release-1.35.x

PR #6276, opened 2024-08-20
fix(install): Fix Debian post installation script when awscli2 is already installed or the installation failed previously

ashleykleynhans:
More fixes to the post installation script for the Debian package.

* Only installs the specified version of awscliv2 if it is not already installed.
* Cleans up the `./aws` before unzipping `awscliv2.zip` if the installation failed previously and the `./aws` directory already exists.
* Use curl `-s` for downloads for cleaner output.


dbyron-sf (approved):

dbyron-sf:
@mergifyio backport release-1.35.x

PR #6279, opened 2024-08-26
fix(aws): Fix AWS CLI v2 for Alpine Linux

jervi:
The version of the AWS CLI v2 that is currently installed in the `slim` docker images (Alpine based) does not work. Trying to execute the aws binary results in a `bash: /usr/local/bin/aws: No such file or directory` error, even though the binary _is_ present. The error occurs because Alpine Linux doesn't support `glibc`. To fix this we need to install a version of AWS CLI v2 that is using `musl` instead of `glibc`.
Starting from Alpine 3.18, compatible aws-cli binaries are available in the [official package repos](https://pkgs.org/download/aws-cli), meaning we can just `apt add` it.
I've bumped the version to the latest available for Alpine (2.15.57) and also bumped Alpine itself from 3.16 to 3.20 (since aws-cli v2 wasn't available in 3.16, and only 3.20 supports the newest version of aws-cli).


dbyron-sf:
@jervi would you mind updating the commit message + PR description a bit?  Moving from `3.8-alpine3.16` to `3.8-alpine3.20` seems at least worth mentioning.  Could you also explain what bug this fixes, and why?

jervi:
> Could you also explain what bug this fixes, and why?
I've updated the PR description

dbyron-sf:
Thank you!

dbyron-sf (approved):

dbyron-sf:
PS This fixes https://github.com/spinnaker/clouddriver/pull/6156.

dbyron-sf:
It's a little bold backporting this when it bumps alpine from 3.16 to 3.20, but since the aws cli is busted in alpine docker images in 1.35, I'ma do it.

@mergifyio backport release-1.35.x

dbyron-sf:
@mergifyio backport release-1.35.x

dbyron-sf:
Included in [Spinnaker 1.35.4](https://spinnaker.io/changelogs/1.35.4-changelog/).

PR #6282, opened 2024-09-19
feat(gce): support resourceManagerTags in instance template

edgarulg:
Support `resourceManagerTags` in GCE. See: https://cloud.google.com/compute/docs/reference/rest/v1/instanceTemplates/insert

Adds `resourceManagerTags` in the description operation and add it in the instanceTemplateProperties.

Issue related: https://github.com/spinnaker/spinnaker/issues/6931


christosarvanitis:
Are there any meaningful tests that can be added for this?

clouddriver-google/clouddriver-google.gradle
```diff
@@ -10,7 +10,7 @@ dependencies {
   implementation "org.codehaus.groovy:groovy"
   implementation "org.codehaus.groovy:groovy-json"
   implementation "org.apache.commons:commons-lang3"
-  implementation "com.google.apis:google-api-services-compute"
+  implementation "com.google.apis:google-api-services-compute:beta-rev20210525-1.31.5"
```
christosarvanitis, line 13: Do we need the pinned version here since it is part of kork?

edgarulg:
> Are there any meaningful tests that can be added for this?

I am still working on adding test for it but it is challenging because the BasicGoogleDeployHandler class is really hard to test :(

clouddriver-google/clouddriver-google.gradle
```diff
@@ -47,8 +47,15 @@ dependencies {
...
   testImplementation "org.springframework.boot:spring-boot-test"
   testImplementation "org.codehaus.groovy:groovy-test"
+  testImplementation 'org.spockframework:spock-core:2.0-groovy-3.0'
+  testImplementation 'org.powermock:powermock-api-mockito2:2.0.9'
+  testImplementation 'org.powermock:powermock-module-junit4:2.0.9'
 }
 
 configurations.all {
-   resolutionStrategy.force 'com.google.apis:google-api-services-compute:beta-rev20201102-1.30.10'
+   resolutionStrategy {
```
jasonmcintosh, line 56: This doesn't work the way most think it does.  Particularly with the kork PR, these resolution strategy blocks need to be removed.

edgarulg: I can get rid of those but we need to merge and release https://github.com/spinnaker/kork/pull/1203 first to ensure the correct version is installed.

edgarulg:
@jasonmcintosh @dbyron-sf @christosarvanitis I tried to add a test for the resourceManagerTags field I added to GCE but I am not able to test the `BasicGoogleDeployHandler` because the class has design problems and we cannot mock properly. The existing tests for that class are [disabled](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandlerSpec.groovy#L58).
We need to refactor the class to easily test functionality but I don't think is a good idea to refactor the code as part of this change.
I also changed the JacksonFactory to GsonFactory as part of the fixes of upgrading google-api libraries.
I prefer to raise a new issue to refactor the `BasicGoogleDeployHandler`and add the appropiate testing for it.
Any thoughts?

dbyron-sf:
@edgarulg I'm all for the refactor happening in a separate PR, especially if it can happen before this one.

bnewtonius:
> @edgarulg I'm all for the refactor happening in a separate PR, especially if it can happen before this one.

Hi @dbyron-sf. Normally I would be in full-throated agreement. This PR is a minimal change (adding one new field). Refactoring here would introduce more risk and have a greater chance of creating regressions.

We are happy to take a crack at refactoring this to make it more testable soon after.

edgarulg:
I am closing this PR because a lot of changes were added from master and it isn't clear what changes belongs to the `resource-manager-tags`. Please use https://github.com/spinnaker/clouddriver/pull/6287 instead.

PR #6287, opened 2024-09-27
feat(gce): support resource-manager-tags in GCE deployment

edgarulg:
Support resourceManagerTags in GCE. See: https://cloud.google.com/compute/docs/reference/rest/v1/instanceTemplates/insert

Adds resourceManagerTags in the description operation and add it in the instanceTemplateProperties.

Issue related: https://github.com/spinnaker/spinnaker/issues/6931

this PR replaced the [previous PR](https://github.com/spinnaker/clouddriver/pull/6282) in clouddriver because I want to leave only my changes related to `resource-manager-tags`

### DIFF
--- a/clouddriver-google/clouddriver-google.gradle
+++ b/clouddriver-google/clouddriver-google.gradle
@@ -48,7 +48,3 @@ dependencies {
   testImplementation "org.springframework.boot:spring-boot-test"
   testImplementation "org.apache.groovy:groovy-test"
 }
-
-configurations.all {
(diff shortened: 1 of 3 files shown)


edgarulg:
I was not able to add tests as part of my changes because there is not an easy way to mock the BasicGoogleDeployHandler. I think we need to add testing to this cloud provider so I created the following issue to refactor the code: https://github.com/spinnaker/spinnaker/issues/6985
As I'm going to work in more new features of GCE in the next days I can work on the refactor as well but I will work on the refactor after this small change.

link108 (approved):

PR #6288, opened 2024-10-02
feat(gce): add support for hyperdisk in GCE

edgarulg:
We add `hyperdisk-balanced` as an option for GCE deployment in Spinnaker. See: https://cloud.google.com/compute/docs/disks/hyperdisks

GoogleDiskType has the available disk types that Spinnaker can deploy. By adding `hyperdisk-balanced` the instance templates in GCE can now use hyperdisk.


clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleDiskSpec.groovy
```diff
@@ -0,0 +1,63 @@
+/*
```
dbyron-sf, line 1: Please avoid adding new groovy code.  Mind rewriting this in java?

clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleDiskSpec.groovy
```diff
@@ -0,0 +1,63 @@
...
+ * Copyright 2024 Netflix, Inc.
```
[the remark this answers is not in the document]
edgarulg: it was auto-generated. Let me change it.

clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleDiskSpec.groovy
edgarulg: Sure, let me change it to Java.

edgarulg: done

dbyron-sf:
LGTM.  I'd love someone who knows more about the google provider to have a look.

ritek01 (approved):

christosarvanitis:
Looks good, there is a gotcha with the hyperdisks that only specific instance families are supported but i guess a user requesting hyperdisk is aware of that.

christosarvanitis (approved):

edgarulg:
> Looks good, there is a gotcha with the hyperdisks that only specific instance families are supported but i guess a user requesting hyperdisk is aware of that.

That is right, Hyperdisk is only supported in specific instance families. The right place to add a validation would be Deck but that is a nice-to-have thing because we assumed the user knows that hyperdisk is for critical instances. Even in the default instances provided by Spinnaker in GCE it doesn't have C3, C4 families to add validations.

The users that will use hyperdisk is not configuring their instances using deck-ui.

jasonmcintosh (approved):

PR #6290, opened 2024-10-10
refactor(gce): Refactor BasicGoogleDeployHandler to be more readable and testable

edgarulg:
Key point about the refactor.

1. I translated the whole class BasicGoogleDeployHandler into java.
2. I split up the handle method into small single responsibility methods.
3. I defined unit testing for most of the new methods I created to prove functionality.
4. I do not include any new feature or change that was not defined by the BasicGoogleDeployHandler before.

Related to: https://github.com/spinnaker/spinnaker/issues/6985


edgarulg:
I published and deployed a snapshot version with my refactor and I validated a GCE deploy worked as expected. I didn't tests all possible configurations as part of my e2e validation.

edgarulg:
@Mergifyio refresh

edgarulg:
@Mergifyio rebase

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy
```diff
@@ -920,7 +920,7 @@ class GCEUtil {
       : []
   }
 
-  static ServiceAccount buildScheduling(BaseGoogleInstanceDescription description) {
+  static Scheduling buildScheduling(BaseGoogleInstanceDescription description) {
```
dbyron-sf, line 923: How did this ever work before?

edgarulg: In line: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy#L924 
It creates a Scheduling object in runtime. Groovy is flexible enough to allow a different object in runtime and it only pass the object to the InstanceProperties object: 
https://github.com/spinnaker/clouddriver/blob/master/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy#L390
https://github.com/spinnaker/clouddriver/blob/master/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy#L428

that is how this works but I don't like it. It's really hard to follow what is going on with groovy. That is why I moved it to java.

dbyron-sf: OK.  I suppose in some future we'll change GCEUtil to java too.

edgarulg: yes, that is the plan

clouddriver-google/src/main/java/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.java
```diff
@@ -0,0 +1,1391 @@
...
+      BasicGoogleDeployDescription description, InstanceGroupManager instanceGroupManager) {
+    if (description.getDistributionPolicy() != null) {
+      DistributionPolicy distributionPolicy = new DistributionPolicy();
+
+      if (description.getSelectZones()
+          && !description.getDistributionPolicy().getZones().isEmpty()) {
+        log.info(
+            String.format(
+                "Configuring explicit zones selected for regional server group: %s",
+                description.getDistributionPolicy().getZones().get(0)));
```
jasonmcintosh, line 986: get 0 - just the one?  NOT checked old code... but we're streaming below which may mean multiple zones...

edgarulg: In the old code it only logs the first element but you are right. It should log all the selected regions because they are use in the distribution policy.

mattgogerly (approved):

PR #6296, opened 2024-10-17
chore(azure): Adding a verifyAccountHealth configuration

christosarvanitis:
Adding a verifyAccountHealth configuration parameter for Azure provider (similar to what it exists for other providers) which skips the check for valid credentials of each account allowing Clouddriver to start and/or not fail when a bad account exists in the configuration 

```
azure:
  enabled: true
  health:
    verifyAccountHealth: true|false #defaults to true
  accounts:  []
```


jasonmcintosh (approved):
Thanks!!

edgarulg (approved):

PR #6307, opened 2024-11-21
fix(ecs): evaluate targetgroup healthcheck in instance healthcheck

edgarulg:
Deployments using ECS are marked successful by Spinnaker but the ECS service is unhealthy when target group mappings are defined in the configuration.

Here is pipeline execution with the successful deployment in Spinnaker:
Captura de pantalla 2024-11-21 a la(s) 1 30 51 p m
Captura de pantalla 2024-11-21 a la(s) 1 31 35 p m

But In the server group events we can see unhealthy issues.
Captura de pantalla 2024-11-21 a la(s) 1 31 58 p m

Here is the issue in ECS AWS console.
Captura de pantalla 2024-11-21 a la(s) 1 32 52 p m

As part of my changes I search for targetGrouphealths and then use them to set unhealthy statuses to the instance.

### DIFF
--- a/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/services/ContainerInformationService.java
+++ b/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/services/ContainerInformationService.java
@@ -18,16 +18,19 @@ package com.netflix.spinnaker.clouddriver.ecs.services;
 
 import com.amazonaws.services.ec2.model.Instance;
 import com.amazonaws.services.ecs.model.ContainerDefinition;
+import com.amazonaws.services.ecs.model.LoadBalancer;
 import com.amazonaws.services.ecs.model.NetworkBinding;
 import com.amazonaws.services.ecs.model.TaskDefinition;
 import com.netflix.spinnaker.clouddriver.ecs.cache.Keys;
 import com.netflix.spinnaker.clouddriver.ecs.cache.client.ContainerInstanceCacheClient;
 import com.netflix.spinnaker.clouddriver.ecs.cache.client.EcsInstanceCacheClient;
 import com.netflix.spinnaker.clouddriver.ecs.cache.client.ServiceCacheClient;
+import com.netflix.spinnaker.clouddriver.ecs.cache.client.TargetHealthCacheClient;
 import com.netflix.spinnaker.clouddriver.ecs.cache.client.TaskCacheClient;
 import com.netflix.spinnaker.clouddriver.ecs.cache.client.TaskDefinitionCacheClient;
 import com.netflix.spinnaker.clouddriver.ecs.cache.client.TaskHealthCacheClient;
 import com.netflix.spinnaker.clouddriver.ecs.cache.model.ContainerInstance;
+import com.netflix.spinnaker.clouddriver.ecs.cache.model.EcsTargetHealth;
 import com.netflix.spinnaker.clouddriver.ecs.cache.model.Service;
 import com.netflix.spinnaker.clouddriver.ecs.cache.model.Task;
 import com.netflix.spinnaker.clouddriver.ecs.cache.model.TaskHealth;
@@ -37,6 +40,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
@@ -50,6 +54,7 @@ public class ContainerInformationService {
   private final TaskDefinitionCacheClient taskDefinitionCacheClient;
   private final EcsInstanceCacheClient ecsInstanceCacheClient;
   private final ContainerInstanceCacheClient containerInstanceCacheClient;
+  private final TargetHealthCacheClient targetHealthCacheClient;
 
   @Autowired
   public ContainerInformationService(
@@ -59,7 +64,8 @@ public class ContainerInformationService {
       TaskHealthCacheClient taskHealthCacheClient,
       TaskDefinitionCacheClient taskDefinitionCacheClient,
       EcsInstanceCacheClient ecsInstanceCacheClient,
-      ContainerInstanceCacheClient containerInstanceCacheClient) {
+      ContainerInstanceCacheClient containerInstanceCacheClient,
+      TargetHealthCacheClient targetHealthCacheClient) {
     this.ecsCredentialsConfig = ecsCredentialsConfig;
     this.taskCacheClient = taskCacheClient;
     this.serviceCacheClient = serviceCacheClient;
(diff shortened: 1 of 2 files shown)


clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/services/ContainerInformationService.java
```diff
@@ -101,17 +108,30 @@ public List<Map<String, Object>> getHealthStatus(
     // Task-based health
     if (task != null) {
       boolean hasHealthCheck = false;
+      EcsTargetHealth targetHealth = null;
       if (service != null) {
         hasHealthCheck = taskHasHealthCheck(service, accountName, region);
+        LoadBalancer loadBalancer = service.getLoadBalancers().stream().findFirst().orElse(null);
+        String targetGroupKey =
+            Keys.getTargetHealthKey(accountName, region, loadBalancer.getTargetGroupArn());
```
jasonmcintosh, line 116: loadBalancer can be null, which means getting the target group ARN could through an NPE here.  IF an ECS isn't associated with an LB...

clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/services/ContainerInformationServiceSpec.groovy
```diff
@@ -247,6 +252,199 @@ class ContainerInformationServiceSpec extends Specification {
...
+    def serviceName = 'test-service-name'
+    def type = 'loadBalancer'
+
+    def cachedService = new Service(
+      serviceName: serviceName,
+      loadBalancers: [new LoadBalancer()]
+    )
+
+    serviceCacheClient.get(_) >> cachedService
+    taskCacheClient.get(_) >> new Task(lastStatus: "RUNNING", healthStatus: "UNKNOWN")
```
jasonmcintosh, line 420: This doesn't match the test description where heatlh status is "UNKNOWN" ... and if health status is unknown... that seems problematic to mark the server as "HEALTHY"...

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/services/ContainerInformationService.java
```diff
@@ -137,6 +159,22 @@ public boolean taskHasHealthCheck(Service service, String accountName, String re
...
+            .distinct()
+            .collect(Collectors.toList());
+
+    if (statuses.stream().anyMatch(it -> it.equalsIgnoreCase("unhealthy"))) {
+      return "UNHEALTHY";
+    }
+    if (statuses.stream().anyMatch(it -> it.equalsIgnoreCase("healthy"))) {
+      return "HEALTHY";
+    }
+    return "UNKNOWN";
```
jasonmcintosh, line 175: SO there's a few other states... and the return types here don' seem to match the toPlaformHealthState checsk that do the conversions.  E.g. "HEALTHY" on this needs to match platform state of... "RUNNING" I think to get an "Up" status.  SO not sure this would actually work?

clouddriver-ecs/src/test/groovy/com/netflix/spinnaker/clouddriver/ecs/services/ContainerInformationServiceSpec.groovy
edgarulg: By default, if the containerInformation service lacks health check details in the containerDefinition, it considers the deployment healthy as long as the lastStatus is set to RUNNING.

In this case, health checks are being reported by the targetHealthcheck, while the containerDefinition does not include any health check information.

that is why the healtStatus is UNKNOWN but it should look for targetHealthchecks if the containerDefinition does not include healthcheck

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/services/ContainerInformationService.java
edgarulg: This is odd to me as well, I refactor the code to be more cleaner. Basically the targetHealthchecks might have multiple statuses but for this check we only care about "unhealthy" So I refactored my code to only return Down status when unhealthy is found in targetHealthchecks.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/services/ContainerInformationService.java
```diff
@@ -138,13 +153,27 @@ public boolean taskHasHealthCheck(Service service, String accountName, String re
...
       return "Down";
     }
 
+    if (ecsTargetHealth != null) {
+      List<String> statuses =
+          ecsTargetHealth.getTargetHealthDescriptions().stream()
+              .map(tg -> tg.getTargetHealth().getState())
+              .distinct()
+              .collect(Collectors.toList());
+      if (statuses.stream().anyMatch(it -> it.equalsIgnoreCase("unhealthy"))) {
```
jasonmcintosh, line 172: OK so the downside with this still, we're ONLY checking "unhealthy", and if that's NOT returned you fall back onto the ECS task status checks.  This leads to a situation where you could have:
* ECS Task Check:  HEALTHY
* ECS LB CHeck: PENDING/UNKNOWN - as it has NOT passed LB health checks which may be different.

This would put it back into service incorrectly in this case.  As the LB health check if set should trump the ECS task service health checks when configured.

clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/services/ContainerInformationService.java
```diff
@@ -138,13 +153,20 @@ public boolean taskHasHealthCheck(Service service, String accountName, String re
   }
 
   private String toPlatformHealthState(
```
jasonmcintosh, line 155: Might not be a bad idea to "document" this:
* IF ecs TASK health fails, we fail regardless of other checks
* Then if it's got LB cehcks, return the state of those checks.
* Finally fallback to ECS task health for final confirmation
With the ONLY question being if the user in a pipeline says "ignore LB health checks" which I can't recall the path to get to...

jasonmcintosh: NOT a requirement to document this ;) just a "It'd be nice to make sure the logic is TRULY expressed" or more verbose.  BUT I think this overall covers things! MINUS that "how do you disable LB Health checks" call..

jasonmcintosh (approved):
OK i think this finally should solve the issue of ECS tasks NOT checking load balancer health to verify task health.

edgarulg:
I validated latest changes with the application config we can set for instance health. This changes does not take priority if the config is disabled.

shlomodaari:
@jasonmcintosh can we get this backport to 1.36?

jasonmcintosh:
@mergifyio backport release-1.36.x

PR #6333, opened 2025-01-17
fix(aws): CleanupAlarmsAgent cycle to catch exceptions

christosarvanitis:
A bad account or any exception currently stops the CleanupAlarmsAgent. 

Catching the exception and logging it since the purpose of this agent is to cleanup the stale cloudwatch alarms from the aws/ecs accounts in Spinnaker and not to verify if the accounts are valid or not.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleDiskType.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleDiskType.groovy
@@ -18,6 +18,8 @@ package com.netflix.spinnaker.clouddriver.google.model
 
 public enum GoogleDiskType {
   PD_STANDARD("pd-standard", true),
+  PD_EXTREME("pd-extreme", true),    
+  PD_BALANCED("pd-balanced", true),    
   PD_SSD("pd-ssd", true),
   LOCAL_SSD("local-ssd", false),
   HYPERDISK_BALANCED("hyperdisk-balanced", true)
--- a/clouddriver-google/src/main/java/com/netflix/spinnaker/clouddriver/google/compute/InstanceTemplates.java
+++ b/clouddriver-google/src/main/java/com/netflix/spinnaker/clouddriver/google/compute/InstanceTemplates.java
@@ -29,6 +29,8 @@ public class InstanceTemplates {
   private final Compute.InstanceTemplates computeApi;
   private final GoogleNamedAccountCredentials credentials;
   private final GlobalGoogleComputeRequestFactory requestFactory;
+  private static final String defaultView =
+      "FULL"; // https://cloud.google.com/sdk/gcloud/reference/beta/compute/instance-templates/list
 
   InstanceTemplates(
       GoogleNamedAccountCredentials credentials,
@@ -65,7 +67,11 @@ public class InstanceTemplates {
     return new PaginatedComputeRequestImpl<>(
         pageToken ->
             requestFactory.wrapRequest(
-                computeApi.list(credentials.getProject()).setPageToken(pageToken), "list"),
+                computeApi
+                    .list(credentials.getProject())
+                    .setPageToken(pageToken)
+                    .setView(defaultView),
+                "list"),
         InstanceTemplateList::getNextPageToken,
         InstanceTemplateList::getItems);
   }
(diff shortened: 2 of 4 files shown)


clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/agent/CleanupAlarmsAgent.groovy
```diff
@@ -37,7 +37,7 @@ import java.util.regex.Pattern
 
 @Slf4j
 class CleanupAlarmsAgent implements RunnableAgent, CustomScheduledAgent {
-  public static final long POLL_INTERVAL_MILLIS = TimeUnit.HOURS.toMillis(24)
+  public static final long POLL_INTERVAL_MILLIS = TimeUnit.MINUTES.toMillis(3)
```
dbyron-sf, line 40: Why this change?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/agent/CleanupAlarmsAgent.groovy
```diff
@@ -86,36 +86,38 @@ class CleanupAlarmsAgent implements RunnableAgent, CustomScheduledAgent {
...
-            describeAlarmsRequest.withNextToken(result.nextToken)
-          } else {
-            break
+            if (result.nextToken) {
+              describeAlarmsRequest.withNextToken(result.nextToken)
+            } else {
+              break
+            }
           }
+        } catch (Exception e) {
```
dbyron-sf, line 119: Is there a catch block to add "higher up" (where the run method is called) that would (also) help?  Like, what if there exceptions in other agents?

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/agent/CleanupAlarmsAgent.groovy
christosarvanitis: Bad commit on my end. Will revert this

clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/agent/CleanupAlarmsAgent.groovy
christosarvanitis: Higher up there is already a try/catch block in the `RunnableAgent ` initial implementation  but i have pushed a try/catch block in the `CleanupDetachedInstancesAgent` as well which implements a RunnableAgent

dbyron-sf: Seems like `run` is called from [RunnableAgentExecution.executeAgent](https://github.com/spinnaker/clouddriver/blob/f5f4b112f7b8a5d97127b91b859277fab99a6a59/cats/cats-core/src/main/java/com/netflix/spinnaker/cats/agent/RunnableAgent.java#L32) and all the places that call `executeAgent` are similar to [DefaultAgentScheduler](https://github.com/spinnaker/clouddriver/blob/f5f4b112f7b8a5d97127b91b859277fab99a6a59/cats/cats-core/src/main/java/com/netflix/spinnaker/cats/agent/DefaultAgentScheduler.java#L134).  There's a try/catch, but it only updates metrics.  So what we get here is some extra logging, but I don't see how it's going to help with agent scheduling.

christosarvanitis: This cleanup agents are going through each account and trying to do a cleanup of any stale alarms in cloudwatch or any detached instances etc. When a bad account is present in the credentials repository the agent stops and doesnt go through the rest of the accounts.
By adding a try/catch here we are logging that there was a problem with X account and continue to the next account.

dbyron-sf: Aaah yes, the try/catch here and in CleanupDetachedInstancesAgent is inside `getAccounts().each`.  Seems like the consequences of this fix are good...that instead of dying on the first error, we'll continue to clean up other accounts.

dbyron-sf (approved):

dbyron-sf:
@christosarvanitis if you bring this branch up to date, I think it's ready to go.

christosarvanitis:
> @christosarvanitis if you bring this branch up to date, I think it's ready to go.

Done! Thanks @dbyron-sf

PR #6334, opened 2025-01-17
fix(google): Add partner metadata on instanceTemplate properties

edgarulg:
When cloning a server group using the UI. The resourceManagerTags and partnerMetadata are not populated even if the deployment is configured with resourceManagerTags and partnerMetadata.

In clouddriver we include the resourceManagerTags under the instanceTemplate properties but by default the partnerMetadata is not included. See https://cloud.google.com/sdk/gcloud/reference/beta/compute/instance-templates/list

The fix only set the view option to "FULL". This ensure the partnerMetadata is included in the instanceTemplate properties.

Deck PR: https://github.com/spinnaker/deck/pull/10161

### DIFF
--- a/clouddriver-google/src/main/java/com/netflix/spinnaker/clouddriver/google/compute/InstanceTemplates.java
+++ b/clouddriver-google/src/main/java/com/netflix/spinnaker/clouddriver/google/compute/InstanceTemplates.java
@@ -29,6 +29,8 @@ public class InstanceTemplates {
   private final Compute.InstanceTemplates computeApi;
   private final GoogleNamedAccountCredentials credentials;
   private final GlobalGoogleComputeRequestFactory requestFactory;
+  private static final String defaultView =
+      "FULL"; // https://cloud.google.com/sdk/gcloud/reference/beta/compute/instance-templates/list
 
   InstanceTemplates(
       GoogleNamedAccountCredentials credentials,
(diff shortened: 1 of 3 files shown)


clouddriver-google/src/main/java/com/netflix/spinnaker/clouddriver/google/compute/InstanceTemplates.java
```diff
@@ -61,11 +61,13 @@ public GoogleComputeOperationRequest<Compute.InstanceTemplates.Insert> insert(
     return requestFactory.wrapOperationRequest(request, "insert");
   }
 
-  public PaginatedComputeRequest<Compute.InstanceTemplates.List, InstanceTemplate> list() {
+  public PaginatedComputeRequest<Compute.InstanceTemplates.List, InstanceTemplate> list(
+      String view) {
```
jasonmcintosh, line 65: Is this api used anywhere else?  I see the one spot below in the abstract caching agent that passes a fixed value of "FULL", but no place else this list call is updated?

jasonmcintosh: IF it's only ever called in that ONE operation - why not just add that as a fixed value here?

edgarulg: In the clouddriver-google module there is only one call for this API. I wanted to leave it flexible but as it is only one call it's pointless. I will make the "FULL" view as the default one.

jasonmcintosh (approved):

edgarulg:
@Mergify backport release-1.36.x

PR #6344, opened 2025-02-12
feat(aws): configurable duration of assume role sessions

dbyron-sf:
via new per-account configuration property `sessionDurationSeconds` and
```
aws:
  defaultSessionDurationSeconds: (no default value)
```
if neither aws.defaultSessionDurationSeconds is set nor a per-account sessionDurationSeconds, the behavior is the same as before.  The duration of the assume role session is the default provided by the aws sdk.  In version 1.12.261, that's [900 seconds](https://github.com/aws/aws-sdk-java/blob/1.12.261/aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSAssumeRoleSessionCredentialsProvider.java#L48).

A bit of refactoring to use lombok in some places snuck in here as well.


zachsmith1 (approved):

jasonmcintosh:
NICE:  It'd not be bad to issue a CAUTION:  Role session duration... on transitive roles is limited to 1 hour.  On base role sessions, it can be up to 12 hours. I'd go with the 1 hour normally.

dbyron-sf:
> NICE: It'd not be bad to issue a CAUTION: Role session duration... on transitive roles is limited to 1 hour. On base role sessions, it can be up to 12 hours. I'd go with the 1 hour normally.

Any chance you can link to some docs about this?

jasonmcintosh:
Docs on session duration (defaults): https://docs.aws.amazon.com/singlesignon/latest/userguide/howtosessionduration.html#:~:text=The%20minimum%20session%20duration%20is,session%20duration%20of%2012%20hours.

Doc on chaining:
https://repost.aws/knowledge-center/iam-role-chaining-limit

jasonmcintosh:
NOTE too internally there's a "cache" on resources that I think is 15 minutes, so JUST setting session duration MAY NOT impact that cache of the various SDKs (will check).  I'll have to find this - i hit it with the lambda timeouts

jasonmcintosh:
https://github.com/armory-io/clouddriver/blob/master/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/AwsSdkClientSupplier.java#L67 See this.

dbyron-sf:
> https://github.com/armory-io/clouddriver/blob/master/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/AwsSdkClientSupplier.java#L67 See this.

I believe this is one level down.  As in, this is where caching agents, etc. get aws sdk clients, but I don't believe the assume role code uses this.

PR #6347, opened 2025-02-13
fix(build): Uses a hard coded qemu for builds due to a bug in latest …

jasonmcintosh:


jasonmcintosh:
See https://github.com/docker/setup-qemu-action/issues/198 - Testing with older version of QEMU

jasonmcintosh:
https://github.com/tonistiigi/binfmt/issues/240 FYI

PR #6351, opened 2025-02-26
fix(clouddriver-google): Fix logic when Instance Types are available only in selected Zones/Regions

christosarvanitis:
In GCP there are some instance types (like c4-highcpu-X) that are available only in specific Zones per Region (or all of the zones in some other regions). 

- When the user selects a Regional deployment (Distribute instances across multiple zones) and also selects an instance type that is available only in specific zones in that region: 
[Before the fix]: Clouddriver was failing with an error message Instance type not found
[After the fix]: If the instance type selected is available in all the region then it succeeds. If the instance type is not available in all the zones within the region it fails with the appropriate message

-  When the user selects a Regional deployment (Distribute instances across multiple zones) and Selects the Zones explicitly + also selects an instance type that is available only in specific zones in that region:
[Before the fix]: Clouddriver was failing with an error message Instance type not found
[After the fix]: If the user selects only the Zones that the instance type is available then it succeeds. If the user selects any Zone that the instance type is not available then it fails with the appropriate message.

### DIFF
--- a/clouddriver-google/src/main/java/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.java
+++ b/clouddriver-google/src/main/java/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.java
@@ -56,6 +56,7 @@ import com.netflix.spinnaker.clouddriver.google.deploy.GCEUtil;
 import com.netflix.spinnaker.clouddriver.google.deploy.GoogleOperationPoller;
 import com.netflix.spinnaker.clouddriver.google.deploy.SafeRetry;
 import com.netflix.spinnaker.clouddriver.google.deploy.description.BasicGoogleDeployDescription;
+import com.netflix.spinnaker.clouddriver.google.deploy.exception.GoogleOperationException;
 import com.netflix.spinnaker.clouddriver.google.deploy.ops.GoogleUserDataProvider;
 import com.netflix.spinnaker.clouddriver.google.model.GoogleHealthCheck;
 import com.netflix.spinnaker.clouddriver.google.model.GoogleLabeledResource;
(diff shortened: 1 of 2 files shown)


edgarulg (approved):
Thanks for working on this fix. Basically the same behavior is kept but it is returning the appropriate error message based on the zone's availability for the machine. Overall looks good to me.

We need to backport this change because it will help GCE Spinnaker Users to know why their deployment failed but older versions still has the groovy implementation. We can make a manual backport if everyone agree.

clouddriver-google/src/main/java/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.java
```diff
@@ -289,8 +290,45 @@ protected String getMachineTypeNameFromInput(
     if (description.getInstanceType().contains("custom-")) {
       return description.getInstanceType();
     } else {
-      return GCEUtil.queryMachineType(
-          description.getInstanceType(), location, description.getCredentials(), task, BASE_PHASE);
+      List<String> queryZone =
+          description.getRegional()
+              ? (description.getSelectZones()
+                  ? description.getDistributionPolicy().getZones()
```
jasonmcintosh, line 296: CAN distribution policy be null?  IF SO this would potentially introduce an NPE...

jasonmcintosh:
I'd welcome a backport fix on this :)

clouddriver-google/src/main/java/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.java
christosarvanitis: Ideally Deck protects us from this since it is failing the validations but i can change to: 
```
      List<String> queryZone =
          description.getRegional()
              ? (description.getSelectZones() && !description.getDistributionPolicy().getZones().isEmpty()
                  ? description.getDistributionPolicy().getZones()
                  : Collections.singletonList(location))
              : Collections.singletonList(location);
```
Which basically is pipeline definition: 
```
      "distributionPolicy": {
        "targetShape": "EVEN",
        "zones": []
      },
      "regional": true,
      "selectZones": true,
```

christosarvanitis: Updated

jasonmcintosh (approved):

dbyron-sf:
> I'd welcome a backport fix on this :)

Looks like we need a 1.37.1 as soon as we sort out some retrofit issues.  I'm OK with this going there too.  We're already behind schedule on releasing 1.36.next, but I suppose this could go there too.  How far back are you thinking?

christosarvanitis:
Since this part of code was refactored heavily for 1.37.x i created this manual backport for groovy to go on 1.36.x https://github.com/spinnaker/clouddriver/pull/6352

dbyron-sf:
@mergifyio backport release-1.37.x

PR #6352, opened 2025-02-26
fix(clouddriver-google): Fix logic when Instance Types are available only in selected Zones/Regions - manual backport

christosarvanitis:
Implementation pre-refactoring of https://github.com/spinnaker/clouddriver/pull/6351 to release-1.36.x

In GCP there are some instance types (like c4-highcpu-X) that are available only in specific Zones per Region (or all of the zones in some other regions).

- When the user selects a Regional deployment (Distribute instances across multiple zones) and also selects an instance type that is available only in specific zones in that region:
[Before the fix]: Clouddriver was failing with an error message Instance type not found
[After the fix]: If the instance type selected is available in all the region then it succeeds. If the instance type is not available in all the zones within the region it fails with the appropriate message

- When the user selects a Regional deployment (Distribute instances across multiple zones) and Selects the Zones explicitly + also selects an instance type that is available only in specific zones in that region:
[Before the fix]: Clouddriver was failing with an error message Instance type not found
[After the fix]: If the user selects only the Zones that the instance type is available then it succeeds. If the user selects any Zone that the instance type is not available then it fails with the appropriate message.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
@@ -177,6 +177,7 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
     if (description.instanceType.contains('custom-')) {
       machineTypeName = description.instanceType
     } else {
+      // This is to handle when resources aren't in all zones but only some and this causes failures on deploys.
       def queryZone = description.regional ? (description.selectZones ? description.distributionPolicy.getZones() : [location]) : [location]
       queryZone.each { zoneOrLocation ->
         def msg = description.regional ?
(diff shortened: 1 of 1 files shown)


jasonmcintosh:
SO I hate but... can we backport the test too please?

christosarvanitis:
> SO I hate but... can we backport the test too please?

christosarvanitis:
> SO I hate but... can we backport the test too please?

I dont see a way to do this. The refactoring that happened in 1.37.x was mainly due to this groovy class was not testable and we didnt have any test coverage. So adding a test here is like trying to refactor again

radu-dragon (approved):

jasonmcintosh (approved):

jasonmcintosh:
REALLY want builds to run though first on this (and ideally it-test)

jasonmcintosh:
https://github.com/spinnaker/clouddriver/actions/runs/13548922324/job/37867472528 hrmm that's odd I thinkt hey did run but not seeing it in the checks...

jasonmcintosh:
Actually that run is for other.  Need to figure out why PR build isn't running here...

christosarvanitis:
> Actually that run is for other. Need to figure out why PR build isn't running here...

Probably because im raising a PR against a release branch and only release managers + mergify is allowed to do that.

PR #6359, opened 2025-03-18
fix(docker): resolve the issue of `Unrecognized field "child"`

kirangodishala:
This PR addresses https://github.com/spinnaker/spinnaker/issues/7019

Google cloud docker artifact registry responds with additional fields when tags list api is invoked. It used to work with retrofit1 as Gson converter ignored the additional fields while deserializing by default.  After upgrading to retrofit2, the ObjectMapper used for deserializing  was not configured to ignore additional fields. This PR includes the required configuration.


dbyron-sf:
@mergifyio backport release-1.37.x

PR #6379, opened 2025-07-31
fix(google): Add retry and status polling logic

mrusan:
The fix addresses an issue we've encountered, where during red-black deployments subsequent steps could execute before autoscaler creation finishes, causing inconsistent behavior and potential failures on delivery.

Now, the async operations use the GoogleOperationPoller which implements proper retry logic and handles operation status polling.

### DIFF
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
@@ -65,6 +65,7 @@ import com.netflix.spinnaker.moniker.Namer
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
+import javax.annotation.PostConstruct
 
 import static com.google.common.base.Preconditions.checkArgument
 import static com.netflix.spinnaker.clouddriver.google.deploy.GCEUtil.BACKEND_SERVICE_NAMES
@@ -612,10 +613,6 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
           // Wait for regional autoscaler creation to complete before proceeding with deployment
           // Uses GoogleOperationPoller which implements proper retry logic and handles operation status polling
           if (googleDeployDefaults.enableAsyncOperationWait) {
-            log.warn(
-              "[enableAsyncOperationWait]: If you see unjustified long waits or other issues caused by this flag, " +
-              "please drop a note in Spinnaker Slack or open a GitHub Issue with the related details."
-            )
             googleOperationPoller.waitForRegionalOperation(compute, project, region, autoscalerOperation.getName(),
               null, task, "regional autoscaler $serverGroupName", BASE_PHASE)
           }
@@ -658,10 +655,6 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
           // Wait for zonal autoscaler creation to complete before proceeding with deployment
           // Uses GoogleOperationPoller which implements proper retry logic and handles operation status polling
           if (googleDeployDefaults.enableAsyncOperationWait) {
-            log.warn(
-              "[enableAsyncOperationWait]: If you see unjustified long waits or other issues caused by this flag, " +
-              "please drop a note in Spinnaker Slack or open a GitHub Issue with the related details."
-            )
             googleOperationPoller.waitForZonalOperation(compute, project, zone, autoscalerOperation.getName(),
               null, task, "autoscaler $serverGroupName", BASE_PHASE)
           }
@@ -850,6 +843,14 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
       description, credentials, customUserData)
     task.updateStatus BASE_PHASE, "Resolved user data."
     return userData
+  }
+
+  @PostConstruct
+  void logEnableAsyncOperationWaitWarning() {
+    if (googleDeployDefaults?.enableAsyncOperationWait) {
+      log.warn("[enableAsyncOperationWait]: If you see unjustified long waits or other issues caused by this flag, " +
+               "please drop a note in Spinnaker Slack or open a GitHub Issue with the related details.")
+    }
   }
 
   static class GoogleInstanceTemplate implements GoogleLabeledResource {
--- a/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
+++ b/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
@@ -29,6 +29,7 @@ import com.google.api.services.compute.model.InstanceGroupManagerAutoHealingPoli
 import com.google.api.services.compute.model.InstanceProperties
 import com.google.api.services.compute.model.InstanceTemplate
 import com.google.api.services.compute.model.NamedPort
+import com.google.api.services.compute.model.Operation
 import com.netflix.frigga.Names
 import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.cats.cache.Cache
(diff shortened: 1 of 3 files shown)


clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -587,14 +588,31 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
...
+          // Google Cloud autoscaler insert operations are asynchronous and return an Operation object.
+          // 
+          // Per Google Cloud documentation:
+          // "When you perform any requests that modify data, a zoneOperations or regionOperations resource 
+          // is returned, and you can query the operation to check the status of your change."
+          //
+          // Without waiting for autoscaler creation to complete, subsequent deployment steps (health checks, traffic routing)
+          // may execute before the autoscaler is active, leading to inconsistent behavior and potential deployment failures.
+          //
+          // This fix aligns Spinnaker GCP behavior with Spinnaker AWS behavior, where autoscaling group operations are synchronous.
```
dbyron-sf, line 607: This feels like a big enough change that it warrants a feature flag.

jasonmcintosh: SO this is a bug, not a feature.  The current implementation gets an async response.  Instead of polling for the state of that response, it just accepts if it doesn't error & continues.  Turns out... there are cases where that response is a PENDING sorta state that causes failures UNLESS you poll for completion.

jasonmcintosh: The comment isn't exactly correct on the internals of where this fails...

dbyron-sf: Right, but...how many folks are depending on the current behavior?  Or, what if this actually makes things worse (at least temporarily, while we work out the kinks)?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -611,46 +629,102 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
...
           "Load balancer backend service",
           task,
-          [400, 412],
+          [400, 412],  // Retry on Bad Request (400) and Precondition Failed (412)
           [],
           [action: "update", phase: BASE_PHASE, operation: "updateBackendServices", (TAG_SCOPE): SCOPE_GLOBAL],
           registry
         )
+        
+        if (operation) {
```
dbyron-sf, line 681: Under what circumstances is operation null?  Is it worth logging a message or emitting a metric?

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
jasonmcintosh: SO this LITERALLY can cause failures without this polling.  https://github.com/spinnaker/spinnaker/issues/7170 ssee for this

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
jasonmcintosh: The previous invocation would return null always.  Since polling for the state of the operation (i it's there) requires passing oepration information, we have to check if that's null before using the operation.  This null check wasn't needed previously because we weren't polling for state previously.  https://github.com/spinnaker/clouddriver/blob/aa0c51ae74351bd3ae0ce9949212bb5ac5ae073c/clouddriver-google/src/main/java/com/netflix/spinnaker/clouddriver/google/deploy/SafeRetry.java#L89 and below where null was always returned.

dbyron-sf: I see....Wouldn't the world be a better place if SafeRetry.doRetry let the exception bubble up?  Seems like we're effectively swallowing them which doesn't seem great.

jasonmcintosh: Yes I agree ;) I think that's another area that could use some improvement ... probably separate from this.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
dbyron-sf: Yeah, it's a tough call.  I bet there are some setups that are gonna start to fail when this behavior changes -- pipelines are gonna take longer to complete than people expect....even if what people expect is "wrong."  In the spirit of reducing blast radius I think a feature flag is good idea, but I won't hold up the PR for it.

jasonmcintosh: Let's look at FF.  My only concern on FF is...that doing this as a flag may be MORE complicated/risky than without.  That said, I understand the concerns!

mattgogerly: Potential middle ground for a feature flag to opt-out rather than in, for people running into issues? With a request to post in Slack if needing to turn it on. If nobody messages after 1/2 releases, rip off the bandaid?

dbyron-sf: Yeah, I'm good with having the feature flag enabled by default.

mrusan: @mattgogerly @dbyron-sf @jasonmcintosh thank you for reviewing the code and your input - I've added a `enableAsyncOperationWait` flag, enabled by default, along with a log message - haven't found other places where we log such messages, so let me know if there's a better example

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -611,46 +635,108 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
...
+          //
+          // This fix aligns Spinnaker GCP behavior with Spinnaker AWS behavior, where autoscaling group operations are synchronous.
+          Operation autoscalerOperation = timeExecute(compute.autoscalers().insert(project, zone, autoscaler),
                       "compute.autoscalers.insert",
                       TAG_SCOPE, SCOPE_ZONAL, TAG_ZONE, zone)
+          
+          // Wait for zonal autoscaler creation to complete before proceeding with deployment
+          // Uses GoogleOperationPoller which implements proper retry logic and handles operation status polling
+          if (googleDeployDefaults.enableAsyncOperationWait) {
+            log.warn(
```
jasonmcintosh: I'd NOT do a warn.  I'd do a DEBUG.  OR just a log message when INITIALLY set as a WARN.

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
```diff
@@ -587,14 +588,37 @@ class BasicGoogleDeployHandler implements DeployHandler<BasicGoogleDeployDescrip
...
+          // This fix aligns Spinnaker GCP behavior with Spinnaker AWS behavior, where autoscaling group operations are synchronous.
+          Operation autoscalerOperation = timeExecute(
               compute.regionAutoscalers().insert(project, region, autoscaler),
               "compute.regionAutoscalers.insert",
               TAG_SCOPE, SCOPE_REGIONAL, TAG_REGION, region)
+          
+          // Wait for regional autoscaler creation to complete before proceeding with deployment
+          // Uses GoogleOperationPoller which implements proper retry logic and handles operation status polling
+          if (googleDeployDefaults.enableAsyncOperationWait) {
+            log.warn(
```
jasonmcintosh: Same as other - not warn on all operations please!

clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.groovy
mrusan: Oh, certainly. I wasn't sure _how_ verbose this is supposed to be for the users to notice. Done

jasonmcintosh:
FYI I'm i think good with this... so if others are good, think we can get this merged.  We're fixing GCP buidls in master now which are currently a bit broken :(

jasonmcintosh (approved):

dbyron-sf (approved):

PR #6380, opened 2025-08-07
fix(google): Add retry and status polling logic

mrusan:
Addresses https://github.com/spinnaker/spinnaker/issues/7170 similarly to https://github.com/spinnaker/clouddriver/pull/6379 but for the master branch code, which was refactored to Java.

The `enableAsyncOperationWait` flag was omitted for this PR, because part of the operations were already using `googleOperationPoller` correctly, and we're just making the behaviour consistent.

Also covers a couple of regression bugs discovered during implementation and testing.


mrusan:
Closing in favor of https://github.com/spinnaker/spinnaker/pull/7191

